PR #400, opened 2016-12-21
feat(widget-message-meet): add demo package for data entry

adamweeks:


packages/widget-message-meet/README.md
```diff
@@ -96,15 +96,15 @@ The easiest way to get the Spark Message and Meet Widget into your web site is t
   -  Add a `<link />` tag to include `main.css`
 1.  Create a container where you would like to embed the widget and add the following attributes to configure the widget:
   - `data-toggle="spark-message-meet"`: (required)
-  - `data-access-token`: (required) Access token for the user account initiating the messaging session. For testing purposes you can use a developer access token from <https://developers.ciscospark.com>.
```
adamweeks: Need to add this back.

packages/widget-message-meet/src/components/widget-data-input/index.js
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable react/no-set-state, global-require */
```
adamweeks: Need to remove this file and test, not used any more.

packages/widget-message-meet/src/app.js
```diff
@@ -24,7 +25,7 @@ function loadAllWidgets() {
   }
 }
 
-loadAllWidgets();
+document.addEventListener(`DOMContentLoaded`, loadAllWidgets, false);
```
adamweeks: Added this because when using the widget on codepen, it would attempt to load before it was ready.

adamweeks:
Not quite ready yet, but figured I'd give a starting point.

adamweeks:
Ok, now it is ready for round one!

packages/widget-message-meet/src/components/loading-screen/index.js
```diff
@@ -3,12 +3,15 @@ import classNames from 'classnames';
...
+import SparkLogo from '../spark-logo';
 import styles from './styles.css';
 
 export default function LoadingScreen() {
   return (
     <div className={classNames(`loading`, styles.loading)}>
-      <div className={classNames(`spark-logo`, styles.logo)} />
+      <div className={classNames(`spark-logo`, styles.logo)}>
+        <SparkLogo />
+      </div>
```
bzang: This and the removal of the CSS styles looks like a regression. We can't actually have an img tag right now because of how the CDN loads files. HTML uses the HTML location to resolve the `src` field, while CSS uses the location of the CSS file.

packages/widget-message-meet/src/components/spark-logo/index.js
```diff
@@ -0,0 +1,16 @@
...
+import classNames from 'classnames';
+import styles from './styles.css';
+
+import logo from './logo-spark.png';
+
+// eslint-disable-next-line func-style,arrow-body-style
+const SparkLogo = () => {
+  return (
+    <div className={classNames(`spark-logo`, styles.logo)}>
+      <img role="presentation" src={logo} />
```
bzang: See above. Can't use an img tag for this. Images won't load correctly from CDN

packages/widget-message-meet/src/demo/demo.js
```diff
@@ -0,0 +1,135 @@
+/* eslint-disable react/no-set-state, global-require */
```
bzang: What's disable `global-require` for?

packages/widget-message-meet/src/components/spark-logo/index.js
adamweeks: Good call. We should still be able to make this a component though right? Just have the proper css values?

packages/widget-message-meet/src/demo/demo.js
adamweeks: Oh, I was using your code for the background image at one point, but since I'm not now, I'll remove the rule disabling.

packages/widget-message-meet/src/components/loading-screen/index.js
adamweeks: I wasn't aware of the CDN issue. Will change the component to use the css method.

adamweeks: @bzang should be fixed now

PR #402, opened 2017-01-03
test(widget-message-meet): fix mock spark implementation

adamweeks:
Seems to be some change in the way jest mocks are implemented


ianwremmel:
@bzang, can you reread your comment? :)

PR #404, opened 2017-01-09
refactor(widget-message-meet) add immutable conversation to store

adamweeks:


packages/widget-message-meet/src/components/activity-list/index.js
```diff
@@ -1,64 +1,54 @@
...
 import classNames from 'classnames';
-import moment from 'moment';
 import ActivityItem from '../activity-item';
 import DaySeparator from '../day-separator';
 import NewMessagesSeparator from '../new-messages-separator';
 import formatDate from '../../utils/date';
 
+export const ACTIVITY_ITEM_TYPE_ACTIVITY_ITEM = `ACTIVITY_ITEM_TYPE_ACTIVITY_ITEM`;
+export const ACTIVITY_ITEM_TYPE_DAY_SEPARATOR = `ACTIVITY_ITEM_TYPE_DAY_SEPARATOR`;
+export const ACTIVITY_ITEM_TYPE_NEW_MESSAGE_SEPARATOR = `ACTIVITY_ITEM_TYPE_NEW_MESSAGE_SEPARATOR`;
```
bzang: Didn't we say we were gonna change the names of these to something slightly different?

packages/widget-message-meet/src/containers/connected-activity-list/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,4 @@
+exports[`ConnectedActivityList renders properly 1`] = `
+<div
+  className="activity-list" />
+`;
```
bzang: Shouldn't this snapshot still have activities? Or does the mock store not have any?

packages/widget-message-meet/src/containers/scrolling-activity/__snapshots__/index.test.js.snap
```diff
@@ -4,200 +4,7 @@ exports[`ScrollingActivity container renders properly 1`] = `
...
-            <div
-              className="activity-post-action">
-              <div
-                className="action-spacer" />
-            </div>
-          </div>
-        </div>
-      </div>
-    </div>
+      className="activity-list" />
```
bzang: Looks like the activities are missing here as well. Is this because of a shallow render or just that the selector isn't firing correctly?

packages/widget-message-meet/src/components/activity-list/index.js
adamweeks: Yep, must be in the other branch, I'll pull those in.

packages/widget-message-meet/src/containers/connected-activity-list/__snapshots__/index.test.js.snap
adamweeks: The mock store doesn't have any. I updated the actual activity list to have proper data in its tests. I wasn't sure about building out a fully populated store, but we can.

packages/widget-message-meet/src/containers/scrolling-activity/__snapshots__/index.test.js.snap
adamweeks: The selector is firing, it just is in an empty state as is.

adamweeks:
@bzang please re-review

PR #408, opened 2017-01-11
feature(CI-page): use cisService instead of service

aimeex3:
this updates the CI login screen with the spark theme


packages/plugin-credentials/src/config.js
```diff
@@ -15,7 +15,7 @@ export default {
     // eslint-disable-next-line camelcase
     redirect_uri: process.env.CISCOSPARK_REDIRECT_URI,
     scope: process.env.CISCOSPARK_SCOPE,
-    service: `spark`,
+    cisService: `spark`,
```
ianwremmel: After thinking about this a bit, can we keep the config property the unchanged and just update the name of the url parameter? that way it doesn't introduce a breaking change?

PR #413, opened 2017-01-17
Jenkinsfile

ianwremmel:


Jenkinsfile
```diff
@@ -0,0 +1,409 @@
...
+  if (env.SDK_BUILD_DEBUG != null) {
+    dockerEnv+="SDK_BUILD_DEBUG=${env.SDK_BUILD_DEBUG}\n"
+  }
+  if (env.SKIP_FLAKY_TESTS != null) {
+    dockerEnv+="SKIP_FLAKY_TESTS=${env.SKIP_FLAKY_TESTS}\n"
+  }
+  if (env.WDM_SERVICE_URL != null) {
+    dockerEnv+="WDM_SERVICE_URL=${env.WDM_SERVICE_URL}\n"
+  }
+  if (env.WORKSPACE != null) {
```
adamweeks: We'll need to add our new env vars if my merge ever goes through

Gruntfile.js
```diff
@@ -146,7 +148,11 @@ module.exports = function(grunt) {
         '<%= config.src %>/**/*.js',
         '<%= config.test %>/**/*.js',
         'Gruntfile.js'
-      ]
+      ],
+      options: {
+        format: process.env.XUNIT ? 'checkstyle' : 'stylish',
+        outputFile: process.env.XUNIT && 'reports/style/eslint-legacy.xml'
```
bzang: `eslint-legacy`?

Gruntfile.js
```diff
@@ -353,12 +359,6 @@ module.exports = function(grunt) {
...
   if (XUNIT) {
-    var eslint = grunt.config('eslint');
-    eslint.options = eslint.options || {};
-    eslint.options.format = 'junit';
-    eslint.options.outputFile = 'reports/eslint.xml';
-    grunt.config('eslint', eslint);
-
     var jscs = grunt.config('jscs');
     jscs.options.reporter = 'junit';
     jscs.options.reporterOutput = 'reports/jscs.xml';
```
bzang: Should we remove `jscs` at some point? I believe the main project is deprecated and they've merged with eslint.

Jenkinsfile
```diff
@@ -0,0 +1,409 @@
...
+def IS_VALIDATED_MERGE_BUILD = false
+def GIT_COMMIT
+
+def cleanup = { ->
+  archive 'reports/**/*'
+  sh 'rm -f .env'
+
+  if (IS_VALIDATED_MERGE_BUILD && currentBuild.result != 'SUCCESS') {
+    withCredentials([usernamePassword(
+      credentialsId: '386d3445-b855-40e4-999a-dc5801336a69',
```
bzang: Is it ok to have credentials hard coded in here? Should we be using the credentials jenkins plugin (if we have it)?

Jenkinsfile
```diff
@@ -0,0 +1,409 @@
...
+          DOCKER_RUN_OPTS = "${DOCKER_RUN_OPTS} -e NPM_CONFIG_CACHE=${env.WORKSPACE}/.npm"
+          DOCKER_RUN_OPTS = "${DOCKER_RUN_OPTS} --volumes-from=\$(hostname)"
+          DOCKER_RUN_OPTS = "${DOCKER_RUN_OPTS} --user=\$(id -u):\$(id -g)"
+          // DOCKER_RUN_OPTS has some values in it that we want to evaluate on
+          // the node, but image.inside doesn't do subshell execution. We'll use
+          // echo to evaluate once on the node and store the values.
+          DOCKER_RUN_OPTS = sh script: "echo -n ${DOCKER_RUN_OPTS}", returnStdout: true
+          // image.inside uses the -d flag, so we can only use --rm for
+          // bash-started containers
+          env.DOCKER_RUN_OPTS = "${DOCKER_RUN_OPTS} --rm ${DOCKER_IMAGE_NAME}"
```
bzang: Do we have access to docker compose on the boxes? It would probably simplify this if we can just use that to coordinate and spin up containers.

Jenkinsfile
ianwremmel: What are those vars? I'll add them.

Gruntfile.js
ianwremmel: for the deprecated src and test directories

Gruntfile.js
ianwremmel: yes. once the web client stops using the legacy sdk, at least two gruntfiles should go away. I'm pretty sure jscs doesn't get run any more, but since a bunch of stuff should get deleted in the next month or two, i didn't see the need to rock the boat now.

Jenkinsfile
adamweeks: `MESSAGE_DEMO_CLIENT_ID`
`MESSAGE_DEMO_CLIENT_SECRET`

Jenkinsfile
ianwremmel: this is the jenkins credentials plugin. that uuid says which credential to pull from the jenkins credentials database.

Jenkinsfile
ianwremmel: we do, but how would it help? the compose file would only reference once image.

ianwremmel: Also, I'm happy to refine the docker stuff later, but right now, the goal is switch to Jenkinsfile while changing as few other things as possible.

Jenkinsfile
bzang: Ahhhh go it.

Jenkinsfile
bzang: Got ya... I was thinking it'd be easier to provide all the settings in a dockercompose file so that all you would need to do is docker-compose up and it'll bring everything online. But makes sense right now to change as little as possible.

PR #416, opened 2017-01-18
feat(plugin-metrics): add support for new clientmetrics endpoints

Cruig:
Support user identity aliasing and posting pre-login metrics via Spark's Metrics Service


packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+  alias: function(preLoginId) {
+    return this.request({
+      method: `POST`,
+      api: `metrics`,
+      resource: `clientmetrics`,
+      headers: {
+        "X-Prelogin-UserId": preLoginId
+      },
+      body: {},
+      params: [
```
aimeex3: whats the difference from body and params?

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+          value: true
+        }
+      ]
+    });
+  },
+
+  postPreLoginMetric: function(payload, preLoginId) {
+    return this.request({
+      method: `POST`,
+      url: 'https://metrics-a.wbx2.com/metrics/api/v1/clientmetrics-prelogin',
```
aimeex3: would api: `metrics`, and resource: `clientmetrics-prelogin` work here?

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+      headers: {
+        "X-Prelogin-UserId": preLoginId,
+        "content-type": "application/json",
```
aimeex3: does json: true automatically add content-type: application/json to the header?

aimeex3: also should these use backticks instead of quotes?

src/client/services/metrics/metrics.js
```diff
@@ -34,13 +34,49 @@ var MetricsService = SparkBase.extend(
...
+            value: true
+          }
+        ]
+      });
+    },
+
+    postPreLoginMetric: function(payload, preLoginId) {
+      return this.request({
+        method: `POST`,
+        url: 'https://metrics-a.wbx2.com/metrics/api/v1/clientmetrics-prelogin',
```
aimeex3: same comments as above

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
     return this.batcher.request(Object.assign({key}, value));
   },
 
-  submitClientMetrics(eventName, props) {
+  /**
+   * This corresponds to #sendSemiStructured() in the deprecated metrics handler
```
aimeex3: @saurjai3 just giving you a heads up. once carter's PR goes in, the method in the new sdk will be called `submitClientMetrics` instead of `sendSemiStructured()`

aimeex3:
there are also a lot of lint errors :)

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+   * @param eventName
+   * @param props
+   * @param preLoginId
+   */
+  submitClientMetrics(eventName, props, preLoginId) {
```
ianwremmel: This should be singular. It's also slightly redundant. Any way we can avoid saying metrics twice in `spark.metrics.submitClientMetrics`?

packages/plugin-metrics/src/metrics.js
ianwremmel: yea. we should never hardcode full urls.

packages/plugin-metrics/src/metrics.js
ianwremmel: left side should be apostrophes, right side shoul dbe back ticks.

the content type is unnecessary because of `json:true` (set by default in http-core). 

@Cruig, is `x-prelogin-userid` in the cloudapps list of cors-allowed headers? 

Please make the header name lowercase.

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+  },
+
+  postPreLoginMetric: function(payload, preLoginId) {
+    return this.request({
+      method: `POST`,
+      url: 'https://metrics-a.wbx2.com/metrics/api/v1/clientmetrics-prelogin',
+      headers: {
+        "X-Prelogin-UserId": preLoginId,
+        "content-type": "application/json",
+        "charset": "utf-8"
```
ianwremmel: This should be implied and if it's not, it should be in http-core, not a single request.

packages/plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -54,12 +86,12 @@ describe(`plugin-metrics`, () => {
     });
 
     describe(`#submitClientMetrics()`, () => {
-      it(`submits a metric to clientmetrics`, () => {
+      it(`submits a metric to clientmetrics if after auth`, () => {
```
ianwremmel: add a second set of describe nesting here: describe(`before login`), describe(`after login`)

packages/plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -74,7 +106,46 @@ describe(`plugin-metrics`, () => {
...
       });
+      it('posts pre-login metric if before auth', function() {
+        metrics.submitClientMetrics(eventName, mockPayload, preLoginId);
+        assert.calledWith(metrics.postPreLoginMetric, preLoginProps, preLoginId);
+      });
+
+    });
+
+    describe('#postPreLoginMetric()', function() {
+      it('should return request', function() {
```
ianwremmel: Don't use `should` in test statements.

```javascript
describe('#postPreLoginMetric()', function() {
  it('returns an HttpResponse object', function() {
```

packages/plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -74,7 +106,46 @@ describe(`plugin-metrics`, () => {
...
+
+            assert.equal(metric.metricName, eventName);
+            assert.equal(metric.tags.testTag, `tag value`);
+            assert.equal(metric.fields.testField, 123);
+          });
+      });
+    });
 
+    describe('#alias()', function() {
+      it('should return request', function() {
```
ianwremmel: don't use should

src/client/services/metrics/metrics.js
```diff
@@ -34,13 +34,49 @@ var MetricsService = SparkBase.extend(
...
+        },
+        body: payload
+      });
+    },
+
   /**
    * Submits semi-structured metrics
    * @param {string} eventName
-   * @param {Object} props
+   * @param {object} props
```
ianwremmel: Object should be capitalized

src/client/services/metrics/metrics.js
```diff
@@ -34,13 +34,49 @@ var MetricsService = SparkBase.extend(
...
+   * @param {string} [preLoginId]
    */
-  sendSemiStructured: function sendSemiStructured(eventName, props) {
-    var payload = {
+  sendSemiStructured: function sendSemiStructured(eventName, props, preLoginId) {
+    let payload = {
```
ianwremmel: don't use `let` in legacy sdk

packages/plugin-metrics/src/metrics.js
```diff
@@ -26,16 +26,75 @@ const Metrics = SparkPlugin.extend({
...
+      // their metrics will not post from a queue flush in time
+      this.postPreLoginMetric(_payload, preLoginId);
+    } else {
+      return this.clientMetricsBatcher.request(payload);
+    }
+  },
+
+
+  /**
+   * Issue request to alias a user's pre-login ID with their CI UUID
```
ianwremmel: maybe `aliasUser`? Not sure if that's better, but I couldn't figure out what this did until I read the description, then it was obvious.

packages/plugin-metrics/src/metrics.js
Cruig: @ianwremmel that doesn't work because it's prior to auth and so it errors with "`metrics`" is not a known service... how should this work?

packages/plugin-metrics/src/metrics.js
Cruig: fixed with using `qs`

packages/plugin-metrics/src/metrics.js
Cruig: fixed the lowercasing, yes @ianwremmel it's in the cors-allowed (and i've tested to confirm that it works), and fixed to use back ticks

packages/plugin-metrics/src/metrics.js
Cruig: fixed

packages/plugin-metrics/src/metrics.js
Cruig: fixed

packages/plugin-metrics/test/unit/spec/metrics.js
Cruig: fixed

packages/plugin-metrics/test/unit/spec/metrics.js
Cruig: fixed

packages/plugin-metrics/test/unit/spec/metrics.js
Cruig: fixed

src/client/services/metrics/metrics.js
Cruig: fixed

src/client/services/metrics/metrics.js
Cruig: fixed

PR #419, opened 2017-01-21
feat(mercury+board): sharing mercury web socket

tran2:
For now, sharing web socket is hiding behind the feature toggle `web-shared-mercury`
Two changes are in this PR (let me know if I should split this into two PRs)

## Mercury
When `web-shared-mercury` is enabled, mercury will allow sharing its socket connection,
- webSocketUrl changes from having query parameter `bufferState=true` to `mercuryRegistrationStatus=true&registrationRefreshStatus=true` (`registrationRefreshStatus=true` is a server side toggle, which will be removed when we remove feature toggle)

## Whiteboard realtime
- add two methods to allow adding and removing board bindings from mercury connection
- refactor connectByCreatingNewSocket by combining register() and connect() (before, sdk user needs to manually do those steps)

### DIFF
--- a/packages/plugin-board/src/board.js
+++ b/packages/plugin-board/src/board.js
@@ -408,6 +408,64 @@ const Board = SparkPlugin.extend({
       .then((res) => res.body);
   },
 
+  /**
+   * Registers with Mercury for sharing web socket
+   * @memberof Board.BoardService
+   * @param  {Board~Channel} channel
+   * @returns {Promise<Board~Registration>}
+   */
+  registerToShareMercury(channel) {
+    const isSharingMercuryFeatureEnabled = this.spark.device.features.developer.get(`web-sharable-mercury`);
+
+    if (!this.spark.mercury.localClusterServiceUrls) {
+      return Promise.reject(new Error(`\`localClusterServiceUrls\` is not defined, make sure mercury is connected`));
+    }
+    else if (!isSharingMercuryFeatureEnabled || !isSharingMercuryFeatureEnabled.value) {
+      return Promise.reject(new Error(`\`web-sharable-mercury\` is not enabled`));
+    }
+
+    const webSocketUrl = this.spark.device.webSocketUrl;
+    const mercuryConnectionServiceClusterUrl = this.spark.mercury.localClusterServiceUrls.mercuryConnectionServiceClusterUrl;
+
+    const data = {
+      mercuryConnectionServiceClusterUrl,
+      webSocketUrl,
+      action: `REPLACE`
+    };
+
+    return this.spark.request({
+      method: `POST`,
+      api: `board`,
+      resource: `/channels/${channel.channelId}/register`,
+      body: data
+    })
+      .then((res) => res.body);
+  },
+
+  /**
+   * Remove board binding from existing mercury connection
+   * @memberof Board.BoardService
+   * @param  {Board~Channel} channel
+   * @param  {String} binding - the binding as provided in board registration
+   * @returns {Promise<Board~Registration>}
+   */
+  unregisterFromSharedMercury(channel, binding) {
+    const webSocketUrl = this.spark.device.webSocketUrl;
+    const data = {
+      binding,
+      webSocketUrl,
+      action: `REMOVE`
+    };
+
+    return this.spark.request({
+      method: `POST`,
+      api: `board`,
+      resource: `/channels/${channel.channelId}/register`,
+      body: data
+    })
+      .then((res) => res.body);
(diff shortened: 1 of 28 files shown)


packages/plugin-board/src/realtime.js
```diff
@@ -82,9 +82,47 @@ const RealtimeService = Mercury.extend({
...
     return this.socket.send(data);
   },
 
+  /**
+   * Connect and use an exisiting mercury connection
+   * @memberof Board.RealtimeService
+   * @param  {Board~Channel} channel
+   * @returns {Promise<Board~Registration>}
+   */
+  connectToSharedMercury(channel) {
```
tran2: We make this call first when we want to share mercury connection (usually when we open a new whiteboard), then the response will include `sharedWebSocket`:
```JSON
{
   "mercuryConnectionServiceClusterUrl":"...",
   "binding":"...",
   "webSocketUrl":"...",
   "sharedWebSocket":true,
   "action":"REPLACE"
}
```
If `sharedWebSocket` is true, we won't need to do anything as board binding is registered.

If `sharedWebSocket` is false, then we will need to open a second web socket using `webSocketUrl`. I am letting the web-client handle this situation. Do you think it's better to have this logic in the sdk?

ianwremmel: I think so, yes. In particular, I'd like it to be integration tested in the sdk. I don't see a good way to test it in the web client.

packages/plugin-board/src/board.js
```diff
@@ -415,30 +415,34 @@ const Board = SparkPlugin.extend({
...
-    return this.spark.request({
-      method: `POST`,
-      api: `board`,
-      resource: `/channels/${channel.channelId}/register`,
-      body: data
-    })
+        return data;
+      })
+      .then((data) =>
+        this.spark.request({
```
ianwremmel: put implicit returns on the same line as their arrows

packages/plugin-board/src/board.js
```diff
@@ -408,6 +408,66 @@ const Board = SparkPlugin.extend({
...
+        const mercuryConnectionServiceClusterUrl = this.spark.mercury.localClusterServiceUrls.mercuryConnectionServiceClusterUrl;
+
+        const data = {
+          mercuryConnectionServiceClusterUrl,
+          webSocketUrl,
+          action: `REPLACE`
+        };
+
+        return data;
+      })
```
ianwremmel: I don't think there's a need for this then block. can't you just call `spark.request` with `data`?

packages/plugin-board/src/board.js
```diff
@@ -408,6 +408,66 @@ const Board = SparkPlugin.extend({
...
+      .then((data) => this.spark.request({
+        method: `POST`,
+        api: `board`,
+        resource: `/channels/${channel.channelId}/register`,
```
ianwremmel: do channels have urls? would it be more appropriate to use `url: channel.url`?

packages/plugin-board/src/board.js
```diff
@@ -408,6 +408,66 @@ const Board = SparkPlugin.extend({
...
+    const data = {
+      binding,
+      webSocketUrl,
+      action: `REMOVE`
+    };
+
+    return this.spark.request({
+      method: `POST`,
+      api: `board`,
+      resource: `/channels/${channel.channelId}/register`,
```
ianwremmel: same comment about urls

packages/plugin-board/src/realtime.js
ianwremmel: Did this logic move into the sdk yet? I don't see anything reading sharedWebSocket.

packages/plugin-board/src/realtime.js
```diff
@@ -85,7 +89,90 @@ const RealtimeService = Mercury.extend({
...
+    return this.spark.board.registerToShareMercury(channel)
+      .then((res) => {
+        this.boardBindings = [res.binding];
+        this.boardWebSocketUrl = res.webSocketUrl;
+
+        if (!res.sharedWebSocket) {
```
tran2: added logic for when `sharedWebSocket` is false

test/unit/spec/services/board/realtime.js
```diff
@@ -135,43 +135,88 @@ describe('Services', function() {
         });
       });
 
-      describe('#connectToSharedMercury', function() {
-        var registrationInfo;
+      describe.only('when trying to share mercury connection', function() {
```
ianwremmel: remove .only

PR #426, opened 2017-01-27
refactor(common): replace base64 methods implementation with package

### LINKED ISSUE #95: Replace base64 methods in common with those used by node-jose

tran2:
use urlsafe-base64 package as similar to node-jose

resolves #95

### DIFF
--- a/packages/common/src/base64.js
+++ b/packages/common/src/base64.js
@@ -6,6 +6,7 @@
 import UrlSafeBase64 from 'urlsafe-base64';
 
 /**
+ * Converts a string from a base64url-encoded string
  * @param {string} str
  * @returns {string}
  */
@@ -14,17 +15,22 @@ export function fromBase64url(str) {
 }
 
 /**
- * Converts a string to a base64url-encoded string
- * @param {string} str
+ * Converts a string to a base64url-encoded string. It also accepts a buffer
+ * @param {string|buffer} str
  * @returns {string}
  */
 export function toBase64Url(str) {
-  return UrlSafeBase64.encode(new Buffer(str));
+  let buffer = str;
+  if (!Buffer.isBuffer(buffer)) {
+    buffer = new Buffer(buffer);
+  }
+
+  return UrlSafeBase64.encode(buffer);
 }
 
 /**
- * Converts a string to a base64url-encoded string
- * @param {string} str
+ * Converts a string to a base64url-encoded string. It also accepts a buffer
+ * @param {string|buffer} str
  * @returns {string}
  */
 export function encode(str) {
--- a/packages/common/src/base64.js
+++ b/packages/common/src/base64.js
@@ -3,15 +3,14 @@
  * Copyright (c) 2015-2016 Cisco Systems, Inc. See LICENSE file.
  */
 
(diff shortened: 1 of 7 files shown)


packages/common/src/base64.js
```diff
@@ -3,15 +3,14 @@
  * Copyright (c) 2015-2016 Cisco Systems, Inc. See LICENSE file.
  */
 
-import atob from './shims/atob';
```
ianwremmel: also delete the shim files

packages/common/src/base64.js
```diff
@@ -20,10 +19,7 @@ export function fromBase64url(str) {
  * @returns {string}
  */
 export function toBase64Url(str) {
-  return btoa(str)
-    .replace(/\+/g, `-`)
-    .replace(/\//g, `_`)
-    .replace(/\=/g, ``);
+  return UrlSafeBase64.encode(new Buffer(str));
```
ianwremmel: Doesn't this also need to call toString()? or does encode already return a string?

packages/common/test/unit/spec/common.js
```diff
@@ -16,12 +16,19 @@ describe(`common`, () => {
       assert.isDefined(base64);
     });
 
-    it(`does btoa transforms`, () => {
+    it(`encodes string to base64`, () => {
```
ianwremmel: "base64-encodes a string"

packages/common/test/unit/spec/common.js
```diff
@@ -16,12 +16,19 @@ describe(`common`, () => {
...
     });
 
-    it(`does atob transforms`, () => {
-      assert.equal(base64.fromBase64url(`YWJj`), `abc`);
+    it(`can encode buffer to base64`, () => {
+      const buffer = new Buffer(`abc`);
+      assert.equal(base64.toBase64Url(buffer), `YWJj`);
+    });
+
+    it(`decodes base64 string to readable string`, () => {
```
ianwremmel: "base64-decodes a string"

packages/common/test/unit/spec/common.js
```diff
@@ -16,12 +16,19 @@ describe(`common`, () => {
...
     });
 
-    it(`does btoa transforms`, () => {
+    it(`encodes string to base64`, () => {
       assert.equal(base64.toBase64Url(`abc`), `YWJj`);
     });
 
-    it(`does atob transforms`, () => {
-      assert.equal(base64.fromBase64url(`YWJj`), `abc`);
+    it(`can encode buffer to base64`, () => {
```
ianwremmel: "base64-encodes a buffer"

packages/common/src/base64.js
ianwremmel: if str is already a buffer, this could effectively be `new Buffer(Buffer(string))`. Can you add a check to make sure `new Buffer()` is only called if str is not a buffer?

tran2: yes encode already returns a string

linuxwolf:
@tran2 / @ianwremmel FWIW, I'm 👍

:shipit:

PR #429, opened 2017-01-30
Reenable call testing 

adamweeks:
This expands upon #384 and resolves merge conflicts.


adamweeks:
Left a lot of tests skipped and added more to the skipped list currently. The remaining tests pass on sauce. Will work on re-enabling them on the next story/PR.

packages/example-phone/test/automation/spec/call.js
```diff
@@ -3,7 +3,7 @@ import testUsers from '@ciscospark/test-helper-test-users';
 import pkg from '../../../package';
 
 
-describe(`example-phone`, () => {
+describe.skip(`example-phone call`, () => {
```
ianwremmel: rather than putting call, incoming-call, etc in the first describe block, can you add another nest level? the package tests pretty much always have the package name as the first describe.

adamweeks: Sure. I was wondering why each spec file had the same opening describe block!

ianwremmel: yep. it's really helpful for gating other jobs, since the plugin name roughly corresponds to which service is broken.

adamweeks:
Describe blocks have been updated @ianwremmel

adamweeks:
@ianwremmel let me know when we are good to merge these in with the pipeline

PR #432, opened 2017-02-01
fix(logout): fix spark.logout() to fail gracefully

aimeex3:
fixes https://github.com/ciscospark/spark-js-sdk/issues/220

### DIFF
--- a/src/client/device/device.js
+++ b/src/client/device/device.js
@@ -309,7 +309,7 @@ var Device = SparkBase.extend({
     this.logger.info('device: unregister requested');
 
     if (!this.url) {
-      throw new Error('device is not registered');
+      return Promise.reject('device is not registered');
     }
 
     var options = {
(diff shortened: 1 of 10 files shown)


src/spark-core.js
```diff
@@ -144,9 +144,9 @@ var SparkCore = AmpersandState.extend({
     return this.device.remove()
       .catch(function logDeviceRemovalFailure(reason) {
         this.logger.warn(reason);
-      })
+      }.bind(this))
       .then(function clearEncryptStore() {
-        this.encryption.keystore.clear();
+        return this.encryption.keystore.clear();
```
aimeex3: @ianwremmel, is encryption always defined in legacy sdk or should we also check `if (this.encryption)`

arun3528:
nice work aimee

src/client/device/device.js
```diff
@@ -309,7 +309,7 @@ var Device = SparkBase.extend({
     this.logger.info('device: unregister requested');
 
     if (!this.url) {
-      throw new Error('device is not registered');
+      return Promise.reject('device is not registered');
```
mattnorris: 👍

src/spark-core.js
ianwremmel: yep, it's always defined

PR #437, opened 2017-02-02
fix(plugin-encryption): Utilize spark:kms scope

andrewholsted:
Pass spark:kms to spark.credentials.getAuthorization(). @ianwremmel - should there be a unit test for this? I checked for an existing one but couldn't find it. On top of that, I failed to get the tests to run locally following the README. Happy to share the console output.


ianwremmel:
A unit test would be great. Send me the console output and I'll have a look.

ianwremmel:
Fixed in 1.0

PR #438, opened 2017-02-03
feature(user-activation): add user activation through ci

aimeex3:


packages/plugin-user/test/integration/spec/user.js
```diff
@@ -17,51 +17,75 @@ import {nodeOnly} from '@ciscospark/test-helper-mocha';
...
+        assert.isFalse(res.sso);
+      })
+    );
+  });
+
   // This test relies on setting a specific user agent, so it doesn't work in
   // browsers
   nodeOnly(describe)(`#activate()`, () => {
-    it(`completes the mobile signup process`, () => spark.request({
+    xit(`completes the mobile signup process`, () => spark.request({
```
aimeex3: cannot test this yet until server provides an endpoint where we can get a verification token

test/integration/spec/services/user.js
```diff
@@ -179,10 +193,27 @@ describe('Services', function() {
...
+            assert.isFalse(res.sso);
+          });
+      });
+    });
+
+    skipInBrowser(describe)('#activate() @CI', function() {
       createUnauthedSpark();
 
-      it('registers mobile user with a valid email address and activates account @atlas', function() {
+      xit('retrieves a valid user token using verificationToken', function() {
```
aimeex3: same here, cannot test until server has endpoint that gives back verification token

packages/plugin-credentials/src/token.js
```diff
@@ -34,6 +34,10 @@ const Token = SparkPlugin.extend({
     access_token: `string`,
     expires: `number`,
     expires_in: `number`,
+    passwordSet: {
```
ianwremmel: why `passwordSet` instead of `hasPassword`?

packages/plugin-credentials/test/integration/spec/token.js
```diff
@@ -42,6 +42,16 @@ describe(`plugin-credentials`, function() {
           assert.notEqual(token.access_token, supertoken.access_token);
           assert.equal(token.refresh_token, supertoken.refresh_token);
         }));
+
+      it(`sets the previous passwordSet state onto the new token`, () => {
+        supertoken.passwordSet = false;
+        return supertoken.refresh()
+        .then((token) => {
```
ianwremmel: indent `.then`

packages/plugin-user/src/config.js
```diff
@@ -5,14 +5,19 @@
...
+    activationUrl: `https://idbroker.webex.com/idb/token/v1/actions/UserActivation/invoke`,
     batcherWait: 100,
     batcherMaxCalls: 100,
     batcherMaxWait: 1500,
+    setPasswordUrl: `https://identity.webex.com/identity/scim/v1/Users`,
     verifyDefaults: {
-      deviceName: `DESKTOP`,
-      deviceId: `not a mobile device`,
-      pushId: `not a mobile device`
+      reqId: `DESKTOP`
```
ianwremmel: what does reqId mean?

packages/plugin-user/src/user.js
```diff
@@ -28,26 +28,42 @@ const User = SparkPlugin.extend({
...
+      uri: this.config.activationUrl,
       method: `POST`,
       body: options,
-      requiresClientCredentials: true
+      auth: {
+        user: this.spark.config.credentials.oauth.client_id,
+        pass: this.spark.config.credentials.oauth.client_secret,
+        sendImmediately: true
+      },
+      withCredentials: true
```
ianwremmel: I thought we weren't doing this with cookies anymore

packages/plugin-user/src/user.js
```diff
@@ -28,26 +28,42 @@ const User = SparkPlugin.extend({
...
     })
-      .then((res) => res.body);
+      .then((res) => {
+        response = res;
+        return this.getOauthCode();
+      })
+      .then((res) => {
+        const code = res.match(`<title>(.*)</title>`)[1];
```
ianwremmel: are you sure this works? i'm pretty sure when you pass a string to String#match(), it looks for a string match rather than converting it to a regex.

I think what you're looking for is `/<title>(.*?)</title>/`

packages/plugin-user/src/user.js
```diff
@@ -173,6 +189,55 @@ const User = SparkPlugin.extend({
   },
 
   /**
+   * Sets password status on token
+   * @param {Boolean} value
+   * @returns {undefined}
+   */
+  setPasswordStatus(value) {
+    if (this.spark.credentials && this.spark.credentials.supertoken) {
```
ianwremmel: this should call a credentials method rather than talking to supertoken directly. that'll make it easier to gut the credentials plugin(s) later.

packages/plugin-user/src/user.js
```diff
@@ -193,21 +258,50 @@ const User = SparkPlugin.extend({
...
+    }
+
+    if (!options.reqId) {
+      return Promise.reject(new Error(`\`options.reqId\` is required`));
+    }
+
+    let shouldRefreshAccessToken = true;
+    let requiresClientCredentials = false;
+    let headers;
+    const promise = this.spark.credentials.getAuthorization()
```
ianwremmel: Why the `promise` variable instead of just returning immediately?

packages/plugin-user/test/integration/spec/user.js
ianwremmel: use `.skip`, not `xit`. If we can't test it, we can't merge it. do they have a timeline?

aimeex3: a couple of weeks. nikhil is working on it this sprint

packages/plugin-credentials/src/token.js
aimeex3: i added hasPassword in the derived section, but i guess that's not needed :)

packages/plugin-user/src/config.js
aimeex3: reqId is required in the atlas endpoint. it is supposed to be used to check if the email has been delivered. https://wiki.cisco.com/display/WX2/New+Onboarding+Flows
but it is currently not used in the web client

packages/plugin-user/src/user.js
aimeex3: we still are. i don't think CI is changing it

packages/plugin-user/src/user.js
aimeex3: oo interesting it's been working, but i will change it

packages/plugin-user/src/user.js
aimeex3: 👍  removed cookie usage

packages/plugin-user/src/config.js
ianwremmel: Can we use a value here that differentiates us from the native client?

ianwremmel: also, this should come from config, not be hardcoded

packages/plugin-user/src/user.js
ianwremmel: so, withCredentials can be removed, right?

packages/plugin-user/src/user.js
```diff
@@ -173,6 +183,51 @@ const User = SparkPlugin.extend({
...
+   * @param {string} options.password (required)
+   * @param {string} options.userId (required)
+   * @returns {Promise} Resolves with complete user object containing new password
+   */
+  setPassword(options) {
+    options = options || {};
+    if (!options.password) {
+      return Promise.reject(new Error(`\`options.password\` is required`));
+    }
+    if (!options.userId) {
```
ianwremmel: Does this get called before or after device registration? if it's after, then you can read spark.device.userid instead of making it an option param.

packages/plugin-user/src/user.js
```diff
@@ -193,21 +248,49 @@ const User = SparkPlugin.extend({
...
+
+    if (!options.reqId) {
+      return Promise.reject(new Error(`\`options.reqId\` is required`));
+    }
+
+    let shouldRefreshAccessToken = true;
+    let requiresClientCredentials = false;
+    let headers;
+
+    return this.spark.credentials.getAuthorization()
```
ianwremmel: getAuthorization() should be marked as deprecated. I think you want getUserToken()

packages/plugin-user/src/user.js
aimeex3: should get called after device registration. 👍 changed it to `this.spark.device.userId`

packages/plugin-user/src/user.js
aimeex3: whoops yeah

packages/plugin-user/test/integration/spec/user.js
```diff
@@ -11,76 +11,142 @@ import sinon from '@ciscospark/test-helper-sinon';
...
+          const query = url.parse(res.verifyEmailURL).query;
+          const token = querystring.parse(query).t;
+          return assert.isFulfilled(unauthSpark.user.activate({verificationToken: token}));
+        })
+        .then((res) => {
+          assert.property(res, `email`);
+          assert.property(res, `cookieName`);
+          assert.property(res, `cookieValue`);
+          assert.property(res, `cookieDomain`);
+          assert.property(res, `tokenData`);
```
ianwremmel: We only care about tokenData now, right?

aimeex3: yes and email. good point, will remove the cookie values

PR #439, opened 2017-02-03
Add passport support

ianwremmel:
This PR adds support for [passportjs](http://www.passportjs.org/). it got delayed for a while and probably needs to be updated, but should be reasonably easy to adapt to the current code base if so desired.

### DIFF
--- a/packages/passport-ciscospark/test/automation/fixtures/router.js
+++ b/packages/passport-ciscospark/test/automation/fixtures/router.js
@@ -1,0 +1,41 @@
+const express = require(`express`);
+const session = require(`express-session`);
+const CiscoSparkStrategy = require(`../../..`).Strategy;
+const passport = require(`passport`);
+const errorhandler = require(`errorhandler`);
+const {base64} = require(`@ciscospark/common`);
+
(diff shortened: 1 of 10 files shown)


ianwremmel:
Note to self: need a way to add the Spark instance to the request

packages/passport-ciscospark/test/automation/fixtures/router.js
```diff
@@ -0,0 +1,41 @@
...
+  done(null, `id`);
+});
+
+passport.deserializeUser((id, done) => {
+  done(null, userDB);
+});
+
+passport.use(new CiscoSparkStrategy({
+  // eslint-disable-next-line camelcase
+  redirect_uri: `http://127.0.0.1:${process.env.PORT}`
```
adamweeks: Do we need to use the redirect_uri from options here?

ianwremmel: yea, i think this was a remnant from an earlier iteration. I'll fix it.

ianwremmel:
marking as blocked for now; should be revisited after the auth overhaul

PR #440, opened 2017-02-03
test(plugin-phone): expand test coverage

adamweeks:
Ready for review. Will be much cleaner once #429 is merged in.

Still quite a few tests skipped or incomplete, but this is a lot more coverage for the `plugin-phone` as it exists now.


packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -223,47 +320,161 @@ describe(`plugin-phone`, function() {
...
-        it(`is false`);
+        it(`is false`, () => {
+          return call.stopSendingAudio()
+            .then(() => assert.isFalse(call.sendingAudio));
+        });
       });
     });
 
     describe(`#sendingVideo`, () => {
+      let call;
```
ianwremmel: Does it make sense to do these tests based on starting the calls with different constraints rather instead of changing the audio status in the call?

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -223,47 +320,161 @@ describe(`plugin-phone`, function() {
...
+      afterEach(() => {
+        const c = call;
+        call = undefined;
+        return c.hangup().catch((reason) => console.warn(reason));
+      });
+
       describe(`when the local party is receiving Video`, () => {
-        it(`is true`);
+        it(`is true`, () => {
+          return call.startReceivingVideo()
```
ianwremmel: use implicit returns when possible

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -317,18 +576,135 @@ describe(`plugin-phone`, function() {
...
+
+      describe(`in-progress events`, () => {
+        let call;
+        beforeEach(() => {
+          call = spock.spark.phone.dial(mccoy.email);
+          assert.equal(call.status, `initiated`);
+        });
+
+        describe(`on(remoteAudioMuted:change)`, () => {
+          // stopSendingAudio doesn't trigger this change but should
```
ianwremmel: can you change the comment to something more like "when the remote part invokes stop sending audio, the local party doesn't receive the update"? I had to stare at it for a while before I realized the comment was, in fact, valid.

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -317,18 +576,135 @@ describe(`plugin-phone`, function() {
...
+                .then(() => remoteCall.stopSendingAudio())
+                .then(() => remoteCall.hangup()),
+              call.when(`disconnected`)
+                .then(() => {
+                  return assert.callCount(remoteAudioSpy, 1);
+                })
+            ]);
+          });
+
+          // startSendingAudio doesn't trigger this change but should
```
ianwremmel: see previous

packages/plugin-phone/src/call.js
```diff
@@ -389,7 +389,9 @@ const Call = SparkPlugin.extend({
       this.stopListening(this.spark.mercury);
       this.off();
       URL.revokeObjectURL(this.localMediaStreamUrl);
+      this.localMediaStreamUrl = undefined;
```
adamweeks: This is the only "fix" I put in for the plugin to match the tests. We will do more in another PR.

adamweeks:
Reasons for `skip` tests:
- If a call is created and `.hangup()` is called on it before the locus session is created, the promise just waits for the locus session until timeout. We need a way to cancel a locus creation.
- `remoteAudioMuted:change` does not get called when the remote caller changes their audio status (with `.stopSendingAudio()` )
- When calling `.toggleReceivingAudio()`  to change into send only status, the sdp media direction should be `sendonly` but it is actually `inactive` 
- Test users cannot create 1:1 messages to each other (some sort of message signing issue perhaps?).

packages/plugin-phone/test/integration/spec/phone.js
```diff
@@ -65,19 +64,38 @@ describe(`plugin-phone`, function() {
     });
 
     describe(`#deregister()`, () => {
-      it(`disconnects from mercury`);
-      it(`unregisters from wdm`);
-      it(`is a noop when not registered`);
+      it(`disconnects from mercury`, () => {
+        const mercuryDisconnectSpy = sinon.spy(spock.spark.mercury, `disconnect`);
```
ianwremmel: add assertions that mercury is connected to demonstrate the test is valid

packages/plugin-phone/test/integration/spec/phone.js
```diff
@@ -65,19 +64,38 @@ describe(`plugin-phone`, function() {
...
+        return spock.spark.phone.deregister()
+          .then(() => assert.calledOnce(mercuryDisconnectSpy))
+          .then(() => assert.isFalse(spock.spark.mercury.connected, `Mercury is not connected`))
+          .then(() => assert.isFalse(spock.spark.phone.connected, `Mercury (proxied through spark.phone) is not conneted`))
+          .then(() => mercuryDisconnectSpy.restore());
+      });
+
+      it(`unregisters from wdm`, () =>
+        spock.spark.phone.deregister()
```
ianwremmel: implicit returns should go on the same line as the `=>`

packages/plugin-phone/test/integration/spec/phone.js
```diff
@@ -65,19 +64,38 @@ describe(`plugin-phone`, function() {
...
+          .then(() =>
+            spock.spark.device.on(`change:device`)
+              .then(() => assert.isUndefined(spock.spark.device.url))
+          )
+      );
+
+      // TODO: Does not currently noop after multiple calls
+      it.skip(`is a noop when not registered`, () => {
```
ianwremmel: mercury.disconnect() is the noop in question. maybe spy on spark.request (which is currently impossible, iirc) instead and assert that no network calls get made?

packages/plugin-phone/test/integration/spec/phone.js
```diff
@@ -138,7 +154,35 @@ describe(`plugin-phone`, function() {
 
       it(`calls a user by AppID username`);
       it(`calls a user by tropo uri`);
-      it(`calls a user by spark uri`);
+
+      // FIXME: Test users currently cannot create 1:1 messages to each other
+      it.skip(`calls a user by spark uri`, () => {
+        let conversation;
+        let sparkUri;
+        return spock.spark.request({
```
ianwremmel: can you write a testcase somewhere demonstrating the 1:1 messaging problem?

packages/plugin-phone/test/integration/spec/media-state-controls.js
```diff
@@ -144,8 +145,9 @@ describe(`plugin-phone`, function() {
         });
 
         describe(`when the call is not receiving audio`, () => {
```
ianwremmel: Before eliminating the bugzilla comment, please run this test in firefox via sauce. I think you'll find that it fails with a different error

ianwremmel:
@adamweeks  this is merged, right?

ianwremmel:
closing in favor of alternate merge method

PR #442, opened 2017-02-06
fix(logout): sanitize options param in logout function

aimeex3:


packages/plugin-credentials/test/unit/spec/credentials.js
```diff
@@ -307,5 +308,76 @@ describe(`plugin-credentials`, () => {
...
+        spark.credentials._redirect = sinon.spy();
+        assert.isDefined(spark.credentials.supertoken);
+        return spark.credentials.logout()
+          .then(() => {
+            assert.isUndefined(spark.credentials.supertoken.access_token);
+            assert.notCalled(spark.credentials._redirect);
+          });
+      });
+
+      it(`revokes the access token, but does not redirect user`, () => {
```
ianwremmel: wrap this in a describe block indicating `noRedirect: true`

Issue #441, opened 2017-02-06
Plugins cannot replace standard ones (e.g. plugin-logger)


hagemt:
In some scenarios, we'd like to replace the `this.spark.logger` but run into some trouble:

Calling `registerPlugin('logger', ..., { replace: true, ... })` works as expected, but:

Doing so *before* `require('ciscospark')` is "required", since the import constructs an instance.
(registering plugins after instances are created is bad, unless you can guarantee they're unused)
Doing so *after* `require('ciscospark')` will throw an Error if `NODE_ENV !== 'production'`.

IMHO, I think the `interceptors` mechanism is sufficient for adding hooks. (enables quite a bit)
But AFAIK, this doesn't enable us to capture and redirect everything that the SDK would log.

Thanks.

hagemt:
FTR, we hacked around this issue by explicitly registering the "stock" `logger` plugin first, then replacing it with our own, which works for us since we never use the default instance.

joeyyin123:
@hagemt Does the issue still persist on your side? Thanks.

hagemt:
@joeyyin123; will close. Not sure if it was addressed, but the workaround (as above) was fine for us.

PR #444, opened 2017-02-08
fix(spark-core): interceptors ignore plugin-logger

hagemt:
I kept the `console` as a fall-back, just in case; AFAIK, `options.logger` was the wrong reference.

It's not the case (without the work of yet another interceptor) that the `logger` is on `options`.

Effectively, this meant (for us, and likely others) that logging would always fall-back to `console`.

IIRC, the interceptor's `this` reference has a `spark`, on which the `logger` plugin is registered.

Also cross-referencing #441 for an oft-related (and still existing, but minor) issue involving replacing the `logger`. We hacked around that issue by explicitly registering the stock `logger` plugin first, then replacing it with our own, which works for us since we never use the default instance.

### DIFF
--- a/packages/spark-core/src/interceptors/request-logger.js
+++ b/packages/spark-core/src/interceptors/request-logger.js
@@ -16,7 +16,7 @@ export default class RequestLoggerInterceptor extends Interceptor {
    * @returns {RequestLoggerInterceptor}
    */
   static create() {
-    return new RequestLoggerInterceptor(this);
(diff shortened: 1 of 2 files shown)


hagemt:
Second commit adjusts `static create` to be consistent w/ e.g. `AuthInterceptor.create`.

@ianwremmel: does this look in good shape for you to merge? If so:
I remember how to ferry this through the Jenkins (validated the merge).

PR #451, opened 2017-02-15
feat(http-core): add raw description to error response

aimeex3:


packages/http-core/src/http-error.js
```diff
@@ -91,7 +102,8 @@ const HttpError = extendError({
     return message;
   },
   properties: {
-    defaultMessage: `An error was received while trying to fulfill the request`
+    defaultMessage: `An error was received while trying to fulfill the request`,
+    defaultRawDescription: `none`
```
ianwremmel: why `none` rather than `\`\``?

packages/http-core/src/http-error.js
```diff
@@ -122,6 +134,36 @@ function parseObject(body) {
   return message;
 }
 
+/**
+ * @param {object} body
+ * @private
+ * @returns {string}
+ */
+function parseDescription(body) {
```
ianwremmel: I think only SparkHttpError should have the rawDescription field, since it's the one that adds the tracking id and url to the message. on that note, `rawMessage` might be a better field name.

packages/spark-core/src/lib/spark-http-error.js
```diff
@@ -6,17 +6,43 @@
...
 import {HttpError} from '@ciscospark/http-core';
+import {isArray, pick} from 'lodash';
 
 const SparkHttpError = extendError(HttpError, {
   parseFn(res, ...rest) {
+    /* eslint complexity: [0] */
     /* eslint prefer-reflect: [0] */
     let message = HttpError.prototype.parseFn.call(this, res, ...rest);
 
+    let body = res.body;
```
ianwremmel: Why are these changes necessary? Doesn't HttpError already do this?

aimeex3: This parses just the description out of the error. HttpError parses the entire message. Initially I added the description parsing as part of HttpError, but you mentioned that it should be moved to SparkHttpError instead

ianwremmel:
can this be closed?

aimeex3:
yeah this one may need to get revisited for the new sdk

PR #455, opened 2017-02-16
WIP: call testing

ianwremmel:


ianwremmel:
Ok, there's still a lot skipped and I think we've decided to change some things, but let's get this into master.

Future changes: 

- do not renegotiate for `toggleLocalAudio()`/`toggleLocalVide()` (unless we need to establish an audio/video stream).
- refactor the `#remoteMediaStream` and `#localMediaStream` tests to remove the extra tick (using call.when() is introducing an extra tick and therefore allowed one of the more annoying bugs to go undetected)

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -149,11 +149,45 @@ describe(`plugin-phone`, function() {
     });
 
     describe(`#to`, () => {
-      it(`represents the receiving party`);
+      let call;
+      afterEach(() => {
+        const c = call;
+        call = undefined;
```
bzang: Seems a bit awkward. Why does `c` need to be set?

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -223,47 +336,183 @@ describe(`plugin-phone`, function() {
...
+            call.when(`connected`)
+              .then(() => assert.equal(call.status, `connected`))
+              .then(() => assert.isTrue(call.receivingVideo))
+          ]);
+        });
       });
 
       describe(`when the local party is not receiving Video`, () => {
-        it(`is false`);
+        it(`is false`, () => {
```
bzang: Seems like a double negative here. Should this be a part of `when the local party is receiving Video` and `it('is false')`?

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -278,23 +527,71 @@ describe(`plugin-phone`, function() {
...
     describe(`#answer()`, () => {
-      it(`accepts an incoming call`);
+      it(`accepts an incoming call`, () => {
+        const call = mccoy.spark.phone.dial(spock.email);
+        return Promise.all([
+          spock.spark.phone.when(`call:incoming`)
+            .then(([c]) => c.answer()),
+          call.when(`connected`)
+            .then(() => assert.equal(call.status, `connected`))
+        ]);
```
bzang: Isn't this already tested all over the place?

packages/plugin-phone/test/integration/spec/call.js
ianwremmel: because we need to do two things in this function: unset call and return a promise. the alternative is

```javascript
const promise = c.hangup...
call = undefined;
return promise
```

packages/plugin-phone/test/integration/spec/call.js
ianwremmel: the `it` refers to `#receivingVideo`

packages/plugin-phone/test/integration/spec/call.js
ianwremmel: yes, but not explicitly; there are a lot of redundant code paths in order to keep assertions narrow. that said, with some of the bugs i'm aware of, a lot of these redundancies will be getting refactored

ianwremmel: there's also a lot that should be factored into unit tests

PR #456, opened 2017-02-17
check deps on build

ianwremmel:


packages/test-helper-server/package.json
```diff
@@ -18,7 +18,8 @@
     "express": "^4.13.3",
     "mkdirp": "^0.5.1",
     "morgan": "^1.6.1",
-    "multer": "^1.2.0"
+    "multer": "^1.2.0",
+    "uuid": "^3.0.1"
```
adamweeks: These added deps just already missing?

ianwremmel: right. I'm pretty sure they didn't cause problems because of npm3 making them available (or possibly because they were installed in the root node_modules directory)

PR #459, opened 2017-02-20
chore: add message demo environment vars to jenkinsfile

adamweeks:


adamweeks:
Closed due to migration to https://github.com/ciscospark/react-ciscospark

PR #461, opened 2017-02-20
feat(plugin-phone): add Phone#isWebRTCSupported()

ianwremmel:


packages/plugin-phone/test/unit/spec/phone.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe(`plugin-phone`, () => {
+  describe(`Phone`, () => {
+    describe(`#isCallingSupported()`, () => {
+      let spark;
+      beforeEach(() => {
+        spark = new CiscoSpark();
+      });
+      // This is sort of a silly test since we only actually run this test in
+      // browsers that support calling...
```
adamweeks: Could we mock out `DetectRTC.isWebRTCSupported` to have it return false to do a false test?

packages/plugin-phone/src/phone.js
```diff
@@ -70,6 +71,18 @@ const Phone = SparkPlugin.extend({
...
 
   /**
+   * Indicates if the current browser appears to support webrtc calling. Note:
+   * at this time, there's no way to determine if the current browser supports
+   * h264 without asking for camera permissions
+   * @returns {Promise<Boolean>}
+   */
+  isCallingSupported() {
+    return new Promise((resolve) => {
+      resolve(DetectRTC.isWebRTCSupported);
```
adamweeks: I like that it is a promise, allows us to change the method behind the scenes without worrying about implementation.

packages/plugin-phone/test/unit/spec/phone.js
ianwremmel: I thought about that, but it doesn't really prove anything beyond `false === false`. if we care badly enough about mocking, we'd actually need to mock the browser environment that DetectRTC is reading.

PR #462, opened 2017-02-21
test(plugin-phone): introduce wrapper to detect call error events

ianwremmel:


packages/plugin-phone/test/integration/lib/handle-error-event.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * A lot of failures get produce by EventEmitters, which makes them difficult to
```
adamweeks: `get produced`

packages/plugin-phone/test/integration/lib/handle-error-event.js
```diff
@@ -0,0 +1,31 @@
...
+ * detect in tests (they just look like timeouts). This is a test helper that
+ * captures that error and turns it into a rejected promise
+ * @param {Call} call
+ * @param {Function} fn
+ * @returns {Promise}
+ */
+export default function handleErrorEvent(call, fn) {
+  let r;
+  const p = new Promise((resolve, reject) => {
+    r = reject;
```
adamweeks: What is the purpose of this?

ianwremmel: need to hold onto reject outside of the promise so that unbind can be called

PR #463, opened 2017-02-21
feat(plugin-phone): fail fast when an offer does not include h264

ianwremmel:


packages/plugin-phone/src/call.js
```diff
@@ -42,6 +42,28 @@ import {
...
+function ensureH264(offer) {
+  const sdp = transform.parse(offer);
+  const video = find(sdp.media, {type: `video`});
+  if (!video) {
+    throw new Error(`Expected video section in offer but no such section found`);
+  }
+
+  const doesHaveH264 = video.rtp.reduce((hasH264, rtp) => hasH264 || rtp.codec.includes(`264`), false);
+  if (!doesHaveH264) {
+    throw new Error(`Offer includes video section but does not include h264 coded`);
```
bzang: Is that supposed to be "codec"?

ianwremmel: yep

PR #464, opened 2017-02-21
chore(plugin-phone): improve h264 injector

ianwremmel:


tasks/inject-h264.js
```diff
@@ -68,7 +84,8 @@ module.exports = function(grunt) {
       copy(path.join(__dirname, `selenium`, `linux`))
         .then(encode),
       copy(path.join(__dirname, `selenium`, `windows`))
-        .then(encode)
+        .then(encode),
+      rsync(path.join(__dirname, `selenium`, `mac`), path.join(__dirname, `..`, `.tmp`, `selenium`))
     ])
```
bzang: Why do we need to copy the mac path to a temp directory?

ianwremmel: this one is for local testing. using the profile breaks the profile.

PR #466, opened 2017-02-21
fix(plugin-phone): make Call#hangup() a more effective noop

ianwremmel:


packages/plugin-phone/src/call.js
```diff
@@ -181,7 +181,15 @@ const Call = SparkPlugin.extend({
     wasSendingAudio: `boolean`,
     wasSendingVideo: `boolean`,
     wasReceivingAudio: `boolean`,
-    wasReceivingVideo: `boolean`
+    wasReceivingVideo: `boolean`,
+    locusJoinInFlight: {
```
adamweeks: I like it!

PR #467, opened 2017-02-21
Media Unit Testing

ianwremmel:

PR #469, opened 2017-02-22
Media Unit Testing 2

ianwremmel:
Closes #467

### DIFF
--- a/packages/plugin-phone/src/web-rtc-media.js
+++ b/packages/plugin-phone/src/web-rtc-media.js
@@ -1,4 +1,5 @@
 import AmpState from 'ampersand-state';
+
 import {
   acceptAnswer,
   addStream,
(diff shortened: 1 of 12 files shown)


packages/plugin-phone/test/unit/spec/web-rtc-media.js
```diff
@@ -0,0 +1,171 @@
...
+        `audio`,
+        `video`
+      ]
+        .forEach((mediaType) => {
+          [
+            [true, true, false],
+            [true, false, false],
+            [false, true, true],
+            [false, false, true]
+          ].forEach(([sending, receiving, shouldRenegotiate]) => {
```
adamweeks: Wow, that is efficient! Covers lots of cases with very little code.

packages/plugin-phone/src/web-rtc-media.js
```diff
@@ -12,13 +13,64 @@ import {
...
+ * @returns {bool} true if receiving, false if not
+ */
+function getRemoteMediaStatus(kind, pc) {
+  const streams = pc.getRemoteStreams();
+
+  if (streams.length === 0) {
+    return false;
+  }
+
+  const res = streams.reduce((areStreamsFlowing, stream) => {
```
adamweeks: Would a `.some` work better here? I don't think I've used a `.reduce` for a boolean before.

ianwremmel: I hadn't used `some` before. I like it in principle, but my first attempt to switch broke all the tests, so i'm inclined to leave this as-is for now :)

packages/plugin-phone/src/webrtc.js
```diff
@@ -49,87 +30,54 @@ const startSendingMedia = curry((kind, pc) => {
...
-    stream.getTracks().forEach((track) => {
-      if (track.kind === kind) {
-        track.enabled = false;
-        // Store the tracks so we can add them back later (should the user want
-        // to mute/unmute)
-        storeMediaTrackByKindByStream(kind, stream, track);
-        if (pc.removeTrack) {
-          const sender = find(pc.getSenders(), (s) => s.track === track);
-          pc.removeTrack(sender);
+          // The next few lines are a silly hack to deal with chrome not
```
adamweeks: Is there an issue we can reference to check in the future?

ianwremmel: no, but i just wrote myself a note to create one

ianwremmel:
closing in favor of alternate merge method

PR #472, opened 2017-03-02
Docs Change and Copyright Update

bbender:
Changes:

- Help guides moved from /example/ to /guides/ URLs
- Jekyll redirect plugin added to redirect old URLs to new ones
- Bundler/Gemfile added to docs
- Copyright notices updated to 2017 across project


ianwremmel:
note that using custom jekyll plugins means you'll need to change the build process. You'll need to update the jenkinsfile to do the jekyll build on our side and push the result to github pages. Right now, github is doing the jekyll compile

the copyright footer should also be a range and not just the current year, right?

bbender:
`jekyll-redirect-from` is listed as a [supported plugin](https://help.github.com/articles/adding-jekyll-plugins-to-a-github-pages-site/) for GitHub Pages. There are conflicting docs out there about Jekyll plugins and GitHub Pages. This plugin worked in my test repository without requiring static pages.

I also updated the copyright footer to be a range.

docs/_config.yml
```diff
@@ -9,7 +9,7 @@
 title: Cisco Spark JavaScript SDK
 baseurl: "/spark-js-sdk"
 permalink: /:categories/:title/
-url: "ciscospark.github.io/spark-js-sdk"
+url: "http://ciscospark.github.io"
```
ianwremmel: https?

bbender: Good catch. It's updated to `https`.

PR #477, opened 2017-03-08
refactor(plugin-mercury): remove the first ping and use buffer_state as the auth success indicator

ianwremmel:


packages/plugin-mercury/src/socket/socket-base.js
```diff
@@ -341,29 +341,14 @@ export default class Socket extends EventEmitter {
...
-          /* istanbul ignore next */
-          return reject(error);
-        }
-      });
-
-      this.send({
-        id,
-        type: `ping`
-      });
+      };
```
bzang: Should we have a rejected promise in if the wrong thing comes back?

packages/test-helper-mock-web-socket/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.7.34",
   "description": "",
   "license": "MIT",
-  "main": "dist/index.js",
+  "main": "src/index.js",
```
bzang: shouldn't main be `dist`? or does this package never get built independently?

ianwremmel: whoops. will fix

packages/plugin-mercury/src/socket/socket-base.js
ianwremmel: nope. the ping here is just to kick off ping/pong. it's not longer relevant to prove we've authenticated successfully. _ping has it's own logic for addressing mismatched pongs

PR #486, opened 2017-03-15
fix(plugin-conversation): browser encoding of thumbnails changed to atob

adamweeks:


packages/plugin-conversation/src/process-image.browser.js
```diff
@@ -94,7 +92,7 @@ export default function processImage({file, thumbnailMaxWidth, thumbnailMaxHeigh
         },
         file);
       const parts = canvas.toDataURL(`image/png`).split(`,`);
-      const byteString = base64.decode(parts[1]);
+      const byteString = atob(parts[1]);
```
ianwremmel: Add a comment about why we're not using base64. Otherwise, I imagine I'll see this in six months and try to "fix" it.

packages/plugin-conversation/src/process-image.browser.js
```diff
@@ -6,9 +6,7 @@
 
 import {pick} from 'lodash';
 import {orient} from '@ciscospark/helper-image';
-import {base64} from '@ciscospark/common';
-
-/* global Blob, document, Image, URL */
+/* eslint-env browser */
```
ianwremmel: Nice. I feel like this didn't work last time. Or, didn't have *all* the globals.

adamweeks: Yea, I was happy it did.

PR #489, opened 2017-03-21
ffix(plugin-credentials): do not attempt to downscope without a client_id/client_secret

ianwremmel:


ianwremmel:
Updated so that we no longer need a client_secret to downscope and to emit a warning suggesting you should really include a client_id.

PR #491, opened 2017-03-22
test(plugin-wdm): prove we do not get an ios url when refreshing an expired device

ianwremmel:


ianwremmel:
merged via a rebase; closing manually

PR #493, opened 2017-03-23
fix(plugin-phone): RemoteMediaStream fail when localMedia was provided

bzang:


packages/plugin-phone/src/call.js
```diff
@@ -832,15 +832,15 @@ const Call = SparkPlugin.extend({
...
-      this.media.set({
-        audio: options.constraints.audio,
-        video: options.constraints.video,
-        offerToReceiveAudio: options.offerOptions.offerToReceiveAudio,
-        offerToReceiveVideo: options.offerOptions.offerToReceiveVideo
-      });
     }
 
+    this.media.set({
+      audio: options.constraints.audio,
```
ianwremmel: the audio and video lines should stay inside the conditional

PR #496, opened 2017-03-24
feat(widget-message-meeet): Remove widgets from repo

bzang:


bzang:
There may be some, but there is some cross use with the `example-phone` so I figure for now it's safer to leave them in. Is Jest being used in the example? If not we can probably remove that.

ianwremmel:
nope, jest is not used in the example.

PR #497, opened 2017-03-26
Alle

### LINKED ISSUE #305: Investigate moving to alle
Lerna is really handy for a lot of things, but the alle pattern solves some additional problems around build time, maintaining consistent dep versions, and perhaps most importantly, enabling greenkeeper.

- https://github.com/boennemann/alle
- https://github.com/pouchdb/pouchdb

ianwremmel:
Fixes #305


ianwremmel:
# Scripts

These scripts did the heavy lifting.

## Move package folders

```bash
#!/bin/bash
for i in ./packages/*; do
  git mv "${i}" "./packages/node_modules/$(cat ${i}/package.json | jq .name | tr -d '"')";
done
```

## Move dependencies from individual packages to top-level package.json

> Run this with @ianwremmel/tooling.js

```js
const {uniq} = require(`lodash`);
/* eslint-disable */
let dirs = readdir(`./packages/node_modules`).filter((d) => d !== `@ciscospark`);
dirs = dirs.concat(readdir(`./packages/node_modules/@ciscospark`).map((d) => `@ciscospark/${d}`));
const deps = [];

for (const d of dirs) {
  const pkg = require(`./packages/node_modules/${d}/package.json`);
  if (pkg.dependencies) {
    Object.keys(pkg.dependencies).forEach((dep) => {
      if (!dep.includes(`@ciscospark`)) {
        deps.push(`${dep}@${pkg.dependencies[dep]}`);
      }
    });
  }
}

// echo(uniq(deps.sort()).reverse());
const filtered = deps.sort().reverse().reduce((acc, dep) => {
  if (!acc.length) {
    return [dep];
  }
  const [name] = dep.split(`@`);
  if (acc.length && !(acc[acc.length - 1].includes(name))) {
    acc.push(dep);
  }
  return acc;
}, [])

sh(`npm install --save ${filtered.join(' ')}`)
```


## Remove dependencies from individual packages

```js
/* eslint-disable */
let dirs = readdir(`./packages/node_modules`).filter((d) => d !== `@ciscospark`);
dirs = dirs.concat(readdir(`./packages/node_modules/@ciscospark`).map((d) => `@ciscospark/${d}`));
const deps = [];

for (const d of dirs) {
  const pkg = require(`./packages/node_modules/${d}/package.json`);
  delete pkg.dependencies;
  delete pkg.devDependencies;
  delete pkg.peerDependencies;
  writeFile(`./packages/node_modules/${d}/package.json`, `${JSON.stringify(pkg, null, 2)}\n`);
}
```

ianwremmel:
Merge conflicts are caused by the version updates on each build. This PR should be rebased before merging.

ianwremmel:
This looks like a big PR, but if you go through commit by commit, it should be pretty straightforward.

ianwremmel:
This should be merged after #496

CONTRIBUTING.md
```diff
@@ -76,7 +76,7 @@ Install dependencies:
 ```
 # Install top-level dependencies
 npm install
-# Install dependencies for each module in ./packages and locally link unpublished modules as needed
+# Install dependencies for each module in ./packages/node_modules and locally link unpublished modules as needed
 npm run bootstrap
```
bzang: this isn't needed anymore right?

package.json
```diff
@@ -214,7 +302,7 @@
     "karma-mocha-reporter": "^2.0.4",
     "karma-safari-launcher": "^1.0.0",
     "karma-sauce-launcher": "^1.0.0",
-    "lerna": "2.0.0-beta.30",
+    "lerna": "2.0.0-beta.38",
```
bzang: Is lerna still here for publishing?

ianwremmel: yep. baby steps :)

CONTRIBUTING.md
ianwremmel: correct

deps.js
```diff
@@ -0,0 +1,201 @@
...
+      return walkDeps(`${filePath}.js`);
+    }
+    throw err;
+  }
+}
+
+function updateSinglePackage(packagePath) {
+  return findDepsForPackage(packagePath)
+    .then(depsToVersions(require(`./package.json`)))
+    .then(assignDeps(path.join(packagePath, `package.json`)));
```
ianwremmel: nope. behold the power of `curry`. `assignDeps` is a curried function, so, when it's called at line 125, it returns a function that'll be passed to `.then` expecting the other parameter.

circle.yml
```diff
@@ -14,9 +14,9 @@ general:
 
 test:
   pre:
-    - npm run deps:check
     - npm run grunt:circle -- build:
         parallel: true
+    - npm run deps:check
```
adamweeks: Why are we running it after the build?

packages/node_modules/@ciscospark/bin-sauce-connect/bin/sauce-connect
```diff
@@ -1,18 +1,19 @@
 #!/usr/bin/env node
 
-require('babel-register');
-var path = require('path');
-require('dotenv').config({
-  path: path.join(__dirname, '../../../.env')
+const path = require(`path`);
+require(`dotenv`).config({
+  path: path.join(__dirname, `../../../../../.env`)
```
adamweeks: Is the path needed? I thought `dotenv` automatically found it at the root

package.json
```diff
@@ -32,7 +32,7 @@
     "url": "http://sqbu-github.cisco.com:WebExSquared/squared-js-sdk.git"
   },
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "echo 'npm run bootstrap is no longer required. Hooray!'",
```
adamweeks: 👍

circle.yml
ianwremmel: because it doesn't handle import/export, so it needs to run against dist

.gitattributes
```diff
@@ -0,0 +1,3 @@
+packages/node_modules/* linguist-vendored=false
```
adamweeks: Does this actually work? If so, we should use it in the widget repo

packages/node_modules/@ciscospark/bin-sauce-connect/bin/sauce-connect
ianwremmel: because this command isn't necessarily run from the root. i think. you're right, but i also know i needed this at some point.  :)

.gitattributes
ianwremmel: nope

packages/node_modules/@ciscospark/bin-sauce-connect/bin/sauce-connect
adamweeks: Gotcha!

.gitattributes
adamweeks: Darn.

PR #499, opened 2017-03-28
fix(plugin-phone): Fix video/audio state issue on join

bzang:
This fixes issues where you cannot operate on the call once it's been answered.

### DIFF
--- a/packages/plugin-phone/src/call.js
+++ b/packages/plugin-phone/src/call.js
@@ -842,16 +842,13 @@ const Call = SparkPlugin.extend({
     }
 
     if (options.offerOptions) {
-      mediaOptions.offerOptions = {
-        offerToReceiveAudio: options.offerOptions.offerToReceiveAudio,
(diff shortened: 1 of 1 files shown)


packages/plugin-phone/src/call.js
```diff
@@ -832,22 +833,28 @@ const Call = SparkPlugin.extend({
...
-      if (options.offerOptions.offerToReceiveVideo) {
-        this.media.set({offerToReceiveVideo: options.offerOptions.offerToReceiveVideo});
-      }
+      mediaOptions.offerOptions = {
+        offerToReceiveAudio: options.offerOptions.offerToReceiveAudio,
+        offerToReceiveVideo: options.offerOptions.offerToReceiveVideo
+      };
     }
 
+    if (mediaOptions.offerOptions || mediaOptions.constraints) {
```
ianwremmel: you probably don't need this conditional; set should set it as an empty object and do nothing, but this might be a theoretical perf win.

PR #500, opened 2017-03-28
attempt to stabilize call tests

ianwremmel:


packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -429,57 +458,71 @@ describe(`plugin-phone`, function() {
...
-            return Promise.all([
-              call.when(`disconnected`)
-                .then(() => assert.equal(call.status, `disconnected`)),
-              mccoyCall.when(`disconnected`)
-                .then(() => assert.equal(mccoyCall.status, `disconnected`))
-            ]);
-          });
-      }));
+      it(`ends an in-progress call`, () => {
+        const handler = handleErrorEvent(spock.spark.phone.dial(mccoy.email), (call) => {
```
adamweeks: Do you need to `return handler` for the test to complete properly?

adamweeks: Nevermind, I see it *way* down there 👍

packages/plugin-phone/test/integration/spec/call.js
```diff
@@ -429,57 +458,71 @@ describe(`plugin-phone`, function() {
...
+          let mccoyCall;
+          const handler = Promise.all([
```
ianwremmel: that was a mistake. the Promise.all was just suppose to be returned

packages/plugin-phone/src/call.js
```diff
@@ -530,11 +542,19 @@ const Call = SparkPlugin.extend({
       }
     }
 
-    this.spark.phone.register()
+    this.logger.info(this.registered, !!this.spark.device.url, this.connected, this.spark.mercury.connected, this.spark.device.url);
+
+    // Note: mercury.connect() will call device.register() if need. We're not
+    // using phone.register() here because it guarantees a device refresh, which
+    // is probably unnecessary.
+    this.spark.mercury.connect()
```
bzang: if phone.register is called, can it handle not having a Client_secret and fail gracefully?

packages/plugin-phone/src/call.js
```diff
@@ -558,13 +578,13 @@ const Call = SparkPlugin.extend({
...
-    if (!this.locus) {
-      if (this.locusJoinInFlight) {
-        this.logger.info(`call: no locus, waiting for rest call to complete before hanging up`);
-        return this.when(`change:locus`)
-          .then(() => this.hangup());
-      }
+    if (this.locusJoinInFlight) {
+      this.logger.info(`call: locus join in flight, waiting for rest call to complete before hanging up`);
+      return this.when(`change:locusJoinInFlight`)
+        .then(() => this.hangup());
```
bzang: Doesn't seem like you would need to check if `locusJoinInFlight` is false, but could it still be true when a `change:locusJoinInFlight` happens?

packages/plugin-phone/src/call.js
ianwremmel: yea, the client_secret should never be required. downscope may emit warnings, but nothing that should propagate as a failure.

packages/plugin-phone/src/call.js
ianwremmel: line 581 states that join in flight is true, so if a change event fires, it can only become false.

packages/plugin-phone/src/call.js
ianwremmel: i think my pr to quiet those warnings is still waiting on a stable pipeline.

PR #502, opened 2017-03-29
v7-logoutlatency: Fixes logging out latency

saurjai3:


packages/spark-core/src/spark-core.js
```diff
@@ -210,12 +210,6 @@ const SparkCore = AmpState.extend({
   logout(...args) {
     return Promise.resolve()
       .then(() => {
-        if (this.mercury) {
-          return this.mercury.disconnect();
-        }
-        return Promise.resolve();
-      })
-      .then(() => {
```
saurjai3: The mercury disconnect is not needed since as soon as we disconnect, the web-client starts synchronizing due to a mercury state change and results in data fetches. As a result it slows down the logout process. Mercury would anyways be disconnected as soon as the logout is done and the client is refreshed.

ianwremmel: That sounds like a web client bug. this change would cause nodejs apps to never disconnect from mercury, even after calling logout(). mercury emits several different offline events. the web client is probably listening to the wrong one.

packages/spark-core/src/spark-core.js
```diff
@@ -226,7 +226,11 @@ const SparkCore = AmpState.extend({
         this.boundedStorage.clear(),
         this.unboundedStorage.clear()
       ]))
-      .then(() => this.credentials.logout(...args));
+      .then(() => {
+        this.trigger(`client:logout`);
```
ianwremmel: What is listening for this event?

packages/spark-core/src/spark-core.js
```diff
@@ -226,7 +226,11 @@ const SparkCore = AmpState.extend({
...
+        this.credentials.logout(...args);
+        return Promise.resolve();
```
ianwremmel: return the result of logout.

packages/spark-core/src/spark-core.js
saurjai3: We do a StorageAdapter.clear() on this event on the client side.

saurjai3:
opened new PR: https://github.com/ciscospark/spark-js-sdk/pull/549
hence closing this one.

PR #503, opened 2017-03-29
fix(board): ensure board creator does not get authorized to board KRO

tran2:
Currently, KMS server also authorizes the resource creator, in this case it adds
the board creator to board KRO on `createChannel`, which we don't want. We want only
conversation KRO to be in the list of authorization. 
As a result, board creator can have access to board encryption key even after leaving the conversation.

This PR checks if the board creator is authorized, if so remove the auth.
KMS is making changes to this, however, we don't know when that happens.

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-board/src/board.js
+++ b/packages/node_modules/@ciscospark/plugin-board/src/board.js
@@ -5,7 +5,7 @@
  */
 
 import {SparkPlugin, Page} from '@ciscospark/spark-core';
-import {base64} from '@ciscospark/common';
+import {base64, tap} from '@ciscospark/common';
 import Realtime from './realtime';
 import {assign, defaults, chunk, pick, some} from 'lodash';
 import querystring from 'querystring';
@@ -115,7 +115,8 @@ const Board = SparkPlugin.extend({
       resource: `/channels`,
       body: this._prepareChannel(conversation, channel)
     })
-      .then((res) => this._ensureOnlyConversationAuthorization(res.body));
+      .then((res) => res.body)
+      .then(tap((body) => this._ensureOnlyConversationAuthorization(body)));
   },
 
   _prepareChannel(conversation, channel) {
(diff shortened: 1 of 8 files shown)


packages/plugin-board/src/board.js
```diff
@@ -127,6 +130,28 @@ const Board = SparkPlugin.extend({
...
   },
 
+  _ensureOnlyConversationAuthorization(channel) {
+    if (!this._isChannelCreatorAuthorized(channel.creatorId, channel.kmsMessage)) {
+      return Promise.resolve(channel);
+    }
+
+    // remove channel creator from board service kms as we want to link to
+    // conversation KRO only
+    const kmsMessage = {
```
tran2: Good catch, Ian! The code is super nice after combining `process.nextTick` and `tap` 👍

packages/plugin-board/src/board.js
```diff
@@ -127,6 +131,28 @@ const Board = SparkPlugin.extend({
     }, channel);
   },
 
+  _ensureOnlyConversationAuthorization(channel) {
+    // keep this function out of the critical path of creating a board
+    process.nextTick(() => {
+      if (this._isChannelCreatorAuthorized(channel.creatorId, channel.kmsMessage)) {
+        // remove channel creator from board service kms as we want to link to
+        // conversation KRO only
+        const kmsMessage = {
```
ianwremmel: I should've seen this last week, but can you implement this in kms.js? There are already test stubs for `addAuthorization` and `removeAuthorization`, I just never got around to writing them...

packages/plugin-encryption/test/integration/spec/kms.js
```diff
@@ -46,13 +57,85 @@ describe(`Encryption`, function() {
...
+          assert.lengthOf(kro.authorizationUris, 1);
+        }));
+
+      it(`authorizes a user to a key`, () => spark.encryption.kms.addAuthorization({
+        userIds: [anotherUser.spark.device.userId],
+        resourceUri: kro.uri
+      })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, anotherUser.spark.device.userId);
```
ianwremmel: This should also assert that the other user can fetch the key

packages/plugin-encryption/test/integration/spec/kms.js
```diff
@@ -46,13 +57,85 @@ describe(`Encryption`, function() {
...
+          assert.equal(userAuth.authId, anotherUser.spark.device.userId);
+        }));
+
+      it(`deauthorizes a user from a key`, () => spark.encryption.kms.removeAuthorization({
+        authId: anotherUser.spark.device.userId,
+        resourceUri: kro.uri
+      })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, anotherUser.spark.device.userId);
```
ianwremmel: This should also assert that the other user can no longer fetch the key

packages/plugin-encryption/test/integration/spec/kms.js
```diff
@@ -15,8 +15,9 @@ describe(`Encryption`, function() {
   this.timeout(30000);
   describe(`KMS`, () => {
     let spark, user;
```
ianwremmel: can you name your users spock and mccoy to keep with the rest of the test suite? :)

tran2: sure :)

PR #504, opened 2017-03-30
Added the user agent to user plugin tests against the atlas activations endpoint

dgthistle:
To fix failing atlas integration tests

@ianwremmel please review!

### DIFF
--- a/packages/plugin-user/src/user.js
+++ b/packages/plugin-user/src/user.js
@@ -275,9 +275,10 @@ const User = SparkPlugin.extend({
         .catch((err) => Promise.reject(new Error(`failed to set authorization`, err)))
       )
       .then(() => {
-        if (options.spoofUserAgent) {
-          headers['user-agent'] = 'wx2-android';
+        if (typeof window === `undefined`) {
+          headers[`user-agent`] = `spark-js-sdk`;
         }
+
         return this.request({
           service: `atlas`,
           resource: `users/activations`,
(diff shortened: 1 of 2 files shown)


packages/plugin-user/src/user.js
```diff
@@ -274,15 +274,20 @@ const User = SparkPlugin.extend({
...
-        service: `atlas`,
-        resource: `users/activations`,
-        method: `POST`,
-        headers,
-        body: options,
-        shouldRefreshAccessToken,
-        requiresClientCredentials
-      }))
+      .then(() => {
+        if (options.spoofUserAgent) {
```
ianwremmel: how about

```
if (typeof window === `undefined` && ! header[`user-agent`]) {
  header[`user-agent`] = `spark-js-sdk`
}
```

I don't like the idea of masquerading as a different user agent type and I really don't like the idea of being able to opt into it arbitrarily from production code.

dgthistle: Sure. Although, is the user-agent header set from this point in the code when run from the browser? Or does that happen in a lower level browser call?

dgthistle: nevermind, I see you are just checking the headers in case someone sets it somewhere. headers is initialized just above so I don't think that is necessary.

packages/plugin-user/src/user.js
```diff
@@ -274,15 +274,21 @@ const User = SparkPlugin.extend({
...
-        headers,
-        body: options,
-        shouldRefreshAccessToken,
-        requiresClientCredentials
-      }))
+      .then(() => {
+        if (typeof window === `undefined`) {
+          headers[`user-agent`] = `spark-js-sdk`;
+        }
+
```
dgthistle: @ianwremmel good?

dgthistle: No need to check headers[`user-agent`] since headers is initialized right above.

PR #506, opened 2017-03-30
robust close codes

ianwremmel:
This puts us in position to easily add support for additional codes once we have a complete plan of action for how to handle multiple sockets

Note: commented-out code blocks are intentional; the NotFound close code doesn't quite work as intended, but the rest of these changes are still helpful without it.


ianwremmel:
@tran2 do you see anything in here that'll impact board?

PR #512, opened 2017-04-03
lint commit messages

ianwremmel:


ianwremmel:
the unlinted commit is just the version number when publishing a new version. long-term, yes, that should probably fit a pattern, but I'm not sure it should be the changelog pattern. i figure that's a boat we can rock when we figure out how to specify major/minor/patch :)

ianwremmel:
@adamweeks, yes, we've been using the angular conventions, though we should probably copy over the new ones. I used to link their document, but they kept moving it.

PR #513, opened 2017-04-03
feat(plugin-wdm): Added an interceptor for aphelia/subscription service to add required header details

saurjai3:
v7-subscription: Added an interceptor for aphelia/subscription service to add required header details

### DIFF
--- a/packages/plugin-wdm/src/interceptors/device-url.js
+++ b/packages/plugin-wdm/src/interceptors/device-url.js
@@ -1,0 +1,43 @@
+/**!
+*
+* Copyright (c) 2015-2016 Cisco Systems, Inc. See LICENSE file.
+*/
+
+import {Interceptor} from '@ciscospark/http-core';
+
+/**
+* @class
+*/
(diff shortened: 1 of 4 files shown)


packages/plugin-wdm/src/interceptors/device-url.js
```diff
@@ -0,0 +1,43 @@
...
+  /**
+  * Only add the device url header if (a) we have a device url to add and
+  * (b) if there are already other headers in place that will trigger a CORS
+  *     preflight and
+  * (c) we are not calling CI.
+  * @param {Object} options
+  * @returns {boolean}
+  */
+  requiresDeviceUrl(options) {
+    return !!(this.spark.device && this.spark.device.url) && Object.keys(options.headers).length > 0 && options.api !== `oauth` && options.api !== `saml`;
```
ianwremmel: use `service`, not `api`

ianwremmel:
Also, a unit test or two would be good.

PR #514, opened 2017-04-04
	feat: use new ui for authorization flow 

kaprenikhil:
Change a request parameter for authorization request. This allow to display new ui for login page which is in sync with Spark Clients.

Here is login page after the change.

screen shot 2017-04-04 at 1 45 18 pm


kaprenikhil:
I updated commit. Do you want me to open new PR ?

ianwremmel:
no need for a new PR.

kaprenikhil:
Thanks, do I need to push anywhere ?

aimeex3:
err is this correct? we're modifying node_modules for plugin-credentials? also the code looks for `service` as the config parameter:
https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/plugin-credentials/src/credentials/common.js#L90
and
https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/plugin-credentials/src/credentials/common.js#L123-L124
We would have to modify the code as well if we're modifying the config parameter

kaprenikhil:
Looks like there is another config, which must have been used for the code you mentioned - https://github.com/ciscospark/spark-js-sdk/blob/master/src/defaults.js

kaprenikhil:
Do you want me to revert the change I made ? May be its not being used.

ianwremmel:
@aimeex3, yes all code is now in `./packages/node_modules` instead of `./packages`. This allows us to avoid the `npm run bootstrap` step. 

I wouldn't worry too much about backing things out unless they're breaking something. Both credentials plugins are going to be largely replaced in the coming weeks.

PR #515, opened 2017-04-05
feat(tooling): Updated lerna to use exact versions

bzang:
Setting the exact versions on build will allow us to more reliably peg to a specific version of the SDK without new builds affecting parent projects.


adamweeks:
Will this still work with the alle stuff @ianwremmel is working on?

ianwremmel:
@adamweeks, it should, yes

bzang:
This is not necessary anymore as the new `deps.js` only uses exact versions in generating `package.json` files for packages.

PR #517, opened 2017-04-05
fix(logout): resolve promise if device.url and supertoken not defined

aimeex3:


src/client/credentials/authorization.js
```diff
@@ -149,7 +149,7 @@ var Authorization = SparkBase.extend({
         }.bind(this))
     ])
       .then(function revokeSupertoken() {
-        return supertoken.revoke()
+        return supertoken && supertoken.revoke()
```
ianwremmel: How do we get into this state?

aimeex3: i think sometimes we call logout() when the user is already logged out. this happens sometimes when we get a sideboarded link and we need to log out the previous user out, but the device and supertoken is already unregistered/revoked

PR #524, opened 2017-04-10
Exclusively use npm scripts for running static analysis

ianwremmel:


ianwremmel:
I'm pretty sure it's failing because I never created the `lint` task, only the `lint:*` subtasks. I'll fix that :)

PR #527, opened 2017-04-11
chore(tooling): detect next version using git history and npm version

ianwremmel:
Allows manualy version specification via commit message
Infer change type using git history


tooling/version.sh
```diff
@@ -0,0 +1,34 @@
...
+  # Assume this is a patch release
+  NEXT_VERSION_TYPE=patch
+  # If there are breaking changes, make it a minor bump. post-1.0, this logic
+  # will need to be revised
+  BREAKING_COUNT=$(git log --oneline upstream/master.. | grep -c '$BREAKING CHANGE:')
+  if [ $BREAKING_COUNT -gt 0 ]; then
+    NEXT_VERSION_TYPE=minor
+  fi
+
+  # Now that we knw the type, do some silly things to make npm version work
```
bzang: missing `o` in `know`

tooling/version.sh
```diff
@@ -0,0 +1,34 @@
...
+  #
+  # Figure out the current version
+  CURRENT_VERSION=$(jq -r .version lerna.json)
```
bzang: Looks liek `jq` is only available in the docker container. Is that ok?

ianwremmel: it's also on the slave. the docker file that defines it is at https://sqbu-github.cisco.com/WebExSquared/jenkins-slave-containers/blob/master/slaves/spark-js-sdk/Dockerfile

adamweeks:
Looks good to me. Do we want to mention anything in any of the docs about the `#release` tag?

ianwremmel:
we can document it after we go full semantic release. for now, it's something i'd prefer people don't know about. generally, speaking, using it is a bug.

Jenkinsfile
```diff
@@ -387,7 +387,13 @@ ansiColor('xterm') {
                   }
 
                   try {
-                    sh script: "npm run lerna --silent -- publish --skip-npm --skip-git  --yes --repo-version=${version}"
+                    def publishScript = "npm run lerna --silent -- publish --skip-npm --skip-git  --yes --repo-version=${version}"
```
bzang: Just in case lerna overwrites the exact version numbers in the deps.js generated package.json, we should probably add `--exact` to the publish call here.

ianwremmel: done.

PR #530, opened 2017-04-12
test(plugin-phone): add test for media state after disconnect

adamweeks:
This is the failing test that describes the issue we are seeing with the widget upon remote party disconnect.

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/call.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/call.js
@@ -13,7 +13,7 @@ import testUsers from '@ciscospark/test-helper-test-users';
 import handleErrorEvent from '../lib/handle-error-event';
 import {SparkHttpError} from '@ciscospark/spark-core';
 
-import {assertLocusMediaState} from './media-state-controls';
+import assertLocusMediaState from '../lib/assert-locus-media-state';
 
 describe(`plugin-phone`, function() {
   this.timeout(30000);
@@ -189,7 +189,7 @@ describe(`plugin-phone`, function() {
           ]));
         });
 
-        it(`is not sending media`, () => {
+        it(`releases local video and audio`, () => {
           call = spock.spark.phone.dial(mccoy.email);
           assert.equal(call.status, `initiated`);
           return handleErrorEvent(call, () => Promise.all([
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/media-state-controls.js
```diff
@@ -14,7 +14,7 @@ import {skipInFirefox} from '@ciscospark/test-helper-mocha';
 
 import {boolToStatus} from '../..';
 
-function assertLocusMediaState(call, {
+export function assertLocusMediaState(call, {
```
ianwremmel: If we're going to use this in multiple files, let's move it up to the `lib` directory.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/call.js
```diff
@@ -186,6 +188,22 @@ describe(`plugin-phone`, function() {
               .then(() => assert.equal(call.status, `disconnected`))
           ]));
         });
+
+        it(`is not sending media`, () => {
```
ianwremmel: This is to prove that the camera isn't still sending after hangup, right? Can you make the description a little clearer about that goal?

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/call.js
```diff
@@ -186,6 +188,22 @@ describe(`plugin-phone`, function() {
...
+          call = spock.spark.phone.dial(mccoy.email);
+          assert.equal(call.status, `initiated`);
+          return handleErrorEvent(call, () => Promise.all([
+            mccoy.spark.phone.when(`call:incoming`)
+              .then(([c]) => c.reject()),
+            call.when(`disconnected`)
+              .then(() => assertLocusMediaState(call, {
+                sendingAudio: false,
```
ianwremmel: For this set of assertions, I think you might actually want to try to read the state of the streams in the peer connection. Isn't the problem kinda that the `call` thinks it has hungup correctly, but the camera hasn't been released?

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/media-state-controls.js
adamweeks: Yea, that works.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/call.js
adamweeks: I'm not sure I follow on this one. Isn't that what `assertLocusMediaState` does? What would I check otherwise? `call.media.peer.something`?

ianwremmel: yea, assertLocusMediaState makes assertions about the values in the current locus DTO, but not necessarily the state of the local peer connection. I think for this test, you'll need to make assertions around `call.media.peer`.

ianwremmel:
can this be closed?

PR #531, opened 2017-04-12
docs(general): Update README and CONTRIBUTING

bbender:
Update `README.md` to streamline the getting started information and move build and test instructions into `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,10 +58,18 @@ Install tooling dependencies with:
 npm install
 ```
 
-Specify which package to build by using the `PACKAGE` environment variable. To build the SDK, use `ciscospark`:
+Build the SDK:
 
 ```bash
 npm run build
+```
+
+#### Building Packages Individually
+
+To build packages individually, specify the package name in an environment variable, `PACKAGE`, and run: 
+
+```bash
+PACKAGE=PACKAGENAME npm run grunt:package -- build
 ```
 
 ### Running Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,42 +48,59 @@ You can get these values by registering a new integration on the [Spark for Deve
 
 Fork the [spark-js-sdk](https://github.com/ciscospark/spark-js-sdk/) repository and `git clone` your fork:
 
-    git clone https://github.com/your-username/spark-js-sdk.git
+```bash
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,20 +1,96 @@
...
+- `CISCOSPARK_REDIRECT_URI`
+- `CISCOSPARK_SCOPE`
+
+You can get these values by registering a new integration on the [Spark for Developers](https://developer.ciscospark.com/add-integration.html) portal.
+
+### Building the SDK
+
+Fork the [spark-js-sdk](https://github.com/ciscospark/spark-js-sdk/) repository and `git clone` your fork:
+
+    git clone https://github.com/your-username/spark-js-sdk.git
```
ianwremmel: It looks like you're missing backticks for these commands.

CONTRIBUTING.md
```diff
@@ -1,20 +1,96 @@
...
+
+Install tooling dependencies with:
+
+    npm install
+
+Specify which package to build by using the `PACKAGE` environment variable. To build the entire SDK, use `ciscospark`:
```
ianwremmel: The "entire" sentence is not accurate.

CONTRIBUTING.md
```diff
@@ -1,20 +1,96 @@
...
+
+    PACKAGE=PACKAGENAME npm run grunt:package -- express:test test:node
+
+#### Run Browser Tests
+Keeps the browser open so that you can reload set break points and reload the page
+
+    KARMA_DEBUG=true PACKAGE=PACKAGENAME npm run grunt:package -- express:test test:browser
+
+### Git Commit Guidelines
+
```
ianwremmel: I just discovered https://conventionalcommits.org. We should try to rely on that when possible, but I haven't read through all of it yet.

CONTRIBUTING.md
bbender: The `git clone` command will be shown as a code block, though it's not fenced by back-ticks.

CONTRIBUTING.md
bbender: What would be a better command to build the SDK without building each package individually?

bbender: What would be a better command to build the SDK without building each package individually?

bbender:
Merged 26-Apr-2017.

PR #538, opened 2017-04-15
npm-build-docs

ianwremmel:
- [x] fix lint errors
- [x] detect current version number
- [x] do not document WebRTCMedia class methods as bare functions


package.json
```diff
@@ -51,7 +51,10 @@
     "commitmsg": "npm run --silent lint:commitmsg",
     "bootstrap": "echo 'npm run bootstrap is no longer required. Hooray!'",
     "clean": "lerna clean",
-    "build": "./tooling/build.js",
+    "build": "npm run --silent build:js && npm run --silent build:doc:html",
+    "prebuild:doc:html": "rimraf ./docs/api",
+    "build:doc:html": "documentation build packages/node_modules/{ciscospark,@ciscospark/plugin-phone}/src/index.js -f html -o ./docs/api --sort-order alpha",
```
adamweeks: does this only build for `@ciscospark/plugin-phone`?

ianwremmel: and the ciscospark package. those are the only "public" pieces of the sdk right now. HttpCore and SparkCore should get better docs, then they'll be added, too. The other plugins should not be on the docs site.

ianwremmel:
This is very out of date

PR #542, opened 2017-04-16
npm test

ianwremmel:
Over half of the changes here are the deletion of `*/test/index.js` and adjusting the location of imports accordingly. 

[this diff](https://github.com/ciscospark/spark-js-sdk/compare/2006835a56cd4e1615df425c32cfe1b6cdbaf96c...e29ea85ae97dea96e211f6e50cddf047329db069) should make review much easier by omitting thos changes.


ianwremmel:
Known issues:
- [ ] SIGINT doesn't propagate to the karma server
- [ ] There's a retain cycle somewhere that'll wedge the suite open. It looks like it has something to do with the batcher. The `retrieves an avatar url` unit test is enough to reproduce.

karma-ng.conf.js
```diff
@@ -99,23 +109,31 @@ module.exports = function configureKarma(config) {
...
     };
 
-    cfg.browserify.transform.unshift([`browserify-istanbul`, {
-      ignore: [`test-helper*/**`, `**/dist/**`],
-      instrumenter: require(`isparta`)
-    }]);
+    // cfg.browserify.transform.unshift([`browserify-istanbul`, {
+    //   instrumenter: require(`isparta`),
+    //   defaultIgnore: false
+    // }]);
```
bzang: Is browserify still being used?

ianwremmel: yes, but i can't get karma to generate coverage reports, hence the comments.

PR #545, opened 2017-04-17
V7.1 autologout: Fixes issue with forced logout even when in cisco network

saurjai3:
Ref: https://github.com/ciscospark/spark-js-sdk/pull/488

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-wdm/src/device/device.js
+++ b/packages/node_modules/@ciscospark/plugin-wdm/src/device/device.js
@@ -250,6 +250,10 @@ const Device = SparkPlugin.extend({
 
       const timer = setTimeout(() => {
         this.spark.request({
+          headers: {
+            'cisco-no-http-redirect': null,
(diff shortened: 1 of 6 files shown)


packages/node_modules/@ciscospark/plugin-wdm/src/device/device.js
```diff
@@ -250,6 +250,10 @@ const Device = SparkPlugin.extend({
 
       const timer = setTimeout(() => {
         this.spark.request({
+          headers: {
```
ianwremmel: I think changes went in that mean you shouldn't need to explicitly omit these headers. try running your tests without this block.

saurjai3: It seems to run the test cases after removing this block. Hence removing these checks.

ianwremmel:
can this be closed?

saurjai3:

PR #546, opened 2017-04-17
V7.1-contentshare : fixes broken images on content share

saurjai3:


packages/node_modules/@ciscospark/plugin-conversation/src/conversation.js
```diff
@@ -469,10 +469,16 @@ const Conversation = SparkPlugin.extend({
   /**
    * Creates a ShareActivty for the specified conversation
    * @param {Object} conversation
+   * @param {Object} activity
    * @returns {ShareActivty}
    */
-  makeShare(conversation) {
-    return ShareActivity.create(conversation, null, this.spark);
+  makeShare(conversation, activity) {
```
ianwremmel: The idea behind `makeShare` is that the value it returns *is* the provisional activity. Why can't the web client use `makeShare` instead of a bare object?

packages/node_modules/@ciscospark/plugin-conversation/src/process-image.browser.js
```diff
@@ -54,7 +54,7 @@ export function computeDimensions({width, height}, maxWidth, maxHeight) {
  * @returns {Promise<Array>} Buffer, Dimensions, thumbnailDimensions
  */
 export default function processImage({file, thumbnailMaxWidth, thumbnailMaxHeight, enableThumbnails}) {
-  if (!file.type.startsWith(`image`)) {
+  if (!file.type || !file.type.startsWith(`image`)) {
```
ianwremmel: might be more readable as `if (file.type && !file.type.startsWith('image')) {`

ianwremmel:
@bzang does any of this look like it'll cause problems with the widgets?

bzang:
From a widget perspective, this shouldn't cause any issues.

packages/node_modules/@ciscospark/plugin-conversation/src/process-image.browser.js
saurjai3: Above change would not resolve the promise if file.type is undefined.

packages/node_modules/@ciscospark/plugin-conversation/src/conversation.js
saurjai3: It needs to create the activity using the provisional activity created at the web-client passed as the second param, else the clientTempIds won't match and would result in a broken image on the clients.

packages/node_modules/@ciscospark/plugin-conversation/src/process-image.browser.js
ianwremmel: ah, right.

packages/node_modules/@ciscospark/plugin-conversation/src/conversation.js
ianwremmel: Why can't the result of `makeShare()` *be* the provisional activity? When josh and I original talked about `makeShare()`, I'm pretty sure that was the plan and that was why we intentionally left the activity param out of `makeShare()`

saurjai3: I am not comfortable making that change on the web-client side as it appears the plan was not documented anywhere. At this point it would help if we can just go with the above changes and may be revisit it after the sdk v7 upgrade. Please suggest.

ianwremmel: Fine. Make sure you have tests around this in the web client because we'll need to revisit and introduce breaking changes to get `ShareActivity` cleaned up.

ianwremmel:
can this be closed?

saurjai3:

PR #550, opened 2017-04-17
V7.1 textwithimage: displays text when posted with image uploads

saurjai3:


packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
```diff
@@ -233,9 +233,9 @@ const ShareActivity = SparkPlugin.extend({
       verb: `share`,
       object: {
         objectType: `content`,
-        displayName: this.displayName,
-        content: this.content,
-        mentions: this.mentions,
+        displayName: this.object && this.object.displayName ? this.object.displayName : ``,
```
ianwremmel: These properties are defined via the AmpState props hash, so they should be read directly from `this`, not from `this.object`.

ianwremmel:
@bzang does any of this look like it'll cause problems with the widgets?

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
```diff
@@ -333,8 +333,8 @@ ShareActivity.create = function create(conversation, object, spark) {
     parent: spark
   });
 
-  if (files) {
-    files.forEach((file) => share.add(file));
+  if (files && files.items) {
+    files.items.forEach((file) => share.add(file));
```
bzang: Is there a concern that this may break backwards compatibility for others using `.create()` if they expect the `files` object to not have a separate `items` object? As I understand it, this is how the `conversation` service returns files in the `activities` list.

bzang:
We're using some of the methods, but overall this should not affect the widgets since we'll manually calling `share.add()` and forming our own activity and share objects.

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
saurjai3: right bzang, The activities returned from conversation service has files.items. Do you suggest any changes for backward compatibility or are we good with this PR?

ianwremmel: the problem here is this change breaks the interface to makeShare(). It should probably be something like 
```js
const files = files.items || files
files.forEach(...)
```

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
saurjai3: screen shot 2017-04-18 at 9 56 05 am

saurjai3: not sure what changes do you suggest? As per the screenshot the displayName is within this.object.displayName here.

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
saurjai3: sounds good. will change it as recommended above.

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
ianwremmel: I'm suggesting that `ShareActivity#object` should never be defined. When we started down the `ShareActivity` path, the intent was to have all of the client-settable properties of the activity be top-level properties of the `ShareActivity` instance.

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
```diff
@@ -333,9 +333,8 @@ ShareActivity.create = function create(conversation, object, spark) {
     parent: spark
   });
 
-  if (files) {
-    files.forEach((file) => share.add(file));
-  }
+  files = files.items || files;
+  files.items.forEach((file) => share.add(file));
```
ianwremmel: files can't have an `items` property at this point.
you still need to confirm files exists before referencing its properties.

packages/node_modules/@ciscospark/plugin-conversation/src/share-activity.js
saurjai3: At this point we should go with this change and may be revisit the ShareActivity later again.

ianwremmel: Please consider this an unstable API and make sure you have tests for it.

ianwremmel:
can this be closed?

saurjai3:

PR #552, opened 2017-04-18
feat(plugin-people): move plugin out of ciscospark

adamweeks:


packages/node_modules/@ciscospark/plugin-people/src/people.js
```diff
@@ -51,9 +51,13 @@ const People = SparkPlugin.extend({
    *   // => success
    */
   get(person) {
+    if (!person) {
```
adamweeks: @ianwremmel git "moved" this file instead of deleting it, so it appears that it isn't deleted

adamweeks:
@ianwremmel I considered it an initial implementation of the standalone plugin, but I'll rename the PR.

packages/node_modules/@ciscospark/plugin-people/src/people.js
ianwremmel: ah, gotcha. i feel like that used to be more obvious...

ianwremmel: Is that true for all of the test files?

adamweeks: There was only one integration test, but I had actually written all the tests out before I found it, so it looks like git didn't think of it as a move.

PR #553, opened 2017-04-19
fix(@ciscospark/plugin-phone): address race in toggleSending

ianwremmel:
This removes a race condition in toggleSendingAudio/toggleSendingVideo. A future PR will make this more efficient

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/src/call.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/call.js
@@ -375,7 +375,7 @@ const Call = SparkPlugin.extend({
           sdp: offer,
           mediaId: this.mediaId
         }))
-        .then(() => this._fetchExpectedLocus())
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -937,31 +937,42 @@ const Call = SparkPlugin.extend({
   },
 
   /**
-   * The response to a PUT to LOCUS/media may not be fully up-to-dat when we
+   * The response to a PUT to LOCUS/media may not be fully up-to-date when we
```
adamweeks: Did you want to put the prop in the jsdoc? No biggie if it is going to change soon

ianwremmel: I'll get it on the next pass.

PR #554, opened 2017-04-19
test(plugin-board): create new conversation for channel listing test

tran2:
Sometimes karma runs test multiple times on the same browser in
parallel, which creates more channels than expected for one
conversation. This change should protect these tests from mentioned race
condition.

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-board/test/integration/spec/board.js
+++ b/packages/node_modules/@ciscospark/plugin-board/test/integration/spec/board.js
@@ -179,44 +179,54 @@ describe(`plugin-board`, () => {
       });
 
       it(`retrieves all boards for a specified conversation across multiple pages`, () => {
-        let existingChannelsCount = 0;
-        let numChannelsToAdd = 0;
-        const pageLimit = 50;
+        const numChannelsToAdd = 12;
+        const pageLimit = 5;
+        const channelsCreated = [];
+        let channelsReceived = [];
+        let convo;
 
-        return participants[0].spark.board.getChannels(conversation, {
-          channelsLimit: 100
+        return participants[0].spark.conversation.create({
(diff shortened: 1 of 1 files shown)


packages/node_modules/@ciscospark/plugin-board/test/integration/spec/board.js
```diff
@@ -179,44 +179,54 @@ describe(`plugin-board`, () => {
...
-            return participants[0].spark.board.getChannels(conversation, {
+            return participants[0].spark.board.getChannels(convo, {
               channelsLimit: pageLimit
             });
           })
-
           // get boards, page 2
           .then((channelPage) => {
             assert.lengthOf(channelPage.items, pageLimit);
             assert(channelPage.hasNext());
```
ianwremmel: Can you clean this up so it doesn't use bare assertions anywhere? assert.isTrue in particular?

packages/node_modules/@ciscospark/plugin-board/test/integration/spec/board.js
```diff
@@ -179,44 +179,54 @@ describe(`plugin-board`, () => {
...
+            assert(channelPage.hasNext());
+            channelsReceived = channelsReceived.concat(channelPage.items);
             return channelPage.next();
           })
           .then((channelPage) => {
-            assert.lengthOf(channelPage, 1);
+            assert.lengthOf(channelPage, 2);
             assert(!channelPage.hasNext());
+            channelsReceived = channelsReceived.concat(channelPage.items);
+            assert.sameDeepMembers(channelsCreated, channelsReceived);
```
ianwremmel: does sameDeepMembers allow for two arrays to have the same set of items, but in a different order?

tran2: yup, these two arrays have different order

PR #557, opened 2017-04-21
feat(plugin-lyra): add lyra api implementations

tran2:


packages/node_modules/@ciscospark/plugin-lyra/src/spaces.js
```diff
@@ -0,0 +1,393 @@
...
+import {SparkPlugin} from '@ciscospark/spark-core';
+import querystring from 'querystring';
+import {base64} from '@ciscospark/common';
+
+/**
+ * @class
+ * @extends {Lyra}
+ * @memberof Lyra
+ */
+const Spaces = SparkPlugin.extend({
```
ianwremmel: If I'm reading this right, I think this should be split into two plugins:
- space
- device

There shouldn't be any need to include "Space" in method names since the methods are attached to the space plugin.

packages/node_modules/@ciscospark/plugin-lyra/test/integration/spec/spaces.js
```diff
@@ -0,0 +1,255 @@
...
+      .then((space) => {
+        lyraMachine.space = space;
+      }));
+
+    before(`create users`, () => testUsers.create({count: 2})
+      .then((users) => {
+        participants = users;
+        spock = participants[0];
+
+        return Promise.all(map(participants, (participant) => {
```
ianwremmel: avoid `lodash.map()` when mapping arrays; just use `Array#map()`

packages/node_modules/@ciscospark/plugin-lyra/test/integration/spec/spaces.js
```diff
@@ -0,0 +1,255 @@
...
+      displayName: `Test Lyra Conversation`,
+      participants
+    })
+      .then((c) => {
+        conversation = c;
+        return conversation;
+      }));
+
+    describe(`#list()`, () => {
+      it(`returns spaces`, () => assert.isFulfilled(spock.spark.lyra.spaces.list()));
```
ianwremmel: maybe assert it returns an array (with an entry?).

tran2:
Lyra service is a service coordinating devices that act together in a physical meeting room or collaboration space. For the web client, a user can manually pair to a Cisco Spark Board for call/volume control from web.

packages/node_modules/@ciscospark/plugin-lyra/src/lyra.js
```diff
@@ -0,0 +1,58 @@
...
+      resource: `/service_health`
+    })
+      .then((res) => res.body);
+  },
+
+  /**
+   * Gets lyra build info
+   * @returns {Promise<Object>} buildInfo response body
+   */
+  getBuildInfo() {
```
tran2: I guess we don't. Let me remove this.

packages/node_modules/@ciscospark/plugin-lyra/src/spaces.js
tran2: makes sense, let me refactor

packages/node_modules/@ciscospark/plugin-lyra/src/lyra.js
```diff
@@ -0,0 +1,58 @@
...
+  namespace: `Lyra`,
+  children: {
+    spaces: Spaces
+  },
+
+  /**
+   * Pings lyra
+   * @returns {Promise<Object>} ping response body
+   */
+  ping() {
```
tran2: @saurjai3 what do you think? Do we need this?

packages/node_modules/@ciscospark/plugin-lyra/src/occupants.js
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Joins a lyra space, update every 10 minutes to keep alive for MANUAL
+   * @param {Types~LyraSpace} space
+   * @param {string} space.url
+   * @param {object} options
+   * @param {string} options.passType
+   * @returns {Promise}
+   */
+  join(space, options = {passType: `MANUAL`}) {
```
ianwremmel: if join and leave act on a space and manipulate the current user, shouldn't these methods be in the space class?

ianwremmel: I haven't reviewed the whole thing yet, but I want to get this addressed before reading through the rest.

packages/node_modules/@ciscospark/plugin-lyra/src/occupants.js
```diff
@@ -0,0 +1,178 @@
...
+ */
+const Occupants = SparkPlugin.extend({
+  namespace: `Lyra`,
+
+  /**
+   * Joins a lyra space, update every 10 minutes to keep alive for MANUAL
+   * @param {Types~LyraSpace} space
+   * @param {string} space.url
+   * @param {object} options
+   * @param {string} options.passType
```
tran2: It's how a user binds to a Spark Board, there's `MANUAL`, `ULTRASOUND` (proximity)

packages/node_modules/@ciscospark/plugin-lyra/src/occupants.js
tran2: @ianwremmel, the join is basically the user adds himself to the space (when a user manually adds a Spark Board from the web client). I think it makes more sense to call `spark.lyra.occupants.join(space)` than `spark.lyra.spaces.join(space)`.

ianwremmel: `spark.lyra.spaces.join(space)` would be consistent with the conversation plugin. I would expect the first parameter of methods of the occupants plugin to be an `occupant`.

packages/node_modules/@ciscospark/plugin-lyra/src/occupants.js
ianwremmel: gotcha

packages/node_modules/@ciscospark/plugin-lyra/src/occupants.js
tran2: Oh ok, let me make the changes :)

packages/node_modules/@ciscospark/plugin-lyra/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ * @private
+ */
+
+import '@ciscospark/plugin-mercury';
+import '@ciscospark/plugin-encryption';
+import '@ciscospark/plugin-conversation';
+import '@ciscospark/plugin-feature';
+import '@ciscospark/plugin-locus';
```
tran2: nope, I'm moving it to the integration test that needed it instead

packages/node_modules/@ciscospark/plugin-lyra/test/integration/spec/device.js
```diff
@@ -0,0 +1,170 @@
...
+            correlationId: locus.correlationId,
+            deviceUrl: spock.spark.device.url,
+            localMedias: []
+          }
+        });
+      });
+
+      after(`remove binding`, () => spock.spark.lyra.space.unbindConversation(lyraMachine.space, conversation));
+
+      it(`allows muting`, (done) => {
```
ianwremmel: Please change these the imperative form "it mutes", "it unmutes", "it increases the volume", etc

packages/node_modules/@ciscospark/plugin-lyra/test/integration/spec/device.js
```diff
@@ -0,0 +1,170 @@
...
+        return spock.spark.lyra.device.decreaseVolume(lyraMachine.space);
+      });
+
+      it(`allows setting volumne`, (done) => {
+        lyraMachine.spark.mercury.once(`event:lyra.space_audio_volume_set_action`, (event) => {
+          assert.equal(event.data.level, 2);
+          done();
+        });
+
+        return spock.spark.lyra.device.setVolume(lyraMachine.space, 2);
```
ianwremmel: I think if you're using the done callback, you shouldn't return, but I'm not sure.

ianwremmel: There's a helper in the phone plugin for waiting for events in a promisy fashion

tran2: are you talking about using `.when` instead of `.on`?

tran2:
ah thanks, I'll remember to use that next time :)

PR #560, opened 2017-04-24
fix(plugin-feature): make setBundledFeatures merge each toggle object

nickclar:


packages/node_modules/@ciscospark/plugin-feature/src/feature.js
```diff
@@ -81,7 +81,8 @@ const Feature = SparkPlugin.extend({
       resource: `features/users/${this.spark.device.userId}/toggles`,
       body: featureList
     })
-      .then((res) => this.spark.device.features[`user`].add(res.body, {merge: true}));
+      .then((res) => res.body.featureToggles.forEach((featTog) => this.spark.device.features.user.add(featTog, {merge: true})))
```
ianwremmel: There's no guarantee it's a user toggle, right? don't you need to use a property of the `featTog` to determine which collection to add it to?

Also, I *think* there should be a way to bulk add the toggles instead of one at a time; that should reduce the number of change events fired.

packages/node_modules/@ciscospark/plugin-feature/test/integration/spec/feature.js
```diff
@@ -71,9 +72,12 @@ describe(`plugin-feature`, function() {
     }];
 
     it(`sets a value for two user feature toggle`, () => {
+      spark.feature.setFeature(`user`, `key1`, false);
       return spark.feature.setBundledFeatures(featureUpdateArray)
         .then((res) => {
           assert.isDefined(res);
+          assert.equal(find(res.serialize().user, {key: `key1`}).value, `value1`);
+          assert.equal(find(res.serialize().user, {key: `key2`}).value, `value2`);
           assert.becomes(spark.feature.getFeature(`user`, `key1`), `value1`);
```
ianwremmel: These assert.becomes need to return to provide any values. Looks like it slipped through before, but let's get it fixed while you're looking at it. 

```js
return Promise.all([
  assert.becomes(spark.feature.getFeature(`user`, `key1`), `value1`),
  assert.becomes(spark.feature.getFeature(`user`, `key2`), `value2`)
]);
```

packages/node_modules/@ciscospark/plugin-feature/src/feature.js
```diff
@@ -81,7 +84,12 @@ const Feature = SparkPlugin.extend({
       resource: `features/users/${this.spark.device.userId}/toggles`,
       body: featureList
     })
-      .then((res) => this.spark.device.features[`user`].add(res.body, {merge: true}));
+      .then((res) => {
+        const partitionedToggles = partition(res.body.featureToggles, {type: `USER`});
+        this.spark.device.features.user.add(partitionedToggles[0], {merge: true});
+        this.spark.device.features.developer.add(partitionedToggles[1], {merge: true});
+        return this.spark.device.features;
```
ianwremmel: I'm not sure we want to return the features object here. Ideally, developers are only accessing feature values via `spark.features.getFeature`. If we encourage raw access, I could imagine things might get changed inappropriately. Is this return value getting used somewhere?

nickclar: we're not using the return value anywhere critical so I can make that change

packages/node_modules/@ciscospark/plugin-feature/test/integration/spec/feature.js
nickclar: ok. And I'll have to remove the other assert.equal statements since the function will no longer return a response.

ianwremmel:
can this be closed?

nickclar:
Yes, this can be closed, it was merged in this larger PR https://github.com/ciscospark/spark-js-sdk/pull/589

PR #562, opened 2017-04-25
Fix async media

ianwremmel:
Smaller than it looks, the first commit just alphabetizes call.js.

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/src/call.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/call.js
@@ -36,6 +36,8 @@ import boolToStatus from './bool-to-status';
 import WebRTCMedia from './web-rtc-media';
 import uuid from 'uuid';
 
+import {parse} from 'sdp-transform';
+
 /**
  * @event ringing
  * @instance
@@ -506,7 +508,6 @@ const Call = SparkPlugin.extend({
 
     this.listenTo(this.media, `negotiationneeded`, debounce(() => {
       this.media.createOffer()
-        .then(tap(() => this.logger.info(`UPDATE MEDIA NEGOTIATION NEEDED`)))
         .then(() => this._updateMedia({
           sdp: this.media.offerSdp,
           mediaId: this.mediaId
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -8,7 +8,13 @@
 /* global RTCPeerConnection, RTCSessionDescription */
 
 import {SparkPlugin} from '@ciscospark/spark-core';
-import {base64, oneFlight, retry, tap, whileInFlight} from '@ciscospark/common';
+import {
+  base64,
```
bzang: wasn't there an issue with the base64 encoder we were using here a while back?

ianwremmel: yea, but only for a very specific use case; that was when converting an arraybuffer into a base64 string to put in an image tag. base64url didn't work for that case, but generally speaking, base64url is preferred.

ianwremmel: and is, in fact, necessary for encoding hydra ids.

packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -949,6 +949,9 @@ const Call = SparkPlugin.extend({
   },
 
   _updateMedia(payload) {
+    if (payload.sdp && !payload.sdp.includes(`b=`)) {
```
adamweeks: What is `b=`?

ianwremmel: bandwidth. basically linus screws up if we don't include a bandwidth line in the sdp. this check is here to make sure we're always using the bandwidth limited sdp rather than the sdp read from peer.localDescription.sdp.

ianwremmel: though, now that you mention it, that's not quite right. only video should have a b line. I can probably take this check out. it's more of a development assertion than something we actually want at runtime.

adamweeks:
Looks good to me. One day I'll understand all of the webrtc stuff, but not today

packages/node_modules/@ciscospark/plugin-phone/src/call.js
ianwremmel: no, wait, wrong again. both audio and video should have b lines. misread some code.

PR #567, opened 2017-04-27
Sdkv7 avatar

### LINKED ISSUE #564: Avatars are not retrieved and stored correctly when attempting to batch large numbers of requests
The web client's avatar handling can cause a race condition in spark-core.batch that results in incorrect values to be returned from plugin-avatar.retrieveAvatarUrl. The client makes many often duplicate calls to plugin-avatar.retrieveAvatarUrl when processing large lists of people (room activities, read receipts etc). There may be dozen of requests for an equal number of unique uuids, and five to ten duplicate requests for particular uuids (those appearing in multiple lists for example). 

The combination of  many requests with multiple duplicates cause exposes a problem within the spark-core.batch.request method that ultimately results in a boolean true being returned from the avatar URL request instead of the associated string. Batched duplicate requests are resolved once the cached deferred promise is resolved. When the timing is right, a duplicate's resolution will occur *after* the cached deferred promise's cleanup is complete. As currently written, the cleanup simply returns "true" instead of the presumed http response.

Additionally there are a few problems in avatar that need to be addressed:
 - not honoring the cacheControl information returned from the avatar service
 - incorrect parameter types being passed to avatar methods
 - An easily overlooked avatar sizes configuration that could result in incomplete cleanup after a uploading a new avatar.

good-joe:
fix(@ciscospark/plugin-avatar): Avatars are not retrieved and stored correctly when attempting to batch large numbers of requests

Fixes #564 

Web client requests many avatars on load and when entering spaces with large numbers of people. These requests are often duplicated many times as well. Usually a duplicate request through spark-core.batcher.request returns a deferred promise that resolves with the original, non-duplicate request's promise. However, these conditions introduce a race condition that can cause a duplicate to be resolved on an original's promise *after* cleanup code defined in batcher.request has been executed. The cleanup causes the promise to resolve to "true" instead of the request response and introduces issue #564. This problem is addressed in commit 074aa8b85f9baaa0d524a5b23cb4b02fb03236ef.

Two other problems were found in plugin-avatar: the cacheControl property of the avatar services response is ignored and avatar-url-store methods were not being called with the correct signature. These bugs caused the avatar cache to keep urls up to 360000 seconds. Additionally, the cache was not flushed of all urls mapped to uuids after successfully uploading a new avatar.

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
+++ b/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
@@ -53,6 +53,9 @@ const AvatarUrlBatcher = Batcher.extend({
   },
 
   handleItemSuccess(item) {
+    const maxAge = item.response.cacheControl && item.response.cacheControl.indexOf(`max-age=`) !== -1 && Number(item.response.cacheControl.split(`max-age=`)[1]) || undefined;
+    item = {uuid: item.uuid, size: item.size, url: item.response.url, maxAge};
+
     return this.getDeferredForResponse(item)
       .then((defer) => defer.resolve(item));
   },
--- a/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
+++ b/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
@@ -15,7 +15,10 @@ const AvatarUrlBatcher = Batcher.extend({
     return Promise.all(res.options.body.map((req) => {
       return Promise.all(req.sizes.map((size) => {
         const response = res.body[req.uuid] && res.body[req.uuid][size] || undefined;
-        return this.acceptItem({uuid: req.uuid, size, response});
+        return this.acceptItem({
+          response,
+          uuid: req.uuid,
+          size});
       }));
     }));
   },
@@ -53,11 +56,11 @@ const AvatarUrlBatcher = Batcher.extend({
   },
 
   handleItemSuccess(item) {
-    const maxAge = item.response.cacheControl && item.response.cacheControl.indexOf(`max-age=`) !== -1 && Number(item.response.cacheControl.split(`max-age=`)[1]) || undefined;
-    item = {uuid: item.uuid, size: item.size, url: item.response.url, maxAge};
-
     return this.getDeferredForResponse(item)
-      .then((defer) => defer.resolve(item));
+      .then((defer) => defer.resolve({
+        uuid: item.uuid,
+        size: item.size,
+        url: item.response.url}));
   },
 
   fingerprintRequest(item) {
--- a/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
+++ b/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
@@ -56,11 +56,11 @@ export default class AvatarUrlStore {
   /**
    * Adds the given item to the store
    * @param {Object} item
-   * @param {integer} cacheControl
+   * @param {integer} item.maxAge
    * @param {integer} item.size
    * @param {string} item.url
    * @param {string} item.uuid
-   * @returns {Promise<string>}
+   * @returns {Promise<string>} The URL of the added avatar
    */
   add(item) {
     if (!item) {
@@ -78,14 +78,13 @@ export default class AvatarUrlStore {
     if (!item.url) {
       return Promise.reject(new Error(`\`item.url\` is required`));
     }
-    if (!item.cacheControl) {
-      return Promise.reject(new Error(`\`item.cacheControl\` is required`));
+    if (!item.maxAge) {
+      return Promise.reject(new Error(`\`item.maxAge\` is required`));
     }
 
-    // use item suggested cache TTL or configed default if missing
-    setTimeout(this.remove.bind(this, item), item.cacheControl * 1000);
+    setTimeout(this.remove.bind(this, item), item.maxAge * 1000);
     urlByUuid.get(this).set(`${item.uuid} - ${item.size}`, item);
-    return Promise.resolve(item);
+    return Promise.resolve(item.url);
   }
 
   /**
@@ -98,8 +97,7 @@ export default class AvatarUrlStore {
    * @returns {Promise<true>}
    */
   remove(item) {
-    /* eslint no-extra-parens: [0] */
-    const sizes = (item.size && [item.size]) || [40, 50, 80, 110, 135, 192, 640, 1600];
+    const sizes = item.size && [item.size] || [40, 50, 80, 110, 135, 192, 640, 1600];
     sizes.forEach((one) => urlByUuid.get(this).delete(`${item.uuid} - ${one}`));
     return Promise.resolve(true);
   }
--- a/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
+++ b/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
@@ -28,7 +28,7 @@ export default class AvatarUrlStore {
    * @param {object} item
    * @param {string} item.uuid A user uuid
    * @param {integer}item.size the requested size
-   * @returns {Promise<object> Resolves to the avatar item {uuid, url, size, cacheControl}
+   * @returns {Promise<object>} Resolves to the avatar item {uuid, url, size, cacheControl}
    *                           or Rejects on bad param, not in store
    *
    * @memberOf AvatarUrlStore
--- a/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
+++ b/packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
@@ -28,7 +28,8 @@ export default class AvatarUrlStore {
    * @param {object} item
    * @param {string} item.uuid A user uuid
    * @param {integer}item.size the requested size
-   * @returns {Promise} resolves to the URL or rejects if not mapped
+   * @returns {Promise<object> Resolves to the avatar item {uuid, url, size, cacheControl}
+   *                           or Rejects on bad param, not in store
    *
    * @memberOf AvatarUrlStore
    */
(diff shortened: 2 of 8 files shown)


packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
```diff
@@ -61,7 +61,10 @@ const Batcher = SparkPlugin.extend({
         this.prepareItem(item)
           .then((req) => {
             defer.promise = defer.promise
-              .then(() => this.deferreds.delete(idx))
+              .then((res) => {
+                this.deferreds.delete(idx);
+                return res;
+              })
```
good-joe: Returning the result of the predecessor chained promise instead of the value of delete.

packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
```diff
@@ -53,6 +53,9 @@ const AvatarUrlBatcher = Batcher.extend({
   },
 
   handleItemSuccess(item) {
+    const maxAge = item.response.cacheControl && item.response.cacheControl.indexOf(`max-age=`) !== -1 && Number(item.response.cacheControl.split(`max-age=`)[1]) || undefined;
+    item = {uuid: item.uuid, size: item.size, url: item.response.url, maxAge};
+
     return this.getDeferredForResponse(item)
```
good-joe: cacheControl value may have changed its format. Regardless, maxAge is now correctly parsed from the given cache control. Note plugin-avatar will ultimately throw an exception if cacheControl does not exist or fails the parse. This is to notify web devs as soon as possible if the protocol changes

packages/node_modules/@ciscospark/plugin-avatar/src/avatar.js
```diff
@@ -96,7 +94,7 @@ const Avatar = SparkPlugin.extend({
     })
       .then((res) => {
         // invalidate user's cached avatar
-        this.store.remove(this.spark.device.userId);
+        this.store.remove({uuid: this.spark.device.userId});
         return res.url;
```
good-joe: remove was expecting an "item" object, not just a uuid. This prevented cleanup of all URLs associated with uuid

packages/node_modules/@ciscospark/plugin-avatar/src/avatar.js
```diff
@@ -32,12 +32,12 @@ const Avatar = SparkPlugin.extend({
...
+   * @returns {Promise<string>} the avatar URL
    */
   _fetchAvatarUrl(uuid, options) {
-    return this.store.get(uuid, options.size)
-      .catch(() => this.batcher.request(Object.assign({}, {uuid, size: options.size}))
-        .then((item) => this.store.add(defaults(item, {cacheControl: options.cacheControl, url: item.response.url}))));
+    return this.store.get({uuid, size: options.size})
+      .catch(() => this.batcher.request({uuid, size: options.size, maxAge: options.maxAge})
+        .then((item) => this.store.add(item)));
   },
```
good-joe: damnit. maxAge is ignored here. I'll remove shortly

packages/node_modules/@ciscospark/plugin-avatar/src/config.js
```diff
@@ -11,11 +11,6 @@ export default {
     batcherMaxWait: 1500,
 
     /**
-     * @description Milliseconds a cached avatar url is considered valid
-     * @type {number}
-     */
-    cacheExpiration: 60 * 60 * 1000,
-    /**
      * @description default avatar size to retrieve if no size is specified
```
good-joe: cacheControl is now a required property of the request response

packages/node_modules/@ciscospark/plugin-avatar/test/unit/spec/avatar-url-store.js
```diff
@@ -8,15 +8,15 @@ import MockSpark from '@ciscospark/test-helper-mock-spark';
...
+  const item1_40 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 40, url: `http://www-40.example.com`, maxAge: 300};
+  const item1_50 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 50, url: `http://www-50.example.com`, maxAge: 300};
+  const item1_80 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 80, url: `http://www-80.example.com`, maxAge: 300};
+  const item1_110 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 110, url: `http://www-110.example.com`, maxAge: 300};
+  const item1_192 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 192, url: `http://www-192.example.com`, maxAge: 300};
+  const item1_640 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 640, url: `http://www-640.example.com`, maxAge: 300};
+  const item1_1600 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa0`, size: 1600, url: `http://www-1600.example.com`, maxAge: 300};
+  const item2_80 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa1`, size: 80, url: `http://www2.example.com`, maxAge: 300};
+  const item3_80 = {uuid: `88888888-4444-4444-4444-aaaaaaaaaaa2`, size: 80, url: `http://www3.example.com`, maxAge: 300};
```
good-joe: the store only sees maxAge (parsed in the avatar batcher), not the whole cacheControl since cache control also defines (the currently ignored) visibility.

packages/node_modules/@ciscospark/plugin-avatar/test/unit/spec/avatar.js
```diff
@@ -43,15 +43,13 @@ describe(`plugin-avatar`, () => {
...
+                cacheControl: `public max-age=3600`
               }
             }
           },
           statusCode: 200,
           options: {
-            ids: [
-              `88888888-4444-4444-4444-aaaaaaaaaaa0`
-            ],
             body: [
```
good-joe: Ids are not part of the avatar protocol

packages/node_modules/@ciscospark/plugin-avatar/test/unit/spec/avatar.js
```diff
@@ -446,6 +424,112 @@ describe(`plugin-avatar`, () => {
...
+          assert.becomes(avatar.retrieveAvatarUrl(`88888888-4444-4444-4444-aaaaaaaaaaa1`), `https://example.com/88888888-4444-4444-4444-aaaaaaaaaaa1`),
+          assert.becomes(avatar.retrieveAvatarUrl(`88888888-4444-4444-4444-aaaaaaaaaaa4`), `https://example.com/88888888-4444-4444-4444-aaaaaaaaaaa4`),
+          assert.becomes(avatar.retrieveAvatarUrl(`88888888-4444-4444-4444-aaaaaaaaaaa4`), `https://example.com/88888888-4444-4444-4444-aaaaaaaaaaa4`),
+          assert.becomes(avatar.retrieveAvatarUrl(`88888888-4444-4444-4444-aaaaaaaaaaa0`), `https://example.com/88888888-4444-4444-4444-aaaaaaaaaaa0`),
+          assert.becomes(avatar.retrieveAvatarUrl(`88888888-4444-4444-4444-aaaaaaaaaaa5`), `https://example.com/88888888-4444-4444-4444-aaaaaaaaaaa5`)
+        ])
+          .then(() => {
+            assert.callCount(spark.request, 1);
+          });
+      });
```
good-joe: This test is simply a workout of duplicate handling. See general comment on proof of fix wrt batcher returning incorrect values

packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-batcher.js
ianwremmel: I seem to recall the item itself has a cache property; I'm not sure the cache header is the appropriate place to pull the cache data from (since this is a bulk request, it actually seems to me that it's a bug that the response has a cache header at all).

packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
```diff
@@ -78,14 +78,13 @@ export default class AvatarUrlStore {
     if (!item.url) {
       return Promise.reject(new Error(`\`item.url\` is required`));
     }
-    if (!item.cacheControl) {
```
ianwremmel: It looks like we were trying to use item.cacheControl, but didn't bother to parse it. I think in handleItemSuccess you want to look at item.cacheControl, not item.response.cacheControl.

packages/node_modules/@ciscospark/plugin-avatar/src/avatar-url-store.js
```diff
@@ -78,14 +78,13 @@ export default class AvatarUrlStore {
...
+    if (!item.maxAge) {
+      return Promise.reject(new Error(`\`item.maxAge\` is required`));
     }
 
-    // use item suggested cache TTL or configed default if missing
-    setTimeout(this.remove.bind(this, item), item.cacheControl * 1000);
+    setTimeout(this.remove.bind(this, item), item.maxAge * 1000);
     urlByUuid.get(this).set(`${item.uuid} - ${item.size}`, item);
-    return Promise.resolve(item);
+    return Promise.resolve(item.url);
```
ianwremmel: It's kinda odd to put an object into a store, then only get one of its properties back as a result of the add. Can you move the item.url selection to the invocation site?

packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
```diff
@@ -61,7 +61,10 @@ const Batcher = SparkPlugin.extend({
         this.prepareItem(item)
           .then((req) => {
             defer.promise = defer.promise
-              .then(() => this.deferreds.delete(idx))
+              .then((res) => {
```
ianwremmel: You can use `tap()` from `@ciscospark/common` for this sort of thing.

PR #574, opened 2017-05-01
add in-browser constant

ianwremmel:


packages/node_modules/@ciscospark/common/src/in-browser/index.js
```diff
@@ -0,0 +1,13 @@
...
+ * @ignore
+ */
+
+/**
+ * Constants that, through the magic of bundling should indicate if we're
+ * running in a nodejs or a browser environment
+ * @name inBrowser
+ * @type {boolean}
+ */
+export {default as default} from './node';
```
bzang: Does `export {default} from './node';` work here?

ianwremmel: technicall, yes, but it confuses eslint.

ianwremmel:
The edgecases I can think of:
1. electron: it's like a browser but can run native node code; i imagine this isn't the most problematic part of the sdk for electron
2. webpack 1: it *might* ignore the browser field unless you put `browser` in your web pack config. again, this isn't the worst problem you'd see if that's the case.

ianwremmel:
merged via other code

PR #585, opened 2017-05-15
V8 consolidated: Consolidated sdk changes for web-client to move to v8

saurjai3:


saurjai3:
Will open another PR after merging the code.

PR #586, opened 2017-05-15
Remove deprecated stream urls

ianwremmel:


ianwremmel:
Yes. and if all goes according to plan, that'll happen automatically based on the commit message in 

```
feat(@ciscospark/plugin-phone): remove media stream urls 

BREAKING CHANGE: this change will cause apps referencing `call.remoteMediaStreamUrl` or `call.localMediaStreamUrl` to stop rendering the referenced streams. Apps should rely on `call.removeMediaStream` and `call.localMediaStream`
```

PR #589, opened 2017-05-16
V8 consolidated: Consolidated sdk changes for web-client to move to v8

saurjai3:


packages/node_modules/@ciscospark/storage-adapter-local-storage/src/index.js
```diff
@@ -72,7 +72,7 @@ export default class StorageAdapterLocalStorage {
        */
       clear() {
         loggers.get(this).info(`local-storage-store-adapter: clearing localStorage`);
-        return Promise.resolve(localStorage.clear());
+        return Promise.resolve(localStorage.removeItem(`web-client-internal`));
```
ianwremmel: This can't be hard coded to `web-client-internal` as there's no telling what value might be passed by a consumer. The key to remove is the key that the rest of the adapter uses.

packages/node_modules/@ciscospark/plugin-conversation/src/decryption-transforms.js
```diff
@@ -124,8 +132,9 @@ export const transforms = toArray(`inbound`, {
         object[name] = plaintext;
       })
       .catch((reason) => {
-        ctx.spark.logger.warn(`plugin-conversation: failed to decrypt ${name}`);
-        return Promise.reject(reason);
+        ctx.spark.logger.warn(`plugin-conversation: failed to decrypt ${name} `, reason);
+        object[name] = decryptionFailureMessage;
+        Promise.resolve(decryptionFailureMessage);
```
ianwremmel: This needs to return the Promise.

PR #590, opened 2017-05-17
Fix auth bugs

ianwremmel:


ianwremmel:
mostly merged, except for the `make config less magical` commit. I don't actually remember what that was solving and it caused issues, so I'm going to close for now.

PR #592, opened 2017-05-18
Reimplement filterEscape and restore pre 0.7 convo behavior

ianwremmel:
I'm not really convinced this is the right behavior for convo, but it's what the web client does now. After updating the convo tests and a conversation with @pcl, I'm reasonably convinced the right behavior is to reject messaged with disallowed html. That, however, is a PM/UX decision that'll need to be made product wide. Until then, this seems to be the "most correct" of the available options.

### DIFF
--- a/packages/node_modules/@ciscospark/helper-html/src/html.js
+++ b/packages/node_modules/@ciscospark/helper-html/src/html.js
@@ -35,3 +35,5 @@ function noopSync(processCallback, allowedTags, allowedStyles, html) {
 
 export const filter = curry(noop, 4);
 export const filterSync = curry(noopSync, 4);
+export const filterEscape = curry(noop, 4);
+export const filterEscapeSync = curry(noopSync, 4);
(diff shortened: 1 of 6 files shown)


bzang:
Yea, it looks fine, but I agree, not the optimal way to do it... although I don't think that rejecting the message is the right way to do it either. I would think it would be to escape anything that isn't allowed. How do other messaging platforms do it?

ianwremmel:
this is escaping the stuff that isn't allowed, but it still seems od to me that if i type `<b><script>console.log(1)</script></b>` that part of it would be escaped. Though, that does appear to be github's behavior:

<b><script>console.log(1)</script></b>

ianwremmel:
i guess a rejection seems appropriate since we've (currently) taken the stance that users shouldn't enter html directly. therefore, if disallowed html makes it as far as the sdk, it's a bug in the consumer. We should revisit this soon and come up with a better method that both gives control but continues to protect the system.

PR #594, opened 2017-05-18
Fix locus parsing

ianwremmel:


packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -561,19 +539,20 @@ const Call = SparkPlugin.extend({
...
-        this.trigger(`connected`);
-        break;
-      case `disconnected`:
-        this.trigger(`disconnected`);
-        break;
-      default:
-        // do nothing
-      }
+      process.nextTick(() => this.trigger(this.status));
+      // switch (this.status) {
```
adamweeks: Did you mean to leave this commented code in?

ianwremmel:
current implementation fails with timeouts

PR #596, opened 2017-05-20
WIP: Firefox Testing

ianwremmel:
Fixes #544
Fixes #593

This changeset automatically downloads the openh264 plugin and adds it to the browser defs for @ciscospark/plugin-phone. When run locally, this allows us to test firefox with openh264. On sauce labs, however, the inclusion of a Firefox profile seems to cause network requests to hang; the first network request (creating a test user) logs its outbound payload, but never completes and never rrors.


ianwremmel:
fixed via #630

PR #601, opened 2017-05-23
Add non-wedging timers

ianwremmel:


package.json
```diff
@@ -54,7 +54,7 @@
     "prebuild": "./tooling/index.js exec -- rimraf dist",
     "build": "./tooling/index.js build",
     "prebuild:docs": "rimraf ./docs/api",
-    "build:docs": "documentation build --np --project-name ciscospark --project-version 1.x --format html --output ./docs/api --github ./packages/node_modules/ciscospark/src/index.js ./packages/node_modules/@ciscospark/plugin-phone/src/index.js",
+    "build:docs": "documentation build --access public --np --project-name ciscospark --project-version 1.x --format html --output ./docs/api --github ./packages/node_modules/ciscospark/src/index.js ./packages/node_modules/@ciscospark/plugin-phone/src/index.js",
```
adamweeks: What does this change do?

packages/node_modules/@ciscospark/common-timers/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+/**
+ * Wrapper around setTimout which (in node) unrefs the returned timer to avoid
+ * wedging the process open unexpectedly.
+ * @param {Mixed} args
+ * @protected
+ * @returns {Timer|Number}
+ */
+export function safeSetTimeout(...args) {
```
adamweeks: Good call, had no idea this was an issue.

package.json
ianwremmel: it prevents anything tagged with `@protected` from being documented. my goal is for `@public` stuff to go into the docs site and `@public + @protected` to go into readmes. That way, the docs site will explain how to use the officially sdk, but the readmes will explain how to use each plugin and fully explain spark-core, etc.

PR #603, opened 2017-05-26
fix(internal-plugin-user): remove unnecessary checks for hasPassword

aimeex3:
res.body.hasPassword does not exist in the activate() and setPassword() call


aimeex3:
merged in https://github.com/ciscospark/spark-js-sdk/pull/606

PR #606, opened 2017-05-31
V8.1 Fix for hashtags and onBoarding

saurjai3:


packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -210,7 +210,12 @@ const Authorization = SparkPlugin.extend({
     location = cloneDeep(location);
     if (this.spark.getWindow().history && this.spark.getWindow().history.replaceState) {
       Reflect.deleteProperty(location.query, `code`);
-      location.query.state = base64.encode(JSON.stringify(omit(location.query.state, `csrf_token`)));
+      if (isEmpty(omit(location.query.state, `csrf_token`))) {
```
ianwremmel: Please add a test

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -68,9 +68,7 @@ const User = SparkPlugin.extend({
     })
       .then((res) => {
         this.spark.credentials.set({supertoken: res.body.tokenData});
-        if (res.body.hasPassword) {
-          this.hasPassword = true;
-        }
+        this.hasPassword = true;
```
ianwremmel: Is this accurate? I thought the point of calling /activate was to determine if password was set.

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/src/authorization.js
saurjai3: Added the test

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
saurjai3: I checked with Aimee, This is required for the onBoarding to work.

aimeex3: oops sorry - this line should actually be removed. 

`activate()` fetches the auth token so that the user will be able to set their password. when this function is called hasPassword should be false actually.
`setPassword()` should set `hasPassword` to true
`verify()` actually hits the /activate endpoint on atlas (yes confusing name) and this determines if the password was set

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -225,5 +225,48 @@ browserOnly(describe)(`plugin-authorization-browser-first-party`, () => {
           });
       });
     });
+
+    describe(`#_cleanUrl()`, () => {
+      it(`removes the state parameter for an empty key`, () => {
```
ianwremmel: "removes the state parameter when it has no keys"

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -225,5 +225,48 @@ browserOnly(describe)(`plugin-authorization-browser-first-party`, () => {
...
+            }
+          }
+        };
+        sinon.spy(spark.authorization, `_cleanUrl`);
+        spark.authorization._cleanUrl(location);
+        assert.called(spark.getWindow().history.replaceState);
+        assert.equal(spark.getWindow().location.href, ``);
+      });
+
+      it(`updates the state parameter when key is present`, () => {
```
ianwremmel: "keeps the parameter when it has keys"

ianwremmel:
Note that scopes in commit messages should be the full plugin name, including `@ciscospark`

saurjai3:
@ianwremmel , Ready for 2nd review. thanks.

aimeex3:
ah sorry forgot one more param. in `verify()` check for `res.body.sso` too
so if (res.body.hasPassword || res.body.sso), set this.hasPassword to true

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -225,5 +225,48 @@ browserOnly(describe)(`plugin-authorization-browser-first-party`, () => {
...
+            state: {
+              csrf_token: `token`,
+              key: `value`
+            }
+          }
+        };
+        sinon.spy(spark.authorization, `_cleanUrl`);
+        spark.authorization._cleanUrl(location);
+        assert.called(spark.getWindow().history.replaceState);
+        assert.isDefined(spark.getWindow().location.href);
```
ianwremmel: This confirms that `_cleanUrl()`. was called, but I don't see where it confirms the intended behavior

ianwremmel: For example, confirm that the new state has `key` but not `csrf_token`

PR #609, opened 2017-06-06
feat(internal-plugin-presence): initial implementation

adamweeks:


packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,145 @@
...
+      body: {
+        subject: this.spark.internal.device.userId,
+        eventType: status,
+        ttl
+      }
+    })
+      .then((response) => transformSingleStatusResponse(response.body));
+  },
+
+  postCompositions(subjects) {
```
ianwremmel: I'd rather not wrap api calls if we're only removing a single copy of code; unless code is getting duplicated *a lot* i feel like there's value in not having to go searching for the network request when reading these functions.

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -0,0 +1,100 @@
...
+        return spock.spark.internal.presence.get(mccoy.id)
+          .then((presenceResponse) => {
+            assert.property(presenceResponse, mccoy.id);
+          });
+      });
+    });
+
+    describe(`#list()`, () => {
+      it(`returns a list of people matching email address`, () => {
+        return spock.spark.internal.presence.list([mccoy.id, spock.id])
```
ianwremmel: This doesn't look like it does what the description says.

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return this.postSubscriptions([personId], 0);
+  },
+
+  /**
+   * Unsubscribes from a group of people's presence subscription
+   * @param {Array} personIds
+   * @returns {Promise}
+   */
+  unsubscribeList(personIds) {
```
ianwremmel: Oh, I see. the list methods return arrays. I think I'd rather see this implemented via the batcher, but let's talk it through tomorrow (or sometime).

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -0,0 +1,100 @@
...
+          .then((presenceResponse) => {
+            assert.isAbove(Object.keys(presenceResponse).length, 0);
+            assert.property(presenceResponse, mccoy.id);
+            assert.property(presenceResponse, spock.id);
+          });
+      });
+    });
+
+    describe(`#subscribe`, () => {
+      it(`subscribes to and returns current status of a person id`, () => spock.spark.internal.presence.subscribe(mccoy.id)
```
ianwremmel: We should probably get a mercury test in place, if at all possible.

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Gets the current presence status of a given person id
+   * @param {string} personId
+   * @returns {Promise}
+   */
+  get(personId) {
+    if (!personId) {
+      return Promise.reject(new Error(`A person id is required`));
+    }
+    return this.postCompositions([personId]);
```
adamweeks: You can do a GET for a single user, but the response is completely different, so I just used the bulk endpoint for one or many

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,145 @@
...
+  },
+
+  /**
+   * Subscribes to a list of people's presence status updates
+   * Updates are sent via mercury events `apheleia.subscription_update`
+   * @param {Array} personIds
+   * @param {number} subscriptionTtl - Requested length of subscriptions in seconds.
+   * @returns {Promise}
+   */
+  subscribeList(personIds, subscriptionTtl = defaultSubscriptionTtl) {
```
adamweeks: You can subscribe to all members in a room/space that way you will get status updates for everyone over mercury.

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
adamweeks: Oops. Copy pasta fail.

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
adamweeks: Yea, good call.

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
adamweeks: That is what I'm doing on the non-plugin version for the widget. I can do that here too if you'd rather. https://github.com/ciscospark/react-ciscospark/blob/master/packages/node_modules/%40ciscospark/redux-module-presence/src/actions.js#L86

adamweeks:
@ianwremmel ready for round 2

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,127 @@
...
+    else {
+      subjects = [personIds];
+    }
+    return this.spark.request({
+      method: `POST`,
+      service: `apheleia`,
+      resource: `compositions`,
+      body: {subjects}
+    })
+      .then((response) => transformStatusResponses(response.body.statusList));
```
ianwremmel: Please apply the transforms as payload transformations instead of invoking them directly.

adamweeks: I don't know what you mean by this. Do you have an example you can point me to?

ianwremmel: take a look at the team and conversation plugins.

packages/node_modules/@ciscospark/internal-plugin-presence/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+import config from './config';
+
+registerInternalPlugin(`presence`, Presence, {
+  payloadTransformer: {
+    predicates: [
+      {
+        name: `normalizeSingleStatusResponse`,
+        direction: `inbound`,
+        test(ctx, response) {
+          return Promise.resolve(has(response, `body.eventType`));
```
ianwremmel: Is this adequately specific? i feel like other things may have an `eventType`, though I can't actually name anything at the moment

packages/node_modules/@ciscospark/internal-plugin-presence/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+        },
+        extract(response) {
+          return Promise.resolve(response);
+        }
+      },
+      {
+        name: `normalizeStatusResponses`,
+        direction: `inbound`,
+        test(ctx, response) {
+          return Promise.resolve(has(response, `body.statusList`) || has(response, `body.responses`));
```
ianwremmel: Again, are these test specific enough? Should something in the first item of each list be checked to confirm it's the type you expect it to be?

packages/node_modules/@ciscospark/internal-plugin-presence/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+            statuses = response.body.statusList;
+          }
+          else if (has(response, `body.responses`)) {
+            statuses = response.body.responses;
+          }
+          else {
+            return Promise.resolve();
+          }
+          // Convert array of statuses to an object
+          const processed = {};
```
ianwremmel: Consider `reduce` instead of `forEach`

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,127 @@
...
+    else {
+      subjects = [personIds];
+    }
+    return this.spark.request({
+      method: `POST`,
+      service: `apheleia`,
+      resource: `compositions`,
+      body: {subjects}
+    })
+    .then((response) => response.body);
```
ianwremmel: based on the transform above, won't this require the consumer to then lookup the result for a single user in the response? maybe we need a `get()` and a `list()`...

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
```diff
@@ -0,0 +1,127 @@
...
+        subjects,
+        subscriptionTtl,
+        includeStatus: true
+      }
+    })
+    .then((response) => response.body);
+  },
+
+  /**
+   * Unsubscribes from a single person's presence subscription
```
ianwremmel: This description isn't quite accurate.

packages/node_modules/@ciscospark/internal-plugin-presence/src/index.js
adamweeks: Is all traffic processed through the reducers or just the presence plugin traffic?

ianwremmel: all traffic. the idea was to let us add new apis that would *just work* without needing to make sdk changes.

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
adamweeks: The idea was to have all presence responses be in the format of:
```js
{
  userId: {
    presence: {}
  }
}
```
We could do just the presence part for single users

packages/node_modules/@ciscospark/internal-plugin-presence/src/index.js
adamweeks: Ok, i'll try to add more tests to make sure it is the right object we are looking for.

packages/node_modules/@ciscospark/internal-plugin-presence/src/presence.js
adamweeks: It was at one point 😂

adamweeks:
@ianwremmel ready for another round

PR #615, opened 2017-06-09
WebRTC Cleanup

ianwremmel:
# Implementation
- [x] general cleanup
- [x] investigate if removal of remote streams fires negotiationneeded in ff
- [x] add dynamic assertions for firefox
- [x] figure out why chrome has one test that fails if all tests run
- [x] test re-re-negotiation
- [x] lint
- [x] squash
- [x] remove lockfile
- [x] support providing constraint to setMedia
- [x] support providing track to setMedia
- [x] move media-engine-webrtc unit tests to start of build due to high duration
- [x] skip media-engine-webrtc in gating builds since it's just unit tests
- [ ] change name back to adapter

# Utilization

The phone plugin needs the following additions in order to avoid changing its public api

## These need to be evented
- [ ] media.remoteMediaStream
- [ ] media.localMediaStream
- [x] media.receivingAudio
- [x] media.receivingVideo
- [x] media.sendingAudio
- [x] media.sendingVideo

## This needed to be readable
- [x] media.videoConstraint.facingMode.ideal

# Phone Plugin Changes
- [x] use `media.setMedia()` instead of `media.set()`


packages/node_modules/@ciscospark/common/src/evented.js
```diff
@@ -0,0 +1,35 @@
+import makeTemplateContainer from './template-container';
+
+const data = new (makeTemplateContainer(WeakMap, Map))();
+
+/**
+ * Given a class property, this decorator change it into a setter/getter pair;
```
bzang: `...changes it into...`

packages/node_modules/@ciscospark/media-engine-webrtc/browsers.js
```diff
@@ -0,0 +1,53 @@
...
+        platform: `OS X 10.12`,
+        browserName: `chrome`,
+        version: `latest`,
+        chromeOptions: {
+          args: [
+            `--use-fake-device-for-media-stream`,
+            `--use-fake-ui-for-media-stream`
+          ]
+        }
+      },
```
bzang: Only macos?

packages/node_modules/@ciscospark/media-engine-webrtc/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@ciscospark/media-engine-webrtc",
+  "version": "1.0.3",
+  "description": "",
+  "license": "MIT",
+  "main": "dist/index.js",
+  "devMain": "src/index.js",
+  "repository": "https://github.com/ciscospark/spark-js-sdk/tree/master/packages/media-engine-webrtc",
+  "engines": {
+    "node": ">=4"
```
bzang: Should this be `>= 6` like the other packages?

packages/node_modules/@ciscospark/media-engine-webrtc/browsers.js
ianwremmel: we haven't setup the windows openh264 stuff yet; once we have that, we'll add it here.

packages/node_modules/@ciscospark/media-engine-webrtc/package.json
ianwremmel: yea, but i think you're mixing PRs; I think prod has them all at 4 and one of the other PRs you reviewed moves them all to six.

ianwremmel:
because of the update to adapter.js, we need to delay merging this until the phone plugin consumes it.

PR #616, opened 2017-06-09
refactor(internal-plugin-metrics): enable pre-login metrics

saurjai3:


packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -80,15 +80,18 @@ const Metrics = SparkPlugin.extend({
...
-    return this.request({
-      method: `POST`,
-      api: `metrics`,
-      resource: `clientmetrics-prelogin`,
-      headers: {
-        "x-prelogin-userid": preLoginId
-      },
-      body: payload
-    });
+    return this.spark.credentials.getUserToken()
```
ianwremmel: There's probably no reason to try getting a user token. Given this is a *pre login metric*, I can't imagine you would ever not hit the `catch` case.

saurjai3:
@ianwremmel , Ready for another review... thanks.

PR #623, opened 2017-06-16
Fixes plugin support and oneFlight

saurjai3:
Fixes plugin support and oneFlight

### DIFF
--- a/packages/node_modules/@ciscospark/common/src/one-flight.js
+++ b/packages/node_modules/@ciscospark/common/src/one-flight.js
@@ -52,8 +52,7 @@ export default function oneFlight(...params) {
       }
 
       /* eslint no-invalid-this: [0] */
-      let flight;
-      flight = flights.get(this, target, key);
+      let flight = flights.get(this, target, key);
       if (flight) {
         const message = `one flight: attempted to invoke ${prop} while previous invocation still in flight`;
         /* instanbul ignore else */
--- a/packages/node_modules/@ciscospark/common/src/one-flight.js
+++ b/packages/node_modules/@ciscospark/common/src/one-flight.js
@@ -44,15 +44,15 @@ export default function oneFlight(...params) {
    * @returns {Object}
    */
   function oneFlightDecorator(target, prop, descriptor) {
-    let key = prop;
+    const key = prop;
     descriptor.value = wrap(descriptor.value, function oneFlightExecutor(fn, ...args) {
+      let innerKey = key;
       if (keyFactory) {
-        const keyFactoryValue = keyFactory(...args);
-        key = keyFactoryValue || `${key}_${keyFactoryValue}`;
+        innerKey = `${innerKey}_${keyFactory(...args)}`;
       }
 
       /* eslint no-invalid-this: [0] */
-      let flight = flights.get(this, target, key);
+      let flight = flights.get(this, target, innerKey);
       if (flight) {
         const message = `one flight: attempted to invoke ${prop} while previous invocation still in flight`;
         /* instanbul ignore else */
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/common/test/unit/spec/one-flight.js
```diff
@@ -5,6 +5,43 @@ import {assert} from '@ciscospark/test-helper-chai';
 
 describe(`common`, () => {
   describe(`@oneFlight`, () => {
+    it(`returns new promise on function with undefined keyFactory`, () => {
```
ianwremmel: What happens if you call a oneflighted function twice without a keyfactory? isn't that the scenario that was causing you a problem?

saurjai3: The scenario which was causing the issue is that -> 
OneFlight was called on funcC1() with keyFactory as undefined. The promise is not resolved yet.
Now when you call funcC2() with keyFactory as undefined then it would assume that there is already a oneFlight i.e from above function call and would return that promise instead of actually calling funcC2(). I think that is a wrong behavior since it should allow calling funcC2(). Remember that both funcC1() and fncC2() are on the same object.

The fix:
1) My suggestion is that the keyFactory should never be called if undefined, as the weakmaps do not support undefined keys. If it does then it tends to lock the entire object and not that specific function.
2) If the keyFactory is undefined allow calling the function.

packages/node_modules/@ciscospark/common/src/one-flight.js
```diff
@@ -51,7 +51,10 @@ export default function oneFlight(...params) {
       }
 
       /* eslint no-invalid-this: [0] */
-      let flight = flights.get(this, target, key);
+      let flight;
+      if (key) {
```
ianwremmel: `key` should only be undefined if `keyFactory` returns `undefined`, but that doesn't mean call is not in flight. Assuming `undefined` is not a valid key (and we need to confirm that), then, really, the solution needs to fix that key rather than assuming flight exists.

note that `WeakMappedMappedMap` is a WeakMap of Maps of Maps, meaning that `undefined` needs to be a valid key for `Map`s, not necessarily `WeakMaps`. `WeakMappedMappedMap` might be expressed as `WeakMap<Object, Map<Object, Map<string, Promise>>>` in e.g. typescript.

`flights.get(this, target, key)` expands to something like

```js
const targets = flights.get(this);
const flightsForTarget = targets.get(target);
const flight = flightsForTarget.get(key);
```

packages/node_modules/@ciscospark/common/src/one-flight.js
```diff
@@ -47,7 +47,8 @@ export default function oneFlight(...params) {
     let key = prop;
     descriptor.value = wrap(descriptor.value, function oneFlightExecutor(fn, ...args) {
       if (keyFactory) {
-        key = keyFactory(...args);
+        const keyFactoryValue = keyFactory(...args);
+        key = keyFactoryValue || `${key}_${keyFactoryValue}`;
```
ianwremmel: there's probably no need for the boolean here. In fact, it's still plausible that two methods could produce the same key. 

This'll probably be the safest option:

```js
key = `${key}_${keyFactor(...args)}`;
```

packages/node_modules/@ciscospark/common/test/unit/spec/one-flight.js
```diff
@@ -5,6 +5,43 @@ import {assert} from '@ciscospark/test-helper-chai';
...
+
+      return Promise.all([
+        c.funcC1(),
+        c.funcC2()
+      ])
+        .then(() => {
+          assert.calledTwice(c.spy);
+        });
+    });
+
```
ianwremmel: Please add another test for the original scenario of calling the same method twice with an undefined key factory result.

PR #624, opened 2017-06-19
feat(internal-plugin-board): allow multiple Realtime connections at once

tran2:
This PR adds the ability to have 2+ realtime connections for both shared mercury or non-shared whiteboards. (to address the case when we have a whiteboard open while viewing another whiteboard)


packages/node_modules/@ciscospark/internal-plugin-board/src/realtime-channel.js
```diff
@@ -0,0 +1,48 @@
...
+      default: false
+    },
+    socket: {
+      type: `object`
+    }
+  },
+
+  send(data) {
+    /* istanbul ignore next */
+    if (!this.socket) {
```
ianwremmel: Is this necessary given that the run time will throw basically the same string as an exception?

packages/node_modules/@ciscospark/internal-plugin-board/src/realtime.js
```diff
@@ -96,13 +87,20 @@ const RealtimeService = Mercury.extend({
...
     }
 
-    // use mercury socket if it is shared
-    return this.spark.internal.feature.getFeature(`developer`, `web-shared-mercury`)
-      .then((isSharingMercuryFeatureEnabled) => {
-        if (isSharingMercuryFeatureEnabled && this.isSharingMercury) {
-          return this.spark.internal.mercury.socket.send(data);
-        }
-        return this.socket.send(data);
+    return realtimeChannel.send(data);
```
ianwremmel: I wonder of realtimeChannel should have a different name. Doesn't "channel" already refer to (approximately) the whiteboard in question?

packages/node_modules/@ciscospark/internal-plugin-board/src/realtime.js
```diff
@@ -148,21 +166,31 @@ const RealtimeService = Mercury.extend({
...
-        this.boardBindings = [res.binding];
-        this.boardWebSocketUrl = res.webSocketUrl;
-
-        if (!res.sharedWebSocket) {
-          return this.connect()
-            .then(() => {
-              this.isSharingMercury = false;
-              return res;
-            });
+      .then(tap(({binding, webSocketUrl}) => {
```
ianwremmel: Given that tap will suppress errors, let's try to get in the habit of commenting why we're suppressing errors.

packages/node_modules/@ciscospark/internal-plugin-board/test/integration/spec/sharing-mercury.js
```diff
@@ -6,20 +6,23 @@
...
+    let mccoy, spock;
+    let uniqueRealtimeData;
 
     before(`create users`, () => testUsers.create({count: 2})
       .then((users) => {
-        participants = users;
+        participants = [spock, mccoy] = users;
 
-        return Promise.all(map(participants, (participant) => {
+        return Promise.all(Array.map(participants, (participant) => {
```
ianwremmel: Why not just call participants.map directly?

packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
```diff
@@ -51,7 +51,7 @@ const Mercury = SparkPlugin.extend({
   },
 
   @oneFlight
-  connect() {
+  connect(webSocketUrl) {
     if (this.connected) {
```
ianwremmel: Let's get some unit tests that confirm the underlying socket is created with webSocketUrl when specified or the default when not (I assume one of those tests probably already exists, but it seems like this change should have needed at least one new test.)

packages/node_modules/@ciscospark/internal-plugin-board/src/realtime.js
tran2: not sure what other name it should have

packages/node_modules/@ciscospark/internal-plugin-board/src/realtime-channel.js
tran2: ok, I'll remove this if block

tran2:
@ianwremmel I've added a new commit to `plugin-mercury` so that on close event `_reconnect` would connect the correct socket url

PR #625, opened 2017-06-20
refactor(@ciscospark/i-plugin-support): call getUserToken on submitLogs

saurjai3:
Fixes failure to upload logs from web-client

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-support/test/unit/spec/support.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-support/test/unit/spec/support.js
@@ -34,4 +34,11 @@ describe(`plugin-support`, function() {
     });
   });
 
+  describe(`#submitLogs()`, () => {
+    it(`calls getUserToken`, () => {
+      spark.internal.support.submitLogs({});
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/internal-plugin-support/test/unit/spec/support.js
```diff
@@ -34,4 +34,11 @@ describe(`plugin-support`, function() {
     });
   });
 
+  describe(`#submitLogs()`, () => {
```
ianwremmel: Is there an integration test for this already? Why didn't it catch this?

saurjai3: yes there is but it is disabled. The comment says that rackspace is broken. Do you want me to enable that and remove the unit test?
screen shot 2017-06-21 at 4 33 15 pm

ianwremmel: ah, gotcha. i remember now. yea, let's try turning that back on.

ianwremmel:
Please turn the disabled test back on, then this is good to merge.

PR #626, opened 2017-06-21
Bundling

ianwremmel:
I *think* 75fa5e8 is the only commit that's not superseded by other PRs.


tooling/browserify/dot-browser.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+/* eslint-disable no-invalid-this */
+
+const debug = require(`debug`)(`tooling:dot-browser`);
+const fs = require(`fs`);
+const path = require(`path`);
+const {wrap} = require(`lodash`);
```
bzang: would it be better to adapt the `const wrap = require(`lodash/wrap`);` syntax here as well just to be consistent with the code that supports tree shaking?

tooling/lib/build.js
```diff
@@ -6,32 +6,104 @@ const {
...
+        dest: {
+          browser: path.resolve(`packages`, `node_modules`, packageName, filename.replace(`src`, `dist/browser`)),
+          main: path.resolve(`packages`, `node_modules`, packageName, filename.replace(`src`, `dist/main`)),
+          module: path.resolve(`packages`, `node_modules`, packageName, filename.replace(`src`, `dist/module`))
+        }
+      });
+    }
+  }
+
+  const config = JSON.parse(await readFile(path.join(process.cwd(), `.babelrc`)));
```
bzang: Would using the sync version of `readFile` (and other fs methods) make more sense than doing async/await?

tooling/browserify/dot-browser.js
ianwremmel: yea, that's probably a good idea.

tooling/lib/build.js
ianwremmel: maybe. `await` is used all over tooling, so it's not trivial to take it out; I think at this point remaining consistent with what's there seems like the way to go.

PR #627, opened 2017-06-23
fix jsdoctrinetest

ianwremmel:


packages/node_modules/ciscospark/src/plugins/webhooks.js
```diff
@@ -30,7 +30,6 @@ const Webhooks = SparkPlugin.extend({
    * @param {Types~Webhook} webhook
    * @returns {Promise<Webhook>}
    * @example
-   * var ciscospark = require('../..');
```
ianwremmel: yea, _setup.js injects it into the scope. it's a bit of hack (in order to get credentials in place), but it's way less of a hack than what it was doing before).

PR #628, opened 2017-06-24
API Docs

ianwremmel:


ianwremmel:
View running demo at https://pathologist-pennants-41583.netlify.com/

ianwremmel:
This is probably going to be simpler to review commit-by-commit, but [this](https://github.com/ciscospark/spark-js-sdk/compare/8437152b7705dae0c9dff9551ce12a28ef30a9fd...ianwremmel:16021630844c2929fe4ee85df2a7c6a1ca089378) removes most of the same-change-in-every-file changes)

bzang:
Minor UX thoughts:
- Glad it's much cleaner and structured better than the previous docs.
- Is there a way to add better syntax coloring into the code blocks?
- What about a button to open or close all of the "Instance members"?
- I can't figure out what the filter box actually does... seems to show or hide sections, but I can't tell how the text I enter matches with the filtered results. 
- Clicking to open the tree in resources also goes to that resource. It may make more sense to have the expand behavior be separate from the "goto section" behavior
- It was a bit easy to get lost in the docs as I click around, can we highlight the section the user is in on the table of contents on the left? (e.g. http://getbootstrap.com/css/)

Nothing major, but some UX improvements that would make the docs even more approachable.

ianwremmel:
@bzang all those improvements sound great, but we'll need to backlog them for now. They require pulling the documentation.js theme into the repo and customizing it; I don't see that as being a small undertaking.

ianwremmel:
One of the changes in here is dropping coverage by a full percentage point at prevent about 250 tests from running (down from about 9800)

PR #629, opened 2017-06-26
tentative: fix(ciscospark): adjust tests to allow API contract change

ianwremmel:
Note: this does not change the SDK API contract, only allows a change to an undocumented REST API contract


ianwremmel:
not needed.

PR #631, opened 2017-06-27
fix(internal-plugin-conversation): handle user and convo creation errors

aimeex3:
Previously, when creating a user, it did not handle the scenario when a user cannot be created. Also, it would fail creating a conversation if there were any invalid users. This PR adds an additional option, which allows the conversation to be created without the invalid users.

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
@@ -10,6 +10,7 @@ import {cloneDeep, defaults, isArray, isObject, isString, last, map, merge, omit
 import {readExifData} from '@ciscospark/helper-image';
 import uuid from 'uuid';
 import querystring from 'querystring';
+import {InvalidUserCreation} from './convo-error';
 import ShareActivity from './share-activity';
 import {EventEmitter} from 'events';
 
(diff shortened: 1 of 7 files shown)


packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -81,16 +82,27 @@ const Conversation = SparkPlugin.extend({
...
+          return options.allowPartialCreation ? undefined : Promise.reject(err);
+        })
+    ))
       .then((participants) => {
+
         participants.unshift(this.spark.internal.device.userId);
-        params.participants = uniq(participants);
+        participants = uniq(participants);
+
+        const validParticipants = participants.filter((participant) => !!participant);
```
ianwremmel: the `!!` is actually unnecessary here.

packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -81,16 +82,27 @@ const Conversation = SparkPlugin.extend({
...
+        params.participants = validParticipants;
 
-        if (params.participants.length === 2 && !(options && options.forceGrouped)) {
+        // check if original participants list was to create a 1:1
+        if (participants.length === 2 && !(options && options.forceGrouped)) {
```
ianwremmel: Does this need to fail if `participants.length === 2 && validParticipants.length !== 2`?

packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -81,16 +82,27 @@ const Conversation = SparkPlugin.extend({
       return Promise.reject(new Error(`\`params.participants\` is required`));
     }
 
-    return Promise.all(params.participants.map((participant) => this.spark.internal.user.asUUID(participant, {create: true})))
+    return Promise.all(params.participants.map((participant) =>
```
ianwremmel: please put 85 and 86 on the same line

PR #633, opened 2017-06-29
Update chai

ianwremmel:
Remove use of chai-as-promised from any test that'll run in a web browser since it's no longer compatible with IE 11 and [never will be](https://github.com/domenic/chai-as-promised/issues/209).


ianwremmel:
still need to tweak isRejected a bit, but this is close:

```
    var promise = this.then ? this : this._obj;
    return promise
      .then(function(actual) {
        chai.assert(
          false,
          'expected #{this} to be rejected but it was fulfilled with #{act}',
          null,
          expected,
          actual
        );
        return actual;
      },
      function(reason) {
        if (expected) {
          if (expected instanceof RegExp) {
            // this isn't quite doing substring matching yet.
            chai.assert.match(reason, expected);
          }

          if (typeof expected === 'string') {
            chai.assert.include(reason.toString(), expected);
          }

          if (expected instanceof Error) {
            chai.assert.instanceOf(reason, expected);
          }
        }
      }.bind(this));
  });
```

ianwremmel:
The only changes since review are rebasing against master and the indentation fix on the last commit

ianwremmel:
there have been some additional uses of chai-as-promised since this was written but, more concerningly, many of the avatar tests fail now.

ianwremmel:
screen shot 2017-10-10 at 2 01 59 pm

PR #634, opened 2017-06-29
fix(tooling): add package name linter

ianwremmel:
We inadvertently published two bad releases because of a package name collision. This linter ensures that package names match their directory name

### DIFF
--- a/tooling/lint-package-names.sh
+++ b/tooling/lint-package-names.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+PACKAGES=$(echo packages/node_modules/{*,@ciscospark/*} | xargs -n 1 | sed 's/packages\/node_modules\///' | xargs -n 1 | grep -v '^@ciscospark$')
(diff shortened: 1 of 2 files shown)


tooling/lint-package-names.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+PACKAGES=$(echo packages/node_modules/{*,@ciscospark/*} | xargs -n 1 | sed 's/packages\/node_modules\///' | xargs -n 1 | grep -v '^@ciscospark$')
+for PACKAGE in $PACKAGES; do
```
adamweeks: Should we also check to make sure there are no duplicate names? Although, I don't know how that would be possible.

ianwremmel: yea, that's almost the more important goal, but it turns out to be a side effect of this check.

adamweeks: Pretty impossible to have two folders the same name I guess 😆

PR #636, opened 2017-06-29
dependency cleanup

ianwremmel:
This is the first in a series of PRs that break \#626 into more managable chunks


ianwremmel:
superceded and otherwise no longer relevant

PR #647, opened 2017-07-12
npm link instructions

ianwremmel:


ianwremmel:
Note: this doesn't work. 

Another possibility

```bash
rm -rf ./node_modules/@ciscospark ./node_modules/ciscospark
npm ls -g --json | grep '@ciscospark' | grep -v peer | grep '{' | grep -v '{}' | awk '{print $1}' | sed 's#"##g' | sed 's#:##g' | xargs -n 1 npm link
```

in the sdk, you may also need to run
```bash
rm -rf ./packages/node_modules/@ciscospark/*/node_modules
```
but i would avoid that last one if at all possible. If the last step is required, it may be useful to run `npm install` in the web client after all is said and done.

ianwremmel:
These are all a bunch of ways *not* to make monorepos npm-linkable.

PR #648, opened 2017-07-12
feat(authorization): restore appid support

ianwremmel:


adamweeks:
Package lock???

ianwremmel:
ah, crap. I committed that? I'll fix it.

adamweeks:
Darn, I got excited for npm5

PR #653, opened 2017-07-14
Debounce batcher maxCalls

adamweeks:
Two main features in the two different commits. I've tested all of the plugins that are using batcher and there were no changes needed in their tests.

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
+++ b/packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
@@ -158,7 +158,7 @@ describe(`spark-core`, () => {
 
           const promises = [];
           // eslint-disable-next-line no-unmodified-loop-condition
-          for (let i = 0; i < BATCHER_MAX_CALLS + 1; i++) {
+          for (let i = 0; i < BATCHER_MAX_CALLS; i++) {
             result.push(i);
             promises.push(spark.internal.batcher.request(i));
             assert.notCalled(spark.request);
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
```diff
@@ -158,7 +158,7 @@ describe(`spark-core`, () => {
 
           const promises = [];
           // eslint-disable-next-line no-unmodified-loop-condition
-          for (let i = 0; i < BATCHER_MAX_CALLS + 1; i++) {
```
ianwremmel: use `<=` instead of `+1`. it's slightly more efficient and in keeping with the other place where you already changed to `<=`

ianwremmel:
Also, commit message should read like commands not statements of facts.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
adamweeks: I'm not sure what you mean. I changed it from `i < BATCHER_MAX_CALLS + 1` to `i < BATCHER_MAX_CALLS`. Did you want me to make it `i <= BATCHER_MAX_CALLS + 1`?

adamweeks:
@ianwremmel commit messages updated

packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
ianwremmel: Oh, I guess I misread the change. yea, this appears fine as-is.

ianwremmel: and by misread, i meant "thought the removal of `+ 1` was an addition"

PR #654, opened 2017-07-19
feat(@ciscospark/internal-pulgin-metrics): support call diagnostic event

willzxd:
Add a new function `submitCallDiagnosticEvents` in metrics plugin with callDiagnosticEventsBather to support of sending call diagnostic event to `/clientmetrics`.

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
@@ -1,0 +1,43 @@
+/**!
+ *
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ * @private
+ */
+
+import Batcher from './batcher';
+
+const CallDiagnosticEventsBatcher = Batcher.extend({
+  namespace: `Metrics`,
+
+  prepareItem(item) {
+    // networkType should be a enum value: `wifi`, `ethernet`, `cellular`, or `unknown`.
+    // Browsers cannot provide such information right now. However, it is a required field.
+    const origin = {
+      buildType: process.env.NODE_ENV === `production` ? `prod` : `test`,
+      networkType: `unknown`
+    };
+    item.eventPayload.origin = Object.assign(origin, item.eventPayload.origin);
+    return Promise.resolve(item);
+  },
+
+  prepareRequest(queue) {
+    // Add sent timestamp
+    queue.forEach((item) => {
+      item.eventPayload.originTime.sent = new Date().toISOString();
+    });
+    return Promise.resolve(queue);
+  },
+
+  submitHttpRequest(payload) {
+    return this.spark.request({
+      method: `POST`,
+      service: `metrics`,
+      resource: `clientmetrics`,
+      body: {
+        metrics: payload
+      }
+    });
+  }
+});
+
+export default CallDiagnosticEventsBatcher;
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/call-diagnostic-events-batcher.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/call-diagnostic-events-batcher.js
@@ -22,7 +22,7 @@ function promiseTick(count) {
 }
 
 describe(`plugin-metrics`, () => {
-  describe(`ClientMetricsBatcher`, () => {
+  describe(`callDiagnosticEventsBatcher`, () => {
     let spark;
 
     beforeEach(() => {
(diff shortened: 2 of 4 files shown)


packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
```diff
@@ -0,0 +1,42 @@
...
+import Batcher from './batcher';
+
+const CallDiagnosticEventsBatcher = Batcher.extend({
+  namespace: `Metrics`,
+
+  prepareItem(item) {
+    // Add buildType and networkType
+    const origin = {
+      buildType: process.env.NODE_ENV === `production` ? `prod` : `test`,
+      networkType: `unknown`
```
willzxd: networkType: 
enum:
- wifi
 - ethernet
- cellular
- unknown

I think we cannot get this info from browsers right now. However, it is a required field.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
```diff
@@ -8,7 +8,8 @@ import {inBrowser} from '@ciscospark/common';
 export default {
   device: {
     preDiscoveryServices: {
-      metricsServiceUrl: process.env.METRICS_SERVICE_URL || `https://metrics-a.wbx2.com/metrics/api/v1`
+      metricsServiceUrl: process.env.METRICS_SERVICE_URL || `https://metrics-a.wbx2.com/metrics/api/v1`,
+      diagnosticsServiceUrl: process.env.DIAGNOSTICS_SERVICE_URL || `https://metrics-a.wbx2.com/metrics/api/v1`
```
willzxd: Right now the call diagnostic event API is the same as clientmetrics API. However, from Feb to now, the API has been changed three times. I am not sure whether they will change it some day. This is the reason I want to use a variable to present the `diagnosticsServiceUrl` with a default value. 

It is not necessary to put it before the user logs in. But I am not sure where I should put it.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+    item.eventPayload.origin = Object.assign(origin, item.eventPayload.origin);
+    return Promise.resolve(item);
+  },
+
+  prepareRequest(queue) {
+    // Add sent timestamp
+    return Promise.resolve(queue.map((item) => {
```
willzxd: Change to `array.forEach`.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
ianwremmel: gotcha. thanks. Can you put a comment in the code to that effect?

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/call-diagnostic-events-batcher.js
```diff
@@ -0,0 +1,178 @@
...
+          body: undefined,
+          options
+        });
+      };
+      sinon.spy(spark, `request`);
+    });
+
+    let clock;
+    beforeEach(() => {
+      clock = lolex.install(Date.now());
```
ianwremmel: Any reason to not use `local.install()`?

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -150,5 +156,22 @@ describe(`plugin-metrics`, () => {
           });
       });
     });
+
+    describe(`#submitCallDiagnosticEvents()`, () => {
+      it(`submits a call diagnostic event metrics`, () => {
```
ianwremmel: no need for the work "metrics" in this `it` statement

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/call-diagnostic-events-batcher.js
willzxd: Fixed.

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
willzxd: Fixed.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
willzxd: Done.

PR #656, opened 2017-07-24
Correct repo information for internal-plugin-board

SinZ163:
Also, npm still thinks @ciscospark/plugin-board exists still, and uses this broken link too


ianwremmel:
Thanks for the submission! We've actually got a separate set of changes the should merge sometime this week that addresses this in a more holistic fashion (all the modules have this issue at the moment).

PR #657, opened 2017-07-24
Cross Platform npm Commands

mmaines-cisco:
Some of the ESLint npm commands use unix commands. This PR contains the changes for using Javascript files to build the npm commands instead of using .sh files or unix commands


.vscode/settings.json
```diff
@@ -20,5 +20,6 @@
     "**/dist": true,
     "node_modules/**": true,
     "packages/node_modules/**": false
-  }
+  },
+  "files.eol": "\n"
```
mmaines-cisco: On windows machines CRLF was used as the eol which would not pass the elsint tests. Setting this to "\n" ensures that no matter what OS is used (in MS Code at least) the eol will be LF and will pass eslint tests

ianwremmel: Make sure this is also configured in `.editorconfig`. That's really the right way to do this sort of thing (though also support vscode directly doesn't hurt. It's kinda unfortunate vscode doesn't just build in editorconfig.

tooling/lint-package-names.js
```diff
@@ -0,0 +1,53 @@
+// const shelljs = require(`shelljs`);
+const fs = require(`fs`);
+
+const modules = [];
+
+
+// JS equivelent of "set -e" bash command
+// for (var envVariable in process.env ) {
```
ianwremmel: Commented out on purpose?

tooling/lint-package-names.js
```diff
@@ -0,0 +1,53 @@
...
+//     shelljs.exec(`echo ${envVariable}`);
+// }
+
+// Grab the module names of each package in the packages/node_modules 
+// and package/node_modules/@ciscospark directories containing the ciscospark substring
+// Using a Promise to excecute async functions in a sync manner to pass linting tests
+new Promise((resolve, reject) => {
```
ianwremmel: Implement this file in `tooling/lib` and `tooling/commands`. There are already helpers for enumerating package names and reading package.jsons

build-eslint-command.js
```diff
@@ -0,0 +1,9 @@
+const shelljs = require(`shelljs`);
```
ianwremmel: I don't think you need shelljs. `require('child_process').execSync()` should achieve the same thing.

mmaines-cisco: The main reason for using shelljs here as opposed to execSync is that this file is being linted by eslint and fails via the no-sync enforcement. Is there a proper way to get around this or is using shelljs ok in this instance?

tooling/lint-package-names.js
mmaines-cisco: In the original script the set -e command was the first command run but it never printed outputted anything when run locally. In the JS equivelent here, it lists out ALL environment variable. In the interest of matching the original output from the .sh script i commented this part out but did not wanted to remove it completely if this was in fact intended behavior. Is the outputted list of all environment variables on the machine intended behavior here?

build-eslint-command.js
adamweeks: You can disable eslint for a particular line using `// eslint-disable-next-line no-sync`. Just try to put a code comment above it for why you disabled it.
https://stackoverflow.com/questions/27732209/turning-off-eslint-rule-for-a-specific-line#29592334

mmaines-cisco: Thanks Adam that helps a lot

PR #662, opened 2017-07-26
feat(@ciscospark/internal-plugin-board): add delete partial contents

tran2:
for cases where we want to keep the background image and delete only the annotations

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-people/package.json
+++ b/packages/node_modules/@ciscospark/plugin-people/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ciscospark/plugin-people",
-  "version": "1.12.2",
+  "version": "1.12.3",
   "description": "",
   "license": "MIT",
   "author": "Adam Weeks <adweeks@cisco.com>",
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
```diff
@@ -245,7 +265,7 @@ const Board = SparkPlugin.extend({
           type: contentType,
           encryptionKeyUrl
         },
-          pick(res, `file`, `payload`)
+          pick(res, [`file`, `payload`])
```
ianwremmel: Generally avoid the array option when passing literals. it's an extra array allocation that doesn't add value.

packages/node_modules/@ciscospark/internal-plugin-board/test/unit/spec/board.js
```diff
@@ -239,6 +239,25 @@ describe(`plugin-board`, () => {
     });
   });
 
+  describe(`#deletePartialContent()`, () => {
+
+    before(() => {
+      spark.request.reset();
+      return spark.internal.board.deletePartialContent(channel, [data1]);
```
ianwremmel: I missed this before, sorry. The function under test should generally be called in the test method, rather than before method. This test probably doesn't need a before hook.

tran2: oh yea, you're right. Thanks!

PR #667, opened 2017-08-02
fix(avatar): add defaultAvatar param and fetch all sizes of avatars

aimeex3:
This PR exposes whether an avatar is a default avatar or not. Also, when an avatar is requested, it fetches all sizes of the avatar and stores them into the cache so that they will all match when different sizes are fetched and expire the same time

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-avatar/src/avatar.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-avatar/src/avatar.js
@@ -6,6 +6,7 @@
 
 import AvatarUrlBatcher from './avatar-url-batcher';
 import AvatarUrlStore from './avatar-url-store';
+import {oneFlight} from '@ciscospark/common';
 import {SparkPlugin} from '@ciscospark/spark-core';
 import {defaults} from 'lodash';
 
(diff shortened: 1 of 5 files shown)


packages/node_modules/@ciscospark/internal-plugin-avatar/src/avatar.js
```diff
@@ -36,8 +36,24 @@ const Avatar = SparkPlugin.extend({
    */
   _fetchAvatarUrl(uuid, options) {
     return this.store.get({uuid, size: options.size})
-      .catch(() => this.batcher.request({uuid, size: options.size}))
-        .then((item) => this.store.add(defaults({cacheControl: options.cacheControl}, item)));
+      .catch(() => Promise.all([
+        // fetch all possible sizes of avatar and store in cache
+        this.batcher.request({uuid, size: 40}),
```
ianwremmel: move these sizes to the avatar config and then just loop over them here.

ianwremmel: you'll need a set of unit tests that asserts the number of api calls that go out. Take a look at Xiaodan's PR that for how to spy on spark.request.

aimeex3: looks like avatar tests already check that spark.request is only called once. i updated the mocked return values to make them work with my changes

PR #671, opened 2017-08-04
chore(tooling): update build pipeline to npm 5

ianwremmel:


ianwremmel:
maybe. it might actually be enough to just let jenkins run `npm install` and commit `package-lock.json` at that point since we're not directly merging PRs.

ianwremmel:
but, that's also part of why i haven't `package-lock` yet :)

ianwremmel:
blocked by https://github.com/npm/npm/issues/16807

docker/builder/Dockerfile
```diff
@@ -23,6 +23,8 @@ RUN tar -xzf "sc-${SAUCE_CONNECT_VERSION}-linux.tar.gz"
 RUN ln -s "${SC_BINARY}" /usr/local/bin/sc
 RUN chmod ugo+w /sauce
 
+RUN npm install -g npm@latest
```
adamweeks: Should we do `npm@5` instead?

ianwremmel: maybe, but it doesn't much matter right now. until https://github.com/npm/npm/issues/16807 gets fixed, we can't use npm 5 in the pipeline.

ianwremmel:
closing for now.

PR #672, opened 2017-08-04
centralize sauce logic

ianwremmel:


adamweeks:
One thing that would be cool would be the ability to have a command that sets up the tunnel, runs command over tunnel, then closes tunnel when done.

ianwremmel:
yea. i used to have that instead of the three commands, but i found the overhead wasn't worth it; we could maybe enhance run to do that if you don't already have a tunnel running a print a warning that tells you about start and stop. that can come later though. this was mostly motivated by wanting to replace `docker/builder/cmd.sh` with `docker/builder/cmd.js`. There are a lot of logging improvements we can make if we can hook directly into the tooling api instead of just throwing out shell commands.

PR #678, opened 2017-08-09
feat(@ciscospark/plugin-people): add support for people.get('me')

### LINKED ISSUE #676: people.get("me")  no longer works
when using the ciscospark nodejs lib people.get("me") no longer works.  This has broken the botkit library.

The problem is in the @ciscospark module.

https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/%40ciscospark/plugin-people/src/people-batcher.js#L69

Should be:

resource: `people/${ids}`

Also the handleHttpSuccess function should check to see if there are items...because the get("me") will not return an array of items.

adamweeks:
fixes #676

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-people/src/people.js
+++ b/packages/node_modules/@ciscospark/plugin-people/src/people.js
@@ -3,7 +3,7 @@
  */
 
 import {SparkPlugin, Page} from '@ciscospark/spark-core';
-import {base64} from '@ciscospark/common';
+import {base64, oneFlight} from '@ciscospark/common';
 
 import PeopleBatcher from './people-batcher';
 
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/plugin-people/src/people.js
```diff
@@ -58,6 +58,13 @@ const People = SparkPlugin.extend({
     if (!person) {
       return Promise.reject(new Error(`A person with an id is required`));
     }
+    if (person === `me`) {
```
ianwremmel: should this be moved to a private method and wrapped in `@oneFlight`?

packages/node_modules/@ciscospark/plugin-people/test/integration/spec/people.js
```diff
@@ -39,6 +39,14 @@ describe(`plugin-people`, function() {
             assert.equal(peopleResponse.emails[0], mccoy.email);
           });
       });
+
+      it(`gets the current user with "get('me')"`, () => {
```
ianwremmel: Maybe also add test that does `get('me')` and `get(mccoy.id)` in a promise.all. it's overkill given the current implementation, but from a black-box point of view, it'll be a good test to have in place long-term.

adamweeks:
@ianwremmel ready for round 2

PR #682, opened 2017-08-18
Shared metrics context (Client -> SDK, SDK -> Client)

UWTech:
Add shared context for metrics between Client and SDK

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
@@ -15,7 +15,9 @@ const Metrics = SparkPlugin.extend({
     callDiagnosticEventsBatcher: CallDiagnosticEventsBatcher
   },
 
-  metricMap: new Map(),
+  props: {
+    metricMap: Map
+  },
 
   CLIENT_METRIC: `clientMetric`,
   SDK_METRIC: `sdkMetric`,
@@ -152,7 +154,7 @@ const Metrics = SparkPlugin.extend({
   },
 
   /**
-  * Registers a metric to be used to later update or call finish method
+  * Registers a client metric to be used to later update or call finish method
   * @param {string} eventName name of the metric
   * @param {string} id unique identifier
   * @param {Object} props to populate the metric with
@@ -169,7 +171,7 @@ const Metrics = SparkPlugin.extend({
   * @param {Object} props to populate the metric with
   * @returns {bool} true if metric registered, false otherwise
   */
-  registerSdkMetric(eventName, id, props) {
+  registerMetric(eventName, id, props) {
     return this._registerMetric(eventName, id, props, this.SDK_METRIC);
   },
 
@@ -203,7 +205,7 @@ const Metrics = SparkPlugin.extend({
    * @param {string} id of the metric to be sent
    * @returns {Object} HttpResponse object
    */
-  clientSendRegisteredMetric(id) {
+  sendClientRegisteredMetric(id) {
     if (!this.metricIsType(id, this.CLIENT_METRIC)) {
       return false;
     }
@@ -217,7 +219,7 @@ const Metrics = SparkPlugin.extend({
    * @param {string} id of the metric to be sent
    * @returns {Object} HttpResponse object
    */
-  sdkSendRegisteredMetric(id) {
+  sendRegisteredMetric(id) {
     if (!this.metricIsType(id, this.SDK_METRIC)) {
       return false;
     }
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
@@ -7,7 +7,6 @@ import Batcher from './batcher';
 import ClientMetricsBatcher from './client-metrics-batcher';
 import CallDiagnosticEventsBatcher from './call-diagnostic-events-batcher';
 import {deprecated} from '@ciscospark/common';
-import _ from 'lodash';
 
 const Metrics = SparkPlugin.extend({
   children: {
@@ -32,7 +31,13 @@ const Metrics = SparkPlugin.extend({
     return this.batcher.request(Object.assign({key}, value));
   },
 
-  createPayload(eventName, props) {
+  /**
+   * generates a named metrics payload
+   * @param {string} eventName
+   * @param {Object} props
+   * @returns {Object} payload
+   */
+  _createPayload(eventName, props) {
     const payload = {metricName: eventName};
     if (props.tags) {
       payload.tags = props.tags;
@@ -54,7 +59,7 @@ const Metrics = SparkPlugin.extend({
    * @returns {Object} HttpResponse object
    */
   submitClientMetrics(eventName, props, preLoginId) {
-    const payload = this.createPayload(eventName, props);
+    const payload = this._createPayload(eventName, props);
 
     if (preLoginId) {
       const _payload = {
@@ -67,10 +72,10 @@ const Metrics = SparkPlugin.extend({
       // their metrics will not post from a queue flush in time
       return this.postPreLoginMetric(_payload, preLoginId);
     }
-    return this.sendBatchMetric(payload);
+    return this._sendBatchMetric(payload);
   },
 
-  sendBatchMetric(payload) {
+  _sendBatchMetric(payload) {
     return this.clientMetricsBatcher.request(payload);
   },
 
@@ -121,10 +126,6 @@ const Metrics = SparkPlugin.extend({
     return this.metricMap.set(id, tuple);
   },
 
-  getMetrics() {
-    return this.metricMap;
-  },
-
   getMetric(id) {
     return this.metricMap.get(id);
   },
@@ -133,10 +134,18 @@ const Metrics = SparkPlugin.extend({
     return this.metricMap.delete(id);
   },
 
-  registerMetric(eventName, id, props, type) {
+  /**
+   * registers (stores) a metric if it does not exist
+   * @param {string} eventName
+   * @param {string} id
+   * @param {Object} props
+   * @param {string} type
+   * @returns {bool}
+   */
+  _registerMetric(eventName, id, props, type) {
     if (!this.getMetric(id)) {
       // hash name and props into table/object
-      const payload = this.createPayload(eventName, props);
+      const payload = this._createPayload(eventName, props);
       return this.storeMetric(id, payload, type);
     }
     return false;
@@ -149,8 +158,8 @@ const Metrics = SparkPlugin.extend({
   * @param {Object} props to populate the metric with
   * @returns {bool} true if metric registered, false otherwise
   */
-  clientRegisterMetric(eventName, id, props) {
-    return this.registerMetric(eventName, id, props, this.CLIENT_METRIC);
+  registerClientMetric(eventName, id, props) {
+    return this._registerMetric(eventName, id, props, this.CLIENT_METRIC);
   },
 
   /**
@@ -160,8 +169,8 @@ const Metrics = SparkPlugin.extend({
   * @param {Object} props to populate the metric with
   * @returns {bool} true if metric registered, false otherwise
   */
-  sdkRegisterMetric(eventName, id, props) {
-    return this.registerMetric(eventName, id, props, this.SDK_METRIC);
+  registerSdkMetric(eventName, id, props) {
+    return this._registerMetric(eventName, id, props, this.SDK_METRIC);
   },
 
   /**
@@ -180,7 +189,7 @@ const Metrics = SparkPlugin.extend({
     }
 
     // update metric
-    _.assign(props.fields, metricProps.fields);
+    Object.assign(props.fields, metricProps.fields);
 
     return this.storeMetric(id, props, type);
   },
@@ -198,7 +207,7 @@ const Metrics = SparkPlugin.extend({
     if (!this.metricIsType(id, this.CLIENT_METRIC)) {
       return false;
     }
-    const httpResp = this.sendBatchMetric(this.getMetric(id).metric);
+    const httpResp = this._sendBatchMetric(this.getMetric(id).metric);
     this.deleteMetric(id);
     return httpResp;
   },
@@ -212,7 +221,7 @@ const Metrics = SparkPlugin.extend({
     if (!this.metricIsType(id, this.SDK_METRIC)) {
       return false;
     }
-    const httpResp = this.sendBatchMetric(this.getMetric(id).metric);
+    const httpResp = this._sendBatchMetric(this.getMetric(id).metric);
     this.deleteMetric(id);
     return httpResp;
   }
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
@@ -16,7 +16,9 @@ const Metrics = SparkPlugin.extend({
   },
 
   props: {
-    metricMap: Map
+    metricMap: {
+      type: `object`
+    }
   },
 
   CLIENT_METRIC: `clientMetric`,
@@ -138,6 +140,10 @@ const Metrics = SparkPlugin.extend({
 
   /**
    * registers (stores) a metric if it does not exist
+   * returns false if the metric has been registered by another client
+   * operation, or if the metric has been registered by SDK.
+   * On false return, user can use secondary hashing algorithm to store under
+   * different key if desired, or handle as they feel appropriate.
    * @param {string} eventName
    * @param {string} id
    * @param {Object} props
@@ -155,6 +161,10 @@ const Metrics = SparkPlugin.extend({
 
   /**
   * Registers a client metric to be used to later update or call finish method
+  * returns false if the metric has been registered by another client
+  * operation, or if the metric has been registered by SDK.
+  * On false return, user can use secondary hashing algorithm to store under
+  * different key if desired, or handle as they feel appropriate.
   * @param {string} eventName name of the metric
   * @param {string} id unique identifier
   * @param {Object} props to populate the metric with
@@ -166,6 +176,10 @@ const Metrics = SparkPlugin.extend({
 
   /**
   * Registers a metric to be used to later update or call finish method
+  * returns false if the metric has been registered by another client
+  * operation, or if the metric has been registered by SDK.
+  * On false return, user can use secondary hashing algorithm to store under
+  * different key if desired, or handle as they feel appropriate.
   * @param {string} eventName name of the metric
   * @param {string} id unique identifier
   * @param {Object} props to populate the metric with
@@ -177,6 +191,10 @@ const Metrics = SparkPlugin.extend({
 
   /**
    * Updates a registered metric, if it exists
+   * returns false if the metric has been registered by another client
+   * operation, or if the metric has been registered by SDK.
+   * caller can process false response as appropriate
+   * (e.g.: secondary hash method lookup, etc.)
    * @param {string} id lookup key for metric
    * @param {Object} metricProps to be used for updating metric
    * @returns {bool} true if the metrics exists and was updated, false otherwise
@@ -197,7 +215,8 @@ const Metrics = SparkPlugin.extend({
   },
 
   metricIsType(id, type) {
-    return this.getMetric(id) && this.getMetric(id).type === type;
+    const metric = this.getMetric(id);
+    return metric && metric.type === type;
   },
 
   /**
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+  clientRegisterMetric(eventName, id, props) {
+    return this.registerMetric(eventName, id, props, this.CLIENT_METRIC);
+  },
+
+  /**
+  * Registers a metric to be used to later update or call finish method
+  * @param {string} eventName name of the metric
+  * @param {string} id unique identifier
+  * @param {Object} props to populate the metric with
+  * @returns {bool} true if metric registered, false otherwise
```
ianwremmel: Is this bool expected to provide value outside of tests? If not, a spy may be more appropriate.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+    const storedMetric = this.getMetric(id);
+    const props = storedMetric ? storedMetric.metric : null;
+    const type = storedMetric ? storedMetric.type : null;
+
+    if (!props || !type || !metricProps) {
+      return false;
+    }
+
+    // update metric
+    _.assign(props.fields, metricProps.fields);
```
ianwremmel: Use `Object.assign`. Generally, you want to avoid using lodash methods if there's a native method available.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -55,9 +67,12 @@ const Metrics = SparkPlugin.extend({
       // their metrics will not post from a queue flush in time
       return this.postPreLoginMetric(_payload, preLoginId);
     }
-    return this.clientMetricsBatcher.request(payload);
+    return this.sendBatchMetric(payload);
   },
 
+  sendBatchMetric(payload) {
```
ianwremmel: This should probably be prefixed with an `_`.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -26,14 +32,7 @@ const Metrics = SparkPlugin.extend({
...
 
-  /**
-   * This corresponds to #sendSemiStructured() in the deprecated metrics handler
-   * @param {string} eventName
-   * @param {Object} props
-   * @param {string} preLoginId
-   * @returns {Object} HttpResponse object
-   */
-  submitClientMetrics(eventName, props, preLoginId) {
+  createPayload(eventName, props) {
```
ianwremmel: Please add doc block (to all new methods).

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+  },
+
+  /**
+  * Registers a metric to be used to later update or call finish method
+  * @param {string} eventName name of the metric
+  * @param {string} id unique identifier
+  * @param {Object} props to populate the metric with
+  * @returns {bool} true if metric registered, false otherwise
+  */
+  clientRegisterMetric(eventName, id, props) {
```
ianwremmel: Let's call this `registerClientMetric()`.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+  },
+
+  /**
+  * Registers a metric to be used to later update or call finish method
+  * @param {string} eventName name of the metric
+  * @param {string} id unique identifier
+  * @param {Object} props to populate the metric with
+  * @returns {bool} true if metric registered, false otherwise
+  */
+  sdkRegisterMetric(eventName, id, props) {
```
ianwremmel: Let's call this `registerMetric()`. 

If there's already a client metric, this should behave like `updateRegisteredMetric()`, right?

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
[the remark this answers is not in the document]
UWTech: Done

UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+
+  getMetric(id) {
+    return this.metricMap.get(id);
+  },
+
+  deleteMetric(id) {
+    return this.metricMap.delete(id);
+  },
+
+  registerMetric(eventName, id, props, type) {
```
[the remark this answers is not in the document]
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: This is an abstraction of the underlying implementation so we can change the functionality of the method later. It is not strictly necessary, but prevents us from having to use the same preprocessing for both client and SDK metrics.

Is just a simple wrapper.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+  metricIsType(id, type) {
+    return this.getMetric(id) && this.getMetric(id).type === type;
+  },
+
+  /**
+   * Sends a registered metric if it exists
+   * @param {string} id of the metric to be sent
+   * @returns {Object} HttpResponse object
+   */
+  clientSendRegisteredMetric(id) {
```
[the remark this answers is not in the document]
UWTech: So this method actually sends the metric. Register simply creates and stores the metric

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
+    this.deleteMetric(id);
+    return httpResp;
+  },
+
+  /**
+   * Sends a registered metric if it exists
+   * @param {string} id of the metric to be sent
+   * @returns {Object} HttpResponse object
+   */
+  sdkSendRegisteredMetric(id) {
```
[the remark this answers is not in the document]
UWTech: So the idea here is that if you are the original creator of the metric, be you the client or the sdk, you are the party that has ultimate authority to send the metric. In order to otherwise know the role of the party, we would have to pass a param to the method.

Using a wrapper like this ensures that we are not unintentionally sending a metric that is registered and owned by the client in the SDK, or vice versa. 

I'm not sure I understand the argument for only having a single send metric method. Are you suggesting a paramaterized version where we send the identifier as an argument, (e.g.: "SDK" or "Client")?

So that would look like sendMetric(id, requestingParty){
if(getMetric(id).type != requestingParty) {
// party does not own this metric, cannot send, only update!
return false
}
else {
send metric
}

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -172,5 +173,150 @@ describe(`plugin-metrics`, () => {
...
+        assert.isFalse(metrics.sdkSendRegisteredMetric(`2`));
+      });
+      it(`does not send null metric`, () => {
+        assert.isFalse(metrics.sdkSendRegisteredMetric(`4`));
+      });
+    });
+
+    describe(`#sendUnstructured`, () => {
+      it(`sends an unstructured metric`, () => {
+        metrics.sendUnstructured(`foo`, `bar`)
```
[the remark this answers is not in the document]
UWTech: This "test" was just filler. I was just trying to generate coverage because the pre-check was failing. I was only going for coverage rather than value. Figured that was okay since the method was deprecated and previously uncovered by tests anyway.

I'll double check the others to be sure though.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: The intention here is for the bool to potentially be used by whatever consumer is requesting a metric be registered. This could be used to determine if there is some problem with the metric registration/deletion process since this will be managed by way of register/update/send methods.

In the event a metric has already been registered with that ID, the consumer could read the FALSE return, and take some action (console log, send failure metric, secondary hashing, etc.).

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -7,6 +7,7 @@ import Batcher from './batcher';
 import ClientMetricsBatcher from './client-metrics-batcher';
 import CallDiagnosticEventsBatcher from './call-diagnostic-events-batcher';
 import {deprecated} from '@ciscospark/common';
+import _ from 'lodash';
```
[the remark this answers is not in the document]
UWTech: Removed all together. Using Object.assign instead.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -15,6 +16,11 @@ const Metrics = SparkPlugin.extend({
     callDiagnosticEventsBatcher: CallDiagnosticEventsBatcher
   },
 
+  metricMap: new Map(),
```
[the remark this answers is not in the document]
UWTech: Wouldn't we want to share the context information across instances?

I could conceive of situations where we'd want to share the information across different modules that may not necessarily be talking to the same instance.

I believe in the case of the web-client, we do have a single instance of Spark-JS-SDK, but seems it might be useful.

Is the thinking this is not expected behavior from most people; e.g.: if I create a new instance, I expect all the context information to be fresh, and that the instance does not consider the other metrics I've started to create for the purposes of sharing between the SDK and the client?

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +114,107 @@ const Metrics = SparkPlugin.extend({
...
     };
     return this.callDiagnosticEventsBatcher.request(event);
+  },
+
+  storeMetric(id, metricProps, metricType) {
+    const tuple = {metric: metricProps, type: metricType};
+    return this.metricMap.set(id, tuple);
+  },
+
+  getMetrics() {
```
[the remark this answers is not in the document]
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -172,5 +173,150 @@ describe(`plugin-metrics`, () => {
...
+        metrics.registerMetric(`testNullType`, `3`, testPayload, null);
+        assert.isFalse(metrics.updateRegisteredMetric(`3`, `someValue`));
+      });
+    });
+
+    describe(`#clientSendRegisteredMetric`, () => {
+      it(`sends client metric`, () => {
+        metrics.registerMetric(`horse`, `1`, testPayload, `clientMetric`);
+        metrics.clientSendRegisteredMetric(`1`)
+        .then(() => {
```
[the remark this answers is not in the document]
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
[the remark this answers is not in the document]
ianwremmel: the metrics plugin will be instantiated once for each sdk instance. Typically, you would only make a second sdk instance if you had a different set of credentials (e.g. for a node web server that handles multiple user requests). In those cases, you definitely don't want to share the metrics map across instances.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
ianwremmel: yea, understood. i was thinking it made more sense for the method called by the sdk to be a pretty generic name, and then the method called by consumers would be obviously for consumers.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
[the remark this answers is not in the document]
ianwremmel: gotcha, yea, i misread the name a bit. similarly to the sdk method above, it seems to me the use-by-sdk methods should be generically named and the use-by-client methods should have the "client" moniker. that said, `sendClientRegisteredMetric()` would be more appropriate (method names should generally be verbs or, at least, verb-like).

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: Right, right; got it. Since we're thinking about this in the context of the SDK, the generic method should be used by the SDK, whereas the explicitly differentiated one would be for external users.

I can change the names of the SDK ones, sure.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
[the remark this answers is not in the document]
ianwremmel: i was still thinking there'd be two send methods: `sendRegisteredMetric()` and `sendRegisteredClientMetric()`. `sendRegisteredMetric()` should be a noop if `registerClientMetric()` is called for that key before `registerMetric()` is called for that key. That way, the sdk can report a given dataset without client input via 

- sdk: `registerMetric()`
- sdk: `updateMetric()` 
- sdk: `sendMetric()`

should a client choose to participate, the flow would be 

- client: `registerClientMetric()`
- sdk: `registerMetric()`
- sdk: `updateMetric()` 
- sdk: `sendMetric()` (noop in this case)
- client: `sendClientMetric()`

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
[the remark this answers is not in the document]
ianwremmel: yea, but that would mean that, should this test start failing at some point, we'd never know.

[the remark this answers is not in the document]
UWTech: fixed

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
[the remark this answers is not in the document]
UWTech: done

UWTech:
Addressed PR comments.

Not super familiar with Ampersand, so let me know if there is some other pattern you'd like followed for the map instantiation.

Cheers.

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -172,5 +173,150 @@ describe(`plugin-metrics`, () => {
...
+            assert.property(metric, `metricName`);
+            assert.property(metric, `tags`);
+            assert.property(metric, `fields`);
+
+            assert.equal(metric.metricName, `horse`);
+            assert.equal(metric.tags.success, true);
+            assert.equal(metric.fields.perceivedDurationInMillis, testPayload.fields.perceivedDurationInMillis);
+          });
+      });
+      it(`does not send non-sdj metric`, () => {
```
ianwremmel: "sdk"

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -99,6 +121,111 @@ const Metrics = SparkPlugin.extend({
...
+    }
+
+    // update metric
+    Object.assign(props.fields, metricProps.fields);
+
+    return this.storeMetric(id, props, type);
+  },
+
+  metricIsType(id, type) {
+    return this.getMetric(id) && this.getMetric(id).type === type;
```
ianwremmel: store the result of `getMetric()` so that you're only doing one lookup. you could also use lodash's `get()`

```javascript
return get(this.getMetric(id), `type`) === type;
```

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -172,5 +173,150 @@ describe(`plugin-metrics`, () => {
...
+      });
+      it(`does not send null metric`, () => {
+        assert.isFalse(metrics.sendRegisteredMetric(`4`));
+      });
+    });
+
+    describe(`#sendUnstructured`, () => {
+      it(`sends an unstructured metric`, () => {
+        return metrics.sendUnstructured(`foo`, `bar`)
+        .then(() => {
```
ianwremmel: indent `.then()`

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -172,5 +173,150 @@ describe(`plugin-metrics`, () => {
...
+        assert.isFalse(metrics.sendRegisteredMetric(`2`));
+      });
+      it(`does not send null metric`, () => {
+        assert.isFalse(metrics.sendRegisteredMetric(`4`));
+      });
+    });
+
+    describe(`#sendUnstructured`, () => {
+      it(`sends an unstructured metric`, () => {
+        return metrics.sendUnstructured(`foo`, `bar`)
```
ianwremmel: remove the return, move `metrics...` up a line, and use the arrow function's implicit return.

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
UWTech: done
Not sure there is a big cost differential between to two O(1) lookups versus comparing local variable, and one uses more bytes, but is a small enough change.

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/src/metrics.js
```diff
@@ -15,6 +15,13 @@ const Metrics = SparkPlugin.extend({
     callDiagnosticEventsBatcher: CallDiagnosticEventsBatcher
   },
 
+  props: {
+    metricMap: Map
```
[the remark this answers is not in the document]
UWTech: done

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
UWTech: done

UWTech:
Made changes as requested.

UWTech:
Created README

UWTech:
@ianwremmel  any word on when we might get this merged in?

No specific task we have right now that requires this, but would be nice to get this in for use around metrics we may need.

Cheers.

PR #684, opened 2017-08-23
fix(helper-html): make the javascript protocol filter case insensitive

### LINKED ISSUE #683: Anchor tags with href protocol set to javascript isn't stripped out if capital letters are used
`<a href='javascript:console.log("hello");'>bad</a>` is stripped out but if you change the case
`<a href='JaVaScRiPt:console.log("hello");'>bad</a>` is not removed.

mwegman:
Fixes #683

### DIFF
--- a/packages/node_modules/@ciscospark/helper-html/src/html.shim.js
+++ b/packages/node_modules/@ciscospark/helper-html/src/html.shim.js
@@ -109,8 +109,8 @@ function _filterSync(processCallback, allowedTags, allowedStyles, html) {
           node.removeAttribute(attrName);
         }
         else if (attrName === `href` || attrName === `src`) {
-          const attrValue = node.attributes.getNamedItem(attrName).value.trim();
-          if (attrValue.indexOf(`javascript:`) === 0) {
+          const attrValue = node.attributes.getNamedItem(attrName).value.trim().toLowerCase();
+          if (attrValue.indexOf(`javascript:`) === 0 || attrValue.indexOf(`vbscript:`) === 0) {
             reparent(node);
           }
         }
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/helper-html/test/unit/spec/html.js
```diff
@@ -109,7 +109,7 @@ skipInNode(describe)(`html`, () => {
     },
     {
       it: `filters javascript: from a href`,
-      input: `<p><a href="javascript:window.close(); return false">click here</a></p>`,
```
adamweeks: Could we keep the original test case with an all lowercase test and make these mixed case tests a new test?

packages/node_modules/@ciscospark/helper-html/src/html.shim.js
```diff
@@ -109,8 +109,8 @@ function _filterSync(processCallback, allowedTags, allowedStyles, html) {
           node.removeAttribute(attrName);
         }
         else if (attrName === `href` || attrName === `src`) {
-          const attrValue = node.attributes.getNamedItem(attrName).value.trim();
-          if (attrValue.indexOf(`javascript:`) === 0) {
+          const attrValue = node.attributes.getNamedItem(attrName).value.trim().toLocaleLowerCase();
```
ianwremmel: please use `toLowerCase()` instead of `toLocaleLowerCase()`. We use `toLowerCase()` everywhere else. I'm open to consistently change to `toLocaleLowerCase()`, but IIRC, it actually has unexpected behaviors and doesn't help us in this case (since 'vbscript:' or 'javascript:' aren't localized).

PR #685, opened 2017-08-25
chore(tooling): use proper collation for next-version detection

ianwremmel:


tooling/lib/version.js
```diff
@@ -167,6 +193,8 @@ async function getChangeType() {
  * @returns {string}
  */
 function increment(type, version) {
+  debug(`incrementing ${version} by ${type}`);
```
mwegman: Do you want to leave in your debug statement?

ianwremmel: yea. there's a bunch of them scattered throughout ./tooling. [debug](https://www.npmjs.com/package/debug) is a handy package for writing really verbose logging that doesn't print by default. This'll only output if the appropriate environment variable is set.

PR #688, opened 2017-08-30
fix(@ciscospark/plugin-phone): rely on server info to detect calls

ianwremmel:
Solves *part* of the bug that will result from delta events, but not all of it.


ianwremmel:
Delta events are locus notifications from mercury that only include part of a locus (the DTO is getting too big for the websocket channel to practically carry). Delta events are diffs of the previous locus to the current locus. It turns out we never should have been relying on participant count to determine if it's a "call". With that mistake combined with the possibility of delta events before we're ready to fully support delta events, a delta event could have zero participants (because the participants didn't change), thus causing the logic this PR fixes to assume the call had ended.

adamweeks:
Cool, thanks for the info @ianwremmel!

PR #689, opened 2017-08-31
fix(@ciscospark/http-core): upgrade package "qs"

CiscoTalentIT:
A security vulnerability was reported for the existing version of package qs. Upgraded it to fix the issue.

 High severity vulnerability found on qs@5.2.1
- desc: Prototype Override Protection Bypass
- info: https://snyk.io/vuln/npm:qs:20170213

### DIFF
--- a/packages/node_modules/@ciscospark/helper-html/package.json
+++ b/packages/node_modules/@ciscospark/helper-html/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ciscospark/helper-html",
-  "version": "1.9.3",
+  "version": "1.10.1",
   "description": "HTML Utiltities",
   "license": "MIT",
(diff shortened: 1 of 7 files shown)


ianwremmel:
Please use interactive rebase to remove extraneous commits. Also note that the scope should be the full package name `@ciscospark/http-core`. The full commit message should be lowercase and less than 72 characters on the first line. The message should be imperative present tense (i.e. state what you're doing, not what you did). In this case, `upgrade package "qs"`. Additional details about the vulnerability can go in the body of the commit message.

if you run `npm install`, a precommit hook should be installed which'll lint the commit message before letting you commit it.

(I'm rather pedantic about commit messages because we eventually intend to generate changelogs from them)

Finally, who are you? What's your relationship with Cisco Spark?

PR #696, opened 2017-09-11
Call Memberships

ianwremmel:


ianwremmel:
note: the first three commits have been reviewed elsewhere

packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
```diff
@@ -6,11 +6,14 @@ import {Batcher} from '@ciscospark/spark-core';
 
 /**
  * @class
+ * @ignore
```
adamweeks: TIL: `@ignore`

packages/node_modules/@ciscospark/plugin-phone/src/call-memberships.js
```diff
@@ -0,0 +1,30 @@
+import AmpCollection from 'ampersand-collection';
+import {find} from 'lodash';
+
+import CallMempership from './call-membership';
```
adamweeks: `s/CallMempership/CallMembership`

packages/node_modules/@ciscospark/plugin-phone/src/state-parsers.js
```diff
@@ -70,10 +70,85 @@ export function joinedOnThisDevice(spark, locus) {
...
+  const users = new Map();
+  const devices = new Map();
+
+  for (const participant of locus.participants) {
+    if (participant.type === `USER`) {
+      users.set(participant.url, participant);
+    }
+    else if (participant.type === `RESOURCE_ROOM`) {
+      devices.set(participant.url, participant);
+    }
```
adamweeks: Do we want to track unknown participants?

ianwremmel: no one has been able to tell me what `UNKNOWN` means, so, at this time, `USER` and `RESOURCE_ROOM` seem to be the only types that provide useful information. The fourth type, `MEETING_BRIDGE` has been officially established to provide no client-useful information (though I can imagine a future where it provides UI-relevant information to call but not any specific member)

packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
ianwremmel: it's an alias for `@private`, but I'm trying to differentiate between things I don't want in the public docs and things that should not be documented. The whole people batcher class should be documented and used by internal code, it just doesn't belong on the developer site.

packages/node_modules/@ciscospark/plugin-phone/src/state-parsers.js
adamweeks: Gotcha.

PR #697, opened 2017-09-13
refactor(@ciscospark/plugin-people): add optional showAllTypes flag

UWTech:
add optional showAllTypes flag for Hydra people request

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-people/src/config.js
+++ b/packages/node_modules/@ciscospark/plugin-people/src/config.js
@@ -7,8 +7,10 @@ export default {
     batcherWait: 100,
     batcherMaxCalls: 10,
     batcherMaxWait: 1500,
-    // optional flag that requires Hydra to send every type field,
-    // even if the type is not "person" (e.g.: SX10, webhook_intergation, etc.)
+    /**
+     * optional flag that requires Hydra to send every type field,
+     * even if the type is not "person" (e.g.: SX10, webhook_intergation, etc.)
+     */
     showAllTypes: false
   }
 };
--- a/packages/node_modules/@ciscospark/plugin-people/src/config.js
+++ b/packages/node_modules/@ciscospark/plugin-people/src/config.js
@@ -6,6 +6,9 @@ export default {
   people: {
     batcherWait: 100,
     batcherMaxCalls: 10,
-    batcherMaxWait: 1500
+    batcherMaxWait: 1500,
+    // optional flag that requires Hydra to send every type field,
+    // even if the type is not "person" (e.g.: SX10, webhook_intergation, etc.)
+    showAllTypes: false
   }
 };
--- a/packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
+++ b/packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
@@ -9,9 +9,6 @@ import {Batcher} from '@ciscospark/spark-core';
  */
 const PersonUUIDRequestBatcher = Batcher.extend({
   namespace: `People`,
-  // optional flag that requires Hydra to send every type field,
-  // even if the type is not "person" (e.g.: SX10, webhook_intergation, etc.)
-  showAllTypes: false,
 
   /**
    * @param {HttpResponseObject} res
@@ -69,7 +66,7 @@ const PersonUUIDRequestBatcher = Batcher.extend({
   submitHttpRequest(ids) {
     return this.spark.request({
       service: `hydra`,
-      resource: `people/?id=${ids}&showAllTypes=${this.showAllTypes}`
+      resource: `people/?id=${ids}&showAllTypes=${this.config.showAllTypes}`
     });
   }
 });
--- a/packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
+++ b/packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
@@ -9,6 +9,9 @@ import {Batcher} from '@ciscospark/spark-core';
  */
 const PersonUUIDRequestBatcher = Batcher.extend({
   namespace: `People`,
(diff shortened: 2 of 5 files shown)


packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
```diff
@@ -9,6 +9,9 @@ import {Batcher} from '@ciscospark/spark-core';
  */
 const PersonUUIDRequestBatcher = Batcher.extend({
   namespace: `People`,
+  // optional flag that requires Hydra to send every type field,
+  // even if the type is not "person" (e.g.: SX10, webhook_intergation, etc.)
+  showAllTypes: false,
```
ianwremmel: Put this in `plugin-people/src/config.js` instead of here.

packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
```diff
@@ -66,7 +69,7 @@ const PersonUUIDRequestBatcher = Batcher.extend({
   submitHttpRequest(ids) {
     return this.spark.request({
       service: `hydra`,
-      resource: `people/?id=${ids}`
+      resource: `people/?id=${ids}&showAllTypes=${this.showAllTypes}`
```
ianwremmel: instead of `this.showAllTypes`, you want `this.config.showAllTypes`

packages/node_modules/@ciscospark/plugin-people/src/people.js
```diff
@@ -142,7 +144,8 @@ const People = SparkPlugin.extend({
    *   });
    *   // => success
    */
-  list(options) {
+  list(options, showAllTypes = false) {
+    this.batcher.showAllTypes = showAllTypes;
```
ianwremmel: If list gets called twice with showAllTypes = true and showAllTypes = false, depending on timing, you could get unexpected behavior in the response.

packages/node_modules/@ciscospark/plugin-people/src/people-batcher.js
```diff
@@ -9,6 +9,9 @@ import {Batcher} from '@ciscospark/spark-core';
  */
 const PersonUUIDRequestBatcher = Batcher.extend({
   namespace: `People`,
+  // optional flag that requires Hydra to send every type field,
```
ianwremmel: Does `showAllTypes` add the `type` field to the response, or does it cause the response to include more than just `type == person` people?

UWTech: The "showAllTypes" flag will include the "type" field in cases where it otherwise would have been omitted. For example, in the case of a "webhook_integration" the type field is omitted entirely without the inclusion of the additional query parameter "showAllTypes=true".

The type field for "person" will always be included without any special query parameters necessary.

packages/node_modules/@ciscospark/plugin-people/src/people.js
UWTech: That is true; I was trying to avoid having to extend the "query" parameter, and thought this edge case wasn't a big deal.

I'll try to think of a simple way to make this work on a per request basis without a race condition, and without having to extend request.

packages/node_modules/@ciscospark/plugin-people/src/config.js
```diff
@@ -6,6 +6,9 @@ export default {
   people: {
     batcherWait: 100,
     batcherMaxCalls: 10,
-    batcherMaxWait: 1500
+    batcherMaxWait: 1500,
+    // optional flag that requires Hydra to send every type field,
```
ianwremmel: Use jsdoc formatted comments. config isn't well documented yet, but check a few of the other plugin configs and you should find one that has all the required doc tags.

packages/node_modules/@ciscospark/recipe-private-web-client/src/index.js
```diff
@@ -18,6 +18,7 @@ import '@ciscospark/internal-plugin-support';
 import '@ciscospark/internal-plugin-team';
 import '@ciscospark/internal-plugin-user';
 import '@ciscospark/internal-plugin-wdm';
+import '@ciscospark/plugin-people';
```
ianwremmel: Do you need to put the showAllTypes config in here or can that be done in the web client?

packages/node_modules/@ciscospark/plugin-people/src/config.js
UWTech: Done.

packages/node_modules/@ciscospark/recipe-private-web-client/src/index.js
UWTech: Good point. I think I should go ahead and do it here.

UWTech: Actually, we're already configuring the logger in web-client, so to be consistent, I guess I'll go ahead and do the configuration for people plugin there as well.

UWTech:
@ianwremmel  any timeframe when this can be merged in?

We need this addition to the recipe in order to integrate Hydra. Seems like a low risk change to me; can we get this in soon?

Cheers.

UWTech:
@ianwremmel , can I assume your last comment is an approval, or should I wait for the official check box here?

Hoping to merge this in soon, as the Hydra changes in web-client depend on it, thanks.

packages/node_modules/@ciscospark/recipe-private-web-client/src/index.js
UWTech: FYI, I do have a local branch for web-client ready to go once this is merged in that I have set the config for. Verified everything gets sent as expected.

ianwremmel:
already merged

UWTech:
Merged in.

PR #584, opened 2017-05-15
First pass at room dialing

ianwremmel:

PR #698, opened 2017-09-14
Group Calling

ianwremmel:
Closes #584

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
@@ -1,0 +1,166 @@
+/*!
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import '@ciscospark/plugin-phone';
+
+import {assert} from '@ciscospark/test-helper-chai';
+import sinon from '@ciscospark/test-helper-sinon';
+import CiscoSpark from '@ciscospark/spark-core';
+import testUsers from '@ciscospark/test-helper-test-users';
+import handleErrorEvent from '../lib/handle-error-event';
+import {browserOnly, maxWaitForEvent} from '@ciscospark/test-helper-mocha';
+
+if (process.env.NODE_ENV !== `test`) {
+  throw new Error(`Cannot run the plugin-phone test suite without NODE_ENV === "test"`);
+}
+
+browserOnly(describe)(`plugin-phone`, function() {
+  this.timeout(60000);
+
+  describe(`Phone`, () => {
+    let mccoy, spock;
+    before(`create users and register`, () => testUsers.create({count: 2})
+      .then((users) => {
+        [mccoy, spock] = users;
+        spock.spark = new CiscoSpark({
+          credentials: {
+            authorization: spock.token
+          }
+        });
+
+        mccoy.spark = new CiscoSpark({
+          credentials: {
+            authorization: mccoy.token
+          }
+        });
+
+        return Promise.all([
+          spock.spark.phone.register(),
+          mccoy.spark.phone.register()
+        ]);
+      }));
+
+    let ringMccoy;
+
+    beforeEach(() => {
+      ringMccoy = sinon.spy();
+      mccoy.spark.phone.on(`call:incoming`, ringMccoy);
+    });
+
+    beforeEach(() => {
+      spock.spark.config.phone.enableExperimentalGroupCallingSupport = true;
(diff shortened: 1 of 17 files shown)


packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
```diff
@@ -0,0 +1,166 @@
...
+      ringMccoy = sinon.spy();
+      mccoy.spark.phone.on(`call:incoming`, ringMccoy);
+    });
+
+    beforeEach(() => {
+      spock.spark.config.phone.enableExperimentalGroupCallingSupport = true;
+      mccoy.spark.config.phone.enableExperimentalGroupCallingSupport = true;
+    });
+
+    afterEach(() => {
```
adamweeks: Is there a reason why we are unsetting these after each test?

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
```diff
@@ -219,7 +198,7 @@ browserOnly(describe)(`plugin-phone`, function() {
           }));
 
       // TODO [SSDK-233, SSDK-508, SSDK-56, SSDK-149, SSDK-3] need to figure out
-      // what entitlements are required to dial sip addresses. Also, might be
+      // what entitlements are required to a dial sip addresses. Also, might be
```
adamweeks: Very minor but probably meant to be `to dial a sip`

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
```diff
@@ -0,0 +1,166 @@
...
+    after(`unregister spock and mccoy`, () => Promise.all([
+      spock && spock.spark.phone.deregister()
+        .catch((reason) => console.warn(`could not disconnect spock from mercury`, reason)),
+      mccoy && mccoy.spark.phone.deregister()
+        .catch((reason) => console.warn(`could not disconnect mccoy from mercury`, reason))
+    ]));
+
+
+    describe(`#dial()`, () => {
+      it(`calls a room by hydra room id`, () => spock.spark.request({
```
adamweeks: Do we want to test a room with more than 2 members? I know it isn't technically a 1:1, but it seems like it might behave that way with calling.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
```diff
@@ -0,0 +1,166 @@
...
+                      return maxWaitForEvent(1000, `change:state`, call);
+                    })
+                    .then(() => assert.equal(call.state, `active`)),
+                  mccoy.spark.phone.when(`call:incoming`)
+                ])
+                  // eslint-disable-next-line no-unused-vars
+                  .then(([_, [mc]]) => {
+                    mccoyCall = mc;
+
+                    // assert.isFalse(call.isCall);
```
adamweeks: Commented code?

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
ianwremmel: It's good practice in case these users get used elsewhere. I've made attempts to reuse test users in separate files in the past, so just in case we try that again, we want to make sure they're in their original state.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
ianwremmel: There's another test in the next PR that does have 3 users, but since it's not explicitly a 1:1 room, it should behave like a group call for all practical purposes.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
adamweeks: Gotcha, thanks!

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/multi-party.js
ianwremmel: Will fix it in the next PR.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
ianwremmel: will fix in next PR

PR #700, opened 2017-09-20
sip & dtmf

ianwremmel:


packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -500,6 +500,12 @@ const Call = SparkPlugin.extend({
       fn() {
         return isCall(this.locus);
       }
+    },
+    supportsDtmf: {
+      deps: [`locus`],
+      fn() {
+        return this.locus && this.locus.self && this.locus.self.enableDTMF;
```
adamweeks: Easier to use lodash `get` here?

packages/node_modules/@ciscospark/internal-plugin-locus/src/locus.js
```diff
@@ -223,6 +224,20 @@ const Locus = SparkPlugin.extend({
       });
   },
 
+  sendDtmf(locus, tones) {
```
adamweeks: jsdoc?

PR #702, opened 2017-09-20
Remove Legacy SDK V2

ianwremmel:
Closes #649


ianwremmel:
Yea. I moved everything to regular dependencies since there were too many cases where it's not really clear which type it should. The dependency types have no impact on what we publish, so it seemed like it made more sense to have them in one list instead of two.

PR #703, opened 2017-09-20
feat(@ciscospark/i-p-conversation): optionally defer decryption

ianwremmel:


ianwremmel:
@saurjai3 and @aimeex3, I added you to the review to see if you think this'll be useful for the web client. it augments `Conversation#list()` to be able to return non-decrypted conversations. Each conversation will have a decrypt method bound to it that'll let you take control of when you want to start decrypting.

ianwremmel:
At the moment, you can tell if a conversation is encrypted by checking if it has a `decrypt()` property (invoked `decrypt()` will remove the method), but I'm open to other options.

PR #704, opened 2017-09-21
API Documentation Updates

### LINKED ISSUE #680: Add support for posting messages in markdown
It would be very nice if when [posting a message](https://ciscospark.github.io/spark-js-sdk/api/#messages) via the api it would support markdown.

bbender:
- Standardize wording across objects
- Add `text`/`markdown` properties to `MessageObject` (closes #680)
- Change several property types from `uuid` to `string`
- Correct several `@link` links

### DIFF
--- a/packages/node_modules/ciscospark/src/ciscospark.js
+++ b/packages/node_modules/ciscospark/src/ciscospark.js
@@ -51,7 +51,7 @@ const CiscoSpark = SparkCore.extend({
 CiscoSpark.version = PACKAGE_VERSION;
 
 /**
- * Create a new CiscoSpark instance
+ * Create a new {@link CiscoSpark} instance
  *
  * @example
  * <caption>Create a new CiscoSpark instance configured for your OAuth client</caption>
--- a/packages/node_modules/ciscospark/src/ciscospark.js
+++ b/packages/node_modules/ciscospark/src/ciscospark.js
@@ -51,7 +51,7 @@ const CiscoSpark = SparkCore.extend({
 CiscoSpark.version = PACKAGE_VERSION;
 
 /**
- * Create a new ciscospark instance
+ * Create a new CiscoSpark instance
  *
  * @example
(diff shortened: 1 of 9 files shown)


packages/node_modules/ciscospark/src/ciscospark.js
```diff
@@ -51,7 +51,7 @@ const CiscoSpark = SparkCore.extend({
 CiscoSpark.version = PACKAGE_VERSION;
 
 /**
- * Create a new ciscospark instance
+ * Create a new CiscoSpark instance
```
ianwremmel: This should stay capitalized and actually be `Create a new {@link CiscoSpark} instance`

`CiscoSpark` is the class name `ciscospark` is a potential name of the instance.

packages/node_modules/ciscospark/src/plugins/memberships.js
```diff
@@ -6,22 +6,27 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
  * @property {string} roomId - The room ID
- * @property {uuid} personId - The person ID
+ * @property {string} personId - The person ID
  * @property {email} personEmail - The email address of the person / room member
- * @property {boolean} isModerator - Indicates whether the specified person should be a room moderator.
- * @property {boolean} isMonitor - Indicates whether the specified member is a room monitor.
- * @property {isoDate} created - The date and time that this membership was created.
+ * @property {boolean} isModerator - Indicates whether the specified person should be a room moderator
+ * @property {boolean} isMonitor - Indicates whether the specified member is a room monitor
+ * @property {isoDate} created - The date and time that this membership was created
```
ianwremmel: We should typedef isoDate somewhere. and maybe come up with better capitalization :)

bbender: This has been addressed.

PR #707, opened 2017-09-25
style(all): apply indent rules

ianwremmel:
No material changes, just finally found the eslint rules that'll make indents work right. Other than `.eslintrc.yml`, every change here was made via `eslint --fix`

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -2,3 +2,10 @@ extends: "@ciscospark"
 globals:
   PACKAGE_VERSION: false
 root: true
+rules:
+  indent:
+    - error
+    - 2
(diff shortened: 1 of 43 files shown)


adamweeks:
Can we update this in the eslint package too?

ianwremmel:
Once mason gets the pipeline set up, that's the plan. I don't want to introduce conflicts with what he's doing, so i'm keeping all rule changes in local files for now.

ianwremmel:
Oh, one thing:

> every change here was made via eslint --fix

I also disabled the indent rule in the automation tests because wd is a lot more readable if you follow very different indent rules.

PR #710, opened 2017-09-25
fix(tooling): publish all packages when tooling changes

ianwremmel:


tooling/lib/version.js
```diff
@@ -108,8 +108,18 @@ exports.set = async function set(version, {all, lastLog}) {
...
-    .filter((p) => ![`docs`, `legacy`, `tooling`].includes(p));
+
+  // First, get all the packages we should update
+  let packages = Array.from(all ? await list() : await updated.updated({dependents: true}));
+  // if we used updated() and it told us there's a tooling update, then there's
+  // a chance we changed a dependency or build rule, so we need to republish
+  // everything;
+  if (!all && packages.includes(`tooling`)) {
+    all = true;
+    packages = Array.from(all ? await list() : await updated.updated({dependents: true}));
```
adamweeks: Since you're setting `all = true` do we need the conditional?

ianwremmel: nah, not really. that was just copy-and-paste. i'll fix it.

PR #714, opened 2017-10-03
feat(@ciscospark/internal-plugin-conversation): set/unset space property

aimeex3:
Add ability to set/unset ANNOUNCEMENT tag to conversations.
https://sqbu-github.cisco.com/WebExSquared/cloud-apps/wiki/Announcement-Only-Spaces-Design

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
@@ -994,6 +994,36 @@ const Conversation = SparkPlugin.extend({
   };
 });
 
+/**
+ * Sets/unsets space property for convo
+ * @param {Object} conversation
+ * @param {string} tag
+ * @param {Activity} activity
+ * @returns {Promise<Activity>}
+ */
+[
+  `setSpaceProperty`,
+  `unsetSpaceProperty`
+].forEach((fnName) => {
+  const verb = fnName.startsWith(`set`) ? `set` : `unset`;
+  Conversation.prototype[fnName] = function submitSpacePropertyActivity(conversation, tag, activity) {
+    if (!isString(tag)) {
+      return Promise.reject(new Error(`\`tag\` must be a string`));
+    }
+
+    return this._inferConversationUrl(conversation)
+      .then(() => this.prepare(activity, {
+        verb,
+        target: this.prepareConversation(conversation),
+        object: {
+          tags: [tag],
+          objectType: `spaceProperty`
+        }
+      }))
+      .then((a) => this.submit(a));
+  };
+});
+
 [
   `tag`,
   `untag`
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -574,6 +574,25 @@ const Conversation = SparkPlugin.extend({
   },
 
   /**
+   * Sets announcement mode for convo
+   * @param {Object} conversation
+   * @param {ShareActivity|Array<File>} activity
+   * @returns {Promise<Activity>}
+   */
+  setAnnouncement(conversation, activity) {
```
ianwremmel: Let's see about generalizing this a bit. I'm assuming there will be additional spaceProperties to set at some future point (though this is atleast the second verb we've added for setting tags so I have little hope will stick to this interface). 

Take a look at the bottom of conversation.js where we generate a bunch of functions because they're identical except for verb.

packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -574,6 +574,25 @@ const Conversation = SparkPlugin.extend({
   },
 
   /**
+   * Sets announcement mode for convo
+   * @param {Object} conversation
+   * @param {ShareActivity|Array<File>} activity
```
ianwremmel: Copy and paste?

packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -678,6 +697,25 @@ const Conversation = SparkPlugin.extend({
     }, activity);
   },
 
+  /**
+   * Unsets announcement mode for convo
+   * @param {Object} conversation
+   * @param {ShareActivity|Array<File>} activity
```
ianwremmel: Copy and paste?

ianwremmel:
also, commit message scope should include `@ciscospark`

packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -994,6 +994,36 @@ const Conversation = SparkPlugin.extend({
...
+[
+  `setSpaceProperty`,
+  `unsetSpaceProperty`
+].forEach((fnName) => {
+  Conversation.prototype[fnName] = function submitObjectActivity(conversation, tag, activity) {
+    if (!isString(tag)) {
+      return Promise.reject(new Error(`\`tag\` must be a string`));
+    }
+
+    const verb = fnName.startsWith(`set`) ? `set` : `unset`;
```
ianwremmel: Move this outside of `submitObjectActivity` so that it only does this check twice instead of everytime this function gets called.

PR #717, opened 2017-10-05
Update Sinon

ianwremmel:


ianwremmel:
yea. chai-as-promised will never be compatible with IE11, so I had to remove it. It's still there for wd to use, but depending on how today goes, we might be on webdriver.io by next week.

PR #716, opened 2017-10-05
The WebRTC PR to close them all

ianwremmel:
Closes #615
Closes #602
Closes #706

PR #719, opened 2017-10-07
WIP: Samples

ianwremmel:
- [x] move media engine fixes to #716
- [x] wait for modals instead of relying on arbitrary delay
- [ ] add media assertions
- [x] setup wdio for use on jenkins (sauce, firefox profile, etc) (samples don't need to tested independently; wdio can handle all of the sample packages at one shot - maybe move the samples to the top level?)
- [x] make chrome work with fake media
- [ ] ~publish bundle.js?~ 
- [x] publish samples to netlify
- [ ] README for each sample
- [ ] ~move samples folder to top level?~ (breaks test helpers)
- [ ] ensure samples get tested at the right times (given the dependency tree won't trigger sample tests'


ianwremmel:
this is spread across other PRs and effective complete

PR #720, opened 2017-10-09
feat(429 retry after handling): handle 429 retry after from idBroker

UWTech:
add regex to parse API name form idbroker url
add map to store API to retry after time
add methods to check for retry after
add tests

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/index.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/index.js
@@ -42,6 +42,7 @@ export {default as RequestLoggerInterceptor} from './interceptors/request-logger
 export {default as RequestTimingInterceptor} from './interceptors/request-timing';
 export {default as SparkTrackingIdInterceptor} from './interceptors/spark-tracking-id';
 export {default as SparkUserAgentInterceptor} from './interceptors/spark-user-agent';
+export {default as RateLimitInterceptor} from './interceptors/rate-limit';
 
 export {default as Batcher} from './lib/batcher';
 export {default as Page} from './lib/page';
--- a/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
@@ -1,0 +1,128 @@
+/*!
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {Interceptor} from '@ciscospark/http-core';
+
+// contains the system time in seconds at which the retry after associated with a 429 expires
+// mapped by the API name, e.g.: idbroker.webex.com/doStuff would be mapped as 'doStuff'
+const rateLimitExpiryTime = new WeakMap();
+
+// extracts the common identity API being called
+const idBrokerRegex = /.*idbroker.webex.com\/([^\/]+)/;
+
+/**
+ * @class
+ */
+export default class RateLimitInterceptor extends Interceptor {
+
+  /**
+   * @returns {RateLimitInterceptor}
+   */
+  static create() {
+    return new RateLimitInterceptor({spark: this});
+  }
+
+  /**
+  * constructor
+  * @param {mixed} args
+  * @returns {Exception}
+  */
+  constructor(...args) {
+    super(...args);
+    rateLimitExpiryTime.set(this, new Map());
+  }
+    /**
+   * @see {@link Interceptor#onRequest}
+   * @param {Object} options
+   * @returns {Object}
+   */
+  onRequest(options) {
+    if (this.isRateLimited(options.uri)) {
+      return Promise.reject(
+        new Error(`API rate limited ${options.uri}`)
+      );
+    }
+    return Promise.resolve(options);
+  }
+  /**
+   * @see {@link Interceptor#onResponseError}
+   * @param {Object} options
+   * @param {Error} reason
+   * @returns {Object}
+   */
+  onResponseError(options, reason) {
+    if (reason.statusCode === 429 && options.uri.includes(`idbroker`)) {
+      // set the retry after in the map
+      this.setRateLimitExpiry(options.uri, this.getRetryAfterTime(options));
+    }
+    return Promise.reject(reason);
+  }
+
+  /**
+   * @param {object} options associated with the request
+   * @returns {number} retry after time in seconds
+   */
+  getRetryAfterTime(options) {
+    const retryAfter = options.headers[`retry-after`] || null;
+    if (retryAfter === null || retryAfter <= 0) {
+      return 60;
+    }
+    else if (retryAfter > 3600) {
+      return 3600;
+    }
+    return retryAfter;
+  }
+
+  /**
+   * Sets the system time at which the rate limiting
+   * will expire in the rateLimitExpiryTime map.
+   * Assumes retryAfter is in seconds (S)
+   * @param {string} uri API issuing the rate limiting
+   * @param {number} retryAfter seconds until rate limiting expires
+   * @returns {bool} true is value was successfully set
+   */
+  setRateLimitExpiry(uri, retryAfter) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime() / 1000;
+    const expiry = currTimeSec + retryAfter;
+    const dict = rateLimitExpiryTime.get(this);
+    return apiName && dict && dict.set(apiName, expiry);
+  }
+
+  /**
+   * returns true if the API is currently rate limited
+   * @param {string} uri
+   * @returns {Boolean} indicates whether or not the API is rate currently rate limited
+   */
+  getRateLimitStatus(uri) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime() / 1000;
+    const dict = rateLimitExpiryTime.get(this);
+    return dict && dict.get(apiName) === null ? false : currTimeSec < dict.get(apiName);
+  }
+
+  /**
+   * splits the URI and returns the API name of idBroker
+   * @param {string} uri
+   * @returns {string}
+   */
+  getApiName(uri) {
+    const results = uri === null ? null : uri.match(idBrokerRegex);
+    return results === null ? null : results[1];
+  }
+
+  /**
+   * check URI against list of currently rate limited
+   * URIs, and determines if retry-after
+   * @param {String} uri pattern to check
+   * @returns {bool}
+   */
+  isRateLimited(uri) {
+    // determine if the URI is associated with a common identity API
+    if (uri && uri.includes(`idbroker`)) {
+      return this.getRateLimitStatus(uri);
+    }
+    return false;
+  }
+}
--- a/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
@@ -9,7 +9,7 @@ import {Interceptor} from '@ciscospark/http-core';
 const rateLimitExpiryTime = new WeakMap();
 
 // extracts the common identity API being called
-const idBrokerRegex = /.*idbroker.webex.com\/([^\/]+)/;
+const idBrokerRegex = /.*(idbroker|identity).webex.com\/([^\/]+)/;
 
 /**
  * @class
@@ -52,9 +52,9 @@ export default class RateLimitInterceptor extends Interceptor {
    * @returns {Object}
    */
   onResponseError(options, reason) {
-    if (reason.statusCode === 429 && options.uri.includes(`idbroker`)) {
-      // set the retry after in the map
-      this.setRateLimitExpiry(options.uri, this.getRetryAfterTime(options));
+    if (reason.statusCode === 429 && (options.uri.includes(`idbroker`) || options.uri.includes(`identity`))) {
+      // set the retry after in the map, setting to milliseconds
+      this.setRateLimitExpiry(options.uri, this.extractRetryAfterTime(options) * 1000);
     }
     return Promise.reject(reason);
   }
@@ -63,7 +63,7 @@ export default class RateLimitInterceptor extends Interceptor {
    * @param {object} options associated with the request
    * @returns {number} retry after time in seconds
    */
-  getRetryAfterTime(options) {
+  extractRetryAfterTime(options) {
     const retryAfter = options.headers[`retry-after`] || null;
     if (retryAfter === null || retryAfter <= 0) {
       return 60;
@@ -77,17 +77,17 @@ export default class RateLimitInterceptor extends Interceptor {
   /**
    * Sets the system time at which the rate limiting
    * will expire in the rateLimitExpiryTime map.
-   * Assumes retryAfter is in seconds (S)
+   * Assumes retryAfter is in milliseconds
    * @param {string} uri API issuing the rate limiting
-   * @param {number} retryAfter seconds until rate limiting expires
+   * @param {number} retryAfter milliseconds until rate limiting expires
    * @returns {bool} true is value was successfully set
    */
   setRateLimitExpiry(uri, retryAfter) {
     const apiName = this.getApiName(uri);
-    const currTimeSec = new Date().getTime() / 1000;
+    const currTimeSec = new Date().getTime();
     const expiry = currTimeSec + retryAfter;
     const dict = rateLimitExpiryTime.get(this);
-    return apiName && dict && dict.set(apiName, expiry);
+    return apiName && dict.set(apiName, expiry);
   }
 
   /**
@@ -97,9 +97,9 @@ export default class RateLimitInterceptor extends Interceptor {
    */
   getRateLimitStatus(uri) {
     const apiName = this.getApiName(uri);
-    const currTimeSec = new Date().getTime() / 1000;
+    const currTimeSec = new Date().getTime();
     const dict = rateLimitExpiryTime.get(this);
-    return dict && dict.get(apiName) === null ? false : currTimeSec < dict.get(apiName);
+    return dict.get(apiName) === null ? false : currTimeSec < dict.get(apiName);
   }
 
   /**
@@ -109,7 +109,9 @@ export default class RateLimitInterceptor extends Interceptor {
    */
   getApiName(uri) {
     const results = uri === null ? null : uri.match(idBrokerRegex);
-    return results === null ? null : results[1];
+    // group 0 = full match of URL, group 1 = identity or idbroker base
+    // group 2 = api name
+    return results === null ? null : results[2];
   }
 
   /**
@@ -120,7 +122,7 @@ export default class RateLimitInterceptor extends Interceptor {
    */
   isRateLimited(uri) {
     // determine if the URI is associated with a common identity API
-    if (uri && uri.includes(`idbroker`)) {
+    if (uri && (uri.includes(`idbroker`) || uri.includes(`identity`))) {
       return this.getRateLimitStatus(uri);
     }
     return false;
--- a/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
@@ -54,24 +54,28 @@ export default class RateLimitInterceptor extends Interceptor {
   onResponseError(options, reason) {
     if (reason.statusCode === 429 && (options.uri.includes(`idbroker`) || options.uri.includes(`identity`))) {
       // set the retry after in the map, setting to milliseconds
-      this.setRateLimitExpiry(options.uri, this.extractRetryAfterTime(options) * 1000);
+      this.setRateLimitExpiry(options.uri, this.extractRetryAfterTime(options));
     }
     return Promise.reject(reason);
   }
 
   /**
    * @param {object} options associated with the request
-   * @returns {number} retry after time in seconds
+   * @returns {number} retry after time in milliseconds
    */
   extractRetryAfterTime(options) {
+    // 1S * 1K === 1MS
+    const milliMultiplier = 1000;
     const retryAfter = options.headers[`retry-after`] || null;
+    // set 60 retry if no usable time defined
     if (retryAfter === null || retryAfter <= 0) {
-      return 60;
+      return 60 * milliMultiplier;
     }
+    // set max to 3600 S (1 hour) if greater than 1 hour
     else if (retryAfter > 3600) {
-      return 3600;
+      return 3600 * milliMultiplier;
     }
-    return retryAfter;
+    return retryAfter * milliMultiplier;
   }
 
   /**
@@ -84,10 +88,15 @@ export default class RateLimitInterceptor extends Interceptor {
    */
   setRateLimitExpiry(uri, retryAfter) {
     const apiName = this.getApiName(uri);
+
+    if (!apiName) {
+      return false;
+    }
+
     const currTimeMilli = new Date().getTime();
     const expiry = currTimeMilli + retryAfter;
     const dict = rateLimitExpiryTime.get(this);
-    return apiName && dict.set(apiName, expiry);
+    return dict.set(apiName, expiry);
   }
 
   /**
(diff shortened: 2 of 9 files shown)


UWTech:
@ianwremmel  here are the basics so far.

Cheers.

UWTech:
Created interceptor and UTs

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+      this.setRateLimitExpiry(options.uri, this.getRetryAfterTime(options));
+    }
+    return Promise.reject(reason);
+  }
+
+  /**
+   * @param {object} options associated with the request
+   * @returns {number} retry after time in seconds
+   */
+  getRetryAfterTime(options) {
```
ianwremmel: let's name this something that doesn't start with get, since it's not a getter for a value related to the interceptor's state

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+   * Sets the system time at which the rate limiting
+   * will expire in the rateLimitExpiryTime map.
+   * Assumes retryAfter is in seconds (S)
+   * @param {string} uri API issuing the rate limiting
+   * @param {number} retryAfter seconds until rate limiting expires
+   * @returns {bool} true is value was successfully set
+   */
+  setRateLimitExpiry(uri, retryAfter) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime() / 1000;
```
ianwremmel: Date.now() should be more efficient; note that dividing by 1000 turns the int into a float, so you probably want to wrap the result with `Math.floor`

ianwremmel: (alternatively, it may be easier to work in milliseconds everywhere and just convert the retry-after header from seconds)

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * returns true if the API is currently rate limited
+   * @param {string} uri
+   * @returns {Boolean} indicates whether or not the API is rate currently rate limited
+   */
+  getRateLimitStatus(uri) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime() / 1000;
+    const dict = rateLimitExpiryTime.get(this);
```
ianwremmel: take a look at the TemplateContainer construct (usage of it will probably be more helpful than looking at its definition). it takes care of some of the existence checking for you.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+    return dict && dict.get(apiName) === null ? false : currTimeSec < dict.get(apiName);
```
ianwremmel: This is a bit complicated for a ternary

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * splits the URI and returns the API name of idBroker
+   * @param {string} uri
+   * @returns {string}
+   */
+  getApiName(uri) {
+    const results = uri === null ? null : uri.match(idBrokerRegex);
```
ianwremmel: This is a bit complicated for a ternary

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * check URI against list of currently rate limited
+   * URIs, and determines if retry-after
+   * @param {String} uri pattern to check
+   * @returns {bool}
+   */
+  isRateLimited(uri) {
+    // determine if the URI is associated with a common identity API
+    if (uri && uri.includes(`idbroker`)) {
```
ianwremmel: note that some common identity apis are also hosted on identity.webex.com

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,219 @@
...
+          function sleep(ms) {
+            return new Promise((resolve) => setTimeout(resolve, ms));
+          }
+
+          async function sleepForExpiry() {
+            await sleep(2000);
+          }
+
+          interceptor.setRateLimitExpiry(`https://idbroker.webex.com/horse/v1/myID`, 1);
+          sleepForExpiry().then(() => {
```
ianwremmel: - promises need to be returned for mocha to wait for them
- .then statements should always be on the following line and indented one level from the above statement (unless the above statement is a `.then` or a `.catch`.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,219 @@
...
+        it(`returns true when API is rate limited`, () => {
+          interceptor.setRateLimitExpiry(`https://idbroker.webex.com/horse/v1/myID`, new Date().getTime() * 2);
+          assert.equal(interceptor.isRateLimited(`https://idbroker.webex.com/horse/v1/myID`), true);
+        });
+        it(`returns false when previously rate limited API expiry time has been met`, () => {
+          function sleep(ms) {
+            return new Promise((resolve) => setTimeout(resolve, ms));
+          }
+
+          async function sleepForExpiry() {
```
ianwremmel: you've got a few different sleep definitions in this file, how about consolidating them at the top as one function.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,219 @@
...
+            await sleep(2000);
```
ianwremmel: Awesome as async/await is, we're keeping it out of source and test code for now.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {string} uri API issuing the rate limiting
+   * @param {number} retryAfter seconds until rate limiting expires
+   * @returns {bool} true is value was successfully set
+   */
+  setRateLimitExpiry(uri, retryAfter) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime() / 1000;
+    const expiry = currTimeSec + retryAfter;
+    const dict = rateLimitExpiryTime.get(this);
+    return apiName && dict && dict.set(apiName, expiry);
```
ianwremmel: Doesn't checking of `dict` is valid hide a real error?

ianwremmel: I think you'll also want to use safeSetTimeout to setup cache eviction, right? Otherwise, especially as this expands to other services, you'll end up with a memory leak.

UWTech: How would I end up with a memory leak?

These are all keyed by API name, so any new entries would overwrite the old entry.

For example, dostuff/1234 and dostuff/4567 map to the same key "dostuff". I don't see why we need cache eviction for an ephemeral structure that is gone after the app shuts down. 

I can remove the check for dict, I added it because you requested it be put in in the examples provided.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: I'm not sure I follow, what is complicated here?

If the value is not in the dict (is null), the result is false, the API is not limited. Else, if the current system time is less than the expiry time, true, we are rate limited. If the currsystem time is not less than the expiry, meaning it is greater than the expiry, the answer is false.

If we didn't have to check for a null value, there wouldn't even need to be a ternary, it would just be return currTimeSec < dict.get(apiName);

I wanted to make sure I wasn't making any assumptions about how null was handled, so rather than doing a direct comparison against a potentially null value, I'm just using a ternary to return false.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: This is the use case for ternary. I don't follow what is complicated here.

It has been verified with unit tests.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: Thanks, I'll add that.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: I could skip this ternary and return null, but the second ternary is definitely necessary.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
UWTech: Done.

Not sure how I ended up with the replication and two different methods.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
UWTech: Done.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: I did remove the check for a null dict per your earlier comment.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: Done

UWTech:
Addressed code review feedback

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,130 @@
...
+   * Sets the system time at which the rate limiting
+   * will expire in the rateLimitExpiryTime map.
+   * Assumes retryAfter is in milliseconds
+   * @param {string} uri API issuing the rate limiting
+   * @param {number} retryAfter milliseconds until rate limiting expires
+   * @returns {bool} true is value was successfully set
+   */
+  setRateLimitExpiry(uri, retryAfter) {
+    const apiName = this.getApiName(uri);
+    const currTimeSec = new Date().getTime();
```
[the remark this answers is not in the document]
UWTech: The change was made, but the variable name was not updated to milli ( I changed the method docs to reflect this, but didn't update the variable name.)

I can get that real quick.

packages/node_modules/@ciscospark/spark-core/src/index.js
```diff
@@ -42,6 +42,7 @@ export {default as RequestLoggerInterceptor} from './interceptors/request-logger
 export {default as RequestTimingInterceptor} from './interceptors/request-timing';
 export {default as SparkTrackingIdInterceptor} from './interceptors/spark-tracking-id';
 export {default as SparkUserAgentInterceptor} from './interceptors/spark-user-agent';
+export {default as RateLimitInterceptor} from './interceptors/rate-limit';
```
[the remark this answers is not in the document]
UWTech: I'll take a look, thanks.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: There is code that prevents Keying nulls, so shouldn't be a concern.

The code being more compact with ternary is more readable imo, and since they're functionally equivalent,  is it really such a big deal to have ternary?

We're talking about the difference between grepping 7 or 8 lines of code if we use if statements versus 2 lines using ternary operators.

This strikes me as a coding style thing rather than a concern about functional correctness. Is there a style rule saying no ternary operators?

I could add a check in getRateLimitStatus that short circuits and returns if API name is null, but that would be pretty clunky and would obfuscate the method's function imo.  So at line 100 I could add:
if(apiName === null)
 return false;

However, since the expectation is the apiName returned is never null due to the fact we check the Regex against the URI before calling get rate limit status, you are now adding an extra null check on every single positive hit to any idBroker/identity calls. We would expect that apiName is never null, so optimizing with a short circuit for the unexpected case where it is null would be *less* efficient overall.

packages/node_modules/@ciscospark/spark-core/src/index.js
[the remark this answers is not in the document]
UWTech: Done.

UWTech:
@ianwremmel  any further concerns here?

Getting pinged asking when this is getting merged in.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
bzang: Just to chime in here.

I think there is definitely some code style preference here, but also from an architecture standpoint there are some best practices we can use. So, sure there are checks in place higher up the stack that'll short circuit certain behavior, but I think if we're trying to ensure code isolation, then it's important to have checks in place for odd behaviors in case the code calling these Classes get funky. For example, in this implementation, I would think you'd want to have checks in `onResponseError` and `onRequest` for if `options` and `options.uri` are even defined.

I don't personally have an issue with ternaries, and I think if we actually put together a style guide, we'd explicitly encourage them where it makes sense, but compactness doesn't necessarily mean more understandable, and here it definitely took me some extra time to figure out what the ternary is doing.

UWTech: There is no check for options === null in any of the onError interceptor methods.

Provided there is error checking prior to inputting a value into the method at some stage, it should be fine. The most appropriate place to put the logic that does input validation is prior to adding to whatever data structure you want to use to store.

I don't see a legitimate use case for someone accessing the map outside of the accessors provided in the module. Any additional interactions needed should be added after examining the current assumptions and expectations with respect to the map. If people are adding methods, and messing with the map wily-nilly, no amount of preparation is going to prevent problems.

Additionally the module is reasonably small, so it shouldn't be hard for someone to grok the code  and make the appropriate changes.

The discussion around ternary operators is not one of compactness versus excessive catch statements per se; given that the work has been done, and the code is functionally correct, and we agree it is a style preference, I don't think it makes sense to write it again.

I could do an informal survey here at my site, and we'd probably find that most people prefer ternary operators, perhaps at your site, you don't find ternary as easy to understand as lengthy if-then statements and explicit null checks; I don't think one or another is necessarily right or wrong so much as it isn't necessary to reconstitute the solution.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * returns true if the API is currently rate limited
+   * @param {string} uri
+   * @returns {Boolean} indicates whether or not the API is rate currently rate limited
+   */
+  getRateLimitStatus(uri) {
+    const apiName = this.getApiName(uri);
+    const currTimeMilli = new Date().getTime();
+    const dict = rateLimitExpiryTime.get(this);
+    return dict.get(apiName) === null ? false : currTimeMilli < dict.get(apiName);
```
ianwremmel: when would `dict.get(apiName) === null` be true? WeakMaps return `undefined` if the specified key can't be found.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
ianwremmel: > given that the work has been done, and the code is functionally correct, and we agree it is a style preference

We don't agree it's *just* "a style preference". Two of the maintainers of this project have raised concerns about readability. We're not saying "don't use ternaries", but we are concerned that these particular ternaries are difficult to grok.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,130 @@
...
+   * @param {string} uri API issuing the rate limiting
+   * @param {number} retryAfter milliseconds until rate limiting expires
+   * @returns {bool} true is value was successfully set
+   */
+  setRateLimitExpiry(uri, retryAfter) {
+    const apiName = this.getApiName(uri);
+    const currTimeMilli = new Date().getTime();
+    const expiry = currTimeMilli + retryAfter;
+    const dict = rateLimitExpiryTime.get(this);
+    return apiName && dict.set(apiName, expiry);
```
ianwremmel: Wouldn't the `apiName` check make more sense around line 87 to avoid potentially doing needless work if `apiName` is falsy?

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * @see {@link Interceptor#onResponseError}
+   * @param {Object} options
+   * @param {Error} reason
+   * @returns {Object}
+   */
+  onResponseError(options, reason) {
+    if (reason.statusCode === 429 && (options.uri.includes(`idbroker`) || options.uri.includes(`identity`))) {
+      // set the retry after in the map, setting to milliseconds
+      this.setRateLimitExpiry(options.uri, this.extractRetryAfterTime(options) * 1000);
```
ianwremmel: The `*1000` might make more sense inside `extractRetryAfterTime`; that way, all functions' inputs and outputs are in the same unit.

packages/node_modules/@ciscospark/spark-core/src/index.js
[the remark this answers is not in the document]
ianwremmel: I still don't see a change that adds the rate limit interceptor to the list of active interceptors.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,291 @@
...
+          interceptor.setRateLimitExpiry(`https://idbroker.webex.com/horse/v1/myID`, new Date().getTime() * 2);
+          assert.equal(interceptor.isRateLimited(`https://idbroker.webex.com/horse/v1/myID`), true);
+        });
+        it(`returns true when identity API is rate limited`, () => {
+          interceptor.setRateLimitExpiry(`https://identity.webex.com/horse/v1/myID`, new Date().getTime() * 2);
+          assert.equal(interceptor.isRateLimited(`https://identity.webex.com/horse/v1/myID`), true);
+        });
+        it(`returns false when previously rate limited idbroker API expiry time has been met`, () => {
+          interceptor.setRateLimitExpiry(`https://idbroker.webex.com/horse/v1/myID`, 1);
+          return sleep(2000)
```
ianwremmel: Take a look at our use of `lolyx` for mock timers so that tests can be written synchronously (and run significantly faster).

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,291 @@
...
+
+      describe(`#getApiName`, () => {
+        it(`returns null when there is no regex match`, () => {
+          assert.equal(interceptor.getApiName(`https://fakeuir.com/testFake`), null);
+        });
+        it(`returns null when the argument is null`, () => {
+          assert.equal(interceptor.getApiName(null), null);
+        });
+        it(`returns idbroker API name if there is a match`, () => {
+          assert.equal(interceptor.getApiName(`https://idbroker.webex.com/horse/v1/myID`), `horse`);
```
ianwremmel: Is the goal to stop using an entire hostname when we get back a 429 or to stop using the particular endpoint that sent the 429?

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,291 @@
...
+        });
+        it(`returns identity API name if there is a match`, () => {
+          assert.equal(interceptor.getApiName(`https://identity.webex.com/horse/v1/myID`), `horse`);
+        });
+      });
+
+      describe(`#getRateLimitStatus`, () => {
+        it(`returns false if API name is not in rate limit expiry map`, () => {
+          assert.equal(interceptor.getRateLimitStatus(`https://fakeuir.com/testFake`), false);
```
ianwremmel: Let's use `example.com` for all domains we don't control. Yes, these are unit tests, but we're using a lot of real domains for testing throughout the product that have definitely caused wierd issues.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
```diff
@@ -0,0 +1,291 @@
...
+            headers: {'retry-after': 7200}}), 3600);
+        });
+        it(`returns retry-after as is if 0 < retry-after < 3600`, () => {
+          assert.equal(interceptor.extractRetryAfterTime({
+            headers: {'retry-after': 55}}), 55);
+        });
+      });
+
+      describe(`#onRequest`, () => {
+        it(`Rejects idbroker request if API is rate limited`, () => {
```
ianwremmel: See question above about rate-limiting a service vs an endpoint

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: Ternary operators can't get much simpler than this. From my perspective you are essentially dictating style.

I'll go ahead and change these, but if you don't want ternary operators being used, you need to have an official rule, and not ask me to change functionally correct code that is already done.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
UWTech: As discussed in earlier commit messages, the specific API associated with a hostname is limited. The two host names that we've been asked to honor 429s from are the ID broker endpoints.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
UWTech: Answered earlier.

packages/node_modules/@ciscospark/spark-core/src/index.js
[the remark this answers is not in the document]
UWTech: Right I hadn't pushed the commit since there was still disagreement on the use of ternary operators. I have added it to spark-js-sdk/packages/node_modules/@ciscospark/spark-core/src/spark-core.js, so should show up when I push.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: Changed.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: Ah, gotcha. I'll change that to undefined, thanks Ian.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/interceptors/rate-limit.js
ianwremmel: gotcha. sorry, that still seems weird to me, but if that's the plan of record, good enough.

packages/node_modules/@ciscospark/spark-core/src/interceptors/rate-limit.js
UWTech: I don't really like the idea of the method mutating the form of the data which is used as input, but I don't have strong objections. Consistency is important as well, so I can make that change.

UWTech:
@Ian got rid of the ternary operators; you're making an old embedded systems developer sad, but I can avoid their use in the future if it is going to be so controversial. I honestly find them much more readable, and when they are not nested to an insane degree, I prefer them.

Updated to use lolex, moved some stuff around, and I did add the interceptor in the core.

UWTech:
@ianwremmel Tagged the wrong Ian on last comment....

got rid of the ternary operators; you're making an old embedded systems developer sad, but I can avoid their use in the future if it is going to be so controversial. I honestly find them much more readable, and when they are not nested to an insane degree, I prefer them.

Updated to use lolex, moved some stuff around, and I did add the interceptor in the core.

Cheers.

ianwremmel:
already merged

UWTech:
Merged. Sorry about skipping cleanup here.

PR #724, opened 2017-10-12
feat(@ciscospark/i-p-encryption): get auth list of the resource

huichecisco:
need to get authorization list of the resource, so we can authorize user to the resource if the user is not authorized, also only de-authorize the user that is authorized.

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
@@ -153,7 +153,7 @@ const KMS = SparkPlugin.extend({
    * @param {string} options.kroUri
    * @returns {Array<authId>}
    */
-  getAuthorizationList({kro, kroUri}) {
+  listAuthorizations({kro, kroUri}) {
     kroUri = kroUri || kro.uri;
     /* istanbul ignore if */
     if (!kroUri) {
@@ -165,12 +165,7 @@ const KMS = SparkPlugin.extend({
     })
       .then((res) => {
         this.logger.info(`kms: retrieved authorization list`);
-        const authIds = [];
-        const authorizations = res.authorizations;
-        authorizations.forEach((authorization) => {
-          authIds.push(authorization.authId);
-        });
-        return authIds;
+        return res.authorizations;
       });
   },
 
--- a/packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
@@ -147,6 +147,34 @@ const KMS = SparkPlugin.extend({
   },
 
   /**
+   * Retrieve a list of users that have been authorized to the KRO
+   * @param {Object} options
+   * @param {KMSResourceObject} options.kro the target kro
+   * @param {string} options.kroUri
+   * @returns {Array<authId>}
+   */
+  getAuthorizationList({kro, kroUri}) {
+    kroUri = kroUri || kro.uri;
+    /* istanbul ignore if */
+    if (!kroUri) {
+      return Promise.reject(new Error(`\`kro\` or \`kroUri\` is required`));
+    }
+    return this.request({
+      method: `retrieve`,
+      uri: `${kroUri}/authorizations`
+    })
+      .then((res) => {
+        this.logger.info(`kms: retrieved authorization list`);
+        const authIds = [];
+        const authorizations = res.authorizations;
+        authorizations.forEach((authorization) => {
+          authIds.push(authorization.authId);
+        });
+        return authIds;
+      });
+  },
+
+  /**
    * Deauthorizes a user or KRO from a KRO
    * @param {Object} options
    * @param {string} options.userId
--- a/packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
@@ -91,6 +91,89 @@ describe(`Encryption`, function() {
         }));
     });
 
+    /**
+     *Test getAuthorizationList function
+     *Setup: Create a resource, then authorize a user to a key, then authorize a resource to a key
+     *Test:
+     *1)Invoke getAuthorizationList on the resource, verify that the function returns the array that contains two authorized items
+     *2)Remove the authorized user from the resource, then invoke getAuthorizationList again, verify that the returned array doesn't have the user.
+     *3)Remove the authorized resource from the resource again, then invoke getAuthorizationList again, verify that he returned array doesn't have the resource.
+     */
+    describe(`#getAuthorizationList()`, () => {
+      let boundedKeyUri, kro, otherKro;
+      const testUserId = mccoy.spark.internal.device.userId;
+      let testResourceId;
+
+      before(`creates a resource`, () => spark.internal.encryption.kms.createUnboundKeys({count: 1})
+        .then(([key]) => spark.internal.encryption.kms.createResource({
+          key
+        }))
+        .then((k) => {
+          kro = k;
+          boundedKeyUri = kro.keyUris[0];
+          assert.lengthOf(kro.authorizationUris, 1);
+        }));
+
+      before(`authorizes a user to a key`, () => spark.internal.encryption.kms.addAuthorization({
+        userIds: [testUserId],
+        kroUri: kro.uri
+      })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, testUserId);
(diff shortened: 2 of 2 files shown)


packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,89 @@ describe(`Encryption`, function() {
...
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, otherKro.uri);
+          return spark.internal.encryption.kms.getAuthorizationList({kroUri: kro.uri})
+            .then((authIds) => {
+              assert.equal(authIds.length, 2);
+              assert.notEqual(authIds.indexOf(testUserId), -1);
+              assert.notEqual(authIds.indexOf(testResourceId), -1);
+            });
+        }));
+
```
ianwremmel: remove extra blank line

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,89 @@ describe(`Encryption`, function() {
...
+      it(`remove the user and verify this user is not in the authorization list `, () => spark.internal.encryption.kms.removeAuthorization({
+        userId: testUserId,
+        kroUri: kro.uri
+      })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, testUserId);
+          return spark.internal.encryption.kms.getAuthorizationList({kroUri: kro.uri})
+            .then((authIds) => {
+              assert.equal(authIds.indexOf(testUserId), -1);
```
ianwremmel: use `assert.include`

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,89 @@ describe(`Encryption`, function() {
...
+            });
+        }));
+
+
+      it(`remove the resource and verify this resource is not in the authorizaiton list`, () => spark.internal.encryption.kms.removeAuthorization({
```
ianwremmel: Isn't this the same test as above? Is one of these supposed to assert the user is in the list and another that they are not?

packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
```diff
@@ -147,6 +147,34 @@ const KMS = SparkPlugin.extend({
   },
 
   /**
+   * Retrieve a list of users that have been authorized to the KRO
+   * @param {Object} options
+   * @param {KMSResourceObject} options.kro the target kro
+   * @param {string} options.kroUri
+   * @returns {Array<authId>}
+   */
+  getAuthorizationList({kro, kroUri}) {
```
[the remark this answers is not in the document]
huichecisco: updated name to listAuthorizations

packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
```diff
@@ -147,6 +147,34 @@ const KMS = SparkPlugin.extend({
...
+    }
+    return this.request({
+      method: `retrieve`,
+      uri: `${kroUri}/authorizations`
+    })
+      .then((res) => {
+        this.logger.info(`kms: retrieved authorization list`);
+        const authIds = [];
+        const authorizations = res.authorizations;
+        authorizations.forEach((authorization) => {
```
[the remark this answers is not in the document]
huichecisco: Yes, agree.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: updated the code and removed the extra blank line

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: updated the test code accordingly due to the code change. Also used the assert.include to check the member of array.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: One removes user from authorization list, another one  removes the resource from the resource from the authorization list.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,88 @@ describe(`Encryption`, function() {
...
+          testResourceId = otherKro.uri;
+          return spark.internal.encryption.kms.addAuthorization({
+            authIds: [testResourceId],
+            kro
+          });
+        })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, otherKro.uri);
+          return spark.internal.encryption.kms.listAuthorizations({kroUri: kro.uri})
```
ianwremmel: this block should probably be an `it()` test instead of part of the the test setup

packages/node_modules/@ciscospark/internal-plugin-encryption/src/kms.js
```diff
@@ -147,6 +147,34 @@ const KMS = SparkPlugin.extend({
...
+    if (!kroUri) {
+      return Promise.reject(new Error(`\`kro\` or \`kroUri\` is required`));
+    }
+    return this.request({
+      method: `retrieve`,
+      uri: `${kroUri}/authorizations`
+    })
+      .then((res) => {
+        this.logger.info(`kms: retrieved authorization list`);
+        const authIds = [];
```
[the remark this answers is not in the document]
huichecisco: updated code to return res.authorizations.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,88 @@ describe(`Encryption`, function() {
...
+            .then((authorizations) => {
+              assert.equal(authorizations.length, 2);
+              assert.include(authorizations, {uri: auth.uri, authId: testUserId, resourceUri: auth.resourceUri});
```
ianwremmel: assert.include uses strict equality, so this will never hold (since the object defined in parameter 2 isn't in `authorizations`).  If we were on the most recent version of chai, you'd be able to use `assert.deepInclude()`, but since we're not there yet, try something like `assert.include(authorizations.map((a) => a.authId), testUserId`

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -91,6 +91,88 @@ describe(`Encryption`, function() {
...
+      it(`remove the resource and verify this resource is not in the authorizaiton list`, () => spark.internal.encryption.kms.removeAuthorization({
+        userId: testResourceId,
+        kroUri: kro.uri
+      })
+        .then(([auth]) => {
+          assert.equal(auth.resourceUri, kro.uri);
+          assert.equal(auth.authId, testResourceId);
+          return spark.internal.encryption.kms.listAuthorizations({kroUri: kro.uri})
+            .then((authorizations) => {
+              assert.notInclude(authorizations, {uri: auth.uri, authId: testResourceId, resourceUri: auth.resourceUri});
```
ianwremmel: Similar to the statement about deepInclude above, you'll want to map authorizations down to primitives and just compare to testResourceId

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
ianwremmel: got it. took me a while spot the subtle difference.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: Yes, i moved  this part as a separate test case instead of as the part of test suite setup. 
Initially i am concerned about the order of the tests. It seems that the order of execution for test cases are the order that test cases are laid out.

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: updated accordingly

packages/node_modules/@ciscospark/internal-plugin-encryption/test/integration/spec/kms.js
huichecisco: updated accordingly

PR #729, opened 2017-10-19
feat(@ciscospark/i-p-board): allow delete channel from a spark space

tran2:
Mainly remove the link between a whiteboard and a spark space.

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
@@ -135,18 +135,9 @@ const Board = SparkPlugin.extend({
    * @returns {Promise}
    */
   deleteChannel(conversation, channel) {
-    return this.spark.request({
-      method: `PUT`,
-      uri: `${channel.aclUrl}/links`,
-      body: this._prepareDeleteChannel(conversation, channel)
-    })
-      .then((res) => res.body);
-  },
-
-  _prepareDeleteChannel(conversation, channel) {
     // remove the ACL link between conversation and board
     // remove conversation auth from board KRO in kms message
-    return {
+    const body = {
       aclLinkType: `INCOMING`,
       linkedAcl: conversation.aclUrl,
       kmsMessage: {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
```diff
@@ -128,6 +128,37 @@ const Board = SparkPlugin.extend({
...
+  deleteChannel(conversation, channel) {
+    return this.spark.request({
+      method: `PUT`,
+      uri: `${channel.aclUrl}/links`,
+      body: this._prepareDeleteChannel(conversation, channel)
+    })
+      .then((res) => res.body);
+  },
+
+  _prepareDeleteChannel(conversation, channel) {
```
ianwremmel: Is this something likely to be used anywhere other than deleteChannel? if not, not sure it needs to be its own method.

packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
```diff
@@ -128,6 +128,37 @@ const Board = SparkPlugin.extend({
...
+    // remove the ACL link between conversation and board
+    // remove conversation auth from board KRO in kms message
+    return {
+      aclLinkType: `INCOMING`,
+      linkedAcl: conversation.aclUrl,
+      kmsMessage: {
+        method: `delete`,
+        uri: `${channel.kmsResourceUrl}/authorizations?${querystring.stringify({authId: conversation.kmsResourceObjectUrl})}`
```
ianwremmel: note that request accepts a `qs` property and takes care of stringification if you give it a hash automatically

tran2: this uri is for the kmsMessage though, would request take care of that?

ianwremmel: yep, keep 
```js
uri: `${channel.kmsResourceUrl}/authorizations`
```
then just put the querystring params in the `qs` hash.

PR #730, opened 2017-10-20
fix(@ciscospark/common): remove console logging

adamweeks:
`logger` should be console if nothing else is defined.

### DIFF
--- a/packages/node_modules/@ciscospark/common/src/one-flight.js
+++ b/packages/node_modules/@ciscospark/common/src/one-flight.js
@@ -58,10 +58,11 @@ export default function oneFlight(...params) {
         if (this && this.logger) {
           this.logger.info(message);
         }
-        else {
-          /* eslint no-console: [0] */
+        else if (process.env.NODE_ENV !== `production`) {
+          // eslint-disable-next-line no-console
           console.info(message);
         }
+
         return flight;
       }
 


packages/node_modules/@ciscospark/common/src/one-flight.js
```diff
@@ -58,10 +58,6 @@ export default function oneFlight(...params) {
         if (this && this.logger) {
           this.logger.info(message);
         }
-        else {
```
ianwremmel: Let's do this with check on NODE_ENV and/or CISCOSPARK_LOG_LEVEL rather than removing it entirely. It is *useful*, if *annoying* :)

adamweeks: I don't see the pattern of the logger being used in conjunction with NODE_ENV anywhere else. Shouldn't we just rely on the logger config to determine if it should actually log?

ianwremmel: The problem seems to be that this thing isn't finding `this.logger` and is therefor falling back to console. So, we need to either (a) figure out why this isn't picking up `this.logger` or (b) throw a little bit of logic into the fallback for when it can't find `this.logger`.

adamweeks: Is this the only place were `this.logger` may not be found? Doesn't seem like we check it any where else we use it, but of course this is in a decorator

adamweeks:
Changed to check NODE_ENV

packages/node_modules/@ciscospark/common/src/one-flight.js
ianwremmel: yea, i'm pretty sure it is.

adamweeks:
@ianwremmel could i `#ignore-tooling` on this?

ianwremmel:
it wouldn't help; you're changing the common package, on which all other packages depend. `#ignore-tooling` just means "ignore changes to the tooling dir and select other files for determining what to test"

adamweeks:
Closed as to being handled in another PR

PR #733, opened 2017-10-25
Fix presence mercury event tests

adamweeks:


packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -20,6 +25,11 @@ describe(`plugin-presence`, function() {
         spock.spark = new CiscoSpark({
           credentials: {
             authorization: users[0].token
+          },
+          config: {
+            logger: {
+              level: `error`
```
ianwremmel: ?

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -64,6 +74,8 @@ describe(`plugin-presence`, function() {
     });
 
     describe(`#subscribe`, () => {
+      afterEach(() => spock.spark.internal.presence.unsubscribe([mccoy.id, spock.id]));
```
ianwremmel: `spock && spock.spark` makes explaining jenkins output a lot easier when test user creation fails

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -81,6 +93,32 @@ describe(`plugin-presence`, function() {
             assert.equal(presenceResponse.responses[1].subject, spock.id);
           })
       );
+
+      it(`should receive a mercury event for a subscribed person's change`, () => {
+        const presenceUpdateSpy = sinon.spy();
+        spock.spark.internal.mercury.on(`event:apheleia.subscription_update`, presenceUpdateSpy);
```
ianwremmel: Is the spy doing anything that waiting for the event doesn't?

ianwremmel: oh, i see, this is where you're getting the value of the subscription event. `eventEvent()` should give you the event as well, so that might be an easier place to grab the property from.

packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
adamweeks: I'm not sure I follow this.

adamweeks:
@ianwremmel ready for round 2

PR #736, opened 2017-10-31
WIP: Samples 2

ianwremmel:
This is like #719, but is not based off of #716 (and therefore not blocked by its mergeability (or lack thereof).

> Note: this code is under active development and is likely to be rebased a few times. Please do not fork off of this branch at this time.
- [ ] Samples
  - [x] authentication
  - [x] ~single party calling (deprecated)~
  - [x] single party calling (with group calling enabled)
    - [ ] demonstrate how to split tracks in chrome
  - [ ] multi party calling
- [ ] Tooling
  - [x] publish samples via netlify
  - [x] rename all sample-related npm scripts to start with (instead of end with) "samples:"


.nvmrc
```diff
@@ -0,0 +1 @@
+v7.10.1
```
bbender: Are we now recommending at least Node 7.10?

package.json
```diff
@@ -77,6 +77,13 @@
     "get-next-version": "./tooling/index.js version next --always",
     "deps:check": "./tooling/index.js exec -- dependency-check ./package.json",
     "deps:generate": "./deps.js",
+    "build:samples": "echo 'Deprecated. Please use npm run samples:build'; exit 1",
+    "serve:samples": "echo 'Deprecated. Please use npm run samples:serve'; exit 1",
+    "test:samples": "echo 'Deprecated. Please use npm run samples:test'; exit 1",
```
bbender: Can these three deprecated commands be removed?

webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+  output: {
+    filename: `bundle.js`,
+    library: `ciscospark`,
+    libraryTarget: `var`,
+    path: __dirname,
+    sourceMapFilename: `[file].map`
+  },
+  devtool: `source-map`,
+  devServer: {
+    https: true,
```
bbender: Do we need to use HTTPS for localhost?

.nvmrc
ianwremmel: nah, this was to get netlify to use a version that would work with nodegit; it didn't actually help (i removed nodegit), so i might be able to remove this file. For the moment, saying we support active LTS versions of node should be adequate, though we're not currently setup to test against that many node versions in a reasonable amount of time.

package.json
ianwremmel: I may just make them interchangeable

webpack.config.js
ianwremmel: i've found it to be a little hit-or-miss, but in general, yes. various browsers really don't like to let you get camera access if you're not on https.

ianwremmel: That said, it's mostly just for camera access; i don't think anything else needs it.

packages/node_modules/samples/browser-auth-implicit/test/wdio/spec/implicit-grant-authentication.js
```diff
@@ -0,0 +1,49 @@
...
+      assert.equal(browserSpock.getText(`#authentication-status`), `Not Authenticated`);
+    });
+
+    step(`initiates implicit grant authentication flow`, () => {
+      browserSpock.click(`[title="authenticate"]`);
+    });
+
+    step(`enters user login information`, () => {
+      // Enter Email
+      browserSpock.waitForExist(`input[type=email]`);
```
adamweeks: I've switched to using `.waitUntil` because it allows you to provide a failure message. Otherwise the logs just say "promise rejected". Not a blocker, but something to keep in mind while working with webdriver.io

packages/node_modules/samples/browser-single-party-call/index.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+  <title>Sample: Single Party Calling</title>
+  <body>
+    <h1>Single Party Calling</h1>
+    <p>This sample demonstrates 1:1 calling.</p>
+    <p>In case your computer doesn't have a camera attached, you can load this
```
adamweeks: Should we instruct how to do the fake media on chrome?

packages/node_modules/samples/browser-single-party-call/test/wdio/spec/dial-and-reject.js
```diff
@@ -0,0 +1,50 @@
...
+      }));
+
+    step(`loads the app`, () => {
+      browser.url(`/packages/node_modules/samples/browser-single-party-call?usefakemedia`);
+    });
+
+    step(`connects mccoy's browser`, () => {
+      browserMccoy.waitUntil(() => browserMccoy.getTitle() === `Sample: Single Party Calling`);
+      browserMccoy.execute(setValue, `[placeholder="Your access token"]`, mccoy.token.access_token);
+      browserMccoy.click(`[title="connect"]`);
```
adamweeks: We've been creating test helper elements objects that contain all these locator strings in the widgets. So you would import it then use `elements.connectButton`. Almost like a page-objects lite.

packages/node_modules/samples/browser-single-party-call/index.html
ianwremmel: i'm actually leaning toward removing it entirely. it's lead to more confusion than i expected and i really only put it in there because i was being lazy while working from my mac mini.

packages/node_modules/samples/browser-single-party-call/test/wdio/spec/dial-and-reject.js
ianwremmel: historically, i've been against that strategy because it mostly just gives you even more files to look at to understand a test. that said, i accept that it makes sense when you're writing a large number of tests to interact with a small number of elements. Given the narrow scope of these tests, do you think it adds much of any thing?

wdio.conf.js
```diff
@@ -0,0 +1,366 @@
...
+  // },
+  //
+  // Test runner services
+  // Services take over a specific job you don't want to take care of. They enhance
+  // your test setup with almost no effort. Unlike plugins, they don't add new
+  // commands. Instead, they hook themselves up into the test process.
+  services: CI ? [
+    `sauce`,
+    `static-server`,
+    `webpack`
```
adamweeks: Ha, the last time I tried this, they didn't even support webpack 2 configs. Good to see that they are working now.

packages/node_modules/samples/browser-single-party-call/test/wdio/spec/dial-and-reject.js
adamweeks: I only brought it up because I saw the selectors repeated across the different test scenarios. It doesn't really make sense for a one time use element selector. I do think these tests would benefit from it by reducing the amount of "magic strings" throughout.

ianwremmel: fair enough. I'd say, let's leave it as-is for now, and maybe once we've got a few more samples built, we can standardize them all a bit better than they're looking right now and then add those helpers.

adamweeks: Sure thing

index.html
```diff
@@ -3,12 +3,12 @@
     <head></head>
     <body>
         <script type="text/javascript">
-            
+
             console.log("location: ", window.location);
 
-            window.location.href = "http://localhost:8000/packages/node_modules/sample-browser-auth-implicit/" + window.location.hash;
-        
+            window.location.href = "http://localhost:8000/packages/node_modules/samples/browser-auth-implicit/" + window.location.hash;
```
adamweeks: Should we get the domain/port from window location and use it instead of localhost?

packages/node_modules/samples/browser-auth-implicit/test/wdio/spec/implicit-grant-authentication.js
```diff
@@ -23,15 +25,33 @@ describe(`samples/browser-auth-implicit`, () => {
...
+
+        browserSpock.setValue(`#IDToken2`, spock.password);
+        browserSpock.click(`#Button1`);
+
+        if (!browserSpock.isVisible(`.generic-error`)) {
+          break;
+        }
+      }
+
+      // I feel like there's a better way to do this...
```
adamweeks: Could you check to see if you got redirected after a successful login? if the `.generic-error` class is on the redirected page, it could cause a false negative

packages/node_modules/samples/browser-single-party-call/test/wdio/spec/dial-and-reject.js
```diff
@@ -14,6 +14,10 @@ describe(`samples/browser-single-party-call`, () => {
         mccoy = users[1];
       }));
 
+    // the inner-most function is executed in the window environment
+    // eslint-disable-next-line no-undef
+    before(() => browser.execute(() => window.reload));
```
adamweeks: Why not just `browser.reload()`?

packages/node_modules/samples/browser-auth-implicit/test/wdio/spec/implicit-grant-authentication.js
ianwremmel: if '.generic-error' is on the redirected page, we want it to fail.

ianwremmel:
hmm, ok, yea. the redirect uri and generated index.html don't work for the auth sample.

packages/node_modules/samples/browser-single-party-call/test/wdio/spec/dial-and-reject.js
ianwremmel: interesting...thought i looked for that. I'll make the change.

PR #738, opened 2017-11-07
feat(@ciscospark/i-p-board): prevent deletion of active whiteboard

tran2:
- add lockChannelForDeletion to lock channel before deleting or to tell
  whether a channel cannot be locked since it is in use
- move encryption related unit tests to unit/spec/encryption

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
@@ -173,7 +173,10 @@ const Board = SparkPlugin.extend({
   lockChannelForDeletion(channel) {
     return this.spark.request({
       method: `POST`,
-      uri: `${channel.channelUrl}/lock?intent=delete`
+      uri: `${channel.channelUrl}/lock`,
+      qs: {
+        intent: `delete`
+      }
     })
       .then((res) => res.body);
   },
--- a/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
@@ -132,9 +132,11 @@ const Board = SparkPlugin.extend({
    * @memberof Board.BoardService
    * @param  {Conversation~ConversationObject} conversation
    * @param  {Board~Channel} channel
+   * @param  {Object} options
+   * @param  {Object} options.preventDeleteActiveChannel Returns error if channel is in use
    * @returns {Promise}
    */
-  deleteChannel(conversation, channel) {
+  deleteChannel(conversation, channel, options = {}) {
     // remove the ACL link between conversation and board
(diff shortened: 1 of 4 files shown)


packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
```diff
@@ -147,10 +149,31 @@ const Board = SparkPlugin.extend({
...
+   * @param  {Board~Channel} channel
+   * @returns {Promise}
+   */
+  lockChannelForDeletion(channel) {
     return this.spark.request({
-      method: `PUT`,
-      uri: `${channel.aclUrl}/links`,
-      body
+      method: `POST`,
+      uri: `${channel.channelUrl}/lock?intent=delete`
```
ianwremmel: is this a case where qs can actually be used? (unlike my confusion last time)

packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
```diff
@@ -147,10 +149,31 @@ const Board = SparkPlugin.extend({
       aclLinkOperation: `DELETE`
     };
 
+    let promise = Promise.resolve();
+    if (options.preventDeleteActiveChannel) {
+      promise = this.lockChannelForDeletion(channel);
```
ianwremmel: Why does this need to be done on the client? Why doesn't the service send a 409 if the channel is in use?

packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
ianwremmel: does the channel have a links array? is there a lock link that can dropped directly into the url instead of synthesizing it?

tran2: Channel doesn't have links array. Let me put in a request for that?

ianwremmel: nah, it's fine, i just saw that you stopped using  `${channel.aclUrl}/links`, so i wanted to make sure we didn't have something like that already.

packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
tran2: The current delete (aka remove channel from a space) is done by the client sending `remove ACL Links` to KMS directly. The board service does not have control over KMS so this `/lock?intent=delete` from the board service is a helper for clients.

packages/node_modules/@ciscospark/internal-plugin-board/src/board.js
tran2: oh it's just weird diff issues... I didn't remove that part of the code. It belonged in a different function

PR #739, opened 2017-11-08
feat(@ciscospark/spark-core): add error support for empty params

adamweeks:
Batcher #handleHttpError could receive an object or string that isn't a SparkHttpError

Hopefully this resolves https://voxeolabs.atlassian.net/browse/SSDK-1475

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
@@ -2,6 +2,7 @@
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
 
+import {has} from 'lodash';
 import SparkPlugin from './spark-plugin';
 import {
   cappedDebounce,
@@ -167,11 +168,16 @@ const Batcher = SparkPlugin.extend({
    * @returns {Promise<undefined>}
    */
   handleHttpError(reason) {
-    const msg = reason.message || reason.body || reason;
-    return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
-      .then((defer) => {
-        defer.reject(msg);
-      })));
+    if (reason instanceof SparkHttpError) {
+      if (has(reason, `options.body.map`)) {
+        return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
+          .then((defer) => {
+            defer.reject(reason);
+          })));
+      }
+    }
+    this.logger.error(`http error handler called without a SparkHttpError object`, reason);
+    return Promise.reject(reason);
   },
 
   /**
--- a/packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
+++ b/packages/node_modules/@ciscospark/spark-core/test/unit/spec/lib/batcher.js
@@ -295,5 +295,11 @@ describe(`spark-core`, () => {
         });
       });
     });
+
+    describe(`#handleHttpError()`, () => {
+      it(`handles a non SparkHttpError object passed`, () => {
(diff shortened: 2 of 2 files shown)


packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
```diff
@@ -166,12 +166,15 @@ const Batcher = SparkPlugin.extend({
...
    * @returns {Promise<undefined>}
    */
-  handleHttpError(reason) {
+  handleHttpError(reason = `An undefined http error has occurred.`) {
     const msg = reason.message || reason.body || reason;
-    return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
-      .then((defer) => {
-        defer.reject(msg);
-      })));
+    if (reason && reason.options && reason.options.body && reason.options.body.map) {
```
ianwremmel: use lodash.has?

ianwremmel: I think we might want to log something indicating that we hit the http error handler without an http error; that *should* imply that the error happened before the request was ever submitted.

adamweeks:
The batcher that was reported with this error was the metrics batcher which uses this one:
```return Reflect.apply(Batcher.prototype.handleHttpError, this, [reason]);```

File: packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js [hunk 2]
```diff
@@ -167,11 +168,17 @@ const Batcher = SparkPlugin.extend({
...
+      const msg = reason.message || reason.body || reason;
+      if (has(reason, `options.body.map`)) {
+        return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
+          .then((defer) => {
+            defer.reject(msg);
+          })));
+      }
+    }
+    this.logger.error(`http error handler called without a SparkHttpError object`);
+    return Promise.reject(`An undefined http error has occurred.`);
```
ianwremmel: this should be wrapped in `new Error()`

ianwremmel: I think there's a lint rule for that, but we don't have it enabled.

packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
```diff
@@ -167,11 +168,16 @@ const Batcher = SparkPlugin.extend({
...
-    const msg = reason.message || reason.body || reason;
-    return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
-      .then((defer) => {
-        defer.reject(msg);
-      })));
+    if (reason instanceof SparkHttpError) {
+      if (has(reason, `options.body.map`)) {
+        return Promise.all(reason.options.body.map((item) => this.getDeferredForRequest(item)
+          .then((defer) => {
+            defer.reject(reason);
```
adamweeks: @ianwremmel ready for one more round. I changed this to reject with the http error as we discuss as well.

packages/node_modules/@ciscospark/spark-core/src/lib/batcher.js
```diff
@@ -167,11 +168,16 @@ const Batcher = SparkPlugin.extend({
...
+          })));
+      }
+    }
+    this.logger.error(`http error handler called without a SparkHttpError object`);
+    return Promise.reject(new Error(`An undefined http error has occurred.`));
```
ianwremmel: ok, saw one more issue; if reason is not a `SparkHttpError`, it may still carry useful information. we should certainly log it and can probably just reject with it directly.

ianwremmel:
This is merged

PR #740, opened 2017-11-08
Encryption fixes

ianwremmel:


packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -129,7 +129,12 @@ const User = SparkPlugin.extend({
       resource: `users`
     })
       .then((res) => res.body)
-      .then(tap((user) => this.recordUUID(user)));
+      .then(tap((user) => this.recordUUID({
+        id: user.id,
+        // CI endpoints don't use the same user format as actors, so, email may
+        // be in one of a few fields
+        emailAddress: user.email || user.emailAddress
```
adamweeks: we're having to massing conversation/user fields in the widget too...

packages/node_modules/@ciscospark/test-helper-mocha/src/index.js
```diff
@@ -87,6 +91,15 @@ module.exports = {
     return inFirefox() ? mochaMethod.skip : mochaMethod;
   },
 
+  skipInIE: function skipInIE(mochaMethod) {
```
adamweeks: Nice!

packages/node_modules/@ciscospark/plugin-logger/src/logger.js
```diff
@@ -128,6 +129,20 @@ const Logger = SparkPlugin.extend({
   },
 
   /**
+   * Determines if the current level allows logs at the speicified level to be
```
adamweeks: `speicified` not fixed here

packages/node_modules/@ciscospark/internal-plugin-encryption/README.md
```diff
@@ -1,3 +1,81 @@
...
+  .then(() => {
+    spark.internal.encryption.kms.createUnboundKeys({count: 1})
+      .then(([key]) => {
+        return spark.internal.encryption.kms.createResource({
+          userIds: [],
+          key
+        })
+      })
+      .then((kro) => {
+
```
adamweeks: Maybe some comments as to what the return values for all of these are?

ianwremmel: ah crap. i thought that docs thing was done when I pulled it in here. apparently I never finished it...

I think I'll be pulling that out entirely for now.

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
ianwremmel: ?

adamweeks: wow, total keyboard fail on that. I don't even know what I was trying to say other than we are doing something similar on widgets.

PR #742, opened 2017-11-11
feat(@ciscospark/i-p-wdm): support Spark High Availability Design

tran2:
- feature toggle `web-ha-messaging`
- when we get a new catalog, we parse service and serviceHostMap and
  encapsulate into serviceCatalog collection
- getWebSocketUrl should now replace webSocketUrl due to its ability
  to take datacenter availability into account
- add new method to mark current service url as fail and get a new one,
  it cycles through all the host and when all hosts fail, it requests
  new service registration info from wdm

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
@@ -113,10 +113,16 @@ const Mercury = SparkPlugin.extend({
 
   _prepareUrl(webSocketUrl) {
     if (!webSocketUrl) {
-      webSocketUrl = this.spark.internal.device.currentWebSocketUrl;
+      webSocketUrl = this.spark.internal.device.webSocketUrl;
     }
 
-    return this.spark.internal.feature.getFeature(`developer`, `web-shared-mercury`)
+    return this.spark.internal.feature.getFeature(`developer`, `web-ha-messaging`)
+      .then((haMessagingEnabled) => {
+        if (haMessagingEnabled) {
+          webSocketUrl = this.spark.internal.device.currentWebSocketUrl;
+        }
+      })
+      .then(() => this.spark.internal.feature.getFeature(`developer`, `web-shared-mercury`))
       .then((webSharedMercury) => {
         webSocketUrl = url.parse(webSocketUrl, true);
         Object.assign(webSocketUrl.query, {
@@ -196,8 +202,14 @@ const Mercury = SparkPlugin.extend({
           return callback(reason);
         }
         else if (reason instanceof ConnectionError) {
-          this.logger.info(`mercury: received a generic connection error, will try to connect to another datacenter`);
-          return this.spark.internal.device.markUrlFailedAndGetNew(`mercuryConnection`)
+          return this.spark.internal.feature.getFeature(`developer`, `web-ha-messaging`)
+            .then((haMessagingEnabled) => {
+              if (haMessagingEnabled) {
+                this.logger.info(`mercury: received a generic connection error, will try to connect to another datacenter`);
+                return this.spark.internal.device.markUrlFailedAndGetNew(`mercuryConnection`);
+              }
+              return null;
+            })
             .then(() => callback(reason));
         }
 
--- a/packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
@@ -119,8 +119,12 @@ const Mercury = SparkPlugin.extend({
     return this.spark.internal.feature.getFeature(`developer`, `web-ha-messaging`)
       .then((haMessagingEnabled) => {
         if (haMessagingEnabled) {
-          webSocketUrl = this.spark.internal.device.currentWebSocketUrl;
+          return this.spark.internal.device.getWebSocketUrl();
         }
+        return webSocketUrl;
+      })
+      .then((wsUrl) => {
+        webSocketUrl = wsUrl;
       })
       .then(() => this.spark.internal.feature.getFeature(`developer`, `web-shared-mercury`))
       .then((webSharedMercury) => {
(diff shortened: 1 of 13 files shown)


packages/node_modules/@ciscospark/internal-plugin-mercury/src/mercury.js
```diff
@@ -193,6 +195,12 @@ const Mercury = SparkPlugin.extend({
           this.backoffCall.abort();
           return callback(reason);
         }
+        else if (reason instanceof ConnectionError) {
+          this.logger.info(`mercury: received a generic connection error, will try to connect to another datacenter`);
+          return this.spark.internal.device.markUrlFailedAndGetNew(`mercuryConnection`)
```
ianwremmel: maybe instead of marking a service failed, we should mark a url as failed and let the code in wdm take care of figuring out the service.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
```diff
@@ -282,6 +351,27 @@ const Device = SparkPlugin.extend({
     }
   },
 
+  _updateServiceCatalog(newRegistration) {
+    Object.keys(newRegistration.services).forEach((service) => {
+      const uri = newRegistration.services[service];
+      const u = Url.parse(uri);
+      const hosts = newRegistration.serviceHostMap.hostCatalog[u.host];
+      const existingService = this.serviceCatalog.get(service);
+      if (existingService) {
```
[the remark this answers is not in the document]
tran2: Oh, thanks! I overlook the fact that you can add remove, add, merge as an option to set. Looks like it does what I want by adding the option `{remove: false}` since the I wanted to add/merge here.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
```diff
@@ -99,6 +123,36 @@ const Device = SparkPlugin.extend({
     return Promise.resolve(this._getServiceUrl(this.config.preDiscoveryServices, service));
   },
 
+  markUrlFailedAndGetNew(service) {
```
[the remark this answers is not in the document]
tran2: yes, let me do that

tran2:
@ianwremmel my bad, right now it should be behind this feature toggle `web-ha-messaging`

packages/node_modules/@ciscospark/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -453,6 +476,15 @@ describe(`plugin-mercury`, () => {
           assert.notMatch(wsUrl, /multipleConnections/);
         }));
 
+      describe(`when web-ha-messasing is enabled`, () => {
```
ianwremmel: messaging is spelled with a g :)

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
```diff
@@ -7,7 +7,9 @@ import {safeSetTimeout} from '@ciscospark/common-timers';
 import {omit} from 'lodash';
 import util from 'util';
 import FeaturesModel from './features-model';
+import ServiceCollection from './service-collection';
 import {persist, waitForValue, SparkPlugin} from '@ciscospark/spark-core';
+import * as Url from 'url';
```
ianwremmel: no need for `*` -> `import Url from 'url';`

ianwremmel: also, `url` is typically lowercase

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/service-model.js
```diff
@@ -0,0 +1,156 @@
+/*!
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import AmpState from 'ampersand-state';
+import {defaults, isObject, some, find} from 'lodash';
+import url from 'url';
+
+const ServiceModel = AmpState.extend({
```
ianwremmel: Please add jsdoc blocks to class and methods.

packages/node_modules/@ciscospark/internal-plugin-wdm/test/unit/spec/device.js
```diff
@@ -85,6 +101,133 @@ describe(`plugin-wdm`, () => {
           assert.becomes(device.getServiceUrl(`valid`), `http://example.com/valid-service`)
         ]);
       });
+
+      describe(`when web-ha-messasing is enabled`, () => {
+        it(`returns the service url with a different host if available`, () => {
```
ianwremmel: What happens when we've run out of hosts? What happens if there's only one host?

packages/node_modules/@ciscospark/internal-plugin-mercury/test/unit/spec/mercury.js
tran2: oops.. I blame auto complete

packages/node_modules/@ciscospark/internal-plugin-wdm/test/unit/spec/device.js
tran2: When we've ran out of hosts, it's going to call `this.refresh()`, this logic and the logic to cycle host is in `markUrlFailedAndGetNew`

Here, for `getServiceUrl`, it'll just use the current host

ianwremmel: gotcha. thanks!

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
tran2: There are some use of `url` in the code so I had to change the name of the package to `Url` to prevent conflicts.

PR #746, opened 2017-11-14
feat(@ciscospark/i-p-lyra): allow sending proximity token

tran2:
Help support audio-pairing

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-lyra/src/lyra.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-lyra/src/lyra.js
@@ -11,8 +11,19 @@ const Lyra = SparkPlugin.extend({
   children: {
     space: Space,
     device: Device
-  }
+  },
 
+  sendUltraSoundToken(token) {
+    return this.spark.request({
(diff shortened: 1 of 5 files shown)


packages/node_modules/@ciscospark/internal-plugin-lyra/src/lyra.js
```diff
@@ -11,8 +11,19 @@ const Lyra = SparkPlugin.extend({
   children: {
     space: Space,
     device: Device
-  }
+  },
 
+  sendUltraSoundToken(token) {
```
ianwremmel: Please add doc block

ianwremmel: What's in the response body? I wonder if we can give the method a name gives a better idea why we're calling it instead of just what it's doing.

tran2: I renamed to `getAdvertisedEndpoint` so it makes more sense hopefully :)
Also added @typedef so we know how the body looks like

PR #753, opened 2017-11-17
WIP: Publish Auth Sample

ianwremmel:


packages/node_modules/@ciscospark/spark-core/src/spark-core.js
```diff
@@ -382,17 +382,21 @@ const SparkCore = AmpState.extend({
    * `onBeforeLogout` handlers
    * @returns {Promise}
    */
-  logout(...args) {
+  logout(options, ...rest) {
```
bzang: this makes me want to use `kwargs` in JS

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -291,36 +291,40 @@ const Credentials = SparkPlugin.extend({
...
+
     const tokens = this.userTokens.models;
-    const supertoken = this.supertoken;
 
-    this.unset(`supertoken`);
+    try {
+      this.unset(`supertoken`);
+    }
+    catch (err) {
+      this.logger.warn(`credentials: failed to cleaer supertoken`, err);
```
bzang: typo `cleaer`

packages/node_modules/samples/browser-auth-implicit/app.js
```diff
@@ -4,8 +4,29 @@
...
+}
+
+// Inititate the SDK
+// IN A PRODUCTION APP YOU SHOULD NOT HARDCODE THESE VALUES BUT INSTEAD LOAD
+// THEM FROM THE ENVIRONMENT AT BUILD TIME. SECRETS AND OTHER CREDENTIALS SHOULD
+// NOT BE COMMITTED TO THE CODEBASE.
+const spark = ciscospark.init({
+  config: {
+    credentials: {
+      client_id: `C7c3f1143a552d88d40b2afff87600c366c830850231597fb6c1c1e28a5110a4f`,
```
bzang: Given the warning, does it make sense to hard code it here, or provide some instructions on how to add it into envvars?

ianwremmel: I thought about that. The problem is that none of the sample apps are compiled, they're loaded as-is. My goal was to have as little friction as possible to interpret the sample code. 

I guess an alternative might be to put a form at the top of the page for them to add their app credentials, but then the sample would get a whole lot bigger.

PR #754, opened 2017-11-17
feat(@ciscospark/spark-core): refresh token at 60-90% of expiration

aimeex3:


packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -341,6 +371,10 @@ const Credentials = SparkPlugin.extend({
     const supertoken = this.supertoken;
     const tokens = this.userTokens.models;
 
+    // clear refresh timer
+    if (this.refreshTimer) {
+      clearTimeout(this.refreshTimer);
```
ianwremmel: I think you also want to unset the refresh time when you clear it, right? Otherwise `this.refreshTimer` would be truthy at unexpected times.

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -143,6 +159,15 @@ const Credentials = SparkPlugin.extend({
     Reflect.apply(SparkPlugin, this, args);
   },
 
+  downscope(scope) {
```
ianwremmel: This should probably be _underscore prefixed; `getToken(scope)` is how a consumer should cause a downsope

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -267,6 +287,12 @@ const Credentials = SparkPlugin.extend({
           this.supertoken = attrs.authorization;
         }
       }
+
+      // schedule refresh
+      if (this.supertoken) {
+        const timeoutLength = this._calcRefreshTimeout(this.supertoken.expires_in);
```
ianwremmel: There's an edge case this doesn't cover. when you load a token from storage, expires_in doesn't mean anything any more (since it's a duration from the time the token was received).

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
ianwremmel: no, don't do that. we're about the outlaw underscore prefixing :)

PR #757, opened 2017-11-22
Multi-party Samples

ianwremmel:
- [x] sample tests don't have a dependency list, so they don't get tested
- [x] ~sample tests seem to fail~ the single party `dial and reject` test leaves the browsers in a bad state
- [x] ~sample tests seem to fail~ one of the multi-party tests seems to leave the browsers in a bad state
- [ ] sample tests that don't produce junit don't seem to stop builds


packages/node_modules/samples/browser-multi-party-call/app.js
```diff
@@ -0,0 +1,152 @@
...
+  });
+
+  call.once(`remoteMediaStream:change`, () => {
+    document.getElementById(`remote-view`).srcObject = call.remoteMediaStream;
+  });
+
+  // Once the call ends, we'll want to clean up our UI a bit
+  call.on(`disconnected`, () => {
+    // Remove the streams from the UI elements
+    document.getElementById(`self-view`).srcObject = document.getElementById(`remote-view`).srcObject = undefined;
```
adamweeks: Not a big fan of the multi assign.

packages/node_modules/samples/browser-multi-party-call/test/wdio/spec/dial-before-connect.js
```diff
@@ -0,0 +1,114 @@
...
+
+      browserSpock.waitForExist(`.listening`);
+    });
+
+    step(`connects mccoy's browser`, () => {
+      browserMccoy.waitUntil(() => browserMccoy.getTitle() === `Sample: Multi Party Calling`);
+      browserMccoy.execute(setValue, `[placeholder="Your access token"]`, mccoy.token.access_token);
+      browserMccoy.click(`[title="connect"]`);
+
+      // TODO how do we wait for a modal to open?
```
adamweeks: You could wait for [#alertText](http://webdriver.io/api/protocol/alertText.html) to return a value, but doesn't seem to do much more than what you are doing now.

packages/node_modules/samples/browser-multi-party-call/test/wdio/spec/normal-dialing.js
```diff
@@ -0,0 +1,120 @@
...
+
+    step(`loads the app`, () => {
+      browser.reload();
+      browser.url(`/browser-multi-party-call/`);
+    });
+
+    step(`connects mccoy's browser`, () => {
+      browserMccoy.waitUntil(() => browserMccoy.getTitle() === `Sample: Multi Party Calling`);
+      browserMccoy.execute(setValue, `[placeholder="Your access token"]`, mccoy.token.access_token);
+      // Debouncing before connecting ruduces flakiness
```
adamweeks: `reduces`

PR #760, opened 2017-11-22
Change Test Helpers to es6 syntax

ianwremmel:


packages/node_modules/@ciscospark/test-helper-appid/src/create-user.browser.js
```diff
@@ -2,25 +2,20 @@
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
 
-'use strict';
 
-require('whatwg-fetch');
-var makeLocalUrl = require('@ciscospark/test-helper-make-local-url');
+require(`whatwg-fetch`);
```
adamweeks: were we going to switch over to back ticks? I thought we were sticking with single quotes except for those with template literals in them

ianwremmel: we are, but those rules weren't ready when i made this PR. I'm adding the new eslint stuff onto this PR, so I'll be requesting a second review in a little while.

ianwremmel:
this is done

PR #761, opened 2017-11-22
Improve XUnit Reports

ianwremmel:
We've got three different sources of xunit reports: mocha, karma, and wdio. Each produces a subtly different format that Jenkins parses differently. This PR is an attempt to have all tests appear in the Jenkins UI as something like "Package Name > Describe Blocks... > It Block > Browser Name or NodeJS".

### DIFF
--- a/karma-ng.conf.js
+++ b/karma-ng.conf.js
@@ -159,18 +159,28 @@ function makeConfig(packageName, argv) {
 
   if (argv && argv.xunit) {
     cfg.junitReporter = {
+      // className defaults to the first describe block
       classNameFormatter(browser) {
         let name = argv.package;
-
-        name = `${name}.${browser.name
-          .replace(/ /g, `_`)
-          .replace(/\./g, `_`)}`;
+        name = name;
+        // name = `${name}.${browser.name
+        //   .replace(/ /g, `_`)
+        //   .replace(/\./g, `_`)}`;
 
         return name;
       },
+
       outputFile: `${packageName}.xml`,
       outputDir: `reports/junit/karma`,
+
+      // Specifying 'suite' fills in the testuite.package attribute. Jenkins
+      // combines testsuite.name with testsuite.package
       suite: packageName,
+
+      // useBrowserName=true doesn't seem to effect the output format, but does
+      // put reports in folders by browser. In other words, this needs to be set
+      // in order for karma to produce reports for more than just the last
+      // browser to complete.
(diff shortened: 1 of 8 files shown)


ianwremmel:
Remaining problems:
- [ ] karma tests stick the browser name after the class name and before the test name. ideally, we want the browser name all the way at the end
- [ ] nodejs tests should stick "node" in the same place that karma puts the browser name
- [ ] something caused all the logger tests to screw up because the count on console.* calls changed
- [ ] error descriptions are over-escaped

See https://sqbu-jenkins.cisco.com:8443/job/team/job/web-sdk/job/spark-js-sdk--validated-merge--pipeline2/1826/ for examples

ianwremmel:
I'm marking this as reference because to solve the jenkins issue, we really need to customize the karma junit reporter's output.

The karma-junit-reporter always puts the browser name in `testsuite.name`. This behavior is not currently customizable. Jenkins concatenates `testuite.name` with `testuite.package`, so simply setting `testsuite.package` doesn't really help either. 

Ideally, I think we want `testsuite.name === process.env.PACKAGE`, `testsuite.package === undefined`, `testcase.className === <first describe block>`, and `testcase.name=test'

ianwremmel:
```#!/bin/bash

set -e

rm -rf ./reports

npm run test -- --package @ciscospark/common --xunit
npm run test -- --package @ciscospark/plugin-logger --xunit

# find ./reports -name '*documentation*' | xargs rm -f

OUT=""
for FILE in $(find ./reports/junit -name '*.xml'); do
  OUTFILE_NAME="$(echo "${FILE}" | sed -s 's#/#_#g' | tr -d '()' | sed -s 's#\._#_#g')"
  B64=$(base64 < "${FILE}")

  OUT+="sh \"echo '"
  OUT+="${B64}"
  OUT+="' | base64 -d > "
  OUT+="${OUTFILE_NAME}"
  OUT+='"'
  OUT+="
"


# OUT+="    sh 'mkdir -p $(dirname ${FILE}) && echo \"$(cat ${FILE} | base64)\" | base64 -d > \"${FILE}\"'"
# OUT+="
# "


# echo ''
# echo "${FILE}"
# echo ''
# cat "${FILE}" | base64
# echo ''
# echo ''

done
# echo "${OUT}"
echo "${OUT}" | pbcopy

# sh 'echo "ABC" > "FILE"'

PR #762, opened 2017-11-27
fix(@ciscospark/internal-plugin-metrics): add timestamp to metrics

aimeex3:


packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -31,7 +31,8 @@ describe(`plugin-metrics`, () => {
         testTag: `tag value`
       },
       metricName: eventName,
-      type: `behavioral`
+      type: `behavioral`,
+      timestamp: Date.now()
```
ianwremmel: This could be flaky if the mock and the src code happen to call Date.now() on a transition from one millisecond to the next. You should be able to use lolex to set a particular "now" so that the test is deterministic.

aimeex3:
only timestamp was requested to get added to the client metrics https://wiki.cisco.com/display/WX2/Influx+-+Operational+Metrics

packages/node_modules/@ciscospark/internal-plugin-metrics/test/unit/spec/metrics.js
```diff
@@ -93,9 +119,30 @@ describe(`plugin-metrics`, () => {
...
       describe(`before login`, () => {
         it(`posts pre-login metric`, () => {
-          metrics.submitClientMetrics(eventName, mockPayload, preLoginId);
-          assert.calledWith(metrics.postPreLoginMetric, preLoginProps, preLoginId);
-          assert.calledOnce(spark.credentials.getClientToken);
+          const date = Date.now();
+          const promise = metrics.submitClientMetrics(eventName, mockPayload, preLoginId);
+          return promiseTick(50)
+            .then(() => clock.tick(config.metrics.batcherWait))
+            .then(() => assert.isFulfilled(promise))
```
ianwremmel: don't use isFulfilled. Tran is working on removing chai-as-promised because the latest version will never be compatible with Internet Explorer. just returning the promise should work.

PR #764, opened 2017-11-27
build(npm): disable package-lock.json

ianwremmel:
Greenkeeper does not play as nicely with package-lock.json as we would like, so this disables it for the timebeing

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


ianwremmel:
yea, I know greenkeeper lockfile exists, but the way it works requires putting secrets in places I really don't want to. I also had mixed success getting it working in other projects. We can revisit at some point, but this'll do for now.

PR #771, opened 2017-11-30
Stabilize call tests

ianwremmel:
It looks like there are several issues
- Some call tests are flaky because of a timing problem at the end of the test (we sometimes call hangup after the call has hung itself up, thus producing a promise that will never resolve).
- The boundary between enableExperimentalGroupCalling be false or true is not explicit enough and tests rely on deprecated fields that may no longer be updated as expected.

In addition to the new set of tests, this PR will
- [x] Add additional mute/unmute event tests
- [x] Induce some harsher breakages when enableExperimentalGroupCalling is true
  - Can we avoid a semver major bump since it is "Experimental"?
- [x] Remove some of the existing tests as they'll be replaced by the new Event Model tests.


packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/events-model.js
```diff
@@ -0,0 +1,659 @@
...
+              })
+          ]);
+
+          assert.equal(mc.me.state, `notified`);
+          assert.equal(findMembership(mc, spock).state, `connected`, `mccoy expects spock to be be connected when mccoy receives the call`);
+
+          await Promise.all([
+            expectMembershipDeclinedEvent(sc, mccoy.id, `spock expects mccoy to decline the call`),
+            expectMembershipDeclinedEvent(mc, mccoy.id, `mccoy expects mccoy to decline the call`),
+            // expectDeclinedEvent(mc, `mccoy expects to decline`),
```
adamweeks: Commented line needed?

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/events-model.js
```diff
@@ -0,0 +1,659 @@
...
+          assert.equal(mc.state, `active`, `mccoy expects the call to remain active after mccoy leaves`);
+
+          await Promise.all([
+            expectMembershipDisconnectedEvent(sc, spock.id, `spock expects spock to disconnect`),
+            expectInactiveEvent(sc, `spock expects the call to go inactive after leaving the call`),
+            sc.hangup()
+          ]);
+
+          assert.equal(sc.me.state, `disconnected`, `spock expects to be disconnected after spock leaves`);
+          // assert.equal(findMembership(mc, spock).state, `disconnected`, `mccoy expects to be disconnected after spock leaves`);
```
adamweeks: More commented assertions

packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -903,15 +901,29 @@ const Call = SparkPlugin.extend({
...
-    this.on(`change:state`, () => {
-      if (this.config.enableExperimentalGroupCallingSupport) {
+      this.on(`change:state`, () => {
         process.nextTick(() => this.trigger(this.state));
-      }
-    });
+      });
+      this.on = wrap(this.on, (fn, eventName, ...rest) => {
+        if (deprecatedCallEventNames.includes(eventName)) {
+          throw new Error(`The "${eventName}" event is no longer valid with "enableExperimentalGroupCallingSupport===true"`);
```
adamweeks: May need a doc note that these events are **hard** deprecated now.

ianwremmel: Good call

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/events-model.js
ianwremmel: Yea, I'll add some comments on these. These are things we should be asserting, but can't at this time because we shutdown the event listeners when you disconnect. Before or as part of 2.0, I want to change the call event model so that calls can be rejoined (and, therefore, we won't turn the listeners off until it's appropriate). Unfortunately, that depends on understanding our memory leak issue first.

PR #782, opened 2017-12-08
feat(@ciscospark/i-p-wdm): switch server when current server fails

tran2:
add server error interceptor to switch server url when a REST request
fails

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-wdm/src/interceptors/server-error.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-wdm/src/interceptors/server-error.js
@@ -1,0 +1,51 @@
+/*!
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {Interceptor} from '@ciscospark/http-core';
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/internal-plugin-wdm/src/interceptors/server-error.js
```diff
@@ -0,0 +1,51 @@
...
+              success: false
+            },
+            tags: {
+              action: `failed`,
+              error: reason.message,
+              url: options.uri
+            }
+          });
+
+          return this.spark.internal.device.markUrlFailedAndGetNew(options.uri)
```
ianwremmel: Should this request be replayed instead of failed?

tran2: Do you mean retry this same request with the new server?

ianwremmel: yea

good-joe:
This PR should be closed, replaced by a nearly identical PR from jofuhrma

PR #783, opened 2017-12-08
feat(@ciscospark/media-engine-webrtc): add serialize function

adamweeks:


packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -686,6 +686,15 @@ export default class WebRTCMediaEngine {
   triggerNegotiationNeeded() {
     this.trigger(`negotiationneeded`);
   }
+
+  /**
+   * Returns a string when attempting to serialize object
+   * (Applies mostly this is a property of AmpState object)
```
ianwremmel: wording?

packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -686,6 +686,15 @@ export default class WebRTCMediaEngine {
...
+   * @returns {String}
```
ianwremmel: lowercase

packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
adamweeks: Deep. I re-edited that three times. Guess it wasn’t a charm

adamweeks: Derp

PR #785, opened 2017-12-12
fix(@ciscospark/plugin-phone): do not synthesize idsfor non-people

ianwremmel:


packages/node_modules/@ciscospark/plugin-phone/src/state-parsers.js
```diff
@@ -115,11 +115,18 @@ export function participantsToCallMemberships(spark, locus) {
  * @returns {CallMembership}
  */
 export function participantToCallMembership(spark, locus, participant) {
+  let personId, personUuid;
+  if (!participant.person.isExternal) {
```
adamweeks: What does `isExternal` represent in this case?

PR #792, opened 2017-12-14
WIP: better code coverage

ianwremmel:


ianwremmel:
So, I've got this working with a few caveats:
- I can't get it to instrument `./src` files *and pass*. It looks like the `istanbul` transform breaks the `builtin-extend` transform. 
- This only collects coverage from the browser tests (currently we're only collecting node test coverage)
- `nyc report` expects all raw files to be in a single directory (not nested) so the current file structure that we use for doing differential testing and copying coverage from previous builds won't quite work.

PR #794, opened 2017-12-15
WebRTC Stats Stream

ianwremmel:
This started out as "make a sample demonstrating mute and unmute", but ended up turning into "provide the hooks necessary to prove audio and video are flowing without a human being involved in the test".

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/src/call.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/call.js
@@ -1522,7 +1522,6 @@ const Call = SparkPlugin.extend({
    * @returns {StatsStream}
    */
   getRawStatsStream() {
-    // TODO stop polling after hangup
     return new StatsStream(this.media.pc);
   },
 
--- a/packages/node_modules/@ciscospark/plugin-phone/src/call.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/call.js
@@ -4,6 +4,9 @@
 
 /* eslint-env browser: true */
 
+import {debounce, defaults, find, get, wrap} from 'lodash';
+import {parse} from 'sdp-transform';
+import uuid from 'uuid';
 import {SparkPlugin} from '@ciscospark/spark-core';
 import {
   base64,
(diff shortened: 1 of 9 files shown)


packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -1513,6 +1514,25 @@ const Call = SparkPlugin.extend({
...
     });
+  },
+
+  /**
+   * Returns a {@link Readable} that emits {@link Call#media.pc}'s
+   * {@link RTCStatsReport} every second.
+   * @returns {StatsStream}
+   */
+  getRawStatsStream() {
+    // TODO stop polling after hangup
```
adamweeks: This TODO will be in another PR?

packages/node_modules/samples/browser-single-party-call-with-mute/test/wdio/spec/normal-dialing.js
```diff
@@ -0,0 +1,95 @@
...
+      assert.equal(browserSpock.getText('#outgoing-video-stats'), noStreamText);
+      browserSpock.click('[title="start sending video"]');
+      waitForSpecificText(browserSpock, '#outgoing-video-stats', noStreamText, true);
+    });
+
+    it('ends the call', () => {
+      browser.pause(5000);
+      // TODO add assertions around streams
+
+      browserSpock.click('[title="hangup"]');
```
adamweeks: This test doesn't test anything.

packages/node_modules/@ciscospark/plugin-phone/src/call.js
ianwremmel: actually, it's already done, i just need to remove the TODO

packages/node_modules/samples/browser-single-party-call-with-mute/test/wdio/spec/normal-dialing.js
ianwremmel: yea, none of these guys really do yet. I'm going to tackle that in the PR i'm starting right now (unifying sample code, creating more informative test helpers, etc).

ianwremmel:
merged via #795

PR #795, opened 2017-12-21
clean up sample tests

ianwremmel:
- [x] make `setValue()` a wdio command
- [x] make `waitForSpecificText()` a wdio command
- [x] fix lint warnings
- [x] reset at start of tests
- [x] remove all direct uses of `waitFor()`
- [x] remove all direct uses of `assert.*()`
- [x] normalize `index.html` and `app.js`


packages/node_modules/samples/browser-multi-party-call/test/wdio/spec/dial-and-reject.js
```diff
@@ -13,12 +12,14 @@ describe('sample-browser-multi-party-call', () => {
     const browserSpock = browser.select('browserSpock');
     const browserMccoy = browser.select('browserMccoy');
 
-    before('reload browser', () => browser.refresh());
+    before('reload browser', () => {
+      browser.select('browserSpock').reload();
```
adamweeks: You have the `const`'s up above, do you need to do the select over again? Also, `reload` spawns a whole new selenium test, did you mean to do a `refresh`?

packages/node_modules/samples/browser-multi-party-call/test/wdio/spec/dial-before-connect.js
```diff
@@ -13,12 +12,14 @@ describe('sample-browser-multi-party-call', () => {
     const browserSpock = browser.select('browserSpock');
     const browserMccoy = browser.select('browserMccoy');
 
-    before('reload browser', () => browser.refresh());
+    before('reload browser', () => {
+      browser.select('browserSpock').reload();
```
adamweeks: Same thing here, `reload` vs `refresh`

wdio.helpers.d/assertions.js
```diff
@@ -0,0 +1,50 @@
+const addCommand = require('./lib/add-command');
```
adamweeks: These helpers would be useful in the react repo too. Should we make them a package or put them in the web styleguide repo?

packages/node_modules/samples/browser-multi-party-call/test/wdio/spec/dial-and-reject.js
ianwremmel: it was mostly a matter of copying this block into each test. wasn't really paying attention to what was or was not already defined.

Yes, I'm intentionally restarting selenium. The tests were getting stuck in weird states; it about doubles the test duration, but they're way more stable.

wdio.helpers.d/assertions.js
ianwremmel: probably, but we really need a template for new standalone projects with semantic-release fully automated.

ianwremmel: we could put them in the styleguide, but then we'd probably need to justify it as a "style rule" somehow, though there is something to be said for disallowing `browser.waitUntil` in spec files. That still runs into the issue of not having a fully automated release process.

PR #796, opened 2017-12-22
fix(@ciscospark/spark-core): use while loop to remove token instead

tran2:
the forEach here won't remove all tokens due to this.userTokens.models
being modified during forEach, replacing with the while loop to remove
all the tokens

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
@@ -337,7 +337,6 @@ const Credentials = SparkPlugin.extend({
    */
   invalidate() {
     this.logger.info('credentials: invalidating tokens');
-    const tokens = this.userTokens.models;
 
     // clear refresh timer
     if (this.refreshTimer) {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -350,14 +349,14 @@ const Credentials = SparkPlugin.extend({
       this.logger.warn('credentials: failed to clear supertoken', err);
     }
 
-    tokens.forEach((token) => {
+    while (this.userTokens.models.length) {
       try {
-        this.userTokens.remove(token);
+        this.userTokens.remove(this.userTokens.models[0]);
```
aimeex3: is there a case where userTokens.remove fails? would it make sense to clear all the tokens using `this.userTokens.reset()`?

tran2: @aimeex3 yea, Ian said that there was a reason why he looped it initially but he forgot. And I didn't want to find out why so :P

ianwremmel: it may have something to do with the way the storage layer is rigged up. I don't remember for sure, but it's possible that `reset()` doesn't cause the storage layer to remove stuff.

PR #799, opened 2018-01-03
Plug the Memory Leak

ianwremmel:


ianwremmel:
I think I've got the main leak fixed. There still appears to be a slight leak, but it may just be how the garbage collector works. 

before:
about 1600 iterations
in `f(x)=ax+b`, `a=14093`
screen shot 2018-01-04 at 9 11 25 am

after:
about 5000 iterations
in `f(x)=ax+b`, `a=174`
screen shot 2018-01-04 at 9 19 05 am

interestingly, the after data has significantly more iterations because it runs quite a bit faster, which I assume to be from a sizeable reduction of gc-pauses, even though I'm not sure that makes sense.

packages/node_modules/@ciscospark/spark-core/src/lib/spark-plugin.js
```diff
@@ -121,19 +123,21 @@ const SparkPlugin = AmpState.extend({
...
+   * @param {Object} attrs
+   * @param {Object} options
+   * @returns {undefined}
+   */
   initialize(...args) {
     Reflect.apply(AmpState.prototype.initialize, this, args);
 
     // This is a bit of a hack to allow makeStateDataType work
-    let cloned = false;
+    this._dataTypes = cloneDeep(this._dataTypes);
```
adamweeks: img

ianwremmel:
blocker: something unrelated is messing up the refreshTimer unit test in chrome.

ianwremmel:
previous blocker resolved; now dealing with weird code coverage changes

PR #800, opened 2018-01-05
Separate ciscospark plugins into standalone packages

ianwremmel:
- [x] document auth plugins
- [x] document config properties correctly
- [x] fix public/private docs
- [x] fix describe nesting


packages/node_modules/@ciscospark/plugin-phone/src/config.js
```diff
@@ -2,46 +2,47 @@
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
 
+/**
+ * @typedef {Object} PhoneConfig
+ * @property {number} [audioBandwidthLimit=64000]
+ * @property {number} [videoBandwidthLimit=1000000]
+ * @property {boolean} [enableExperimentalGroupCallingSupport=false] - Group
```
adamweeks: Should `hangupIfLastActive` be defined here with the HangupBehavior type?

packages/node_modules/@ciscospark/plugin-logger/src/config.js
```diff
@@ -2,18 +2,22 @@
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
 
+/**
```
adamweeks: I like these defined up here, makes it much easier to read.

packages/node_modules/@ciscospark/plugin-logger/src/logger.js
```diff
@@ -149,6 +156,7 @@ const Logger = SparkPlugin.extend({
    * @instance
    * @memberof Logger
    * @private
+   * @memberof Logger
```
adamweeks: Already have `@memberof` two lines up

packages/node_modules/@ciscospark/plugin-logger/src/config.js
ianwremmel: haha, it wasn't really by choice; it was the only way i could get the docs generator to nest things in anyway approaching something reasonable.

packages/node_modules/@ciscospark/plugin-phone/src/config.js
ianwremmel: the `@memberof` in `HangupBehavior` takes care of it and renders slightly better. documentation.js still seems to be my favorite doc generator, but i've definitely gotten more familiar with its rough patches than I'd like.

PR #801, opened 2018-01-06
fix(@ciscospark/spark-core): catch token revocation failure

aimeex3:
if token.revoke fails, credentials.refresh() should still succeed. failure happens when there is an old token that was not cleaned up when logging out
  
I think this PR also addresses the error when logging out: https://github.com/ciscospark/spark-js-sdk/pull/796, but might be good to have both changes in. Even if token.revoke fails for a different reason, we still want credentials.refresh() to succeed

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
@@ -13,7 +13,7 @@ import {
   whileInFlight
 } from '@ciscospark/common';
 import {safeSetTimeout} from '@ciscospark/common-timers';
-import {cloneDeep, isObject} from 'lodash';
+import {clone, cloneDeep, isObject} from 'lodash';
 
 import SparkPlugin from '../spark-plugin';
 import {persist, waitForValue} from '../storage/decorators';
@@ -383,7 +383,7 @@ const Credentials = SparkPlugin.extend({
     this.logger.info('credentials: refresh requested');
 
     const supertoken = this.supertoken;
-    const tokens = this.userTokens.models;
+    const tokens = clone(this.userTokens.models);
 
     // This is kind of a leaky abstraction, since it relies on the authorization
     // plugin, but the only alternatives I see are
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -404,15 +404,18 @@ const Credentials = SparkPlugin.extend({
...
-          return this.downscope(token.scope)
-            // eslint-disable-next-line max-nested-callbacks
-            .then((t) => {
-              this.userTokens.add(t);
+        return Promise.all(tokens.map((token) => this.downscope(token.scope)
+          // eslint-disable-next-line max-nested-callbacks
+          .then((t) => {
+            this.userTokens.add(t);
+            this.logger.info(`credentials: revoking token for ${token.scope}`);
+            try {
```
ianwremmel: How were you seeing this fail? I would have expected a rejected promise, not a throw exception.

aimeex3: i think it throws an exception here. I debugged it with the dev tools and it fails at this line in `token.revoke()`: https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/token.js#L418

screen shot 2018-01-05 at 6 40 02 pm

ianwremmel: Gotcha. @tran2 does https://github.com/ciscospark/spark-js-sdk/pull/796 fix this? I feel like you were looking into this error before Christmas, but I can't remember exactly what that PR addresses.

tran2: @ianwremmel PR #796 should fix the issue where token.revoke() fails due to logout not removing all the tokens

aimeex3: 👍 #796 fixes underlying issue because token is not cleaned up when logging out. Added this in case token.revoke() fails for different reasons, but i'm okay to close this PR if it's not necessary

tran2: If somehow token was in a bad state, token.revoke() would fail. I think it makes sense to handle when it fails but I'm not sure how since I'm not familiar with this flow:
- Why was the revoke necessary here?
- If `token.revoke()` fails, the token won't get revoked (by sending a request to the backend) would there be any security issues since it might still be in the local storage?

aimeex3: @ianwremmel? do you think this is still necessary? I think revoke was called here to revoke the previous token after getting the new token. maybe we can remove it from localStorage if it fails... i think it calls userTokens.reset prior to this, but @ianwremmel mentioned that that may not remove it from localStorage.

PR #803, opened 2018-01-09
WIP: ROAP

ianwremmel:
- [x] extract roap_session into JS
- [x] extract roap_session tests into JS
- [x] move roap-session to `helper-roap`
- [ ] change plugin-locus to handle passing ROAP messages instead of just SDPs
- [ ] integrate roap-session with webrtc-media-engine and/or plugin-phone. (I'm still debating the right abstraction. On the one hand, ROAP deals with signalling, so perhaps should be added at the plugin-phone layer; on the other hand, ROAP deals with sdp exchange, so perhaps belongs at the media layer).


ianwremmel:
Blocked on account of a missing Locus API.

PR #806, opened 2018-01-22
feat(@ciscospark/plugin-phone): add #listActiveCalls method

adamweeks:


SCRIPTS.md
```diff
@@ -62,6 +62,12 @@ Test a single package and generate coverage and xunit reports
 npm test -- --package @ciscospark/spark-core --coverage --xunit
 ```
 
+Keeps the browser open so that you can reload set break points and reload the page
```
ianwremmel: wording

packages/node_modules/@ciscospark/plugin-phone/src/phone.js
```diff
@@ -174,6 +174,31 @@ const Phone = SparkPlugin.extend({
...
+   * @returns {Promise<Call[]>}
+   */
+  listActiveCalls() {
+    return this.spark.internal.locus.list()
+      .then((loci) => {
+        const calls = [];
+        if (!loci) {
+          return calls;
+        }
+        loci.forEach((locus) => {
```
ianwremmel: Instead of creating a call for each locus and then checking if it's been emitted, can we check if it's been emitted and, if not, create it? then once you've iterated over the retrieved loci, you can return something like `Array.from(this.emittedCalls)`. Part of the goal here is to ensure only one call is ever created for a given audio-video experience.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
```diff
@@ -240,6 +240,54 @@ browserOnly(describe)('plugin-phone', function () {
...
+            .then((result) => {
+              assert.isTrue(Array.isArray(result));
+              assert.equal(result.length, 1);
+              const activeCall = result[0];
+              assert.equal(activeCall.id, connectedCall.id);
+              assert.equal(activeCall.id, connectedCall.id);
+            });
+        });
+
+        it('should add call to emittedCalls collection', () => {
```
ianwremmel: avoid using "should" in test descriptions;`it('adds the call...)`

packages/node_modules/@ciscospark/plugin-phone/src/phone.js
ianwremmel: oh, i guess `Array.from` wouldn't work. I was thinking `emittedCalls` was a `Set`, but something like `clone(this.emittedCalls.models)`, maybe.

SCRIPTS.md
adamweeks: Copied from `CONTRIBUTING.md`, will fix both.

adamweeks:
@ianwremmel ready for round 2

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
```diff
@@ -240,6 +240,69 @@ browserOnly(describe)('plugin-phone', function () {
...
+          spock.spark.phone.listActiveCalls()
+            .then((callList) => {
+              assert.isTrue(Array.isArray(callList));
+              assert.equal(callList.length, 0);
+            }));
+      });
+
+      describe('when a ongoing call exists', () => {
+        let connectedCall;
+        beforeEach('create an existing call', () => handleErrorEvent(spock.spark.phone.dial(mccoy.email), (call) => expectCallIncomingEvent(mccoy.spark.phone)
```
ianwremmel: this might be flaky. I think you need the `expectCallIncomingEvent` down in the `Promise.all()`.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
```diff
@@ -240,6 +240,69 @@ browserOnly(describe)('plugin-phone', function () {
...
+            .then((result) => {
+              assert.isTrue(Array.isArray(result));
+              assert.equal(result.length, 1);
+              const activeCall = result[0];
+              assert.equal(activeCall.id, connectedCall.id);
+              assert.equal(activeCall.id, connectedCall.id);
+            });
+        });
+
+        it('adds call to emittedCalls collection', () => {
```
ianwremmel: Please add missing "the"s to test statements. "adds the call to emitte...", "does not add the call to the emitted...", etc

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
adamweeks: This was taken from an existing test, line 268 on `call.js`

adamweeks: Should I change that one too to increase stability?

adamweeks: Fixed this one for now.

packages/node_modules/@ciscospark/plugin-phone/test/integration/spec/phone.js
adamweeks: done

PR #809, opened 2018-01-25
Stateless Spark Plugin

ianwremmel:
Note: once this is approved I plan to 
1. remove the last two commits
2. rebase against master
3. apply the two transforms

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/spark-core-plugin-mixin.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/spark-core-plugin-mixin.js
@@ -58,15 +58,17 @@ export default function mixinSparkCorePlugins(State, config, interceptors) {
       }));
     }
 
+    // Only mess with the plugin's derived properties if it's an amp-state plugin
+    if (constructor.prototype._definition) {
+      if (constructor.prototype._definition.ready) {
+        const {fn, depList} = State.prototype._derived.ready;
+        const def = {
+          deps: depList.concat(`${name}.ready`),
+          fn
+        };
 
-    if (constructor.prototype._definition.ready) {
-      const {fn, depList} = State.prototype._derived.ready;
-      const def = {
-        deps: depList.concat(`${name}.ready`),
-        fn
-      };
-
-      createDerivedProperty(State.prototype, 'ready', def);
+        createDerivedProperty(State.prototype, 'ready', def);
+      }
     }
   };
 
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/spark-internal-core-plugin-mixin.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/spark-internal-core-plugin-mixin.js
@@ -51,15 +51,17 @@ export default function mixinSparkInternalCorePlugins(State, config, interceptor
       }));
     }
 
+    // Only mess with the plugin's derived properties if it's an amp-state plugin
+    if (constructor.prototype._definition) {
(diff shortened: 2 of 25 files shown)


packages/node_modules/@ciscospark/spark-core/src/lib/spark-core-plugin-mixin.js
```diff
@@ -58,15 +58,17 @@ export default function mixinSparkCorePlugins(State, config, interceptors) {
...
-      const {fn, depList} = State.prototype._derived.ready;
-      const def = {
-        deps: depList.concat(`${name}.ready`),
-        fn
-      };
-
-      createDerivedProperty(State.prototype, 'ready', def);
+    // Only mess with the plugin's derived properties if it's an amp-state plugin
+    if (constructor.prototype._definition) {
+      if (constructor.prototype._definition.ready) {
```
bzang: Could this be combined into:
``` js
if (constructor.prototype._definition &&
  constructor.prototype._definition.ready) {
```

packages/node_modules/@ciscospark/spark-core/src/lib/spark-internal-core-plugin-mixin.js
```diff
@@ -51,15 +51,17 @@ export default function mixinSparkInternalCorePlugins(State, config, interceptor
...
-      const {fn, depList} = State.prototype._derived.ready;
-      const def = {
-        deps: depList.concat(`${name}.ready`),
-        fn
-      };
-
-      createDerivedProperty(State.prototype, 'ready', def);
+    // Only mess with the plugin's derived properties if it's an amp-state plugin
+    if (constructor.prototype._definition) {
+      if (constructor.prototype._definition.ready) {
```
bzang: Same as above.
Also there seems to be quiet a bit of duplicate code between the internal and non-internal versions of this. Is there a way to simplify this?

tooling/codemods/legacy-to-stateless-spark-plugin.js
```diff
@@ -0,0 +1,157 @@
...
+      ]
+    })
+    .replaceWith(() => j.callExpression(
+      j.identifier('super'),
+      [
+        j.spreadElement(j.identifier('args'))
+      ]
+    ));
+
+  return `${root
```
bzang: this formatting makes my head hurt... initially thought there was something missing.

packages/node_modules/@ciscospark/spark-core/src/lib/spark-core-plugin-mixin.js
ianwremmel: yea, it could be. I liked the semantic difference of "check if this is a SparkPlugin vs StatelessSparkPlugin", then check if it has a ready property. I can certainly change it.

packages/node_modules/@ciscospark/spark-core/src/lib/spark-internal-core-plugin-mixin.js
ianwremmel: Probably. When I did the internal/external split, I think I tried a few things that were all pretty terrible, but it's certainly worth investigating again. That said, I've got some idea for making the plugin system more explicit, so, that might all be worth doing in a bigger refactor.

tooling/codemods/legacy-to-stateless-spark-plugin.js
ianwremmel: yea. one of our eslint rules does that. i like it most of the time, but I'm not a fan of what it does to template literals.

PR #821, opened 2018-02-13
standard readmes

ianwremmel:


packages/node_modules/@ciscospark/recipe-private-web-client/README.md
```diff
@@ -1,27 +1,33 @@
...
-  config: ...
-});
+```bash
+npm install --save @ciscospark/recipe-private-web-client
 ```
 
-### Development
+## Usage
+
+Don't use this. Seriously. Just don't.
```
adamweeks: LOL

ianwremmel:
I'm pretty sure the docs that got removed were very badly generated from code comments and actually didn't communicate things (a single method might have been documented as if it were a bare function and so forth). I'd like to see us start generating `## API` sections automatically, but that's a bigger set of work because we actually have to document code. (for example #817 and #646). Are there any specific docs that you were concerned with? I may well have overwritten soemthing of value since there were so many packages to change at once.

adamweeks:
I didn't see anything beside method names that got overwritten, so we should be good.

PR #823, opened 2018-02-15
Add ability to provide more options to appid user

adamweeks:
# Add ability to provide more options to appid user

## Description

JWT generation was not using the parameters as documented on the [developer portal](https://developer.ciscospark.com/guest-issuer.html). These changes will allow us to match the style as well as gives some new options.

There was also an issue with `dotenv` importing order that has been resolved.

## Type of Change

### DIFF
--- a/tooling/lib/test/index.js
+++ b/tooling/lib/test/index.js
@@ -7,8 +7,8 @@ const debug = require('debug')('tooling:test');
 
 const dotenv = require('dotenv');
 
-dotenv.config({path: '.env.default'});
 dotenv.config();
+dotenv.config({path: '.env.default'});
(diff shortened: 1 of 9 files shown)


tooling/lib/test/index.js
```diff
@@ -7,8 +7,8 @@ const debug = require('debug')('tooling:test');
 
 const dotenv = require('dotenv');
 
-dotenv.config({path: '.env.default'});
```
ianwremmel: wait, really? how has anything ever worked? :)

adamweeks: I'm guessing manual env vars having the actual keys you need

PR #824, opened 2018-02-15
mid call screen sharing

ianwremmel:
Closes #723 
Closes #652


packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -260,12 +284,15 @@ export default class WebRTCMediaEngine {
...
         // exception up the stack
-        resolve(this._getUserMedia());
+        resolve(this._getUserMedia()
+          .then(tap(() => this.logger.info('gum completed'))));
         return;
       }
 
       resolve();
     })
+      .then(() => {
```
adamweeks: Is this to get just another tick in?

ianwremmel: nope. looks like i pulled out some dev code, but missed the then block. will fix

PR #826, opened 2018-02-15
 feat(samples): add browser-auth-jwt sample

adamweeks:
Includes most of #823 and will rebase once merged.


packages/node_modules/samples/browser-auth-jwt/app.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-env browser */
+/* global ciscospark */
+/* eslint-disable camelcase */
+/* eslint-disable no-console */
+/* eslint-disable require-jsdoc */
+
+
+// Inititate the SDK
+const spark = ciscospark.init();
+const authBtn = document.getElementById('authBtn');
```
ianwremmel: nitpick, but throught out the rest of the samples, ids are kebab case.

adamweeks: The html was directly copied from `browser-auth-implicit` which has the same id for the authorize button.

ianwremmel: ah, gotcha. I should have caught it in that PR. mind switching both of them for consistency?

PR #830, opened 2018-02-20
feat(@ciscospark/plugin-phone): add call:created event

adamweeks:
A few things on this PR:

* A new phone event `call:created` was added
* Whenever a locus event happens to a call we aren't aware of via `emittedCalls`, the event will trigger
* Event names have been exported as an object `phoneEvents` to reduce the amount of "magic strings"
* Refactored the register locus listing to use the `listActiveCalls` method
* Expanded integration tests

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/src/phone.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/phone.js
@@ -7,8 +7,25 @@ import WebRTCMediaEngine from '@ciscospark/media-engine-webrtc';
 import {defaults, get} from 'lodash';
 
 import Call from './call';
-import {isCall, makeInternalCallId, shouldRing} from './state-parsers';
+import {isCall, shouldRing} from './state-parsers';
 import Calls from './calls';
+
+export const events = {
+  CALL_CREATED: 'call:created',
+  CALL_INCOMING: 'call:incoming'
+};
+
+/**
+ * Call Created Event
+ *
+ * Emitted when a call begins outside of the sdk
+ *
+ * @event call:created
+ * @instance
+ * @memberof Phone
+ * @type {Object}
+ * @property {Call} call The created call
+ */
 
 /**
  * Incoming Call Event
@@ -115,25 +132,13 @@ const Phone = SparkPlugin.extend({
           return Promise.resolve();
         }
         return Promise.all([
+          this.emittedCalls.reset(),
           this.spark.internal.mercury.when('event:mercury.buffer_state')
             .then(([message]) => {
               if (message.data.bufferState.locus === 'UNKNOWN') {
-                return this.spark.internal.locus.list();
+                return this.listActiveCalls();
               }
               return Promise.resolve();
-            })
-            .then((loci) => {
-              if (!loci) {
-                return;
-              }
-              // eslint-disable-next-line max-nested-callbacks
-              loci.forEach((locus) => {
-                this.trigger('call:incoming', Call.make({
-                  locus
-                }, {
-                  parent: this.spark
-                }));
-              });
             }),
           this.spark.internal.mercury.connect()
         ]);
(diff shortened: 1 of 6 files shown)


packages/node_modules/@ciscospark/plugin-phone/src/phone.js
```diff
@@ -192,8 +198,11 @@ const Phone = SparkPlugin.extend({
           if (!this.emittedCalls.get(makeInternalCallId(locus))) {
             const callItem = Call.make({locus, parent: this.spark});
             calls.push(callItem);
-            this.trigger('call:incoming', callItem);
             this.emittedCalls.add(callItem);
+            this.trigger(events.CALL_CREATED, callItem);
+            if (direction(locus) === 'in' && !joined(locus)) {
```
adamweeks: This *might* be considered a breaking change if someone was counting on the `call:incoming` event to (incorrectly) notify them of an existing call.

packages/node_modules/@ciscospark/plugin-phone/src/call.js
```diff
@@ -808,6 +808,7 @@ const Call = SparkPlugin.extend({
 
   // The complexity in createOrJoinLocus is largely driven up by fairly readable `||`s
   /* eslint-disable complexity */
+  /* eslint-disable no-param-reassign */
```
ianwremmel: This was just a warning, right? let's not disable warnings, that way way when we decide to fix them, we know where there are.

ianwremmel: (except for warnings that we're never going to fix. For example, payload transforms rely on a bunch of things that generate warnings and that'll probably never change)

packages/node_modules/@ciscospark/plugin-phone/src/phone.js
```diff
@@ -239,20 +248,26 @@ const Phone = SparkPlugin.extend({
...
+    const call = Call.make({
+      locus: event.data.locus
+    }, {
+      parent: this.spark
+    });
+    this.emittedCalls.add(call);
 
-      this.trigger('call:incoming', call);
+    // Trigger events as necessary
+    this.trigger(events.CALL_CREATED, call);
```
ianwremmel: Do I correctly read this as emit `CALL_CREATED` once for every call and further emit `CALL_INCOMING` once if appropriate?

packages/node_modules/@ciscospark/plugin-phone/src/phone.js
```diff
@@ -192,8 +198,11 @@ const Phone = SparkPlugin.extend({
           if (!this.emittedCalls.get(makeInternalCallId(locus))) {
             const callItem = Call.make({locus, parent: this.spark});
             calls.push(callItem);
-            this.trigger('call:incoming', callItem);
             this.emittedCalls.add(callItem);
+            this.trigger(events.CALL_CREATED, callItem);
+            if (direction(locus) === 'in' && !joined(locus)) {
+              this.trigger(events.CALL_INCOMING, callItem);
```
ianwremmel: given that we've got the `call:created` event, I feel like this should probably just mirror the `onLocusEvent` logic, right?

packages/node_modules/@ciscospark/plugin-phone/src/call.js
adamweeks: I thought it was an error but I’ll double check

packages/node_modules/@ciscospark/plugin-phone/src/phone.js
adamweeks: My concern was that the `alertType.action` might be not correct when doing a `locus.list` but it looks like it is correct after inspecting the network response. I'll have them use a shared function.

adamweeks:
@ianwremmel requested changes have been made

PR #848, opened 2018-03-13
chore(tooling): add option to include samples

adamweeks:


tooling/lib/npm.js
```diff
@@ -48,5 +49,8 @@ exports.getDistTag = async function getDistTag(packageName) {
       debug('Something went wrong, but we had to use --silent, so it\'s hard to tell what', err);
     }
   }
+  else if (includeSamples && packageName === 'samples') {
+    return pkg.version;
```
andrewholsted: Do you need to do the fancy tag check is just returning the package.json sufficient? Does samples need to be versioned if it's private?

tooling/lib/npm.js
```diff
@@ -9,10 +9,11 @@ const {exec} = require('./async');
 /**
  * Returns the dist-tag for the specified package
  * @param {string} packageName
+ * @param {bool} [includeSamples = false]
  * @private
  * @returns {Promise<string>}
  */
-exports.getDistTag = async function getDistTag(packageName) {
+exports.getDistTag = async function getDistTag(packageName, includeSamples = false) {
```
bzang: To keep consistent with the other tooling methods, let's put these into a single object instead of 2 separate params. (i.e. `getDistTab({packageName, includeSamples = false}) { }`)

package.json
```diff
@@ -74,7 +74,7 @@
     "distsrc": "sed -ibak 's#\"main\": \"dist/index.js#\"main\": \"src/index.js#' packages/node_modules/{*,@ciscospark/*}/package.json && rimraf packages/node_modules/{*,@ciscospark/*}/*bak",
     "srcdist": "sed -ibak 's#\"main\": \"src/index.js#\"main\": \"dist/index.js#' packages/node_modules/{*,@ciscospark/*}/package.json && rimraf packages/node_modules/{*,@ciscospark/*}/*bak",
     "publish:docs": "grunt --gruntfile Gruntfile.concurrent.js publish:docs",
-    "get-next-version": "babel-node ./tooling/index.js version next --always",
+    "get-next-version": "babel-node ./tooling/index.js version next --always --includeSamples",
```
bzang: Is there a case where we wouldn't want to include samples in the version calculation?

adamweeks: I have no idea, but I wanted to make it configurable

tooling/lib/npm.js
adamweeks: Samples definitely gets version bumped so we should check it. Not sure why we are going through npm to get the version instead of just package.json

tooling/lib/npm.js
adamweeks: Done

adamweeks:
@bzang ready again

PR #849, opened 2018-03-13
fix(internal-plugin-avatar): profile image orientation

saurjai3:
# Pull Request Template

## Description

Fixes issue with the image orientation on profile avatars. The code for the same was earlier available as part of internal-plugin-conversation, but that is moved to helper-image so that the same code with minimal changes could be utilized for internal-plugin-avatar.

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-2864

## Type of Change

## Test Coverage

- Test images to reproduce and verify the existing bug could be referred from: https://github.com/recurser/exif-orientation-examples

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/helper-image/src/process-image.js
+++ b/packages/node_modules/@ciscospark/helper-image/src/process-image.js
@@ -12,6 +12,7 @@ import {pick} from 'lodash';
  * @param {Number} options.thumbnailMaxWidth
  * @param {Number} options.thumbnailMaxHeight
  * @param {Boolean} options.enableThumbnails
+ * @param {Object} options.logger
  * @returns {Promise<Array>} Buffer, Dimensions, thumbnailDimensions
  */
 export default function processImage({
--- a/packages/node_modules/@ciscospark/helper-image/package.json
+++ b/packages/node_modules/@ciscospark/helper-image/package.json
@@ -10,6 +10,10 @@
   "engines": {
     "node": ">=4"
   },
+  "browser": {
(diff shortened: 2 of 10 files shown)


packages/node_modules/@ciscospark/helper-image/src/process-image.js
```diff
@@ -12,9 +12,11 @@ import {pick} from 'lodash';
  * @param {Number} options.thumbnailMaxWidth
  * @param {Number} options.thumbnailMaxHeight
  * @param {Boolean} options.enableThumbnails
+ * @param {Object} options.logger
  * @returns {Promise<Array>} Buffer, Dimensions, thumbnailDimensions
  */
-export default function processImage({
+// eslint-disable-next-line import/prefer-default-export
```
adamweeks: Any reason why the default was removed? [Our reasoning](https://github.com/ciscospark/web-styleguide/tree/master/javascript#modules--prefer-default-export) behind the rule.

saurjai3: process-image.browser.js has two functions detectFileType() and processImage() which are called from outside the helper-image package and hence the default is removed. The default is also removed from process-image.js to keep both the node and browser version in sync.

adamweeks: Sounds good. Could you please make a note in comments above the disable for that? I think we are good once those changes are in place.

PR #853, opened 2018-03-21
Update Install & Test Instructions

mattnorris:
# Update Install & Test Instructions

## Description

Updated installation and testing instructions for contributors. Added a `BUILD-ISSUES.md` file to reference and more information on setting up SauceLabs to run the SDK tests. 

## Type of Change

- This change requires a documentation update

## Test Coverage

Tests are not required because these are only documentation changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,9 +40,10 @@ Please provide sufficient logging around the issue which you are reporting as th
 
 Before you can build the Cisco Webex JS SDK, you will need the following dependencies:
 
-- [Node.js](https://nodejs.org/) 6.x (LTS)
+- [Node.js](https://nodejs.org/) (LTS)
   - We recommend using [nvm](https://github.com/creationix/nvm) (or [nvm-windows](https://github.com/coreybutler/nvm-windows))
     to easily switch between Node.js versions
+  - Install the latest Node.js Long Term Support using `nvm install --lts`
 - [Git](https://git-scm.com/)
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -74,8 +75,23 @@ Build the SDK:
...
 ```
 
+*Build issues?* See [BUILD-ISSUES.md](./BUILD-ISSUES.md) for help.
+
 ### Running Tests
 
+#### SauceLabs
+
+The SDK uses [SauceLabs](https://saucelabs.com/) to run its tests. Sign in to retrieve your *USERNAME* and *ACCESS KEY* from [User Settings](https://saucelabs.com/beta/user-settings) and add them to your `.env` file: 
+
```
bzang: Is sauce labs actually required? My understanding is for most people they don't need sauce credentials and can run tests locally. Sauce is only used for CI.

mattnorris: When I tried to run the basic (complete) suite of tests it failed because it needed the username and access key (and tunnel ID, ultimately) from Sauce. I also picked a couple packages at random to scope it and it failed as well. Maybe not all packages need it, but getting the basics up an running requires it. 

It looks like there are 61 packages. I could do some digging and see what requires what.

adamweeks: You should be able to run all the tests locally without sauce from my understanding as well. Curious to see your findings.

mattnorris:
An oldie but a goodie... 😉 

@bzang revised my SauceLabs comments now that I understand a bit more about them.

PR #855, opened 2018-03-21
fix(spark-core): emit InvalidRequestError, if refresh token expires

saurjai3:
# Pull Request Template

## Description

The change fixes the issue where the refresh_token is expired for a user, as a result when a request for refreshing the access_token is made, it fails too with an error "The refresh token provided is expired, revoked, malformed, or invalid". This results in hanging of the web-client. We would want the user to be logged out when this condition is met.

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-5100

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
@@ -425,6 +425,23 @@ const Credentials = SparkPlugin.extend({
       })
       .then(() => {
         this.scheduleRefresh(this.supertoken.expires);
+      })
+      .catch((error) => {
+        const {InvalidRequestError} = grantErrors;
+        if (error instanceof InvalidRequestError) {
+          // Error: The refresh token provided is expired, revoked, malformed, or invalid. Hence emit an event to the client, an opportunity to logout.
+          this.unset('supertoken');
+          while (this.userTokens.models.length) {
+            try {
+              this.userTokens.remove(this.userTokens.models[0]);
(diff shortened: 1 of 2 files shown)


bzang:
@saurjai3 Is that the correct behavior? I would expect a logout only if both the refresh token and the access token are expired, not just the refresh token.

saurjai3:
absolutely @bzang , the access token would already be expired which would result in calling credentials.refresh(). If that fails because of the already expired refresh token, then that's the point we would do logout(conditional check on auth.js to ensure that the error is of type InvalidRequestError).

adamweeks:
Can we get some tests added for this feature?

saurjai3:
added the test case.

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/credentials.js
```diff
@@ -420,6 +420,23 @@ const Credentials = SparkPlugin.extend({
       })
       .then(() => {
         this.scheduleRefresh(this.supertoken.expires);
+      })
+      .catch((error) => {
+        const {InvalidRequestError} = grantErrors;
```
aimeex3: should this be defined above? don't need to set it in a variable every time it hits the catch

saurjai3:
All tests on spark-core passed locally.

PR #859, opened 2018-03-30
fix(media-engine-webrtc): allow replacement of tracks in call

### LINKED ISSUE #858: Replacing localMediaStream causes uncaught exception
## Expected Behavior
When we programmatically change the localMediaStream directly using something like `call.localMediaStream = stream` the stream should update correctly.

## Current Behavior
An uncaught exception at this line: https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/%40ciscospark/media-engine-webrtc/src/engine.js#L676

## Possible Solution (we welcome any logical suggestion)
Issue is because `removeTrack()` is being called on `this.pc`, so we remove the track from the peerConnection and `existing.track` no longer exists when being removed from `localMediaStream`.

Reversing the removal calls solves the issue:
```
this.localMediaStream.removeTrack(existing.track);
this.pc.removeTrack(existing);
```

## Steps to Reproduce
1. Establish a call
2. Create a new media stream with a different device
3. Set the new media stream

bzang:
## Description

Remove the old track from the localMediaStream first before completely removing it from the peerConnection.

Fixes #858

## Type of Change

## Test Coverage

Test coverage for this is dicey. Need some help seeing how we can test this.

**Test Configuration**:
* SDK Version: `v1.29.5`
* Node/Browser Version: `v8.9.4`
* NPM Version: `5.7.1`

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
+++ b/packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
@@ -672,8 +672,8 @@ export default class WebRTCMediaEngine {
     const existing = this.pc.getSenders().find((s) => s.track.kind === track.kind && s.track !== track);
     if (existing) {
       this.logger.info(`removing previous ${track.kind} from local media stream`);
-      this.pc.removeTrack(existing);
       this.localMediaStream.removeTrack(existing.track);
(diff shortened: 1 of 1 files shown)


bzang:
Superseded by #910

PR #862, opened 2018-04-04
fix(plugin-phone): change locus media muting toggles

adamweeks:
## Description

It was determined that we were doing audio/video muting incorrectly. This lines up the phone plugin with what locus is expecting during muting events.

Fixes # SSDK-1841

## Type of Change


adamweeks:
Blocked by webrtc media engine failures from Firefox 59

mattnorris:
Outdated. Fixed in `plugin-meetings`.

PR #874, opened 2018-04-18
fix(media-engine-webrtc): add support for firefox 59

adamweeks:
Lots of changes in the way sdp's are generated in firefox 59, as well as how getSenders responds.


packages/node_modules/@ciscospark/media-engine-webrtc/renegotiation-support.md
```diff
@@ -0,0 +1,50 @@
...
+| end sendrecv | SKIP           | Good           | SKIP           | N/A            |
+
+### Chrome Video
+
+|              | start inactive | start recvonly | start sendonly | start sendrecv |
+| ------------ | -------------- | -------------- | -------------- | -------------- |
+| end inactive | N/A            | Good           | Good           | Good           |
+| end recvonly | Good           | N/A            | SKIP           | Good           |
+| end sendonly | Good           | Good           | N/A            | Good           |
+| end sendrecv | Good           | Good           | SKIP           | N/A            |
```
mattnorris: Nice writeup 👍

packages/node_modules/@ciscospark/media-engine-webrtc/test/unit/spec/engine.js
```diff
@@ -179,7 +191,7 @@ browserOnly(describe)('media-engine-webrtc', function () {
                   engine.stop();
                 });
 
-                step(`initiates a session with audio=${audioStart} and video=${videoStart}`, () => {
+                (((browser.chrome && (flakyInChrome || brokenInChrome)) || (browser.firefox && brokenInFirefox)) ? it.skip : step)(`initiates a session with audio=${audioStart} and video=${videoStart}`, () => {
```
mattnorris: This may be more readable broken out into a few lines or variables, .e.g., `skipChrome = browser.chrome && (flaklyInChrome || brokenInChrome);`

packages/node_modules/@ciscospark/media-engine-webrtc/renegotiation-support.md
adamweeks: Thanks, I didn't have a lot of info on why the chrome tests were skipped other than the comments in the code, so I didn't go into as much detail.

packages/node_modules/@ciscospark/media-engine-webrtc/test/unit/spec/engine.js
adamweeks: Yea, I just copied what was existing, but I'll see if I can make it a bit more readable

adamweeks: Fixed

packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -429,7 +431,7 @@ export default class WebRTCMediaEngine {
           // Note: Doesn't like removing senders whose tracks are not live, so
           // t.stop() has to come after removeTrack()
           try {
-            this.pc.removeTrack(this.pc.getSenders().find((s) => s.track === t));
+            this.pc.removeTrack(this.pc.getSenders().find((s) => s.track && s.track === t));
```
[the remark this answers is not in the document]
adamweeks: Yes, FF59 adds another sender to the response which doesn't have a track property. I still haven't figured out what this new sender is for.

packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -440,8 +442,8 @@ export default class WebRTCMediaEngine {
             }
             this.logger.warn(err);
           }
-          this.logger.info(`setting receiving${capitalize[kind]} to false`);
-          this[`sending${capitalize[kind]}`] = false;
+          this.logger.info(`setting sending${capitalize[kind]} to ${constraint}`);
+          this[`sending${capitalize[kind]}`] = constraint;
```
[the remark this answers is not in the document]
adamweeks: It will hit, but it looks like `constraint` will always be false. I could switch it back to hard coded boolean, but I like that we know where the value is coming from.

packages/node_modules/@ciscospark/media-engine-webrtc/src/engine.js
```diff
@@ -669,11 +671,13 @@ export default class WebRTCMediaEngine {
...
     this.logger.info(`preparing to add ${track.kind} to local media stream`);
-    const existing = this.pc.getSenders().find((s) => s.track.kind === track.kind && s.track !== track);
+    const existing = this.pc.getSenders().find((s) => s.track && s.track.kind === track.kind && s.track !== track);
     if (existing) {
       this.logger.info(`removing previous ${track.kind} from local media stream`);
+      // Track object is removed from existing after calling `pc.removeTrack`
+      const existingTrack = existing.track;
       this.pc.removeTrack(existing);
-      this.localMediaStream.removeTrack(existing.track);
+      this.localMediaStream.removeTrack(existingTrack);
```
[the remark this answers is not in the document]
adamweeks: `existingTrack` is still around but the `existing` object will no longer have a `track` property. There was a note somewhere about peer connection not liking removing tracks that are already removed from local media stream, so I kept the same order by doing it this way.

Guess I totally missed it in your PR

adamweeks:
@bzang can you point to where you would want the comments? re:

>You may also want to put some inline comments in the tests where the returned SDP doesn't match setMedia options. Maybe pointing to the readme you created?

bzang:
@adamweeks  I was thinking above the tests that have a mismatch to just say what's being passed into setMedia and what the expected value in the offer is. Maybe here, for example? https://github.com/webex/spark-js-sdk/pull/874/files#diff-70c763e5d4c06b4b5cf0436f3305e414R686
I'm just thinking someone who hasn't ready the readme (like me originally) might find it very strange that what goes in isn't what ends up being set.

adamweeks:
@bzang ah ok, for the screenshare stuff, yea, I'll put some comments around it with the changes due to FF59

bzang:
@adamweeks that was an example, really any place where there's a mismatch could probably use a little comment on why there's a mismatch.

adamweeks:
@bzang added some more docs as comments

PR #880, opened 2018-05-04
Add @webex/sparkd package

mattnorris:
# Pull Request

## Description

Adds the first @webex org package to publish. `sparkd` is not a dependency of any other package and is in low, specialized use, making it a good candidate to test rebranding. 

Introduces a change to the test script so that specific packages are ignored by browser tests. Some packages are meant to run only in Node, so we now have the ability to configure that and avoid confusion when running tests via command line. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SSDK-1910

## Type of Change

## Test Coverage

```
# Ran to confirm browser tests were ignored implicitly. 
npm test -- --package @webex/sparkd 
npm test -- --package @ciscospark/sparkd 

# Ran to confirm browser tests were ignored explicitly. 
npm test -- --package @webex/sparkd --node

# Ran to confirm browser tests could still be run explicitly. 
npm test -- --package @webex/sparkd --browser

# Ran to confirm browser and Node tests were tested implicitly in other packages. 
npm test -- --package @webex/plugin-rooms
```

**Test Configuration**:
* SDK Version: 1.31
* Node/Browser Version: 8.10.0
* NPM Version: 5.8.0

## Checklist:

### DIFF
--- a/tooling/commands/test.js
+++ b/tooling/commands/test.js
@@ -12,6 +12,21 @@ const spawn = require('../util/spawn');
 const {report} = require('../util/coverage');
 const {start, stop} = require('../util/server');
 
+/**
+ * Returns true if the given package should be tested
+ * in the browser. (Some packages are intended for use
+ * only in Node.)
+ * @param {String} packageName
+ * @returns {Boolean}
+ */
+function shouldTestInBrowser(packageName) {
+  const noBrowserPackages = [
+    '@ciscospark/sparkd',
+    '@webex/sparkd'
+  ];
+  return !noBrowserPackages.includes(packageName);
+}
+
 module.exports = {
   command: 'test',
   desc: 'Run the test suite',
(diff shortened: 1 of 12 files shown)


tooling/commands/test.js
```diff
@@ -12,6 +12,21 @@ const spawn = require('../util/spawn');
...
+ * Returns true if the given package should be tested
+ * in the browser. (Some packages are intended for use
+ * only in Node.)
+ * @param {String} packageName
+ * @returns {Boolean}
+ */
+function shouldTestInBrowser(packageName) {
+  const noBrowserPackages = [
+    '@ciscospark/sparkd',
+    '@webex/sparkd'
```
bzang: Nice! Only thought would be to store the list as a constant or config somewhere.

bzang:
Should this be named `teamsd` instead?

mattnorris:
@bzang deprecating will be done manually via npm after this is published. 

Regarding the name, most likely `webexproxyd` or something like that. I've spoken with @bbender about naming. We may have more clarity next week.

tooling/commands/test.js
mattnorris: I thought about that, but wanted to reduce the mental overhead of looking elsewhere for the exclusions. If this becomes a large list it makes sense to have it as part of a config.

PR #888, opened 2018-05-21
chore(testing): add ability to limit browsers

adamweeks:
## Description

Since firefox is a source of pain currently, this change will allow us to provide a command line variable to limit which browser the tests are run on. 

## Type of Change

## Test Coverage

- [X] Phone plugin tests with only chrome

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/SCRIPTS.md
+++ b/SCRIPTS.md
@@ -54,6 +54,12 @@ Test a single package, but only in a browser
 npm test -- --package @ciscospark/spark-core --browser
 ```
 
+Test a single package, but only in a specific browser
+
+```bash
+BROWSER=chrome npm test -- --package @ciscospark/spark-core --browser
+```
+
 Test a single package and generate coverage and xunit reports
 
 ```bash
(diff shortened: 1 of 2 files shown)


SCRIPTS.md
```diff
@@ -54,6 +54,12 @@ Test a single package, but only in a browser
 npm test -- --package @ciscospark/spark-core --browser
 ```
 
+Test a single package, but only in a specific browser
+
+```bash
+USE_BROWSER=chrome npm test -- --package @ciscospark/spark-core --browser
```
bzang: Is the `BROWSER` flag already used? If not, I'd suggest that since it would match with the widgets project.

adamweeks: My main concern was that there would be two "browser"s, one env var for limiting the browser and one command line flag that said to run the browser suites. If you don't think that it will cause confusion, I will change.

mattnorris: We could change `--browser` to `--browser-tests` to be more clear if it is an issue. I suspect this would have to change widget-wise too.

adamweeks: Ok, I'll make the change to `BROWSER=chrome` and leave `--browser` for now and see how it is received.

PR #937, opened 2018-08-06
feat(plugin-phone): remove call:created event for inactive loci

adamweeks:
# Pull Request Template

## Description

Call objects will still emit locus events, but the main phone pluginshould not create calls for inactive locus events.

Fixes # SPARK-18814

## Type of Change

## Test Coverage

- [X] Plugin-phone tests

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-phone/src/phone.js
+++ b/packages/node_modules/@ciscospark/plugin-phone/src/phone.js
@@ -244,17 +244,20 @@ const Phone = SparkPlugin.extend({
    * @returns {undefined}
    */
   onLocusEvent(event) {
-    // We only want to handle calls we are not aware of
-    if (this.emittedCalls.has(event.data.locus)) {
+    const {locus} = event.data;
+
+    // We only want to handle calls we are not aware of in an active state
+    if (this.emittedCalls.has(locus) || locus.fullState.state.toLowerCase() === 'inactive') {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


packages/node_modules/@ciscospark/plugin-phone/src/phone.js
```diff
@@ -244,17 +244,20 @@ const Phone = SparkPlugin.extend({
    * @returns {undefined}
    */
   onLocusEvent(event) {
-    // We only want to handle calls we are not aware of
-    if (this.emittedCalls.has(event.data.locus)) {
+    const {locus} = event.data;
+
+    // We only want to handle calls we are not aware of in an active state
+    if (this.emittedCalls.has(locus) || locus.fullState.state.toLowerCase() === 'inactive') {
```
bzang: Based on the flows we covered a couple weeks ago, it is possible to go from `inactive` to something else and that behavior should be reflected by events. If that's the case, will this logic still work for that condition?

adamweeks: The actual call object will continue to emit events as locus events, but the phone plugin will not emit new call events (`call:created` and `call:incoming`) for inactive locus objects.

bzang: Ahhh, that's right. Call locus events are handled elsewhere.

PR #939, opened 2018-08-10
feat(plugin-meetings): initial plugin scaffolding

bzang:
I created the initial scaffolding for the meetings plugin. Should be pretty self explanatory, but let me know if there's any weak spots where I can beef up the docs or comments.

All of our actual plugin code will go into `src` folder. Eventually we'll remove the `ciscospark.js` file once we hook in all the plugin code. For now, there's a `Meetings` class to get us started.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "test:samples": "echo 'Deprecated. Please use npm run samples:test'; npm run --silent samples:test",
     "samples:build": "babel-node ./tooling/index.js build --only-samples",
     "samples:serve": "webpack-dev-server",
+    "serve:package": "webpack-dev-server --config webpack/webpack.config.js --hot --inline --history-api-fallback",
     "presamples:test": "rimraf bundle.js bundle.js.map",
     "samples:test": "NODE_ENV=test wdio",
     "tooling": "babel-node ./tooling/index.js",
@@ -223,6 +224,7 @@
     "webpack": "^4.12.0",
     "webpack-cli": "^3.0.8",
     "webpack-dev-server": "^3.1.2",
+    "webpack-merge": "^4.1.4",
     "webrtc-adapter": "^6.1.0",
     "whatwg-fetch": "^2.0.3",
     "ws": "^4.0.0",
@@ -263,5 +265,8 @@
     "Jenkinsfile": "lint:jenkinsfile",
     "*package.json": "lint:package-names"
   },
-  "snyk": true
+  "snyk": true,
+  "devDependencies": {
+    "html-webpack-plugin": "^3.2.0"
+  }
 }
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -263,5 +265,8 @@
     "Jenkinsfile": "lint:jenkinsfile",
     "*package.json": "lint:package-names"
   },
-  "snyk": true
+  "snyk": true,
```
arun3528: Can we disable as this doesn't work on windows ?

packages/node_modules/@webex/plugin-meetings/README.md
```diff
@@ -0,0 +1,43 @@
...
+npm install --save @webex/plugin-meetings
+```
+
+## Usage
+
+This is a plugin for the Cisco Webex JS SDK . Please see our [developer portal](https://developer.webex.com/sdks-and-widgets.html) and the [API docs](https://webex.github.io/spark-js-sdk/api/) for full details.
+
+## Development
+
+To use `webpack-dev-server` to load this package, run `npm run serve:package -- --env.package @webex/plugin-meetings` from the root of this repo.
```
arun3528: do you think it will work on widows machine ?

bzang: The syntax may need to change, but it should. There may be a need to install the linux subsystem to do this.

package.json
bzang: We can't get rid of it yet since we need it to meet security requirements. Once we update to the npm@>6, we can get rid of it in place of `npm-audit`

mattnorris: @bzang are you aware of anything that will fail if we start enforcing npm 6?

bzang: The only thing that I've been able to find is that the minimum version of node is 6. Currently we supposedly still support back to v4. I think it's reasonable to force developers to move to version of node that's newer.

package.json
```diff
@@ -263,5 +265,8 @@
...
+  "devDependencies": {
```
adamweeks: Did we ever figure out why we didn't have devDependencies before?

webpack/webpack.config.base.js
```diff
@@ -0,0 +1,51 @@
...
+      }
+    ]
+  },
+  plugins: [
+    new webpack.EnvironmentPlugin({
+      CISCOSPARK_LOG_LEVEL: 'log',
+      DEBUG: '',
+      NODE_ENV: process.env.NODE_ENV || 'development',
+      CISCOSPARK_ACCESS_TOKEN: process.env.CISCOSPARK_ACCESS_TOKEN,
+      TO_PERSON: process.env.TO_PERSON,
```
adamweeks: Probably don't need this env var

package.json
bzang: It's an sdk so everything is a normal dependency? I should probably move these over as a normal dependency.

PR #959, opened 2018-08-30
Remove Snyk & add support for `npm audit`

mattnorris:
# Pull Request

## Description

Snyk security auditing has trouble with Windows. To have full-featured meetings, they must work on Edge. This removes the Snyk requirement and adds support for `npm audit` which is able to do "in-house" security scans.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-33139

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #960, opened 2018-08-30
chore(tooling): update circle ci to v2

bzang:
## Description

Circle CI is deprecating v1 of their config. This updates us to v2.

Fixes SPARK-32635

## Type of Change


adamweeks:
@mattnorris 
>will this take care of the config you mentioned for #959 ?

Not quite, because @bzang didn't add the greenkeeper update sections. We'll have to add those after the package.lock is enabled.

I think we should just go ahead and do both of the PRs and then we can setup another for greenkeeper package lock updates.

PR #971, opened 2018-09-20
fix(plugin-phone): add devices to memberships

adamweeks:
If a device is not associated to a user, it should be considered its own membership and treated as one.

This is to fix the issue of a 1:1 call to a device not registering the device as a member and the call never ends if the device hangs up first.

# Pull Request Template

## Description

Fixes # SPARK-29430

## Type of Change

## Test Coverage

- [X] plugin-phone

**Test Configuration**:
* SDK Version v1.41.0
* Node/Browser Version 8.11.3
* NPM Version 5.6.0

## Checklist:


adamweeks:
@mattnorris not that I'm aware of. We would need access to a TP unit for testing and control of the TP unit to have it hang up.

I can add some unit tests to the function though, that way we'll get our expectations testing.

PR #983, opened 2018-10-03
fix(i-p-conversation): auth header on key rotation

saurjai3:
## Description
Currently when the convo DTO does not contain the encryptionKeyUrl but still has a valid KRO, the JS sdk fails with a 401 while doing the key rotation. This change would fix the 401 during this scenario.

## Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
@@ -249,6 +249,10 @@ const Conversation = SparkPlugin.extend({
           }, omit(options, 'id', 'user', 'url'))
         };
 
+        // to prevent a 401, this will add the auth token while doing
+        // key rotation
+        params.runWhitelistedDomains = true;
+
         return Promise.resolve(user ? this.spark.internal.user.asUUID(user) : null)
           .then((userId) => {
             if (userId) {


packages/node_modules/@ciscospark/internal-plugin-conversation/src/conversation.js
```diff
@@ -249,6 +249,10 @@ const Conversation = SparkPlugin.extend({
           }, omit(options, 'id', 'user', 'url'))
         };
 
+        // to prevent a 401, this will add the auth token while doing
+        // key rotation
+        params.runWhitelistedDomains = true;
```
mattnorris: 1. Can you think of any time when we *wouldn't* want to do this? 
2. Please [add a test](https://github.com/webex/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/internal-plugin-conversation/test/integration/spec/get.js#L75).

saurjai3:
We do not need these changes. web-ha-messaging was enabled on my integration profile due to which the conversation url was getting changed to https://conv.intb3.ciscospark.com which did not match the service catalog and hence failing with a 401. The feature toggle is not ready for release yet and hence should be disabled before trying the key rotation changes. Once I disabled the feature flag, it is working similar to other clients, hence cancelling these changes.

PR #996, opened 2018-10-10
Remove /env from the README 

taymoork2:
## Description

The node sample was using `ciscospark` instead of `ciscospark/env`,
`ciscospark/env` is the recommended way to use the SDK in a non-browser environment.
However, I suggest that we start to move away (and possible deprecation) from `ciscospark/env` and have people move to just `ciscospark` and have users explicitly `init()` the library.
`/env`'s grabbing of the access token and initializing the SDK "automagically" isn't the best and having two different ways to initiate the SDK is just confusing at the end of the day.
I've included a section that still tells users how to use `ciscospark/env`

I'm also suggesting moving away from the Browserify sample and move to a Parcel sample.

Fixes [38320](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-38320)

## Type of Change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,104 +33,122 @@ npm install --save ciscospark
 ## Usage
 
 To use the SDK, you will need Cisco Webex credentials. If you do not already have a Cisco Webex account, visit
-[Cisco Webex for Developers](https://developer.webex.com/) to create your account and retrieve your access token.
+[Cisco Webex for Developers](https://developer.webex.com/) to create your account and retrieve your **_access token_**.
 
 See [the detailed docs](https://webex.github.io/spark-js-sdk/) for more usage examples.
 
-You will need to set the following environment variable:
-
-- `CISCOSPARK_ACCESS_TOKEN`
-
 ```javascript
 const ciscospark = require(`ciscospark`);
-const spark = ciscospark.init({
+const teams = ciscospark.init({
   credentials: {
-    access_token: process.env.CISCOSPARK_ACCESS_TOKEN
+    access_token: <your webex teams access token>
   }
 });
 
-spark.rooms
+// Create a room with the title "My First Room"
+// Add Alice and Bob to the room
+// Send a **Hi Everyone** message to the room
+teams.rooms.create({ title: `My First Room` }).then(room => {
+  return Promise.all([
+    ciscospark.memberships.create({
+      roomId: room.id,
+      personEmail: `alice@example.com`
+    }),
+    ciscospark.memberships.create({
+      roomId: room.id,
+      personEmail: `bob@example.com`
+    })
+  ]).then(() =>
+    ciscospark.messages.create({
+      markdown: `**Hi Everyone**`,
+      roomId: room.id
+    })
+  );
+});
+```
+
+#### _A note on browser usage_
+
+We do not provide a built version of the SDK that includes `window.ciscospark`.
+In-browser usage is almost the same as Node.js, but it handles the user authentication flow for you. See the [browser guide](https://webex.github.io/spark-js-sdk/guides/browsers/) for more information.
+
+If you're already using a bundler (like [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org/)) you can simply import/require the package and use the above snippet and assign the initialized `team` variable to `window.webexteams`.
+For a quick example, we'll use [Parcel](https://parceljs.org/) to bundle the SDK for a website. For any more information and questions on how to use Parcel, please head to their [website](https://parceljs.org/).
+
+1. Create `index.js`.
+
+```javascript
+import { init as initTeams } from 'ciscospark';
+
+// Initialize the SDK and make it available to the window
+const teams = (window.webexteams = initTeams({
+  credentials: {
+    access_token: <your webex teams access token>
+  }
+}));
+
+// Create a room with the title "My First Room"
+teams.rooms
   .create({
     title: 'My First Room!'
   })
-  .catch(function(error) {
+  .catch((error) => {
     console.error(error);
   });
 
-spark.rooms
+// Filter for "My First Room" from the last 10 rooms
+teams.rooms
   .list({
     max: 10
   })
-  .then(function(rooms) {
-    var room = rooms.items.filter(function(room) {
-      return room.title === 'My First Room!';
-    })[0];
+  .then((rooms) => {
+    // Destructure room properties for it's id (aliased to roomId) and title
+    const { id: roomId, title } = rooms.items.filter(
+      room => room.title === 'My First Room!'
+    )[0];
 
-    spark.messages.create({
-      text: 'Hello World!',
-      roomId: room.id
+    // Post message "Hello World!" to "My First Room!"
+    teams.messages.create({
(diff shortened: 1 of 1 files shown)


taymoork2:
`docs/_posts/guides/9999-01-01-getting-started.md` & `documentation/ciscospark.md` both need to be updated because they both include `ciscospark/env`

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
-      ciscospark.memberships.create({
-        roomId: room.id,
-        personEmail: `bob@example.com`
-      }),
-    ])
-      .then(() => ciscospark.messages.create({
-        markdown: `**Hi Everyone**`,
-        roomId: room.id
-      }));
+const spark = ciscospark.init({
```
mattnorris: Maybe we should take this opportunity to call this `const teams` instead. Or will that be too confusing with `ciscospark`. What do you think?

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+  credentials: {
+    access_token: process.env.CISCOSPARK_ACCESS_TOKEN
+  }
+});
+
+spark.rooms
```
mattnorris: Why the change from `Promise.all`?

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+      });
+  })
+  .catch(function(error) {
+    console.error(error);
+  });
+
 ```
 
-### Browsers
+### _A note on Browser usage_
```
mattnorris: Lowercase "browser"

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+
+We do not provide a pre-built version of the SDK that includes a `window.ciscospark`.
```
mattnorris: We do not provide a built version of the SDK that includes `window.ciscospark`.

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+If you're already using a bundler (like [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org/)) you can simply import/require the package and use the above snippet and assign the initialized `spark` variable to `window.ciscospark`.
```
mattnorris: If you're already using a bundler like [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org/), you can simply import/require the package, initialize it, and assign it to `window.ciscospark`.

INCLUDE A SMALL SNIPPET HERE SHOWING WHAT YOU JUST DESCRIBED

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+In-browser usage is pretty much the same as Node.js usage, with the addition of handling the user authentication flow for you. See the guide on the [docs site](https://webex.github.io/spark-js-sdk/guides/browsers/) for more information.
```
mattnorris: In-browser usage is almost the same as Node.js, but it handles the user authentication flow for you. See the [browser guide](https://webex.github.io/spark-js-sdk/guides/browsers/) for more information.

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+
+For a quick example on how to use the SDK on a website, we'll be using the [Parcel Bundler](https://parceljs.org/). For any more information and questions on how to use Parcel, please head to their [website](https://parceljs.org/).
```
mattnorris: We will use [Parcel](https://parceljs.org/) to bundle the SDK for a website.

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+
+1. Put the above javascript snippet into a `index.js`
+  a. If you'd like to have `window.ciscospark` available, change the init line to `const spark = window.ciscospark = ciscospark.init`
```
mattnorris: I would rewrite this to just show what the code would look like. Instead of having them all on the same line, break it out into two and use a comment on the bottom line that just says something like, `// Make available to the window.`

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+2. Create a `index.html` with the snippet below
```
mattnorris: Remove "a".

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+3. Run `parcel index.html` in your terminal
+4. Go to [http://localhost:1234](http://localhost:1234) and open the developer console to see the output
```
mattnorris: Put periods at the end of these sentences.

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
+ </body>
+</html>
+```
 
-We do not provide a pre-built version of `ciscospark`.
+#### `ciscospark/env`
 
-If you've already got a commonjs or es6 build process in place, you can simply
-use `const ciscospark = require('ciscospark')`.
+`ciscospark/env` is primarily used for the Node.js environment and will continue to be supported, however we suggest that you import/require `ciscospark`.
```
mattnorris: Let's not point people to `ciscospark/env`. I'd rather do things explicitly. 

It may be more appropriate to include a note elsewhere (maybe another doc or .md file) about it and that it will no longer be supported.

README.md
```diff
@@ -37,59 +37,101 @@ To use the SDK, you will need Cisco Webex credentials. If you do not already hav
...
-        personEmail: `bob@example.com`
-      }),
-    ])
-      .then(() => ciscospark.messages.create({
-        markdown: `**Hi Everyone**`,
-        roomId: room.id
-      }));
+const spark = ciscospark.init({
+  credentials: {
+    access_token: process.env.CISCOSPARK_ACCESS_TOKEN
```
adamweeks: Maybe we can put a comment here that the access token doesn't have to come from an env var with this setup.

README.md
```diff
@@ -32,64 +32,124 @@ npm install --save ciscospark
 
 ## Usage
 
+#### _[Still using `ciscospark/env`?](documentation/ciscospark.md#nodejs)_
```
adamweeks: Can you move this to the bottom of the usage section? Let's hide it a bit more.

README.md
```diff
@@ -32,64 +32,124 @@ npm install --save ciscospark
...
-      }),
-    ])
-      .then(() => ciscospark.messages.create({
-        markdown: `**Hi Everyone**`,
-        roomId: room.id
-      }));
-  });
+const teams = ciscospark.init({
+  credentials: {
+    access_token: process.env.ACCESS_TOKEN
```
adamweeks: Let's go with the usage below from your browser usage example. 

```suggestion
    access_token: <your webex teams access token>
```

PR #1017, opened 2018-11-09
refactor(plugin-wdm): change data structure

stanjiawang:
# Pull Request Template

## Description

Changed wdm service ampersand data structure since the server side has changed the map of services. registration.services obsoleted and will be deprecated soon in the response message when registering to wdm. We should use registration.serviceHostMap.serviceLinks as the new map and use service name as key to lookup uri in serviceLinks and then use host name as key to lookup in the registration.serviceHostMap.hostCatalog.

server side specs: https://wiki.cisco.com/display/WX2/Spark+Client+Behavior
wdm response: https://sqbu-github.cisco.com/gist/sgaikwad/879d8506fef720ebdf3f52b876190d8e

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-38919

## Type of Change
refactor

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
@@ -7,7 +7,8 @@ import {inBrowser} from '@ciscospark/common';
 export default {
   device: {
     preDiscoveryServices: {
-      metricsServiceUrl: process.env.METRICS_SERVICE_URL || 'https://metrics-a.wbx2.com/metrics/api/v1'
+      metricsServiceUrl: process.env.METRICS_SERVICE_URL || 'https://metrics-a.wbx2.com/metrics/api/v1',
+      metrics: process.env.METRICS_SERVICE_URL || 'https://metrics-a.wbx2.com/metrics/api/v1'
     }
   },
   metrics: {
--- a/packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
@@ -320,10 +320,6 @@ const Device = SparkPlugin.extend({
 
   _getServiceUrl(target, service) {
     /* istanbul ignore if */
-    if (!target) {
-      return Promise.reject(new Error('`target` is a required parameter'));
-    }
-
     if (!service) {
       return Promise.reject(new Error('`service` is a required parameter'));
     }
@@ -335,6 +331,11 @@ const Device = SparkPlugin.extend({
         return Promise.resolve(s.url);
       }
     }
+
+    if (!target) {
+      return Promise.reject(new Error('`target` is a required parameter'));
+    }
+
     return Promise.resolve(target[`${service}ServiceUrl`]);
   },
 
@@ -373,7 +374,7 @@ const Device = SparkPlugin.extend({
   },
 
   @oneFlight
-  @waitForValue('@')
+    @waitForValue('@')
   refresh() {
     this.logger.info('device: refreshing');
 
@@ -406,7 +407,7 @@ const Device = SparkPlugin.extend({
   },
 
   @oneFlight
-  @waitForValue('@')
+    @waitForValue('@')
   register() {
     /* eslint no-invalid-this: [0] */
     this.logger.info('device: registering');
@@ -431,7 +432,7 @@ const Device = SparkPlugin.extend({
   },
 
   @oneFlight
-  @waitForValue('@')
+    @waitForValue('@')
   unregister() {
     this.logger.info('device: unregistering');
 
--- a/packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
@@ -4,7 +4,7 @@
 
 import {oneFlight} from '@ciscospark/common';
 import {safeSetTimeout} from '@ciscospark/common-timers';
-import {omit} from 'lodash';
+import {omit, find} from 'lodash';
 import util from 'util';
 import FeaturesModel from './features-model';
 import ServiceCollection from './service-collection';
@@ -78,6 +78,7 @@ const Device = SparkPlugin.extend({
     reportingSiteDesc: 'string',
     reportingSiteUrl: 'string',
     searchEncryptionKeyUrl: 'string',
+    // services is obsoleted and will be deprecated
     services: {
       // Even though @jodykstr will tell you the docs claim you don't need to
       // initialize `object` properties, the docs lie.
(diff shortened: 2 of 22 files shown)


packages/node_modules/@webex/plugin-meetings/src/ciscospark.js
```diff
@@ -16,7 +16,7 @@ CiscoSpark.version = PACKAGE_VERSION;
 
 CiscoSpark.init = function init(attrs = {}) {
   const config = merge({}, {
-    hydraServiceUrl: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1',
+    hydra: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1',
```
andrewholsted: Changing these config value names is going to be a breaking change for third party developers if they overwrite or define these urls. @mattnorris do you know the likelihood of that?

@stanjiawang is there any way to make this backward compatible with a deprecation warning?

adamweeks: Agreed. `hydraServiceUrl` is used in a few different plugins and probably shouldn't be changed unless there is a real need to.

adamweeks:
Could we get a bit better description on the commits? From a documentation standpoint, I have no idea what `fix UT` means.

packages/node_modules/@webex/plugin-meetings/src/ciscospark.js
mattnorris: The contrary to this is altering URLs for integration, which we just went through with BTS. However, if we ensure that Hydra's URL is changed depending on another environment variable, it will work. 

Do we know the cases where Hydra will have to be integration?

packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
```diff
@@ -7,7 +7,7 @@ import {inBrowser} from '@ciscospark/common';
 export default {
   device: {
     preDiscoveryServices: {
-      metricsServiceUrl: process.env.METRICS_SERVICE_URL || 'https://metrics-a.wbx2.com/metrics/api/v1'
+      metrics: process.env.METRICS_SERVICE_URL || 'https://metrics-a.wbx2.com/metrics/api/v1'
```
mattnorris: I'd rather not change these from other plugins. If we decide the `ServiceUrl` suffix is too verbose, we should change it in the repo in another PR.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
```diff
@@ -229,15 +229,18 @@ const Device = SparkPlugin.extend({
 
   @waitForValue('@')
   isService(service) {
-    return this._isService(this.services, service)
+    if (service === 'idbroker') {
```
mattnorris: Why is `idbroker` excluded?

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
```diff
@@ -281,8 +283,9 @@ const Device = SparkPlugin.extend({
       return Promise.reject(new Error('`uri` is a required parameter'));
     }
 
-    for (const value of Object.values(target)) {
-      if (value && uri.startsWith(value)) {
+    for (const key of Object.keys(target)) {
+      const serviceUrl = target[key];
+      if (key !== 'idbroker' && uri.startsWith(serviceUrl)) {
```
mattnorris: Same question as above. I think adding a comment as to why `idbroker` is special will clarify.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/interceptors/device-url.js
```diff
@@ -29,7 +29,7 @@ export default class DeviceUrlInterceptor extends Interceptor {
 
     return Promise.resolve(options.service || this.spark.internal.device.determineService(options.uri))
       .then((service) => {
-        if (service && service !== 'oauth' && service !== 'saml') {
+        if (service && service !== 'oauth' && service !== 'saml' && service !== 'idbroker') {
```
mattnorris: idbroker

packages/node_modules/@ciscospark/internal-plugin-wdm/test/unit/spec/device.js
```diff
@@ -42,9 +42,9 @@ describe('plugin-wdm', () => {
 
       describe('#_updateServiceCatalog', () => {
         it('refreshes the device if serviceHostMap is not defined and ha is enabled', () => {
-          device.services = {
-            validServiceUrl: 'http://example.com/valid-service',
-            anotherServiceUrl: 'http://another.com/another-service'
+          device.serviceHostMap.serviceLinks = {
+            valid: 'http://example.com/valid-service',
```
mattnorris: I prefer the less ambiguous `validServiceUrl`. Please see my comments at the beginning.

packages/node_modules/@webex/plugin-meetings/src/ciscospark.js
stanjiawang: @andrewholsted @adamweeks @mattnorris Thanks a lot for your comments! As I mentioned in the PR description, the key to the service URL map has been changed on the server side. The res.services(which keys are serviceName + "ServiceUrl") is obsoleted and will be deprecated soon in the response message when registering to wdm and we should use res.serviceHostMap.serviceLinks(which keys are just the service name without the redundant "ServiceUrl").

https://sqbu-github.cisco.com/gist/sgaikwad/879d8506fef720ebdf3f52b876190d8e

The obsoleted map below:
screen shot 2018-11-14 at 11 37 18 am

The new map we should use:
screen shot 2018-11-14 at 11 35 57 am

packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
stanjiawang: @mattnorris Thanks for your comment. I would rather not change this suffix either, haha. The reason I need to change this service name is that service side changed the key(remove the suffix ServiceUrl) of the map and I have to change some functions in wdm module to make it compatible with the new map. Since the metricsServiceUrl  in preDiscoveryServices in hard coded, I think we could probably change it to make it consistent with the message on server side. Any thought on this?

packages/node_modules/@ciscospark/internal-plugin-wdm/test/unit/spec/device.js
stanjiawang: @mattnorris Same reason above, make it compatible and consistent with the new map in the message received from the server side.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
stanjiawang: @mattnorris Sorry I didn't mention this in the description. Thanks for bringing this up.

https://sqbu-github.cisco.com/gist/sgaikwad/879d8506fef720ebdf3f52b876190d8e

If you compare the old service map with the new service map carefully, you could find that there is NO idbroker in the old service map and you can see the idbroker service in the new map. 

This isService is called in function requiresCredentials in file:
spark-js-sdk/blob/master/packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js which decide whether we should add auth to the message header when we connect to idbroker (CI). So the idbroker is a special case for this function and it should not be considered as a service.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/device/device.js
stanjiawang: Same reason above.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/interceptors/device-url.js
stanjiawang: Same reason above.

stanjiawang:
> Could we get a bit better description on the commits? From a documentation standpoint, I have no idea what `fix UT` means.

@adamweeks 
Sorry for the ambiguous description on the commits. I will squash the commits and re-write the description when we are good with the changes. Thanks a lot for your suggestion.

packages/node_modules/@ciscospark/internal-plugin-metrics/src/config.js
mattnorris: @stanjiawang  Thank you for clarifying. I missed this point in the description.

stanjiawang: @mattnorris Sure! I will try to replace all the service name without "ServiceUrl" in the repo. Thanks!

stanjiawang:
@mattnorris Yes! I put the source code behind the web-ha-messaging feature toggle so that the code could work either with or without the feature toggle on. Also, it will have less impact on the current unit test and integration test. Let me know if you have any concerns. Thanks!

mattnorris:
Not closed automatically

PR #1034, opened 2018-11-20
[WIP] test(samples): react conversion for meetings sample

taymoork2:
Some problems at the moment 

Fails connected to Locus on Chrome (getting a 401)
Can't debug `app.react.js` on Firefox for firefox reasons (it can't find `app.react.js` in the debugger)
When running `leave()` to disconnect, the camera and mic still run


packages/node_modules/samples/browser-meetings/react/app.js
```diff
@@ -0,0 +1,332 @@
...
+          // reported it
+          throw err;
+        });
+    }
+  };
+
+  connect = (e) => {
+    e.preventDefault();
+
+    spark.meetings.create(this.state.connectAddress).then(async (meeting) => {
```
andrewholsted: This doesn't need to be async since there is no `await`

mattnorris:
I think this has been surpassed by other approaches.

PR #1035, opened 2018-11-20
fix(search): remove query check

phatv:
# Pull Request Template

## Description

Remove query check from search request.

This allows us to search for all items in a convo rather than having to match with a string. For example, I needed a search that returns all files in a conversation. 

Fixes # (issue)

## Type of Change

## Test Coverage

- [x] Made a search request without a query in the body. Received a successful response with all results.
- [x] Made a search request with a query in the body. Received a successful response that returned results with the query. 

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/internal-plugin-search/src/search.js
+++ b/packages/node_modules/@ciscospark/internal-plugin-search/src/search.js
@@ -44,10 +44,6 @@ const Search = SparkPlugin.extend({
     /* eslint max-nested-callbacks: [0] */
     options = options || {};
 
-    if (!options.query) {
-      return Promise.resolve([]);
-    }
-
     let promise = Promise.resolve();
     if (!this.spark.internal.device.searchEncryptionKeyUrl) {
       promise = this.bindSearchKey();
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/internal-plugin-search/src/search.js
```diff
@@ -44,10 +44,6 @@ const Search = SparkPlugin.extend({
     /* eslint max-nested-callbacks: [0] */
     options = options || {};
 
-    if (!options.query) {
-      return Promise.resolve([]);
-    }
```
andrewholsted: What's the expected use case here @phatv? Is that the only way the files search works with a given convo? Would it be better to say `!options.query || <other param`?

phatv: Searching for all items. Query is the search string and I don't want to specify that because we want all files. The other params I'm using are sharedIn, type, and limit. None of these are required by all searches and can be left out so checking for another specific param doesn't seem useful.

phatv: I could check if the object is not empty so we know that it is searching for something. Like `Object.keys(test).length === 0`?

andrewholsted: I meant will this break any existing search functionality in the client (not your new code)

adamweeks:
@phatv could you fill out the PR template so we know why these changes are necessary?

PR #1037, opened 2018-11-27
Mute v2

chrisadubois:


packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
...
+};
+
+Media.localAudioTracks = (audio, stream) => {
+  let audioTrack;
+  if (stream) {
+    if (stream.getAudioTracks) {
+      [audioTrack] = stream.getAudioTracks();
+    }
+  }
+  if (audioTrack) {
```
andrewholsted: Can all this logic move inside the previous if block?

```
if (stream.getAudioTracks) {
      const audioTrack = stream.getAudioTracks();
      if(audioTrack) {
        audioTrack.enabled = !audio.mute
        return true;
       }
    }
```

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,99 @@
...
+      data: {
+        muted: undefined,
+        self: undefined
+      },
+      methods: {
+        getMuted() {
+          return this.muted;
+        },
+        getSelf() {
+          return this.self;
```
andrewholsted: What is `self` here? I see it's set with `audio.self` but not sure what that is

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,99 @@
...
+        },
+        onAfterToggle(transition, audio, meeting) {
+          if (audio.self) {
+            this.self = true;
+          }
+          if (transition.from !== transition.to) {
+            MeetingUtil.remoteUpdateAudioVideo(audio, {mute: meeting.video.muted}, meeting)
+              .then((resolution) => {
+                Media.localTracks(audio, {mute: meeting.video.muted}, meeting.localMediaStream);
+                this.muted = audio.mute;
```
andrewholsted: What happens if `Media.localTracks` fails to mute here? Will out state get out of sync?

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,99 @@
...
+        },
+        plugins: [
+          new StateMachineHistory({max: 2})
+        ]
+      }
+    });
+  }
+};
+
+export default AudioStateMachine;
```
andrewholsted: This feels really complex for a simple binary property. @mattnorris?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -188,6 +190,22 @@ export default class Meeting extends StatelessSparkPlugin {
...
 
+  getAudioMuted() {
+    return this.audio.getMuted();
+  }
+
+  getAudioSelf() {
+    return this.audio.getSelf();
+  }
+
+  getVideoMuted() {
```
andrewholsted: Should this be `isVideoMuted()`? Does it return a boolean or something else? Same goes for audio above and I'm not sure what self is

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -213,10 +231,10 @@ export default class Meeting extends StatelessSparkPlugin {
   setRemoteStreams(pc) {
     // Receiving remote media streams
     pc.ontrack = (event) => {
-      [this.mediaStream, this.localMediaStream] = event.streams;
-
-      this.trigger('media:ready', {
-        type: 'remote',
+      [this.mediaStream] = event.streams;
+      document.getElementById('remotevideo').srcObject = this.mediaStream;
```
andrewholsted: Please remove. This isn't necessary if we handle the events in the app itself as a developer would

packages/node_modules/@webex/plugin-meetings/src/constants.js
```diff
@@ -64,6 +65,39 @@ export const MEETING_STATE_MACHINE = {
     ERROR: 'ERROR'
   }
 };
+export const EVENT_TRIGGERS = {
+  MEDIA_AUDIO: 'media:audio',
```
andrewholsted: Can we either add a `:` to the key here or change the value to match? It will be incongruent with other things that use underscores.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
...
+  if (mediaId) {
+    return [{
+      localSdp: JSON.stringify({
+        audioMuted: audio.mute,
+        videoMuted: video.mute
+      }),
+      mediaId
+    }];
+  }
+  return {};
```
andrewholsted: We shouldn't return two different types `[]` vs `{}`

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,99 @@
...
+
+const AudioStateMachine = {
+  create() {
+    return new StateMachine({
+      transitions: [
+        {
+          name: MEETING_AUDIO_STATE_MACHINE.TRANSITIONS.INIT,
+          from: '*',
+          to: function (audio, meeting) {
+            return handleTransition(this.state, audio, meeting);
```
andrewholsted: We don't actually use this.state (value) in the method above. Pedantic, but it makes more sense to do `this.state = handleTransition(audio, meeting)` and then `return this.state`; We're just overwriting the value in the helper method which isn't intuitive when looking at the method.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -253,6 +272,47 @@ export default class Meeting extends StatelessSparkPlugin {
...
 
+  startAudioVideo(mediaDirection) {
+    this.audio.init({mute: !mediaDirection.sendAudio, self: true}, this);
+    this.video.init({mute: !mediaDirection.sendVideo, self: true}, this);
+  }
+
+  muteAudio() {
+    const options = {};
+    options.mute = true;
+    options.self = true;
```
andrewholsted: Is there a reason you broke these out instead of writing

```
this.toggleAudio({
  mute: true,
  self: true
})
```

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -253,6 +272,47 @@ export default class Meeting extends StatelessSparkPlugin {
...
+  }
+
+  muteVideo() {
+    const options = {};
+    options.mute = true;
+    options.self = true;
+    this.toggleVideo(options);
+  }
+
+  unmuteVideo() {
```
andrewholsted: I like the syntax that is exposed to the user here.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -390,13 +450,7 @@ export default class Meeting extends StatelessSparkPlugin {
...
    */
   leave() {
-    return this.meetingRequest.leaveMeeting({
-      locusUrl: this.locusUrl,
-      selfId: this.selfId,
-      correlationId: this.correlationId,
-      resourceId: null,
-      deviceUrl: this.deviceUrl
-    });
+    return MeetingUtil.leaveMeeting(this);
```
andrewholsted: I'd rather be explicit about what options we are passing here instead of passing the entire class. We can discuss at standup if you disagree.

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -18,10 +22,39 @@ MeetingUtil.parseLocus = (response) => {
...
+    }
+    return meeting.meetingRequest.remoteAudioVideoToggle({
+      locusUrl: meeting.locusUrl,
+      selfId: meeting.selfId,
+      localMedias,
+      deviceUrl: meeting.deviceUrl,
+      correlationId: meeting.correlationId
+    });
+  }
+  return Promise.reject(new Error('You need a meeting object.'));
```
andrewholsted: Please move this to the top of the method and return early. That removes a level of indentation and you don't have to go back up in the method to see what case hits here.

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -18,10 +22,39 @@ MeetingUtil.parseLocus = (response) => {
...
+};
+
+MeetingUtil.leaveMeeting = (meeting) => {
+  if (!meeting) {
+    return Promise.reject(new Error('You need a meeting object.'));
+  }
```
andrewholsted: yayyy. like this.

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -40,7 +73,7 @@ MeetingUtil.joinMeeting = (meeting) => {
         throw err;
       });
   }
-  return Promise.reject();
+  return Promise.reject(new Error('You need a meeting object.'));
```
andrewholsted: Same comment as above

packages/node_modules/@webex/plugin-meetings/src/meeting/video.js
```diff
@@ -0,0 +1,99 @@
...
+        },
+        plugins: [
+          new StateMachineHistory({max: 2})
+        ]
+      }
+    });
+  }
+};
+
+export default VideoStateMachine;
```
andrewholsted: Same comments from the audio file.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
  */
 const Media = {};
 
+Media.localTracks = (audio, video, stream) => {
+  Media.localAudioTracks(audio, stream);
+  Media.localVideoTracks(video, stream);
+};
+
+Media.localAudioTracks = (audio, stream) => {
```
mattnorris: Maybe name this `hasLocalAudioTracks`? From the current name it looks like it should return the actual tracks (an array or object) but we are returning a boolean after some manipulation.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
  */
 const Media = {};
 
+Media.localTracks = (audio, video, stream) => {
```
mattnorris: This calls other functions that return booleans, yet it doesn't return anything. Is its purpose to manipulate data further down? It should probably have a verb or return a boolean (see my note below).

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
...
+    }
+  }
+  if (videoTrack) {
+    videoTrack.enabled = !video.mute;
+    return true;
+  }
+  return false;
+};
+
+Media.generateLocalMedias = (mediaId, audio, video) => {
```
mattnorris: Do these need to be `async`?

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,99 @@
...
+        onAfterInit(transition, audio, meeting) {
+          this.muted = audio.mute;
+          this.self = audio.self;
+          meeting.trigger(EVENT_TRIGGERS.MEDIA_AUDIO, {
+            type: EVENT_TYPES.local,
+            status: this.state,
+            muted: this.muted,
+            action: MEETING_AUDIO_STATE_MACHINE.TRANSITIONS.INIT
+          });
+          console.log(`audio:state#onAfterInit->${transition.event} fired! State changed from '${transition.from}' to '${transition.to}' with transition '${transition.transition}''.`);
```
mattnorris: `this.logger.info` rather than `console.log`

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
mattnorris: @andrewholsted my understanding from talking with @chrisadubois is that it will allow us to recover from a myriad things happening as these complex calls occur. That's oversimplifying it, and we can discuss, but I believe this exposes the life cycles Chris wants to be able to reverse history and recover from errors. 

@chrisadubois can you elaborate?

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
chrisadubois: yeah may or may not need it, but it's to determine whether or not the action was remote or by self

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
chrisadubois: Media.localTracks can't really "fail" in the same way as a request could fail, but we can add a try catch block

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
chrisadubois: that's exactly correct matt, as discussed from the last meeting.

And the properties this.audio.muted will always have a source of truth. for details one can dig into the state

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
chrisadubois: I did, i wonder why it didn't get included

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: No I don't disagree, but we just need to get clear there and do it everywhere

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: sure, pick one 'is' or 'get'

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
chrisadubois: so weird, i made these changes last night, must have not committed

packages/node_modules/@webex/plugin-meetings/src/media/index.js
chrisadubois: yep, teehee, stole this one direct from web client

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
andrewholsted: If it can't fail then it's not an issue. Was just trying to understand where we would fail/rollback

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
andrewholsted: `is` please

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
chrisadubois: following state machine standard here where the to function returns what value should be set because this.state gets automatically set from the to: function

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: ?? lets discuss tomorrow during the sanity check -- want to have the one that got set up for the guest client

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: nope -- changed

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: done

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: done

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: 👍

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
chrisadubois: done for the rest

packages/node_modules/@webex/plugin-meetings/src/meeting/video.js
chrisadubois: changed to promise based, and removed history necessity

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
...
+Media.setLocalTracks = (audio, video, stream) => {
+  Media.setLocalAudioTracks(audio, stream);
+  Media.setLocalVideoTracks(video, stream);
+};
+
+Media.setLocalAudioTracks = (audio, stream) => {
+  let audioTrack;
+  if (stream) {
+    if (stream.getAudioTracks) {
+      [audioTrack] = stream.getAudioTracks();
```
andrewholsted: You don't have to declare audioTrack outside of the if block any more

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -15,6 +15,52 @@ import {
...
+        audioTrack.enabled = !audio.mute;
+        return true;
+      }
+    }
+  }
+  return false;
+};
+
+Media.setLocalVideoTracks = (video, stream) => {
+  let videoTrack;
```
andrewholsted: same comment as above

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
```diff
@@ -0,0 +1,101 @@
+import StateMachine from 'javascript-state-machine';
+import StateMachineHistory from 'javascript-state-machine/lib/history';
+
+import Media from '../media';
+import MeetingUtil from '../meeting/util';
+import {MEETING_AUDIO_STATE_MACHINE, EVENT_TRIGGERS, EVENT_TYPES} from '../constants';
+
+const handleTransition = (value, audio, meeting) => {
+  if (audio.mute && audio.self) {
+    value = MEETING_AUDIO_STATE_MACHINE.STATES.MUTE_SELF;
```
andrewholsted: value is `this.state`. Is it intentional to change that value in addition to returning it from the function? Totally may be, I just figured it would be one or the other.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
andrewholsted: Sure check `/samples/browser-meetings/app.js` line 118 or so. It listens for the event and does the same thing you are doing on 235

packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js
chrisadubois: afaik, javascript does pass by value for strings? since this.state is a string type, value is a copy of the value stored in string. it then returns and alters state internal javascript-state-machine from to: function(...)
short summary, yes, it's intentional to get a copy of it, know what it is before it would've been changed, change it *if* it should be, and return the result to set what state should be, in some cases, there is no change, in which case no transition lifecycle events are triggered

PR #1050, opened 2018-12-14
Syntax check and sanitize access token

taymoork2:
## Description

Added some syntax checking when using an [access] token during the init process to minimize errors.
Check for extra spaces and/or lack of spaces when adding `Bearer` to the access token property value, remove the potential errors caused by syntax mishaps, suggest a cleaned version and also suggest that Bearer isn't necessary (both logged to the console as warn and info respectively).

Was tested using browser samples.
Tested have not been written up for this (yet)

Fixes [SPARK-34900](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-34900)

## Type of Change

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/spark-core/src/spark-core.js
+++ b/packages/node_modules/@ciscospark/spark-core/src/spark-core.js
@@ -181,37 +181,6 @@ const SparkCore = AmpState.extend({
   },
 
   /**
-   *
-   * Check if access token is correctly formated and correct if it's not
-   * Warn user if token string has errors in it
-   * @param {string} token
-   * @returns {string}
-   */
-  bearerValidator(token) {
-    if (token.includes('Bearer') && token.split(' ').length - 1 === 0) {
-      console.warn(
-        "Your access token does not have a space between 'Bearer' and the token, please add a space to it or replace it with this already fixed version:\n\n",
-        token.replace('Bearer', 'Bearer ').replace(/\s+/g, ' ')
-      );
-      console.info("Tip: You don't need to add 'Bearer' to the access_token field. The token by itself is fine");
-
-      return token.replace('Bearer', 'Bearer ').replace(/\s+/g, ' ');
-    }
-    // Allow elseIf return
-    // eslint-disable-next-line  no-else-return
-    else if (token.split(' ').length - 1 > 1) {
-      console.warn(
-        `Your access token has ${token.split(' ').length - 2} too many spaces, please use this format:\n\n`,
-        token.replace(/\s+/g, ' ')
-      );
-      console.info("Tip: You don't need to add 'Bearer' to the access_token field, the token by itself is fine");
-      return token.replace(/\s+/g, ' ');
-    }
-
-    return token.replace(/\s+/g, ' '); // Clean it anyway (just in case)
-  },
-
-  /**
    * @instance
    * @memberof SparkCore
    * @param {[type]} args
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/spark-core/src/spark-core.js
```diff
@@ -120,6 +122,24 @@ const SparkCore = AmpState.extend({
         });
 
       if (typeof get(attrs, 'credentials.access_token') === 'string') {
+        if (get(attrs, 'credentials.access_token').includes('Bearer') && get(attrs, 'credentials.access_token').split(' ').length - 1 === 0) {
+          this.logger.warn(
+            "Your access token does not have a space between 'Bearer' and the token, please add a space to it or replace it with this already fixed version\n\n",
+            get(attrs, 'credentials.access_token').trim().replace('Bearer', 'Bearer ').replace(/\s+/g, ' ')
+          );
+          this.logger.info("Tip: You don't need to add 'Bearer' to the access_token field, the token by itself is fine");
```
mattnorris: "... field. The token..."

packages/node_modules/@ciscospark/spark-core/src/spark-core.js
```diff
@@ -120,6 +122,24 @@ const SparkCore = AmpState.extend({
...
+
+          set(attrs, 'credentials.access_token', get(attrs, 'credentials.access_token').trim().replace('Bearer', 'Bearer ').replace(/\s+/g, ' '));
+        }
+        else if (get(attrs, 'credentials.access_token').split(' ').length - 1 > 1) {
+          this.logger.warn(
+            `Your access token has one too many spaces (${get(attrs, 'credentials.access_token').split(' ').length - 2}), please clean it up or replace it with this already cleaned version\n\n`,
```
mattnorris: - Instead of "one too many", should we just check for more than one and correct it, then say "too many"? 
- ". Please use this format: \n\n"

packages/node_modules/@ciscospark/spark-core/src/spark-core.js
```diff
@@ -120,6 +122,24 @@ const SparkCore = AmpState.extend({
         });
 
       if (typeof get(attrs, 'credentials.access_token') === 'string') {
+        if (get(attrs, 'credentials.access_token').includes('Bearer') && get(attrs, 'credentials.access_token').split(' ').length - 1 === 0) {
```
mattnorris: I think this ability is a perfect candidate to break out into the appropriate functions and unit test them.

taymoork2: The unit test for this is not that simple.
Would have written tests for this but it's this weird reduce function.

But agree with that this should be broken out into its own function

mattnorris: I think you could encapsulate the string replace logic and test different scenarios against it.

taymoork2: Alright added those tests

packages/node_modules/@ciscospark/spark-core/src/spark-core.js
```diff
@@ -396,6 +365,37 @@ const SparkCore = AmpState.extend({
...
+  /**
+   *
+   * Check if access token is correctly formated and correct if it's not
+   * Warn user if token string has errors in it
+   * @param {string} token
+   * @returns {string}
+   */
+  bearerValidator(token) {
+    if (token.includes('Bearer') && token.split(' ').length - 1 === 0) {
+      console.warn(
```
mattnorris: Instead of `console`, use the `logger`.

packages/node_modules/@ciscospark/spark-core/test/unit/spec/spark-core.js
```diff
@@ -99,6 +99,85 @@ describe('Spark', () => {
       });
   });
 
+  describe('initializes with Bearer Token', () => {
```
mattnorris: This structure could be simplified (because you're verifying the same information) by looping through an array of objects `[{access_token: 'Bearer1234', testName: 'initializes...'}, {access_token: 'Bearer 1234', ...}]`.

packages/node_modules/@ciscospark/spark-core/src/spark-core.js
taymoork2: `this.logger` is not in scope because bearerValidator gets called during the constructor and it seems that the logger gets added to the initialized variable/object somewhere else (at least not in the `spark-core` file)

mattnorris:
Not automatically closed

PR #1066, opened 2019-01-02
fix(meetings): remove scheduled meetings and use just in samples

chrisadubois:


packages/node_modules/@webex/plugin-meetings/sample/app.js
```diff
@@ -8,6 +8,8 @@
 /* eslint-disable no-console */
 /* eslint-disable require-jsdoc */
 
+import '@ciscospark/internal-plugin-calendar';
```
andrewholsted: Does this work? I didn't think we could import anything inside of `/sample/` but maybe that was the old way

chrisadubois: it seems to work, yeah i tested it

chrisadubois:
does it make sense to do the removal? yeah it will change joining a scheduled meeting because you will have to extract it's sipURi or convoId and join via those methods

packages/node_modules/@webex/plugin-meetings/sample/app.js
andrewholsted: ok. For consistency, I think we should move it to ciscospark.js where we require other plugins. Isn't that what @arun3528 did for device search?

chrisadubois:
ok done

chrisadubois:
@mattnorris am I not approved to merge PRs into master?

mattnorris:
@chrisadubois you have to submit your changes through gauntlet after receiving approval from a DX core team member (me, Adam, Taymoor, etc.). Since we're merging into master now we need to double-check its impact. After that: 

```
git push gauntlet HEAD:master
```

PR #1074, opened 2019-01-14
participant list, invite guest, admit guest, meeting link, 1:1 incoming call (answer/reject), alerted call, and ringing events

chrisadubois:
ready to review


packages/node_modules/@webex/plugin-meetings/src/locus-events/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * @param {Object} data
+   * @returns {undefined}
+   * @memberof LocusEvents
+   */
+  locusEvent(data) {
+    const {eventType} = data;
+    console.log('Mercury-Locus-Event', eventType);
+    if (eventType === LOCUSEVENT.MESSAGE_ROAP) {
```
mattnorris: Why does a ROAP message simply return?

packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
```diff
@@ -49,7 +49,7 @@ export default class MeetingInfo extends StatelessSparkPlugin {
    * Fetches meeting info from the server
    * @param {String} destination one of many different types of destinations to look up info for
    * @param {String} type optional type to match up with the destination value
-   * @returns {Object} returns a meeting info object
+   * @returns {Promise} returns a meeting info object
```
mattnorris: 👍

packages/node_modules/@webex/plugin-meetings/src/meeting-info/request.js
```diff
@@ -16,6 +16,7 @@ export default class MeetingInfoRequest extends StatelessSparkPlugin {
    * type can also be specified as an option and be of the list SIP_URI,MEETING_ID,LOCUS_ID,PERSONAL_ROOM,MEETING_LINK,ONE_ON_ONE,MEDIA_SIP_URI,CONVERSATION_URL,TEMP_SIP_URI
    * with the desination matching
    * @returns {Promise} returns a promise that resolves/rejects the result of the request
+   * @throws {Error} if the options are not valid and complete
```
mattnorris: 👍

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -27,19 +26,15 @@ const MeetingInfoUtil = {};
  * @returns {String} returns a sipURL
```
mattnorris: Update this to `boolean` if you're now returning that.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -63,7 +58,12 @@ MeetingInfoUtil.generateOptions = (from) => {
     };
   }
   const options = {};
-  if (MeetingInfoUtil.isSipUri(destination)) {
+  // TODO: convert to sip URI?
```
mattnorris: Is there a particular "default" format you find is most useful to pass around internally?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -46,10 +49,54 @@ export default class Meeting extends StatelessSparkPlugin {
...
+  admit(participantIds) {
+    return this.participants.admitParticipants(participantIds);
+  }
+
+  parseLocus(locus) {
+    // parse all the object and attach it
+    this.locusUrl = locus.url;
+    this.type = this.getLocusType(locus.info.locusTags);
+    this.partner = locus.participants.find((p) => (locus.self && p.identity !== locus.self.identity)
+      && (locus.participants.length <= 2 || (p.type === 'USER' && !p.removed))) || this.partner;
```
mattnorris: Is this `USER` a part of the constant file? If it isn't, no need to add. But if it is and is used elsewhere, it may be good to use here too.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -137,24 +192,37 @@ export default class Meeting extends StatelessSparkPlugin {
...
+    this.locusUrl = locus.locusUrl || locus.url;
     this.locusId = locus.locusId;
     this.selfId = locus.selfId;
     this.mediaId = locus.mediaId;
+    this.participants.set(mtgLocus);
+  }
+
+  getLocusType(locusTags) {
+    // Parse all the value and get the type of meeting
+    if (locusTags.find((tag) => tag === 'ONE_ON_ONE')) {
```
mattnorris: Same question on the constant as above.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -137,24 +192,37 @@ export default class Meeting extends StatelessSparkPlugin {
...
+      return 'ONE_ON_ONE';
+    }
+    return null;
```
mattnorris: What's the advantage of returning `null` here rather than `undefined` or an empty string?

packages/node_modules/@webex/plugin-meetings/src/member/index.js
```diff
@@ -1,197 +0,0 @@
...
-/* eslint-disable */
-export default class Member {
-  /**
-   * Create a Member.
-   * @param {string} id
-   */
-  constructor({id, isGuest, isHost, isAudioMuted, isVideoMuted, devices}) {
-    this.id = id;
-    // TODO: Make this thenable?
-    this.isGuest = isGuest;
```
mattnorris: Are "members" changing to "participants"? Will we be able to distinguish between people *participating* in the meeting (attending) vs. those just on the meeting invite? Will participants have these same characteristics, e.g., isHost, etc.?

packages/node_modules/@webex/plugin-meetings/src/participants/collection.js
```diff
@@ -0,0 +1,68 @@
...
+const ParticipantCollection = {
+  participants: {},
+
+  /**
+   * @param {*} locus the initial join locus object to set up the participants
+   * @returns {Object} the ParticipantCollection once initialized
+   * @memberof ParticipantCollection
+   */
+  create(locus) {
+    this.participants.selfList = [];
```
chrisadubois: this.participants = {selfList ...

packages/node_modules/@webex/plugin-meetings/src/participants/index.js
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   *
+   * @param {Object} attrs
+   * @param {Object} options
+   * @memberof Participants
+   */
+  constructor(attrs, options) {
+    super({}, options);
+    this.participantRequest = new ParticipantRequest({}, options);
+    this.locusUrl = null;
```
chrisadubois: add a TODO to update all the modules with a locusUrl change and add a updateLocusUrl function

packages/node_modules/@webex/plugin-meetings/src/participants/util.js
```diff
@@ -0,0 +1,181 @@
...
+    if (locusParticipant.hideInRoster) {
+      return;
+    }
+    if (ParticipantUtil.handleAddToSelfList(participants, locusParticipant, selfUserIdentity)) {
+      return;
+    }
+    if (ParticipantUtil.handleAddToHostList(participants, locusParticipant, hostIdentity, selfUserIdentity)) {
+      return;
+    }
+    ParticipantUtil.handledAddToLobbyMeetingUnjoinedLists(participants, locusParticipant);
```
chrisadubois: consider optimizing the order of the function calls here, and add an else { put into not meeting }
put a TODO

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -114,11 +156,34 @@ export default class Meeting extends StatelessSparkPlugin {
...
+        this.type = MeetingUtil.getLocusType(locus.info.locusTags);
+      }
+      if (locus.participants && locus.self) {
+        this.partner = MeetingUtil.getLocusPartner(locus.participants, locus.self);
+      }
+      if (this.partner) {
+        this.sipUri = this.partner.person.sipUrl || this.partner.person.emailAddress || this.partner.person.email;
+      }
+    }
+    this.setLocus(locus);
```
andrewholsted: Why do we call set locus outside of the if(locus) block? Seems like this would/could fail if locus is null in the `setLocus` method

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -407,13 +497,16 @@ export default class Meeting extends StatelessSparkPlugin {
     this.resourceId = options.resourceId;
     this.setCorrelationId(this.id);
     // this.state.join();
+    this.spark.meetings.notifyRinging(this, OUTGOING);
     return MeetingUtil.joinMeeting(this)
       .then((response) => {
-        this.setLocusJoin(response);
+        this.setLocus(response);
+        this.spark.meetings.notifyRinging(this, null, false);
```
andrewholsted: I'm not sure how I feel about calling the plugin internally. I think this might be a better case to either use the event emitter or possibly restructure things. Could you emit the notify ringing event inside of `meeting` instead `meetings` since the ring should only be specific to a particular meeting?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -482,6 +575,41 @@ export default class Meeting extends StatelessSparkPlugin {
...
+   * @memberof Meeting
+   */
+  decline(reason = BUSY) {
+    return this.meetingRequest
+      .declineMeeting({
+        locusUrl: this.locusUrl,
+        deviceUrl: this.deviceUrl,
+        reason
+      })
+      .then((response) => this.spark.meetings.destroy(this, response));
```
andrewholsted: Do we want to destroy on a decline? I'm thinking of a case where I've been invited to a meeting but am on another call. I may want to decline to finish the current call and then hop in via obtp to the one I was invited in. I'm not sure if this is just an edge case though

packages/node_modules/@webex/plugin-meetings/sample/app.js
```diff
@@ -418,25 +482,26 @@ document.getElementById('join').addEventListener('click', () => {
   }
 
   meeting.join({resourceId: deviceId});
+
+  meeting.participants.on('participants:created', (participants) => {
```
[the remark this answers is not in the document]
chrisadubois: +1

packages/node_modules/@webex/plugin-meetings/src/locus-events/index.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * @param {Object} data
+   * @returns {undefined}
+   * @memberof LocusEvents
+   */
+  locusEvent(data) {
+    const {eventType} = data;
+    console.log('Mercury-Locus-Event', eventType);
```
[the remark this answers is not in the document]
chrisadubois: just debugging for now

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: I'll take a look.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: This is just for 1:1 call, so in this case we would want to. For space meeting, we would not want to destroy, it will go through a different flow.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: Yeah I thought about this and wasn't 100%, but was what I could come up with.
Here's the reasoning for doing it on meetings: It's an incoming 1:1 call, client doesn't know for sure which meeting object to listen to. However, we could maybe change the synchronization/sequence of it so that it's for sure AFTER the meeting_added event occurs and then emit the event directly from the meeting object. restructuring in that way would make logical sense to me (though this does as well) what other options for restructuring were you thinking of?

andrewholsted: I like your suggestion. Mainly because we _have_ to create a meeting before they can act on it, so it seems like we should create the meeting and then immediately emit the notify ringing (possibly with a flag on the create method?). 

I wasn't thinking about the incoming case so I don't see another way other than what you proposed above. It's possible to merge this since it's so meaty and then pick that one peace up separately.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
chrisadubois: I can't recall why I did that but I changed it.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -8,9 +8,12 @@ import AudioStateMachine from '../meeting/audio';
 import VideoStateMachine from '../meeting/video';
 import PeerConnectionManager from '../peer-connection-manager';
 import MeetingRequest from '../meeting/request';
+import Participants from '../participants/index';
```
andrewholsted: /index isn't needed here

PR #1087, opened 2019-02-04
feat(federation): make federation config changes

glenatwork:
## Description

Changes for Federation.

The “credentials” config object is changed to an Ampersand object that uses the 'derived' properties concept to generate the CI URL from the current (user’s) CI host name and/or default URL.

A flag “federation” is added to the config “credentials” object to enable federation functionality. 

The method in the SDK that calls the Atlas activations API, which returns a user’s home CI, is the verify() method. This method now automatically updates the internal CI in the config based on what is returned by Alas. 

To maintain the current CI in use on SSO redirect, the "state" query parameter is used to hold the current CI in use.

The SDK activate() method requires the email address to be passed in, in the case of Federation enabled, so that it can call verify() to get the user's home CI. This can be distinguished from the PIN validation case since the PIN comes in as a different parameter than the token used with the email address.

The SDK setPassword() also, temporarily, requires the email address, but, again, only in the case of Federation flag is enabled.  This is due to dependency on WDM, which until Phase 2, does not return the correct CI URL.  The email will not be required for setPassword in the near future.

A setConfig() method is added to spark-core to allow the client to dynamically update config, including “credentials”, if desired.

Fixes # SPARK-32903

## Type of Change

Note: Changes in API signature (requiring email) are only in the case of Federation flag enabled

## Test Coverage

SDK:

- [X] spark-core (added test for setConfig)
- [X] plugin-authorization-browser-first-party (added tests to confirm unique flows for Federation)
- [X] internal-plugin-user (added tests for EU users, users with/without Federation)

Web Client
- [X] Tested US and EU user login in Federated integration environment

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

Note: I don't think Phase 1 of Federation will be doc'd at this time


packages/node_modules/@ciscospark/spark-core/src/config.js
```diff
@@ -4,71 +4,177 @@
...
+export default {
+  maxAppLevelRedirects: 10,
+  maxLocusRedirects: 5,
+  maxAuthenticationReplays: 1,
+  maxReconnectAttempts: 1,
+  onBeforeLogout: [],
+  trackingIdPrefix: 'spark-js-sdk',
+  trackingIdSuffix: '',
+  AlternateLogger: undefined,
+  credentials: new CredentialInfo({
```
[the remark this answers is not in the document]
glenatwork: Good question.  Looking at the examples in the docs, the children are Ampersand objects:
https://ampersandjs.com/docs/#ampersand-state-children
In the SDK, SparkInternalCore in property internal is also an Ampersand object.
So it would seem having the child be an Ampersand object is the model.
Regardless it raises the question whether config should be changed to a child property too (like SparkInternalCore).  It's currently still a session property.  I will look into that.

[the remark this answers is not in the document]
glenatwork: I don't think we want to declare config as a child.  This will mean (config change) events are propagated up which may create unnecessary noise.

packages/node_modules/@ciscospark/test-helper-test-users/src/index.js
```diff
@@ -12,8 +12,8 @@ if (typeof Promise === 'undefined') {
 /* eslint-env mocha */
 /* eslint camelcase: [0] */
 
-const _ = require('lodash');
 const assert = require('assert');
+const _ = require('lodash');
```
mattnorris: This may complain in linting. I believe it checks for alphabetizing.

packages/node_modules/@ciscospark/test-helper-test-users/src/index.js
```diff
@@ -156,5 +156,9 @@ module.exports = {
     }
 
     return _remove(users);
-  }
+  },
+
+  // Org for Federation test users
+  EU_PRIMARY_ORG_ID: 'c1b781f7-b70b-43f7-bac6-6590d98d50fa'
```
mattnorris: Is this okay to put here, or should this be in `.env`?

packages/node_modules/@ciscospark/spark-core/test/unit/spec/spark-core.js
```diff
@@ -122,6 +123,19 @@ describe('Spark', () => {
...
+    it('updates the config', () => {
+      const config = {credentials: {prop: true}};
+
+      const spark = new Spark();
+
+      assert.isUndefined(spark.config.credentials.prop);
+
+      spark.setConfig(config);
+
+      assert.isTrue(spark.config.credentials.prop);
```
mattnorris: 👍

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/automation/spec/authorization-code-grant.js
```diff
@@ -7,105 +7,123 @@
...
-          .waitForElementByCssSelector('#access-token:not(:empty)')
-            .text()
-              .then((text) => {
-                accessToken = text;
-                assert.isString(accessToken);
-                assert.isAbove(accessToken.length, 0);
-                return browser;
-              })
-          .waitForElementByCssSelector('[title="Refresh Access Token"]')
+// Run the tests with and without Fderation enabled,
```
mattnorris: "Federation"

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/automation/spec/authorization-code-grant.js
```diff
@@ -7,105 +7,123 @@
...
+  {it: 'with US user without Federation enabled', EUUser: false, attrs: {}},
+  {it: 'with US user with Federation enabled', EUUser: false, attrs: {config: {credentials: {federation: true}}}}
+];
+
+runs.forEach((run) => {
+  describe(`plugin-authorization-browser-first-party ${run.it}`, function () {
+    this.timeout(120000);
+
+    describe('Authorization', () => {
+      describe.skip('Authorization Code Grant', () => {
```
mattnorris: Remove .skip

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -49,27 +50,43 @@ const User = SparkPlugin.extend({
...
-        return res.body;
-      });
+
+    // if we have options.email and options.verificationToken
+    // and Federation flag is enabled, flag that we need to
+    // lookup user's CI.
+    const needsCILookup = this.spark.config.credentials.federation && options.email && options.verificationToken;
+    return this.getUserCI(options.email, needsCILookup).then(() => {
+      const activateOptions = Object.assign({}, options);
+      delete activateOptions.email;
```
mattnorris: Why delete `email` here?

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -272,6 +296,15 @@ const User = SparkPlugin.extend({
       return Promise.reject(new Error('`options.email` is required'));
     }
 
+    // For Federation, there is only one Atlas, which we get the uesr's CI from
```
mattnorris: "user's"

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -289,10 +322,57 @@ const User = SparkPlugin.extend({
...
       });
   },
 
+
+  /**
+   * If the passed-in condition is true, retrieve the user's
+   * CI and update config.
+   * Otherwise, do nothing, leave currently configured CI.
+   * @param {string} email (required)
+   * @param {boolean} condition (required)
```
mattnorris: Let's rename `condition` to what it does. Maybe `suppressEmail`?

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -289,10 +322,57 @@ const User = SparkPlugin.extend({
...
+   */
+  getUserCI(email, condition) {
+    let verifyPromise;
+    if (condition) {
+      // call verify first to get the user's CI, but suppress sending another email
+      const verifyOptions = {
+        email,
+        suppressEmail: true
+      };
+      verifyPromise = this.verify(verifyOptions).then((res) => Promise.resolve(res.userEntities));
```
mattnorris: I think returning here is fine rather than assigning and returning at the end.

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
```diff
@@ -289,10 +322,57 @@ const User = SparkPlugin.extend({
...
+    }
+    else {
+      verifyPromise = Promise.resolve(
```
mattnorris: See comment above.

packages/node_modules/@ciscospark/test-helper-test-users/src/index.js
glenatwork: This file is no longer changed

packages/node_modules/@ciscospark/test-helper-test-users/src/index.js
glenatwork: Moved to env file

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
glenatwork: The API call takes all the same passed-in args but it chokes on having the email included.

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
glenatwork: It's whether we lookup the user's CI or just return what is already config'd.  Changed it to reflect that it triggers the CI lookup.

packages/node_modules/@ciscospark/plugin-authorization-browser-first-party/test/automation/spec/authorization-code-grant.js
glenatwork: This test wasn't run originally.

packages/node_modules/@ciscospark/test-helper-test-users/src/index.js
mattnorris: @glenatwork is there a reason not to put it in `.env.default`?

packages/node_modules/@ciscospark/internal-plugin-user/src/user.js
mattnorris: 👌

PR #1090, opened 2019-02-06
Adding new eDiscovery plugin

decryan1984:
# Pull Request Template

## Description

Adding a new plugin for ediscovery v2, the purpose of which is to interact with the new ediscovery service to provide a compliance officer with the means to create and download compliance reports.

Fixes # (issue)
NA

## Type of Change
Adding an entirely new plugin: internal-plugin-ediscovery in packages/node_modules/@ciscospark

## Test Coverage

A set of unit tests have been added in internal-plugin-ediscovery/test/unit/spec, they can be run from root with the following command:
npm test -- --package @ciscospark/internal-plugin-ediscovery --node
They can also be debugged through the browser using this command:
npm test -- --package @ciscospark/internal-plugin-ediscovery --browser --karmaDebug

- [x] createReportSuccess
- [x] createReportWithNoParamFailure
- [x] getReportsSuccess
- [x] getReportSuccess
- [x] getReportWithNoParamFailure
- [x] deleteReportSuccess
- [x] deleteReportWithNoParamFailure
- [x] restartReportSuccess
- [x] restartReportWithNoParamFailure
- [x] getContentSummaryPagedSuccess
- [x] getContentSummaryWithNoParamFailure
- [x] mapPosts
- [x] mapEvents
- [x] mapShares
- [x] addActivities

**Test Configuration**:
* Node 10.13.0
* NPM Version 6.4.1

N.B. the plugin has undergone significant manual testing using the test client we have created, which can be found here: https://sqbu-github.cisco.com/WebExSquared/ediscovery/tree/master/jted

## Checklist:


packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+   * @param {Array<String>} spaceIds - A list of space ids relevant to the report
+   * @param {Object} range - Contains the start time and end time defining the search period
+   * @returns {Promise<ResponseEntity>} Http response containing the new report record
+   */
+  @waitForValue('@')
+  createReport(name, description, emails, keywords, spaceNames, spaceIds, range) {
+    if (!emails || !keywords || !spaceNames || !spaceIds || !range) {
+      throw Error('Invalid parameter(s)');
+    }
+
```
decryan1984: Before the report request can be sent to the ediscovery service, any potentially sensitive parameters must be encrypted and any email addresses converted to UUID, this is to prevent PII from being stored server side.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    return decryptNextActivities(this).then(() => reportGenerator.saveAs());
+  },
+
+  /**
+   * Decrypt the activities found in the report content
+   *
+   * @param {Object[]} activities An array of activities with encrypted messages
+   * @returns {Promise<[any]>} The activities with their messages decrypted
+   */
```
decryan1984: Depending on the activity type, the content will de decrypted and in the case of shares and boards, downloaded.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+   *
+   * @param {UUID} reportId - Id of the report being requested
+   * @returns {List<Activities>} HttpResponse
+   */
+  @waitForValue('@')
+  generateReport(reportId) {
+    if (!reportId) {
+      throw Error('Invalid parameter(s)');
+    }
+
```
decryan1984: Retrieves report content from the ediscovery service and decrypts it until the complete report is done.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+      resource: `reports/${reportId}`
+    })
+      .then((res) => res)
+      .catch((reason) => {
+        this.logger.error(`Restart Report Error ${reason}`);
+        return reason;
+      });
+  },
+
+
```
decryan1984: Content summary consists of a list of all the space ids included in the report, and the file count, participant count etc. for each. This may be too many to effectively display in a UI at once, hence the paged API.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+      encryptedName: '',
+      encryptedDescription: '',
+      encryptedKeywords: [],
+      encryptionkeyUrl: ''
+    };
+
+    if (!name && !description && (!keywords || keywords.length <= 0)) {
+      return Promise.resolve(encryptedRequestInfo);
+    }
+
```
decryan1984: A key is created and bound to a KRO in order to ensure it does not expire while the report is still valid.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/report-generator.js
```diff
@@ -0,0 +1,133 @@
...
+const ReportGenerator = SparkPlugin.extend({
+  namespace: 'Ediscovery',
+
+  props: {
+    reportId: 'string',
+    zip: 'object',
+    folder: 'object',
+    spaces: 'object'
+  },
+
```
decryan1984: Activities are currently mapped to a mime format and stored as EML files which are downloaded to the users hard disk.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+   * @param {String} description - A textual summary of the reports purpose
+   * @param {Array<String>} emails - A list of user emails relevant to the report
+   * @param {Array<String>} keywords - A list of search terms relevant to the report
+   * @param {Array<String>} spaceNames - A list of space names relevant to the report
+   * @param {Array<String>} spaceIds - A list of space ids relevant to the report
+   * @param {Object} range - Contains the start time and end time defining the search period
+   * @returns {Promise<ResponseEntity>} Http response containing the new report record
+   */
+  @waitForValue('@')
+  createReport(name, description, emails, keywords, spaceNames, spaceIds, range) {
```
andrewholsted: Once you get beyond a few arguments, it's typically better to just have the method accept an object. Helps with ordering and future proofing.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    if (!emails || !keywords || !spaceNames || !spaceIds || !range) {
+      throw Error('Invalid parameter(s)');
```
andrewholsted: I think you should add a `new` in front of `Error`. I haven't seen it this way before.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    }
+
+    const self = this;
+    const encryptionPromise = self._encryptRequestParams(name, description, keywords);
+    const uuidPromise = self._asUUIDs(emails);
+
+    return Promise.all([encryptionPromise, uuidPromise])
+      .then((values) => {
+        const body = reportRequest;
+        body.name = values[0].encryptedName;
```
andrewholsted: I don't think you want to assign `body` to `reportRequest` and then modify it. You will end up changing `reportRequest` by reference. Here is an example

```
const report = {
	a: 'foo',
	b: 'bar'
}

const body = report;

body.a = 'bar';

console.log('body', body);
console.log('report', report);
```

yields
```
body { a: 'bar', b: 'bar' }
report { a: 'bar', b: 'bar' }
```

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+   * @returns {Promise<ResponseEntity>} Http response containing the new report record
+   */
+  @waitForValue('@')
+  createReport(name, description, emails, keywords, spaceNames, spaceIds, range) {
+    if (!emails || !keywords || !spaceNames || !spaceIds || !range) {
+      throw Error('Invalid parameter(s)');
+    }
+
+    const self = this;
+    const encryptionPromise = self._encryptRequestParams(name, description, keywords);
```
andrewholsted: It's an anitpattern to use `self`. We should use arrow functions or other means to still reference `this`

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+  },
+
+  /**
+   * Retrievse the user UUIDs associated with each supplied email. If an email with no user
+   * is found an error is thrown. If multiple such emails exist only the first encountered
+   * throws an error.
+   *
+   * @param {Array<String>} emails - A list of emails to resolve
+   * @returns {Array<Promise<UUID>>} The UUIDs associated with the emails
+   */
```
andrewholsted: Nice comments/jsdocs in general

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    if (!emails) {
+      throw Error('Invalid parameter');
+    }
+
+    if (emails.length <= 0) {
+      return Promise.resolve([]);
+    }
+
+    const promises = [];
+    emails.forEach((userEmail) => {
```
andrewholsted: if `bluebird` is included in the sdk you can use `Promise.map()` to handle some of this broilerplate with pushing promises.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+  /**
+   * Encrypts any potentially sensitive information relating to the report request
+   * The encrypted info is then sent to the eDiscovery service and only decrypted when retrieved again by the client
+   *
+   * @param {String} name - A label to identify the report
+   * @param {String} description - A textual summary of the reports purpose
+   * @param {Array<String>} keywords - A list of search terms relevant to the report
+   * @returns {Promise<encryptedRequestInfo>} Object containing encrypted versions of the original parameters
+   */
+  _encryptRequestParams(name, description, keywords) {
```
andrewholsted: These encrypt/decrypt methods might be better suited as transforms. @mattnorris? I'm not sure. I know they work on incoming requests

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    }
+
+    return this.request({
+      method: 'GET',
+      service: 'ediscovery',
+      resource: `reports/${reportId}`
+    })
+      .then((res) => {
+        if (res.body.reportRequest && res.body.reportRequest.encryptionKeyUrl && res.body.reportRequest.keywords) {
+          return this._decryptRequestParams(res.body.reportRequest.encryptionKeyUrl, res.body.reportRequest.name, res.body.reportRequest.description, res.body.reportRequest.keywords)
```
andrewholsted: definitely think a transform is the preferred sdk style here.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+    return this.request({
+      method: 'GET',
+      service: 'ediscovery',
+      resource: 'reports'
+    })
+      .then((res) => {
+        const promises = [];
+        for (let i = 0; i < res.body.length; i += 1) {
+          if (res.body[i].reportRequest && res.body[i].reportRequest.encryptionKeyUrl && res.body[i].reportRequest.keywords) {
+            promises.push(this._decryptRequestParams(res.body[i].reportRequest.encryptionKeyUrl, res.body[i].reportRequest.name, res.body[i].reportRequest.description, res.body[i].reportRequest.keywords)
```
andrewholsted: same as above

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+
+    return this.request({
+      method: 'DELETE',
+      service: 'ediscovery',
+      resource: `reports/${reportId}`
+    })
+      .then((res) => res)
+      .catch((reason) => {
+        this.logger.error(`Delete Report Error ${reason}`);
+        return reason;
```
andrewholsted: I think you might want to rethrow the err reason here otherwise it will fail silently

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+   *
+   * @param {Object[]} activities An array of activities with encrypted messages
+   * @returns {Promise<[any]>} The activities with their messages decrypted
+   */
+  _decryptActivities(activities) {
+    const promises = [];
+    for (let i = 0; i < activities.length; i += 1) {
+      const activity = activities[i];
+      if (activity.verb === 'post' && activity.objectDisplayName && activity.encryptionKeyUrl) {
+        const textPromise = this.spark.internal.encryption.decryptText(activity.encryptionKeyUrl, activity.objectDisplayName, {onBehalfOf: activity.actorId})
```
andrewholsted: Is there any reason we can't use the existing decrypts for activities?

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/report-generator.js
```diff
@@ -0,0 +1,133 @@
...
+   */
+  saveAs() {
+    if (!this.zip) {
+      throw Error('No activities added to report');
+    }
+
+    return this.zip.generateAsync({type: 'blob'})
+      .then((content) => {
+        // Download to the user's browser
+        saveAs(content, `${this.reportId}.zip`);
```
andrewholsted: is there a reason to use this over the native `fs` methods? I'm not against bringing in dependencies, just trying to understand it better.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Good idea, in fact, we already have a reportRequest object we could use: 
https://github.com/webex/spark-js-sdk/pull/1090/files#diff-bfd2923745106b14ceb4ea04fd757d5c

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Happy to go with convention, but I think this is the same as using _new_: http://es5.github.io/#x15.11.1
In general I've tried to avoid using _new_ where possible.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: I think we should make the reportRequest into a proper class, use it for the function param and create another for this body object.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Nice, will look into it

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Definitely

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: I wasn't aware of an existing decrypt method for activities, is there something available in a plugin aside from encryption?

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: I'm not familiar with a transform pattern, can you point me towards an example?

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/report-generator.js
decryan1984: I think @gannons may be in a better position to discuss that

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -0,0 +1,520 @@
...
+/*!
+ * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {waitForValue, SparkPlugin} from '@ciscospark/spark-core';
+import reportRequest from './report-request';
+import ReportGenerator from './report-generator';
+
+/**
+ * @class Ediscovery is used by compliance officers to run compliance reports
```
mattnorris: I would probably rename to `EDiscovery` to be consistent with other names within the repo. This name is a unique though.

packages/node_modules/@ciscospark/internal-plugin-ediscovery/src/report-generator.js
gannons: The native fs module is stripped when the sdk is wrapped by browserify. i.e. As the browse does not allow direct file system access. There is a replacement module called browserify-fs. But it works on an in-memory file system.

The saveAs(..) function prompts the browse to kick off a download.

PR #1092, opened 2019-02-14
 added domains for whitelisting

arun3528:
# Pull Request Template
Just added new domains for whitelisting
## Description

Added fix for web client integration issue .
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-57301

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


packages/node_modules/@ciscospark/spark-core/src/config.js
```diff
@@ -73,8 +73,11 @@ export default {
     preDiscoveryServices: {
       hydraServiceUrl: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1'
     },
+    validateDomains: process.env.VALIDATE_DOMAINS,
```
andrewholsted: you should add a `|| false` here too

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
```diff
@@ -61,14 +61,6 @@ export default class AuthInterceptor extends Interceptor {
       return Promise.resolve(false);
     }
 
-    // returns true if uri is in whitelistedServiceDomains and is requested
-    if (options.uri && options.runWhitelistedDomains) {
```
andrewholsted: this will be a breaking change for any dev that doesn't update their code, right?

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
```diff
@@ -78,9 +70,26 @@ export default class AuthInterceptor extends Interceptor {
...
         if (options.uri) {
-          return this.spark.internal.device.isServiceUrl(options.uri);
+          // If service Url then return or else check if whitelisted
+          return this.spark.internal.device.isServiceUrl(options.uri)
+            .then((isServiceUrl) => {
+              if (isServiceUrl) {
+                return true;
+              }
+              if (options.addAuthHeader === undefined) {
+                options.addAuthHeader = true;
```
andrewholsted: I don't think we want to add it by default unless the options.addAuthHeader is present.

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/token.js
```diff
@@ -269,6 +269,7 @@ const Token = SparkPlugin.extend({
     return this.spark.request({
       method: 'POST',
       uri: this.config.tokenUrl,
+      addAuthHeader: false,
```
andrewholsted: false should be the default so this isn't necessary

packages/node_modules/@ciscospark/spark-core/src/config.js
mwegman: If you are worried about `undefined` here, do `!!process.env.VALIDATE_DOMAINS`

packages/node_modules/@ciscospark/spark-core/src/lib/credentials/token.js
arun3528: This is necessary if we make the value as true . We still should make this as false for token request

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
arun3528: No if a developer doesn't pass the header and the global allow i false we dont add the header or do whitelisting check

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
arun3528: ya changed it and relays on the global default

packages/node_modules/@ciscospark/spark-core/src/config.js
arun3528: its a string for now added  || "false" . any suggestion on how to do it . I dont think we can pass boolean

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
mattnorris: Why do we need `addAuthHeader`? I thought the internals of `isServiceUrl` were changing and checking the new config option.

packages/node_modules/@ciscospark/spark-core/src/config.js
mattnorris: I did something similar for [skipping flaky tests](https://github.com/webex/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/test-helper-mocha/src/index.js#L60). What's the best way to accomplish this?

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
```diff
@@ -78,9 +70,27 @@ export default class AuthInterceptor extends Interceptor {
         if (options.service) {
           return this.spark.internal.device.isService(options.service);
         }
-
         if (options.uri) {
-          return this.spark.internal.device.isServiceUrl(options.uri);
+          // If service Url then return or else check if whitelisted
+          return this.spark.internal.device.isServiceUrl(options.uri)
+            .then((isServiceUrl) => {
+              if (isServiceUrl) {
```
mattnorris: @arun3528 please see my comment below. This part seems redundant. Shouldn't the function `isServiceUrl` be modified to accommodate this case? Otherwise the logic is split into two files.

arun3528: @mattnorris  currently the isServiceUrl only checks if the url exists in the serviceCatalog

The whitelisting check  against the whitelist domains was already here i just moved it around.

addAuthHeader will be false if the env variabled it set . There is one request for which we dont want to add the header at all . For that request we need to pass the header anyways 

We can have a call tomorrow . may be there is a better solution

packages/node_modules/@ciscospark/internal-plugin-wdm/src/config.js
```diff
@@ -12,7 +12,11 @@ export default {
     },
     defaults: {
       name: process.title || inBrowser && 'browser' || 'javascript',
-      deviceType: 'UNKNOWN'
+      deviceType: 'WEB',
+      model: 'modelx',
```
arun3528: web-model

packages/node_modules/@ciscospark/spark-core/src/interceptors/auth.js
mattnorris: Resolved: it's clearer to explicitly separate the service catalog logic from the whitelist logic instead of conflating the two.

packages/node_modules/@ciscospark/internal-plugin-wdm/src/config.js
```diff
@@ -12,7 +12,11 @@ export default {
     },
     defaults: {
       name: process.title || inBrowser && 'browser' || 'javascript',
-      deviceType: 'UNKNOWN'
+      deviceType: 'WEB',
```
mattnorris: @arun3528 let's name them the following: 

```
deviceType: 'WEB'
model: 'js-sdk'
localizedModel: 'js-sdk'
systemName: 'JS_SDK'

arun3528:
@mattnorris  fixed the config file.

PR #1103, opened 2019-02-27
tests for meetings/meeting and docs

chrisadubois:


packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,235 @@
...
+          });
+        });
+      });
+      describe('#create', () => {
+        beforeEach(() => {
+          spark.meetings.createMeeting = sinon.stub().returns(Promise.resolve(true));
+        });
+        it('calls createMeeting and returns its promise', async () => {
+          const create = spark.meetings.create('test', 'test');
+          assert.exists(create.then);
```
andrewholsted: Nice assurance of a promise

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,235 @@
...
+        describe('successful MeetingInfo.#fetchMeetingInfo', () => {
+          beforeEach(() => {
+            spark.meetings.meetingInfo.fetchMeetingInfo = sinon.stub().returns(Promise.resolve(true));
+          });
+          it('creates the meeting from a successful meeting info fetch', async () => {
+            const meeting = await spark.meetings.createMeeting('test', 'test');
+            assert.instanceOf(meeting, Meeting, 'createMeeting should eventually resolve to a Meeting Object');
+            assert.calledOnce(spark.meetings.meetingInfo.fetchMeetingInfo);
+            assert.calledOnce(MeetingsUtil.extractDestination);
+            assert.calledOnce(MeetingsUtil.getMeetingAddedType);
```
andrewholsted: I don't know if we need to get this granular on asserting individual steps inside of `create`. The most important thing is we assert on what the expected return value is (and that it resolves/rejects a promise).

chrisadubois: I'll add 2 tests, one detailed and one contract based.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+          describe('after #setReachability', () => {
+            beforeEach(() => {
+              spark.meetings.setReachability();
+              const reachabilityMocker = spark.meetings.getReachability();
+              sinon.stub(reachabilityMocker, 'gatherReachability').returns(true);
+            });
+            it('gets the reachability data instance from spark.meetings', () => {
+              const reachability = spark.meetings.getReachability();
+              assert.exists(reachability, 'reachability is defined because #setReachability has been called');
+              assert.typeOf(reachability, 'object', 'should be an object');
```
andrewholsted: I would rather assert that specific methods exist instead of asserting it's an instanceOf or that it's an object. If we make a breaking change to the API contract, this won't help us. In this case, just asserting that there is a .getReachability() is probably enough. Then we will have separate unit tests for that component later down the line

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+            staticMeetings = null;
+          });
+          describe('MeetingCollection getByKey proxies', () => {
+            beforeEach(() => {
+              MeetingCollection.getByKey = sinon.stub().returns(true);
+            });
+            describe('#getMeetingByType', () => {
+              it('gets the Meeting instance from MeetingCollection using type and value', () => {
+                staticMeetings.getMeetingByType('test', 'test');
+                assert.calledOnce(MeetingCollection.getByKey);
```
andrewholsted: Instead of asserting that in internal function was called, we should assert that `test` was returned. I would prefer to insert an actual object (multiple actually) in the `beforeEach` then assert that the correct object is returned.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+          });
+          describe('MeetingCollection getAll proxies', () => {
+            beforeEach(() => {
+              MeetingCollection.getAll = sinon.stub().returns(true);
+            });
+            describe('#getAllMeetings', () => {
+              it('calls MeetingCollection to get all meetings with supplied options', () => {
+                staticMeetings.getAllMeetings({test: 'test'});
+                assert.calledOnce(MeetingCollection.getAll);
+                assert.calledWith(MeetingCollection.getAll, {test: 'test'});
```
andrewholsted: Same as above. Insert multiple meetings in before each and assert they are all returned

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+              });
+            });
+            describe('#getActiveMeetings', () => {
+              it('calls MeetingCollection to get all active meetings using the active option', () => {
+                staticMeetings.getActiveMeetings();
+                assert.calledOnce(MeetingCollection.getAll);
+                assert.calledWith(MeetingCollection.getAll, {active: true});
```
andrewholsted: same as above

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+              });
+            });
+            describe('#getScheduledMeetings', () => {
+              it('calls MeetingCollection to get all scheduled meetings using the scheduled option', () => {
+                staticMeetings.getScheduledMeetings();
+                assert.calledOnce(MeetingCollection.getAll);
+                assert.calledWith(MeetingCollection.getAll, {scheduled: true});
```
andrewholsted: same as above

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+              MeetingCollection.getByKey = sinon.stub().returns({locusInfo: {onFullLocus}});
+            });
+            it('tests the sync meeting calls for existing meeting', async () => {
+              await spark.meetings.syncMeetings();
+              assert.calledOnce(spark.meetings.request.getActiveMeetings);
+              assert.calledOnce(MeetingCollection.getByKey);
+              assert.calledOnce(onFullLocus);
+              assert.calledWith(spark.meetings.request.getActiveMeetings);
+              assert.calledWith(MeetingCollection.getByKey, 'locusUrl', 'test');
+              assert.calledWith(onFullLocus, {url: 'test'});
```
andrewholsted: I don't think we want to assert on individual internal methods being called. We should be able to change those and as long as the end result is the same, the test should pass.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+      describe('#listenInternal', () => {
+        it('listens to the decline 1:1 internal event', (done) => {
+          const spy = sinon.spy(spark.meetings, 'destroy');
+          spark.meetings.listenInternal();
+          Events.emit('DESTROY_MEETING_DECLINE_1_1', {meeting: {id: 'test'}, response: 'test', type: 'test'});
+          assert.calledOnce(spy);
+          done();
+        });
+      });
+      describe('#handleLocusMercury', () => {
```
andrewholsted: looks good!

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -0,0 +1,321 @@
...
+          });
+          it('should setup the meeting by difference event', async () => {
+            await spark.meetings.handleLocusEvent({locus: 'test', eventType: 'locus.difference', locusUrl: 'test'});
+            assert.calledOnce(MeetingCollection.getByKey);
+            assert.calledWith(MeetingCollection.getByKey, 'locusUrl', 'test');
+            assert.calledOnce(initialSetup);
+            assert.calledWith(initialSetup, 'test', {locusInfo: {initialSetup}});
+          });
+          it('should setup the meeting by difference event', async () => {
+            await spark.meetings.handleLocusEvent({locus: 'test', eventType: 'test', locusUrl: 'test'});
```
andrewholsted: this is purely an internal handling correct? an end user wouldn't invoke that method?

chrisadubois: correct

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
chrisadubois: hmmmmm, i agree with you, but we also want to make sure when we make changes, that the function is doing everything it was initially supposed to, so if somebody changes something internal, that was initially meant to also be called (for example, unsetlocalStream AND unset localShare), we would have broken behavior.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
chrisadubois: okay, will take a look.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -0,0 +1,715 @@
...
+      logger: {log},
+      meetings: {config: {mediaSettings: {}}}
+    };
+    beforeEach(() => {
+      meeting = new Meeting(
+        {
+          userId: 'test',
+          resource: 'test',
+          deviceUrl: 'test',
+          locus: {url: 'test'}
```
andrewholsted: I think it's helpful to put realistic looking data in tests. Otherwise you get a string of `test` everywhere and it's possible it's not the value you think it is but some other prop.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -0,0 +1,715 @@
...
+            assert.calledOnce(meeting.getMediaStreams);
+          });
+        });
+        describe('rejection', () => {
+          it('should error out and return a promise', async () => {
+            meeting.join = sinon.stub().returns(Promise.reject());
+            meeting.getMediaStreams = sinon.stub().returns(true);
+            await meeting.joinWithMedia();
+            assert.calledOnce(meeting.join);
+            assert.notCalled(meeting.getMediaStreams);
```
andrewholsted: doesn't joinWithMedia call getMediaStreams internally? I don't know why we would assert that we don't (or why it passes)

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
chrisadubois: yeah, i like the idea of adding assert.exists checks for the public api/event functions

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
chrisadubois: likely commenting on a resolution vs rejection describe block

```
        describe('resolution', () => {
          it('should success and return a promise', async () => {
            meeting.join = sinon.stub().returns(Promise.resolve('test'));
            meeting.getMediaStreams = sinon.stub().returns(Promise.resolve(['test', 'test']));
            meeting.addMedia = sinon.stub().returns(Promise.resolve('test'));
            await meeting.joinWithMedia();
            assert.calledOnce(meeting.join);
            assert.calledOnce(meeting.getMediaStreams);
          });
        });
        describe('rejection', () => {
          it('should error out and return a promise', async () => {
            meeting.join = sinon.stub().returns(Promise.reject());
            meeting.getMediaStreams = sinon.stub().returns(true);
            await meeting.joinWithMedia();
            assert.calledOnce(meeting.join);
            assert.notCalled(meeting.getMediaStreams);
          });
        });
```

chrisadubois:
@andrewholsted -- yeah i definitely feel the overboardness, not arguing with you on that front.

I do feel it adds some locked-down-ness to the files, lets go over in standup tomorrow and decide as a group?

Also I added tests for MeetingCollection in a separate file that addressed your concerns for pulling values out of the Collection.

packages/node_modules/@webex/plugin-meetings/src/meetings/collection.js
```diff
@@ -40,6 +39,7 @@ const MeetingCollection = {
    *
    * @param {Object} options An object containg properties to filter by
    * @param {Boolean} options.active Only return active meetings
+   * @param {Boolean} options.scheduled Only return scheduled meetings
```
chrisadubois: remove

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -668,29 +696,27 @@ export default class Meeting extends StatelessSparkPlugin {
...
+   * @param {MediaStream} options.localShare
    * @returns {Promise}
    * @memberof Meeting
    */
   addMedia(options = {}) {
     const {localStream, localShare} = options;
     // eslint-disable-next-line no-warning-comments
     // TODO wire into default config. There's currently an issue with the stateless plugin or how we register
-    this.mediaProperties.setMediaDirection(Object.assign(this.config.mediaSettings, options.mediaSettings));
+    this.mediaProperties.setMediaDirection(process.env.NODE_ENV === 'test' ? {} : Object.assign(this.config.mediaSettings, options.mediaSettings));
```
chrisadubois: let's edit this with mockSpark instance so that meeting has a spark.meetings.config.mediaSettings stored due to StatelessSparkPlugin

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/collection.js
```diff
@@ -0,0 +1,69 @@
...
+  describe('meetings collection', () => {
+    beforeEach(() => {
+      MeetingCollection.meetings = {};
+    });
+    describe('#get', () => {
+      beforeEach(() => {
+        MeetingCollection.meetings.test = 'test';
+      });
+      it('should get the meeting', () => {
+        assert.equal(MeetingCollection.get('test'), 'test');
```
chrisadubois: missed on the checking for size of the collection

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/collection.js
```diff
@@ -0,0 +1,69 @@
...
+        MeetingCollection.meetings.test = 'test';
+        MeetingCollection.meetings.test2 = 'test2';
+      });
+      it('should get all the meeting', () => {
+        assert.equal(Object.keys(MeetingCollection.meetings).length, 2);
+        assert.deepEqual(Object.keys(MeetingCollection.getAll()), ['test', 'test2']);
+      });
+      describe('with filters', () => {
+        beforeEach(() => {
+          MeetingCollection.meetings.test = {id: 'test', active: true};
```
chrisadubois: change id to uuid format, don't test beyond what values are used from the meeting object

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/collection.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+      it('should get the meeting', () => {
+        assert.equal(Object.keys(MeetingCollection.meetings).length, 1);
+        assert.equal(MeetingCollection.remove('test'), 'test');
+        assert.equal(Object.keys(MeetingCollection.meetings).length, 0);
+      });
+    });
+    describe('#getAll', () => {
+      beforeEach(() => {
+        MeetingCollection.meetings.test = 'test';
```
chrisadubois: using 'test' is problematic, use like id1 or id2, or value1 value2, or pmr pmr1, pmr2 etc

PR #1104, opened 2019-03-04
fix(encryption): kms onbehalfof not working as expected

gannons:
# Pull Request Template

## Description

When the ediscovery plugin uses the encryption plugin to decrypt a set of activities, the decryption can fail intermittently, due to an error such as:

> Unhandled promise rejection KmsError: "User 93b4bd25-ed19-4c72-81a4-a9c8b6241708 doesn't have authorization to access resource kms://ciscospark.com/resources/c6fab088-c06e-49f1-938f-810cc0a9a92c
> KMS_RESPONSE_STATUS: 403
> KMS_REQUEST_ID: 41c43f23-3e7e-4f5c-8ade-e3b00a69dcf5"

This occurs when the incorrect user id is used to request the decryption. In the example below, the activity clearly has the actor id 91facb42-c96f-441d-b45c-718ba5a71786, which is sent to the encryption plugin, however, the user id attached to the KMS request to get the key is for a different user (who is not involved in the same activity).

There were two roots to this issue.

1) The KMS Context is a global object and  holds a property for the user id. The id is, by default, set to the logged in user. But when making a request on behalf of someone else it needs to be the id of that person. We set it in an async callback. This means that if multiple requests come in they can interfere with each other. That causes the incorrect userId to be used for some requests.

2) The user id is set within a _getContext() method. This method is annotated with a @oneFlight decorator which ensures the SDK does not spam a service. It does this by caching the initial promise and returning it to subsequent requests. i.e. At least until the promise is resolved. To avoid the unwanted caching we can differentiate the requests based on their parameters. e.g.

    @oneFlight({
     keyFactory: ({uri, onBehalfOf}) => `${uri}/${onBehalfOf}`
    })

However a simple solution to both of the above is to move out of the _getContext() method and to create a copy of the context to update.

## Type of Change

## Test Coverage

A test that reproduced the issue is included in the PR.

## Checklist:


gannons:
> Are there other scenarios where we would be making a KMS request with a different user id?

There are - I'm not familiar with them but I know the search services have switched over. The conversation service will/is doing work also to use it. But neither are using it from the spark-js-sdk.

> Do clients need to make changes to pick up this difference?

No client changes should be required. This was specifically a caching bug in the sdk. Client usage should see no change. AFAIK the only client using onbehalfof is ediscovery.js.

PR #1111, opened 2019-03-12
Meeting link fix

chrisadubois:
fix meeting link lookup regex and add retry logic


packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -28,14 +30,21 @@ const MeetingInfoUtil = {};
 MeetingInfoUtil.isMeetingLink = (value) => {
   // eslint-disable-next-line no-warning-comments
   // TODO: lets remove regex from this equation and user URI matchers and such
-  const regex = /(.*).webex.com\/(?:meet|m)\/(.*)/;
-  const m = regex.exec(value);
+  const m = MEETING_LINK_SIP_REGEX.exec(value);
```
andrewholsted: Can we _not_use a regex please? As the TODO says, we should just parse the url that way we can assert on various elements of the uri. 

```
const meetingLink = new URL(url);
if (meetingLink.hostname === 'foo') { }
```

etc. Is there value to the above?

chrisadubois: yeah i mean i want to find a more complete solution than URL because we'd still need a regex to ensure that *.webex.com is a a correct regex
and that meet or m or join appears
and that we can extract the user from /meet/<user>

I agree we should get away from regex, i added that TODO, we just didn't have cycles to work on it at the time.

chrisadubois: and we need to ensure not to include the www for converting links to sip URI

andrewholsted: I would rather do this now than punt on it. Otherwise the regex will live forever. I can help contribute the code.

packages/node_modules/@webex/plugin-meetings/src/constants.js
```diff
@@ -57,6 +57,7 @@ export const LOCUS_URL = 'locusUrl';
 export const LEFT = 'LEFT';
 
 export const MEETING_LINK = 'MEETING_LINK';
+export const MEETING_LINK_SIP_REGEX = /(?:(?:http|https):\/\/)?(?:www\.)?(\w+).webex.com\/(?:meet|join|m)\/(.+)/;
```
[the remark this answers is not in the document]
chrisadubois: yeah i can spend the time to remove the regex

packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
```diff
@@ -45,21 +45,52 @@ export default class MeetingInfo extends StatelessSparkPlugin {
...
-      type,
-      conversationServiceUrl: this.spark.internal.device.services.conversationServiceUrl
+    const options = this.fetchInfoOptions(destination, type);
+    return this.requestFetchInfo(options).catch((error) => {
+      // retry meeting info fetch logic
+      // if it failed first as a meeting link
+      // try converting it to a sip address
+      // and doing the look up as a SIP_URI against server
+      if (options.type === MEETING_LINK) {
+        return this.requestFetchInfo(this.fetchInfoOptions(MeetingInfoUtil.convertLinkToSip(destination), SIP_URI));
```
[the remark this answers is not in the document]
chrisadubois: it is broken down O.o

packages/node_modules/@webex/plugin-meetings/src/constants.js
```diff
@@ -57,6 +59,7 @@ export const LOCUS_URL = 'locusUrl';
 export const LEFT = 'LEFT';
 
 export const MEETING_LINK = 'MEETING_LINK';
+// export const MEETING_LINK_SIP_REGEX = /(?:(?:http|https):\/\/)?(?:www\.)?(\w+).webex.com\/(?:meet|join|m)\/(.+)/; // TODO: remove
```
mattnorris: Remove now? If it's needed for reference, maybe place it in a doc.

PR #1116, opened 2019-03-18
Remove incompatible & unused browser tests

mattnorris:
# Pull Request 

## Description

Remove IE11 tests and old, unused browser tests. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-64534

## Type of Change

## Checklist:

### DIFF
--- a/browsers-ng.js
+++ b/browsers-ng.js
@@ -49,12 +49,6 @@ module.exports = function(packageName, argv) {
         browserName: 'MicrosoftEdge',
         version: 'latest'
       },
-      sl_ie_11_win7: {
-        base: 'SauceLabs',
-        platform: 'Windows 7',
-        browserName: 'internet explorer',
-        version: '11'
-      },
       sl_safari_latest_osx13: {
         base: 'SauceLabs',
         platform: 'macOS 10.13',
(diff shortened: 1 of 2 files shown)


taymoork2:
Missed two files
`packages/node_modules/@ciscospark/plugin-phone/browsers.js`
`packages/node_modules/@ciscospark/media-engine-webrtc/browsers.js`

mattnorris:
@taymoork2 I was going to leave those because they have the latest FF instead of an arbitrary older one. They are commented out though for what I assume would be quick troubleshooting. Is that worth it to keep?

taymoork2:
@mattnorris up to you, nothing changes if we keep it or remove it since they're already commented out

mattnorris:
@taymoork2 let's keep them for now and reevaluate when we do a holistic review.

taymoork2:
Asking this question for future reference

Why are we getting rid of IE tests on saucelabs?

mattnorris:
Actually... we don't have to. There are problems when running them in development mode locally because they are not transpiled correctly by Babel on-the-fly. But when transpiled to `dist` they do work for the most part. 

I'll reference these commits later when cleaning up the available browsers we actually don't test on.

mattnorris:
Resurrecting after our conversation at standup.

PR #1119, opened 2019-03-19
Emit messages "created" events

mattnorris:
# Pull Request 

## Description

Emit messages `created` events when a new activity is received from Mercury. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-62543

## Type of Change

## Test Coverage

```
DEBUG=messages npm test -- --package @ciscospark/plugin-messages --node
```

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/common/src/uuid-utils.js
+++ b/packages/node_modules/@ciscospark/common/src/uuid-utils.js
@@ -27,10 +27,3 @@ export function deconstructHydraId(id) {
     type: payload.pop()
   };
 }
-
-export const hydraTypes = {
-  MESSAGE: 'MESSAGE',
-  PEOPLE: 'PEOPLE',
-  ROOM: 'ROOM',
-  TEAM: 'TEAM'
-};
--- a/packages/node_modules/@ciscospark/common/src/uuid-utils.js
+++ b/packages/node_modules/@ciscospark/common/src/uuid-utils.js
@@ -1,0 +1,36 @@
+import {encode, decode} from './base64';
+
+/**
+ * Constructs a Hydra ID for a given UUID and type.
+ *
+ * @export
+ * @param {string} type one of PEOPLE, TEAM, ROOM
+ * @param {any} id
+ * @returns {string}
+ */
+export function constructHydraId(type, id) {
+  return encode(`ciscospark://us/${type.toUpperCase()}/${id}`);
+}
+
+/**
+ * Deconstructs a Hydra ID and provides the UUID.
+ *
+ * @export
+ * @param {String} id
+ * @returns {String}
+ */
+export function deconstructHydraId(id) {
+  const payload = decode(id).split('/');
+
+  return {
+    id: payload.pop(),
+    type: payload.pop()
+  };
+}
+
+export const hydraTypes = {
+  MESSAGE: 'MESSAGE',
+  PEOPLE: 'PEOPLE',
+  ROOM: 'ROOM',
+  TEAM: 'TEAM'
+};
--- a/packages/node_modules/@ciscospark/plugin-messages/src/messages.js
+++ b/packages/node_modules/@ciscospark/plugin-messages/src/messages.js
@@ -5,6 +5,8 @@
 import {SparkPlugin, Page} from '@ciscospark/spark-core';
 import {isArray} from 'lodash';
 
+const MESSAGES_CREATED = 'messages:created';
+
 /**
  * @typedef {Object} MessageObject
  * @property {string} id - (server generated) Unique identifier for the message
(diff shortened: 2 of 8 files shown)


packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -28,6 +41,72 @@ import {isArray} from 'lodash';
...
+   * Initializer
+   * @param  {...any} args
+   * @returns {undefined}
+   */
+  initialize(...args) {
+    Reflect.apply(SparkPlugin.prototype.initialize, this, args);
+
+    // Listen to new activities from Mercury.
+    this.listenTo(this.spark.internal.mercury, 'event:conversation.activity',
+      (event) => this.onConversationActivityEvent(event));
```
andrewholsted: I feel like this code should go inside the listen method.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -28,6 +41,72 @@ import {isArray} from 'lodash';
...
+  /**
+   * Trigger a "messages:created" event.
+   * @param {Object} event
+   * @returns {undefined}
+   */
+  onConversationActivityEvent(event) {
+    const {activity} = event.data;
+
+    // Reply activities are not currently supported
+    if (activity.type === API_ACTIVITY_TYPE.REPLY) {
```
andrewholsted: Are these threaded replies?

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
andrewholsted: And you could optionally call listen automatically to prevent an end user from having to do it?

mattnorris: That's what I was going back and forth on. I can hook up mercury by default, but wasn't sure if I should.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
mattnorris: Yes. I figured I'd handle that while I was in there.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
mattnorris: @andrewholsted I moved listening into one function.

packages/node_modules/@ciscospark/common/src/uuid-utils.js
```diff
@@ -0,0 +1,36 @@
...
+export function deconstructHydraId(id) {
+  const payload = decode(id).split('/');
+
+  return {
+    id: payload.pop(),
+    type: payload.pop()
+  };
+}
+
+export const hydraTypes = {
```
adamweeks: Should we move these to the constants file?

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -28,6 +42,71 @@ import {isArray} from 'lodash';
...
+      .then(() => this.listenTo(this.spark.internal.mercury, INCOMING_MESSAGES_CREATED,
+        (event) => this.onConversationActivityEvent(event)));
+  },
+
+  /**
+   * Trigger a "created" event.
+   * @param {Object} event
+   * @returns {undefined}
+   */
+  async onConversationActivityEvent(event) {
```
adamweeks: I don't see where this needs to be async

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -28,6 +42,71 @@ import {isArray} from 'lodash';
...
+      }
+    }
+  },
+
+  /**
+   * Constructs an event detail object for messages:created
+   * @export
+   * @param {Object} activity from mercury
+   * @param {boolean} fromHydra get decrypted text from Hydra by default
+   * @returns {Object} constructed event
```
adamweeks: Would be great to have a bit more info on the shape of the object returned.

packages/node_modules/@ciscospark/plugin-messages/test/integration/spec/messages.js
```diff
@@ -215,5 +217,86 @@ describe('plugin-messages', function () {
...
+            tags: [],
+            roomType: 'group'
+          }
+        ].forEach((input) => {
+          it(`creates a messages event for a ${input.roomType} room`, () => {
+            const activityEvent = makeActivityEvent('post', input.tags);
+            const {activity} = activityEvent.data.data;
+            const event = spark.messages.createMessagesEventData(activity);
+
+            debug(`messages event created: ${JSON.stringify(event)}`);
```
adamweeks: This needed?

packages/node_modules/samples/browser-socket/index.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Sample: Socket</title>
+    <link rel="stylesheet" href="https://unpkg.com/spectre.css/dist/spectre.min.css">
```
adamweeks: Would be awesome to use collab-ui here

packages/node_modules/@ciscospark/plugin-messages/test/integration/spec/messages.js
mattnorris: It only prints to screen if you use `DEBUG=messages` from the CLI. I found it helpful.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
mattnorris: It doesn't have to anymore. I had it fetching something else at one point. Good catch!

packages/node_modules/samples/browser-socket/index.html
mattnorris: Agreed. I just needed something quick and supported. I would rather not include it as another build step. I may use https://unpkg.com/@collab-ui/core/css/collab-ui.min.css when revising.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
jpjpjp: @mattnorris, when we are all said and done with this work, ideally we'll also have events coming out of the memberships and rooms module as well.    We could add a listen() method to each of those modules as well, but are there inefficiencies on having 3 different plugins all listening for all the conversation activity events on the same socket?

If not that is certainly the most readable way to do this implementation.  If so it seems like we'd still like to have the developer do something proactive to indicate that they want message (or membership, or room) events, but to have a common internal socket plugin that routes the appropapriate conversation activities to the appropriate module handler.

mattnorris: @jpjpjp good point; they will all listen to the same socket, so there won't be inefficiencies.

mattnorris:
Did not close automatically

PR #1126, opened 2019-04-02
Feature message deleted event

jpjpjp:
# Pull Request Template

## Description

Updated plugin-messages to call messages.on('deleted') when a user deletes a previously posted message.   Developers of custom clients can use this handle this event by updating their GUI to represent the deleted message

Fixes # (issue)

## Type of Change

## Test Coverage
TBD....

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -95,14 +107,34 @@ const Messages = SparkPlugin.extend({
...
-  createMessagesEventData(activity) {
+  messageCreatedEventData(activity) {
     return {
       id: constructHydraId(hydraTypes.MESSAGE, activity.id),
       roomId: constructHydraId(hydraTypes.ROOM, activity.target.id),
       roomType: activity.target.tags.includes('ONE_ON_ONE') ? 'direct' : 'group',
       text: activity.object.displayName,
       personId: constructHydraId(hydraTypes.PEOPLE, activity.actor.id),
       personEmail: activity.actor.emailAddress || activity.actor.entryEmail,
+      personOrgId: constructHydraId(hydraTypes.ORGANIZATION, activity.actor.orgId),
```
mattnorris: This isn't part of the [Hydra payload](https://developer.webex.com/docs/api/v1/messages/get-message-details) from what I can tell. We have to be careful about including things out-of-spec. Maybe there is a property object we can group things into like `experimental: {}`.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -95,14 +107,34 @@ const Messages = SparkPlugin.extend({
    * @param {Object} activity from mercury
    * @returns {Object} constructed event
    */
-  createMessagesEventData(activity) {
+  messageCreatedEventData(activity) {
```
mattnorris: Why rename? It actually is creating a new object of `messages` data.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -16,7 +16,11 @@ import {isArray} from 'lodash';
 
 const debug = require('debug')('messages');
 
-const OUTGOING_MESSAGES_CREATED = 'created';
+const MESSAGE_CREATED = 'created';
+const MESSAGE_DELETED = 'deleted';
+// This actually generates all kind of activities related events
+// Should we give this  a name that represents that
+//const CONVERSATION_ACTIVITY = 'event:conversation.activity';
```
mattnorris: 👍

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
jpjpjp: You are right.   OrgId is there on the membership, but NOT the messages.   I'll remove this.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
jpjpjp: I wanted to differentiate between the event for the message created event and the message deleted event, but I from your question can see that the create part of the name refers to what the function is doing (not the contents of the event its creating).

What about createMessageCreatedEventData and createMessageDeletedEventData?  (or a single createMessageEventData, with a second activity parameter)?

mattnorris: We're wading into Java naming territory 😉 

Let's simplify them to these, because we're already within the "messages" plugin and file: 
- `getCreatedEvent`
- `getDeletedEvent`

jpjpjp: Will do!

mattnorris:
@jpjpjp is this also handled by #1142 ?

jpjpjp:
Changes incorporated in feat_room_events PR

PR #1127, opened 2019-04-02
Add file attachments to Mercury messages

mattnorris:
# Pull Request 

## Description

Add file attachments to Mercury messages. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-66906

## Type of Change

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/common/src/uuid-utils.js
+++ b/packages/node_modules/@ciscospark/common/src/uuid-utils.js
@@ -34,7 +34,7 @@ export function deconstructHydraId(id) {
 
 /**
  * Returns file URLs for the activity, adhering to Hydra details,
- * e.g., https://api.ciscospark.com/v1/contents/Y2lzY29zcGF...",
+ * e.g., https://api.ciscospark.com/v1/contents/Y2lzY29zcGF...
  * @see https://developer.webex.com/docs/api/v1/messages/get-message-details
  * @param {Object} activity from mercury
  * @returns {Array} file URLs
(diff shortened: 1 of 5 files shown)


jpjpjp:
Looks good!   I noticed that you are building a files field for EVERY message event and then have this code in the new getHydraFiles() method:
if (!files) {
    return undefined;
  }

This results in every message that DOESN'T have a file attachment getting having and undefined "files" attribute.

Would it be better to create the message event payload without files first and then add this logic before sending it:

if (activity.files) {
    messagesEventData.files = getHydraFiles(activity);
  }

packages/node_modules/@ciscospark/common/src/uuid-utils.js
```diff
@@ -27,3 +31,29 @@ export function deconstructHydraId(id) {
...
+ */
+export function getHydraFiles(activity) {
+  const hydraFiles = [];
+  const {files} = activity.object;
+
+  if (files) {
+    const {items} = files;
+
+    // Note: Generated ID is dependent on file order.
+    for (let i = 0; i < items.length; i += 1) {
```
taymoork2: Could this be a `.forEach` or `.map` on the items array?

taymoork2: unless this is dependant on being as fast as possible

mattnorris: Unfortunately, no, it's dependent on position. Any other way has unused vars.

taymoork2: This is me being extra for the sake of being extra, but couldn't you do something like this?
```javascript
return activity.object.files
  ? activity.object.files.items.map(item =>
      hydraFiles.push(
        `${hydraBaseUrl}/contents/${constructHydraId(hydraTypes.CONTENT, `${activity.id}/${i}`)}`
      )
    )
  : [];
```

PR #1135, opened 2019-04-11
Feature membership events

jpjpjp:
# Pull Request Template

## Description

Add membership events  -- DON'T MERGE THIS.  REFER TO https://github.com/webex/spark-js-sdk/pull/1142

## Type of Change

## Test Coverage

No tests yet.  Since the goal here is to have these events match the webhooks, an ideal test would be to create a space, add some members, modify some (ie: make moderators), send and receive some read receites etc.  The framework should register for both the SDK events and the webhooks and then it can compare them.   Need to learn more about how to create users to support a test like this.

## Checklist:

- [x] My code follows the style guidelines of this project -- I think.  I got way more lint complaints on than on my previous pull.  Did the lint config for this project change recently?


packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -2,7 +2,21 @@
...
+  constructHydraId,
+  hydraTypes
+} from '@ciscospark/common';
+
+const debug = require('debug')('memberships');
+
+const MEMBERSHIP_CREATED = 'created';
+const MEMBERSHIP_DELETED = 'deleted';
+const MEMBERSHIP_UPDATED = 'updated';
+const CONVERSATION_ACTIVITY = 'event:conversation.activity';
```
mattnorris: We may want to put this in a common file now that we're using it in multiple places.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+   * Connect to the web socket to listen to incoming messages.
+   * @returns {Promise}
+   */
+  listen() {
+    // Create a common envelope that we will wrap all events in
+    this.eventEnvelope = {
+      resource: 'memberships',
+      // orgId -- TODO get from this.spark.internal.user
+      // createdBy -- TODO get from this.spark.internal.user
+      ownedBy: 'creator',
```
mattnorris: Is `'creator'` a TODO placeholder or the actual value?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+      status: 'active',
+      created: new Date().toISOString(),
```
mattnorris: Because `listen` only happens once, I would leave off `created` and add it when you handle the event. (I'll keep reading, but I'm not sure `eventEnvelope` should be a class instance.)

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+  },
+
+  /**
+   * Trigger a membership related events.
+   * @param {Object} event
+   * @returns {undefined}
+   */
+  onWebexApiEvent(event) {
+    const {activity} = event.data;
+    console.log(activity);
```
mattnorris: This should be removed. You can use `debug(activity)` if it's valuable for debugging.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+
+    // Reply activities are not currently supported
+    if (activity.type === API_ACTIVITY_TYPE.REPLY) {
```
mattnorris: We actually don't need this in either file now that I think about it. We should just act on the types we need and ignore the rest.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    }
+  },
+
+  /**
+   * Constructs event data object for the "membership created" event,
+   * adhering to Hydra's Webehook data for an "created" event on the "membership" resource.
+   * @param {Object} activity from mercury
+   * @returns {Object} constructed event
+   */
+  getCreatedEvent(activity) {
```
mattnorris: `getCreatedEvent`, `getDeletedEvent`, and `getUpdatedEvent` can be further simplified by passing a second argument to `getCommonSdkEvent`: the value you want on the `.event` property.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +38,148 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    sdkEvent.data.created = activity.published;
+    return sdkEvent;
+  },
+
+  /**
+   * Constructs common event data object for all membership events
+   * @param {Object} activity from mercury
+   * @returns {Object} constructed event
+   */
+  getCommonSdkEvent(activity) {
```
mattnorris: Is this envelope needed for the `messages` plugin too?

packages/node_modules/samples/browser-socket/app.js
```diff
@@ -69,17 +69,39 @@ document.getElementById('credentials').addEventListener('submit', (event) => {
...
             console.log(message);
           });
         })
         .catch((err) => {
           console.error(`error listening to messages: ${err}`);
           updateStatus(false);
         });
+      webex.memberships.listen()
+        .then(() => {
+          console.log('connected');
```
mattnorris: Is the intent to replace these logs with assertions at some point?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: 'creator' is the actual value.  You can see this is the value set in the envelope of most webhooks.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Agree.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Agree.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Agree.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Good point.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: It is.  Will discuss with you how to implement it there.

packages/node_modules/samples/browser-socket/app.js
jpjpjp: I was just scaffolding of your original sample.   I think we ultimately need a better sample

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: As per our discussion, leaving this on to align with the webhook envelope information.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Removed

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Simplified.

packages/node_modules/@ciscospark/common/src/constants.js
```diff
@@ -1,16 +1,45 @@
-export const API_ACTIVITY_TYPE = {
-  REPLY: 'reply'
+export const WEBEX_EVENT = {
+  TEAMS_ACTIVITY: 'event:conversation.activity'
```
andrewholsted: is having a `.` in an event standard practice? I've seen the `:` and camelCase strings.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    // Create a common envelope that we will wrap all events in
+    this.eventEnvelope = {
+      resource: SDK_EVENT_RESOURCE.MEMBERSHIPS,
+      // id -- webhook id concept does not correlate to SDK socket event
+      // name -- webhook name concept does not correlate to SDK socket event
+      // targetUrl -- targetUrl concept does not correlate to SDK socket event
+      // secret -- secret concept does not correlate to SDK socket event
+      ownedBy: SDK_EVENT_OWNER.CREATOR,
+      status: SDK_EVENT_STATUS.ACTIVE,
+      created: new Date().toISOString(),
```
andrewholsted: This will send the same `created` date for all events, no?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+      data: {}
+    };
+    const that = this;
```
andrewholsted: With `=>` you shouldn't need to assign `this` to `that`. What's the main use case here?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+
+  /**
+   * Constructs the data object for an event on the memberships resource,
+   * adhering to Hydra's Webehook data structure memberships.
+   * @param {Object} activity from mercury
+   * @param {Object} event type of "webhook" event
+   * @returns {Object} constructed event
+   */
+  getMembershipEvent(activity, event) {
+    const sdkEvent = JSON.parse(JSON.stringify(this.eventEnvelope));
```
andrewholsted: Is this just a way to clone the data? You might also consider clone or clone deep from lodash if the sdk includes it.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -1,8 +1,20 @@
...
+  API_ACTIVITY_VERB,
+  SDK_EVENT_RESOURCE,
+  SDK_EVENT,
+  SDK_EVENT_OWNER,
+  SDK_EVENT_STATUS,
+  hydraTypes,
+  constructHydraId
+} from '@ciscospark/common';
+
+const debug = require('debug')('memberships');
```
andrewholsted: What does this do? Seems like it should be for dev only?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Removed this (also from plugin-messages)

packages/node_modules/@ciscospark/common/src/constants.js
jpjpjp: This is the underlying conversations nomenclature that exists today.    We created the WEBEX_EVENT.TEAMS_ACTIVITY enums to abstract that a way a bit from the average SDK user/contributor

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Yes, this is the desired behavior as part of an effort to have the SDK Event payloads match as closely as possible to the payload of the existing webhooks.  

Today each webhook payload has an "envelope" which includes info about the webhook itself including when it was created.  In the webhook world this corresponds to the initial POST /webhooks call.  In the SDK world it corresponds to the memberships.listen() call.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Thats what I thought too, but then I found that this.evenlopeData was undefined.   I'll try to see if I can get rid of that.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Yes, happy to use another preferred method for style or efficiency.  Can you recommend an approach?

adamweeks: `const sdkEvent = Object.assign({}, this.eventEnvelope);` will create a new object for you.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    const sdkEvent = JSON.parse(JSON.stringify(this.eventEnvelope));
+    let member = '';
+
+    sdkEvent.event = event;
+    sdkEvent.data.created = activity.published;
+    sdkEvent.actorId = constructHydraId(hydraTypes.PEOPLE, activity.actor.id);
+    sdkEvent.data.roomId =
+      constructHydraId(hydraTypes.ROOM, activity.target.id);
+    sdkEvent.data.roomType =
+      activity.target.tags.includes('ONE_ON_ONE') ? 'direct' : 'group';
```
adamweeks: I thought `ONE_ON_ONE` was a constant somewhere, but maybe that is just in the widgets.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    if ((activity.object.roomProperties) && (activity.object.roomProperties.isModerator)) {
+      sdkEvent.data.isModerator = true;
+    }
+    else {
+      sdkEvent.data.isModerator = false;
+    }
+    sdkEvent.data.isMonitor = false;
+    // isRoomHidden
```
adamweeks: What is this for?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    }
+    else {
+      // For a read receipt the person is the "actor" or the one who did the reading
+      member = 'actor';
+      // And the "object" is the message that was last seen
+      sdkEvent.data.lastSeenId =
+        constructHydraId(hydraTypes.MESSAGE, activity.object.id);
+    }
+    sdkEvent.data.id = constructHydraId(hydraTypes.MEMBERSHIP,
+      `${activity[member].entryUUID}:${activity.target.id}`);
```
adamweeks: Nice reverse engineering!

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    sdkEvent.data.personId =
+      constructHydraId(hydraTypes.PEOPLE, activity[member].entryUUID);
```
adamweeks: We should verify that `activity[member]` exists before continuing, otherwise we'll get crashes.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
```diff
@@ -74,11 +73,6 @@ const Messages = SparkPlugin.extend({
   onConversationActivityEvent(event) {
     const {activity} = event.data;
 
-    // Reply activities are not supported.
```
adamweeks: How are we supporting replies now?

packages/node_modules/samples/browser-socket/app.js
```diff
@@ -66,20 +66,45 @@ document.getElementById('credentials').addEventListener('submit', (event) => {
...
         })
         .catch((err) => {
           console.error(`error listening to messages: ${err}`);
           updateStatus(false);
         });
+      webex.memberships.listen()
+        .then(() => {
+          console.log('listening to membership events');
+          updateStatus(true);
+          webex.memberships.on('created', (membership) => {
```
adamweeks: Would be awesome to use the `SDK_EVENT` constants here.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Thanks @adamweeks I tried that but when I use Object.assign the eventEnvelope gets updated.   Here's an example from a node terminal:
`
> sdkEnvelope = {foo: 'bar', data: {}} 
{ foo: 'bar', data: {} } 
> sdkEvent = Object.assign({}, sdkEnvelope) 
{ foo: 'bar', data: {} } 
> sdkEvent.data.key = "value1" 
'value1' 
> sdkEvent 
{ foo: 'bar', data: { key: 'value1' } } 
> sdkEnvelope 
{ foo: 'bar', data: { key: 'value1' } }
`
This doesn't happen when I use the JSON.stringify approach.   Is this bad/inefficient?

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: The debug module?  It provides a way for developers to see any debug output from the modules.  This is there for most of the public modules (I followed the debug model that Matt used when implementing the message event)

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +36,146 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+    return this.spark.internal.mercury.connect()
+      .then(() => this.listenTo(this.spark.internal.mercury,
+        WEBEX_EVENT.TEAMS_ACTIVITY,
+        (event) => this.onWebexApiEvent(event)));
+  },
+
+  /**
+   * Trigger a membership related events.
+   * @param {Object} event
+   * @returns {undefined}
```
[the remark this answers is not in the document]
jpjpjp: Done, thanks.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Ended up resolving this with lodash cloneDeep.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: I added it as a constant in the common module in my latest commit.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: It is a deprecated field that we still return in the webhook payload.   If @joesankey agrees we can leave it out of the payloads generated by the SDK.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Good point.  Code looks a little more complex now.  Will ask you for another review.

packages/node_modules/@ciscospark/plugin-messages/src/messages.js
jpjpjp: We still aren't  Matt just convinced himself that we didn't need to have this line of code anymore.

packages/node_modules/samples/browser-socket/app.js
jpjpjp: Good idea.  Updated.

mattnorris:
See #1142 instead.

PR #1142, opened 2019-04-20
Feat membership events

jpjpjp:
# Pull Request Template

## Description

Add membership events -- new PR with cherry picked commits thanks to help from @adamweeks   Should address all comments from https://github.com/webex/spark-js-sdk/pull/1135

New events:
- membership:created
- membership:deleted
- membership:updated  -- change in moderator status or isRoomHidden status
- membership:seen -- read receipt

Event DTO is as similar as possible to the membership webhooks

## Type of Change

## Test Coverage

No tests yet. Since the goal here is to have these events match the webhooks, an ideal test would be to create a space, add some members, modify some (ie: make moderators), send and receive some read receites etc. The framework should register for both the SDK events and the webhooks and then it can compare them. Need to learn more about how to create users to support a test like this.

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
+++ b/packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
@@ -2,7 +2,21 @@
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
 
-import {SparkPlugin, Page} from '@ciscospark/spark-core';
+import { SparkPlugin, Page } from '@ciscospark/spark-core';
+import {
+  WEBEX_API_EVENTS,
+  API_ACTIVITY_TYPE,
+  API_ACTIVITY_VERB,
+  constructHydraId,
+  hydraTypes
+} from '@ciscospark/common';
+
+const debug = require('debug')('memberships');
+
+const MEMBERSHIP_CREATED = 'created';
+const MEMBERSHIP_DELETED = 'deleted';
+const MEMBERSHIP_UPDATED = 'updated';
+const CONVERSATION_ACTIVITY = 'event:conversation.activity';
 
 /**
  * @typedef {Object} MembershipObject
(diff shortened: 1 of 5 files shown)


packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -1,8 +1,23 @@
...
+  SDK_EVENT_RESOURCE,
+  SDK_EVENT,
+  SDK_EVENT_OWNER,
+  SDK_EVENT_STATUS,
+  hydraTypes,
+  constructHydraId
+} from '@ciscospark/common';
+
+const debug = require('debug')('memberships');
+const _ = require('lodash');
```
mattnorris: Instead of importing all of lodash, you can import just what you need, `cloneDeep`. This helps keep the package size down.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +39,177 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+      // id -- webhook id concept does not correlate to SDK socket event
+      // name -- webhook name concept does not correlate to SDK socket event
+      // targetUrl -- targetUrl concept does not correlate to SDK socket event
+      // secret -- secret concept does not correlate to SDK socket event
+      ownedBy: SDK_EVENT_OWNER.CREATOR,
+      status: SDK_EVENT_STATUS.ACTIVE,
+      created: new Date().toISOString(),
+      data: {}
+    };
+    const that = this;
```
mattnorris: You don't need to do this with arrow functions. See the section ["Functionality: lexical scoping “this”"](https://ultimatecourses.com/blog/es6-arrow-functions-syntaxes-and-lexical-scoping).

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +39,177 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+        that.eventEnvelope.createdBy = person.id;
+        that.eventEnvelope.orgId = person.orgId;
+      }).catch((e) => {
+        console.error(`Unable to get person info for memberships \
+          event envelope ${e.message}`);
+      });
+
+    // Register to listen to events
+    return this.spark.internal.mercury.connect()
+      .then(() => this.listenTo(this.spark.internal.mercury,
```
mattnorris: The carriage returns here will probably get flagged by linting.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +39,177 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+   * Constructs the data object for an event on the memberships resource,
+   * adhering to Hydra's Webehook data structure memberships.
+   * @param {Object} activity from mercury
+   * @param {Object} event type of "webhook" event
+   * @returns {Object} constructed event
+   */
+  getMembershipEvent(activity, event) {
+    const sdkEvent = _.cloneDeep(this.eventEnvelope);
+    let member = '';
+    let space = '';
```
mattnorris: Is there a case where it's valid for these to be empty strings? If not, declare as `let member;`, etc.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
```diff
@@ -24,6 +39,177 @@ import {SparkPlugin, Page} from '@ciscospark/spark-core';
...
+        constructHydraId(hydraTypes.MESSAGE, activity.object.id);
+    }
+    else if (activity.verb === API_ACTIVITY_VERB.HIDE) {
+      // For a hide activity the person is also the "actor"
+      member = API_ACTIVITY_FIELD.ACTOR;
+      space = API_ACTIVITY_FIELD.OBJECT; // but the space is now the "object"
+    }
+    else {
+      // For most memberships events the person is the 'object" or the one whose membership it is
+      member = API_ACTIVITY_FIELD.OBJECT;
```
mattnorris: Great explanations 👍

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Will do.  Andrew suggested lodash as opposed to a library free use of JSON.stringify for a deep copy.  I found other spots in the code that import lodash already so thought this approach would not add any package size.

grep -r "lodash" *.js
deps.js:const _ = require('lodash');

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Actually, I added it because the linter complained about the line length.   (This is actually what your code looks like in messages.js)

Hopefully the new multi-line brace function will make all parties happy.

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Will do

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
jpjpjp: Removed

packages/node_modules/@ciscospark/plugin-memberships/src/memberships.js
mattnorris: Got it. If it's already included, that's fine then.

PR #1167, opened 2019-05-09
Add ability to enable & disable presence

mattnorris:
# Pull Request 

## Description

Add ability to enable & disable presence and a convenience method to see its state. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-53219

## Type of Change

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-presence/src/presence.js
+++ b/packages/node_modules/@webex/internal-plugin-presence/src/presence.js
@@ -23,16 +23,28 @@ const Presence = SparkPlugin.extend({
     batcher: PresenceBatcher
   },
 
+  /**
+   * Enables presence feature
+   * @returns {Promise<boolean>} resolves with true, if successful
+   */
   enable() {
     return this.spark.internal.feature.setFeature(USER, USER_PRESENCE_ENABLED, true)
       .then((response) => response.value);
   },
 
+  /**
+   * Disables presence feature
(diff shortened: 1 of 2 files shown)


packages/node_modules/@ciscospark/internal-plugin-presence/test/integration/spec/presence.js
```diff
@@ -40,6 +40,29 @@ describe('plugin-presence', function () {
 
     afterEach('deregister spock with mercury', () => spock.spark.internal.mercury.disconnect());
 
+    describe('#isEnabled()', () => {
```
adamweeks: I would prefer to see this as its own test and verifying from the features instead of relying on the other  `#enable` and `#disable` functions. As well as having the `#enable` test not rely on `#isEnabled`.  It feels somewhat self verifying instead of unit testing each function. Not really a blocker though, just more of a personal preference.

mattnorris: @adamweeks so even though they're really helper methods that call the features methods, you would explicitly use the features methods in the tests?

mattnorris:
@adamweeks please take another 👀  at the tests

adamweeks:
> @adamweeks please take another 👀 at the tests

@mattnorris looks good!

PR #1175, opened 2019-05-16
feat(internal-plugin-conversation): reflect thread server change

willzxd:
# Pull Request Template

## Description

Reflect server changes.

When posting an activity, we will add `parent` object in the `activity` object.

## Type of Change

## Test Coverage

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
@@ -442,7 +442,7 @@ const Conversation = SparkPlugin.extend({
           objectType: 'activity',
           clientTempId: uuid.v4(),
           actor: this.spark.internal.device.userId
-        }, pick(activity, 'activityType', 'parentActivityId'));
+        }, pick(activity, 'activityType', 'parentActivityId', 'parent'));
 
         if (isString(act.actor)) {
           act.actor = {


packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -442,7 +442,7 @@ const Conversation = SparkPlugin.extend({
           objectType: 'activity',
           clientTempId: uuid.v4(),
           actor: this.spark.internal.device.userId
-        }, pick(activity, 'activityType', 'parentActivityId'));
+        }, pick(activity, 'activityType', 'parentActivityId', 'parent'));
```
mattnorris: @willzxd I actually just updated the test for this. Did you verify it works with this change?

willzxd: I saw that. 

Current server behavior is client side is able to post thread replies that contain `activityType` and `parentAcitivtyId` and/or `parent`. 
1 - 2 month later server behavior will change. It will only accept post thread replies that contain `parent`.
The server side will always return responses that contain `parent`.

So we need to be compatible with both to avoid breaking changes.
Updated wiki here: https://wiki.cisco.com/pages/viewpage.action?pageId=236794024

willzxd:
Screen Shot 2019-05-20 at 1 52 33 PM
pass locally.

PR #1179, opened 2019-05-21
fix(ediscovery): adding bulk support for emails to uuid conversion

decryan1984:
## Description
Previously, when we needed to convert an email address to a user id we called an API in the People plugin for each email we required. For large amounts of emails (the client currently allows up to 500 emails as parameters in each report), this resulted in many requests and some of those requests were rate-limited and received 429 responses. This fix switches over to a SCIM api in hydra which allows batching of emails in a single request. It can process 50 emails in a single request, so we may still have to make multiple requests, but not on a scale which we expect to cause issues.

Also making changes to the way some errors are logged/reported during some flows in the plugin.

## Fixes
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-72848

## Test Coverage

This scenario is currently only covered by test scripts that are run in the load test environment

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
+++ b/packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
@@ -2,7 +2,7 @@
 /*!
  * Copyright (c) 2015-2017 Cisco Systems, Inc. See LICENSE file.
  */
-
+import {InvalidEmailAddressError} from './ediscovery-error';
 import {waitForValue, SparkPlugin} from '@webex/webex-core';
 import {oneFlight} from '@webex/common';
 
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -219,6 +219,94 @@ const EDiscovery = SparkPlugin.extend({
...
+      throw Error('The maximum number of emails that can be converted is 500');
+    }
+
+    let orgId = '';
+
+    // find the org id for the compliance officer, as this is required for the SCIM request
+    return this.spark.people.get('me')
+      .then((res) => {
+        // Org id is base64 encoded and of the format Y2lzY29zcGFyazovL3VzL1BFT1BMRS81ZDU5Yjc5NS02ZmEyLTQ2NTQtOGVjMi03NjlkYjE1YzBkOWU
+        const decodedOrgId = Buffer.from(res.orgId, 'base64').toString();
```
mattnorris: There are some base64 utilities at `@webex/common/base64.js` you can use.

decryan1984:
I don't believe adding to our existing unit tests will add much value here Matt. Because of the nature of the SCIM API, we've mainly been conducting tests in our integration and load test environment.

PR #1193, opened 2019-06-03
fix(media): find content mediaShare

robstax:
## Description
previously, when attempt to grab the "content" mediaShare, we assumed it was at the `[0]` index. the order of these mediaShares however is not guaranteed to have content at `[0]`, so instead of `[0]`
use `find`. this meant that when emitting a `MEMBERS_CONTENT_UPDATE` event, we were not correctly reporting who ended/started sharing.

## Type of Change
bug fix

## Test Coverage
Take over someone else's share. Check the payload from the `MEMBERS_CONTENT_UPDATE` event. instead of `null` values, it will have member ids as expected.

## Checklist:


robstax:
@chrisadubois ideally, i would like to keep them separate.

PR #1201, opened 2019-06-07
Adding retry handling for rate limiting and timeouts to ediscovery plugin

decryan1984:
## Description

Adding retry handling for rate limiting and timeouts to ediscovery plugin. The retry behaviour can be configured, but by default it will attempt 3 retries, with exponentially increasing delays of 4, 9 and 16 seconds.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-71491

## Type of Change

New feature (non-breaking change which adds functionality)

## Test Coverage

Manual testing required simulation of errors that trigger retry mechanism

## Checklist:


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -221,6 +222,25 @@ const EDiscovery = SparkPlugin.extend({
...
+      throw Error('Undefined parameter');
+    }
+
+    return this.request({
+      method: 'GET',
+      service: 'hydra',
+      resource: `scim/${orgId}/Users?filter=${filter}`
+    });
+  },
+
```
decryan1984: Isolating the hydra request from the batching and error handling logic

packages/node_modules/@webex/internal-plugin-ediscovery/src/retry.js
```diff
@@ -0,0 +1,34 @@
...
+          retryIntervalInSeconds = reason.headers['retry-after'];
+        }
+        console.error(`Request #${retryCount} error: ${reason.statusCode}. Attempting retry #${retryCount} in ${retryIntervalInSeconds} seconds`);
+
+        return requestWithRetries(ctx, func, args, retryCount, retryIntervalInSeconds, maxRetries);
+      }
+
+      return Promise.reject(reason);
+    });
+}
```
decryan1984: New reusable function that takes a request and can call it multiple times if rate-limited, or the service is unresponsive.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -221,6 +222,25 @@ const EDiscovery = SparkPlugin.extend({
...
+  @waitForValue('@')
+  /**
+   * Retrieves user information from CI based on a query filter
+   * @param {UUID} orgId - Id of the organisation for which users are being requested
+   * @param {String} filter - Query containing information to match users on (id, username, displayName etc.)
+   * @returns {Promise<ResponseEntity<Resources>>} Http response containing the user information
+   */
+  getUsers(orgId, filter = '') {
+    if (!orgId) {
+      throw Error('Undefined parameter');
```
mattnorris: Instead of a generic parameter message, would this be more helpful as "organization ID needed to get users"?

packages/node_modules/@webex/internal-plugin-ediscovery/src/retry.js
mattnorris: There is a retry in `packages/node_modules/@webex/common/src/retry.js`. Can this core one be used or modified to keep retry logic in one place?

decryan1984: I don't think it can Matt, it doesn't seem to provide everything we need in terms of retrying various responses status codes and error handling. It doesn't appear to be used anywhere else in the sdk, and I couldn't find any tests to illustrate how it should be used, but I'm also reluctant to modify it. I wasn't aware this class existed, I was under the impression that each plugin had to handle retries in their own way, e.g. plugin-meetings.

PR #1219, opened 2019-06-24
Meetings samples conversion

adamweeks:
# Meetings Samples Conversion

## Description

Updating the phone plugin samples to be meetings plugin samples.

## Type of Change

### DIFF
--- a/packages/node_modules/samples/browser-single-party-call/index.html
+++ b/packages/node_modules/samples/browser-single-party-call/index.html
@@ -57,7 +57,7 @@
     </div>
 
     <!-- /bundle.js is the bundled Webex JavaScript SDK -->
-    <script src="/bundle.js"></script>
+    <script src="/meetings.bundle.js"></script>
     <!-- app.js is your application code -->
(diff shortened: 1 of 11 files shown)


packages/node_modules/samples/browser-single-party-call/index.html
```diff
@@ -57,7 +57,7 @@ <h1>Single Party Calling</h1>
     </div>
 
     <!-- /bundle.js is the bundled Webex JavaScript SDK -->
-    <script src="/bundle.js"></script>
+    <script src="/meetings.bundle.js"></script>
```
taymoork2: Instead of doing `meetings.bundle.js`, we need to get rid of it.
Make `meetings.bundle.js` --> `bundle.js`

taymoork2: I can make a git patch and you can apply it.
Probably easier than writing out what needs to be replaced

adamweeks:
Closed due to local branch instead of fork. See #1234 instead.

PR #1223, opened 2019-06-25
feat(feature): handle feature update event from mercury

aimeex3:
handle event:featureToggle_update from mercury

# Pull Request Template

## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [x] hooked up web-client with local sdk build, changed global notifications settings, tested that the mercury event gets triggered, and updates the feature collection

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-feature/src/feature.js
+++ b/packages/node_modules/@webex/internal-plugin-feature/src/feature.js
@@ -38,6 +38,34 @@ const Feature = WebexPlugin.extend({
   },
 
   /**
+   * Handles a feature toggle update from the server.
+   * @param {Object} envelope
+   * @returns {undefined}
+   */
+  handleFeatureUpdate(envelope) {
+    if (envelope && envelope.data) {
+      const feature = envelope.data.featureToggle;
+      const keyType = feature.type.toLowerCase();
+
+      if (keyType === 'user' || keyType === 'developer') {
+        this.webex.internal.device.features[keyType].add([feature], {merge: true});
+      }
+    }
+  },
+
+  /**
+   * Register to listen for incoming feature events
+   * @instance
+   * @returns {Promise}
+   */
+  listen() {
+    return this.webex.internal.mercury.connect()
+      .then(() => {
+        this.listenTo(this.webex.internal.mercury, 'event:featureToggle_update', this.handleFeatureUpdate);
+      });
+  },
+
+  /**
    * Issues request to server to set a value for a feature toggle.
    * @param {string} keyType <developer|user>
    * @param {string} key
--- a/packages/node_modules/@webex/internal-plugin-feature/src/index.js
+++ b/packages/node_modules/@webex/internal-plugin-feature/src/index.js
@@ -1,8 +1,10 @@
 /*!
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
+import '@webex/internal-plugin-mercury';
 
 import {registerInternalPlugin} from '@webex/webex-core';
+
 import Feature from './feature';
 import config from './config';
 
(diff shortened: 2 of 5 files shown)


packages/node_modules/@webex/internal-plugin-feature/src/feature.js
```diff
@@ -37,6 +37,22 @@ const Feature = WebexPlugin.extend({
     return Promise.resolve(feature.value);
   },
 
+  /**
+   * Issues request to server to set a value for a feature toggle.
```
andrewholsted: is this jsdoc correct?

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
```diff
@@ -37,6 +37,22 @@ const Feature = WebexPlugin.extend({
...
+   * @param {Object} envelope
+   * @returns {Promise} Refreshes the local device and resolves with the features endpoint's response.
+   */
+  handleFeatureUpdate(envelope) {
+    if (envelope && envelope.data) {
+      const feature = cloneDeep(envelope.data.featureToggle);
```
andrewholsted: why do we need to clone here? Do we mutate the object somewhere else?

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
```diff
@@ -93,6 +109,10 @@ const Feature = WebexPlugin.extend({
   initialize(...args) {
     Reflect.apply(WebexPlugin.prototype.initialize, this, args);
 
+    this.webex.once('ready', () => {
```
andrewholsted: Is it possible that the `ready` event has fired before this plugin has initialized? I don't think so, but wanted to check.

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
aimeex3: oops forgot to update. thanks for catching

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
aimeex3: yeah when it gets added to the feature collection, it parses the object and mutates it. I see 2 events with the same id come through here. Not sure why there's 2 and the second one is the one that gets mutated

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
aimeex3: not sure, i thought ready meant that the plugins have been initialized. is there another event to check explicitly that this plugin is initialized? @mattnorris @taymoork2?

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
aimeex3: i see where it's getting emitted twice: https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/internal-plugin-mercury/src/mercury.js#L442-L443

I think maybe server usually sends one or the other, but this looks like it sends both so it's the same object

aimeex3: should i add a check in mercury.js, if the `namespace` === `data.eventType`, only emit 1 event? @mattnorris @taymoork2 @adamweeks

adamweeks: @aimeex3 I think that should work!

packages/node_modules/@webex/internal-plugin-feature/src/feature.js
taymoork2: `plugin-meetings` does use `this.webex.once('ready'...` internally
Unless matt has something to add, I'd ask @arun3528 why they used it internally

mattnorris: `ready` fires after *all* the plugins have initialized. 

Is there a reason not to use `listenToAndRun` like below?

aimeex3: hmm even if `this.webex.internal.mercury` was not defined yet, `listenToAndRun` would work? I think mercury wasn't defined yet so i wanted to wait for it to be initialized

aimeex3:
merged

PR #1229, opened 2019-06-28
feat(convo): update convo specific APIs

mwegman:
- Instead of going to the generic convo server for convo specific
  actions, use the convo specific API.

# Pull Request Template

## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
@@ -14,25 +14,34 @@ import uuid from 'uuid';
 import {InvalidUserCreation} from './convo-error';
 import ShareActivity from './share-activity';
 
+const idToUrl = new Map();
 
 const Conversation = WebexPlugin.extend({
   namespace: 'Conversation',
 
+  /**
+   * @param {Object} conversation
+   * @param {Object} object
+   * @param {Object} activity
+   * @returns {Promise}
+   */
   acknowledge(conversation, object, activity) {
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
+
     if (!isObject(object)) {
       return Promise.reject(new Error('`object` must be an object'));
     }
 
-    return this._inferConversationUrl(conversation)
-      .then(() => this.prepare(activity, {
-        verb: 'acknowledge',
-        target: this.prepareConversation(conversation),
-        object: {
-          objectType: 'activity',
-          id: object.id,
-          url: object.url
-        }
-      }))
+    return this.prepare(activity, {
+      verb: 'acknowledge',
+      target: this.prepareConversation(convoWithUrl),
+      object: {
+        objectType: 'activity',
+        id: object.id,
+        url: object.url
+      }
+    })
       .then((a) => this.submit(a));
   },
 
@@ -46,11 +55,13 @@ const Conversation = WebexPlugin.extend({
    * @returns {Promise<Activity>}
    */
   add(conversation, participant, activity) {
-    return this._inferConversationUrl(conversation)
-      .then(() => this.webex.internal.user.asUUID(participant, {create: true}))
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
+
+    return this.webex.internal.user.asUUID(participant, {create: true})
       .then((id) => this.prepare(activity, {
         verb: 'add',
-        target: this.prepareConversation(conversation),
+        target: this.prepareConversation(convoWithUrl),
         object: {
           id,
           objectType: 'person'
@@ -130,17 +141,26 @@ const Conversation = WebexPlugin.extend({
       });
   },
 
+  /**
+   * delete content
+   * @param {Object} conversation
+   * @param {Object} object
+   * @param {Object} activity
+   * @returns {Promise}
+   */
   delete(conversation, object, activity) {
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
+
     if (!isObject(object)) {
       return Promise.reject(new Error('`object` must be an object'));
     }
 
-    return this._inferConversationUrl(conversation)
-      .then(() => this.prepare(activity, {
-        verb: 'delete',
-        target: this.prepareConversation(conversation),
-        object: pick(object, 'id', 'url', 'objectType')
-      }))
+    return this.prepare(activity, {
+      verb: 'delete',
+      target: this.prepareConversation(convoWithUrl),
+      object: pick(object, 'id', 'url', 'objectType')
+    })
       .then((a) => this.submit(a));
   },
 
@@ -289,16 +309,27 @@ const Conversation = WebexPlugin.extend({
   /**
    * Fetches a single conversation
    * @param {Object} conversation
+   * @param {String} [conversation.url] The URL where the conversation is located.
+   * @param {String|UUID} [conversation.user] The user to look up in the conversation service
+   *   If specified, the user lookup will take precedence over the url lookup
    * @param {Object} options
    * @returns {Promise<Conversation>}
    */
-  get(conversation, options) {
-    return this._inferConversationUrl(conversation)
-      .then(() => {
-        const {user, url} = conversation;
+  get(conversation, options = {}) {
+    const {user} = conversation;
+    let urlPromise;
 
-        options = options || {};
+    try {
+      const url = !user ? this.getConvoUrl(conversation) : '';
 
+      urlPromise = Promise.resolve({url});
+    }
+    catch (noUrl) {
+      urlPromise = this._inferConversationUrl(conversation);
+    }
+
+    return urlPromise
+      .then(({url}) => {
         const params = {
           qs: Object.assign({
             uuidEntryFormat: true,
@@ -348,7 +379,10 @@ const Conversation = WebexPlugin.extend({
    * @returns {Promise<Activity>}
    */
   leave(conversation, participant, activity) {
-    return this._inferConversationUrl(conversation)
+    const convoWithUrl =
+      Object.assign({}, conversation, {url: this.getConvoUrl(conversation)});
+
+    return Promise.resolve()
       .then(() => {
         if (!participant) {
           participant = this.webex.internal.device.userId;
@@ -357,7 +391,7 @@ const Conversation = WebexPlugin.extend({
         return this.webex.internal.user.asUUID(participant)
           .then((id) => this.prepare(activity, {
             verb: 'leave',
-            target: this.prepareConversation(conversation),
+            target: this.prepareConversation(convoWithUrl),
             object: {
               id,
               objectType: 'person'
@@ -387,7 +421,14 @@ const Conversation = WebexPlugin.extend({
       resource: 'conversations',
       qs: omit(options, 'deferDecrypt'),
       deferDecrypt: options.deferDecrypt
-    });
+    })
+      .then((results) => {
+        for (const convo of results) {
+          idToUrl.set(convo.id, convo.url);
+        }
+
+        return results;
+      });
   },
 
   /**
@@ -401,16 +442,24 @@ const Conversation = WebexPlugin.extend({
       service: 'conversation',
       resource: 'conversations/left',
       qs: options
-    });
+    })
+      .then((results) => {
+        for (const convo of results) {
+          idToUrl.set(convo.id, convo.url);
+        }
+
+        return results;
+      });
   },
 
   /**
    * List activities for the specified conversation
    * @param {Object} options
+   * @param {String} options.conversationUrl URL to the conversation
    * @returns {Promise<Array<Activity>>}
    */
   listActivities(options) {
-    return this._listActivities(Object.assign(options, {mentions: false}));
+    return this._listActivities(Object.assign(options, {resource: 'activities'}));
   },
 
   /**
@@ -419,7 +468,11 @@ const Conversation = WebexPlugin.extend({
    * @returns {Promise<Array<Activity>>}
    */
   listMentions(options) {
-    return this._listActivities(Object.assign(options, {mentions: true}));
+    return this._list({
+      service: 'conversation',
+      resource: 'mentions',
+      qs: omit(options, 'mentions')
+    });
   },
 
   /**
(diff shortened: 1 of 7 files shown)


mattnorris:
@mwegman is this version still relevant? Or should we close this in lieu of the greater changes we spoke of.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -593,6 +648,31 @@ const Conversation = WebexPlugin.extend({
...
 
+  /**
+   * Get url from convo object. If there isn't one, get it from the cache
+   *
+   * @param {Object} convo convo object with either a url or id
+   * @returns {String} url for the specific convo
+   */
+  getConvoUrl(convo) {
+    if (convo.url) {
+      return convo.url;
```
mattnorris: Does it make sense to update the cache here? It may not have been set earlier.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -698,10 +780,13 @@ const Conversation = WebexPlugin.extend({
    * @returns {Promise}
    */
   unassign(conversation, activity) {
-    return this._inferConversationUrl(conversation)
+    const url = this.getConvoUrl(conversation);
```
mattnorris: These two lines are repeated several times. It may make sense to encapsulate them in a method.

packages/node_modules/@webex/internal-plugin-team/src/team.js
```diff
@@ -2,9 +2,10 @@
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
 
+import querystring from 'querystring';
```
mattnorris: Why move this?

packages/node_modules/@webex/internal-plugin-team/src/team.js
```diff
@@ -95,14 +96,16 @@ const Team = WebexPlugin.extend({
    * @param {boolean} options.includeAllTeamMembers
    * @returns {Promise} Resolves with the newly created conversation
    */
-  createConversation(team, params, options) {
-    options = options || {};
+  createConversation(team, params, options = {}) {
```
mattnorris: 👍

packages/node_modules/@webex/internal-plugin-team/src/team.js
```diff
@@ -123,39 +126,43 @@ const Team = WebexPlugin.extend({
...
    * @param {TeamObject} team
+   * @param {String} team.url
    * @returns {Promise} Resolves with an array of conversations
    */
-  listConversations(team) {
-    return this._inferTeamUrl(team)
-      .then(() => this.webex.request({
-        uri: `${team.url}/conversations`
-      }))
+  listConversations({url}) {
```
mattnorris: Interesting; I haven't seen this in a method header before.

packages/node_modules/@webex/internal-plugin-team/src/team.js
mwegman: I did an `npm --fix` in this file. According to the eslint rule, this line needed to move

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: That's a good idea. If the user passes in an `id` and a `url`, we could update the cache

packages/node_modules/@webex/internal-plugin-team/src/team.js
mwegman: `url` is the only property I needed from the object.

packages/node_modules/@webex/internal-plugin-team/src/team.js
mattnorris: We must bow to our robot overlords 🤖

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -14,19 +14,29 @@ import uuid from 'uuid';
...
   acknowledge(conversation, object, activity) {
+    const url = this.getConvoUrl(conversation);
+    const space = Object.assign({}, conversation, {url});
+
     if (!isObject(object)) {
       return Promise.reject(new Error('`object` must be an object'));
     }
 
-    return this._inferConversationUrl(conversation)
+    return Promise.resolve()
```
andrewholsted: Is server side HA coming at the same time as federation or is that a separate initiative? cc @yalo as this makes the bug you are researching not relevant.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -593,6 +661,35 @@ const Conversation = WebexPlugin.extend({
...
+  /**
+   * Get url from convo object. If there isn't one, get it from the cache
+   *
+   * @param {Object} convo convo object with either a url or id
+   * @returns {String} url for the specific convo
+   */
+  getConvoUrl({id, url}) {
+    if (url) {
+      if (id) {
+        idToUrl.set(id, url);
```
andrewholsted: Kind of weird naming on `getConvoUrl` when it's also a setter?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: I was told the `url` in the DTO points towards the server-side HA version of the host. I see it points toward `conv-a` instead of `conv.a6` or `conv.a7`

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: Updating the cache was a last minute side effect to the `getConvoUrl`. I can move this out if we don't like side effects.

adamweeks: Yea, I'd prefer it to be specifically set

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -14,19 +14,29 @@ import uuid from 'uuid';
...
 
+  /**
+   * @param {Object} conversation
+   * @param {Object} object
+   * @param {Object} activity
+   * @returns {Promise}
+   */
   acknowledge(conversation, object, activity) {
+    const url = this.getConvoUrl(conversation);
+    const space = Object.assign({}, conversation, {url});
```
adamweeks: The `space` var name is a bit jarring to me, personally. It doesn't seem to fit with the `conversation` plugin. Perhaps something like `conversationWithUrl` would be more readable.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: So everywhere `getConvoUrl` is called, immediately call `updateConvoUrlCache`?

adamweeks: @mwegman  I see that @mattnorris actually requested you to do it this way, so nevermind my request

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -14,19 +14,29 @@ import uuid from 'uuid';
...
 import ShareActivity from './share-activity';
 
+const idToUrl = new Map();
 
 const Conversation = WebexPlugin.extend({
   namespace: 'Conversation',
 
+  /**
+   * @param {Object} conversation
+   * @param {Object} object
```
adamweeks: What is this object? Could you add a bit more info?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -14,19 +14,29 @@ import uuid from 'uuid';
...
+   * @param {Object} activity
+   * @returns {Promise}
+   */
   acknowledge(conversation, object, activity) {
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
```
adamweeks: Sorry, I didn't notice this before, but it seems like the `convoWithUrl` is used a lot right before calling `prepareConversation`. Does it make sense to move the `getConvoUrl` to within `prepareConversation`?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -46,11 +56,14 @@ const Conversation = WebexPlugin.extend({
    * @returns {Promise<Activity>}
    */
   add(conversation, participant, activity) {
-    return this._inferConversationUrl(conversation)
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
+
+    return Promise.resolve()
```
adamweeks: Since `this.webex.internal.user.asUUID` returns a promise, you could omit this line and just return:
```js
return this.webex.internal.user.asUUID
```

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -130,15 +143,25 @@ const Conversation = WebexPlugin.extend({
       });
   },
 
+  /**
+   * delete content
+   * @param {Object} conversation
+   * @param {Object} object
```
adamweeks: I still don't know what this `object` object is.
img

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -255,13 +278,21 @@ const Conversation = WebexPlugin.extend({
...
-    return this._inferConversationUrl(conversation)
-      .then(() => {
-        const {user, url} = conversation;
+  get(conversation, options = {}) {
+    const {user} = conversation;
+    let urlPromise;
 
-        options = options || {};
+    try {
+      const url = !user ? this.getConvoUrl(conversation) : '';
```
adamweeks: Probably need to add to the jsdoc what all the different formats of conversation this method accepts.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -736,15 +839,25 @@ const Conversation = WebexPlugin.extend({
...
   },
 
+  /**
+   * Update an existing activity
+   * @param {Object} conversation
+   * @param {Object} object
+   * @param {Object} activity
+   * @returns {Promise}
+   */
   update(conversation, object, activity) {
```
adamweeks: We sure love our ambiguous `object` in the SDK. I know it isn't your naming convention, but probably could use some more help in the jsdoc at least

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -902,15 +1021,21 @@ const Conversation = WebexPlugin.extend({
...
-    return this._list({
-      service: 'conversation',
-      resource: options.mentions ? 'mentions' : 'activities',
-      qs: omit(options, 'mentions')
-    });
+    const id = options.conversationId;
+    const url = this.getConvoUrl({url: options.conversationUrl, id});
+    const {resource} = options;
+
+    return Promise.resolve()
```
adamweeks: Could do `return this._list` since `_list` returns a promise.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1054,10 +1179,14 @@ const Conversation = WebexPlugin.extend({
   'unmute'
 ].forEach((verb) => {
   Conversation.prototype[verb] = function submitSimpleActivity(conversation, activity) {
-    return this._inferConversationUrl(conversation)
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = this.prepareConversation(Object.assign({}, conversation, {url}));
+
+    return Promise.resolve()
```
adamweeks: `this.prepare` returns a promise, so we could just return that instead.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1102,10 +1234,13 @@ const Conversation = WebexPlugin.extend({
       return Promise.reject(new Error('`tag` must be a string'));
     }
 
-    return this._inferConversationUrl(conversation)
+    const url = this.getConvoUrl(conversation);
+    const convoWithUrl = Object.assign({}, conversation, {url});
+
+    return Promise.resolve()
```
adamweeks: `this.prepare` returns a promise, so we could just return that instead.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1124,9 +1259,10 @@ const Conversation = WebexPlugin.extend({
       return Promise.reject(new Error('`object` must be an object'));
     }
 
-    const c = this.prepareConversation(conversation);
+    const url = this.getConvoUrl(conversation);
+    const c = this.prepareConversation(Object.assign({}, conversation, {url}));
 
-    return this._inferConversationUrl(conversation)
+    return Promise.resolve()
```
adamweeks: `this.prepare` returns a promise, so we could just return that instead.

packages/node_modules/@webex/internal-plugin-team/src/team.js
```diff
@@ -95,14 +96,16 @@ const Team = WebexPlugin.extend({
...
+    if (!team.url) {
+      throw Error('`team.url` is required');
+    }
 
     if (!params.displayName) {
       return Promise.reject(new Error('`params.displayName` is required'));
     }
 
-    return this._inferTeamUrl(team)
+    return Promise.resolve()
```
adamweeks: Don't need the Promise here, could just return `this._ensureGeneralConversation`

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: I was trying to keep the lines short but I can move it within `convoWithUrl`

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mwegman: I added this to remove the warnings, I can remove them if we don't want partial comments

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mattnorris: @adamweeks @mwegman we don't have to include this to avoid the side effects. I thought it might make sense to update the cache at this point for convenience, but it may be too magical and go against the grain of a getter.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
mattnorris: @adamweeks is it necessary to specify each type? Or is it okay as long as it conforms to this shape, i.e., `url`?

adamweeks: @mattnorris  Just that it needs either an `id` or `url`. Also, it looks like if the conversation object has a `user` property, it sets the url to empty so it would be good to have that in the docs as a way this method supports it as well.

mwegman: I didn't add the `id` parameter because that is getting phased out. One shouldn't use a regular convo `id` to fetch a conversation

PR #1260, opened 2019-07-18
fix(meeting): removed console logs

akoushke:
Removed extra logs from the meeting plugin.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
@@ -548,8 +548,7 @@ MeetingUtil.setPeerConnectionEvents = (meeting) => {
   // In case ICE fail
   const {peerConnection} = meeting.mediaProperties;
 
-  peerConnection.oniceconnectionstatechange = (ev) => {
-    console.log('ICE CHANGE ', ev);
+  peerConnection.oniceconnectionstatechange = () => {
     switch (peerConnection.iceConnectionState) {
       case ICE_STATE.CHECKING:
         Metrics.postEvent({event: eventType.ICE_START, meeting});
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -140,7 +140,6 @@ export default class Meetings extends WebexPlugin {
       this.meetingCollection.getByKey(CORRELATION_ID, MeetingsUtil.checkForCorrelationId(this.webex.internal.device.url, data.locus)) ||
       this.meetingCollection.getByKey(SIP_URI, data.locus.self && data.locus.self.callbackInfo && data.locus.self.callbackInfo.callbackAddress);
 
-      console.log('LOCUS EVENT ', JSON.stringify(data));
```
arun3528: Looks good to me . In future we might have to put more debug logs here for triaging . Sometimes due to a server event client might end call or stop media . To keep track of it we need to check the mercury event coming in

mattnorris: We can use `debug()`. It's very helpful and controlled via command line environment variable.

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -548,8 +548,7 @@ MeetingUtil.setPeerConnectionEvents = (meeting) => {
   // In case ICE fail
   const {peerConnection} = meeting.mediaProperties;
 
-  peerConnection.oniceconnectionstatechange = (ev) => {
-    console.log('ICE CHANGE ', ev);
```
chrisadubois: oops must've missed these last 2 when I did the cleanup. You can also send them to LoggerProxy.trace, which is kind of what I would prefer, but both of these logs are fairly noisy.

PR #1270, opened 2019-07-29
fix(ediscovery): adding optional timeout parameters

willward:
# Pull Request Template

## Description

In the ediscovery plugin I'm adding an optional parameter to each API to allow a timeout to be set, the optional parameter has a default value of 10 seconds

Fixes # (issue)
SPARK-83465 "ETIMEDOUT and then next download hung"

## Type of Change

## Test Coverage

Ran ediscovery tests using this command : 
```
 npm test -- --packages @webex/internal-plugin-ediscovery --node
```
added new tests in content.js and report.js of the ediscovery plugin tests: 
- GetContent timeout defaults to 10s
- GetContent timeout defaults to 10s when other options are specified
- GetContent timeout can be overwritten to 5s with timeout as the only option
- GetContent timeout can be overwritten to 5s when other options are specified
- GetReport timeout defaults to 10s
- GetReport timeout can be overwritten to 5s

**Test Configuration**:
* SDK Version - v1.72.3
* NPM Version - 6.4.1

## Checklist:


willward:
@decryan1984 @gannons can you take a first pass please

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -2,12 +2,12 @@
 /*!
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
```
willward: I think eslint did this re-ordering of imports

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -77,81 +79,90 @@ const EDiscovery = WebexPlugin.extend({
...
    * @param {UUID} reportId - Id of the report which contains the content
-   * @param {int} offset - Start position from which to retrieve content
-   * @param {int} size - Number of activities to retrieve
+   * @param {Object} page - object with offset and size parameters indicating the start position from which to retrieve
+   *                        content and the number of activities to retrieve
+   * @param {int} timeoutMs - optional connection timeout in milliseconds, defaults to 10s
    * @returns {Promise<ResponseEntity<[Activity]>>} Http response containing the activities
    */
-  getContent(reportId, {offset, size} = {}) {
+  getContent(reportId, page, timeoutMs = 10000) {
```
willward: I change the object literal to be a named parameter as i saw a test failure, but looking over now i see i accidentally removed the default value for this object `= {}`, so i'm going to fix this and see if it resolves the test failure also

willward: yes, switching back to `{offset, size} = {}` works fine, will go with that

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -77,81 +79,88 @@ const EDiscovery = WebexPlugin.extend({
...
    * Retrieves content associated with a report
    * @param {UUID} reportId - Id of the report which contains the content
-   * @param {int} offset - Start position from which to retrieve content
-   * @param {int} size - Number of activities to retrieve
+   * @param {Object} {offset, size} - Start position from which to retrieve content and number of activities to retrieve
+   * @param {int} timeoutMs - optional connection timeout in milliseconds, defaults to 10s
    * @returns {Promise<ResponseEntity<[Activity]>>} Http response containing the activities
    */
-  getContent(reportId, {offset, size} = {}) {
+  getContent(reportId, {offset, size} = {}, timeoutMs = 10000) {
```
gannons: Given the current issue with getContent delays I assume you'll be specifying an non-default value in the downloader

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -28,9 +28,10 @@ const EDiscovery = WebexPlugin.extend({
   /**
    * Creates a compliance report with a specific set of search parameters
    * @param {Object} reportRequest - A set of criteria for determining the focus of the search
+   * @param {int} timeoutMs - optional connection timeout in milliseconds, defaults to 10s
    * @returns {Promise<ResponseEntity>} Http response containing the new report record
    */
-  createReport(reportRequest) {
+  createReport(reportRequest, timeoutMs = 10000) {
```
decryan1984: Since this is an optional parameter, would it make more sense to wrap it in an options object as is done below for the paged APIs?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -28,19 +31,24 @@ const EDiscovery = WebexPlugin.extend({
...
    * @param {Object} reportRequest - A set of criteria for determining the focus of the search
+   * @param {Object} options - can specify this option :
+   *                           timeoutMs - connection timeout in milliseconds, defaults to 10s
    * @returns {Promise<ResponseEntity>} Http response containing the new report record
    */
-  createReport(reportRequest) {
+  createReport(reportRequest, options = {}) {
     if (!reportRequest) {
       throw Error('Undefined parameter');
     }
```
willward: updated to a named options parameter and using defaults from lodash to set timeoutMs if undefined, i still need to assign a default empty object in the param definition

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -35,6 +35,47 @@ describe('EDiscovery Content API Tests', () => {
 
       return result;
     });
+
```
willward: added new tests

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/report.js
```diff
@@ -80,11 +80,38 @@ describe('EDiscovery Report API Tests', () => {
       return result;
     });
```
willward: added new tests

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
willward: the timeout in the request module is for connection and first byte transferred, if the overall transfer of data takes > 10s then that's ok, it won't be killed by the timeout

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
willward: updated

willward:
> Looking good. A few things I'd like to see:
> 
> * anywhere it makes sense to add the timeout into an options object instead of an additional param, please do that (like functions that already them)
> * since we are sharing a default timeout, could you make it a variable instead of manually doing 10000?  i.e. `const defaultTimeout = 10000;` `timeoutMs = defaultTimeout`
> * would like to see some unit tests around handling the timeout to make sure defaults and custom values are passed

@adamweeks, i think i have addressed the 3 of these items now in my latest commit, see what you think

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -28,19 +31,24 @@ const EDiscovery = WebexPlugin.extend({
   /**
    * Creates a compliance report with a specific set of search parameters
    * @param {Object} reportRequest - A set of criteria for determining the focus of the search
+   * @param {Object} options - can specify this option :
+   *                           timeoutMs - connection timeout in milliseconds, defaults to 10s
```
adamweeks: To get this option to display in jsdoc readers, you can change this line to:
`* @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 10s`

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -77,84 +85,109 @@ const EDiscovery = WebexPlugin.extend({
   /**
    * Retrieves information relating to a specified report
    * @param {UUID} reportId - Id of the report being requested
+   * @param {Object} options - can specify this option :
```
adamweeks: Same thing here with the jsdoc

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -28,19 +31,24 @@ const EDiscovery = WebexPlugin.extend({
   /**
    * Creates a compliance report with a specific set of search parameters
    * @param {Object} reportRequest - A set of criteria for determining the focus of the search
+   * @param {Object} options - can specify this option :
+   *                           timeoutMs - connection timeout in milliseconds, defaults to 10s
    * @returns {Promise<ResponseEntity>} Http response containing the new report record
    */
-  createReport(reportRequest) {
+  createReport(reportRequest, options = {}) {
```
mattnorris: Rather than rely on lodash, can you use `createReport(reportRequest, options = {timeoutMs: defaultTimeoutMs})` and then the spread operator in the body?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
willward: updated the jsdoc now, thanks for the tip

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
willward: Hi Matt, I don't understand how I can use the spread operator in this content ( due to lack of familiarity with it ). If you wouldn't mind pasting in a snippet i can update to using it, thanks

willward: in some of the methods options can have up to 3 properties, getContent is a good example. 

I want to allow for options to be passed in as e.g. {offset:0, size:100} and then have the timeoutMs added to this object if it is undefined in the object supplied to the method.

I also wanted to keep a consistent approach for all the methods.

willward: maybe something like this in the body ? : 
```
const defaultSettings = {timeoutMs: defaultTimeoutMs}
options = {...options, ...defaultSettings}
```

willward: updated now, have a look see what you think

willward:
@adamweeks @mattnorris - I made this change because we had hangs in out application due to the timeout of 0 ( infinite timeout ) set in http-core/request-shim.js : 
```
function setDefaults(params, o) {
    const defs = {
      cors: true,
      // raynos/xhr defaults withCredentials to true if cors is true, so we need
      // to make it explicitly false by default
      withCredentials: false,
      timeout: 0
    };

    defaults(params, pick(o, Object.keys(defs)), defs);
  }
```
I can't think of a good reason to have an infinite timeout as the default in the SDK, would you be open to changing this default also in http-core ?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -35,12 +34,13 @@ const EDiscovery = WebexPlugin.extend({
...
    */
-  createReport(reportRequest, options = {}) {
+  createReport(reportRequest, options) {
     if (!reportRequest) {
       throw Error('Undefined parameter');
     }
 
-    defaults(options, {timeoutMs: defaultTimeoutMs});
+    // use spread operator to set default options
+    options = {...defaultOptions, ...options};
```
mattnorris: 👍

PR #1275, opened 2019-07-30
Spark 80557 deprecate wdm call u2c directly

InteractiveTimmy:
# Pull Request

## Description

The first major step in removing the **wdm** service host map dependency from other plugins. This new plugin handles and updates the service host map via the **u2c** endpoint instead of the **wdm** endpoint.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-80557

## Type of Change

## Test Coverage

- [x] Generated new tests [unit]
- [x] Tested against other plugins
- [x] Tested in live samples
- [x] Tested against node instance

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:


packages/node_modules/@webex/internal-plugin-services/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Install
+
+```bash
+npm install --save @webex/internal-plugin-services
+```
+
+## Usage
+
+```js
```
adamweeks: Could you add a note about the added env var?

packages/node_modules/@webex/internal-plugin-services/src/interceptors/service.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    return options;
+  }
+  /* eslint-enable no-param-reassign */
+
+  /* eslint-disable class-methods-use-this */
+  /**
+   * Verify that all required parameters have been specified.
+   * @param {object} options
```
adamweeks: Could you add the params the options object is expected to have as new jsdoc lines like:
```
* @param {String} options.resource
```

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -0,0 +1,227 @@
...
+    return new Promise((resolve) => {
+      if (this.status.ready) { resolve(); }
+      this.once('ready', () => { resolve(); });
+    });
+  },
+
+  /**
+   * **Internal**
+   * Safely load one or more `ServiceUrl`s into this `Services` instance.
+   * @param  {...ServiceUrl} services
```
adamweeks: I think this should be `@param {Array<ServiceUrl>}`

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+    });
+
+    return this;
+  },
+
+  /**
+   * **Internal**
+   * Safely unload one or more `ServiceUrl`s into this `Services` instance
+   * @param  {...ServiceUrl} services
```
adamweeks: I think this should be `@param {Array<ServiceUrl>}`

packages/node_modules/@webex/internal-plugin-services/test/unit/spec/services.js
```diff
@@ -0,0 +1,93 @@
+/*!
+ * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {assert} from '@webex/test-helper-chai';
+// import {cloneDeep} from 'lodash';
```
adamweeks: You can remove this

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -0,0 +1,227 @@
...
+
+        return Promise.resolve();
+      });
+  },
+
+  /**
+   * Wait until this plugin is ready
+   * @returns {Promise<void>}
+   */
+  waitUntilReady() {
```
adamweeks: Really like this handy promise!

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -0,0 +1,227 @@
...
+    ready: false
+  },
+
+  _serviceUrls: [],
+
+  _preDiscoveryServiceUrls: [
+    new ServiceUrl({
+      name: 'u2c',
+      defaultUrl: process.env.U2C_SERVICE_URL ||
+        'https://u2c.wbx2.com/u2c/api/v1',
```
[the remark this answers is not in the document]
InteractiveTimmy: Not sure about this one, This currently follows the same setup as other `config.js` files for other plugins. I'm also not sure if production URLs are injected anywhere other than the `config.js` files for other plugins. Let me know if we should move this to a `config.js` file.

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -0,0 +1,227 @@
...
+   * @returns {ServiceUrl}
+   */
+  _getUrl(name) {
+    return this._serviceUrls.find((su) => su.name === name);
+  },
+
+  /**
+   * Get a service url from the current services list by name.
+   * @param {string} name
+   * @returns {string|void}
```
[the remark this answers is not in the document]
InteractiveTimmy: I believe `undefined` may work better here, as it reflects the value is `undefined` vs being set to an empty string. I do, however, understand this would make the function return a consistent datatype. Which would be preferred in this case?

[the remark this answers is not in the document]
mattnorris: Let's do `undefined`.

packages/node_modules/@webex/internal-plugin-services/src/services.js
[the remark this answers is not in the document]
mattnorris: It's not specific or a blocker to this plugin, but it occurred to me after reading through and seeing the various Prod & Int URLs in various places.

packages/node_modules/@webex/internal-plugin-services/src/services.js
```diff
@@ -31,7 +31,7 @@ const Services = WebexPlugin.extend({
    * **Internal**
    * Creates an object where the keys are the service
    * names and the values are the service urls.
-   * @param {boolean} [priorityHost]
+   * @param {boolean} [priorityHost] // use highest priority host if set to `true`
```
mattnorris: Replace `//` with `-`

PR #1279, opened 2019-08-01
Error handling for unencrypted activities and spaces

decryan1984:
## Description

Some activities and spaces that pass through the transforms for getContent and getContentSummary may not need to be decrypted, they may not have have content or the content may be unencrypted. They may be encrypted but not have an encryption key url which is required for decryption.

Adding handling for these case so that we do not attempt to decrypt these activities and spaces and can avoid errors occurring because of this.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-80456

## Type of Change

## Test Coverage

All existing tests have been run, and 9 new tests have been added to ensure we do not attempt to decrypt any objects that do not require decryption or that we know cannot be decrypted:

**Test Configuration**:
* SDK Version - v1.72.3
* NPM Version - 6.4.1

## Checklist:


packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -201,8 +220,7 @@ class Transforms {
         }
 
         // Decrypt activity message if present
-        // For 'add' activities the objectDisplayName is a user id which does not need to be decrypted
-        if (activity.verb !== 'add' && activity.objectDisplayName && activity.encryptionKeyUrl) {
+        if (activity.objectDisplayName) {
```
decryan1984: Add/Leave/etc activities will no longer make it this far, and neither will activities without an encryptionKeyUrl

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -339,9 +367,10 @@ class Transforms {
         space.spaceName = decryptedSpaceName;
       })
       .catch((reason) => {
-        ctx.webex.logger.error(`Decrypt space name error for space ${space.spaceId}: ${reason}`);
-        // add error property to space info - this error will be recorded in the downloader
-        space.error = reason;
+        ctx.webex.logger.warn(`Decrypt space name error for space ${space.spaceId}: ${reason}`);
+        // this can fail if we attempt to decrypt an unencrypted spacename, so we report a warning rather than an error
+        // add warn property to space info - this warning will be recorded in the downloader
+        space.warn = reason;
```
decryan1984: If we pass an unencrypted string here it manifests as a SyntaxError: Unexpected token I in JSON at position 0. We could potentially interpret this as a warning and anything else as an error but I'm not sure if that's guaranteed to be the only error we receive.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -186,6 +190,21 @@ class Transforms {
           activity.participantDisplayNames = space.participantDisplayNames.filter((value) => value !== activity.actorDisplayName).toString();
         }
 
+        // only post and share activities have content which needs to be decrypted
+        if (activity.verb !== 'post' && activity.verb !== 'share') {
```
gannons: ['post', 'share'].includes(acitivit.verb)

may be more flexible for the future. I imagine we'll be adding an 'edit' verb (assuming it's not covered by post).

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -186,6 +190,21 @@ class Transforms {
...
+        }
+
+        if (!activity.encryptionKeyUrl) {
+          // The encryptionKeyUrl may not be necessary if the activity content is unencrypted, so we treat it as a warning, rather than an error
+          const reason = `Activity ${activity.activityId} cannot be decrypted due to a missing encryption key url`;
+
+          ctx.webex.logger.warn(reason);
+          activity.warn = reason;
+
+          return Promise.resolve(object);
```
gannons: This could be a warning or an error but you always treat it as a warning. i.e. As it may be plaintext or encrypted.

I think anything we can't decrypt is an error as the cloud is meant to be using encryption for everything.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -319,6 +337,16 @@ class Transforms {
...
     }
 
+    if (!space.encryptionKeyUrl) {
+      // The encryptionKeyUrl may not be necessary if the space name is unencrypted, so we treat it as a warning, rather than an error
+      const reason = `Space ${space.spaceId} cannot be decrypted due to a missing encryption key url`;
+
+      ctx.webex.logger.warn(reason);
+      space.warn = reason;
+
+      return Promise.resolve(object);
```
gannons: All the download errors are already, essentially, treated as warnings

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
gannons: AFAIK decryption never fails. Instead it'll rolls/shifts meaningless encrypted bytes to more meaningless bytes.

But the code does an auth check. This essentially compares a hash to a hash we stored. If they don't match it throws an error.

That means this code is checking if the hashes match and when they don't threats it as a warning. I think it needs to be an error.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
gannons: 👍

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -338,11 +331,8 @@ class Transforms {
     }
 
     if (!space.encryptionKeyUrl) {
-      // The encryptionKeyUrl may not be necessary if the space name is unencrypted, so we treat it as a warning, rather than an error
-      const reason = `Space ${space.spaceId} cannot be decrypted due to a missing encryption key url`;
-
-      ctx.webex.logger.warn(reason);
-      space.warn = reason;
+      // If the encryptionKeyUrl is empty we assume the space name is unencrypted
+      ctx.webex.logger.info(`Space ${space.spaceId} cannot be decrypted due to a missing encryption key url`);
```
gannons: Looking at the else statement below it'd be neater to bring it out as an if conditional like this one.

e.g.

    else {
      const reason = `No user available with which to decrypt space ${space.spaceId}`;
      ctx.webex.logger.error(reason);
      space.error = reason;
    }

could become 

    if (!space.onBehalfOfUser) {.....}

That'd put all your end conditionals "up front". But this is not a requirement for acceptance.

PR #1284, opened 2019-08-06
Add max timeout for renegotiating ECDH keys

mattnorris:
# Pull Request 

## Description

Add a max timeout for renegotiating ECDH keys so we don't end up in a loop getting new keys. 

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-84651

## Type of Change

## Test Coverage

`npm run test -- --package @webex/internal-plugin-encryption --node`

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
+++ b/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
@@ -455,7 +455,11 @@ const KMS = WebexPlugin.extend({
             return Promise.reject(reason);
           }
 
-          if (timeout >= this.config.kmsMaxTimeout) {
+          // Peek ahead to make sure we don't reset the timeout if the next timeout
+          // will exceed the maximum timeout for renegotiating ECDH keys.
+          const nextTimeout = timeout * 2;
+
+          if (timeout >= this.config.kmsMaxTimeout && nextTimeout < this.config.ecdhMaxTimeout) {
             this.logger.info('kms: exceeded maximum KMS request retries; negotiating new ecdh key');
 
             /* istanbul ignore else */
--- a/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
+++ b/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
@@ -18,6 +18,8 @@ const contexts = new WeakMap();
 const kmsDetails = new WeakMap();
 const partialContexts = new WeakMap();
 
+const consoleDebug = require('debug')('kms');
+
 /**
  * @class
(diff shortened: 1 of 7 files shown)


mattnorris:
@decryan I also added a test for invalid tokens.

packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
```diff
@@ -444,8 +446,15 @@ const KMS = WebexPlugin.extend({
...
 
+          consoleDebug(`timeout ${timeout}`);
           timeout *= 2;
 
+          if (timeout >= this.config.ecdhMaxTimeout) {
+            this.logger.info('kms: exceeded maximum KMS request retries');
+
+            return Promise.reject(reason);
+          }
+
```
decryan1984: Will this condition ever be met? If the timeout keeps getting doubled it will go to 6, 12, 24, 48, which is still less than 96, so it will skip this, but it'll meet the condition below as it's greater than 32, then get set back to 0 and start all over again?

packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
```diff
@@ -444,9 +446,20 @@ const KMS = WebexPlugin.extend({
...
           timeout *= 2;
 
-          if (timeout >= this.config.kmsMaxTimeout) {
+          if (timeout >= this.config.ecdhMaxTimeout) {
+            this.logger.info('kms: exceeded maximum KMS request retries');
+
+            return Promise.reject(reason);
+          }
+
+          // Peek ahead to make sure we don't reset the timeout if the next timeout
```
mattnorris: @decryan1984 good catch! I think this will close that loophole.

packages/node_modules/@webex/internal-plugin-encryption/test/integration/spec/kms.js
```diff
@@ -10,6 +10,8 @@ import WebexCore from '@webex/webex-core';
 import testUsers from '@webex/test-helper-test-users';
 import uuid from 'uuid';
 
+const debug = require('debug')('kms');
```
adamweeks: Do we have some sort of standard or documentation on these different debug levels?

mattnorris: There isn't really. This particular way is used sparingly within the SDK. I find it useful because the DEBUG logs are colored differently so they contrast against the verbose logs. 

Once the log epic is addressed, these may not be needed.

PR #1285, opened 2019-08-07
add default connection timeout to http-core

willward:
# Pull Request Template

## Description

adding a default http timeout of 60 seconds on requests in the js-sdk. The timeout is for connection to the server and first byte of response received.

Fixes # (issue)

SPARK-83465 - ETIMEDOUT and then next download hung

## Type of Change

## Test Coverage

added 2 new tests to `request.js` in `http-core/test/integration/spec` : 
- "ensure default connection timeout is set"
- "ensure default connection timeout can be overwritten"

ran new and existing tests using command `npm test -- --packages @webex/http-core --node`

**Test Configuration**:
* SDK Version - v1.74.5
* NPM Version - 6.4.1

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/http-core/src/index.js
+++ b/packages/node_modules/@webex/http-core/src/index.js
@@ -3,6 +3,7 @@
  */
 
 import {assign, curry, defaults as lodashDefaults, isString} from 'lodash';
+
 import HttpStatusInterceptor from './interceptors/http-status';
 import _request from './request';
 
@@ -38,6 +39,10 @@ const protorequest = curry(function protorequest(defaultOptions, options) {
 
   if (!options.json) {
     Reflect.deleteProperty(options, 'json');
+  }
+
+  if (!options.timeout) {
+    options.timeout = 60000; // default 60 second timeout
   }
 
   options.logger = options.logger || console;
--- a/packages/node_modules/@webex/http-core/test/integration/spec/request.js
+++ b/packages/node_modules/@webex/http-core/test/integration/spec/request.js
@@ -46,6 +46,16 @@ describe('http-core', function () {
           assert.deepEqual(res.body, {isObject: true});
         }));
 
+      it('ensure default connection timeout is set', () => request({uri: makeLocalUrl('/')})
+        .then((res) => {
+          assert.equal(res.options.timeout, 60000);
+        }));
+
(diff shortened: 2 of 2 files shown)


packages/node_modules/@webex/http-core/src/index.js
```diff
@@ -40,6 +41,10 @@ const protorequest = curry(function protorequest(defaultOptions, options) {
     Reflect.deleteProperty(options, 'json');
   }
 
+  if (!options.timeout) {
+    options.timeout = 60000; // default 60 second timeout
```
mattnorris: Let's move this into a `config.js` in the same folder. [Encryption](https://github.com/mattnorris/spark-js-sdk/blob/3ba8c4c442b757808f9dfd4cc4b991013cb260cb/packages/node_modules/%40webex/internal-plugin-encryption/src/config.js#L17) has some good examples.

willward:
Hi Matt,
             I might need to be extra careful on this default, after adding the ediscovery timeouts we hit an issue calling one of the APIs with latency greater than the timeout, I thought this would be ok as the timeout is for socket connection and first byte transferred. But in this case the API was taking up to 30 seconds processing and was not returning any response until all the processing was done and so it triggered the timeout. This could trip us up for other APIs if we introduce such a widespread default. Perhaps it still ok to proceed but double it up to 2 minutes.

Can you think of any approach where the consumer of the SDK could choose the default timeout at init time or something like this, so that we could enforce a default timeout in our app but not affect other consumers ?

mattnorris:
@willward wouldn't you just leave the code as-is and pass `options.timeout`, i.e., make no changes to the code? That would give you control and not affect other consumers. 

I also just added [a limit to prevent an infinite loop](https://github.com/webex/webex-js-sdk/pull/1284) while renegotiating keys. Does this solve your original problem?

willward:
> @willward wouldn't you just leave the code as-is and pass `options.timeout`, i.e., make no changes to the code? That would give you control and not affect other consumers.
> 
> I also just added [a limit to prevent an infinite loop](https://github.com/webex/webex-js-sdk/pull/1284) while renegotiating keys. Does this solve your original problem?

@mattnorris yes i think that will help alright, there is also one other api we use in the encryption plugin ( _fetchDownloadUrl ) which does the http request of POST files/download/endpoints which does not have a timeout set which i was trying to cover with this. Probably better if it is done specific for that API as opposed to this wide ranging default ( which i now think is much too risky ). I'll close this PR thanks for looking into it.

PR #1291, opened 2019-08-09
Improving efficiency of content summary caching

decryan1984:
# Pull Request Template

## Description

When the downloader app currently calls getContent to retrieve the activities for a report, it implicitly (within the sdk) kicks off a call to getContentSummary before calling getContent. This is to ensure that the content summary is cached before the decryption of activities has begun. Since decrypting activities requires a space name, participants and an onBehalfOfUser, it is vital that this cache is fully populated, otherwise the client will have to make a call to getContentSummaryBySpaceId for every single activity it is decrypting. We have seen instances where this cache was not fully present, and this can result in the downloader being rate-limited.

We are modifying this behaviour so that the content summary cache is only used on a per space basis, and if there are gaps in the cache, they will now be filled by calls to getContentSummaryBySpaceId:

1. Replace the ContentSummary cache with a ContentSummaryBySpaceId cache. We do not need to read from it during calls to getContentSummary, just write to it (by space id). We should read and write from/to it when making calls to ContentSummaryBySpaceId. This way if an item is not present in the cache for some reason, we should populate it on the next call.
2. Use the one flight annotation to prevent multiple network calls to ContentSummaryBySpaceId for the same info while the cache is waiting to be populated.
3. Remove the implicit call to getContentSummary from getContent in the sdk, and add one to the downloader before getContent is called.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-86991

## Type of Change

## Test Coverage

All existing tests have been run and are passing.

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

Changes in the downloader have not yet been published, it is better to make them when there is an SDK version available with these changes


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -8,7 +8,7 @@ import Map from 'babel-runtime/core-js/map';
 
 import {InvalidEmailAddressError} from './ediscovery-error';
 
-const defaultOptions = {timeoutMs: 10000}; // 10s default timeout
+const defaultOptions = {timeoutMs: 30000}; // 30s default timeout
```
decryan1984: Default timeout increased as >10s latency was seen for content and content summary APIs

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -193,51 +193,38 @@ const EDiscovery = WebexPlugin.extend({
...
+      service: 'ediscovery',
+      resource: `reports/${reportId}/contents`,
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    });
   },
 
   @waitForValue('@')
-  @oneFlight({keyFactory: ({reportId}) => `${reportId}`})
+  @oneFlight({keyFactory: (reportId) => `${reportId}`})
```
decryan1984: This error may explain some of the unusual behaviour we've seen around getContentSummary

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -193,51 +193,38 @@ const EDiscovery = WebexPlugin.extend({
...
+   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 90s
    * @returns {Promise<ResponseEntity<ContentSummary>>} Http response containing the content summary
    */
   getContentSummary(reportId, options) {
     if (!reportId) {
       throw Error('Undefined parameter');
     }
 
+    const contentSummaryDefaultOptions = {timeoutMs: 90000};
+
```
decryan1984: Large space participant lists have caused latency issues for this API, increasing timeout here while we fix those server side

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -193,51 +193,38 @@ const EDiscovery = WebexPlugin.extend({
...
 
     let offset = 0;
     const size = 100;
     let contentSummary = [];
 
-    // If the content summary for this report has already been cached then it can be retrieved from there instead of making a network call to ediscovery
-    if (this.contentSummaryCache.has(reportId) && this.contentSummaryCache.get(reportId).size > 0) {
-      return Array.from(this.contentSummaryCache.get(reportId).values());
-    }
-
```
decryan1984: We no longer user the entire cache for the contentSummary, if a partial error occurred we would be missing spaces and not necessarily know about it

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
willward: where the error was the extra set of braces in the annotation

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
sregger: This was causing the ediscovery service to be hammered with a large number of getContentSummaryBySpaceId requests which caused that API to slow down.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -286,17 +285,37 @@ const EDiscovery = WebexPlugin.extend({
...
+          catch (error) {
+            this.logger.error(`Error adding ${res.body.spaceId} to contentSummaryBySpaceIdCache: ${error}`);
+          }
+        }
+        else {
+          this.logger.error('Error adding undefined space to contentSummaryBySpaceIdCache');
+        }
+
+        return res.body;
+      });
```
sregger: Good idea to handle the `catch` scenario? Or is it more appropriate to let the upper level handle it?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -248,11 +235,22 @@ const EDiscovery = WebexPlugin.extend({
       })
         .then((res) => {
```
sregger: Should this code handle the failure scenario in a `catch`?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -286,17 +285,37 @@ const EDiscovery = WebexPlugin.extend({
     options = {...defaultOptions, ...options};
```
willward: maybe we should the comment here to use the word space instead of report since the cache is now per space

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: I think we should let the upper layer handle it, that seems to be the general pattern adopted by the other plugins. If there are special cases we need to handle (e.g. the invalid emails in createReport), we can add catch handlers for them, but otherwise it's best to keep this class as straightforward as possible.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: I don't believe so, see full answer below

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: good catch, done

willward:
looks good to me @decryan1984

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -193,51 +193,38 @@ const EDiscovery = WebexPlugin.extend({
...
 
   @waitForValue('@')
-  @oneFlight({keyFactory: ({reportId}) => `${reportId}`})
+  @oneFlight({keyFactory: (reportId) => `${reportId}`})
   /**
    * Retrieves a list of the conversations relevant to a specified report
    * @param {UUID} reportId - Id of the report being requested
    * @param {Object} options - optional parameters for this method
-   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 10s
+   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 90s
```
adamweeks:
```suggestion
   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 30s
```

adamweeks: Oh sorry, I see this isn't using the `defaultOptions`

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -248,11 +235,22 @@ const EDiscovery = WebexPlugin.extend({
       })
         .then((res) => {
           // The results are written to the contentSummaryCache as information for a space is likely to be accessed very frequently for decryption
```
adamweeks: This comment needs to be updated

adamweeks:
@decryan1984 I see that you have marked that there are breaking changes. Is there some documentation that needs to be updated for these?

decryan1984:
> @decryan1984 I see that you have marked that there are breaking changes. Is there some documentation that needs to be updated for these?

They're only breaking in the sense that the downloader app now needs to cache the content summary explicitly Adam, we will make that change when we update the app to the new sdk version with this fix.

adamweeks:
I have approved this PR to go through via `ci-skip` due to the testing changes that will be added to `getContentSummary`.

PR #1295, opened 2019-08-13
Removing recursive looping from getContentSummary API

decryan1984:
# Pull Request Template

## Description

While the downloader app handles the paging of the getContent API (it makes requests repeatedly until all of the content has been retrieved and written to the report), the getContentSummary API is used differently in that the paging is managed in the plugin, it loops internally, retrieving all of the content summary info and returning it as one collection, saving the downloader from having to do it.

We are simplifying the code around this API to simply return a page of data, removing the recursive looping. We have also aligned the return types for getContentSummary, getContentSummaryBySpaceId and the contentSummaryBySpaceId cache so that they all return a response type object (res.body.x).

Also, we are now using the @oneflight annotation for getContent and adding options parameter to its key to ensure we don't return an incorrect set of results.

Adding JSDoc for transforms and several new tests, in particular for getContentSummary, which was not covered previously due to the recursion.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-76030

## Type of Change

## Test Coverage

~14 new tests added, all pass

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

Ediscovery downloader app will be updated when this sdk change is available: https://sqbu-github.cisco.com/WebExSquared/ediscovery_downloader/pull/194


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -287,7 +257,7 @@ const EDiscovery = WebexPlugin.extend({
 
     // If the content summary for this space has already been cached then it can be retrieved from there instead of making a network call to ediscovery
     if (this.contentSummaryBySpaceIdCache.has(reportId) && this.contentSummaryBySpaceIdCache.get(reportId).has(spaceId)) {
-      return this.contentSummaryBySpaceIdCache.get(reportId).get(spaceId);
+      return {body: this.contentSummaryBySpaceIdCache.get(reportId).get(spaceId), statusCode: 200};
```
decryan1984: A response from the cache should mirror a response from the server

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -176,6 +176,7 @@ const EDiscovery = WebexPlugin.extend({
   },
 
   @waitForValue('@')
+  @oneFlight({keyFactory: (reportId, options) => String(reportId + (options && options.offset && options.size ? options.offset + options.size : ''))})
```
decryan1984: Oneflight key has to take the offset into account

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -21,8 +21,6 @@ describe('EDiscovery Content API Tests', () => {
 
   describe('GetContent Tests', () => {
     it('GetContent succeeds', async () => {
-      // getContent calls getContentSummary internally, so we need to stub it here
-      webex.internal.ediscovery.getContentSummary = sinon.stub().returns(Promise.resolve());
```
decryan1984: No longer necessary thankfully

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -80,23 +74,126 @@ describe('EDiscovery Content API Tests', () => {
   });
 
   describe('GetContentSummary Tests', () => {
-    // The fact that the getContentSummary APIs return the response body object (which is a ContentSummary object)
-    // rather than the entire response means we must take a different approach to verifying to verifying these tests.
-    // If possible, this should be altered as part of SPARK-76030 to align with the other APIs
```
decryan1984: Catching up on test coverage for getContentSummary

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -141,7 +257,7 @@ describe('EDiscovery Content API Tests', () => {
       };
       const result = webex.internal.ediscovery.getContentSummaryBySpaceId(uuid, uuid)
         .then((res) => {
-          expect(res).to.equal(mockCachedSpace);
+          expect(res.body).to.equal(mockCachedSpace);
```
decryan1984: getContentSummaryBySpaceId now returns a response type result even if it hits the cache

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -167,13 +283,39 @@ describe('EDiscovery Content API Tests', () => {
...
+      const result = webex.internal.ediscovery.getContentSummaryBySpaceId(uuid, uuid)
+        .then((res) => {
+          expect(res).to.equal(mockResponse);
+          assert.callCount(webex.internal.ediscovery.contentSummaryBySpaceIdCache.has, 2);
+          assert.callCount(spaceIdMap.has, 1);
+          assert.notCalled(spaceIdMap.set);
+        });
+
+      return result;
+    });
```
decryan1984: This should cover one of the conditionals that Adam noticed was missing

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -299,24 +269,65 @@ const EDiscovery = WebexPlugin.extend({
...
+  getClientConfig(options) {
+    // use spread operator to set default options
+    options = {...defaultOptions, ...options};
+
+    return this.request({
+      method: 'GET',
+      service: 'ediscovery',
+      resource: 'reports/clientconfig',
+      timeout: options.timeoutMs
+    });
```
decryan1984: This API will let us dynamically configure aspects of the client behaviour for better performance, for example, we can tweak the page size of the getContentSummary API based on service latency metrics etc.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -8,7 +8,7 @@ import Map from 'babel-runtime/core-js/map';
 
 import {InvalidEmailAddressError} from './ediscovery-error';
 
-const defaultOptions = {timeoutMs: 30000}; // 30s default timeout
+const defaultOptions = {offset: 0, size: 100, timeoutMs: 30000}; // 30s default timeout
```
andrewholsted: These should generally be specified in the plugin config so they could be overwritten by users when they instantiate the sdk. I know this was already here. Just mentioning what the current pattern is.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -203,78 +204,47 @@ const EDiscovery = WebexPlugin.extend({
...
+      service: 'ediscovery',
+      resource: `reports/${reportId}/contents/summary`,
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    })
+      .then((res) => {
+        const cachePromise = this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body);
 
-    return getNextContentSummaryPage(this);
+        return Promise.all([cachePromise])
```
andrewholsted: Does ._writeSpacesToCon... return an array of promises?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
andrewholsted: Does the cache ever expire, or need to?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
andrewholsted: It doesn't appear to, so wondering about the promise.all instead of chaining on the then

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
andrewholsted: For for own curiosity, why is it no longer needed? Did it make a network request before and now it doesn't?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Good point Andrew, I'll change to a then

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Right now it lasts for as long as the downloader client is open, it shouldn't be a problem for now, but we may have to come back and add a cleanup mechanism at some point.

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
decryan1984: GetContent had a baked in call to getContentSummary in order to populate the cache, this was over complicating the flow so we've removed it from the plugin and now control it from the client

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -203,78 +204,47 @@ const EDiscovery = WebexPlugin.extend({
   },
 
   @waitForValue('@')
-  @oneFlight({keyFactory: (reportId) => `${reportId}`})
+  @oneFlight({keyFactory: (reportId, options) => String(reportId + (options && options.offset && options.size ? options.offset + options.size : ''))})
```
mattnorris: Because this is used multiple times and use of the offset is important to how this works, it would be good to break this function out and document it with JSDoc.

packages/node_modules/@webex/internal-plugin-ediscovery/src/config.js
```diff
@@ -4,5 +4,6 @@
 
 export default {
   ediscovery: {
+    defaultOptions: {offset: 0, size: 100, timeoutMs: 30000}
```
decryan1984: Moved default config out of ediscovery.js

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -4,11 +4,19 @@
...
+/**
+ * Creates a unique oneflight key for a request from the reportId and options params
+ * It is important that the offset params are included if present to ensure paged requests get back the correct data
+ * @param {Object} reportId - A set of criteria for determining the focus of the search
+ * @param {Object} options - optional parameters for this method
+ * @returns {String} oneFlight key which is a composite of the request parameters
+ */
+function createOneFlightKey(reportId, options) {
+  return String(reportId + (options && options.offset && options.size ? options.offset + options.size : ''));
+}
```
decryan1984: New function for oneFlight key with options param

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -298,25 +272,65 @@ const EDiscovery = WebexPlugin.extend({
...
+   * @returns {Promise} - Promise resolution indicating operation is complete
+   */
+  _writeSpacesToContentSummaryBySpaceIdCache(reportId, spaces) {
+    if (!reportId || !spaces || !spaces.length) {
+      return Promise.resolve();
+    }
+
+    if (!this.contentSummaryBySpaceIdCache.has(reportId)) {
+      // wipe the cache - we will only need to store one reports data at any one time
+      this.contentSummaryBySpaceIdCache.clear();
```
decryan1984: I've added a basic expiry mechanism, since each downloader client can only retrieve content for one report at a time, we can clear the cache whenever a new report comes in

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -80,23 +76,129 @@ describe('EDiscovery Content API Tests', () => {
...
+
+    it('GetContentSummary initialises contentSummaryBySpaceIdCache if uninitialised', async () => {
+      const mockResponse = {body: [{spaceId: uuid}]};
+
+      webex.internal.ediscovery.request = sinon.stub().resolves(mockResponse);
+
+      webex.internal.ediscovery.contentSummaryBySpaceIdCache = {
+        set: sinon.stub(),
+        has: sinon.stub().returns(false),
+        clear: sinon.stub()
```
decryan1984: clears now needed on cache stub

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -203,78 +212,43 @@ const EDiscovery = WebexPlugin.extend({
...
-    return getNextContentSummaryPage(this);
+    return this.request({
+      method: 'GET',
+      service: 'ediscovery',
+      resource: `reports/${reportId}/contents/summary`,
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    })
+      .then((res) => this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body)
+        .then(() => Promise.resolve(res)));
```
decryan1984: Switched from Promise.all to chained thens

mattnorris: I don't think you need the additional `.then`. If `_writeSpacesToContentSummaryBySpaceIdCache` returns a resolved promise, you can just return it. 

```
.then((res) => this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body));
```

decryan1984: But since _writeSpacesToContentSummaryBySpaceIdCache doesn't return res, the response won't be passed back to the caller right?

mattnorris: Ah, I see. I think the nested `then`s threw me off a bit. I would rewrite this to use `await` to make it clearer. 

```
.then((res) => {
  await this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body); 
  return Promise.resolve(res); 
});
```
You're not returning the result of that write to cache, but rather waiting for it to write before simply returning the response.

decryan1984: Can await be used from within WebexPlugin?

mattnorris: Yes, here is an [example](https://github.com/webex/webex-js-sdk/pull/1304/files#diff-4839efc2cf969a22d9e796384dcf75bdR1141).

decryan1984: Thanks Matt, adding async to the enclosing function in the then did the trick

PR #1324, opened 2019-08-30
feat(plugin-meetings): add #register method

adamweeks:
# Pull Request Template

## Description

Adds `meetings.register` method to do the device registration, websocket connect, and locus event listening. 

BREAKING CHANGE: No longer automatically listens to locus events upon initialization.

Fixes # SPARK-86902

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Coverage

- [X] Plugin-meetings
- [X] Samples

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
@@ -50,6 +50,14 @@ import MeetingsUtil from './util';
    */
 
 /**
+ * Meetings Registered Event
+ * Emitted when the meetings instance has been registered and listening
+ * @event meetings:registered
+ * @instance
+ * @memberof Meetings
+ */
+
+/**
     * Meeting Removed Event
     * Emitted when a meeting was removed from the cache of meetings
     * @event meeting:removed
@@ -121,6 +129,16 @@ export default class Meetings extends WebexPlugin {
        * @memberof Meetings
        */
       this.reachability = null;
+
+      /**
+       * If the meetings plugin has been registered and listening via {@link Meetings#register}
+       * @instance
+       * @type {Boolean}
+       * @public
+       * @memberof Meetings
+       */
+      this.registered = false;
+
       this.onReady();
       Metrics.initialSetup(this.meetingCollection, this.webex.version);
     }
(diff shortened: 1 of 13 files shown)


packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -263,19 +281,51 @@ export default class Meetings extends WebexPlugin {
         StaticConfig.set(this.config);
         LoggerConfig.set(this.config.logging);
         LoggerProxy.set(this.webex.logger);
-        if (this.webex.canAuthorize) {
```
arun3528: Do we need to add this check back  ?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -263,19 +281,51 @@ export default class Meetings extends WebexPlugin {
...
+      if (!this.webex.canAuthorize) {
+        LoggerProxy.logger.error('meetings->register#ERROR, Unable to register, SDK cannot authorize');
+
+        return Promise.reject(new Error('SDK cannot authorize'));
+      }
+
+      return this.webex.internal.device.register()
+        .then(() => this.webex.internal.mercury.connect())
+        .then(() => {
           this.listenForEvents();
```
arun3528: Does the register happens when plugin is loaded ? if yes should `listenForEvents` be here

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
adamweeks: No, since we aren't doing anything that requires an authorized user on plugin init now.

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
adamweeks: Register does not happen automatically with the plugin. The issue that this PR is attempting to resolve is due to the fact that the meetings plugin automatically will listen to events from locus, even if the SDK user isn't concerned with them. Because meetings plugin is part of the main "webex" package, we need to make things like listening for events less automatic and more direct.

packages/node_modules/samples/browser-plugin-meetings/app.js
```diff
@@ -46,17 +46,14 @@ function connect() {
     });
   }
 
-  if (!webex.internal.device.registered) {
-    webex.internal.device
-      .register()
+  // Register our device with Webex cloud
+  if (!webex.meetings.registered) {
```
adamweeks: @chrisadubois here is the change you requested

robstax: would it make sense to move this conditional `webex.meetings.registered` inside of `webex.meetings.register()`?
so like inside `register()` if `webex.meetings.registered` just reject or something? what happens if the user accidentally tries to register twice?

adamweeks: Good call. I will follow the pattern of the other plugins and log that it is already registered and immediately resolve the promise.

PR #1327, opened 2019-09-05
fix(webex-core): add default options interceptor

willward:
# Pull Request Template

## Description

This change adds a new Interceptor for setting default options for http-core request that will apply for every request made by the SDK. The consumer of the SDK can set the options when calling WebexSdk.init like this : 
```
webex = WebexSdk.init({
      credentials: {
        supertoken: superToken
      },
      config: {
        credentials: {
          client_id,
          client_secret
        },
        defaultRequestOptions: {
          timeout: 15000
        }
      }
    });
```
For the ediscovery downloader application we will use this feature to set : 
- a custom http agent for the request which performs ocsp revocation checking
- potentially a default request timeout

Fixes # (issue)
will be used to fix  https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-91060

## Type of Change

## Test Coverage

- add 3 new tests in `@webex/webex-core/test/unit/spec/interceptors/default-options.js`

**Test Configuration**:
* SDK Version - v1.79.0
* NPM Version - v6.4.1

## Checklist:


packages/node_modules/@webex/webex-core/src/interceptors/default-options.js
```diff
@@ -0,0 +1,56 @@
...
+    const {defaultRequestOptions: defaultOptions} = this.webex.config;
+
+    if (!defaultOptions) {
+      return options;
+    }
+
+    Object.keys(defaultOptions).forEach((key) => {
+      // don't override any existing option properties
+      if (!Object.keys(options).includes(key)) {
+        options[key] = defaultOptions[key];
```
InteractiveTimmy: Is there additional intent to append default header key:values in default options? If so, this may not work here, as it does not deep merge/check.

packages/node_modules/@webex/webex-core/src/interceptors/default-options.js
```diff
@@ -0,0 +1,56 @@
...
+ *   webex = WebexSdk.init({
+ *     credentials: {
+ *       supertoken: superToken
+ *     },
+ *     config: {
+ *       credentials: {
+ *         client_id,
+ *         client_secret
+ *       },
+ *       defaultRequestOptions: {
```
InteractiveTimmy: I like how this can be implemented at the instance level. Might be able to shorten this to just `requestOptions`. Need additional input from the team.

packages/node_modules/@webex/webex-core/src/interceptors/default-options.js
willward: the use cases i had was only for top level properties so didn't want to over complicate it, but if you think a deep merge would be better i can look into it.

packages/node_modules/@webex/webex-core/src/interceptors/default-options.js
adamweeks: `defaultRequestOptions` works for me and is descriptive enough to not be confusing

willward:
> All generated code looks good to me. My concern with an interceptor like this is that its scope may be too flexible. Additionally, other interceptors sometimes have conditionals that prevent them from modifying variables that already exist, so if the defaultOptions interceptor sets specific key:values, the requests will bypass the conditionals.

perhaps a whitelist of allowed properties that can be set by this interceptor would help to reduce the scope ?

PR #1340, opened 2019-09-18
chore(eslint): disable underscore dangle rule

adamweeks:
The plugins use the underscored function names to
indicate private functions. This is a highly debated
pattern, but is consistent across the sdk.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -37,7 +37,7 @@ rules:
   no-multi-assign: warn
   no-param-reassign: warn
   no-shadow: warn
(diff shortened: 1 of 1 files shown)


taymoork2:
I think it's fine to keep it as a warn, since it doesn't impede on the current developer experience unless you're working on older code that was put in before the warn was added.
The warn theoretically should stop future developers from using it.

Then again I'm ok with waiting for this to go in after the pipeline goes in, but would also want to take it as an opportunity to take a look at our current eslint setup as a whole

PR #1343, opened 2019-09-19
Create service validation methods

InteractiveTimmy:
# Pull Request Template

## Description

Create validation methods to identify that a url is a part
of the current service catalog. Additionally, append methods
that can be utilized to retrieve other service data, such as
`clusterId`.

Fixes # (issue)
SPARK-94197
SPARK-86858

## Type of Change

## Test Coverage

- [ ] Added Node NPM Tests and Validated they succeeded
- [ ] Ran a live sample to confirm everything worked

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
+++ b/packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
@@ -26,7 +26,7 @@ export default class RedirectInterceptor extends Interceptor {
    * @returns {Object}
    */
   onRequest(options) {
-    if (typeof options.uri === 'string') {
+    if (options && options.uri && typeof options.uri === 'string') {
       if (options.uri.includes('https://idbroker') ||
         options.uri.includes(this.webex.config.credentials.samlUrl) ||
         options.uri.includes(this.webex.config.credentials.tokenUrl) ||
--- a/packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
+++ b/packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
@@ -26,12 +26,13 @@ export default class RedirectInterceptor extends Interceptor {
    * @returns {Object}
    */
   onRequest(options) {
-    if (options.uri && options.uri.includes('https://idbroker')) {
-      return options;
-    }
-
-    if (options.uri.includes(this.webex.config.credentials.samlUrl) || options.uri.includes(this.webex.config.credentials.tokenUrl) || options.uri.includes(this.webex.config.credentials.authorizeUrl)) {
-      return options;
+    if (typeof options.uri === 'string') {
+      if (options.uri.includes('https://idbroker') ||
+        options.uri.includes(this.webex.config.credentials.samlUrl) ||
+        options.uri.includes(this.webex.config.credentials.tokenUrl) ||
+        options.uri.includes(this.webex.config.credentials.authorizeUrl)) {
+        return options;
+      }
     }
 
     // If cisco-no-http-redirect is already set, don't overwrite it
--- a/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
@@ -140,10 +140,6 @@ const ServiceCatalog = AmpState.extend({
           break;
         }
       }
-
-      if (clusterId) {
-        break;
-      }
     }
 
     return clusterId;
--- a/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
@@ -163,7 +163,7 @@ const ServiceCatalog = AmpState.extend({
    * @returns {string} service.name
    * @returns {string} service.url
    */
-  findServiceFromClusterId({clusterId, priorityHost = true, serviceGroup}) {
+  findServiceFromClusterId({clusterId, priorityHost = true, serviceGroup} = {}) {
     const serviceUrls = (typeof serviceGroup === 'string') ?
       this.serviceGroups[serviceGroup] || [] :
       [
--- a/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
@@ -1,3 +1,5 @@
+import Url from 'url';
+
 import AmpState from 'ampersand-state';
 
 import ServiceUrl from './service-url';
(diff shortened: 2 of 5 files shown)


packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,110 @@ const ServiceCatalog = AmpState.extend({
...
+    let clusterId;
+    let serviceUrlObj;
+
+    Object.keys(this.serviceGroups).forEach((key) => {
+      this.serviceGroups[key].forEach((service) => {
+        serviceUrlObj = Url.parse(service.defaultUrl);
+
+        if (serviceUrlObj.host === incomingUrlObj.host &&
+          service.hosts.length > 0) {
+          clusterId = service.hosts[0].id;
```
adamweeks: Should we set this up so that it breaks out of the loops once it finds an ID?

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,110 @@ const ServiceCatalog = AmpState.extend({
...
+
+    return undefined;
+  },
+
+  /**
+   * Find a service based on the provided url.
+   * @param {string} url - Must be parsable by `Url`
+   * @returns {serviceUrl} - ServiceUrl assocated with provided url
+   */
+  findServiceFromUrl(url) {
```
adamweeks: Since this returns a service url, would a better name be `findServiceUrlFromUrl`? Maybe some details in the jsdoc as the difference in the url from the param vs the url that is returned.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,110 @@ const ServiceCatalog = AmpState.extend({
...
+   * Search over all service groups and return a service value from a provided
+   * clusterId. Currently, this method will return either a service name, or a
+   * service url depending on the `value` parameter. If the `value` parameter
+   * is set to `name`, it will return a service name to be utilized within the
+   * Services plugin methods.
+   * @param {object} params
+   * @param {string} params.clusterId - clusterId of found service
+   * @param {boolean} [params.priorityHost] - returns priority host url if true
+   * @param {string} [params.serviceGroup] - specify service group
+   * @param {string} [params.value] - select value of service to return
```
adamweeks: Does it make things easier if we returned an object with both the name and url instead of having to select one or the other via param?

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
InteractiveTimmy: There isn't a built-in break method for `Array.forEach()`. This loop would need to be converted to a `for(iterator, conditional, increment)` loop. Let me know if you think this is necessary.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
InteractiveTimmy: Addressed.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
adamweeks: Yea, let's change it to the `for` iterator so we can break out of the loops.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
InteractiveTimmy: addressed.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
InteractiveTimmy: addressed

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -314,6 +314,44 @@ const Services = WebexPlugin.extend({
...
+  },
+
+  /**
+   * Get a service value from a provided clusterId. This method will
+   * return an object containing both the name and url of a found service.
+   * @param {object} params
+   * @param {string} params.clusterId - clusterId of found service
+   * @param {boolean} [params.priorityHost] - returns priority host url if true
+   * @param {string} [params.serviceGroup] - specify service group
+   * @returns {object} - {name: string, url: string}
```
adamweeks:
```suggestion
   * @returns {object} service
   * @returns {string} service.name
   * @returns {string} service.url
```

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,135 @@ const ServiceCatalog = AmpState.extend({
...
+   * @param {string} url - Must be parsable by `Url`
+   * @returns {string} - ClusterId of a given url
+   */
+  findClusterId(url) {
+    const incomingUrlObj = Url.parse(url);
+    let clusterId;
+    let serviceUrlObj;
+
+    for (const key of Object.keys(this.serviceGroups)) {
+      for (const service of this.serviceGroups[key]) {
```
mwegman: Get rid of the surrounding for loop and change this to `for (const service of Object.values(this.serviceGroups)) {`

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,135 @@ const ServiceCatalog = AmpState.extend({
...
+        serviceUrlObj = Url.parse(service.defaultUrl);
+
+        if (serviceUrlObj.host === incomingUrlObj.host &&
+          service.hosts.length > 0) {
+          clusterId = service.hosts[0].id;
```
mwegman: Do a break here. You have a `clusterId`

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,135 @@ const ServiceCatalog = AmpState.extend({
...
+        }
+
+        for (const host of service.hosts) {
+          if (incomingUrlObj.host === host.host && host.id) {
+            clusterId = host.id;
```
mwegman: Move the break up here.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,135 @@ const ServiceCatalog = AmpState.extend({
...
+   * service url depending on the `value` parameter. If the `value` parameter
+   * is set to `name`, it will return a service name to be utilized within the
+   * Services plugin methods.
+   * @param {object} params
+   * @param {string} params.clusterId - clusterId of found service
+   * @param {boolean} [params.priorityHost] - returns priority host url if true
+   * @param {string} [params.serviceGroup] - specify service group
+   * @returns {object} - {name: string, url: string}
+   */
+  findServiceFromClusterId({clusterId, priorityHost, serviceGroup}) {
```
mwegman: I would think we would want to default to true for the `priorityHost`. We could change it to
`findServiceFromClusterId({clusterId, priorityHost = true, serviceGroup} = {}) {`

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
InteractiveTimmy: Moved the breakpoint.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,116 @@ const ServiceCatalog = AmpState.extend({
...
+        }
+
+        for (const host of service.hosts) {
+          if (incomingUrlObj.host === host.host && host.id) {
+            clusterId = host.id;
+            break;
+          }
+        }
+
+        if (clusterId) {
```
adamweeks: This test can be removed since the break was moved up.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -106,6 +108,116 @@ const ServiceCatalog = AmpState.extend({
...
+          break;
+        }
+      }
+
+      if (clusterId) {
```
adamweeks: This test can be removed since the break was moved up.

adamweeks:
Could you rebase on master to get rid of the merge commits? Also, looks like the unit tests are currently failing.

packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -180,9 +159,11 @@ const ServiceCatalog = AmpState.extend({
    * @param {string} params.clusterId - clusterId of found service
    * @param {boolean} [params.priorityHost] - returns priority host url if true
    * @param {string} [params.serviceGroup] - specify service group
-   * @returns {object} - {name: string, url: string}
+   * @returns {object} service
+   * @returns {string} service.name
+   * @returns {string} service.url
    */
-  findServiceFromClusterId({clusterId, priorityHost, serviceGroup}) {
+  findServiceFromClusterId({clusterId, priorityHost = true, serviceGroup}) {
```
mwegman: It's nice to have `... = {}` at the end of the parameters just in case someone calls `findServiceFromCluserId()` with nothing.

InteractiveTimmy: Implimented.

packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
```diff
@@ -26,12 +26,13 @@ export default class RedirectInterceptor extends Interceptor {
    * @returns {Object}
    */
   onRequest(options) {
-    if (options.uri && options.uri.includes('https://idbroker')) {
-      return options;
-    }
-
-    if (options.uri.includes(this.webex.config.credentials.samlUrl) || options.uri.includes(this.webex.config.credentials.tokenUrl) || options.uri.includes(this.webex.config.credentials.authorizeUrl)) {
-      return options;
+    if (typeof options.uri === 'string') {
```
adamweeks: If options is undefined, it will throw an error, let's add a check: 

```suggestion
    if (options.uri && typeof options.uri === 'string') {
```

mwegman: or you can do `onRequest(options = {}) {` on line 28

PR #1351, opened 2019-09-23
Share 409 error

arun3528:
# Pull Request Template

## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
@@ -362,6 +362,21 @@ export default class LocusInfo extends EventsScope {
           }
         );
       }
+      // If you are  guest and you are removed from the meeting
+      // You wont get any further events
+      else if (this.parsedLocus.self.removed) {
+        // Check if we need to send an event
+        this.emitScoped(
+          {
+            file: 'locus-info',
+            function: 'isMeetingActive'
+          },
+          EVENTS.DESTROY_MEETING,
+          {
+            shouldLeave: false
+          }
+        );
+      }
     }
     else {
       LoggerProxy.logger.warn('LocusInfo#isMeetingActive-> Meeting Type is unknown.');
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -1970,9 +1970,13 @@ export default class Meeting extends StatelessWebexPlugin {
         })
         .then(() => new Promise((resolve, reject) => {
           let timerCount = 0;
+
           // eslint-disable-next-line func-names
           // eslint-disable-next-line prefer-arrow-callback
-          const joiningTimer = setInterval(function () {
+          if (this.type === _CALL_) {
+            resolve();
+          }
+          const joiningTimer = setInterval(() => {
             timerCount += 1;
             if (this.meetingState === FULL_STATE.ACTIVE) {
               clearInterval(joiningTimer);
@@ -1983,7 +1987,7 @@ export default class Meeting extends StatelessWebexPlugin {
               clearInterval(joiningTimer);
               reject(new Error('Meeting is still not active '));
             }
-          }.bind(this), 1000);
+          }, 1000);
         }))
         .then(() => this.roap
           .sendRoapMediaRequest({
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -37,7 +37,8 @@ import {
   MEETING_STATE_MACHINE,
   _JOIN_,
   STATS,
-  MEETING_STATE
+  MEETING_STATE,
+  FULL_STATE
 } from '../constants';
 import ParameterError from '../common/errors/parameter';
 import MediaError from '../common/errors/media';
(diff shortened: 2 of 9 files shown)


packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -362,6 +362,21 @@ export default class LocusInfo extends EventsScope {
           }
         );
       }
+      // If you are  guest and you are removed from the meeting
+      // You wont get any further events
```
andrewholsted: I thought that was the intended behavior? Once you're booted you shouldn't care about the meeting any more.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1967,17 +1968,36 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            })
-            .catch((err) => {
-              this.logger.error('Meeting:index#join --> Error joining the call on send roap media request, ', err);
-              throw err;
-            }))
+        .then(() => new Promise((resolve, reject) => {
+          let timerCount = 0;
+          // eslint-disable-next-line func-names
+          // eslint-disable-next-line prefer-arrow-callback
+          const joiningTimer = setInterval(function () {
```
andrewholsted: why couldn't you use an arrow function here?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1967,17 +1968,36 @@ export default class Meeting extends StatelessWebexPlugin {
...
+            if (this.meetingState === FULL_STATE.ACTIVE) {
+              clearInterval(joiningTimer);
+              resolve();
+            }
+
+            if (timerCount === 4) {
+              clearInterval(joiningTimer);
+              reject(new Error('Meeting is still not active '));
+            }
+          }.bind(this), 1000);
```
andrewholsted: This effectively just waits 4 seconds for the meeting to be Active? I guess for our specific case we know we need to wait for the device to join but for others it may be different?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1967,17 +1968,36 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        }))
+        .then(() => this.roap
+          .sendRoapMediaRequest({
+            sdp: this.mediaProperties.peerConnection.sdp,
+            roapSeq: this.roapSeq,
+            meeting: this // or can pass meeting ID
+          })
+          .then(() => {
+            console.log('send share request ', new Date());
```
andrewholsted: was this just for debugging? Should it use the sdk logger?

packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
```diff
@@ -272,7 +272,7 @@ class Metrics {
       errorCode = 4008;
     }
 
-    return this.createErrorPayloadWithErrorCode(
+    return this.generateErrorPayload(
```
andrewholsted: thanks for the fix

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
arun3528: There are space meeting scenario where you are in IDLE state but you will get mercury event . In this case the meeting is still active but he is removed so we nee to remove the meeting object

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: `this` object was undefined . SO i had to put the function with bind . Any other suggestions !?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: In 98% of the case the other user is JOINED with in a second so there wont be a wait. 
After /call api we get a mercury event which tells the meeting state to be JOINED. And then the /media request

We can fix this in two ways 
1) Set some flag on the addMedia wait for the mercury state change than do the api call.
     needs more work on store , action and if for some reason mercury event doesn't come the addMedia wont get triggered and user wont know it.
2) The above approach, I felt it was simple assuming the join event will surely come with in 4 seconds and if not we throw an promise error so the user can handle it.

Lets check with adam once and see what he thinks about it

arun3528: Logically /call /media and /mediaShare needs to be 3 separate calls but internally we do all three together

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1967,17 +1968,37 @@ export default class Meeting extends StatelessWebexPlugin {
...
+            resolve();
+          }
+          const joiningTimer = setInterval(() => {
+            timerCount += 1;
+            if (this.meetingState === FULL_STATE.ACTIVE) {
+              clearInterval(joiningTimer);
+              resolve();
+            }
+
+            if (timerCount === 4) {
```
chrisadubois: I do not like this

PR #1357, opened 2019-09-26
fix(meetings): fix members:update event for recording

chrisadubois:
fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-95606


packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -0,0 +1,34 @@
+const controlsUtils = {};
```
adamweeks: Not a blocker, but to match the other util files in the meetings plugin, this should be capitalized.

```suggestion
const ControlsUtils = {};
```

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * parse the relevant host values that we care about: id
+ * @param {Object} controls
```
adamweeks: Can we add a `@typedef` for the controls object's shape?
Then we can do 
```suggestion
 * @param {ControlsObject} controls
```

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -0,0 +1,34 @@
...
+ * @returns {Object} parsed host or null if host was undefined
```
adamweeks:
```suggestion
 * @returns {Object} parsedObject - parsed host or null if host was undefined
 * @returns {string} parsedObject.recordingId
```

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -0,0 +1,34 @@
...
+    };
+  }
+
+  return null;
+};
+
+/**
+ * Extract the id from the record controls object
+ * @param {Object} controls
+ * @returns {String}
```
adamweeks:
```suggestion
 * @returns {String|null}
```

packages/node_modules/@webex/plugin-meetings/src/member/util.js
```diff
@@ -126,8 +126,27 @@ MemberUtil.isMuted = (status, accessor) => {
...
+      return controls.record.meta.modifiedBy;
+    }
+  }
+
+  return null;
+};
+
 /**
  * @param {Object} participant the locus participant
+ * @param {Object} locus the meeting instance
```
adamweeks: JSDoc updated, but function wasn't

packages/node_modules/@webex/plugin-meetings/src/member/index.js
```diff
@@ -215,7 +215,6 @@ export default class Member {
       this.email = MemberUtil.extractEmail(participant);
       this.isAudioMuted = MemberUtil.isAudioMuted(participant);
       this.isVideoMuted = MemberUtil.isVideoMuted(participant);
-      this.isRecording = MemberUtil.isRecording(participant);
```
adamweeks: Would it be easier to move the logic from `processIsRecording` to this function?

packages/node_modules/@webex/plugin-meetings/src/member/util.js
```diff
@@ -126,8 +126,27 @@ MemberUtil.isMuted = (status, accessor) => {
   return null;
 };
 
+/**
+ * utility method for getting the recording member for later comparison
+ * @param {Object} controls
+ * @returns {String}
```
adamweeks:
```suggestion
 * @returns {String|null}
```

packages/node_modules/@webex/plugin-meetings/src/member/util.js
```diff
@@ -126,8 +126,27 @@ MemberUtil.isMuted = (status, accessor) => {
...
+ */
+MemberUtil.getRecordingMember = (controls) => {
+  if (!controls) {
+    return null;
+  }
+  if (controls.record && controls.record.recording) {
```
adamweeks: You could do this instead of the additional if on 139
```suggestion
  if (controls.record && controls.record.recording && controls.record.meta) {
```

packages/node_modules/@webex/plugin-meetings/src/member/util.js
chrisadubois: yeah i thought i had pushed this one, the jsdoc is inaccurate

packages/node_modules/@webex/plugin-meetings/src/member/index.js
chrisadubois: No -- in this case that breaks the current pattern set up

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
chrisadubois: hostUtils, selfUtils, mediaSharesUtils etc are all camel case

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
chrisadubois: It does not have a consistent shape, it is different depending on what happens in the meeting, but I will update the shape for what we use as of now

packages/node_modules/@webex/plugin-meetings/src/member/util.js
chrisadubois: done

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
adamweeks: In locus-info you are importing them like this: 
```js
import ControlsUtils from '../locus-info/controlsUtils';
import InfoUtils from '../locus-info/infoUtils';
import FullState from '../locus-info/fullState';
import SelfUtils from '../locus-info/selfUtils';
import HostUtils from '../locus-info/hostUtils';
``` 
I'd like for them to be consistent with how they are imported. No need to fix the others in this PR, though.

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
adamweeks: I see in `getId` you are calling it `LocusControls`. Let's just use that for now:
```suggestion
 * @param {LocusControls} controls
```

chrisadubois:
closed as dupe of https://github.com/webex/webex-js-sdk/pull/1360

PR #1361, opened 2019-10-03
Create test-users package

adamweeks:
# Test Users

## Description

Migrates the test users legacy package to a new package that lives within the Webex repo.

New features added:

* Uses the request library from http-core
* Ability to override env vars via params
* Documentation!

Note: this PR is simply for the new package. Once approved and merged, existing tests will be converted over to use the new library.

## Type of Change

## Test Coverage

- [X] `npm run test -- --packages @webex/test-users`


packages/node_modules/@webex/test-users/README.md
```diff
@@ -0,0 +1,63 @@
...
+```javascript
+import {createTestUser, removeTestUser} from '@webex/test-users';
+
+createTestUser({displayName: 'Test User'}).then((myTestUser) => {
+  // Do something amazing with myTestUser
+
+  // When done, remove the test user
+  removeTestUser(myTestUser)
+});
+```
```
InteractiveTimmy: Glad we're adding usage notes.

packages/node_modules/@webex/test-users/src/generate-random-string.js
```diff
@@ -0,0 +1,17 @@
...
+  const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+  let text = '';
+
+  for (let i = 0; i < length; i += 1) {
+    text += chars.charAt(Math.floor(Math.random() * chars.length));
+  }
+
+  return text;
+}
```
InteractiveTimmy: Would it be better to use the UUID utility here?

packages/node_modules/@webex/test-users/src/index.js
```diff
@@ -0,0 +1,290 @@
...
+      client_id: clientId,
+      client_secret: clientSecret
+    },
+    headers: {
+      // Note: we can't request's auth hash here because this endpoint expects
+      // us to send the auth header *without including "Basic "* before the
+      // token string
+      authorization: btoa(`${clientId}:${clientSecret}`)
+    }
+  })
```
InteractiveTimmy: Great use of decoupling here to pull the params away from other resources. This allows the plugin to potentially work outside of all other plugins.

packages/node_modules/@webex/test-users/src/index.js
```diff
@@ -0,0 +1,290 @@
...
+ * Fetches credentials to talk to the test_users_s endpoint
+ *
+ * Caches result in `clientToken` variable for multiple runs
+ * @param {Object} options
+ * @param {string} options.clientId
+ * @param {string} options.clientSecret
+ * @param {string} options.idbrokerUrl
+ * @private
+ * @returns {String}
+ */
```
InteractiveTimmy: Should we change this to a `@typedef` for `options`?

packages/node_modules/@webex/test-users/src/index.js
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Makes a request authorized with client credentials
+ * @param {Object} options
+ * @param {Object} options.body
+ * @param {string} options.body.clientId
+ * @param {string} options.body.clientSecret
+ * @param {string} options.body.idbrokerUrl
+ * @private
+ * @returns {Promise<HttpResponseObject>}
```
InteractiveTimmy: Another `@typedef` opportunity?

packages/node_modules/@webex/test-users/src/index.js
```diff
@@ -0,0 +1,290 @@
...
+  * @property {string} type
+  * @property {Array.<string>} entitlements
+  * @property {string} orgId
+  * @property {AccessTokenObject} token
+  */
+
+/**
+ * Creates a test user
+ * @param {CreateUserOptions} options
+ * @returns {Promise.<TestUserObject>}
```
InteractiveTimmy: In relation to previous comments, should we only create `@typedef` references on larger objects as displayed here?

packages/node_modules/@webex/test-users/src/index.js
adamweeks: I didn't create one for this because it really is just a subset of the required params from `CreateUserOptions` typedef.

packages/node_modules/@webex/test-users/src/index.js
adamweeks: Yep! That's what I was going for here. Since this is meant to be used outside of the actual "webex" ecosystem.

packages/node_modules/@webex/test-users/src/index.js
adamweeks: We should definitely use `@typedef` for large objects on public functions. For some of the private functions that received the options object but only used a small subset of the params, I simply defined the params the function used.

.env.default
```diff
@@ -25,6 +25,7 @@ IDENTITY_BASE_URL="https://identitybts.webex.com"
 U2C_SERVICE_URL="https://u2c-intb.ciscospark.com/u2c/api/v1"
 WDM_SERVICE_URL="https://wdm-intb.ciscospark.com/wdm/api/v1"
 WHISTLER_API_SERVICE_URL="https://whistler.onint.ciscospark.com/api/v1"
+WEBEX_TEST_USERS_CONVERSATION_SERVICE_URL="https://conversation-intb.ciscospark.com/conversation/api/v1"
```
adamweeks: This will be a breaking change for anyone switching over to the new test users library. Previously, the URL was hard coded based on the WDM url. This package will require it to be explicitly set in either this env var or passed as a config option to `createTestUser`.

packages/node_modules/@webex/test-users/README.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+npm install --save @webex/test-users
+```
+
+## Usage
+
+With the test users library, you can create and remove Webex test users:
+
+```javascript
+import {createTestUser, removeTestUser} from '@webex/test-users';
```
adamweeks: Should we add some more usage options to this section?

packages/node_modules/@webex/test-users/src/generate-random-string.js
adamweeks: I'll give it a shot, the only place it is used is for generating a password.

PR #1363, opened 2019-10-07
Breaking Change: meetings | federation convoId -> convoUrl

chrisadubois:
The above is being added because currently, convoId will not work at all. We have to do some testing and documentation changes, as with federation/fedramp convoId is no longer present (and no longer uniquely identified a conversation)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
@@ -95,7 +95,7 @@ export default class MeetingInfo extends StatelessWebexPlugin {
     return MeetingInfoUtil.generateOptions({
       destination,
       type,
-      conversationServiceUrl: this.webex.internal.device.services.conversationServiceUrl
+      webex: this.webex
     });
   }
 
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
@@ -2,13 +2,13 @@ import url from 'url';
 
 import btoa from 'btoa';
 
+import ParameterError from '../common/errors/parameter';
 import LoggerProxy from '../common/logs/logger-proxy';
 import {
   _SIP_URI_,
   _PERSONAL_ROOM_,
   _MEETING_ID_,
   _CONVERSATION_URL_,
-  CONVERSATIONS,
   _LOCUS_ID_,
   _MEETING_LINK_,
   HTTP_VERBS,
@@ -19,6 +19,7 @@ import {
   ALTERNATE_REDIRECT_TRUE,
   DIALER_REGEX,
   WEBEX_DOT_COM,
+  CONVERSATION_SERVICE,
   WWW_DOT,
   JOIN,
   MEET,
@@ -61,6 +62,8 @@ MeetingInfoUtil.isMeetingLink = (value) => {
 
   return hostNameBool && pathNameBool;
 };
+
+MeetingInfoUtil.isConversationUrl = (value, webex) => webex.internal.services.getClusterId(value).endsWith(CONVERSATION_SERVICE);
 
 MeetingInfoUtil.convertLinkToSip = (value) => {
   const parsedUrl = MeetingInfoUtil.getParsedUrl(value);
@@ -108,7 +111,7 @@ MeetingInfoUtil.isPhoneNumber = (phoneNumber) => {
 };
 
 MeetingInfoUtil.generateOptions = (from) => {
-  const {destination, type, conversationServiceUrl} = from;
+  const {destination, type, webex} = from;
 
   if (type) {
     return {
@@ -130,9 +133,12 @@ MeetingInfoUtil.generateOptions = (from) => {
     options.type = _SIP_URI_;
     options.destination = destination;
   }
-  else {
+  else if (MeetingInfoUtil.isConversationUrl(destination, webex)) {
     options.type = _CONVERSATION_URL_;
-    options.destination = `${conversationServiceUrl}/${CONVERSATIONS}/${destination}`;
+    options.destination = destination;
+  }
+  else {
+    throw new ParameterError('MeetingInfo is fetched with meeting link, sip uri, phone number, or a conversation url.');
   }
 
   return options;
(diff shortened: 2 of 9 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1083,7 +1083,7 @@ export default class Meeting extends StatelessWebexPlugin {
   parseMeetingInfo(meetingInfo) {
     // MeetingInfo will be undefined for 1:1 calls
     if (meetingInfo && meetingInfo.body && !(meetingInfo.errors && meetingInfo.errors.length > 0)) {
-      this.convoId = meetingInfo.body.convoId || this.convoId;
+      this.convoUrl = meetingInfo.body.conversationUrl || this.convoUrl;
```
adamweeks: In the jsdoc it is defined as `meetingInfo.conversationUrl` but we are looking at `meetingInfo.body.conversationUrl`. Should those be changed in the jsdoc?

packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
```diff
@@ -95,7 +95,7 @@ export default class MeetingInfo extends StatelessWebexPlugin {
     return MeetingInfoUtil.generateOptions({
       destination,
       type,
-      conversationServiceUrl: this.webex.internal.device.services.conversationServiceUrl
+      webex: this.webex
```
andrewholsted: why pass the full webex now instead?

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -130,9 +135,17 @@ MeetingInfoUtil.generateOptions = (from) => {
...
-  else {
+  else if (MeetingInfoUtil.isConversationUrl(destination, webex)) {
     options.type = _CONVERSATION_URL_;
-    options.destination = `${conversationServiceUrl}/${CONVERSATIONS}/${destination}`;
+    if (destination.includes(`${CONVERSATION_URL_API}${CONVERSATION_URL_GET_API}`)) {
+      options.destination = destination;
+    }
+    else {
+      options.destination = destination.replace(CONVERSATION_URL_API, `${CONVERSATION_URL_API}${CONVERSATION_URL_GET_API}`);
+    }
```
andrewholsted: I don't understand this logic. if destination doesn't include `conversation/api/v1/conversations` replace `conversation/api/v1` with `conversation/api/v1/conversations`. Is this specifically for the case where destination was an id? If so, I think I would rather check if destination is a UUID and then do the logic, or make it a breaking change. this appears to be making it backwards compatible (which is ok for long standing APIs, but for new stuff, I don't mind s hipping breaking changes)

chrisadubois: it's to allow users to pass a conversation id with and without the GET API included. I'm fine with doing it with just a conversation_url.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/index.js
chrisadubois: Because now we can't read from that, we have to get the clusterId and verify that the convo url is the cluster being looked at.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
andrewholsted: Correct. So here we are making it backward compatible, but other places we are removing support for it. Like this change

```
 // If the destination is a UUID, assume it is an internal conversation ID.
  if (UUID_REG.test(dest)) {
    return dest;
  }

  // Next, assume it is a valid Hydra room ID.
  // If decoding fails, it is a SIP or other destination.
  // if the id is not present we will use the default destination
  return deconstructHydraId(dest).id || dest;
  return dest;
```

I'm fine just enforcing convoUrl and no longer accepting convoId as a param. @adamweeks? Otherwise we do need to default to the cluster url. I would prefer that destination check explicitly for a UUID instead of checking if it contains the url if that's the case we are checking for.

mwegman: If the destination doesn't have `conversations` in it. What does it look like? It's not a URL to a conversation.

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
```diff
@@ -26,15 +21,7 @@ MeetingsUtil.extractDestination = (destination, type) => {
...
 
-  // If the destination is a UUID, assume it is an internal conversation ID.
-  if (UUID_REG.test(dest)) {
-    return dest;
-  }
-
-  // Next, assume it is a valid Hydra room ID.
-  // If decoding fails, it is a SIP or other destination.
-  // if the id is not present we will use the default destination
-  return deconstructHydraId(dest).id || dest;
```
arun3528: Are we removing this because we are going to use the convo URL, And will the convo ID in the convo Url need deconstruction as it could be public facing

chrisadubois: arun this was removed because it was broken and causing bugs. and we also cannot infer by convoId alone. so we are going to just use convo url

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
chrisadubois: I think i can dig in, we might not even need to splice in the /conversations portion, i will check

adamweeks: I agree, we should not support UUID convo id at all.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -130,9 +135,17 @@ MeetingInfoUtil.generateOptions = (from) => {
     options.type = _SIP_URI_;
     options.destination = destination;
   }
-  else {
+  else if (MeetingInfoUtil.isConversationUrl(destination, webex)) {
     options.type = _CONVERSATION_URL_;
-    options.destination = `${conversationServiceUrl}/${CONVERSATIONS}/${destination}`;
+    if (destination.includes(`${CONVERSATION_URL_API}${CONVERSATION_URL_GET_API}`)) {
```
adamweeks: Do we even need this if/else block? If it is a conversation url, can we just `options.destination = destination;`?

chrisadubois: no we do not, thats what I was saying above.

adamweeks:
Samples tests are failing at an unusual point. Can you verify?

PR #1371, opened 2019-10-09
feat(test-helper-test-users): convert to test-users

adamweeks:
# Pull Request Template

## Description

Migrates the test-helper-test-users package to use the built in test-user library. 

## Type of Change

## Test Coverage

Ran tests for multiple plugins and samples. All run successfully.

### DIFF
--- a/packages/node_modules/@webex/test-helper-test-users/src/index.js
+++ b/packages/node_modules/@webex/test-helper-test-users/src/index.js
@@ -2,29 +2,14 @@
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
 
-
-/* istanbul ignore next */
-if (typeof Promise === 'undefined') {
-  // eslint-disable-next-line global-require
-  require('es6-promise').polyfill();
-}
-
 /* eslint-env mocha */
 /* eslint camelcase: [0] */
 
-const _ = require('lodash');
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/test-helper-test-users/src/index.js
```diff
@@ -79,50 +64,6 @@ function _create(options) {
   }
 }
 
-/**
- * Gets users from environment variable
- * @param {Object} options
- * @returns {Array<User>}
- */
-function _extractFromEnv(options) {
```
adamweeks: This function will never be called because the test users library is now accessible by everyone. We can add this feature back in if it ever is requested.

packages/node_modules/@webex/test-helper-test-users/src/index.js
```diff
@@ -2,29 +2,14 @@
...
-try {
-  /* eslint global-require: [0] */
-  TestUsers = require('@ciscospark/test-users-legacy');
-}
-catch (e) {
-  // ignore
-}
+const _ = require('lodash');
+const retry = require('@webex/test-helper-retry');
+const {createTestUser, removeTestUser} = require('@webex/test-users');
```
adamweeks: This is the main change of this PR. Instead of using `test-users-legacy`.

Issue #1366, opened 2019-10-08
@webex/plugin-meetings: Calling updateVideo and updateAudio results in incorrect mute states reported


antti2000:
## Expected Behavior
Calling both meeting.updateVideo and meeting.updateAudio with new local camera and microphone streams within a short period of time (almost consecutively), should result in correct isAudioMuted / isVideoMuted states being sent through members:update callback.

## Current Behavior
Calling both meeting.updateVideo and meeting.updateAudio with new local camera and microphone streams within a short period of time results in incorrect isAudioMuted / isVideoMuted states being sent through members:update callback. The states do not become eventually consistent (through new events) with the sender client. Instead the sender client thinks it is sending both audio and video feeds, but the receiving clients think sender has microphone, camera or both muted (behavior intermittent). Audio stream comes through to the receiving clients pretty much always, video stream sometimes.

If you after a while (> 1s) mute the microphone or camera on the sender client, the audio / video mute states are resolved to correct ones through new events sent to members:update.

Also, adding a delay of approx 50ms - 100ms between the meeting.updateVideo and meeting.updateAudio calls seems to resolve this issue as well.

## Possible Solution (we welcome any logical suggestion)
Each updateVideo/updateAudio call should result in a separate members.update event being sent with the information representing changes occurred due to each individual call. Alternatively, the updates from both calls could potentially be combined into a single result payload that correctly describe the mute audio / video state.

## Steps to Reproduce
1. Start / join meeting with no local media
2. Call updateVideo and updateAudio in a row with local camera and microphone streams
3. Examine the events sent for members:update event for isAudioMuted and isVideoMuted

## Environment Details
* SDK Version 1.80.17
* Node/Browser Version 77.0.3865.90

chrisadubois:
Hi @antti2000 

`updateAudio` and `updateVideo` are unrelated to `isAudioMuted` and `isVideoMuted`. They are not correlated in terms of the SDK members:update event'ing for plugin-meetings.

in your statement you say, "Each updateVideo/updateAudio call should result in a separate members.update event being sent with the information representing changes occurred due to each individual call" -- this is incorrect. neither of these correlate to members update events, they are influenced by media streams only, and only would be specified to media:* events. If the user has previously muted the video or audio, and then they call an updateAudio/Video, the state should persist, the user should not see a members:update event, but the value will remain the same, and video/audio should not be sent if they were muted beforehand.

[...]

PR #1373, opened 2019-10-10
fix(media): prevent unstable media stream updates

chrisadubois:
fixes: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-98172
fixes: https://github.com/webex/webex-js-sdk/issues/1366


adamweeks:
I understand how this fix helps, but maybe we can do something more proactive for the developer to let them know that media can update besides throwing an error?  Would a `canUpdate` computed property or function be worthwhile? Something like:

```js
canUpdate() {
  return this.mediaProperties.peerConnection.signalingState === SDP.STABLE;
}
```

chrisadubois:
Will add the function

PR #1376, opened 2019-10-11
Add proxy support for mercury web socket connections

decryan1984:
# Pull Request Template

## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


packages/node_modules/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -17,6 +19,8 @@ import {
   // NotFound
 } from './errors';
 
+const HttpsProxyAgent = require('https-proxy-agent');
```
gannons: Does the `https-proxy-agent` module support browsers? It does not mention it in the top level documentation.

Should stick with `import` in the sdk instead of `require(..)`.

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket-base.js
```diff
@@ -193,7 +195,7 @@ export default class Socket extends EventEmitter {
       const WebSocket = Socket.getWebSocketConstructor();
 
       this.logger.info('socket: creating WebSocket');
-      const socket = new WebSocket(url);
+      const socket = new WebSocket(url, [], options);
```
gannons: What was options used for before? i.e. See you did not add it.

decryan1984: It was used above with the Reflect.defineProperty to see the options on the socket object, but this does not appear to propagate them to the ws, since we could not create the web socket via the proxy without this change.

packages/node_modules/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -3,10 +3,13 @@
  */
 
 import url from 'url';
+
 import {WebexPlugin} from '@webex/webex-core';
 import {deprecated, oneFlight} from '@webex/common';
 import {camelCase, get, set} from 'lodash';
 import backoff from 'backoff';
+import HttpsProxyAgent from 'https-proxy-agent';
+
```
decryan1984: I didn't re-order these, I assume something is changing them as part of the commit process?

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket-base.js
```diff
@@ -193,7 +195,7 @@ export default class Socket extends EventEmitter {
       const WebSocket = Socket.getWebSocketConstructor();
 
       this.logger.info('socket: creating WebSocket');
-      const socket = new WebSocket(url);
+      const socket = new WebSocket(url, options);
```
decryan1984: I'm not sure how the reflect.defineProperty is supposed to work above, but we found that the options had to be passed to the websocket constructor for the agent to take effect.

webpack.config.js
```diff
@@ -28,7 +28,9 @@ module.exports = (env = process.env.NODE_ENV || 'production') => ({
     }
   },
   node: {
-    fs: 'empty'
+    fs: 'empty',
+    net: 'empty',
```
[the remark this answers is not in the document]
decryan1984: It works for me locally, from checking online it seemed to be a fairly widely used solution to the 'Module not found: Error: Can't resolve 'net'...' problem

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket-base.js
decryan1984: Yes, I'd assume that was always the intention

packages/node_modules/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -3,10 +3,13 @@
  */
 
 import url from 'url';
+
 import {WebexPlugin} from '@webex/webex-core';
 import {deprecated, oneFlight} from '@webex/common';
 import {camelCase, get, set} from 'lodash';
 import backoff from 'backoff';
+import HttpsProxyAgent from 'https-proxy-agent';
```
decryan1984: This module can only be used when running in node, it will not work in the browser.

adamweeks: What happens when run on the browser? The mercury plugin is utilized by the browser based sdk in multiple places.

adamweeks:
I know the readme for this plugin has nothing in it, but since we are in the process of updating all of them, could you add a section on the added environment variables what what they do in the "USAGE" section. While we don't have a "standard" documentation of them yet, something like this would be helpful: https://github.com/webex/webex-js-sdk/tree/master/packages/node_modules/%40webex/test-helper-test-users#environment-defaults

packages/node_modules/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -179,15 +178,10 @@ const Mercury = WebexPlugin.extend({
...
-
-        if (proxyAddress) {
+        // if a proxy agent has been configured we will add it here to successfully open a websocket
+        if (this.webex.config.defaultRequestOptions && this.webex.config.defaultRequestOptions.proxyAgent) {
           this.logger.info('mercury: setup proxy agent');
-
-          const proxyOption = url.parse(proxyAddress);
-
-          options.agent = new HttpsProxyAgent(proxyOption);
+          options.agent = this.webex.config.defaultRequestOptions.proxyAgent;
```
gannons: May be better to create a separate defaultRequestOptions. e.g. defaultMercuryRequestOptions.

decryan1984:
> I know the readme for this plugin has nothing in it, but since we are in the process of updating all of them, could you add a section on the added environment variables what what they do in the "USAGE" section. While we don't have a "standard" documentation of them yet, something like this would be helpful: https://github.com/webex/webex-js-sdk/tree/master/packages/node_modules/%40webex/test-helper-test-users#environment-defaults

I've removed all references to those environmental variables Adam, we now configure the agent in our client and inject it using the defaultRequestOption model that Will added recently

adamweeks:
> I've removed all references to those environmental variables Adam, we now configure the agent in our client and inject it using the defaultRequestOption model that Will added recently

Nice @decryan1984, I like that approach. Any chance you could add a basic proxy usage guide to the mercury readme?

decryan1984:
Test jobs are failing on jenkins, but internal-plugin-Mercury tests were run locally and passed, so it appears to be an issue on the pipeline

packages/node_modules/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -400,6 +400,76 @@ describe('plugin-mercury', () => {
...
+            .then(() => {
+              assert.isTrue(mercury.connected, 'Mercury is connected');
+              assert.isFalse(mercury.connecting, 'Mercury is not connecting');
+              assert.calledWith(socketOpenStub, sinon.match(/ws:\/\/example.com/), sinon.match.has('agent',
+                sinon.match.has('proxy', sinon.match.has('href', testProxyUrl))));
+            });
+        });
+
+        it('connects to Mercury without proxy agent', () => {
+          const promise = mercury.connect();
```
adamweeks: I think you need to `delete webex.config.defaultMercuryOptions;` here so the config doesn't get passed through.

packages/node_modules/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -400,6 +400,76 @@ describe('plugin-mercury', () => {
...
+
+          assert.isFalse(mercury.connected, 'Mercury is not connected');
+          assert.isTrue(mercury.connecting, 'Mercury is connecting');
+          mockWebSocket.open();
```
adamweeks: Are these two tests duplicates?

decryan1984: Yeah, good catch, they were slightly different to begin with but have converged, I'll remove one

packages/node_modules/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
decryan1984: Will clean up that config after each test to be on the safe side

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket-base.js
adamweeks: It does look like you "fixed the glitch" here. Let's hope no one was counting on the bug!

img

packages/node_modules/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -400,6 +400,65 @@ describe('plugin-mercury', () => {
...
+          return promise
+            .then(() => {
+              assert.isTrue(mercury.connected, 'Mercury is connected');
+              assert.isFalse(mercury.connecting, 'Mercury is not connecting');
+              assert.calledWith(socketOpenStub, sinon.match(/ws:\/\/example.com/), sinon.match.has('agent',
+                sinon.match.has('proxy', sinon.match.has('href', testProxyUrl))));
+            });
+        });
+
+        it('connects to Mercury using https proxy agent', () => {
```
adamweeks: This is a duplicate test.

decryan1984: I guess you're right, without the HTTP_PROXY/HTTPS_PROXY env vars, there's no need for two different tests here

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket-base.js
decryan1984: Yeah, my bad on this one. Although I did try reverting this back to passing just the url and the Jenkins failures still happened. Anyway, at this point I'm just happy it's passing!

packages/node_modules/@webex/internal-plugin-mercury/src/socket/socket.js
```diff
@@ -2,9 +2,10 @@
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
 
-import Socket from './socket-base';
 import WS from 'ws';
 
+import Socket from './socket-base';
+
```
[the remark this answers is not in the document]
decryan1984: Most of these import reshuffles seem to be done automatically during the commit process

PR #1377, opened 2019-10-14
SPARK-97477: [JS SDK] Make a call to /bytes API while downloading file

meeshaT:
## Description

In _**encryption.js**_ I have added a call to **_/bytes_** API in **_callDownloadUrl()** function. If **_/bytes_** API return response code **302** then another GET request will be made by the browser to URL present in **LOCATION** attribute in headers. If this GET request is successful then it will go in **then** block where the call to **URL** present in **LOCATION** attribute in **headers** will be made again to download the file & then decrypt it, If GET request call returns error or **_/bytes_** API call returns response code other then **_302_** our flow will go in catch block where we reject the promise with the error object back to the client.

**_NOTE_**: As per Anti-Malware Scanning feature **/bytes API** call should return **302** if file is **not infected**, **410** if file **is infected** and **423** if file scanning is **in process**.

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules****


packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -113,17 +101,36 @@ const Encryption = WebexPlugin.extend({
...
-          this.logger.warn('encryption: could not determine download url for `scr.loc`; attempting to download `scr.loc` directly');
+        return _this2._callDownloadUrl(url, scr, res);
+      });
+  },
 
-          return scr.loc;
-        }
-        this.logger.info('encryption: retrieved download url for encrypted file');
+  _callDownloadUrl(url, scr, result) {
+    const _this2 = this;
```
InteractiveTimmy: I'm not too sure why `this` is being reinstated here. Typically, we should avoid this type of declaration wherever possible. In cases in which `this` needs to be a different object in memory, it can be associated using `method.bind(myObj);` where `myObj` would be aliased as `this`.

Alternatively, since these appear to be class methods, simply removing these lines and utilizing `this` may be a solution.

Also note, Javascript makes namespace assignments mutable for non-primitive data, so both `this` and `_this2` point to the same space in memory, making them identical in every way even after data has been changed on either namespace.

packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -113,17 +101,36 @@ const Encryption = WebexPlugin.extend({
...
-      });
+    return this.request({
+      method: 'GET',
+      uri: url
+    }).then(() => _this2._helperFunction(scr, result))
+      .catch((err) => Promise.reject(err));
+  },
+
+  _helperFunction: function _helperFunction(scr, result) {
+    const _this = this;
```
InteractiveTimmy: See my other comment on reinstating `this`.

packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -113,17 +101,36 @@ const Encryption = WebexPlugin.extend({
...
-        return url;
-      });
+    return this.request({
+      method: 'GET',
+      uri: url
+    }).then(() => _this2._helperFunction(scr, result))
+      .catch((err) => Promise.reject(err));
+  },
+
+  _helperFunction: function _helperFunction(scr, result) {
```
InteractiveTimmy: It appears that JSDoc is missing from this method, take a look at [this file](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/webex-core/src/lib/services/services.js#L55) for reference.

packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -113,17 +101,36 @@ const Encryption = WebexPlugin.extend({
...
-        if (!url) {
-          this.logger.warn('encryption: could not determine download url for `scr.loc`; attempting to download `scr.loc` directly');
+        return _this2._callDownloadUrl(url, scr, res);
+      });
+  },
 
-          return scr.loc;
-        }
-        this.logger.info('encryption: retrieved download url for encrypted file');
+  _callDownloadUrl(url, scr, result) {
```
InteractiveTimmy: This method is missing its JSDoc, see [this file](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/webex-core/src/lib/services/services.js#L55) for an example.

packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -72,28 +72,16 @@ const Encryption = WebexPlugin.extend({
...
-      .then((res) => this.decryptBinary(scr, res.body));
+    const promise = this._fetchDownloadUrl(scr);
 
     proxyEvents(shunt, promise);
 
     return promise;
   },
 
   _fetchDownloadUrl(scr) {
+    const _this2 = this;
```
InteractiveTimmy: I see `_this2` is [called later](https://github.com/webex/webex-js-sdk/pull/1377/files#diff-22291d397df84ad20147dd45e8209ce4R106) when it could be called as `this`.

packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -113,17 +101,36 @@ const Encryption = WebexPlugin.extend({
       }
     })
       .then((res) => {
-        const url = res.body.endpoints[scr.loc];
+        const url = scr.loc;
 
-        if (!url) {
-          this.logger.warn('encryption: could not determine download url for `scr.loc`; attempting to download `scr.loc` directly');
+        return _this2._callDownloadUrl(url, scr, res);
```
InteractiveTimmy: `_this2` should be able to be just `this`, mentioned [here](https://github.com/webex/webex-js-sdk/pull/1377/files#diff-22291d397df84ad20147dd45e8209ce4R83).

PR #1379, opened 2019-10-14
fix(meeting): join with recording notification

chrisadubois:
fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-98302

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
@@ -33,9 +33,16 @@ import Parser from '../locus-info/parser';
 export default class LocusInfo extends EventsScope {
   constructor(updateMeeting, locus = {}, webex, meetingId) {
     super();
+    this.emitChange = false;
+    this.parsedLocus = {
+      states: []
+    };
+    this.webex = webex;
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -33,6 +33,7 @@ import Parser from '../locus-info/parser';
 export default class LocusInfo extends EventsScope {
   constructor(updateMeeting, locus = {}, webex, meetingId) {
     super();
+    this.updateControls(locus.controls);
```
adamweeks: Could we move this to ~line 58 with the other `update` functions to keep them grouped together? If not, let's add a comment as to why this is inconsistent.

chrisadubois: we cannot, so i will add a comment

chrisadubois: i'll replace this.controls with this.updateControls

adamweeks:
Looks like this needs a rebase, sorry @chrisadubois

chrisadubois:
we are putting this in a larger bug fix PR as one.

PR #1383, opened 2019-10-15
test(samples): add more wait for h264

adamweeks:


adamweeks:
@arun3528 has a more complicated idea on how to approach h264 downloading, but this is an interim step to get green builds again.

Issue #1255, opened 2019-07-17
@webex/plugin-meetings: cannot call muteAudio if there is no local video stream, or muteVideo if there is no local audio stream


jnetterf:
## Expected Behavior
`meeting.muteAudio()` should not cause an error if there is an audio stream, but no video stream.

`meeting.muteVideo()` should not cause an error if there is a video stream, but no audio stream.

## Current Behavior
Using the below code, I get the following exception, with the following partial trace:
```
TypeError: "meeting.audio is null"
    doToggle video.js:49
    onBeforeToggle video.js:130
    observeEvents state-machine.js:511
    observeEvents state-machine.js:507
    transit state-machine.js:450
    fire state-machine.js:425
    <anonymous> state-machine.js:624
    muteVideo index.js:1541
```

I also get an error about `meeting.video` when trying to mute audio without a video stream.

## Possible Solution (we welcome any logical suggestion)
Either `meeting.audio` and `meeting.video` should always be set, or the following callsites should not assume they are always set:
 - https://github.com/webex/webex-js-sdk/blob/d0441388bcf6d16934a37af8e670cbf630e1e1c7/packages/node_modules/@webex/plugin-meetings/src/meeting/video.js#L23
 - https://github.com/webex/webex-js-sdk/blob/d0441388bcf6d16934a37af8e670cbf630e1e1c7/packages/node_modules/@webex/plugin-meetings/src/meeting/audio.js#L23

## Steps to Reproduce
Run something like this:
```
const meeting = await setupWebexAndGetAMeeting();
await meeting.join();

const [localVideoStream, _] = await meeting.getMediaStreams({
  sendAudio: false,
  sendVideo: true,
  sendShare: false,
});

await meeting.addMedia({
  mediaSettings: {
    receiveVideo: true,
    receiveAudio: true,
    receiveShare: true,
    sendVideo: true,
    sendAudio: false,
    sendShare: false,
  },
  localStream: localVideoStream,
  localShare: null,
});

await meeting.muteVideo();
```

See the above stack trace.

## Environment Details
```
    "@webex/internal-plugin-wdm": "^1.66.2",
    "@webex/plugin-authorization": "^1.66.2",
    "@webex/plugin-device-manager": "^1.67.1",
    "@webex/plugin-meetings": "^1.67.0",
    "@webex/webex-core": "^1.66.2"
```

PR #1394, opened 2019-10-28
fix(meetings): mute if stream undefined fix

chrisadubois:
fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-100796


adamweeks:
Fixes #1255

PR #1402, opened 2019-11-04
Spark 102791/append additional cluster info

InteractiveTimmy:
# Pull Request

## Description

A recent change has appended new clusters to the U2C service host map that aren't being mapped to any specific service. The current methods used to map the services only map hosts with the same root host address as the key. In order to map the additional hosts, the hostIds from the host catalog must be analyzed against the service names in the serviceLinks section.

Fixes # SPARK-102783

## Type of Change

## Test Coverage

- [x] Modified and appended tests to the current plugin's test suite
- [x] Ran plugin test suite to confirm changes worked
- [x] Ran live usability tests to confirm that the changes function as expected

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/services/services.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/services.js
@@ -327,20 +327,68 @@ const Services = WebexPlugin.extend({
    * @returns {object}
    */
   _formatReceivedHostmap(serviceHostmap) {
-    const output = [];
-    let item = {};
+    // map the host catalog items to a formatted hostmap
+    const formattedHostmap = Object.keys(serviceHostmap.hostCatalog).reduce(
+      (accumulator, key) => {
+        if (serviceHostmap.hostCatalog[key].length === 0) {
+          return accumulator;
+        }
 
+        const serviceName = serviceHostmap.hostCatalog[key][0].id.split(':')[3];
+        const defaultUrl = serviceHostmap.serviceLinks[serviceName];
+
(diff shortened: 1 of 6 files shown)


packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -327,20 +327,49 @@ const Services = WebexPlugin.extend({
    * @returns {object}
    */
   _formatReceivedHostmap(serviceHostmap) {
-    const output = [];
-    let item = {};
+    let serviceItem;
```
mwegman: This doesn't need to be declared here. It looks like it is only used within the `map` call and the `forEach` call. You could declare it within the respective loops.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -327,20 +327,49 @@ const Services = WebexPlugin.extend({
...
+    const unmappedHosts = Object.keys(serviceHostmap.hostCatalog).filter(
+      (key) => mappedLinks.every(
+        (mappedLink) => mappedLink.defaultHost !== key
+      )
+    )
+      // map the filtered hosts against their service names
+      .map((key) => ({
+        serviceName: serviceHostmap.hostCatalog[key][0].id.split(':')[3],
+        hosts: serviceHostmap.hostCatalog[key]
+      }));
```
mwegman: Instead of looping over twice with `filter` and `map`, you can do it once using `reduce`

PR #1417, opened 2019-11-18
feat(plugin-phone): use services catalog for conversation url

adamweeks:
# Pull Request Template

## Description

Converting the phone plugin to use services plugin methods instead of the deprecated device plugin methods.

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-106018

## Test Coverage

- [X] `@webex/plugin-phone --integration`

### DIFF
--- a/packages/node_modules/@webex/plugin-phone/src/call.js
+++ b/packages/node_modules/@webex/plugin-phone/src/call.js
@@ -1010,46 +1010,12 @@ const Call = WebexPlugin.extend({
     this.locusJoinInFlight = true;
     this.logger.info('call: dialing');
 
-    let target = invitee;
-
-    if (base64.validate(invitee)) {
-      // eslint-disable-next-line no-unused-vars
-      const parsed = base64.decode(invitee).split('/');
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/plugin-phone/src/call.js
```diff
@@ -1061,6 +1027,46 @@ const Call = WebexPlugin.extend({
     return this;
   },
 
+  /**
+   * Gets the targe from the invitee
```
InteractiveTimmy: While `targe` does reference `target`, should we change this?

packages/node_modules/@webex/plugin-phone/src/call.js
```diff
@@ -1061,6 +1027,46 @@ const Call = WebexPlugin.extend({
...
+   * @param {string} invitee
+   * @returns {Promise.<string>} The locus target
+   */
+  getTargetFromInvitee(invitee) {
```
InteractiveTimmy: Great way to isolate this block of code as a method.

PR #1433, opened 2019-11-23
Spark 106008/convert webex core

InteractiveTimmy:
# Pull Request

## Description

This pull request encompasses all changes needed to fully decouple the `wdm` services catalog from `webex-core`. This includes adding new helper methods, adding new tests, as well as updating documentation.

Fixes # SPARK-106008

## Type of Change

## Test Coverage

- [x] Generated new tests to include all the added features
- [x] Updated all existing tests within the plugin to address the fix changes.
- [x] Ran `npm test -- --packages @webex/webex-core` with 100% passing tests.

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/README.md
+++ b/packages/node_modules/@webex/webex-core/README.md
@@ -30,7 +30,9 @@ To utilize the basic functionality of the `services` plugin that is bound to the
 
 The following environment variables are used by this plugin:
 
-- `U2C_SERVICE_URL` - stores the service catalog collecting url, typically the **U2C** service
+- `HYDRA_SERVICE_URL` - Stores the public hydra api url for managing Webex Teams resources.
+- `U2C_SERVICE_URL` - Stores the service catalog collecting url, typically the **U2C** service.
+- `VALIDATE_DOMAINS` - When `true`, the services plugin will utilize an *allowed domains* list to allow additional domains that don't exist within the service catalogs to be reached via service requests.
 
 ### Configuration
 
@@ -40,13 +42,25 @@ The `services` plugin that is bound to the `webex-core` plugin upon initializati
 const webex = new Webex({
   config: {
     services: {
-      hydra: 'https://api.ciscospark.com/v1'
+      // Services that are available before catalog retrieval.
+      discovery: {
+        hydra: 'https://api.ciscospark.com/v1'
+      },
+
+      // Validate domains against the allowed domains.
+      validateDomains: true,
+
+      // The allowed domains to validate domains against.
+      allowedDomains: ['allowed-domain']
     }
   }
 });
 ```
 
-The default configuration includes the **U2C** url as defined by environment variable `U2C_SERVICE_URL` or the default `https://u2c.wbx2.com/u2c/api/v1`
+The default configuration includes the following service urls:
+
+- `U2C_SERVICE_URL` [ **U2C** ] - `https://u2c.wbx2.com/u2c/api/v1`
+- `HYDRA_SERVICE_URL` [ **Hydra** ] - `https://api.ciscospark.com/v1`
 
 ## Maintainers
 
--- a/packages/node_modules/@webex/webex-core/src/config.js
+++ b/packages/node_modules/@webex/webex-core/src/config.js
@@ -19,7 +19,46 @@ export default {
   AlternateLogger: undefined,
   credentials: new CredentialsConfig(),
   services: {
-    u2c: process.env.U2C_SERVICE_URL || 'https://u2c.wbx2.com/u2c/api/v1'
+    /**
+     * A list of services that are available prior to catalog collection.
+     *
+     * @type {Object}
+     */
+    discovery: {
+      /**
+       * The hydra discovery url.
+       *
+       * @type {string}
+       */
+      hydra: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1',
+
+      /**
+       * The u2c discovery url
+       *
+       * @type {string}
+       */
+      u2c: process.env.U2C_SERVICE_URL || 'https://u2c.wbx2.com/u2c/api/v1'
+    },
+
+    /**
+     * When true, enable allowed domain validation.
+     *
+     * @type {boolean}
+     */
+    validateDomains: process.env.VALIDATE_DOMAINS === 'true' || false,
+
+    /**
+     * Contains a list of allowed domain host addresses.
+     *
+     * @type {Array<string>}
+     */
+    allowedDomains: [
+      'webexkubed.com',
+      'wbx3.com',
+      'wbx2.com',
+      'ciscospark.com',
+      'webex.com'
+    ]
   },
   device: {
     preDiscoveryServices: {
@@ -33,7 +72,7 @@ export default {
       'wbx3.com',
       'wbx2.com',
       'ciscospark.com',
-      'ciscospark.com'
+      'webex.com'
     ]
   },
   payloadTransformer: {
--- a/packages/node_modules/@webex/webex-core/src/interceptors/auth.js
+++ b/packages/node_modules/@webex/webex-core/src/interceptors/auth.js
@@ -50,64 +50,59 @@ export default class AuthInterceptor extends Interceptor {
   }
 
   /**
-   * Determines if the provided options object needs an auth header
+   * Determines if the provided options object needs an authorization header.
+   *
    * @param {Object} options
    * @returns {Promise<boolean>}
    */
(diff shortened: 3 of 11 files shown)


packages/node_modules/@webex/webex-core/src/config.js
```diff
@@ -19,7 +19,46 @@ export default {
...
+     * @type {boolean}
+     */
+    validateDomains: process.env.VALIDATE_DOMAINS === 'true' || false,
+
+    /**
+     * Contains a list of whitelisted host addresses.
+     *
+     * @type {Array<string>}
+     */
+    whitelist: [
```
adamweeks: Could we rename this to a more inclusive name like `allowedHosts`?

packages/node_modules/@webex/webex-core/src/config.js
```diff
@@ -19,7 +19,46 @@ export default {
...
+      'webexkubed.com',
+      'wbx3.com',
+      'wbx2.com',
+      'ciscospark.com',
+      'ciscospark.com'
```
adamweeks: This is a duplicate

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
```diff
@@ -50,64 +50,58 @@ export default class AuthInterceptor extends Interceptor {
...
+   * Determines if the provided options object needs an authorization header.
+   *
    * @param {Object} options
    * @returns {Promise<boolean>}
    */
   requiresCredentials(options) {
-    if (options.service === 'u2c') {
-      if (options.resource.includes('limited')) {
+    // Validate that authorization is necessary.
+    if (options.addAuthHeader === false) {
```
adamweeks: Could we do the opposite here as well? If addAuthHeader is true, resolve true?

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
```diff
@@ -50,64 +50,58 @@ export default class AuthInterceptor extends Interceptor {
...
-                if (matchingDomains.length) {
-                  return Promise.resolve(true);
-                }
-              }
-
-              return false;
-            });
-        }
+    // Validate that the domain of the uri is whitelisted.
+    if (services.hasWhitelist() && (uri && services.isWhitelistUrl(uri))) {
```
adamweeks:
```suggestion
    if (services.hadApprovedDomains() && (uri && services.isApprovedUrl(uri))) {
```

packages/node_modules/@webex/webex-core/src/config.js
InteractiveTimmy: This can be done and makes more sense, let me get to this change.

packages/node_modules/@webex/webex-core/src/config.js
InteractiveTimmy: Currently, this will be a duplicate until we get to removing service methods from `internal-plugin-wdm`. In the mean time, we can do this if it helps any:

```js
const config = {
  ...
  services: {
    whitelist: [ ... ]
  }
  ...
};

config.device.whitelist = config.services.whitelist;
```

This saves memory, but increases diff.

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
InteractiveTimmy: If this interceptor is to match the same output when intercepting, we can not. If the `addAuthHeader` option is anything but false, it will check to confirm it's a service and then check to confirm if it's in the whitelist before adding the auth header.

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
InteractiveTimmy: This can be implemented.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -455,6 +480,18 @@ const Services = WebexPlugin.extend({
...
+  /**
+   * Determine if a provided url is in the catalog's allowed domains.
+   *
+   * @param {string} url - The url to match allowed domains against.
+   * @returns {boolean} - True if the url provided is allowed.
+   */
+  isAllowedDomainUrl(url) {
+    const catalog = this._getCatalog();
+
+    return !!(catalog.findAllowedDomain(url)) || false;
```
mwegman: You don't need the `|| false`

InteractiveTimmy: Good catch! Changes have been made.

packages/node_modules/@webex/webex-core/README.md
```diff
@@ -30,7 +30,8 @@ To utilize the basic functionality of the `services` plugin that is bound to the
 
 The following environment variables are used by this plugin:
 
-- `U2C_SERVICE_URL` - stores the service catalog collecting url, typically the **U2C** service
+- `HYDRA_SERVICE_URL` - Stores the public hydra api url for managing Webex Teams resources.
+- `U2C_SERVICE_URL` - Stores the service catalog collecting url, typically the **U2C** service.
```
adamweeks: Could you also add the `VALIDATE_DOMAINS` env var?

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -288,6 +288,18 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#hasService()', () => {
```
adamweeks: Since these don't interact with any actual services, should they be unit tests instead of integration?

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -332,6 +344,31 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#isAllowedDomainUrl()', () => {
```
adamweeks: These seem like unit tests as well

packages/node_modules/@webex/webex-core/test/unit/spec/interceptors/auth.js
```diff
@@ -136,136 +141,157 @@ describe('webex-core', () => {
...
+          };
+        });
+
+        it('resolves to false when services plugin does not exist',
+          () => {
+            delete webex.internal.services;
+
+            return interceptor.requiresCredentials({
+              uri: `${services.hydra}/ping`
+            });
```
adamweeks: You didn't assert anything in this test

packages/node_modules/@webex/webex-core/test/unit/spec/interceptors/auth.js
```diff
@@ -136,136 +141,157 @@ describe('webex-core', () => {
...
+          })
+            .then((response) => assert.isTrue(response)));
+
+        it('resolves to false with a non-allowed uri',
+          () => interceptor.requiresCredentials({
+            uri: 'https://not-allowed/resource'
+          })
+            .then((response) => assert.isFalse(response)));
+
+        afterEach('remove services plugin', () => {
```
adamweeks: Could you move this to immediately after the `beforeEach` hook for consistency with the other tests? (Trying to find an eslint rule like this jest one: https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/prefer-hooks-on-top.md)

packages/node_modules/@webex/webex-core/test/unit/spec/services/service-catalog.js
```diff
@@ -54,6 +54,62 @@ describe('webex-core', () => {
...
+        catalog.setAllowedDomains(domains);
+      });
+
+      it('returns a an array of allowed hosts', () => {
+        const list = catalog.getAllowedDomains();
+
+        assert.match(domains, list);
+      });
+
+      afterEach('remove allowed domains', () => {
```
adamweeks: Move this after the `beforeEach` please

packages/node_modules/@webex/webex-core/test/unit/spec/services/service-catalog.js
```diff
@@ -72,6 +128,36 @@ describe('webex-core', () => {
...
+          'example-e',
+          'example-f'
+        ];
+
+        catalog.setAllowedDomains(newValues);
+
+        assert.notDeepInclude(domains, newValues);
+      });
+
+      afterEach('remove allowed domains', () => {
```
adamweeks: Move this after the `beforeEach` please

PR #1436, opened 2019-11-25
chore(tooling): ignore package.json changes in diff

adamweeks:


tooling/commands/ci.js
```diff
@@ -45,8 +45,8 @@ module.exports = {
   },
   handler: wrapHandler(async (argv) => {
     let packages = argv.integration ?
-      await updated({dependents: true, npm: true}) :
-      await updated({npm: true});
+      await updated({dependents: true, npm: !!process.env.CI}) :
```
adamweeks: This is supposed to run on ci, so can we leave this as a manual `true`?

tooling/modified-packages.js
```diff
@@ -60,8 +60,8 @@ const updated = async ({dependents, npm}) => {
 
 const modified = async (argv) => {
   let changedPackages = argv.integration ?
-    await updated({dependents: true, npm: true}) :
-    await updated({npm: true});
+    await updated({dependents: true, npm: !!process.env.CI}) :
```
adamweeks: This should be the only place that checks the `CI` env var

tooling/commands/ci.js
taymoork2: This file actually doesn't run anymore since we moved to a bash based loop

tooling/lib/updated.js
```diff
@@ -18,7 +18,7 @@ const {diff} = require('./git');
  * npm instead of upstream/master
  * @returns {Promise<Array<string>>}
  */
-exports.updated = async function updated({dependents, npm}) {
+exports.updated = async function updated({dependents, npm = !!process.env.CI}) {
```
taymoork2: I'm ok with making this empty instead of defaulted false if there is no value
I just added it cuz it seemed "appropriate"

PR #1447, opened 2019-12-09
WIP : feat(internal-plugin-encryption): add support for multi-cluster file download

Akashdeep-Sarin:
## Description

This PR adds support for multi-cluster file download. Before this PR, files are downloaded from the files-service tied to the `internal.device._values -> serviceHostMap`.  
In some use cases (i.e. federation) the file location may be stored on different file services. The PR points the file service to the one received in the endpoint URL.

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
+++ b/packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
@@ -3,6 +3,7 @@
  */
 
 import {EventEmitter} from 'events';
+import url from 'url';
 
 import {WebexPlugin} from '@webex/webex-core';
 import {proxyEvents, tap, transferEvents} from '@webex/common';
@@ -120,11 +121,15 @@ const Encryption = WebexPlugin.extend({
     const inputBody = {
       endpoints: [scr.loc]
     };
+    const endpointUrl = url.parse(scr.loc);
+
+    // hardcode the url to use 'https' and the file service '/v1/download/endpoints' api
+    endpointUrl.protocol = 'https';
+    endpointUrl.pathname = '/v1/download/endpoints';
 
     return this.request({
       method: 'POST',
-      service: 'files',
-      resource: 'download/endpoints',
+      uri: url.format(endpointUrl),
       body: options ? {
         ...inputBody,
         allow: options.params.allow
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -120,11 +121,12 @@ const Encryption = WebexPlugin.extend({
     const inputBody = {
       endpoints: [scr.loc]
     };
+    const endpointUrl = url.parse(scr.loc);
+    const serviceUrl = `${endpointUrl.protocol}//${endpointUrl.hostname}/v1/download/endpoints`;
```
andrewholsted: I don't think we need to declare the endpoint.protocol? Everything should be over https and I wouldn't want us trying to retrieve anything over http. Feels like there might be a better way to do this via the service links but maybe not. Will defer to @InteractiveTimmy. This seems like a simple solution if it works.

Akashdeep-Sarin: Done!

Akashdeep-Sarin: @andrewholsted  -  I removed the `protocol`, but leads to code failure. Seems like protocol is mandatory for `uri`. So, I am reverting it.
@InteractiveTimmy - any inputs.. ?

lalli-flores: @Akashdeep-Sarin, if as @andrewholsted states that we don't want to retrieve anything over HTTP, you can hardcode the protocol to HTTPS, so it would only look for resources over HTTPS

InteractiveTimmy: If the hostname is not always from the home cluster [i.e. it is possible to get an EU hostname when you're on US clusters in this use case], then this may be the best way to perform these actions. the URL dependency can be used to make this process a bit easier and clearer.

```js
endpointUrl.host = ...;
endpointUrl.protocol  = ...;
endpoint.path = ...;
endpointUrl = url.format(endpointUrl);
```

See Node's documentation on URL [here](https://nodejs.org/api/url.html#url_url_strings_and_url_objects)

lalli-flores:
@Akashdeep-Sarin, can you add a unit tests that check that the `serviceURL` is properly formed? The CI is currently failing because there are no tests in this plug-in

lalli-flores:
@Akashdeep-Sarin , I re-ran the CI but integration tests are still failing

Akashdeep-Sarin:
@lalli-flores , I have rebased and pushed... Its passing now... 👍

PR #1452, opened 2019-12-16
feat(ediscovery): add new APIs for content related requests to remote clusters

decryan1984:
## Description

Adding three new ediscovery request APIs (getRemoteContent, getRemoteContentSummary and getRemoteContentSummaryBySpaceId) that take report urls as parameters rather than report ids. These will allow us to choose which ediscovery service in which federated environment we request data from.

Fixes # (issue)

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-107157

## Type of Change

## Test Coverage

All existing unit tests were run, 26 new tests added to cover the new APIs. 15 new integration tests have been added, 9 to cover existing API functionality, 6 to cover the new url option behaviour.

**Test Configuration**:
* SDK Version: latest
* Node/Browser Version 10.17
* NPM Version 6.11.3

## Checklist:


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -318,6 +437,24 @@ const EDiscovery = WebexPlugin.extend({
...
+      const uuids = reportUrl.match(/[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}/);
+
+      if (uuids && uuids.length > 0) {
+        return uuids[0];
+      }
+    }
+
+    return '';
+  },
+
```
decryan1984: New function used by the three new APIs to allow them to extract the report id from the url parameter

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -211,6 +211,40 @@ const EDiscovery = WebexPlugin.extend({
...
+    options = {...this.config.defaultOptions, ...options};
+
+    return this.request({
+      method: 'GET',
+      url: `${reportUrl}/contents`,
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    });
+  },
+
```
decryan1984: To work with ediscovery reports with content in multiple federated environments we may need to make requests outside the home cluster to retrieve that content, these APIs allow us to achieve that by setting the url property on teh request parameters, rather than the service and resource we previously used.

packages/node_modules/@webex/internal-plugin-ediscovery/test/integration/spec/ediscovery.js
```diff
@@ -1,5 +1,211 @@
...
+        await complianceUser.webex.internal.ediscovery.getRemoteContentSummaryBySpaceId(invalidUrl, spaceId)
+          .then(() => {
+            assert.fail('Expected error response due to invalid url');
+          })
+          .catch((error) => {
+            assert.equal(error.statusCode, 0);
+          });
+      });
+    });
+  });
```
decryan1984: Most of these tests check for 404s from ediscovery, with specific messages that show they have hit the service but there was no report/content available. Because of several issue around report creation and expiry, user validation against CI and the asynchronous way in which report content is retrieved from the service, it is difficult to test happy path scenarios without creating unstable tests.

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -357,4 +549,145 @@ describe('EDiscovery Content API Tests', () => {
...
+        .then((res) => {
+          expect(res).to.equal(mockResponse);
+          assert.callCount(webex.internal.ediscovery.contentSummaryBySpaceIdCache.has, 2);
+          assert.callCount(spaceIdMap.has, 1);
+          assert.notCalled(spaceIdMap.set);
+        });
+
+      return result;
+    });
+  });
```
decryan1984: Most of these tests are modified versions of the existing API tests, with the exception of the 'fails with a url that does not contain a report id' case

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -244,6 +278,47 @@ const EDiscovery = WebexPlugin.extend({
...
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    })
+      .then(async (res) => {
+        this.logger.error('report id is ' + reportId);
+
+        await this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body);
+
+        return Promise.resolve(res);
+      });
```
mwegman: You could rewrite this as:
``` javascript
.then(tap((res) => {
  this.logger.error('report id is ' + reportId);

  return this._writeSpacesToContentSummaryBySpaceIdCache(reportId, res.body);
}));
```
Make sure `tap` is imported from `@webex/common`

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -282,6 +357,50 @@ const EDiscovery = WebexPlugin.extend({
...
+    return this.request({
+      method: 'GET',
+      url: `${reportUrl}/contents/summary/${spaceId}`,
+      timeout: options.timeoutMs
+    })
+      .then(async (res) => {
+        await this._writeSpacesToContentSummaryBySpaceIdCache(reportId, [res.body]);
+
+        return Promise.resolve(res);
+      });
```
mwegman: Could write as:
``` javascript
.then(tap((res) => this._writeSpacesToContentSummaryBySpaceIdCache(reportId, [res.body])));
```

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -282,6 +355,46 @@ const EDiscovery = WebexPlugin.extend({
...
+    }
+
+    const reportId = this._getReportIdFromUrl(reportUrl);
+
+    if (!reportId) {
+      throw Error('Report url does not contain a report id');
+    }
+
+    // use spread operator to set default options
+    options = {...this.config.defaultOptions, ...options};
```
InteractiveTimmy: It seems like this block of code is repeated between the appended functions, While it's not required, would it be possible to make a helper function that handles the "setup" part of these methods? Possibly wrap the `reportUrl` and `options` values into a single method that can be called by the others?

decryan1984: It would be nice to remove some of the duplication Tim, but I'd be concerned it would look inconsistent if I changed it for those 5 functions that this block of code appears in and leaving the other 8 as they are. There are a variety of input params across the functions here, I'd rather not create too many helper functions which might end up making it harder to read.

PR #1413, opened 2019-11-14
Creation of "devices" plugin to replace WDM service methods

InteractiveTimmy:
# Deprecate WDM Service Methods

## Description

This is a draft pull request for the new Devices plugin that will deprecate the original WDM plugin and its service methods. This is being presented as a draft pull request to allow for gated validation points along the way. Ultimately, this added plugin could either stay unique, leaving legacy support for WDM, or, replace WDM and it's a namespace.

Fixes # spark-102183

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Test Coverage

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Generated a testing suite [TODO]

**Test Configuration**:
* Node/Browser Version v8.16.0
* NPM Version v6.4.1

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1462, opened 2019-12-20
Feat/devices generate registration methods

InteractiveTimmy:
# Pull Request

## Description

This is a rework of the [draft pull request](https://github.com/webex/webex-js-sdk/pull/1413) to rework `internal-plugin-wdm`.

This pull request includes all of the registration methods for the devices plugin as well as their associated tests.

Fixes # (issue)

## Type of Change

## Test Coverage

- [x] Generated new tests and tested the package's suite.

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-devices/src/devices.js
+++ b/packages/node_modules/@webex/internal-plugin-devices/src/devices.js
@@ -331,9 +331,135 @@ const Devices = WebexPlugin.extend({
 
   // Registration method members
 
-  refresh() {}, // Refreshes the device's registration.
-  register() {}, // Registers a device.
-  unregister() {}, // Unregisters the device.
+  /**
+   * Refresh the current registered device if able.
+   *
+   * @returns {Promise<void, Error>}
+   */
+  refresh() {
+    this.logger.info('devices: refreshing');
+
+    // Validate that the device can be registered.
+    return this.canRegister(true)
+      .then(() => {
+        // Validate if the device is not registered and register instead.
+        if (!this.registered) {
+          this.logger.info('devices: device not registered, registering');
+
+          return this.register();
+        }
+
+        // Merge body configurations, overriding defaults.
+        const body = {
+          ...(this.serialize()),
+          ...(this.config.body ? this.config.body : {})
+        };
+
+        // Remove unneeded properties from the body object.
+        delete body.features;
+        delete body.mediaCluster;
+
+        // Append a ttl value if the device is marked as ephemeral.
+        if (this.config.ephemeral) {
+          body.ttl = this.config.ephemeralDeviceTTL;
+        }
+
+        // Merge header configurations, overriding defaults.
+        const headers = {
+          ...(this.config.defaults.headers ? this.config.defaults.headers : {}),
+          ...(this.config.headers ? this.config.headers : {})
+        };
+
+        return this.request({
+          method: 'PUT',
+          uri: this.url,
+          body,
+          headers
+        })
+          .then((response) => this.processRegistrationSuccess(response))
+          .catch((reason) => {
+            // Handle a 404 error, which indicates that the device is no longer
+            // valid and needs to be registered as a new device.
+            if (reason.statusCode === 404) {
+              this.logger.info('devices: refresh failed, device is not valid');
+              this.logger.info('devices: attempting to register a new device');
+
+              this.clear();
+
+              return this.register();
+            }
+
+            return Promise.reject(reason);
+          });
+      });
+  },
+
+  /**
+   * Register or refresh a device depending on the current device state. Device
+   * registration utilizes the services plugin to send the request to the
+   * **WDM** service.
+   *
+   * @returns {Promise<void, Error>}
+   */
+  register() {
+    this.logger.info('devices: registering');
+
+    // Validate that the device can be registered.
+    return this.canRegister(true)
+      .then(() => {
+        // Validate if the device is already registered and refresh instead.
+        if (this.registered) {
+          this.logger.info('devices: device already registered, refreshing');
+
+          return this.refresh();
+        }
+
+        // Merge body configurations, overriding defaults.
+        const body = {
+          ...(this.config.defaults.body ? this.config.defaults.body : {}),
+          ...(this.config.body ? this.config.body : {})
+        };
+
+        // Merge header configurations, overriding defaults.
+        const headers = {
+          ...(this.config.defaults.headers ? this.config.defaults.headers : {}),
+          ...(this.config.headers ? this.config.headers : {})
+        };
+
+        // This will be replaced by a `create()` method.
+        return this.request({
+          method: 'POST',
+          service: 'wdm',
+          resource: 'devices',
+          body,
+          headers
+        })
+          .then((response) => this.processRegistrationSuccess(response));
+      });
+  },
+
+  /**
+   * Unregister the current registered device if available. Unregistering a
+   * device utilizes the services plugin to send the request to the **WDM**
+   * service.
+   *
+   * @returns {Promise<void, Error>}
+   */
+  unregister() {
+    this.logger.info('devices: unregistering');
+
+    if (!this.registered) {
+      this.logger.warn('devices: not registered');
+
+      return Promise.resolve();
+    }
+
+    return this.request({
+      uri: this.url,
+      method: 'DELETE'
+    })
+      .then(() => this.clear());
+  },
 
   // Helper method members
 
@@ -439,7 +565,41 @@ const Devices = WebexPlugin.extend({
     Reflect.apply(WebexPlugin.prototype.clear, this, args);
   },
 
-  getWebSocketUrl() {}, // Retrieves the priority-mapped web socket url.
+  /**
+   * Get the current websocket url with the appropriate priority host.
+   *
+   * @param {boolean} [wait=false] - Willing to wait on a valid url.
+   * @returns {Promise<string, Error>} - The priority-mapped web socket url.
+   */
+  getWebSocketUrl(wait = false) {
+    this.logger.info('devices: getting the current websocket url');
+
+    // Destructure the services plugin for ease of reference.
+    const {services} = this.webex.internal;
+
+    // Validate if the method should wait for registration.
+    if (wait) {
+      return this.waitForRegistration()
+        .then(() => services.convertUrlToPriorityHostUrl(this.webSocketUrl))
+        .catch((error) => {
+          this.logger.warn(error.message);
+
+          return Promise.reject(new Error(
+            'devices: failed to get the current websocket url'
+          ));
+        });
+    }
+
+    // Validate if the device is registered.
+    if (!this.registered) {
+      return Promise.reject(new Error(
+        'devices: cannot get websocket url, device is not registered'
+      ));
+    }
+
+    return Promise.resolve()
+      .then(() => services.convertUrlToPriorityHostUrl(this.webSocketUrl));
+  },
 
   /**
    * Process a successful device registration.
@@ -450,8 +610,15 @@ const Devices = WebexPlugin.extend({
   processRegistrationSuccess(response) {
     this.logger.info('devices: received registration payload');
 
+    // Clone the response body for service cleaning.
+    const body = {...response.body};
+
+    // Clean service data.
+    delete body.services;
+    delete body.serviceHostMap;
+
     // Assign the recieved DTO from **WDM** to this device.
-    this.set(response.body);
+    this.set(body);
 
     // Validate if device is ephemeral and setup refresh timer.
     if (this.config.ephemeral) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/internal-plugin-devices/src/devices.js
```diff
@@ -450,8 +624,15 @@ const Devices = WebexPlugin.extend({
...
+    // Clone the response body for service cleaning.
+    const body = {...response.body};
+
+    // Clean service data.
+    delete body.services;
+    delete body.serviceHostMap;
+
     // Assign the recieved DTO from **WDM** to this device.
-    this.set(response.body);
+    this.set(body);
```
InteractiveTimmy: This section was added to properly remove the service data from the `WDM` DTO

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -30,9 +30,9 @@ describe('plugin-devices', () => {
       devices = webex.internal.devices;
     });
 
-    // This will be implemented in the same pull request that `unregister()` is
-    // implemented.
-    afterEach('attempt to unregister the device', () => {});
+    // Do nothing on error, as some methods utilize sinon for `unregister()`.
+    afterEach('attempt to unregister the device', () => devices.unregister()
+      .catch(() => {}));
```
InteractiveTimmy: In some cases, when overriding `unregister()` in a test, it makes this method return a rejected promise. This will attempt to unregister and move on if it fails.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
```diff
@@ -331,9 +331,149 @@ const Devices = WebexPlugin.extend({
...
+
+              this.clear();
+
+              return this.register();
+            }
+
+            return Promise.reject(reason);
+          });
+      })
+      .catch((error) => {
```
adamweeks: Would this catch the promise rejection on line 392 for the 404 error? If so, the error message isn't correct.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
```diff
@@ -331,9 +331,149 @@ const Devices = WebexPlugin.extend({
...
+        return this.request({
+          method: 'POST',
+          service: 'wdm',
+          resource: 'devices',
+          body,
+          headers
+        })
+          .then((response) => this.processRegistrationSuccess(response));
+      })
+      .catch((error) => {
```
adamweeks: This too might catch the request error from 437 and print an incorrect error message.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
```diff
@@ -519,7 +700,30 @@ const Devices = WebexPlugin.extend({
     }, duration * 1000);
   },
 
-  waitForRegistration() {}, // Eventing method queue actions after registration.
+  /**
+   * Wait for the device to be registered.
+   *
+   * @param {number} [timeout] - The maximum duration to wait, in seconds.
```
adamweeks:
```suggestion
   * @param {number} [timeout=10] - The maximum duration to wait, in seconds.
```

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
```diff
@@ -519,7 +700,30 @@ const Devices = WebexPlugin.extend({
...
+    return new Promise((resolve, reject) => {
+      if (this.registered) {
+        resolve();
+      }
+
+      const timeoutTimer = safeSetTimeout(() => reject(
+        new Error('devices: timeout occured while waiting for registration')
+      ), timeout * 1000);
+
+      this.once('registration:success', () => {
```
adamweeks: Is this a constant we can use?

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -342,7 +342,104 @@ describe('plugin-devices', () => {
...
+                assert.include(url, 'mercury');
+              }));
+        });
+
+        describe('when the device is not registered', () => {
+          describe('when the device successfully registers', () => {
+            beforeEach(() => {
+              // This needs to not be returned, as it must be pending during the
+              // it clause.
+              devices.register();
```
adamweeks: This seems like it could cause flakiness depending on speed of the test runner. I'd like to see this in the `it` block as maybe part of a `Promise.all` instead.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -342,7 +342,104 @@ describe('plugin-devices', () => {
...
+              return devices.getWebSocketUrl(wait)
+                .then((url) => {
+                  assert.isDefined(url);
+                  assert.isTrue(services.isServiceUrl(url));
+                  assert.include(url, 'mercury');
+                });
+            });
+          });
+
+          describe('when the device never registers', () => {
```
adamweeks: This describe block seems extraneous. "when the device is not registered, it should return a rejected promise" works.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
adamweeks: I'd prefer to see the `afterEach` block of the test that is modifying it revert the change instead of this generic error state handling.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -342,7 +342,104 @@ describe('plugin-devices', () => {
...
+        beforeEach('set wait', () => {
+          wait = false;
+        });
+
+        describe('when the device is registered', () => {
+          beforeEach('register the device', () => devices.register());
+
+          describe('when the priority host can be mapped', () => {
+            it('should resolve the promise with the websocket url', () => {
+              assert.isDefined(services.get('wdm'));
```
adamweeks: There are a few of these (what I like to call) "setup assertions" throughout. They are good to help when writing tests, but might fit better in the `beforeEach` block so they don't confuse the actual test assertions.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -412,8 +521,341 @@ describe('plugin-devices', () => {
...
+
+              return devices.refresh()
+                .then(() =>
+                  assert.calledWith(devices.request, sinon.match.has(
+                    'uri', devices.url
+                  )));
+            });
+
+            it('should send a PUT request', () => {
+              assert.isTrue(devices.registered);
```
adamweeks: These tests from within this describe block are those "setup assertions" that are confusing to read.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -412,8 +521,341 @@ describe('plugin-devices', () => {
...
+
+            it('should return a resolved promise',
+              () => assert.isFulfilled(devices.refresh()));
+
+            it('should call \'processRegistrationSuccess()\'', () => devices.refresh()
+              .then(() => assert.called(devices.processRegistrationSuccess)));
+          });
+
+          describe('when the device fails to refresh', () => {
+            beforeEach('setup \'register()\' stub', () => {
```
adamweeks: Is this stub only used in the `should attempt to register a new device` test from line 668? If so, it probably can be moved to within the `it` block.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -412,8 +521,341 @@ describe('plugin-devices', () => {
...
+
+              return devices.register()
+                .then(() =>
+                  assert.calledWith(devices.request, sinon.match.has(
+                    'method', 'POST'
+                  )));
+            });
+          });
+
+          describe('when the device is successfully registers', () => {
```
adamweeks:
```suggestion
          describe('when the device is successfully registered', () => {
```

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -472,7 +914,97 @@ describe('plugin-devices', () => {
...
+            return devices.unregister()
+              .then(() =>
+                assert.calledWith(devices.request, sinon.match.has(
+                  'method', 'DELETE'
+                )));
+          });
+        });
+
+        describe('when the device unregistration request is successful', () => {
+          it('should clear the device properties', () => devices.unregister()
```
adamweeks:
```suggestion
          it('should clear the device url', () => devices.unregister()
```

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: I believe that is what it currently does.
```js
describe('when the device never registers', () => {
  it('should return a rejected promise', () => {
    assert.isFalse(devices.registered); // assert that it never registered
    assert.isRejected(devices.getWebSocketUrl(wait)); // assert that it rejected
  });
});
```
Let me know if you have an example or more info on how I could improve this.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -472,7 +914,97 @@ describe('plugin-devices', () => {
...
+        beforeEach('register the device', () => devices.register());
+
+        it('should return a resolved promise',
+          () => assert.isFulfilled(devices.waitForRegistration()));
+      });
+
+      describe('when the device is not registered', () => {
+        describe('when the device registers', () => {
+          beforeEach('trigger a pending device registration', () => {
+            devices.register();
```
adamweeks: This is another potential race condition. Let's move it into the it and Promise.all with `waitForRegistration`. Something like:

```js
return Promise.all([devices.register(), devices.waitForRegistration()])
  .then(() => assert.isTrue(devices.registered));
```

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
adamweeks: You can leave the `it` but remove the entire describe block of `describe('when the device never registers', () => {`. The surrounding describe block is sufficient.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Implementing this.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
InteractiveTimmy: This can be fixed by simply removing the catch block to follow the `this.canRegister()` method. This will simplify the registration method's logic as well.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
InteractiveTimmy: I removed this catch block, see the message from above similar comment.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
InteractiveTimmy: Adding this JSDoc comment.

packages/node_modules/@webex/internal-plugin-devices/src/devices.js
InteractiveTimmy: Yes, adding now.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: I can see how this could cause flakiness. I have moved all unmanaged promises to their appropriate sections.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Reorganized these statements to assert during the beforeEach() methods to properly organize them into the "setup" stage.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Moved these like the before comment.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Moved it to the specific `it` statement it's exclusive to.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Good catch, changed.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Changed this to match the scope of the `it()` method.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: Moved these to a `Promise.all()`.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
InteractiveTimmy: I see, made this change.

packages/node_modules/@webex/internal-plugin-devices/test/integration/spec/devices.js
```diff
@@ -17,24 +17,21 @@ describe('plugin-devices', () => {
...
-    // This will be implemented in the same pull request that `unregister()` is
-    // implemented.
-    afterEach('attempt to unregister the device', () => {});
+          devices = webex.internal.devices;
+        }));
 
-    after('remove users', () => testUsers.remove([user]));
+    afterEach('unregister the device and remove test users',
+      () => devices.unregister()
+        .then(() => testUsers.remove([user])));
```
InteractiveTimmy: There is a bug when using a `before()` clause with creating users, see SPARK-112453

PR #1466, opened 2020-01-15
feat(calendar): handle /calendar API not returning meeting notes and participants

prsaroha:
## Description

Calendar will soon stop sending encrypted notes and participant lists that exceed 20k length from the currently used /calendar API. They'll instead be `null`, and 2 new APIs have been provided to fetch notes and participants with meeting id, which will have to be used in those cases.

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-100549

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
+++ b/packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
@@ -42,6 +42,7 @@
  * @memberof Calendar
  */
 
+import btoa from 'btoa';
 import {WebexPlugin} from '@webex/webex-core';
 
 import CalendarCollection from './collection';
@@ -237,6 +238,32 @@ const Calendar = WebexPlugin.extend({
   },
 
   /**
+   * Retrieves an array of meeting participants for the meeting id
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting participants
+   */
+  getParticipants(id) {
+    return this.request({
+      method: 'GET',
+      service: 'calendar',
+      resource: `calendarEvents/${btoa(id)}/participants`
+    });
+  },
+
+  /**
+   * Retrieves a collection of meetings based on the request parameters
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting notes
+   */
+  getNotes(id) {
+    return this.request({
+      method: 'GET',
+      service: 'calendar',
+      resource: `calendarEvents/${btoa(id)}/notes`
+    });
+  },
+
+  /**
    * Retrieves a collection of meetings based on the request parameters
    * @param {Object} options
    * @param {Date} options.fromDate the start of the time range
@@ -252,7 +279,33 @@ const Calendar = WebexPlugin.extend({
       resource: 'calendarEvents',
       qs: options
     })
-      .then((res) => res.body.items);
+      .then((res) => {
+        const meetingObjects = res.body.items;
+        const promises = [];
+
+        meetingObjects.forEach((meeting) => {
+          if (!meeting.encryptedNotes) {
+            promises.push(
+              this.getNotes(meeting.id)
+                .then((notesResponse) => {
+                  meeting.encryptedNotes = notesResponse.body.encryptedNotes;
+                })
+            );
+          }
+
+          if (!meeting.encryptedParticipants) {
+            promises.push(
+              this.getParticipants(meeting.id)
+                .then((notesResponse) => {
+                  meeting.encryptedParticipants = notesResponse.body.encryptedParticipants;
+                })
+            );
+          }
+        });
+
+        return Promise.all(promises)
+          .then(() => meetingObjects);
+      });
   }
 });
 
--- a/packages/node_modules/@webex/internal-plugin-calendar/src/index.js
+++ b/packages/node_modules/@webex/internal-plugin-calendar/src/index.js
@@ -17,6 +17,26 @@ registerInternalPlugin('calendar', Calendar, {
   payloadTransformer: {
     predicates: [
       {
+        name: 'transformMeetingNotes',
+        direction: 'inbound',
+        test(ctx, response) {
+          return Promise.resolve(has(response, 'body.encryptedNotes'));
+        },
+        extract(response) {
+          return Promise.resolve(response.body);
+        }
+      },
+      {
+        name: 'transformMeetingParticipants',
+        direction: 'inbound',
+        test(ctx, response) {
+          return Promise.resolve(has(response, 'body.encryptedParticipants'));
+        },
+        extract(response) {
+          return Promise.resolve(response.body);
+        }
+      },
+      {
         name: 'transformMeetingArray',
         direction: 'inbound',
         test(ctx, response) {
@@ -66,11 +86,11 @@ registerInternalPlugin('calendar', Calendar, {
             return Promise.resolve();
           }
 
-          // Decrypt participant properties
-          const decryptedParticipants = object.encryptedParticipants.map((participant) => Promise.all([
+          // Decrypt participant properties if meeting object contains participants
+          const decryptedParticipants = object.encryptedParticipants ? object.encryptedParticipants.map((participant) => Promise.all([
             ctx.transform('decryptTextProp', 'encryptedEmailAddress', object.encryptionKeyUrl, participant),
             ctx.transform('decryptTextProp', 'encryptedName', object.encryptionKeyUrl, participant)
-          ]));
+          ])) : [];
 
           return Promise.all([
             ctx.transform('decryptTextProp', 'encryptedSubject', object.encryptionKeyUrl, object),
@@ -82,6 +102,44 @@ registerInternalPlugin('calendar', Calendar, {
             ctx.transform('decryptTextProp', 'spaceURI', object.encryptionKeyUrl, object),
             ctx.transform('decryptTextProp', 'spaceURL', object.encryptionKeyUrl, object)
           ].concat(decryptedParticipants));
+        }
+      },
+      {
+        name: 'transformMeetingNotes',
+        direction: 'inbound',
+        fn(ctx, object) {
+          if (!object) {
+            return Promise.resolve();
+          }
+
+          if (!object.encryptionKeyUrl) {
+            return Promise.resolve();
+          }
+
+          return Promise.all([
+            ctx.transform('decryptTextProp', 'encryptedNotes', object.encryptionKeyUrl, object)
+          ]);
+        }
+      },
+      {
+        name: 'transformMeetingParticipants',
+        direction: 'inbound',
+        fn(ctx, object) {
+          if (!object) {
+            return Promise.resolve();
+          }
+
+          if (!object.encryptionKeyUrl || !object.encryptedParticipants) {
+            return Promise.resolve();
+          }
+
+          // Decrypt participant properties
+          const decryptedParticipants = object.encryptedParticipants.map((participant) => Promise.all([
+            ctx.transform('decryptTextProp', 'encryptedEmailAddress', object.encryptionKeyUrl, participant),
+            ctx.transform('decryptTextProp', 'encryptedName', object.encryptionKeyUrl, participant)
+          ]));
+
+          return Promise.all(decryptedParticipants);
         }
       }
     ]
--- a/packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
+++ b/packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
@@ -65,6 +65,67 @@ function makeMockMeetingPayload(params) {
   return payload;
 }
 
+/**
+ * produces a mock notes object
+ * @param {Object} params
+ * @returns {Notes}
+ */
+function makeMockNotesPayload(params) {
+  const payload = {
+    id: params.meetingId,
+    seriesId: params.seriesId,
+    links: [
+      {
+        href: `https://calendar-example.com/calendar/api/v1/${params.locusId}`,
+        rel: 'self'
+      }
+    ],
+    encryptedNotes: params.meetingNotes
+  };
+
+  return payload;
+}
+
+/**
+ * produces a mock participants object
+ * @param {Object} params
+ * @returns {Participants}
+ */
+function makeMockParticipantsPayload(params) {
+  const payload = {
+    id: params.meetingId,
+    seriesId: params.seriesId,
+    links: [
+      {
+        href: `https://calendar-example.com/calendar/api/v1/${params.locusId}`,
+        rel: 'self'
+      }
+    ],
+    encryptedParticipants: [
+      {
+        id: params.participants.participant1.id,
+        type: 'PERSON',
+        responseType: 'NO_RESPONSE',
+        participantType: 'OPTIONAL',
+        orgId: params.participants.participant1.orgId,
+        encryptedEmailAddress: params.participants.participant1.emailAddress,
+        encryptedName: params.participants.participant1.name
+      },
+      {
+        id: params.participants.participant2.id,
+        type: 'PERSON',
+        responseType: 'UNKNOWN_RESPONSE',
+        participantType: 'REQUIRED',
+        orgId: params.participants.participant2.orgId,
+        encryptedEmailAddress: params.participants.participant2.emailAddress,
+        encryptedName: params.participants.participant2.name
+      }
+    ]
+  };
+
+  return payload;
+}
+
 function postToWhistler(webex, type, payload) {
   return retry(() => webex.request({
     method: 'POST',
@@ -87,6 +148,32 @@ function postToWhistler(webex, type, payload) {
 
       return res;
     }));
+}
+
+function postNotesToWhistler(webex, type, payload) {
+  return retry(() => webex.request({
+    method: 'POST',
+    uri: `${process.env.WHISTLER_API_SERVICE_URL}/calendarEvent`,
+    body: payload,
+    qs: {
+      changeType: type,
+      useProduction: typeof process.env.WDM_SERVICE_URL === 'undefined' || process.env.WDM_SERVICE_URL.includes('wdm-a.wbx2.com')
+    }
+  })
+    .then((res) => res.body));
+}
+
+function postParticipantsToWhistler(webex, type, payload) {
+  return retry(() => webex.request({
+    method: 'POST',
+    uri: `${process.env.WHISTLER_API_SERVICE_URL}/calendarEvent`,
+    body: payload,
+    qs: {
+      changeType: type,
+      useProduction: typeof process.env.WDM_SERVICE_URL === 'undefined' || process.env.WDM_SERVICE_URL.includes('wdm-a.wbx2.com')
+    }
+  })
+    .then((res) => res.body));
 }
 
 describe.skip('plugin-calendar', () => {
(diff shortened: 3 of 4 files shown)


packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
```diff
@@ -34,6 +37,32 @@ const Calendar = WebexPlugin.extend({
...
+   * Retrieves a collection of meetings based on the request parameters
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting notes
+   */
+  getNotes(id) {
+    return window.squared.request({
+      method: 'GET',
+      service: 'calendar',
+      resource: `calendarEvents/${btoa(id)}/notes`
+    }).then((res) => res);
```
InteractiveTimmy: `window.squared.request()` does not exist in a Webex instance unless it is mounted this way. Please use `this.request()` to send requests, as that is where the request method is locally mounted by the extended `WebexPlugin` class.

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
```diff
@@ -34,6 +37,32 @@ const Calendar = WebexPlugin.extend({
...
+   * Retrieves an array of meeting participants for the meeting id
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting participants
+   */
+  getParticipants(id) {
+    return window.squared.request({
+      method: 'GET',
+      service: 'calendar',
+      resource: `calendarEvents/${btoa(id)}/participants`
+    }).then((res) => res);
```
InteractiveTimmy: This `.then((res) => res);` is redundant in this case, please remove this `.then()` statement.

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
InteractiveTimmy: This `.then((res) => res);` is redundant in this case, please remove this `.then()` statement.

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
```diff
@@ -65,22 +65,40 @@ registerInternalPlugin('calendar', Calendar, {
...
-          ]));
+          // Check if notes and participants were sent with meeting object, and fetch if not
+          if (!object.encryptedNotes) {
+            object.encryptedNotes = Calendar.prototype.getNotes(object.id);
+          }
+          if (!object.encryptedParticipants) {
+            object.encryptedParticipants = Calendar.prototype.getParticipants(object.id);
+          }
+
+          return Promise.all([object.encryptedNotes, object.encryptedParticipants])
```
InteractiveTimmy: Is it possible for either of these promises to reject? No change is needed if they cannot be rejected, otherwise, we should discuss what needs to occur in case of rejection.

prsaroha: Only scenario I can think of is if a network error occurs and the Promise stays pending. Otherwise, since we're using a meeting id that we just received, these shouldn't reject.

prsaroha:
@InteractiveTimmy I made the request changes and updated this PR, except for adding the tests for it. I tried testing locally but I get timeout errors running unit tests after adding my client id and secret to the env file. I can't see where my unit tests failed in Circle CI either. Also, I only updated internal-plugin-calendar, and the test file for that has all the test cases inside a describe.skip() so I'm not sure if that's where my PR would fail.
image

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
```diff
@@ -34,6 +37,32 @@ const Calendar = WebexPlugin.extend({
...
     })
       .then((res) => res.body.items);
+  },
+
+  /**
+   * Retrieves an array of meeting participants for the meeting id
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting participants
+   */
+  getParticipants(id) {
```
InteractiveTimmy: This method needs tests to validate the implementation works as expected.

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
```diff
@@ -34,6 +37,32 @@ const Calendar = WebexPlugin.extend({
...
+      resource: `calendarEvents/${btoa(id)}/participants`
+    });
+  },
+
+  /**
+   * Retrieves a collection of meetings based on the request parameters
+   * @param {String} id
+   * @returns {Promise} Resolves with an object of meeting notes
+   */
+  getNotes(id) {
```
InteractiveTimmy: This method needs tests to validate the implementation works as expected.

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
```diff
@@ -65,22 +65,40 @@ registerInternalPlugin('calendar', Calendar, {
             return Promise.resolve();
           }
 
-          // Decrypt participant properties
-          const decryptedParticipants = object.encryptedParticipants.map((participant) => Promise.all([
-            ctx.transform('decryptTextProp', 'encryptedEmailAddress', object.encryptionKeyUrl, participant),
-            ctx.transform('decryptTextProp', 'encryptedName', object.encryptionKeyUrl, participant)
-          ]));
+          // Check if notes and participants were sent with meeting object, and fetch if not
```
InteractiveTimmy: These changes may require modification/refactor of the tests to make sure it works as expected and does not fail during CI process. Please make changes to the testing files.

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
```diff
@@ -65,22 +65,40 @@ registerInternalPlugin('calendar', Calendar, {
...
+          if (!object.encryptedNotes) {
+            object.encryptedNotes = Calendar.prototype.getNotes(object.id);
+          }
+          if (!object.encryptedParticipants) {
+            object.encryptedParticipants = Calendar.prototype.getParticipants(object.id);
+          }
```
mwegman: Don't fetch new data in a transform. Instead you can have the `list` command fetch participants and notes if they aren't available or make the user do a separate fetch. Add a new transform for the participant fetch and the notes fetch.

prsaroha: I wrote a transform each for notes and participants, and moved the work to `.then()` of `list` fetch.

Unit tests failed but I can't see where/why?
image

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -90,9 +90,53 @@ function postToWhistler(webex, type, payload) {
...
+      assert.equal(createdMeeting.meeting.itemId, payload.id);
+
+      return res;
+    }));
+}
+
+describe('plugin-calendar', () => {
   describe('Calendar', () => {
-    describe('#list()', function () {
+    describe.skip('#list()', function () {
```
mwegman: Why are you skipping these tests?

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -90,9 +90,53 @@ function postToWhistler(webex, type, payload) {
...
+    method: 'POST',
+    service: 'whistler',
+    resource: 'calendarEvent',
+    body: payload,
+    qs: {
+      changeType: type,
+      useProduction: typeof process.env.WDM_SERVICE_URL === 'undefined' || process.env.WDM_SERVICE_URL.includes('wdm-a.wbx2.com')
+    }
+  })
+    .then((res) => {
```
mwegman: Minor note:
```suggestion
    .then(({body: createdMeeting}) => {
```
this would replace the next line as well

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -90,9 +90,53 @@ function postToWhistler(webex, type, payload) {
...
```
mwegman: Same minor note as above

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
prsaroha: There was only 1 test in this file earlier, and it was inside a `describe.skip()`, so I moved the `.skip()` on the inside to that test, added my tests to the outer describe block and removed the `.skip()` from there (line 93)

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
prsaroha: I have to return `res` from the `then()`, so I'll have to make a change in the return statement as well

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
```diff
@@ -82,6 +103,44 @@ registerInternalPlugin('calendar', Calendar, {
...
+          if (!object) {
+            return Promise.resolve();
+          }
+
+          if (!object.encryptionKeyUrl) {
+            return Promise.resolve();
+          }
+
+          // Decrypt participant properties
+          const decryptedParticipants = object.encryptedParticipants.map((participant) => Promise.all([
```
adamweeks: This seems to be the same transform that is in `transformMeeting`, why the duplication?

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -90,9 +90,53 @@ function postToWhistler(webex, type, payload) {
...
+      // Test response is using our parameters
+      const createdMeeting = res.body;
+
+      assert.isDefined(createdMeeting);
```
adamweeks: Please move these assertions into a test instead of the helper function. A failed assertion here would be difficult to track down due to it being embedded in other tests.

For example, if the object returned didn't have the meeting notes, we would get a failure on the test `'retrieves the meeting notes for the given meetingId'` but the failure would be on the request, not the implementation.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -279,5 +323,112 @@ describe.skip('plugin-calendar', () => {
...
+            'squaredCallInitiation',
+            'squaredInviter',
+            'squaredRoomModeration',
+            'webExSquared',
+            'squaredFusionCal'
+          ]
+        }
+      })
+        .then((users) => {
+          console.log('Enter then');
```
adamweeks: These logs are helpful during development, but please remove them here.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -279,5 +323,112 @@ describe.skip('plugin-calendar', () => {
...
+          console.log('Got creator, spock, and mccoy');
+
+          webex = new WebexCore({
+            credentials: {
+              authorization: creator.token
+            },
+            config: {
+              services: {
+                discovery: {
+                  whistlerServiceUrl: process.env.WHISTLER_API_SERVICE_URL || 'https://calendar-whistler.onint.ciscospark.com:8084/api/v1'
```
adamweeks: Is this service not provided by the U2C catalog? Our tests could run on multiple environments.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -279,5 +323,112 @@ describe.skip('plugin-calendar', () => {
...
+                }
+              }
+            }
+          });
+          console.log('Got webexcore');
```
adamweeks: Please remove

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
prsaroha: Michael suggested to write new transforms for decrypting meeting notes and participants that are fetched separately, instead of fetching in the existing transform where it would decrypt via the current process.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
prsaroha: I was using the existing code in that file as reference, I'll move it elsewhere or remove it.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
prsaroha: Will do, sorry about that

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
prsaroha: I'm not very sure of how the whistler service is working, so I just used the existing code as reference when writing my own test cases. I'll check if the catalog can be used instead.

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
mwegman: @prsaroha said that the participants can come from the meeting request if they are less than a certain number. This transform would be for the participants request if it didn't show up in the original meeting request.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
mwegman: The first thing you do in the `then` statement following this is grab the `body` property and assign it to `createdMeeting`. You could skip that step.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -279,5 +323,112 @@ describe.skip('plugin-calendar', () => {
...
+
+        const meetingParams = {
+          meetingId: meetingID,
+          seriesId: seriesID,
+          locusId: locusID,
+          meetingNotes
+        };
+
+        return postNotesToWhistler(webex, 'CREATE', makeMockMeetingPayload(meetingParams))
+          .then((res) => {
```
mwegman: If you follow my suggestion above, this could be:
```suggestion
          .then((createMeeting) => {
```
and you can delete the next line

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -279,5 +323,112 @@ describe.skip('plugin-calendar', () => {
...
+          seriesId: seriesID,
+          locusId: locusID,
+          participants: {
+            participant1: mccoy,
+            participant2: spock
+          }
+        };
+
+        return postParticipantsToWhistler(webex, 'CREATE', makeMockMeetingPayload(meetingParams))
+          .then((res) => {
```
mwegman: Same as above

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
```diff
@@ -90,9 +90,53 @@ function postToWhistler(webex, type, payload) {
...
+    }
+  })
+    .then((res) => {
+      // Test response is using our parameters
+      const createdMeeting = res.body;
+
+      assert.isDefined(createdMeeting);
+      assert.equal(createdMeeting.meeting.meetingNotes, payload.meetingNotes);
+
+      return res;
```
mwegman:
```suggestion
      return createMeeting;
```

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
```diff
@@ -82,6 +103,44 @@ registerInternalPlugin('calendar', Calendar, {
...
+      },
+      {
+        name: 'transformMeetingParticipants',
+        direction: 'inbound',
+        fn(ctx, object) {
+          if (!object) {
+            return Promise.resolve();
+          }
+
+          if (!object.encryptionKeyUrl) {
```
adamweeks: Do we need to add a check for the `encryptedParticipants` object here? That way we won't get an error if it doesn't exist and we try to `.map` it on line 137.

```suggestion
          if (!object.encryptionKeyUrl || !object.encryptedParticipants) {
```

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
adamweeks: @prsaroha you are still missing any sort of test coverage for this function. I know that they are used in the `list` function, and the test for it covers them, but if they are public functions, we need tests calling it specifically like an sdk user would.

Another option is that if this and `getNotes` is meant to only be internally used by the `list` function, you could mark these as `@private` in the jsdoc so that the documentation generator will not print them.

packages/node_modules/@webex/internal-plugin-calendar/test/integration/spec/calendar.js
adamweeks: It doesn't look like whistler is a service provided by U2C, so we should be ok with hard coding endpoints, but wary that they could be the cause of flakiness.

packages/node_modules/@webex/internal-plugin-calendar/src/index.js
prsaroha: Will do

packages/node_modules/@webex/internal-plugin-calendar/src/calendar.js
prsaroha: I'll modify the tests I wrote to cover them, looks like I forgot to change them after making changes and writing new transforms

adamweeks: Sounds good. You'll need two new `describe` blocks in the integration suite: one for `describe('#getParticipants')` and one for `describe('#getNotes')`.

adamweeks:
As discussed in the space, since we have no way of integration testing these, we will just move test coverage to unit tests mocking the requests.

packages/node_modules/@webex/internal-plugin-calendar/test/unit/spec/calendar.js
```diff
@@ -0,0 +1,84 @@
...
+
+import {assert} from '@webex/test-helper-chai';
+import Calendar from '@webex/internal-plugin-calendar';
+import MockWebex from '@webex/test-helper-mock-webex';
+import btoa from 'btoa';
+import sinon from 'sinon';
+
+describe('internal-plugin-calendar', () => {
+  describe('Calendar Apis', () => {
+    const webex = new MockWebex({
```
adamweeks: Could you move this initialization into the `beforeEach` block so we're sure we get a fresh webex instance for each test?

```suggestion
    let webex;
```

packages/node_modules/@webex/internal-plugin-calendar/test/unit/spec/calendar.js
```diff
@@ -0,0 +1,84 @@
...
+
+          webex.request = sinon.stub().returns(
+            Promise.resolve({
+              body: {
+                encryptedNotes: 'notes1'
+              }
+            })
+          );
+
+          await webex.internal.calendar.getNotes(id)
```
adamweeks: Lets not mix async and promises:

```suggestion
          const res = await webex.internal.calendar.getNotes(id);
          assert.equal(res.body.encryptedNotes, 'notes1');
```

packages/node_modules/@webex/internal-plugin-calendar/test/unit/spec/calendar.js
```diff
@@ -0,0 +1,84 @@
...
+        it('should fetch the meeting participants', async () => {
+          webex.request = sinon.stub().returns(
+            Promise.resolve({
+              body: {
+                encryptedParticipants: ['participant1']
+              }
+            })
+          );
+
+          await webex.internal.calendar.getParticipants(id)
```
adamweeks:
```suggestion
          const res = await webex.internal.calendar.getParticipants(id);
          assert.equal(res.body.encryptedParticipants.length, 1);
```

adamweeks:
Apologies, but you will need to resolve the conflicts that have arisen from merging #1534

PR #1468, opened 2020-01-16
fix(meetings): add close remote stream on cleanup

adamweeks:
# Pull Request Template

Meetings plugin remote stream fix.

## Description

Remote stream was not being closed during meeting cleanup, causing the media stopped events to not trigger.

## Type of Change


adamweeks:
Tests are failing and no allocation of time to investigate. Will open a JIRA bug for proper tracking.

PR #1470, opened 2020-01-21
Refactor/migrate embargo interceptor

InteractiveTimmy:
# Pull Request

## Description

This pull request moves the embargo interceptor from `internal-plugin-wdm` to `webex-core`, as the `internal-plugin-wdm` will be replaced in an upcoming change.

Fixes # SPARK-111767

## Test Coverage

- [x] Migrated Embarger Interceptor tests as well as generated more tests to improve coverage.
- [x] Ran both `webex-core` and `internal-plugin-wdm` test suites to validate changes worked.

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:


packages/node_modules/@webex/webex-core/src/webex-core.js
```diff
@@ -24,6 +24,7 @@ import UserAgentInterceptor from './interceptors/user-agent';
 import WebexTrackingIdInterceptor from './interceptors/webex-tracking-id';
 import WebexUserAgentInterceptor from './interceptors/webex-user-agent';
 import RateLimitInterceptor from './interceptors/rate-limit';
+import EmbargoInterceptor from './interceptors/embargo';
```
InteractiveTimmy: This properly loads the interceptor into the base package. [see other entries in this file]

packages/node_modules/@webex/webex-core/test/unit/spec/interceptors/embargo.js
```diff
@@ -0,0 +1,139 @@
+import chai from 'chai';
```
InteractiveTimmy: This is an enhanced test that covers additional ground. Reminder, that interceptors can only be tested as a unit tests since they are loaded into non-addressable namespaces in an instance function.

packages/node_modules/@webex/webex-core/test/unit/spec/interceptors/embargo.js
```diff
@@ -0,0 +1,139 @@
...
+
+          beforeEach('set up the device plugin', () => {
+            interceptor.webex.internal.device = {
+              clear: sinon.spy()
+            };
+
+            deviceClear = interceptor.webex.internal.device.clear;
+          });
+
+          it('should not clear the device', () =>
```
adamweeks: Shouldn't this be true according to implementation code:

```js
      if (device) {
        this.webex.internal.device.clear();
      }
```

InteractiveTimmy: If a user is utilizing `webex-core` without the device plugin being mounted, it will fail here since `clear()` will not exists on `undefined` `device`. Previously, this interceptor was mounted with `device`, so that the check wasn't required.

adamweeks: Oh, I'm sorry, I misread the describe blocks. Was thinking this was for the 451 error.

PR #1473, opened 2020-01-22
Refactor/migrate url interceptor

InteractiveTimmy:
# Pull Request

## Description

This change removes the device url interceptor which utilizes the device service host map when making requests using `webex.request()`.

Fixes # SPARK-111766

## Test Coverage

- [x] Updated `internal-plugin-wdm` package tests to accomidate for no url interceptor.
- [x] Ran package tests for both `internal-plugin-wdm` and `webex-core`.

**Test Configuration**:
* NPM Version v6.4.1

## Checklist:


packages/node_modules/@webex/webex-core/src/interceptors/auth.js
```diff
@@ -90,23 +91,29 @@ export default class AuthInterceptor extends Interceptor {
...
+        if (services.hasAllowedDomains() &&
+          (uri && services.isAllowedDomainUrl(uri))) {
+          return true;
+        }
 
-    return Promise.resolve(false);
+        // Return false to indicate authentication is not required.
+        return false;
+      })
+      .catch(() => false);
```
mwegman: I would remove this line. Let something else handle an exception or `Promise.reject`. If we insist on keeping it, we should at least log the error before returning `false`.

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
```diff
@@ -21,100 +21,80 @@ export default class ServiceInterceptor extends Interceptor {
   /* eslint-disable no-param-reassign */
   /**
    * @see Interceptor#onRequest
-   * @param {Object} options
-   * @returns {Object}
+   * @param {Object} options - The request PTO.
+   * @returns {Object} - The mudated request PTO.
```
mwegman: mutated?

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
```diff
@@ -21,100 +21,80 @@ export default class ServiceInterceptor extends Interceptor {
...
-            }
+  /* eslint-disable class-methods-use-this */
+  /**
+   * Generate a usable request uri string from a service url and a resouce.
+   *
+   * @param {string} serviceUrl - The service url.
+   * @param {string} resource - The resouce to be appended to the service url.
+   * @returns {string} - The combined service url and resource.
+   */
+  generateUri(serviceUrl, resource) {
```
mwegman: Why don't we move this method out of the class? We wouldn't need the `eslint-disable` above.

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
```diff
@@ -21,100 +21,80 @@ export default class ServiceInterceptor extends Interceptor {
...
-          })
-          .catch((e) => Promise.reject(e));
-      }
+  /**
+   * Normalizes request options relative to service identification.
+   *
+   * @param {Object} options - The request PTO.
+   * @returns {Object} - The mudated request PTO.
+   */
+  normalizeOptions(options) {
```
mwegman: Same as above

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
```diff
@@ -21,100 +21,80 @@ export default class ServiceInterceptor extends Interceptor {
...
-   * Verify that all required parameters have been specified.
-   * @param {object} options
-   * @param {string} options.resource
-   * @returns {void}
+   * Validates that the appropriate options for this interceptor are present.
+   *
+   * @param {Object} options - The request PTO.
+   * @returns {Object} - The mudated request PTO.
    */
   validateOptions(options) {
```
mwegman: Same as above

InteractiveTimmy: Decided against moving these, as it would require a special-case export to perform unit tests against the methods. Which would be unlike all other interceptors.

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
InteractiveTimmy: Decided against moving these, as it would require a special-case export to perform unit tests against the methods. Which would be unlike all other interceptors.

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
InteractiveTimmy: Decided against moving these, as it would require a special-case export to perform unit tests against the methods. Which would be unlike all other interceptors.

packages/node_modules/@webex/webex-core/src/lib/services/interceptors/service.js
InteractiveTimmy: Fixing this, good catch!

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
InteractiveTimmy: I'm improving this catch statement to properly indicate via the logger the reason why the value was returned as false.

PR #1486, opened 2020-01-31
refactor(internal-plugin-device): Deprecate WDM

InteractiveTimmy:
# Pull Request

## Description

The changes in this pull request are inclusive of overriding all of the existing `@webex/internal-plugin-wdm` plugin's content with new, cleaner logic generated in a previous pull request as `@webex/internal-plugin-devices`. There is also a rename of that plugin to `@webex/internal-plugin-device`, as it now matches the appropriate namespace after overriding the `@webex/internal-plugin-wdm` plugin.

Fixes #
- SPARK-111771: Remap Devices to WDM
- SPARK-111770: Remove WDM Plugin

## Type of Change

## Test Coverage

- [x] Generated additional tests to account for plugin namespace mapping
- [x] Generate new tests for new members within the migrated plugin
- [x] Amended a namespace referencing test to the original plugin.
- [x] Ran all tests for affected plugins with passing results
- [x] Ran all impacted plugin tests to validate that the change did not generate new errors within those plugins.

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:


packages/node_modules/@webex/internal-plugin-wdm/README.md
```diff
@@ -6,12 +6,17 @@
...
 This is an internal Cisco Webex plugin. As such, it does not strictly adhere to semantic versioning. Use at your own risk. If you're not working on one of our first party clients, please look at our [developer api](https://developer.webex.com/getting-started.html) and stick to our public plugins.
 
+- [Deprecation](#deprecation)
 - [Install](#install)
 - [Usage](#usage)
 - [Contribute](#contribute)
 - [Maintainers](#maintainers)
 - [License](#license)
 
+## Deprecation
```
InteractiveTimmy: Simple plugin deprecation message, any feedback on how this should appear is welcome.

packages/node_modules/@webex/internal-plugin-wdm/src/index.js
```diff
@@ -2,28 +2,21 @@
...
-
-import Device from './device';
-import config from './config';
-import DeviceUrlInterceptor from './interceptors/device-url';
-
-registerInternalPlugin('device', Device, {
-  interceptors: {
-    DeviceUrlInterceptor: DeviceUrlInterceptor.create
-  },
+import Device, {
```
InteractiveTimmy: With this type of import/export setup, the `@webex/internal-plugin-wdm` simply points to the new `@webex/internal-plugin-device` so that they are identical in every way. This prevents the need to update every import across all packages.

packages/node_modules/@webex/internal-plugin-wdm/test/integration/spec/webex.js
```diff
@@ -5,33 +5,45 @@
 import '@webex/internal-plugin-wdm';
 
 import {assert} from '@webex/test-helper-chai';
-import sinon from 'sinon';
 import WebexCore from '@webex/webex-core';
 import testUsers from '@webex/test-helper-test-users';
 
-describe('plugin-wdm', function () {
-  this.timeout(30000);
+describe('plugin-wdm', () => {
```
InteractiveTimmy: This file has been updated to test that the `@webex/internal-plugin-device` plugin mounts correctly when utilizing `@webex/internal-plugin-wdm`.

packages/node_modules/@webex/internal-plugin-device/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@webex/internal-plugin-devices",
-  "version": "1.80.96",
+  "name": "@webex/internal-plugin-device",
+  "version": "1.80.95",
```
adamweeks: I think there's a package conflict here that needs to be updated.

packages/node_modules/@webex/internal-plugin-device/test/integration/spec/device.js
```diff
@@ -962,45 +965,45 @@ describe('plugin-devices', () => {
...
 
         it('should return a resolved promise',
-          () => assert.isFulfilled(devices.waitForRegistration()));
+          () => assert.isFulfilled(device.waitForRegistration()));
       });
 
       describe('when the device is not registered', () => {
         beforeEach('verify the device is not registered', () => {
-          assert.isFalse(devices.registered);
+          assert.isFalse(device.registered);
```
adamweeks: Is there a way to set this manually instead of relying on an assert in a before each block?

packages/node_modules/@webex/internal-plugin-wdm/README.md
adamweeks: I think this works for now since we are still planning on supporting users using the `wdm` plugin. 

Once we completely wipe it out, let's use the readme template from one of the `@ciscospark` plugins that points to the new folder.

adamweeks:
Do we need to update the "webex" package to use the new plugin as well? 
https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/webex/src/webex.js#L15

InteractiveTimmy:
We do not, as it's mountable through the `@webex/internal-plugin-wdm` package.

packages/node_modules/@webex/internal-plugin-device/package.json
InteractiveTimmy: I'm not sure what package `@webex/internal-plugin-device` collides with, please let me know if this is the case.

packages/node_modules/@webex/internal-plugin-device/test/integration/spec/device.js
InteractiveTimmy: This can be changed, I'll investigate this further.

InteractiveTimmy: Addressed these, they were redundant assertions that weren't necessary and were determined to be used for test-writing breakpoints.

packages/node_modules/@webex/internal-plugin-device/package.json
InteractiveTimmy: Resolved with a rebase!

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
```diff
@@ -79,6 +79,57 @@ const FeatureModel = AmpState.extend({
...
+    // Generate a return transfer object.
+    let rto = {};
+
+    // Validate that a parameter transfer object was provided.
+    if (!pto) {
+      return rto;
+    }
+
+    // Clone the parameter transfer object into the return transfer object.
+    rto = {...pto};
```
lalli-flores: If you're going to clone it like this, then the above check to see if `pto` is defined is not necessary.
Cloning in line 90 like so `const rto = {..pto};` would create an empty object if there's nothing in `pto`.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
```diff
@@ -79,6 +79,57 @@ const FeatureModel = AmpState.extend({
...
+
+    // Validate that a value was provided.
+    if (rto.val) {
```
lalli-flores: I'm a little confused by this line. If the expected parameter type is `FeatureModel` and the return type is _also_ a `FeatureModel`, why are the attributes different?

packages/node_modules/@webex/internal-plugin-device/src/interceptors/device-url.js
```diff
@@ -24,18 +24,15 @@ export default class DeviceUrlInterceptor extends Interceptor {
...
   onRequest(options) {
     const {headers, service, uri} = options;
-    const {devices, services} = this.webex.internal;
+    const {device, services} = this.webex.internal;
+    const cdu = 'cisco-device-url';
 
-    // Check if header is already set before moving forward
-    if (!devices.url || (headers && !headers['cisco-device-url'])) {
+    // Check if header is already set before moving forward.
+    if (!device.url || (headers && cdu in headers && !headers[cdu])) {
```
lalli-flores: is the `cdu in headers` check necessary? Since you're also checking for `!headers[cdu]`

packages/node_modules/@webex/internal-plugin-wdm/src/index.js
lalli-flores: I think this works for now, but we should probably completely remove usage at some point (sooner rather than later)

packages/node_modules/@webex/internal-plugin-wdm/test/integration/spec/webex.js
```diff
@@ -5,33 +5,45 @@
...
+    });
 
-  describe('onBeforeLogout()', () => {
-    it('unregisters the device', () => webex.logout({noRedirect: true})
-      .then(() => {
-        assert.called(webex.internal.device.unregister);
-        assert.isFalse(webex.internal.device.registered);
-      }));
+    describe('when the plugin is imported', () => {
+      it('should define a \'device\' plugin', () =>
```
lalli-flores:
```suggestion
      it(`should define a 'device' plugin`, () =>
```

OR

```suggestion
      it('should define a "device" plugin', () =>
```

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/features/feature-model.js
```diff
@@ -56,6 +56,117 @@ describe('plugin-devices', () => {
...
+    describe('#parse()', () => {
+      let fixture;
+      let rto;
+
+      beforeEach('initialize the feature model', () => {
+        featureModel = new FeatureModel();
+      });
+
+      describe('when the pto is not defined', () => {
+        beforeEach('get the rto', () => {
```
lalli-flores: Since only one test use this, i think it's fine to have the small setup within the test, instead of creating a describe block for it

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/features/feature-model.js
```diff
@@ -56,6 +56,117 @@ describe('plugin-devices', () => {
...
+        describe('when the value is a number', () => {
+          let num;
+
+          beforeEach('set the value to a number', () => {
+            fixture.val = '1234';
+            num = Number(fixture.val);
+            rto = featureModel.parse(fixture);
+          });
+
+          it('should set the value to a instance of number', () =>
```
lalli-flores: In this tests we have one assertion per test, but on others there are multiple assertions. I would like to keep the testing consistent. which pattern are we following?

adamweeks:
> We do not, as it's mountable through the `@webex/internal-plugin-wdm` package.

I would assume that in the "Deprecate WDM" PR we wouldn't want to keep using the wdm package publicly.

packages/node_modules/@webex/internal-plugin-device/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@webex/internal-plugin-devices",
-  "version": "1.80.99",
+  "name": "@webex/internal-plugin-device",
+  "version": "1.80.95",
```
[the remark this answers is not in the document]
InteractiveTimmy: This is due to package bumping, once this is merged, it should automatically bump the package versions to the appropriate versions.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
```diff
@@ -79,6 +79,57 @@ const FeatureModel = AmpState.extend({
 
   // Ampsersand method members.
 
+  /**
+   * Parse all attribute values and map them to their appropriate members.
+   *
+   * @param {FeatureModel} pto - Attributes to be parsed and mapped.
```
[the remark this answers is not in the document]
InteractiveTimmy: `PTO` => `parameter transfer object`, makes it is technically a `DTO` [common name], but since the method creates a return object `RTO` => `return transfer object`.

Let me know if you have any recommendations on what these namespaces should be named and I will make the change.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
```diff
@@ -79,6 +79,57 @@ const FeatureModel = AmpState.extend({
...
+   * @returns {FeatureModel} - Parsed and mapped attributes.
+   */
+  parse(pto) {
+    // Generate a return transfer object.
+    let rto = {};
```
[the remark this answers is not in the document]
InteractiveTimmy: `PTO` => `parameter transfer object`, makes it is technically a `DTO` [common name], but since the method creates a return object `RTO` => `return transfer object`.

Let me know if you have any recommendations on what these namespaces should be named and I will make the change.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
InteractiveTimmy: I will add a default value of `{}` for the `pto` object.

packages/node_modules/@webex/internal-plugin-device/src/interceptors/device-url.js
InteractiveTimmy: Yes, @taymoork2 and I discussed removing it, but after we merged the changes, other plugins would fail tests.

This is due to how `cdu` has to exist as a key, but return `false` when `!headers[cdu]` is interpreted in order for this interceptor's logic to flow properly.

packages/node_modules/@webex/internal-plugin-wdm/test/integration/spec/webex.js
InteractiveTimmy: A majority of our testing method descriptions in the JSSDK utilize `'` characters when mentioning strings as seen above. ESLint will throw if the ` character is used in a string without the usage of `${value}`. I can make a larger change to correct this, but this will break the style consistency of this test against all of the tests in the JS SDK.

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/features/feature-model.js
InteractiveTimmy: Will refactor this.

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/features/feature-model.js
InteractiveTimmy: I believe I set this up this way to test more aggressively at the lowest child class object, as all upstream classes depend on the consistency of this class. I can refactor this if needed.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
InteractiveTimmy: This method normalizes the feature model DTO. This is also a Ampersand method override that passes itself into the method to normalize data types. I would personally change this if ampersand wasn't required to maintain functionality across the JS SDK. see [this documentation entry](https://ampersandjs.com/docs/#ampersand-state-parse) for more details.

packages/node_modules/@webex/internal-plugin-device/package.json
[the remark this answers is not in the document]
lalli-flores: Alright, thanks for clarification

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
lalli-flores: Even if `pto` is undefined, `{..pto}` would still give you `{}` so no need to add the default value. Although, I don't think it can hurt

lalli-flores:
> > > We do not, as it's mountable through the `@webex/internal-plugin-wdm` package.
> > 
> > 
> > I would assume that in the "Deprecate WDM" PR we wouldn't want to keep using the wdm package publicly.
> 
> If we perform these changes, the scope of this JIRA becomes rather large [since there are about 19 plugins that directly reference `@webex/internal-plugin-wdm`. A separate PR may be safer, but I'm willing to update all of the other plugins if you deem it necessary for the scope of the JIRAs.

I think it needs to be done, but definitely as part of a separate PR

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
lalli-flores: So I guess my question is (for context), why do we have to do this normalization in the first place?

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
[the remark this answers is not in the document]
lalli-flores: yeah, as someone not familiar with the SDK, coming into the code, that is not obvious at all. Like in my other comment, my suggestion would be to rename this to something more explicit

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/features/feature-model.js
lalli-flores: Personally, I prefer one assertion per test, but I saw some other code in this PR that does not follow that format

packages/node_modules/@webex/internal-plugin-wdm/test/integration/spec/webex.js
lalli-flores: Not a big deal, just thought is unnecessary to escape `'` when there are other ways of achieving the same visual effect, that are also more pleasing to the eye in the code

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
InteractiveTimmy: The data transfer object is received from the `wdm` service as a `Record<string, string>` for features. This makes things appear as so:
```js
{
  // ...
  type: 'boolean'
  val: 'true'
  // ...
}
```
This makes comparison operations like so:
```js
if (feature) { /* ... */ }
```
always true. These features must be re-casted as their proper types.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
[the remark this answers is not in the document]
adamweeks: I keep going back and forth but I think I have to agree with @lalli-flores  on this. Even calling it `featureModel` gives a much easier context to process for others. `pto` makes me think I missed something.

packages/node_modules/@webex/internal-plugin-device/src/features/feature-model.js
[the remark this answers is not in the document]
adamweeks: `generatedFeatureModel` conveys the same idea with less ambiguity.

InteractiveTimmy:
Closing this, as the changes have been fully implemented via other pull requests.

lalli-flores:
> Closing this, as the changes have been fully implemented via other pull requests.

Can you add the PRs this was replaced by?

PR #1520, opened 2020-02-17
Fix/host fail flags

InteractiveTimmy:
# Pull Request

## Description

The changes within this pull request are inclusive of the fixes needed to improve host removal/failure. Instead of immediately removing hosts that are failed, simply mark them instead.

## Type of Change

## Test Coverage

- [x] Updated testing suite to validate changes.
- [x] Ran package tests to validate the changes work as expected.

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
@@ -56,9 +56,23 @@ const ServiceUrl = AmpState.extend({
       return this.defaultUrl;
     }
 
-    const filteredHosts = clusterId ?
+    let filteredHosts = clusterId ?
       this.hosts.filter((host) => host.id === clusterId) :
       this.hosts.filter((host) => host.homeCluster);
+
+    const aliveHosts = filteredHosts.filter(
+      (host) => !host.failed
+    );
+
+    filteredHosts = (aliveHosts.length === 0) ?
+      filteredHosts.map(
+        (host) => {
+          /* eslint-disable-next-line no-param-reassign */
+          host.failed = false;
+
+          return host;
+        }
+      ) : aliveHosts;
 
     return this._generateHostUrl(filteredHosts.reduce((previous, current) => (
       (previous.priority > current.priority || !previous.homeCluster) ?
@@ -67,15 +81,13 @@ const ServiceUrl = AmpState.extend({
   },
 
   /**
-   * Attempt to remove a host from this `ServiceUrl`
-   * and return true if the provided `url` had a host
-   * that could be successfully removed or if the url
-   * is assocaited with a defaultUrl, and cannot be
-   * removed.
+   * Attempt to mark a host from this `ServiceUrl` as failed and return true
+   * if the provided url has a host that could be successfully marked as failed.
+   *
    * @param {string} url
    * @returns {boolean}
    */
-  removeHost(url) {
+  failHost(url) {
     if (url === this.defaultUrl) {
       return true;
     }
@@ -83,7 +95,9 @@ const ServiceUrl = AmpState.extend({
     const {hostname} = Url.parse(url);
     const foundHost = this.hosts.find((hostObj) => hostObj.host === hostname);
 
-    this.hosts.splice(this.hosts.indexOf(foundHost), 1);
+    if (foundHost) {
+      foundHost.failed = true;
+    }
 
     return (foundHost !== undefined);
   },
--- a/packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
+++ b/packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
@@ -436,18 +436,22 @@ describe('webex-core', () => {
       let testUrl;
 
       beforeEach('load test url', () => {
+        catalog.clean();
+
         testUrlTemplate = {
           defaultUrl: 'https://www.example-phr.com/api/v1',
           hosts: [
             {
               host: 'www.example-phr-p5.com',
               ttl: -1,
-              priority: 5
+              priority: 5,
+              homeCluster: true
             },
             {
               host: 'www.example-phr-p3.com',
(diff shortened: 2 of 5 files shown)


packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -476,6 +484,23 @@ describe('webex-core', () => {
...
+            const priorityUrl = priorityServiceUrl._getPriorityHostUrl();
+
+            services.markFailedUrl(priorityUrl);
+          }
+        );
+
+        const lastPriorityUrl = priorityServiceUrl._getPriorityHostUrl();
+
+        assert.equal(firstPriorityUrl, lastPriorityUrl);
+      });
```
InteractiveTimmy: This test validates that the hosts are reset and are usable after all have been marked as failed.

packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
```diff
@@ -56,10 +56,24 @@ const ServiceUrl = AmpState.extend({
...
 
+    const aliveHosts = filteredHosts.filter(
+      (host) => !host.failed
+    );
+
+    filteredHosts = (aliveHosts.length === 0) ?
+      filteredHosts.map(
+        (host) => {
+          /* eslint-disable-next-line no-param-reassign */
+          host.failed = false;
```
adamweeks: You could remove the eslint disable by creating a new object and returning:
```suggestion
          const aliveHost = {...host, failed: false};
```

adamweeks: If the idea is to update those hosts in the `this.hosts` array directly, it might be better to make that a specific function since this function is a getter and you don't really want side effects in a getter.

packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
```diff
@@ -83,7 +97,9 @@ const ServiceUrl = AmpState.extend({
     const {hostname} = Url.parse(url);
     const foundHost = this.hosts.find((hostObj) => hostObj.host === hostname);
 
-    this.hosts.splice(this.hosts.indexOf(foundHost), 1);
+    if (foundHost) {
+      foundHost.failed = true;
```
adamweeks: Would it be better to rename this function `failHost` otherwise, you have a mental overhead added to understand that failed is the same as removed?

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -460,13 +464,17 @@ describe('webex-core', () => {
...
+        const priorityUrl = priorityServiceUrl._getPriorityHostUrl();
 
         services.markFailedUrl(priorityUrl);
 
-        assert.isBelow(testUrl.hosts.length, originalServiceUrlHostLength);
+        const failedHost = priorityServiceUrl.hosts.find(
+          (host) => host.failed
+        );
+
+        assert.isDefined(failedHost);
```
adamweeks: Could we be a bit more specific in this assertion and check that `failedHost.url === priorityUrl`?

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -476,6 +484,23 @@ describe('webex-core', () => {
 
         assert.notEqual(priorityUrl, nextPriorityUrl);
       });
+
+      it('should reset hosts once all hosts have been marked failed', () => {
```
adamweeks: This test seems to validate the `_getPriorityHostUrl` feature works. Would a simpler test be to loop through `.hosts` and set failed, then call `markFailedUrl` and check the failed status of all the hosts again?

adamweeks: Ok, I see you're doing this in the unit tests.

packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
InteractiveTimmy: Properly implementing this would require the creation of a `ServiceHost` class, as there isn't a way to provide mutability while also avoiding param reassign on nested objects. This could be done as a part of the planned refactor of this plugin. Examples:

This would create mismatched data:
```js
class ServiceUrl {
  ...
  resetHosts() {
    const preHosts = this.hosts;

    this.hosts = this.hosts.map(
      (host) => ({...host, failed: false})
    );

    if (this.hosts !== preHosts) {
      throw new Error('bad mutability! cannot do!');
    }
}
```

packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
InteractiveTimmy: Yep, implementing this.

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
InteractiveTimmy: I believe I have an implementation for this, generating it now.

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
InteractiveTimmy: In that case, should I leave this as is?

packages/node_modules/@webex/webex-core/src/lib/services/service-url.js
```diff
@@ -75,15 +89,17 @@ const ServiceUrl = AmpState.extend({
    * @param {string} url
    * @returns {boolean}
    */
-  removeHost(url) {
+  failHost(url) {
```
adamweeks: Could you update the jsdoc on this renamed function? It still references removing the host.

PR #1527, opened 2020-02-18
fix(metrics): fix originTime

chrisadubois:


packages/node_modules/@webex/internal-plugin-metrics/src/call-diagnostic-events-batcher.js
```diff
@@ -23,6 +23,7 @@ const CallDiagnosticEventsBatcher = Batcher.extend({
   prepareRequest(queue) {
     // Add sent timestamp
     queue.forEach((item) => {
+      item.eventPayload.originTime = item.eventPayload.originTime || {};
```
adamweeks: Could you add a section in the unit tests that validate this bugfix works? I find it best to write a test that fails before the bugfix implementation, then passes afterwards. Thanks!

PR #1540, opened 2020-02-26
better handling of screensharing takeover

taymoork2:
## Description

screensharing currently doesn't handle a takeover that well, and the `isSharing` variable doesn't get updated
The current solution is to check if a takeover happened and then stop the local share using `stopShare`

NOTE: This is still slightly buggy because once a takeover happens, locus gets conflicted and doesn't properly allow the plugin user to takeover

Fixes [SPARK-116558](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-116558)

## Type of Change

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -762,6 +762,23 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   setUpLocusMediaSharesListener() {
     this.locusInfo.on(EVENTS.LOCUS_INFO_UPDATE_MEDIA_SHARES, (payload) => {
+      const {contentId, disposition} = payload.current;
+
+      // Check if screenshare is being taken over by a remote
+      // If so, terminate the share and allow the remote to be on the floor
+      // Terminating the share allows the user to screenshare again without any issues
+      if (
+        this.isSharing &&
+        this.selfId !== contentId &&
+        disposition !== FLOOR_ACTION.RELEASED
+      ) {
+        this.isSharing = false;
+        this.updateShare({
+          sendShare: false,
+          receiveShare: this.mediaProperties.mediaDirection.receiveShare
+        });
+      }
+
       this.members.locusMediaSharesUpdate(payload);
     });
   }


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -735,6 +735,21 @@ export default class Meeting extends StatelessWebexPlugin {
...
+       */
+
+      const {contentId, disposition} = payload.current;
+
+      if (
+        this.isSharing &&
+        this.selfId !== contentId &&
+        disposition !== FLOOR_ACTION.RELEASED
+      ) {
+        this.stopShare();
```
arun3528: this.stopShare just makes the api call for the floor grant service but you also need to do updateShare(sendShare:false, receiveShare:true).

I guess with you just do updateShare it should handle the `this.stopShare` logic as well

taymoork2:
Tested with a call between chrome <--> native client

1. Chrome calls Native
2. Native starts screenshare
3. Chrome takes over screenshare from Native
4. Native resumes/takes over screenshare from Chrome
	- Chrome screenshare is terminated
	- Errors (floor related) do show up in the console, but do **NOT** cause call to crash out
5. Chrome takes over screenshare from Native _**[FLAKY]**_
	- Takes it a bit for it to register
	- However, it does take over successfully
6. Chrome stops screenshare and ends call

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -141,31 +141,31 @@ export default class LocusInfo extends EventsScope {
   /**
    * updates the locus with full locus object
    * @param {object} locus locus object
-   * @param {sting} eventType particulat locus event
+   * @param {sting} locusEventType particulat locus event
```
adamweeks:
```suggestion
   * @param {string} locusEventType particular locus event
```

adamweeks:
@taymoork2  thanks for fixing the upper scope issue, but could you split the `refactor(meetings): fix redeclared upper scope variable` commit into a different PR since I think this screenshare work will be a bit more involved?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -761,6 +761,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+      // Check if screenshare is being taken over by a remote
+      // If so, terminate the share and allow the remote to be on the floor
+      // Terminating the share allows the user to screenshare again without any issues
+      if (
+        this.isSharing &&
+        this.selfId !== contentId &&
+        disposition !== FLOOR_ACTION.RELEASED
+      ) {
+        this.updateShare({sendShare: false, receiveShare: true});
```
adamweeks: Whenever I try this, I get an error of `'The peer connection is currently negotiation an offer. Please wait for a few seconds and try again.'`. Should we first manually set `this.isSharing = false` and then wait for the media to be ready to stop the share? 

@arun3528 also mentioned that the other clients don't actually stop the share, just pause it, when the floor gets released.

adamweeks: receiveShare should come from the media properties per @arun3528

PR #1541, opened 2020-02-26
feat(meeting): add new media metrics for MQA dashboard analyzer

chrisadubois:
add a new metric type (no jira)


adamweeks:
@chrisadubois these changes look good. Looks like we just need to update our tests for the new values.

adamweeks:
Since this is a Stale PR, we are going to close it. Please reopen when ready 😄

PR #1542, opened 2020-02-27
feat(conversation): respect key rotation

dalipkumar703:
# Pull Request Template
This PR take care of respect key rotation error codes.
## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


dalipkumar703:
> Two things are still needed before approval can be determined
> 
> 1. This needs test(s) to make sure that your changes are working and passing (also for coverage)
> 2. Comments! I'm not entirely sure what this code is trying to achieve and/or how it's achieving it

To give brief, It try to resend failed activity after key rotation when receive error codes ``1400087`` or ``1409004`` . For achieve that decrypting the older activity that got failed and sending it again. It might be content or text activity. So following case for both, I will update it that will follow guide line and cover test case.

@taymoork2 Could you suggest whether for resending lost activity there is another simpler function already ? or Is it possible to send lost activity again without decrypting?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -901,7 +902,60 @@ const Conversation = WebexPlugin.extend({
...
+        let convoUrl;
+
+        if (process.env.NODE_ENV === 'production') {
+          convoUrl = 'https://conv-a.wbx2.com/conversation/api/v1/conversations/';
+        }
+        else {
+          convoUrl = 'https://conversation-intb.ciscospark.com/conversation/api/v1/conversations/';
+        }
+        const transform = decryptTransforms.find((t) => t.name === 'decryptTextProp');
+        const contentTransform = decryptTransforms.find((t) => t.name === 'decryptContent');
```
aimeex3: shouldn't the transformers defined in conversation plugin / index.js file handle the decryption already?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -901,7 +902,60 @@ const Conversation = WebexPlugin.extend({
...
+        const ctxInstance = {webex: this.webex, transform: this.webex.transform, this: this.webex.config};
+
+        // handle when key need to rotate
+        if (error.body && error.body.errorCode === 1400087) {
+          this.request(`${convoUrl + params.body.target.id}/healed?resetKey=true`).then(() => {
```
aimeex3: i think you should be able to use 
webex.request() and specify the service is name and it will automatically get the correct host url based on the service catalog. and the objects should be decrypted (or you may need to modify the predicates in index.js to handle them

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
dalipkumar703: decryption are defined in decrypt-transformers file. And the lost activity is already encrypted to resend it again it require to decrypt it first otherwise user will receive decrypt message.

andrewholsted: I think we should be able to resend using the same props/params as the original send?

dalipkumar703: I tried but user receive encrypted message instead of original message. Therefore, I decrypted again.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
```diff
@@ -208,6 +208,42 @@ describe('plugin-conversation', function () {
...
+          method: 'GET',
+          api: 'conversation',
+          resource: `/conversations/${convo.id}/healed?resetKey=true`
+        })
+          .then(() => webex.request({
+            method: 'GET',
+            api: 'conversation',
+            resource: `conversations/${convo.id}`
+          }))
+          .then((res) => {
```
taymoork2: I think this should be fine, as healed doesn't pass anything to the next request.
And this still passes locally

```suggestion
Promise.all([
	webex.request({
       	method: 'GET',
		api: 'conversation',
		resource: `/conversations/${convo.id}/healed?resetKey=true`
	}),
	webex.request({
        method: 'GET',
        api: 'conversation',
        resource: `conversations/${convo.id}`
    })
]).then(([, res]) => {
```

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -104,5 +109,54 @@ describe('plugin-conversation', () => {
...
+          actor: {
+            objectType: 'person',
+            id: 'id'
+          },
+          object: {
+            displayName: 'updateKey',
+            content: 'updateKey',
+            objectType: 'conversation'
+          }
+        };
```
taymoork2:
```suggestion
        const activity = {
          actor: {
            objectType: 'person',
            id: 'id'
          },
          object: {
            displayName: 'updateKey',
            content: 'updateKey',
            objectType: 'conversation'
          },
          target: {
          	// URL is from from another test that needs to mock encryption service url
            defaultActivityEncryptionKeyUrl: `${process.env.ENCRYPTION_SERVICE_URL || 'https://encryption-a.wbx2.com'}/encryption/api/v1/keys/8a7d3d78-ce75-48aa-a943-2e8acf63fbc9`
          }
        };
```

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -104,5 +109,54 @@ describe('plugin-conversation', () => {
...
```
taymoork2:
```suggestion
        const activity = {
          actor: {
            objectType: 'person',
            id: 'id'
          },
          object: {
            displayName: 'updateKey',
            content: 'updateKey',
            objectType: 'conversation'
          },
          target: {
          	// URL is from from another test that needs to mock encryption service url
            defaultActivityEncryptionKeyUrl: `${process.env.ENCRYPTION_SERVICE_URL || 'https://encryption-a.wbx2.com'}/encryption/api/v1/keys/8a7d3d78-ce75-48aa-a943-2e8acf63fbc9`
          }
        };
```

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
dalipkumar703: does putting it on promise make sure that get convo called later after heal?

taymoork2: So this is the first time i've seen healed and, off the top of my head, don't know where it's documented to see what it does (can send me the link on teams).

taymoork2: `Promise.all` runs them in async and thens `then` after all promises are resolved, so if it needs to be one after the other, i'm fine with the current way its written.
I just don't really like nested requests, but i dont know a cleaner way of writing them besides the `.all`

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -901,9 +903,100 @@ const Conversation = WebexPlugin.extend({
...
-      .then((res) => res.body);
+      .then((res) => res.body)
+      .catch((error) => {
+        // handle when key need to rotate
+        if (error.body && error.body.errorCode === KEY_ROTATION_REQUIRED) {
+          this.webex.request({
+            method: 'GET',
+            api: 'conversation',
+            resource: `conversations/${params.body.target.id}`
+          }).then((res) => {
```
andrewholsted: Why do we do another GET before setting the key to null?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -901,9 +903,100 @@ const Conversation = WebexPlugin.extend({
...
+        else if (
+          error.body &&
+    (error.body.errorCode === KEY_ALREADY_ROTATED || error.body.errorCode === ENCRYPTION_KEY_URL_MISMATCH)
+        ) {
+          // handle when key need to update
+          this.webex.request({
+            method: 'GET',
+            api: 'conversation',
+            resource: `conversations/${params.body.target.id}`
+          }).then((res) => {
```
andrewholsted: This GET request makes sense since we need to get the new key

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -901,9 +903,96 @@ const Conversation = WebexPlugin.extend({
...
+  },
+
+
+  /**
+    * Resend lost share activity
+    * @param {res} res
+    * @param {activity} activity
+    * @returns {Promise}
+    */
+  resendShareActivity(res, activity) {
```
andrewholsted: @aimeex3 @adamweeks @InteractiveTimmy @taymoork2 - this still feels like an antipattern to me. Dalip said the main source of the issue is that `params` gets modified in place when the request is sent/encrypted. What are your thoughts on cloning the request params before sending an activity? It would do it on _every_ send which isn't great but something about this feels like it will be hard to debug and is going against the typical transform patterns

dalipkumar703: @andrewholsted @chrisadubois @taymoork2 I did comparison using cloneDeep for copying activity object and sending message and without cloneDeep and observed time in ms. I did not see major difference but I considered two cases images and text there might be some scenario when it cause large time difference. Let me know if i need to follow clone activity approach.

|without_cloneDeep|without_cloneDeep|with_cloneDeep|with_cloneDeep|
---|---|---|---|
|text(ms) | image(ms)|text(ms)|image(ms)|
|---|---|---|---|
|296|560|477|602
|278|670|371|554
 271|593|413|585
  288|542|375|613
  283|520|382|627
  292|544|518|587
  306|552|421|590
  412||415
  319||434
  344||456
  320||443
  332||456
  358||463
  405||479
  358||501
  350||525
  378||498
  368||571

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -104,5 +109,62 @@ describe('plugin-conversation', () => {
...
+            id: 'id'
+          },
+          object: {
+            displayName: 'updateKey',
+            content: 'updateKey',
+            objectType: 'conversation'
+          },
+          target: {
+            //  mock encryption service url
+            defaultActivityEncryptionKeyUrl: `${process.env.ENCRYPTION_SERVICE_URL || 'https://encryption-a.wbx2.com'}/encryption/api/v1/keys/8a7d3d78-ce75-48aa-a943-2e8acf63fbc9`
```
adamweeks: If the request is mocked, do we need the env var injection here?

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -104,5 +109,62 @@ describe('plugin-conversation', () => {
...
+            id: 'id'
+          },
+          object: {
+            displayName: 'updateKey',
+            content: 'updateKey',
+            objectType: 'conversation'
+          },
+          target: {
+            // URL is from from another test that needs to mock encryption service url
+            defaultActivityEncryptionKeyUrl: `${process.env.ENCRYPTION_SERVICE_URL || 'https://encryption-a.wbx2.com'}/encryption/api/v1/keys/8a7d3d78-ce75-48aa-a943-2e8acf63fbc9`
```
adamweeks: Same question here about the env injection.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
```diff
@@ -208,6 +208,43 @@ describe('plugin-conversation', function () {
         .then((reason) => assert.equal(reason.status, 403)));
     });
 
+    describe('#post() with error handling', () => {
```
adamweeks: Apologies, but it looks like you used the pattern set by the ['#leave() with id only'](https://github.com/webex/webex-js-sdk/blob/c378d07f8622078603384d17e90ef3d04e19a2ff/packages/node_modules/%40webex/internal-plugin-conversation/test/integration/spec/verbs.js#L190) block which is incorrect and must have slipped through previously.
 Could you move this describe block within the `#post()` describe block currently on line 248? Suggested structure:

```js
describe('#post()', () => {
...
  describe('when encryption key needs to rotate', () => {
    it('rotates the key and posts a plain text', () => {
```

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
adamweeks: Could we use the `this.get` function instead of creating an ad hoc request here?

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
dalipkumar703: Not really. will remove

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
dalipkumar703: get function take conversation as parameter but in failed request not having complete convo object.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
dalipkumar703: Okay, will do

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -896,14 +898,38 @@ const Conversation = WebexPlugin.extend({
         async: false
       });
     }
+    const cloneActivity = cloneDeepU(params);
```
adamweeks: Why not use the `cloneDeep` from lodash that is already imported?

dalipkumar703: lodash cloneDeep not able to copy scrObject function toJWE . It has some issue with custom object copying.

dalipkumar703: Though I am giving chance to lodash cloneDeepWith if that works I can remove it.

dalipkumar703: I added customizer function using cloneDeepWith that copy instance function, now not require additional library clone-deep.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -896,14 +898,56 @@ const Conversation = WebexPlugin.extend({
         async: false
       });
     }
+    /**
+   * helper function to cloneDeepWith for copying instance function
+   * @param {T} value
```
adamweeks: What type of object is `{T}`? It should be a javascript type or something defined in a `@typedef` statement.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -896,14 +898,56 @@ const Conversation = WebexPlugin.extend({
...
+   * @returns {null}
```
adamweeks: This function can return an object as well as a null, it needs to be defined here.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -896,14 +898,56 @@ const Conversation = WebexPlugin.extend({
...
+   */
+    // eslint-disable-next-line consistent-return
+    const customActivityCopy = (value) => {
+      const {files} = params.body.object;
```
adamweeks: Why are you using the externally scoped `params` object here? Isn't it being passed in as `value` via your `cloneDeepWith` function?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
dalipkumar703: T can be of any type string, object, Symbol etc. Let me see if there is other representation for generic datatype.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
dalipkumar703: Okay, will update here.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
dalipkumar703: Complete params is not pass every time in value. Value is a property that need to copy from params, it is a recursive operation over params but here i am adding condition to check if value match SCRObject then do copy of few more function that generally avoided. @adamweeks  Do you have better approach in mind to do same ?  Please suggest.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
```diff
@@ -208,14 +208,49 @@ describe('plugin-conversation', function () {
...
 
+
     describe('#post()', () => {
       let message, richMessage;
 
       beforeEach(() => {
         message = 'mccoy, THIS IS A TEST MESSAGE';
         richMessage = `<webex-mention data-object-id="${mccoy.id}" data-object-type="person">mccoy</webex-mention>, THIS IS A TEST MESSAGE`;
       });
+      describe('#post() with error handling', () => {
```
adamweeks:
```suggestion
      describe('when encryption key need to rotate', () => {
```

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
```diff
@@ -208,14 +208,49 @@ describe('plugin-conversation', function () {
...
+        beforeEach(() => {
+          if (convo) {
+            return Promise.resolve();
+          }
+
+          return webex.internal.conversation.create({participants})
+            .then((c) => { convo = c; });
+        });
 
+        it('post a plain text when encryption key need to rotate', () => {
```
adamweeks:
```suggestion
        it('can post a plain text successfully, () => {
```

adamweeks:
After making requested changes, please rebase on upsteam master so that the proper tests run.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/verbs.js
```diff
@@ -208,14 +208,49 @@ describe('plugin-conversation', function () {
...
 
       beforeEach(() => {
         message = 'mccoy, THIS IS A TEST MESSAGE';
         richMessage = `<webex-mention data-object-id="${mccoy.id}" data-object-type="person">mccoy</webex-mention>, THIS IS A TEST MESSAGE`;
       });
+      describe('when encryption key need to rotate', () => {
+        let convo;
+
+        beforeEach(() => {
+          if (convo) {
```
adamweeks: This block is not needed. If the convo object is defined at this point, we have an issue with our test runner or test setup. Please remove.

dalipkumar703: Okay. will remove

PR #1544, opened 2020-02-27
docs(contributing): add samples tests section

adamweeks:
Small documentation update on how samples tests are run and how to best run them locally.


CONTRIBUTING.md
```diff
@@ -177,6 +179,34 @@ To run tests on [SauceLabs](https://saucelabs.com/) locally, you'll need to add
...
+
+```js
+specs: [
+    './wdio.helpers.d/**/*.js',
+    './packages/node_modules/samples/browser-call-with-screenshare/test/wdio/spec/**/*.js'
+  ],
+```
+
+In this example, you would only run the screenshare test.
+Make sure to leave the helpers line in so that helpers get loaded properly.
```
taymoork2:
```suggestion
To run a specific sample test instead of the full suite, append the `--spec` flag to the `samples:test` command and the path to the specific test

```sh
npm run samples:test -- --spec packages/node_modules/samples/browser-call-with-screenshare
```

In this example, you would only run the screenshare test.
```

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -177,6 +179,34 @@ To run tests on [SauceLabs](https://saucelabs.com/) locally, you'll need to add
...
+
+##### Local Samples Tests
+
+If you wish to run the samples tests locally, we suggest changing from the Chrome-to-Firefox multiremote setup to Chrome-to-Chrome.
+
+You can do so by modifying the [wdio.conf.js](./wdio.conf.js) file.
+Simply change the `browserSpock`'s `desiredCapabilities` object to the same as `browserMccoy` (the Chrome instance).
```
taymoork2: Can we just rename these to Chrome and Firefox now...?
I always get confused which is which

taymoork2: ^in a later PR of course

CONTRIBUTING.md
adamweeks: img

CONTRIBUTING.md
adamweeks: Good idea! https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-124460

adamweeks:
@taymoork2 updated with suggested change

PR #1549, opened 2020-03-03
fix(meetings): use async instead of finally

adamweeks:
We’ve had customer reports that the meetings plugin wasn’t working with angular applications, and we were able to recreate the issue around the usage of finally. There might be a better babel config that works with angular, but consider this async conversion a “workaround” to get customers up and running.

I didn’t have full context on the workflow, but I tried to recreate it as best as possible without changing any of the tests. Let me know if I missed something.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
@@ -493,7 +493,7 @@ export default class Meetings extends WebexPlugin {
      * @private
      * @memberof Meetings
      */
-    createMeeting(destination, type = null) {
+    async createMeeting(destination, type = null) {
       const meeting = new Meeting(
         {
           userId: this.webex.internal.device.userId,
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1867,51 +1867,51 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      .finally((reconnect) => {
-        this.reconnectionManager.reset();
+      this.reconnectionManager.reset();
 
-        return Promise.resolve(reconnect);
-      });
+      throw new ReconnectionError('Reconnection failure event', error);
+    }
+
+    this.reconnectionManager.reset();
```
chrisadubois: can we not do a try/catch/finally with the await? that is supported across all browsers

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2648,39 +2648,40 @@ export default class Meeting extends StatelessWebexPlugin {
...
+          uri: content.url,
+          resourceUrl: this.resourceUrl
         });
+      }
+      catch (error) {
+        LoggerProxy.logger.error('Meeting/index->stopShare#Error ', error);
+        throw error;
+      }
+
+      this.isSharing = false;
```
chrisadubois: Same here, can we do a try catch finally?

Also shouldn't we still return a promise resolution/rejection as this is a user interaction call

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -214,7 +214,7 @@ export default class Meetings extends WebexPlugin {
             meeting.locusInfo.initialSetup(data.locus);
           }
         })
-          .finally(() => {
```
arun3528: Create sometime throws an error for the /meetingInfo . Irrespective of that we still send a meeting:added event so we need this in finally

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -508,52 +508,50 @@ export default class Meetings extends WebexPlugin {
 
       this.meetingCollection.set(meeting);
 
-      return this.meetingInfo
-        .fetchMeetingInfo(MeetingsUtil.extractDestination(destination, type), type)
-        .then((info) => {
-          meeting.parseMeetingInfo(info);
-          meeting.meetingInfo = info ? info.body : null;
+      try {
+        const info = await this.meetingInfo.fetchMeetingInfo(MeetingsUtil.extractDestination(destination, type), type);
```
arun3528: This is the function which triggers that

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: I will see if angular has the same issues with finally. This function didn’t return anything on the resolve, so it will behave the same due to it being asynchronous. I switched the promise rejection to a throw for the same reason

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -508,34 +508,28 @@ export default class Meetings extends WebexPlugin {
...
+        // if there is no meeting info we assume its a 1:1 call or wireless share
+        LoggerProxy.logger.info(`Meetings->createMeeting#Error ${err} fetching /meetingInfo for creation.`);
+        LoggerProxy.logger.info('Meetings->createMeeting#Info assuming this destination is a 1:1 or wireless share');
+        // We need to save this info for future reference
+        meeting.destination = destination;
+      }
+      finally {
+        // For type LOCUS_ID we need to parse the locus object to get the information
+        // about the caller and callee
+        // Meeting Added event will be created in `handleLocusEvent`
```
adamweeks: Added a note about where the event will be emitted.

adamweeks:
@chrisadubois @arun3528 I changed the scope on this to just the `createMeeting` function, since it is the only one causing an issue on angular. 

Will investigate others as issues arise, but I know the `stopShare` changes would be too impactful on the work that @taymoork2 is doing.

PR #1551, opened 2020-03-03
Feat/get org from jwt

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request are isolated to the ability to extract a user's OrgId from a provided JWT.

Fixes # [SPARK-125195](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-125195)

## Type of Change

## Test Coverage

- [x] Generated tests to validate the changes work
- [x] Ran the tests to validate the changes work as expected.

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
@@ -5,6 +5,7 @@
 import querystring from 'querystring';
 import url from 'url';
 
+import jwt from 'jsonwebtoken';
 import {
   base64,
   makeStateDataType,
@@ -127,6 +128,31 @@ const Credentials = WebexPlugin.extend({
 
     return `${this.config.authorizeUrl}?${querystring.stringify(options)}`;
     /* eslint-enable camelcase */
+  },
+
+  /**
+   * Extract the OrgId [realm] from a provided JWT.
+   *
+   * @private
+   * @param {string} token - The JWT to extract the OrgId from.
+   * @throws {Error} - If the token does not pass JWT general/realm validation.
+   * @returns {string} - The OrgId.
+   */
+  extractOrgIdFromJWT(token = '') {
+    // Decoded the provided token.
+    const decodedJWT = jwt.decode(token);
+
+    // Validate that the provided token is a JWT.
+    if (!decodedJWT) {
+      throw new Error('unable to extract the OrgId from the provided JWT');
+    }
+
+    if (!decodedJWT.realm) {
+      throw new Error('the provided JWT does not contain an OrgId');
+    }
+
+    // Return the OrgId [realm].
+    return decodedJWT.realm;
   },
 
   /**
--- a/packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
@@ -19,6 +19,26 @@ describe('webex-core', () => {
         user = u;
       }));
 
+    describe('#extractOrgIdFromJWT()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('generate a JWT and Webex Instance', () => {
+        webex = new WebexCore({
+          credentials: user.token
+        });
+
+        credentials = webex.credentials;
+      });
+
+      it('should return the OrgId of the provided JWT', () => {
+        // The access token in a client-auth scenario is a JWT.
+        const token = user.token.access_token;
+
+        assert.equal(credentials.extractOrgIdFromJWT(token), user.orgId);
+      });
+    });
+
     describe('#refresh()', () => {
       nodeOnly(it)('refreshes an access token', () => {
         const webex = new WebexCore({
--- a/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
@@ -228,6 +228,35 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#extractOrgIdFromJWT()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('generate webex and destructure credentials', () => {
+        webex = new MockWebex();
+        credentials = new Credentials(undefined, {parent: webex});
+      });
+
+      it('should return the OrgId of a provided JWT', () => {
+        const jwt = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJyZWFsbSI6Im15LXJlYWxtIn0.U16gzUsaRW1VVikJA2VeXRHPX716tG1_B42oxzy1UMk';
+        const realm = 'my-realm';
+
+        assert.equal(credentials.extractOrgIdFromJWT(jwt), realm);
+      });
+
+      it('should throw if the provided JWT is not valid', () =>
+        assert.throws(() => credentials.extractOrgIdFromJWT('not-valid')));
+
+      it('should throw if the provided JWT does not contain an OrgId', () => {
+        const jwtNoOrg = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c';
+
+        assert.throws(() => credentials.extractOrgIdFromJWT(jwtNoOrg));
+      });
+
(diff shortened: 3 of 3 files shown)


packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -129,6 +130,28 @@ const Credentials = WebexPlugin.extend({
...
+   *
+   * @param {string} token - The JWT to extract the OrgId from.
+   * @returns {string} - The OrgId.
+   */
+  extractOrgIdFromJWT(token = '') {
+    // Split the provided JWT into subsections.
+    const fields = token.split('.');
+
+    // Validate that the provided token is a JWT.
+    if (fields.length !== 3) {
```
lalli-flores: Is there another way to verify that something is a JWT? This check feels a little brittle (from my limited JWT knowledge)

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -129,6 +130,28 @@ const Credentials = WebexPlugin.extend({
...
+      return undefined;
```
lalli-flores: Should you return empty string `''` or `null`?

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -127,6 +128,38 @@ describe('webex-core', () => {
...
+          }
+        });
+
+        credentials = webex.credentials;
+
+        jwt = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJndWVzdC11c2VyLTczNDkiLCJuYW1lIjoiR3Vlc3QgVXNlciBEaXNwbGF5IE5hbWUiLCJpc3MiOiJZMmx6WTI5emNHRnlhem92TDNWekwwOVNSMEZPU1ZwQlZFbFBUaTg1Tm1GaVl6SmhZUzB6WkdOakxURXhaVFV0WVRFMU1pMW1aVE0wT0RFNVkyUmpPV0UiLCJleHAiOiIxNTExMjg2ODQ5In0.xGVs2k-JLGXxZyhMVVGZWlCf5Nu5nz_wT1vaHmH4Etg';
+
+        const fields = jwt.split('.');
+
+        jwtOrgId = JSON.parse(jose.util.base64url.decode(fields[1])).realm;
```
lalli-flores: Instead of re-writing the logic of what `extractOrgIdFromJWT` does, would it be beneficial to hardcode the expected value instead?

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
InteractiveTimmy: Needs to be null in this case, later we will do a:
```js
combinedFunctionForGettingOrgId(token = '') {
  const foundOrgId = extractFromJWT(token) || extractFromUserToken(token);

  if (!foundOrgId) {
    throw new Error('message');
  }
  ...
}

**Comment by InteractiveTimmy** [comment]
After further review, it is impossible to extract the`OrgId` from a JWT unless it was explicitly provided, which our documentation doesn't mention. This being said, we exchange JWT tokens for authorization tokens. I will let this PR go stale while I continue to investigate an alternative solution.

File: packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js (hunk 1 above)
Comment (InteractiveTimmy) [reply packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js]: In this case, the above works well due to how `node-jose` works. If the provided token has three `.` characters, `node-jose` will interpret it. If the interpreted object doesn't have a `realm` key:value`, it returns undefined [as intended].

Comment (InteractiveTimmy) [reply packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js]: In this case, the above works well due to how `node-jose` works. If the provided token has three `.` characters, `node-jose` will interpret it. If the interpreted object doesn't have a `realm` key:value`, it returns undefined [as intended].

File: packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js (hunk 3 above)
Comment (InteractiveTimmy) [reply packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js]: I moved this logic to the appropriate place. With integration test users, we get all the information we need to validate this method works as expected.

File: packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js (hunk 1 above)
Comment (adamweeks) [reply packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js]: Since we are using the `jsonwebtoken` package elsewhere in the SDK, let's use it's [decode](https://www.npmjs.com/package/jsonwebtoken#jwtdecodetoken--options) method.

File: packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js [hunk 4]
```diff
@@ -19,6 +19,32 @@ describe('webex-core', () => {
...
+      });
+
+      it('should return the OrgId of the provided JWT', () => {
+        // The access token in a client-auth scenario is a JWT.
+        const token = user.token.access_token;
+
+        assert.equal(credentials.extractOrgIdFromJWT(token), user.orgId);
+      });
+
+      it('should return undefined if the provided JWT is not valid', () =>
```
adamweeks: I do like the integration test above to verify the token in a client-auth scenario, but these tests feel like they would be better suited as unit tests.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
adamweeks: If the JWT is invalid, we should `throw` an error. If there's no org id in the token, return null.

packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
InteractiveTimmy: This has been moved.

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -228,6 +228,22 @@ describe('webex-core', () => {
...
+    describe('#extractOrgIdFromUserToken()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('generate webex and destructure credentials', () => {
+        webex = new MockWebex();
+        credentials = new Credentials(undefined, {parent: webex});
+      });
+
+      it('should return undefined if the provided JWT is not valid', () =>
```
adamweeks: I know we have the happy path in the integration tests, but could you add one here as well? 

```js
const jwt = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJyZWFsbSI6Im15LXJlYWxtIn0.U16gzUsaRW1VVikJA2VeXRHPX716tG1_B42oxzy1UMk';
const realm = 'my-realm';

assert.isEqual(credentials.extractOrgIdFromJWT(jwt), realm);
```

InteractiveTimmy: Implemented.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
InteractiveTimmy: Discussed this in a Webex Teams 1:1. Seems it was decided to leave this as is based on the parent method that calls these.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
InteractiveTimmy: addressed this.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
lalli-flores: Making decisions over 1:1 chats and giving that as an answer doesn't feel satisfactory, could you explain the reasoning for making such decision?
How a works function shouldn't be dictated by how something else wants to use it, what is right for the function should determined how it is used.

I don't think we should be returning `undefined` when is a purposely set value but that's my opinion

adamweeks: > Needs to be null/undefined [I prefer undefined] in this case, later we will do a:
> 
> ```js
> combinedFunctionForGettingOrgId(token = '') {
>   const foundOrgId = extractFromJWT(token) || extractFromUserToken(token);
> 
>   if (!foundOrgId) {
>     throw new Error('message');
>   }
>   ...
> }
> ```

@lalli-flores, @InteractiveTimmy went over his usage as described in the sample code above and I was convinced.

lalli-flores: @adamweeks thanks for the insight, I would appreciate if conversations on teams PRs happen on a team space instead of 1:1 spaces. If we want to move away from relying on one person to review everything, we should probably open discussions for the team to have an opportunity to engage

PR #1552, opened 2020-03-04
Feat/get org from user token

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request are isolated to the ability to extract a user's OrgId from a provided user token.

Fixes # [SPARK-125197](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-125197)

## Type of Change

## Test Coverage

- [x] Generated tests to validate the changes work
- [x] Ran the tests to validate the changes work as expected.

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
@@ -156,6 +156,27 @@ const Credentials = WebexPlugin.extend({
   },
 
   /**
+   * Extract the OrgId [realm] from a provided user token.
+   *
+   * @private
+   * @param {string} token - The user token to extract the OrgId from.
+   * @throws {Error} - Will throw an error if the provided token is invalid.
+   * @returns {string} - The OrgId.
+   */
+  extractOrgIdFromUserToken(token = '') {
+    // Split the provided token into subsections.
+    const fields = token.split('_');
+
+    // Validate that the provided token has the proper amount of sections.
+    if (fields.length !== 3) {
+      throw new Error('the provided token is not a valid format');
+    }
+
+    // Return the token section that contains the OrgId.
+    return fields[2];
+  },
+
+  /**
    * Generates a Logout URL
    * @instance
    * @memberof Credentials
--- a/packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
@@ -39,6 +39,26 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#extractOrgIdFromUserToken()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('define webex', () => {
+        webex = new WebexCore({
+          credentials: user.token
+        });
+
+        credentials = webex.credentials;
+      });
+
+      it('should return the OrgId when the provided token is valid', () => {
+        // The refresh token is formatted like a normal user token.
+        const token = user.token.refresh_token;
+
+        assert.equal(credentials.extractOrgIdFromUserToken(token), user.orgId);
+      });
+    });
+
     describe('#refresh()', () => {
       nodeOnly(it)('refreshes an access token', () => {
         const webex = new WebexCore({
--- a/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
+++ b/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
@@ -257,6 +257,29 @@ describe('webex-core', () => {
         assert.throws(() => credentials.extractOrgIdFromJWT()));
     });
 
+    describe('#extractOrgIdFromUserToken()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('generate webex and destructure credentials', () => {
+        webex = new MockWebex();
+        credentials = new Credentials(undefined, {parent: webex});
+      });
+
+      it('should return the OrgId of the provided user token', () => {
+        const orgId = 'this-is-an-org-id';
+        const userToken = `000_000_${orgId}`;
+
+        assert.equal(credentials.extractOrgIdFromUserToken(userToken), orgId);
+      });
+
+      it('should throw when provided an invalid token', () =>
+        assert.throws(() => credentials.extractOrgIdFromUserToken('invalid')));
+
+      it('should throw when no token is provided', () =>
+        assert.throws(() => credentials.extractOrgIdFromUserToken()));
+    });
+
     describe('#initialize()', () => {
       it('turns a portal auth string into a configuration object', () => {
         const webex = new MockWebex();


packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -129,6 +129,25 @@ const Credentials = WebexPlugin.extend({
...
+   * @param {string} token - The user token to extract the OrgId from.
+   * @returns {string} - The OrgId.
+   */
+  extractOrgIdFromUserToken(token = '') {
+    // Split the provided token into subsections.
+    const fields = token.split('_');
+
+    // Validate that the provided token has the proper amount of sections.
+    if (fields.length !== 3) {
+      return undefined;
```
adamweeks: This should throw when a token is invalid.

packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
```diff
@@ -19,6 +19,32 @@ describe('webex-core', () => {
...
+      });
+
+      it('should return the OrgId when the provided token is valid', () => {
+        // The refresh token is formatted like a normal user token.
+        const token = user.token.refresh_token;
+
+        assert.equal(credentials.extractOrgIdFromUserToken(token), user.orgId);
+      });
+
+      it('should return undefined when provided an invalid token', () =>
```
adamweeks: Same comment as your other PR, since these tests don't integrate with anything, they would be better suited as unit tests.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
InteractiveTimmy: Not in my implementation plans. See [this JIRA](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-125201) and [this comment](https://github.com/webex/webex-js-sdk/pull/1551#discussion_r387710737)

packages/node_modules/@webex/webex-core/test/integration/spec/credentials/credentials.js
InteractiveTimmy: This has been moved.

packages/node_modules/@webex/webex-core/src/lib/credentials/credentials.js
InteractiveTimmy: This was discussed in a 1:1 chat. It was decided to leave this method as is for another implementation to consume.

lalli-flores: Making decisions over 1:1 chats and giving that as an answer doesn't feel satisfactory, could you explain the reasoning for making such decision?
I agree with @adamweeks that if something is not valid, it should throw an exception, and whatever consumes it should handle the exception how it sees fit

InteractiveTimmy: Here's a larger example showing the logic needed to implement the planned subtask if we were to throw errors on each of the extraction methods and if we did not throw errors on each of the extraction methods:

```js
class credentials {
  /* ... */
  // Method extracs the OrgId from a JWT.
  // This could be marked private.
  extractOrgIdFromJWT(token = '') {}

  // Method extracs the OrgId from a user token.
  // This could be marked private.
  extractOrgIdFromUserToken(token = '') {}

  // Method gets the OrgId if determinable. Public method.
  determineOrgId() {
    const orgId = this.orgId ||
      this.extractOrgIdFromJWT(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.refresh_token);
      /** other ways to get the OrgID from credentials can be amended */

    if (!orgId) {
      throw new Error('excplicit error message');
    }

    return orgId;
  }

  determineOrgIdIfExtractorsThrowErrors() {
    if (this.orgId) {
      return this.orgId;
    }

    try {
      return this.extractOrgIdFromJWT(this.credentials.access_token);
    }
    catch {
      try {
        return this.extractOrgIdFromUserToken(this.credentials.access_token);
      }
      catch {
        try {
          return this.extractOrgIdFromUserToken(this.credentials.refresh_token);
        }
        catch {
          throw new Error('explicit error message');
        }
      }
    }
  }
  /* ... */
}
```

My biggest concern is the excessive nesting of these try/catch blocks. I may be mistaken, is there a better way to handle calling multiple error-throwing methods in the case that you are expecting them to throw a majority of the time?

InteractiveTimmy: Here's a larger example showing the logic needed to implement the planned subtask if we were to throw errors on each of the extraction methods and if we did not throw errors on each of the extraction methods:

```js
class credentials {
  /* ... */
  // Method extracs the OrgId from a JWT.
  // This could be marked private.
  extractOrgIdFromJWT(token = '') {}

  // Method extracs the OrgId from a user token.
  // This could be marked private.
  extractOrgIdFromUserToken(token = '') {}

  // Method gets the OrgId if determinable. Public method.
  determineOrgId() {
    const orgId = this.orgId ||
      this.extractOrgIdFromJWT(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.refresh_token);
      /** other ways to get the OrgID from credentials can be amended */

    if (!orgId) {
      throw new Error('excplicit error message');
    }

    return orgId;
  }

  determineOrgIdIfExtractorsThrowErrors() {
    if (this.orgId) {
      return this.orgId;
    }

    try {
      return this.extractOrgIdFromJWT(this.credentials.access_token);
    }
    catch {
      try {
        return this.extractOrgIdFromUserToken(this.credentials.access_token);
      }
      catch {
        try {
          return this.extractOrgIdFromUserToken(this.credentials.refresh_token);
        }
        catch {
          throw new Error('explicit error message');
        }
      }
    }
  }
  /* ... */
}
```

My biggest concern is the excessive nesting of these try/catch blocks. I may be mistaken, is there a better way to handle calling multiple error-throwing methods in the case that you are expecting them to throw a majority of the time?

InteractiveTimmy: Alternatively, this could just be linear try-catch statements since return is called in each:
```js
try {
  return this.extractOrgIdFromJWT(this.credentials.access_token);
} catch {} // do nothing
/* repeat */
```

lalli-flores: @InteractiveTimmy, you don't need to nest the `try-catch` for each type of token, they can all be under one. In case of failure, we can raise the error and let the developer know what went wrong, instead of failing silently.
Code would be something like

``js
determineOrgId() {
    const orgId = null;

    try {
      orgId = this.orgId ||
      this.extractOrgIdFromJWT(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.access_token) ||
      this.extractOrgIdFromUserToken(this.credentials.refresh_token);
   catch() {
       ...
   }

InteractiveTimmy: In the above block, if `this.extractOrgIdFromJWT(this.credentials.access_token)` throws an error, does it continue to the next line: `this.extractOrgIdFromUserToken(this.credentials.access_token)` or continue to the `catch` statement?

lalli-flores: This is my suggestion 🤷🏻‍♀️, but do what aligns with the code design everywhere, or is more readable (although the latter can be a bit subjective)

lalli-flores: No, you're right, this wouldn't work in this case. Once it gets an error the control flow will go into the `catch`

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -228,6 +228,22 @@ describe('webex-core', () => {
...
+    describe('#extractOrgIdFromUserToken()', () => {
+      let credentials;
+      let webex;
+
+      beforeEach('generate webex and destructure credentials', () => {
+        webex = new MockWebex();
+        credentials = new Credentials(undefined, {parent: webex});
+      });
+
+      it('should throw when provided an invalid token', () =>
```
lalli-flores: I think the "good" case is missing, when the orgID is actually returned

InteractiveTimmy: I did not implement a **good** case for the unit test, but it exists in the integration test [as it requires a valid user token].

lalli-flores: Thanks for adding it!
The value and I see in having a unit test, in addition to the integration is that it allows for developers that didn't write the code to see at glance what expected return values look like. For instance, now I know that the org ID returned is an base64 org ID value and not a name or other type of ID

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -228,6 +228,29 @@ describe('webex-core', () => {
...
+      let credentials;
+      let webex;
+
+      beforeEach('generate webex and destructure credentials', () => {
+        webex = new MockWebex();
+        credentials = new Credentials(undefined, {parent: webex});
+      });
+
+      it('should return the OrgId of the provided user token', () => {
+        const userToken = 'ZjY4NmE4YjgtZmNhMC00MTllLTllNmEtM2M2OGU4ZDg2MzFkMTJkNjVlZjMtYTc0_A52D_584cf4cd-eea7-4c8c-83ee-67d88fc6eab5';
```
adamweeks: This was not a valid token was it?

adamweeks:
Looks like you're getting some unit test failures, could you investigate?

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
InteractiveTimmy: This was an old integration test user token that has expired.

InteractiveTimmy:
> Looks like you're getting some unit test failures, could you investigate?

Seems to be working now, must have been a hickup on circle since the error was related to something in the circle's scripts per the logs.

packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js
adamweeks: Might be beneficial to change the token a bit to one that was never issued anywhere. Just as a safeguard.

PR #1553, opened 2020-03-04
test(samples): mark remote share flaky

adamweeks:
Mark the samples remote screen share tests as flaky as they are consistently failing.

I've created a JIRA for a deeper investigation on the failure reason, but they need to be skipped to stop the blocking of the pipeline for other PRs.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-123941


packages/node_modules/samples/browser-call-with-screenshare/test/wdio/spec/normal-dialing.js
```diff
@@ -79,13 +80,13 @@ describe('samples', () => {
...
 
-      it('starts screensharing remote', () => {
+      flaky(it, process.env.SKIP_FLAKY_TESTS)('starts screensharing remote', () => {
         browserMccoy.click('button[title="share screen"]');
 
         browserMccoy.waitUntil(() => (browserMccoy.getText('#screenshare-tracks') === 'SHARING'), 10000, 'Timed-out waiting for screenshare to start');
       });
 
-      it('stops screensharing remote', () => {
+      flaky(it, process.env.SKIP_FLAKY_TESTS)('stops screensharing remote', () => {
```
taymoork2: I’d only mark this one as flaky, as this is the test that is consistently failing.
The screen does share, however it doesn’t close, and we should at least test that acreensharing works on chrome

PR #1557, opened 2020-03-06
Feat/add orgid to catalog requests

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request are to implement the new credentials orgId retrieval method into the catalog requests that occur at webex init.

Fixes # [SPARK-125198](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-125198)

## Type of Change

## Test Coverage

- [x] Generated new tests.
- [x] Ran tests to validate that the changes work as expected.

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/plugin-authorization-browser/src/authorization.js
+++ b/packages/node_modules/@webex/plugin-authorization-browser/src/authorization.js
@@ -209,7 +209,8 @@ const Authorization = WebexPlugin.extend({
         this.webex.credentials.set({
           supertoken: token
         });
-      });
+      })
+      .then(() => this.webex.internal.services.initServiceCatalogs());
   },
 
   /**
--- a/packages/node_modules/@webex/plugin-authorization-browser/test/integration/spec/authorization.js
+++ b/packages/node_modules/@webex/plugin-authorization-browser/test/integration/spec/authorization.js
@@ -8,6 +8,7 @@ import {assert} from '@webex/test-helper-chai';
 import {createUser} from '@webex/test-helper-appid';
 import WebexCore from '@webex/webex-core';
 import uuid from 'uuid';
+import sinon from 'sinon';
 
 browserOnly(describe)('plugin-authorization-browser', () => {
   describe('Authorization', () => {
@@ -24,9 +25,21 @@ browserOnly(describe)('plugin-authorization-browser', () => {
               .then(() => assert.isTrue(webex.canAuthorize));
           });
       });
+
+      it('should call services#initServiceCatalogs()', () => {
+        const webex = new WebexCore();
+        const userId = uuid.v4();
+        const displayName = `test-${userId}`;
+
+        webex.internal.services.initServiceCatalogs = sinon.spy();
+
+        return createUser({displayName, userId})
+          .then(({jwt}) => webex.authorization.requestAccessTokenFromJwt({jwt}))
+          .then(() => assert.called(webex.internal.services.initServiceCatalogs));
+      });
     });
 
-    describe('\'#refresh', () => {
+    describe.skip('\'#refresh', () => {
       describe('when used with an appid access token', () => {
         it('refreshes the access token', () => {
           const userId = uuid.v4();
--- a/packages/node_modules/@webex/plugin-authorization-node/test/integration/spec/authorization.js
+++ b/packages/node_modules/@webex/plugin-authorization-node/test/integration/spec/authorization.js
@@ -49,9 +49,21 @@ nodeOnly(describe)('plugin-authorization-node', () => {
               .then(() => assert.isTrue(webex.canAuthorize));
           });
       });
+
+      it('should call services#initServiceCatalogs()', () => {
+        const webex = new WebexCore();
+        const userId = uuid.v4();
+        const displayName = `test-${userId}`;
+
+        webex.internal.services.initServiceCatalogs = sinon.spy();
+
+        return createUser({displayName, userId})
+          .then(({jwt}) => webex.authorization.requestAccessTokenFromJwt({jwt}))
+          .then(() => assert.called(webex.internal.services.initServiceCatalogs));
+      });
     });
 
-    describe('\'#refresh', () => {
+    describe.skip('\'#refresh', () => {
       describe('when used with an appid access token', () => {
         it('refreshes the access token', () => {
           const userId = uuid.v4();
--- a/packages/node_modules/@webex/webex-core/src/lib/services/services.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/services.js
@@ -367,16 +367,18 @@ const Services = WebexPlugin.extend({
 
   /**
    * simplified method to update the preauth catalog via email
-   * @param {object} param
-   * @param {string} param.email - must be a standard-format email
+   *
+   * @param {object} query
+   * @param {string} query.email - A standard format email.
+   * @param {string} query.orgId - The user's OrgId.
    * @returns {Promise<void>}
    */
-  collectPreauthCatalog({email} = {}) {
-    if (!email) {
+  collectPreauthCatalog(query) {
+    if (!query) {
       return this.updateServices({from: 'limited', query: {mode: 'DEFAULT_BY_PROXIMITY'}});
     }
 
-    return this.updateServices({from: 'limited', query: {email}});
+    return this.updateServices({from: 'limited', query});
   },
 
   /**
@@ -433,7 +435,9 @@ const Services = WebexPlugin.extend({
       this.webex.credentials.canAuthorize &&
       !catalog.status.postauth.collecting
     ) {
-      return this.updateServices();
+      // Only update services if the preauth catalog is ready.
+      return catalog.waitForCatalog('preauth', timeout, () =>
+        this.updateServices());
     }
 
     return catalog.waitForCatalog(serviceGroup, timeout);
@@ -685,6 +689,75 @@ const Services = WebexPlugin.extend({
   },
 
   /**
+   * Initialize the discovery services and the whitelisted services.
+   *
+   * @returns {void}
+   */
+  initConfig() {
+    // Get the catalog and destructure the services config.
+    const catalog = this._getCatalog();
+    const {services} = this.webex.config;
+
+    // Validate that the services configuration exists.
+    if (services) {
+      // Check for discovery services.
+      if (services.discovery) {
+        // Format the discovery configuration into an injectable array.
+        const formattedDiscoveryServices = Object.keys(services.discovery)
+          .map((key) => ({
+            name: key,
+            defaultUrl: services.discovery[key]
+          }));
+
+        // Inject formatted services into services catalog.
+        catalog.updateServiceUrls('discovery', formattedDiscoveryServices);
+      }
+
+      // Check for allowed host domains.
+      if (services.allowedDomains) {
+        // Store the allowed domains as a property of the catalog.
+        catalog.setAllowedDomains(services.allowedDomains);
+      }
+
+      // Set `validateDomains` property to match configuration
+      this.validateDomains = services.validateDomains;
+    }
+  },
+
+  /**
+   * Make the initial requests to collect the root catalogs.
+   *
+   * @returns {Promise<void, Error>} - Errors if the token is unavailable.
+   */
+  initServiceCatalogs() {
+    this.logger.info('services: initializing initial service catalogs');
+
+    // Destructure the credentials plugin.
+    const {credentials} = this.webex;
+
+    // Init a promise chain. Must be done as a Promise.resolve() to allow
+    // credentials#getOrgId() to properly throw.
+    return Promise.resolve()
+      // Get the user's OrgId.
+      .then(() => credentials.getOrgId())
+      // Begin collecting the preauth/limited catalog.
+      .then((orgId) => this.collectPreauthCatalog({orgId}))
+      .then(() => {
+        // Validate if the token is authorized.
+        if (credentials.canAuthorize) {
+          // Attempt to collect the postauth catalog.
+          return this.updateServices()
+            .catch(() => this.logger.warn(
+              'services: cannot retrieve postauth catalog'
+            ));
+        }
+
+        // Return a resolved promise for consistent return value.
+        return Promise.resolve();
+      });
+  },
+
+  /**
    * Initializer
    *
    * @instance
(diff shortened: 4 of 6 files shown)


packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -684,6 +688,74 @@ const Services = WebexPlugin.extend({
...
+        if (credentials.canAuthorize) {
+          // Attempt to collect the postauth catalog.
+          return this.updateServices()
+            .catch(() => this.logger.warn(
+              'services: cannot retrieve postauth catalog'
+            ));
+        }
+
+        // Return a resolved promise for consistent return value.
+        return Promise.resolve();
```
adamweeks: Is eslint throwing an error here? By not returning anything at the end of the `then` it is the same as returning an empty resolved promise.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -702,50 +774,16 @@ const Services = WebexPlugin.extend({
...
-          .catch(
-            () => this.logger.warn('services: catalog retrieval failed w/auth')
-          );
-      }
-      else {
-        this.collectPreauthCatalog({email: this.webex.config.email});
-      }
+      this.initServiceCatalogs()
+        .catch((error) => this.logger.info(
+          `services: failed to init initial services, ${error.message}`
```
adamweeks: One thing I'd like to think about especially for v2 is a way to communicate errors with the developer outside of just console logging.  Something like `this.webex.emit('error')`.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -702,50 +774,16 @@ const Services = WebexPlugin.extend({
...
-        // primarily in unit testing with the karma suite.
-          .catch(
-            () => this.logger.warn('services: catalog retrieval failed w/auth')
-          );
-      }
-      else {
-        this.collectPreauthCatalog({email: this.webex.config.email});
-      }
+      this.initServiceCatalogs()
+        .catch((error) => this.logger.info(
```
adamweeks: This should be an error that users will need to read to understand why the sdk isn't working.
```suggestion
        .catch((error) => this.logger.error(
```

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: Correct, ESLint states that it needs consistent return values.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: I agree. Along with the larger list of things v2 will have, this is among the highest priority to me.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -684,6 +688,74 @@ const Services = WebexPlugin.extend({
...
+  initServiceCatalogs() {
+    this.logger.info('services: initializing initial service catalogs');
+
+    // Destructure the credentials plugin.
+    const {credentials} = this.webex;
+
+    // Init a promise chain to catch errors.
+    return Promise.resolve()
+      // Get the user's OrgId.
+      .then(() => credentials.getOrgId())
```
lalli-flores: Should you have a `catch` since `getOrgId` might throw an error? or do you want to just let it propagate?

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: addressed

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -684,6 +688,74 @@ const Services = WebexPlugin.extend({
...
+      // Begin collecting the preauth/limited catalog.
+      .then((orgId) => this.collectPreauthCatalog({orgId}))
+      .then(() => {
+        // Validate if the token is authorized.
+        if (credentials.canAuthorize) {
+          // Attempt to collect the postauth catalog.
+          return this.updateServices()
+            .catch(() => this.logger.warn(
```
lalli-flores: What else happens if the services can't be updated? could the client still proceed with whatever it had?

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -684,6 +688,74 @@ const Services = WebexPlugin.extend({
...
+
+    // Init a promise chain to catch errors.
+    return Promise.resolve()
+      // Get the user's OrgId.
+      .then(() => credentials.getOrgId())
+      // Begin collecting the preauth/limited catalog.
+      .then((orgId) => this.collectPreauthCatalog({orgId}))
+      .then(() => {
+        // Validate if the token is authorized.
+        if (credentials.canAuthorize) {
```
lalli-flores: What should happen if the credentials are not authorized?

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -367,16 +367,18 @@ const Services = WebexPlugin.extend({
...
-   * @param {string} param.email - must be a standard-format email
+   *
+   * @param {object} query
+   * @param {string} query.email - A standard format email.
+   * @param {string} query.orgId - The user's OrgId.
    * @returns {Promise<void>}
    */
-  collectPreauthCatalog({email} = {}) {
-    if (!email) {
+  collectPreauthCatalog(query) {
```
andrewholsted: This will be a breaking change for the web client(s). Have we may any further progress of not releasing everything as a minor/patch?

packages/node_modules/@webex/webex-core/src/lib/services/services.js
aimeex3: also what if credentials are not set yet? it won't be able to get an org id

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -684,6 +688,74 @@ const Services = WebexPlugin.extend({
...
+   * @returns {Promise<void, Error>} - Errors if the token is unavailable.
+   */
+  initServiceCatalogs() {
+    this.logger.info('services: initializing initial service catalogs');
+
+    // Destructure the credentials plugin.
+    const {credentials} = this.webex;
+
+    // Init a promise chain to catch errors.
+    return Promise.resolve()
```
aimeex3: this could be `return Promise.resolve(credentials.getOrgId())` so you don't need an empty resolve

packages/node_modules/@webex/webex-core/src/lib/services/services.js
aimeex3: @andrewholsted, i think it still supports email. it looks like this change is to support orgId too

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: @aimeex3, I've made this a pre-requisite for the initial event. This method will no longer be called automatically if there isn't a valid supertoken.

@lalli-flores, now that there's a required check before this is automatically triggered, in the case of this class method, the mentioned method will always return an orgId.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: Yes, if the token is not a fully authorized user token, it will not proceed with just the preauth catalog. This works in the cases of client auth, in which a user token may not be present at the time of `webex:ready`

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: The post-auth catalog should not be collected and then the developer could utilize `services#validateUser()` once the user has provided identification.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: @aimeex3 is correct. Let me know if this presents a problem during any testing.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: Good find, implementing this.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -702,49 +775,20 @@ const Services = WebexPlugin.extend({
...
-      if (this.webex.credentials.canAuthorize) {
-        this.updateServices()
-        // this catch prevents crashing in unique situations found
-        // primarily in unit testing with the karma suite.
-          .catch(
-            () => this.logger.warn('services: catalog retrieval failed w/auth')
-          );
-      }
-      else {
-        this.collectPreauthCatalog({email: this.webex.config.email});
```
aimeex3: is this part still needed? was it added for the guest devices client or was it for something else?

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -702,49 +775,20 @@ const Services = WebexPlugin.extend({
...
+      const {supertoken} = this.webex.credentials;
+
+      // Validate if the supertoken exists.
+      if (supertoken && supertoken.access_token) {
```
aimeex3: just curious would checking `credentials.canAuthorize` suffice? or do you need to check supertoken explicitly?

InteractiveTimmy: If we do not check for supertoken explicitly, it checks for a refresh token as well, which will fail the `credentials#getOrgId()` method.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
InteractiveTimmy: This was added a while ago as a feature that was never documented or utilized in the SDK. A developer would need to do the following to use this:
```js
const webex = new Webex({
  email: 'example@website.com'
});
```
Since this was never documented, let us know if you'd like us to re-implement this as a part of the PR.

aimeex3: it was not utilized by the web client. @robstax can you confirm if this is needed for guest client?

adamweeks:
How does this change impact guest users that are logged in via JWT as described here: https://developer.webex.com/docs/guest-issuer#using-guest-tokens

Is the `requestAccessTokenFromJwt` flow impacted?

InteractiveTimmy:
> How does this change impact guest users that are logged in via JWT as described here: https://developer.webex.com/docs/guest-issuer#using-guest-tokens
> 
> Is the `requestAccessTokenFromJwt` flow impacted?

It is, a fix has been implemented, but one test per authorization plugin has to be skipped due to a race between promises depending on each other.

packages/node_modules/@webex/plugin-authorization-browser/src/authorization.js
```diff
@@ -209,7 +209,8 @@ const Authorization = WebexPlugin.extend({
         this.webex.credentials.set({
           supertoken: token
         });
-      });
+      })
+      .then(() => this.webex.internal.services.initServiceCatalogs());
```
aimeex3: @andrewholsted is this what you were talking about? should this get called in plugin-authorization-browser-first-party as well? i believe first-party would be the web-client

packages/node_modules/@webex/plugin-authorization-browser/test/integration/spec/authorization.js
```diff
@@ -24,9 +25,21 @@ browserOnly(describe)('plugin-authorization-browser', () => {
...
+        webex.internal.services.initServiceCatalogs = sinon.spy();
+
+        return createUser({displayName, userId})
+          .then(({jwt}) => webex.authorization.requestAccessTokenFromJwt({jwt}))
+          .then(() => assert.called(webex.internal.services.initServiceCatalogs));
+      });
     });
 
-    describe('\'#refresh', () => {
+    describe.skip('\'#refresh', () => {
```
aimeex3: any reason these are skipped? seems like these should be fixed

packages/node_modules/@webex/plugin-authorization-node/test/integration/spec/authorization.js
```diff
@@ -49,9 +49,21 @@ nodeOnly(describe)('plugin-authorization-node', () => {
...
+        webex.internal.services.initServiceCatalogs = sinon.spy();
+
+        return createUser({displayName, userId})
+          .then(({jwt}) => webex.authorization.requestAccessTokenFromJwt({jwt}))
+          .then(() => assert.called(webex.internal.services.initServiceCatalogs));
+      });
     });
 
-    describe('\'#refresh', () => {
+    describe.skip('\'#refresh', () => {
```
aimeex3: same here

packages/node_modules/@webex/webex-core/src/lib/services/services.js
robstax: @InteractiveTimmy wanted to call it out. afaik guest anonymous tokens don't technically have an `orgId`, however it still does return a an `orgId`. i dunno if it's a placeholder, or real, or what, but something i wanted to call attention to

packages/node_modules/@webex/plugin-authorization-browser/test/integration/spec/authorization.js
InteractiveTimmy: This will be in a later PR, there's a race condition in which two promises wait for eachother:
`authorization#requestAccessTokenFromJwt()` calls `services#initServiceCatalogs()` requires `credentials#canAuthorize()` which is halted by `authorization#requestAccessTokenFromJwt()`.

Does this specific test impact the web client's auth flow?

packages/node_modules/@webex/plugin-authorization-node/test/integration/spec/authorization.js
InteractiveTimmy: See above comment.

packages/node_modules/@webex/plugin-authorization-browser/test/integration/spec/authorization.js
aimeex3: not this plugin - web client uses plugin-authorization-browser-first-party so if this changes also gets added there, we should make sure the refresh works as expected

InteractiveTimmy: The `Webex.refresh()` method works as expected. The only test within this describe that exists in both of these authorization plugins is:
```
#refresh => when used with an appId access token => refreshes the access token
```
This specifically calls the `requestAccessTokenFromJWT()` which doesn't appear to exist in the `plugin-authorization-browser-first-party` as it doesn't seem to be needed for web client flow [please let me know if this is a utilized method by the web client team]. If `authorization#requestAccessTokenFromJwt()` is not used at all within the web client, then the test case provided by this `describe()` block should not impact the web client.

Please let me know if the `it()` block within the `describe('#refresh')` test describes a flow used by the client and I'll work towards implementing the required flow.

PR #1559, opened 2020-03-09
fix(ediscovery): handle reportids that can be either uuids or urls

gannons:
# Pull Request Template

## Description

As part of federation work we need ediscovery.js to support reaching out to remote (i.e. None home cluster) eDiscovery servers. There is code to support this already. But we discovered that it has problems with the decryption of space activities. Namely the code

 * assumed all requests would have a `resource` property and failed when they did not
 * tried to retrieve a remote space id from the home cluster eDiscovery (which cannot find it)
 * wiped the space summary cache when the remote report summaries were also cached

Fixes # (issue)

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

I'm removing APIs which could breaking existing clients. But as we control the only client of eDiscovery this is not a concern.

## Test Coverage

Introduced appropriate tests and updated existing tests. I also integration this version of ediscovery.js into a local build of the eDiscovery Download Manager.

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
+++ b/packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
@@ -1,9 +1,10 @@
 /* eslint-disable prefer-template */
+/* eslint-disable no-useless-escape */
 /*!
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
 import {waitForValue, WebexPlugin} from '@webex/webex-core';
-import {oneFlight, tap} from '@webex/common';
+import {oneFlight} from '@webex/common';
 
 import {InvalidEmailAddressError} from './ediscovery-error';
 
@@ -187,7 +188,7 @@ const EDiscovery = WebexPlugin.extend({
   @oneFlight({keyFactory: (reportId, options) => createOneFlightKey(reportId, options)})
   /**
    * Retrieves content associated with a report
-   * @param {UUID} reportId - Id of the report which contains the content
+   * @param {UUID|string} reportId - UUID or url of the report which contains the content
    * @param {Object} options - optional parameters for this method
    * @param {number} options.offset - start position from which to retrieve records
    * @param {number} options.size - the number of records to retrieve
@@ -202,44 +203,11 @@ const EDiscovery = WebexPlugin.extend({
     // use spread operator to set default options
     options = {...this.config.defaultOptions, ...options};
 
-    return this.request({
-      method: 'GET',
-      service: 'ediscovery',
-      resource: `reports/${reportId}/contents`,
-      qs: {offset: options.offset, size: options.size},
-      timeout: options.timeoutMs
-    });
-  },
-
-  @waitForValue('@')
-  @oneFlight({keyFactory: (reportUrl, options) => createOneFlightKey(reportUrl, options)})
-  /**
-   * Retrieves content associated with a report
-   * @param {String} reportUrl - full destination address (including report id parameter) for the http request being sent
-   * e.g. 'http://ediscovery-intb.wbx2.com/ediscovery/api/v1/reports/3b10e625-2bd5-4efa-b866-58d6c93c505c'
-   * @param {Object} options - optional parameters for this method
-   * @param {number} options.offset - start position from which to retrieve records
-   * @param {number} options.size - the number of records to retrieve
-   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 30s
-   * @returns {Promise<ResponseEntity<[Activity]>>} Http response containing the activities
-   */
-  getRemoteContent(reportUrl, options) {
-    if (!reportUrl) {
-      throw Error('Undefined parameter');
-    }
-
-    const reportId = this._getReportIdFromUrl(reportUrl);
-
-    if (!reportId) {
-      throw Error('Report url does not contain a report id');
-    }
-
-    // use spread operator to set default options
-    options = {...this.config.defaultOptions, ...options};
+    const [requestOptions] = this._createRequestOptions(reportId, '/contents');
 
     return this.request({
       method: 'GET',
-      url: `${reportUrl}/contents`,
+      ...requestOptions,
       qs: {offset: options.offset, size: options.size},
       timeout: options.timeoutMs
     });
@@ -249,7 +217,7 @@ const EDiscovery = WebexPlugin.extend({
   @oneFlight({keyFactory: (reportId, options) => createOneFlightKey(reportId, options)})
   /**
    * Retrieves a list of the spaces relevant to a specified report
-   * @param {UUID} reportId - Id of the report being requested
+   * @param {UUID|string} reportId - UUID or url of the report being requested
    * @param {Object} options - optional parameters for this method
    * @param {number} options.offset - start position from which to retrieve records
    * @param {number} options.size - the number of records to retrieve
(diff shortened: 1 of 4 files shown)


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -3,7 +3,7 @@
  * Copyright (c) 2015-2019 Cisco Systems, Inc. See LICENSE file.
  */
 import {waitForValue, WebexPlugin} from '@webex/webex-core';
-import {oneFlight, tap} from '@webex/common';
```
gannons: No longer required. Not sure why `tap` was used instead of the more straight forward _older_ APIs.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -202,44 +202,11 @@ const EDiscovery = WebexPlugin.extend({
...
-   * Retrieves content associated with a report
-   * @param {String} reportUrl - full destination address (including report id parameter) for the http request being sent
-   * e.g. 'http://ediscovery-intb.wbx2.com/ediscovery/api/v1/reports/3b10e625-2bd5-4efa-b866-58d6c93c505c'
-   * @param {Object} options - optional parameters for this method
-   * @param {number} options.offset - start position from which to retrieve records
-   * @param {number} options.size - the number of records to retrieve
-   * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 30s
-   * @returns {Promise<ResponseEntity<[Activity]>>} Http response containing the activities
-   */
-  getRemoteContent(reportUrl, options) {
```
gannons: Removed in favour of overloading the `getContent(..)` API to support either a UUID or url. Similar approach was used for the other 2 APIs.

The reason for this is explained later on. But it also removes duplication as these methods are essentially equivalent.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -335,64 +264,34 @@ const EDiscovery = WebexPlugin.extend({
...
-    this.logger.warn(`Cache miss for space ${spaceId} in getContentSummaryBySpaceId`);
-
-    return this.request({
-      method: 'GET',
-      url: `${reportUrl}/contents/summary/${spaceId}`,
-      timeout: options.timeoutMs
-    })
-      .then(tap((res) => this._writeSpacesToContentSummaryBySpaceIdCache(reportId, [res.body])));
+  clearCache() {
+    this.contentSummaryBySpaceIdCache.clear();
```
gannons: This new API will need to be called to clear down the cache. Concern here is that reports with a large number of spaces could exhaust the app's RAM. But the largest reports we've come across will have < 10,000 spaces so it should not be an immediate concern. Still better to handle rather than let a bug/crash creep in.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -449,6 +345,57 @@ const EDiscovery = WebexPlugin.extend({
     return '';
   },
 
+  _isUrl(string) {
+    try {
+      const validUrl = new URL(string);
+
+      return validUrl !== undefined;
```
gannons: Looked at using require('url') as Adam mentioned. But it took me by surprise as

    console.log(url.parse('16bf0d01-b1f7-483b-89a2-915144158fb9'))

generated

    Url {
      protocol: null,
      slashes: null,
      auth: null,
      host: null,
      port: null,
      hostname: null,
      hash: null,
      search: null,
      query: null,
      pathname: '16bf0d01-b1f7-483b-89a2-915144158fb9',
      path: '16bf0d01-b1f7-483b-89a2-915144158fb9',
      href: '16bf0d01-b1f7-483b-89a2-915144158fb9' }

Felt that URL was more direct as it throws a constructor exception. Which I've gotten used to but I recall that seem unusual originally.

    validUrl !== undefined;

To calm eslint :(. It does not like 

    new URL(string)

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -449,6 +345,57 @@ const EDiscovery = WebexPlugin.extend({
...
+    }
+    catch (_) {
```
gannons: I'm surprised eslint lets this through

packages/node_modules/@webex/internal-plugin-ediscovery/src/index.js
```diff
@@ -131,11 +131,9 @@ registerInternalPlugin('ediscovery', EDiscovery, {
           if (!object || !object.body) {
             return Promise.resolve();
           }
-          // the report id is needed in order to retrieve space information for activities with whiteboards so that
-          // the list of users in those spaces can be accessed and the content can be decrypted on their behalf
-          const reportId = object.options.resource.split('/')[1];
 
-          return Promise.all(object.body.map((item) => ctx.transform('decryptReportContent', {body: item}, reportId)));
+          // Always use the report url as this'll resolve correctly for remote reports
+          return Promise.all(object.body.map((item) => ctx.transform('decryptReportContent', {body: item}, object.options.uri)));
```
gannons: This transform calls `getContentSummaryBySpaceId(..)`. By having one method it can always use a `uri` rather than choose between the previous method and `getRemoteContentSummaryBySpaceId(..)`.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -449,6 +346,50 @@ const EDiscovery = WebexPlugin.extend({
...
+    const isUrl = this._isUrl(reportId);
+
+    if (isUrl) {
+      reportUUID = this._getReportIdFromUrl(reportId);
+
+      if (!reportUUID) {
+        throw Error('Report url does not contain a report id');
+      }
+
+      // Ensure the url is formated to
```
taymoork2:
```suggestion
      // Ensure the url is formatted to
```

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -449,6 +346,50 @@ const EDiscovery = WebexPlugin.extend({
...
+      // https://ediscovery.intb1.ciscospark.com/ediscovery/api/v1/reports/16bf0d01-b1f7-483b-89a2-915144158fb9
+      // index.js for example passes the url in as
+      // https://ediscovery.intb1.ciscospark.com/ediscovery/api/v1/reports/16bf0d01-b1f7-483b-89a2-915144158fb9/contents
+      const reportUrl = reportId.substring(0, reportId.lastIndexOf(reportUUID) + reportUUID.length);
+
+      requestOptions = {
+        url: reportUrl + resource
```
taymoork2: I'd move the substring function in here, but the comment is so large that keeping it separate is probably best

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
gannons: It's necessary for this change.

**tldr:** We use a space summary cache for performance. The previous code wiped this cache when the report id changed. In a federated report there can be multiple report ids meaning our cache was wiped when it should not have been.

With the upcoming federation support we're going to run a **local cluster** eDiscovery report. If it finds remote spaces it'll run, potentially several, **remote cluster** reports. But we hide these details behind a single report for the end user.

This change means that when you download a "single" report from ControlHub it can, under the covers, be made up of multiple report ids. Before there was only **one** report id. We took advantage of this and used it to manage our cache.

Namely when we detected the user was retrieving content for a new report id we cleared the existing cache. i.e. As the user was likely finished with those space summaries. But this no longer works with multiple report ids. Hence I'm making the call explicit.

We tend to `pre-populate` the cache for each report. That's important as it prevents the app from hammering the eDiscovery `getContentSummaryBySpaceId(..)` API. With federation we `pre-populate` all **local** and **remote** report spaces. But, as we clear when a new report id is detected, the code ended up only being populated with the details of the _last_ **remote report** id. To compensate for the missing cache details calls were being made to `getContentSummaryBySpaceId(..)`.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -408,9 +308,6 @@ const EDiscovery = WebexPlugin.extend({
     }
 
     if (!this.contentSummaryBySpaceIdCache.has(reportId)) {
-      // wipe the cache - we will only need to store one reports data at any one time
-      this.contentSummaryBySpaceIdCache.clear();
```
gannons: This is the cache clear step that was removed

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -449,6 +346,50 @@ const EDiscovery = WebexPlugin.extend({
     return '';
   },
 
+  _isUrl(string) {
+    return /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&\/\/=]*)/g.test(string);
```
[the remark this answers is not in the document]
sregger: I tried that yesterday but you still get caught out by `\/\/` as changing it to `//` is a build error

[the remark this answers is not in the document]
sregger: > Can you also write a couple test cases that handle UUID

Those tests already exist and as such don't show up in this PR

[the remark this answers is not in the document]
sregger: 350:100  error    Unnecessary escape character: \+                                         no-useless-escape
    350:107  error    Unnecessary escape character: \/                                         no-useless-escape
    350:109  error    Unnecessary escape character: \/                                         no-useless-escape

PR #1600, opened 2020-04-03
Meetings set local video quality

adamweeks:
Adds the ability to set the local video quality during a meeting.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/common/errors/parameter.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/common/errors/parameter.js
@@ -12,10 +12,15 @@ export default class ParameterError extends Error {
   */
   constructor(message = ERROR_DICTIONARY.PARAMETER.MESSAGE, error = null) {
     super(message);
+
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, ParameterError);
+    }
+
     this.name = ERROR_DICTIONARY.PARAMETER.NAME;
     this.sdkMessage = ERROR_DICTIONARY.PARAMETER.MESSAGE;
     this.error = error;
-    this.stack = error ? error.stack : (new Error()).stack;
+
     this.code = ERROR_DICTIONARY.PARAMETER.CODE;
   }
 }
--- a/packages/node_modules/@webex/plugin-meetings/src/constants.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/constants.js
@@ -962,3 +962,52 @@ export const MQA_STATS = {
     systemMaximumCPU: 0
   }
 };
+
+// ****** MEDIA QUALITY CONSTANTS ****** //
+
+export const QUALITY_LEVELS = {
+  LOW: 'LOW',
+  MEDIUM: 'MEDIUM',
+  HIGH: 'HIGH'
+};
+
+export const VIDEO_RESOLUTIONS = {
+  [QUALITY_LEVELS.LOW]: {
+    video: {
+      width: {
+        max: 320,
+        ideal: 320
+      },
+      height: {
+        max: 180,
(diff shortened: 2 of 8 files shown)


packages/node_modules/@webex/plugin-meetings/src/constants.js
```diff
@@ -962,3 +962,44 @@ export const MQA_STATS = {
     systemMaximumCPU: 0
   }
 };
+
+export const VIDEO_RESOLUTIONS = {
```
arun3528, line 966: Should we have a config for what max resolution we should receive

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2773,4 +2774,35 @@ export default class Meeting extends StatelessWebexPlugin {
...
   }
+
+  /**
+   * Sets the quality of the local video stream
+   * @param {String} level {LOW|MEDIUM|HIGH}
+   * @returns {Promise}
+   */
+  setLocalVideoQuality(level) {
+    LoggerProxy.logger.log(`Meeting:index#setLocalVideoQuality --> Setting quality to ${level}`);
+
```
arun3528, line 2785: add a error check if sendVideo is false

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2773,4 +2774,35 @@ export default class Meeting extends StatelessWebexPlugin {
 
     return reject('Meeting:index#sendDTMF --> cannot send DTMF, meeting does not have a connection to the "locus" call control service. Have you joined?');
   }
+
+  /**
+   * Sets the quality of the local video stream
+   * @param {String} level {LOW|MEDIUM|HIGH}
+   * @returns {Promise}
+   */
+  setLocalVideoQuality(level) {
```
arun3528, line 2783: Might need to rethink about this function as its doing nothing other then getUserMedia with lower resolution then doing updateMedia 

Do we still need to have different function to downgrade incoming video and keep it independent of this

adamweeks: The scope of this PR is limited to just the resolution of the sending media. We will have another for the receive media resolution.

packages/node_modules/@webex/plugin-meetings/src/constants.js
adamweeks: All of these configs have max resolutions set, so I'm not sure what you are asking.

arun3528: Was talking about default config . If someone wants to start a meeting by default with lower resolution or video

adamweeks: Good idea! I've created another task for changing settings pre-join as this one is scope towards during call changes. https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-133817

packages/node_modules/@webex/plugin-meetings/src/common/errors/parameter.js
```diff
@@ -12,10 +12,15 @@ export default class ParameterError extends Error {
   */
   constructor(message = ERROR_DICTIONARY.PARAMETER.MESSAGE, error = null) {
     super(message);
+
+    if (Error.captureStackTrace) {
```
lalli-flores, line 16: Can this change be in its own commit? since it doesn't relate to local video quality changes?

adamweeks: Yes, this was actually from another branch that got in here somehow during a rebase.

PR #1601, opened 2020-04-03
fix(ci): fix tap test failure reporting

InteractiveTimmy:
# Pull Request Template

## Description

The scope of this pull request is to address the tap tests not reporting failures when failures are present.

Fixes # [SPARK-133294](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-133294)

## Type of Change

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -337,6 +337,20 @@ jobs:
       - restore_node_modules
       - restore_built_packages
       - run: npm run test --silent -- --packages << parameters.package >> --browsers defaults
+      - run:
+          name: 'Check if tap test for "<< parameters.package >>" failed'
+          command: |
+            mkdir -p reports/junit/karma
+            FAILURES=$(find reports/junit/karma/ -iname '*.xml' -print0 | xargs -0 sed -n 's/.*failures=\"\([^\"]*\).*/\1/p' | xargs)
+
+            echo Failures $FAILURES
+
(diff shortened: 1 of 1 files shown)


taymoork2:
The testing command runs a for loop so its not going to exit like you'd want
It's gonna exit 0 because the loop finished executing all the spawn commands from all the packages that it was given (regardless if its 1 or 10)

Its more work to grab the exit code from tooling + run more tests if more packages are passed than it is to just look in the xml after the fact

lalli-flores:
Alright, thanks for the explanation.

.circleci/config.yml
```diff
@@ -337,6 +337,20 @@ jobs:
...
+      - run:
+          name: 'Check if tap test failed'
+          command: |
+            mkdir -p reports/junit/karma
+            FAILURES=$(find reports/junit/karma/ -iname '*.xml' -print0 | xargs -0 sed -n 's/.*failures=\"\([^\"]*\).*/\1/p' | xargs)
+
+            echo Failures $FAILURES
+
+            if [[ $FAILURES =~ ^([0 ])*$ ]]; then
+              echo 'Tap test succeeded'
```
adamweeks, line 349: Would be nice to have the result have the package it was testing echoed as well:

```suggestion
              echo 'Tap test for << parameters.package >> succeeded'
```

InteractiveTimmy: I added this to the commands being run, please let me know if these changes look okay.

PR #1607, opened 2020-04-08
feat(helper-image): replace height/width with exif info

aimeex3:
# Pull Request Template

## Description

It is possible the server comes back with a height/width different from the exif info. in order to render it correctly, we should use the height/width received from the exif info

Fixes # (issue)

## Type of Change

## Test Coverage

- [x] Uploaded an image to desktop client where image should be rotated. Desktop sends the height/width reversed so we should use the exif height/width instead

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


packages/node_modules/@webex/helper-image/test/unit/spec/index.js
```diff
@@ -22,13 +45,32 @@ describe('helper-image', () => {
         objectType: 'file'
       };
 
-      return file.fetch('/Portrait_7.jpg')
-        .then((f) => {
-          readExifData(sampleFile, f)
```
aimeex3, line 27: we were not returning previously so any failure in the `.then` below would mark the test as passed

packages/node_modules/@webex/helper-image/test/unit/spec/index.js
```diff
@@ -4,11 +4,34 @@
...
-import file from '@webex/test-helper-file';
+import fileHelper from '@webex/test-helper-file';
 import sinon from 'sinon';
+import {browserOnly, nodeOnly} from '@webex/test-helper-mocha';
 
 describe('helper-image', () => {
   describe('readExifData()', () => {
+    let buffer;
+
+    browserOnly(before)(() => fileHelper.fetch('/Portrait_7.jpg')
```
aimeex3, line 15: after fixing returning from the tests, the browser tests were failing because `fileHelper.fetch` in browser fetches the blob instead of a buffer. added this as a workaround to fix the tests
ideally test-helper-file should probably be refactored, but out of scope for this PR

adamweeks:
@aimeex3 

>It is possible the server comes back with a height/width different from the exif info. in order to render it correctly, we should use the height/width received from the exif info

Do we have some use cases and tickets opened with the server team to let them know of this issue? I'd like to make sure this is handled on the server side.

packages/node_modules/@webex/helper-image/src/index.js
```diff
@@ -73,6 +73,8 @@ export function readExifData(file, buf) {
       new ExifImage({image: buf}, (error, exifData) => {
         if (!error && exifData) {
           file.image.orientation = exifData.image.Orientation;
+          file.image.width = exifData.exif.ExifImageWidth;
```
adamweeks, line 76: Is this data populated by the server originally?

aimeex3: it's populated by the clients originally. i meant when we get it back from the server

PR #1613, opened 2020-04-09
test: add media flags for default browsers

adamweeks:
Default browsers for integration tests did not have the proper flags when being run.


adamweeks:
@taymoork2 says this is not working properly

PR #1616, opened 2020-04-10
Apply H264 flags to all browsers

taymoork2:
## Description

Applies H264 to all browsers that don't include it by default
Also allows better flexibility when using `--browsers` flag 

## Type of Change


browsers-ng.js
```diff
@@ -1,15 +1,80 @@
...
+// Also Karma's/Webdriver's `browserNames` are case-sensitive
+// Reformat since we made `argv.browsers` lowercase earlier
+// so it plays nice with Karma/Webdriver
+const format = (browser) => {
+  if (browser.includes('headless')) {
+    // `firefoxheadless` -> `firefoxHeadless`
+    browser = browser.replace('headless', 'Headless');
+  }
+
+  // `chrome` -> `Chrome`
```
lalli-flores, line 13: This will also return `FirefoxHeadless`, is that the expectation? I'd expand the comment for clarity
```suggestion
  // 'chrome' => 'Chrome' or 'firefoxHeadless' => 'FirefoxHeadless'
```

adamweeks:
I'd like to see some sort of documentation around these scripts since you've dove into them now and can explain how they work. Maybe in the contributing or scripts.md?

adamweeks:
We should be able to skip ci this PR as well since no application code is being updated.

PR #1622, opened 2020-04-14
Updating Circle TAP config to use production environment

adamweeks:
The build command as well as the tests were using the integration environment.
This adjusts the TAP tests to build for production servers, as well as points the test runners to the proper test user convo service and EU org id.

No JIRA, came up during incident investigation.

Tested manually with webex-core using the provided env vars and was getting 100% pass rate.

Verified config changes via:
> `circleci config validate`
> Config file at .circleci/config.yml is valid.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,35 @@ jobs:
           paths:
             - packages/node_modules
 
+  build_production:
+    executor: main-executor
+    environment:
+      NODE_ENV: production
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -434,7 +471,7 @@ workflows:
     triggers:
       - schedule:
           # Everyday @ 6PM UTC | 4PM EST | 1PM PST
-          cron: '0 18 * * 1-5'
```
lalli-flores, line 437: Can you update the comment to reflect the time change?

PR #1633, opened 2020-04-21
docs(webex): fix init docs [skip ci]

adamweeks:
Fixing the docs for `Webex.init` usage with an access token.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-136745

### DIFF
--- a/packages/node_modules/webex/src/webex.js
+++ b/packages/node_modules/webex/src/webex.js
@@ -61,9 +61,7 @@ const Webex = WebexCore.extend({
  * @example
  * <caption>Create a new Webex instance configured for a Bot</caption>
  * const webex = Webex.init({
- *   config: {
- *     credentials: `<BOT TOKEN FROM DEVELOPER PORTAL>`
- *   }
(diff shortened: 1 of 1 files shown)


daniel-liuzzi:
Is this change outdated/incorrect? Accodring to the code below, it seems `config` is required:

https://github.com/webex/webex-js-sdk/blob/acaeb17c9c70e6c5367f7e696eea5ba404806e27/packages/node_modules/webex/src/webex.js#L75

lalli-flores:
@daniel-liuzzi, are you suggesting that the format `{ credentials: <TOKEN> }` is the cause of the logger error but when used as `{ config: { credentials: <TOKEN> } }` the error doesn't appear?

daniel-liuzzi:
@lalli-flores No, the error I'm referring to is the `TypeError: Cannot read property 'client_id' of undefined` one. It only happens when using the shorter format. I removed the logger error bit to avoid further confusion. Sorry I didn't make this clearer.

PR #1634, opened 2020-04-21
Feature/add enumeratedevices example to select samples

brunogenaro:
Show our internal use of `enumerateDevices` API in the samples to show how to dynamically set getUserMediaConstraints.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-132106


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2150,6 +2150,21 @@ export default class Meeting extends StatelessWebexPlugin {
     );
   };
 
+  /**
+  *  Checks if the machine has at least one audio or video device
+  * @param {object} [options]
```
adamweeks, line 2155: Let's use this format for documenting the object properties:

```js
* @param {Object} options
* @param {Boolean} options.sendAudio
* @param {Boolean} options.sendVideo
```

packages/node_modules/samples/browser-plugin-meetings/app.js
```diff
@@ -501,32 +501,35 @@ document.getElementById('updateVideo').addEventListener('click', () => {
...
-  if (media.sendVideo) {
-    // Sending a boolean to `getMediaStreams` lets the browser decide which device to use.
-    let video = true;
-
-    // If we have a device specified, tell `getMediaStreams` to use it specifically.
-    if (videoSelect.value) {
-      video = {
-        deviceId: {
-          exact: videoSelect.value
+  meeting.getSupportedDevice({sendAudio: media.sendAudio})
```
adamweeks, line 504:
```suggestion
  meeting.getSupportedDevice({sendVideo: media.sendVideo})
```

packages/node_modules/samples/browser-single-party-call/app.js
```diff
@@ -227,3 +233,37 @@ document.getElementById('dialer').addEventListener('submit', (event) => {
       // Implement error handling here
     });
 });
+
+window.addEventListener('load', () => {
```
adamweeks, line 237: Let's add a bunch of comments in here since this is a sample app. The more verbose the better since consumers will most likely be copy/pasting it into their apps.

packages/node_modules/samples/browser-single-party-call/app.js
```diff
@@ -227,3 +233,37 @@ document.getElementById('dialer').addEventListener('submit', (event) => {
...
+  const audio = document.getElementById('constraints-audio');
+  const video = document.getElementById('constraints-video');
+
+  if (navigator && navigator.mediaDevices && navigator.mediaDevices.enumerateDevices) {
```
adamweeks, line 241: This would be a nice helper function to move into the main `meetings/index.js`. Not necessary for this ticket though. @taymoork2 don't you think?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2150,6 +2150,21 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  *    sendAudio: true/false,
+  *    sendVideo: true/false
+  * }
+  * @returns {Object} {
+  *    sendAudio: true/false,
+  *    sendVideo: true/false
+  *}
+  * @memberof Meetings
+  */
+ getSupportedDevice = (options) => Media.getSupportedDevice(options);
```
adamweeks, line 2166: If the options object is empty, could we default to `sendAudio: true, sendVideo:true`? That way you can just call `meeting.getSupportedDevice()` and find out if you have an audio or video device.

packages/node_modules/samples/browser-single-party-call/app.js
taymoork2: exposing `getSupportedDevice` (which should be renamed to `getSupportedDevices`) should be enough

taymoork2: This is fine because it was to needed be run before the SDK is downloaded

brunogenaro: @taymoork2 Should I rename it now? Wondering if there is people already using this method that would be a breaking change?

taymoork2: Probably the one you're exposing since its the first time it's being exposed as a public api
Internal getSupportedDevice can get renamed later if need be

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
brunogenaro: Done!

packages/node_modules/samples/browser-plugin-meetings/app.js
brunogenaro: ooops! done!

packages/node_modules/samples/browser-single-party-call/app.js
brunogenaro: done!

packages/node_modules/samples/browser-single-party-call/app.js
brunogenaro: done!

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
brunogenaro: done!

PR #1636, opened 2020-04-22
feat(webex-core): enable validateDomains by default

adamweeks:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-137880

By default, domains authentication is disabled and requires a specific configuration on the JS SDK setting validateDomains to true. This setting should be defaulted to true so that services can spin up before being added to the U2C catalog. 

This request was made as a suggestion by Sunil Veluvali while researching issues around SPARK-136704.


packages/node_modules/@webex/webex-core/src/config.js
```diff
@@ -65,7 +65,7 @@ export default {
       hydra: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1',
       hydraServiceUrl: process.env.HYDRA_SERVICE_URL || 'https://api.ciscospark.com/v1'
     },
-    validateDomains: process.env.VALIDATE_DOMAINS === 'true' || false,
+    validateDomains: true,
```
InteractiveTimmy, line 68: Don't need this one, see [this case](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-135715).

adamweeks: The line can be removed completely?

InteractiveTimmy: Yes, as all of these values are no longer used by the device plugin. There's a task to remove this information in our backlog.

lalli-flores: Can you link the task?

PR #1638, opened 2020-04-24
fix(webex-core): service to credentials auth url

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request are focused around generating the credentials config value `authorizeUrl`. An earlier bug fix seemed to have impacted the federation status of this `authorizeUrl` value.

Fixes # [SPARK-137781](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-137781)

## Type of Change

## Test Coverage

- [x] Generated new tests to validate that the added functionality works
- [x] Ran package tests

**Test Configuration**:
* NPM Version v6.13.4

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/services/services.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/services.js
@@ -410,6 +410,15 @@ const Services = WebexPlugin.extend({
     const {idbroker, identity} = this.list(true);
 
     if (idbroker && identity) {
+      const {authorizationString, authorizeUrl} = this.webex.config.credentials;
+
+      // This must be set outside of the setConfig method used to assign the
+      // idbroker and identity url values.
+      this.webex.config.credentials.authorizeUrl = authorizationString ?
+        authorizeUrl :
+        `${idbroker.replace(trailingSlashes, '')}/idb/oauth2/v1/authorize`;
+
+
       this.webex.setConfig({
         credentials: {
           idbroker: {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -410,6 +410,15 @@ const Services = WebexPlugin.extend({
     const {idbroker, identity} = this.list(true);
 
     if (idbroker && identity) {
+      const {authorizationString, authorizeUrl} = this.webex.config.credentials;
+
+      // This must be set outside of the setConfig method used to assign the
+      // idbroker and identity url values.
+      this.webex.config.credentials.authorizeUrl = authorizationString ?
```
andrewholsted, line 417: Why do we do this explicitly here and use `setConfig` below?

InteractiveTimmy: Set config allows the safe mutation of nested object values while maintaining the AmperstandState events:

```js
idbroker: {
  url: `nested object value A` // Only the value changed, object still mutable, event persists
},
identity: {
  url: `nested object value B` // Only the value changed, object still mutable. Event persists
}
```

When attempting to set these values directly, some unwated artifacts occur with eventing. When attempting to set `authorizeUrl` using these values, errors are thrown at runtime / testing.

InteractiveTimmy:
> Thanks for diving into this and trying to get the missing functionality back. This feels a bit like a workaround to me, though. Is there not a way we can do this that is compatible with `setConfig`? If not, and @andrewholsted is good with the changes, then I have nothing blocking as the code looks good and is tested.

There is not, at least within my triaging. This roots into how AmpersandState was used with the credentials config, and is not mounted like other configs are mounted. Without reworking the entire config file and how it is utilized within the credentials plugin, and any downstream plugins, we will have to hard-set the value this way. This is identical to how we set this value within the `Credentials#initialize()` method.

PR #1645, opened 2020-04-28
chore(samples): resize window to match saucelabs' asked resolution

taymoork2:
## Description

This doesn't actually do very much on Saucelabs' side because the recording is still done at like 720p.
But a man can dream.

## Type of Change

## Checklist:


wdio.conf.js
```diff
@@ -317,8 +317,14 @@ exports.config = {
    * @param {Array.<Object>} capabilities list of capabilities details
    * @param {Array.<String>} specs List of spec file paths that are to be run
    */
-  // before(capabilities, specs) {
-  // }
+  // eslint-disable-next-line no-unused-vars
+  before(capabilities, specs) {
```
adamweeks, line 321: Will this affect local debugging too? Do we want to only do this on sauce?

taymoork2: This will affect local debugging too
But all our macbooks should come with a minimum resolution of `2560x1600` so it shouldn't take up all fo the screen
But I can add a `CI` check so it's only limited to Sauce

adamweeks: This works for me, might be good to have the window sizes match on both local and circle.

PR #1651, opened 2020-04-29
fix(meeting): enforce moderator as true only if user sets hostPin

brunogenaro:
## Description
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-135440: Fix issue when a user attempts to call into a space that they created, before a call is ongoing, from the browser-plugin-meetings the attempt will fail with a message indicating that the room is locked.
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-134717: Remove false moderator flag since locus service has requested that instead of sending `moderator: false` during meeting joins, we do not send the flag at all.

Tickets: 
- https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-135440
- https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-134717

This bug was introduced in this Pull Request: https://github.com/webex/webex-js-sdk/commit/588fad7089838af6a85a95374a90abab9cc072d1#diff-e506603fad074bab354b0e9bf78e4b38R68-R90

Steps to test:
1. Create a Space, and add at least one other user. In my testing I created the Space detailed above, and added one Guest user.
2. Obtain the Space information. In order to place the test call, either the "SpaceId" or the Space's "sipAddress" will be needed.
3. Obtain an access_token for the user that created the Space. This can be obtained from an OAuth or the Developer Portal.
4. Navigate to the browser-plugin-meetings.
5. Paste the access_token into the "Credentials" field, then click "connect".
6. Scroll down to the "Dialer" section, then paste either the SpaceId or sipAddress, then click "create"
7. Click "create" within the "Dialer" section, or scroll up slightly and select "join with media".
8. Should it start meeting with no issues


arun3528:
Can we test out the following scenario 

1) Joining Personal PME meeting 
2) create a webex meeting using the webex site (With host Pin)/ See if you do without host pin as well
3) create a scheduled webex PMR meeting 
4) unclaimed PMR () . https://developer.webex.com/docs/api/v1/meetings/create-a-meeting

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -88,8 +88,9 @@ MeetingUtil.cleanOptions = (options, info, owner, userId) => {
     if (MeetingUtil.isOwnerSelf(owner, userId)) {
```
chrisadubois, line 88: I think we should take a look at modifying isOwnerSelf rather than changing the logic here?

PR #1660, opened 2020-05-06
Add ability to override quality on join

taymoork2:
## Description

Today I (somewhat) learned how to use //webrtc-internals

Fixes [#SPARK-133817](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-133817)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -86,6 +86,9 @@ import InMeetingActions from './in-meeting-actions';
  * @property {String} [resourceId]
  * @property {String} [pin]
  * @property {Boolean} [moderator]
+ * @property {String|Object} [meetingQuality]
+ * @property {String} [meetingQuality.local]
+ * @property {String} [meetingQuality.remote]
  */
 
 /**
@@ -1974,6 +1977,43 @@ export default class Meeting extends StatelessWebexPlugin {
       this.wirelessShare = true;
     }
 
+    if (options.meetingQuality) {
+      if (typeof options.meetingQuality === 'string') {
+        if (!QUALITY_LEVELS[options.meetingQuality]) {
+          const errorMessage = `Meeting:index#join --> ${options.meetingQuality} not defined`;
+
+          LoggerProxy.logger.error(errorMessage);
+
+          return Promise.reject(new Error(errorMessage));
+        }
+
+        this.mediaProperties.setLocalQualityLevel(options.meetingQuality);
+        this.mediaProperties.setRemoteQualityLevel(options.meetingQuality);
+      }
+
+      if (typeof options.meetingQuality === 'object') {
+        if (
+          !QUALITY_LEVELS[options.meetingQuality.local] &&
+          !QUALITY_LEVELS[options.meetingQuality.remote]
+        ) {
+          const errorMessage = `Meeting:index#join --> ${
+            options.meetingQuality.local || options.meetingQuality.remote
+          } not defined`;
+
+          LoggerProxy.logger.error(errorMessage);
+
+          return Promise.reject(new Error(errorMessage));
+        }
+
+        if (options.meetingQuality.local) {
+          this.mediaProperties.setLocalQualityLevel(options.meetingQuality.local);
+        }
+        if (options.meetingQuality.remote) {
+          this.mediaProperties.setRemoteQualityLevel(options.meetingQuality.remote);
+        }
+      }
+    }
+
     return MeetingUtil.joinMeetingOptions(this, options).then((join) => {
       this.meetingFiniteStateMachine.join();
 
@@ -2103,7 +2143,12 @@ export default class Meeting extends StatelessWebexPlugin {
    * @todo should be static, or moved so can be called outside of a meeting
    * @memberof Meeting
    */
-  getMediaStreams = (mediaDirection, audioVideo = {}, sharePreferences) => {
+  getMediaStreams = (
+    mediaDirection,
+    // This return an OBJECT {video: {height, widght}}
+    audioVideo = VIDEO_RESOLUTIONS[this.mediaProperties.localQualityLevel],
+    sharePreferences
+  ) => {
     if (
       mediaDirection &&
       (
@@ -2857,6 +2902,11 @@ export default class Meeting extends StatelessWebexPlugin {
       return Promise.reject(new Error(errorMessage));
     }
 
+    // If level is already the same, don't do anything
+    if (level === this.mediaProperties.localQualityLevel) {
+      return Promise.resolve();
+    }
+
     // Set the quality level in properties
     this.mediaProperties.setLocalQualityLevel(level);
 
@@ -2899,6 +2949,11 @@ export default class Meeting extends StatelessWebexPlugin {
       return Promise.reject(new Error(errorMessage));
     }
 
+    // If level is already the same, don't do anything
+    if (level === this.mediaProperties.remoteQualityLevel) {
+      return Promise.resolve();
+    }
+
     // Set the quality level in properties
     this.mediaProperties.setRemoteQualityLevel(level);
 
@@ -2921,22 +2976,27 @@ export default class Meeting extends StatelessWebexPlugin {
       return Promise.reject(new Error(errorMessage));
     }
 
-    // Set the overall quality level
-    this.mediaProperties.setQualityLevel(level);
+    // If level is already the same, don't do anything
+    if (
+      level === this.mediaProperties.localQualityLevel &&
+      level === this.mediaProperties.remoteQualityLevel
+    ) {
+      return Promise.resolve();
+    }
 
     const qualityActions = [];
 
     // Determine the direction of our current media
     const {receiveAudio, receiveVideo, sendVideo} = this.mediaProperties.mediaDirection;
 
-    // If we are receiving media, set the remote media quality
-    if (receiveAudio || receiveVideo) {
-      qualityActions.push(this.setRemoteQualityLevel(level));
-    }
-
     // If we are sending media, set the local media quality
     if (sendVideo) {
       qualityActions.push(this.setLocalVideoQuality(level));
+    }
+
+    // If we are receiving media, set the remote media quality
+    if (receiveAudio || receiveVideo) {
+      qualityActions.push(this.setRemoteQualityLevel(level));
     }
 
     return Promise.all(qualityActions);
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -62,6 +62,7 @@ import InMeetingActions from './in-meeting-actions';
  * @property {boolean} receiveVideo
  * @property {boolean} sendShare
  * @property {boolean} receiveShare
+ * @property {boolean} isSharing
  */
 
 /**
@@ -2169,7 +2170,8 @@ export default class Meeting extends StatelessWebexPlugin {
             {
               ...mediaDirection,
               sendAudio: (mediaDirection.sendAudio === devicePermissions.sendAudio),
-              sendVideo: (mediaDirection.sendVideo === devicePermissions.sendVideo)
+              sendVideo: (mediaDirection.sendVideo === devicePermissions.sendVideo),
+              isSharing: this.isSharing
             },
             audioVideo,
             sharePreferences,
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -2904,6 +2904,8 @@ export default class Meeting extends StatelessWebexPlugin {
 
     // If level is already the same, don't do anything
     if (level === this.mediaProperties.localQualityLevel) {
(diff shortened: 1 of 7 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1979,6 +1983,10 @@ export default class Meeting extends StatelessWebexPlugin {
       this.wirelessShare = true;
     }
 
+    if (options.qualityOverride) {
```
adamweeks, line 1986: This needs to be added to the jsdoc typedef

taymoork2: Addressed: https://github.com/webex/webex-js-sdk/pull/1660/files#diff-7362c0f286448790746a873a7c764a7bR1835

adamweeks:
Could you please also add this functionality to the kitchen sink in this PR for local testing?

adamweeks:
When testing this branch out on the samples, I've done the following but cannot get it to properly constrain:

* connect sdk
* create meeting
* set meeting quality level to low and click "join"
* click "updateMedia"

"overcontstrained media error" happens during the get user media call.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2925,6 +2934,11 @@ export default class Meeting extends StatelessWebexPlugin {
       return Promise.reject(new Error(errorMessage));
     }
 
+    // If level is already the same, don't do anything
```
lalli-flores, line 2937: Is this necessary? what's the cost of reassigning the quality level?

taymoork2: Everything after this if block will cause changes to the active stream
Why mess with the stream if we aren't gonna actually change anything

```js
	// If we are receiving media, set the remote media quality
    if (receiveAudio || receiveVideo) {
      qualityActions.push(this.setRemoteQualityLevel(level));
    }
    // If we are sending media, set the local media quality
    if (sendVideo) {
      qualityActions.push(this.setLocalVideoQuality(level));
    }
```

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1828,6 +1832,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @param {JoinOptions} [options.joinOptions] - see #join()
    * @param {MediaDirection} options.mediaSettings - see #addMedia()
    * @param {AudioVideo} [options.audioVideoOptions] - see #getMediaStreams()
+   * @param {Quality} [options.meetingQuality] - see #setMeetingQuality() -- Run when remote stream is connected
```
adamweeks, line 1835: This typedef is not defined. Please just use `String` or create a typedef.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1979,6 +1984,10 @@ export default class Meeting extends StatelessWebexPlugin {
       this.wirelessShare = true;
     }
 
+    if (options.meetingQuality) {
```
adamweeks, line 1987: `JoinOptions` typedef needs to be updated

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: I like this change, but would also like to see it in the `setRemoteQualityLevel` and `setLocalVideoQuality` commands so they don't call `updateMedia` extraneously if called directly.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1386,6 +1386,10 @@ export default class Meeting extends StatelessWebexPlugin {
 
       this.mediaProperties.remoteStream = streams;
 
+      // Set all stream(s) quality once remote stream's bytes are recieved
+      // This will do nothing if override value isn't given
+      this.setMeetingQuality(this.mediaProperties.qualityLevel);
```
adamweeks, line 1391: I'm not sure why this is needed. Isn't the meeting quality already set at this point?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: Still not addressed

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
taymoork2: I thought this was the agreed spot to set the quality, since calling `setMeetingQuality` on join causes failures due to the remote stream not existing to set it's quality?
So, like the comment says, when the remote stream's bytes are received then set the meeting's quality with the given `meetingQuality` value set at join

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
taymoork2: updated and added

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: It seems like we should be able to set the stream quality at the time of peer connection establishment instead of having to get a standard peer connection at first and then renegotiate a different quality one. 
@arun3528  thoughts?

arun3528: agreed it should be part of addMedia

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -86,6 +86,9 @@ import InMeetingActions from './in-meeting-actions';
  * @property {String} [resourceId]
  * @property {String} [pin]
  * @property {Boolean} [moderator]
+ * @property {String|Object} [meetingQuality]
```
taymoork2, line 89: We decided to make meetingQuality either a string or a object so that the value can be set together or individually

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1974,6 +1977,22 @@ export default class Meeting extends StatelessWebexPlugin {
       this.wirelessShare = true;
     }
 
+    if (options.meetingQuality) {
+      if (typeof options.meetingQuality === 'string') {
```
adamweeks, line 1981: Let's check for a proper string value as well:

```suggestion
      if (typeof options.meetingQuality === 'string' && QUALITY_LEVELS[options.meetingQuality]) {
```

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1974,6 +1977,22 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        this.mediaProperties.setLocalQualityLevel(options.meetingQuality);
+        this.mediaProperties.setRemoteQualityLevel(options.meetingQuality);
+      }
+
+      if (typeof options.meetingQuality === 'object') {
+        if (options.meetingQuality.local) {
```
adamweeks, line 1987: These values need to be validated since the setters on `mediaProperties` don't do any validation.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2920,8 +2953,22 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    this.mediaProperties.setQualityLevel(level);
+    // If level is already the same, don't do anything
+    if (
+      level === this.mediaProperties.localQualityLevel &&
+      level === this.mediaProperties.remoteQualityLevel
+    ) {
+      return Promise.resolve();
+    }
+
+    // Set the overall quality level if the level(s) aren't already set to that value
```
adamweeks, line 2964: You can remove this line since there's no overall quality level any more

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2920,8 +2953,22 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    // Only change what's needed since this will change conenction settings
```
adamweeks, line 2965:
```suggestion
    // Only change what's needed since this will change connection settings
```

packages/node_modules/samples/browser-plugin-meetings/index.html
```diff
@@ -236,6 +236,15 @@ <h2>Then click:</h2>
       </br>
         <input title="create" type="submit" value="create" />
       </br>
+      <p>
+        <label for="setMeetingQualityLevel">Set Meeting Quality Level</label>
```
adamweeks, line 240: Would prefer this to show that it is only being used at join:

```suggestion
        <label for="setMeetingQualityLevel">Meeting Join Quality Level</label>
```

taymoork2: Does the rest of the sample need to reflect that meetingQuality can be either a string or an object? And what's the one we'd go with in the Sample?

packages/node_modules/samples/browser-plugin-meetings/index.html
```diff
@@ -301,26 +311,26 @@ <h2>Then click:</h2>
...
         <p>
-          <label for="setLocalVideoQualityLevel">Set Local Video Quality</label>
-          <select id="setLocalVideoQualityLevel" name="setLocalVideoQualityLevel">
+          <label for="updateLocalVideoQualityLevel">Update Local Video Quality</label>
+          <select id="updateLocalVideoQualityLevel" name="updateLocalVideoQualityLevel">
             <option value="LOW">LOW (320:180)</option>
             <option value="MEDIUM">MEDIUM (640:360)</option>
             <option value="HIGH">HIGH (1280:720)</option>
           </select>
           <button id="setLocalVideoQuality" title="meeting.setLocalVideoQuality()" type="button">meeting.setLocalVideoQuality()</button>
```
adamweeks, line 329: This needs to match the app.js implementation

```suggestion
          <button id="updateLocalVideoQuality" title="meeting.setLocalVideoQuality()" type="button">meeting.setLocalVideoQuality()</button>
```

packages/node_modules/samples/browser-plugin-meetings/app.js
```diff
@@ -1297,29 +1311,29 @@ document.getElementById('sendDtmf').addEventListener('click', () => {
...
     if (meeting) {
-      const level = document.getElementById('setLocalVideoQualityLevel').value;
+      const level = document.getElementById('updateLocalVideoQualityLevel').value;
 
       meeting.setLocalVideoQuality(level);
     }
   });
 
-document.getElementById('setRemoteMeetingQuality')
+document.getElementById('updateRemoteMeetingQuality')
```
adamweeks, line 1332: This element's ID was not updated in `index.html`

adamweeks:
Posted in our space, but also posting here:

Here is my workflow and what is going wrong:
https://localhost:8000/browser-plugin-meetings/
Connect with credentials
Create meeting to space id
Set meeting join quality level to LOW (320x180)
Click “join”
Click “getDevices” and select the correct device
Click “addMedia” with a breakpoint on Media.getMedia (https://github.com/taymoork2/webex-js-sdk/blob/71003a43a99dfa4e972c5814293412ea9831c998/packages/node_modules/@webex/plugin-meetings/src/media/index.js#L341)
See that the media constraints passed to navigator.mediaDevices.getUserMedia are still 1280x720

We should be setting the constraints here to the resolution chosen in quality level

taymoork2:
@adamweeks resolved the issue
Arun suggested that setting the 
`audioVideo` default value needs `VIDEO_RESOLUTIONS[this.mediaProperties.localQualityLevel]` to be set with a `{video: ...}` this caused the breakage in `getMedia` since `VIDEO_RESOLUTIONS` will return a `{video: ...}` anyway, so it was being double wrapped
`getMedia` did a check against video but we didn't throw an error since it would resolve to the default value

packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js
```diff
@@ -339,7 +339,6 @@ pc.setRemoteSessionDetails = (peerConnection, typeStr, remoteSdp, meetingId) =>
 pc.createOffer = (peerConnection, {meetingId, remoteQualityLevel}) => {
   LoggerProxy.logger.log('PeerConnectionManager->createOffer#createOffer: creating a new offer');
 
-  try {
```
taymoork2, line 342: The try/catch doesn't return a promise
The promise itself does have a catch block to return an error anyway

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -715,10 +715,6 @@ describe('plugin-meetings', () => {
           assert.exists(meeting.setMeetingQuality);
         });
 
-        it('should set mediaProperty with the proper level', () => meeting.setMeetingQuality(CONSTANTS.QUALITY_LEVELS.LOW).then(() => {
```
taymoork2, line 718: `setMeetingsQuality` doesn't change the `mediaProperties` directly anymore so this test case will always fail since `setLocalQualityLevel` and `setRemoteQualityLevel` are stubbed and they are handling the property change

PR #1652, opened 2020-04-29
fix(plugin-logger): handle circular references in logger

gannons:
# Pull Request Template

## Description

During a Mercury problem I noticed the following in our test app's logs

    2020-04-29T11:18:31.925Z info: Successfully registered with WDM
    2020-04-29T11:18:31.926Z info: Creating report
    failed to execute Logger#warn RangeError: Maximum call stack size exceeded
        at RegExp.test (<anonymous>)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:119:31)
        at Array.map (<anonymous>)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:94:19)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        ...... 1,000s of repetitions ......
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at walkAndFilter (/Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:122:23)
        at /Users/gannons/Development/ediscovery_downloader/test-scripts/node_modules/@webex/plugin-logger/dist/logger.js:199:14
        at Array.map (<anonymous>)
[...]

PR #1662, opened 2020-05-07
fix(plugin-logger): circular ref should not filter funcs or primatives

gannons:
# Pull Request Template

## Description

Follow up to https://github.com/webex/webex-js-sdk/pull/1652. Noticed that while that PR fixed the recursive loop it also had an unintended side effect. Namely it caused issues like

> failed to execute Logger#warn RangeError [ERR_OUT_OF_RANGE]: The value of "defaultMaxListeners" is out of range. It must be a non-negative number. Received [Circular]

and

> uncaughtException TypeError: Function.prototype.apply was called on [Circular], which is a string and not a function

Realised that the original PR did not take into account primitives and functions were also being walked.

Fixes # (issue)

## Type of Change

## Test Coverage

Added an enhanced unit test to cover this case.

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/plugin-logger/src/logger.js
+++ b/packages/node_modules/@webex/plugin-logger/src/logger.js
@@ -47,7 +47,7 @@ const authTokenKeyPattern = /[Aa]uthorization/;
 function walkAndFilter(object, visited = []) {
   if (visited.includes(object)) {
     // Prevent circular references
-    return '[Circular]';
+    return object;
   }
 
   visited.push(object);
--- a/packages/node_modules/@webex/plugin-logger/test/unit/spec/logger.js
+++ b/packages/node_modules/@webex/plugin-logger/test/unit/spec/logger.js
@@ -409,7 +409,64 @@ describe('plugin-logger', () => {
       object.selfReference = object;
 
       webex.logger.log(object);
-      assert.calledWith(console.log, {Key: 'myKey', selfReference: '[Circular]', string: '-- REDACTED --'});
+
+      const expected = {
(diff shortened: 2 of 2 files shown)


packages/node_modules/@webex/plugin-logger/src/logger.js
```diff
@@ -47,7 +47,7 @@ const authTokenKeyPattern = /[Aa]uthorization/;
 function walkAndFilter(object, visited = []) {
   if (visited.includes(object)) {
     // Prevent circular references
-    return '[Circular]';
+    return object;
```
gannons, line 50: Originally I was going to fix this by checking `object` with `isObject(object)` and only recording the visit in such a case. That also required a check of `!isFunction(object)` to fix

> uncaughtException TypeError: Function.prototype.apply was called on [Circular], which is a string and not a function

so the code looked like

    if (isObject(object) && !isFunction(object)) {
      if (visited.includes(object)) {
        ......

But when I gave it more thought I realised it is possible for a function itself to hold a circular reference (though I'm not sure if that will happen).

I also realised I originally returned `[Circular]` as I believed the objected (which is cloned when sent to the logger) would never be used. Turns out that is not the case.

So now instead of setting `[Circular]` the code just immediately returns the object if it has already been visited. That will still filter out sensitive information, while allowing the object to remain usable, and prevent an unending recursive loop.

gannons:
Looks like a random failure in the samples

> Timed-out waiting for screenshare to start

Can you kick off a rebuild @taymoork2?

PR #1669, opened 2020-05-13
Meetings Plugin - Hostpin to pin conversion

adamweeks:
## Description

Updates the meetings plugin to utilize the `pin` in join requests and no longer uses the deprecated `hostPin` field.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-141185

Manual testing done on both prod and int for the following scenarios:

image


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1962,11 +1962,6 @@ export default class Meeting extends StatelessWebexPlugin {
       });
     }
 
-    // Should PIN collection be followed by pin propt ???
-    if (options.hostPin) {
-      Metrics.pinCollection({meeting: this});
-    }
```
arun3528, line 1968: @chrisadubois  thoughts on this . we can check moderator flag and then send metrics but do we need to send !?

adamweeks: @arun3528 I removed this line because `Metrics.pinCollection` doesn't exist. Surprised there wasn't more errors around it.

We are still sending the metrics on pin collection here: https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/plugin-meetings/src/meeting/util.js#L193

arun3528: thanks @adamweeks

PR #1670, opened 2020-05-13
fix(samples): add all chars to dtmf input

InteractiveTimmy:
# Pull Request

## Description

The changes in this request is scoped to allowing all characters into the input field for DTMF in the meetings sample.

Fixes # [SPARK-143416](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-143416)

## Type of Change

### DIFF
--- a/packages/node_modules/samples/browser-plugin-meetings/index.html
+++ b/packages/node_modules/samples/browser-plugin-meetings/index.html
@@ -297,7 +297,7 @@
         <button id="destroyStats" title="stop filter getStats" type="button">stop getStats</button>
         <p>
           <label for="sendDtmfInput">Send DTMF Tones</label>
-          <input id="sendDtmfInput" name="sendDtmf" value="" placeholder="DTMF Numbers" type="number" />
+          <input id="sendDtmfInput" name="sendDtmf" value="" placeholder="DTMF Numbers" type="text" />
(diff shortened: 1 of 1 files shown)


packages/node_modules/samples/browser-plugin-meetings/index.html
```diff
@@ -297,7 +297,7 @@ <h2>Then click:</h2>
         <button id="destroyStats" title="stop filter getStats" type="button">stop getStats</button>
         <p>
           <label for="sendDtmfInput">Send DTMF Tones</label>
-          <input id="sendDtmfInput" name="sendDtmf" value="" placeholder="DTMF Numbers" type="number" />
+          <input id="sendDtmfInput" name="sendDtmf" value="" placeholder="DTMF Numbers" type="text" />
```
taymoork2, line 300: Do we want to do some kind of validation before sending it up?

Grabbed from https://gist.github.com/d-levin/21df4ab1ad759bb7ecd865100a7909fd
`\d` = 0-9

```suggestion
          <input id="sendDtmfInput" name="sendDtmf" pattern="^[\d#*]+$" value="" placeholder="DTMF Numbers" type="text" />
```

InteractiveTimmy: Per our meeting, we don't, but, I'm going to add it so that it has a tighter scope in the PR.

PR #1672, opened 2020-05-14
Adding new APIs to retrieve and decrypt contentContainer

decryan1984:
# Pull Request Template

## Description
New APIs are now available in the ediscovery service which will allow us to create reports that contain meetings, whiteboards etc. that are not related to a space. This change simply adds those APIs to the plugin and handles the decryption as normal via the transforms. New tests added to cover the new APIs.

Fixes # (issue)
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-139393

## Type of Change

## Test Coverage
All unit tests and integration tests were run locally, including the new tests

**Test Configuration**:
* SDK Version 1.80.210
* Node/Browser Version 10.16.3
* NPM Version  6.9.0

## Checklist:


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
...
+    options = {...this.config.defaultOptions, ...options};
+
+    const [requestOptions, reportUUID] = this._createRequestOptions(reportId, `/contents/container/${containerId}`);
+
+    // If this content container has already been cached then it can be retrieved from there instead of making a network call to ediscovery
+    if (this.contentContainerCache.has(reportUUID) && this.contentContainerCache.get(reportUUID).has(containerId)) {
+      return {body: this.contentContainerCache.get(reportUUID).get(containerId), statusCode: 200};
+    }
+
+    this.logger.warn(`Cache miss for container ${containerId} in contentContainerCache`);
```
gannons, line 388: Might be nice to start sending a metric for misses. But that's not necessary for this PR.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
...
+  },
+
+  /**
+   * The results of a getContentContainer or getContentContainerByContainerId request are written to the contentContainerCache
+   * since information for a container is likely to be accessed very frequently when decrypting activities
+   * @param {UUID} reportId - Id of the report which contains the relevant content summary
+   * @param {Array<Object>} containers - List of the container objects to be written to the cache
+   * @returns {Promise} - Promise resolution indicating operation is complete
+   */
+  _writeToContentContainerCache(reportId, containers) {
```
gannons, line 410: Slightly wish `ContentContainer` had been simplified to `Container`. But that's a matter of taste.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
...
+    if (!reportId || !containers || !containers.length) {
+      return;
+    }
+
+    if (!this.contentContainerCache.has(reportId)) {
+      this.contentContainerCache.set(reportId, new Map());
+    }
+
+    for (let i = 0; i < containers.length; i += 1) {
```
gannons, line 419: for (const container of containers) {....}

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
...
+      const container = containers[i];
+
+      if (container && container.containerId) {
+        try {
+          this.contentContainerCache.get(reportId).set(container.containerId, container);
+        }
+        catch (error) {
+          this.logger.error(`Error adding ${container.containerId} to contentContainerCache: ${error}`);
```
gannons, line 427: Is this possible? i.e. An error is thrown when adding an item to a map.

packages/node_modules/@webex/internal-plugin-ediscovery/src/index.js
```diff
@@ -157,6 +177,28 @@ registerInternalPlugin('ediscovery', EDiscovery, {
 
           return Promise.all(object.body.map((item) => ctx.transform('decryptReportContentSummary', {body: item})));
         }
+      },
+      {
+        name: 'decryptReportContentContainer',
+        direction: 'inbound',
+        fn(ctx, object) {
+          if (!object || !object.body || !object.body.containerName || !object.body.participants || !object.body.participants.length) {
```
gannons, line 185: If a container has no participants don't decrypt the `containerName`. Is that the right behaviour? It may not be possible if the actor is included in the participants (which I forget) as I don't think there is ever a container without an actor (though I'm not 100% about that).

gannons: BTW it looks like a lot of the same checks are done by the transform. Is this check necessary?

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -379,6 +379,52 @@ class Transforms {
...
+      return Promise.resolve(object);
+    }
+
+    return requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
+      [container.encryptionKeyUrl, container.containerName, {onBehalfOf: container.onBehalfOfUser}])
+      .then((decryptedContainerName) => {
+        container.containerName = decryptedContainerName;
+      })
+      .catch((reason) => {
+        ctx.webex.logger.warn(`Decrypt container name error for ${container.containerType} container ${container.containerId}: ${reason}`);
```
gannons, line 421: error?

packages/node_modules/@webex/internal-plugin-ediscovery/test/integration/spec/ediscovery.js
```diff
@@ -206,6 +207,48 @@ describe('internal-plugin-ediscovery', () => {
...
       });
+
+      it('getContentContainer by url succeeds with valid url', async () => {
+        await complianceUser.webex.internal.ediscovery.getContentContainer(validUrl)
+          .then(() => {
+            assert.fail('Expected error response due to invalid report id');
+          })
+          .catch((error) => {
+            assert.equal(error.statusCode, 404);
+            assert.include(error.message, `Report Record with ID[${reportId}] could not be found`);
```
gannons, line 218: Assume the service does return `Report Record` but I would have expected `Content Container`.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
     }
   },
 
+  @waitForValue('@')
+  @oneFlight({keyFactory: (reportId, options) => createOneFlightKey(reportId, options)})
```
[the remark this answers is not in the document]
decryan1984: I'll check it out, if it doesn't work that means other APIs are affected...

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +325,124 @@ const EDiscovery = WebexPlugin.extend({
...
+    const [requestOptions, reportUUID] = this._createRequestOptions(reportId, '/contents/container');
+
+    return this.request({
+      method: 'GET',
+      ...requestOptions,
+      qs: {offset: options.offset, size: options.size},
+      timeout: options.timeoutMs
+    })
+      .then(async (res) => {
+        await this._writeToContentContainerCache(reportUUID, res.body);
```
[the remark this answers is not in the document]
decryan1984: I think if we're going to adopt a new pattern we should do it across all the APIs, probably in a separate PR

packages/node_modules/@webex/internal-plugin-ediscovery/src/index.js
decryan1984: Yeah, good point, this is probably a relic from when the plugin had to determine the onBehalfOf user itself from the participant list, with the service now providing it we should probably just be checking for that. Will remove the duplicate checks here also, makes more sense to do them in the transform itself.

packages/node_modules/@webex/internal-plugin-ediscovery/test/integration/spec/ediscovery.js
decryan1984: The error is thrown because the report record does not exist

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: it can happen if no element is found for reportId and we try to call set on that, but that should be covered in the has check above, so I'll remove

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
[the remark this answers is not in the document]
decryan1984: I've made all the APIs async and switched to using await where applicable

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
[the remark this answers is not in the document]
decryan1984: It was working correctly, I've modified createOneFlightKey because of the new types option for getContentContainers, and added new unit tests to ensure oneFlight behaves correctly.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Turns out the try catch is required

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -328,6 +339,121 @@ const EDiscovery = WebexPlugin.extend({
...
+  },
+
+  /**
+   * Wipe the cache used by eDiscovery to store the space summaries and content containers.
+   * Good practice to clear it down when finished with a report to save RAM.
+   * @returns {undefined}
+   */
+  clearCache() {
+    this.contentSummaryBySpaceIdCache.clear();
+    this.contentContainerCache.clear();
```
decryan1984, line 454: Clearing both caches at the same time

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/content.js
```diff
@@ -256,32 +322,6 @@ describe('EDiscovery Content API Tests', () => {
...
-      webex.request = sinon.stub().returns(Promise.resolve({
-        body: {},
-        statusCode: 400
-      }));
-
-      webex.internal.ediscovery.postAuditLog(uuid, {timeoutMs: 1000})
-        .then((res) => {
-          expect(res.statusCode).equal(400);
-        });
-    });
```
decryan1984, line 284: moved to bottom of file

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -16,7 +16,21 @@ import {InvalidEmailAddressError} from './ediscovery-error';
  * @returns {String} oneFlight key which is a composite of the request parameters
  */
 function createOneFlightKey(reportId, options) {
-  return String(reportId + (options && options.offset && options.size ? options.offset + options.size : ''));
+  let key = String(reportId);
```
gannons, line 19: Assume you need `String(..)` to handle `undefined`?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -98,7 +122,7 @@ const EDiscovery = WebexPlugin.extend({
    * @param {number} options.timeoutMs - connection timeout in milliseconds, defaults to 30s
    * @returns {Promise<ResponseEntity<ReportRecord>>} Http response containing the specified report record
    */
-  getReport(reportId, options) {
+  async getReport(reportId, options) {
```
gannons, line 125: Technically this is not needed. But I won't object to it either.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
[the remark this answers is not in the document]
gannons: 👍 

Actually missed `createOneFlightKey(..)` and read the code as using only the `options` object

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
gannons: How come?

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Yeah, not needed atm but going for consistency across the APIs

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Throws a TypeError without the try catch.

packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
decryan1984: Not likely but better to ensure a string is returned

PR #1678, opened 2020-05-14
fix(auth): web auth is in the service catalog now

chrisadubois:
the title says it all


chrisadubois:
:(

packages/node_modules/@webex/webex-core/src/interceptors/auth.js
```diff
@@ -75,10 +75,6 @@ export default class AuthInterceptor extends Interceptor {
     const {resource, uri} = options;
     const service = options.service || options.api;
 
-    if ((service && service === 'web-authentication') || (name && name === 'web-authentication')) {
```
taymoork2, line 78: So does this now need a auth token?

chrisadubois:
Im confused, where is this left? we confirmed a working change in devices.webex.com?

yes web-auth endpoints require auth-token, if you want to call a web-auth end point that doesn't need auth one would need to add the parameter `addAuthHeader` and set to false

chrisadubois:
The only other alternative I see is that we would have an override for web-authentication where one could add 'addAuthHeader: true` that would override the service name check

packages/node_modules/@webex/webex-core/test/unit/spec/interceptors/auth.js
```diff
@@ -259,17 +259,6 @@ describe('webex-core', () => {
...
-          () => {
-            services = {};
-
-            return interceptor.requiresCredentials({
-              service: 'web-authentication',
-              resource: 'something'
-            })
-              .then((response) => assert.isFalse(response));
-          });
-
```
InteractiveTimmy, line 272: Thank you for removing the test.

PR #1680, opened 2020-05-21
feat(samples): add meetings management

InteractiveTimmy:
# Pull Request

## Description

The changes within this pull request are scoped to the `manage meetings` section of the new meetings samples.

Fixes # [SPARK-143729](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-143729)

## Type of Change

### DIFF
--- a/packages/node_modules/samples/browser-meetings/app.js
+++ b/packages/node_modules/samples/browser-meetings/app.js
@@ -83,3 +83,125 @@ function unregister() {
         'Not Registered';
     });
 }
+
+// Meetings Management Section --------------------------------------------------
+
+const createMeetingDestinationElm = document.querySelector('#create-meeting-destination');
+const createMeetingActionElm = document.querySelector('#create-meeting-action');
+const meetingsJoinDeviceElm = document.querySelector('#meetings-join-device');
+const meetingsJoinPinElm = document.querySelector('#meetings-join-pin');
+const meetingsJoinModeratorElm = document.querySelector('#meetings-join-moderator');
+const meetingsListCollectElm = document.querySelector('#meetings-list-collect');
+const meetingsListElm = document.querySelector('#meetings-list');
+const meetingsAddMediaElm = document.querySelector('#meetings-add-media');
+const meetingsLeaveElm = document.querySelector('#meetings-leave');
+const meetingsCurrentDetailsElm = document.querySelector('#meetings-current-details');
+
+function generateMeetingsListItem(meeting) {
+  const itemElm = document.createElement('div');
+  const joinElm = document.createElement('button');
+  const detailsElm = document.createElement('label');
+
+  itemElm.id = `meeting-list-item-${meeting.id}`;
+  itemElm.key = meeting.id;
+
+  joinElm.onclick = () => joinMeeting(meeting.id);
+  joinElm.type = 'button';
+  joinElm.value = meeting.id;
+  joinElm.innerHTML = 'meeting.join()';
+
+  detailsElm.innerHTML = meeting.destination ||
+    meeting.sipUri ||
+    meeting.id;
+
+  itemElm.appendChild(joinElm);
+  itemElm.appendChild(detailsElm);
+
+  return itemElm;
+}
+
+function collectMeetings() {
+  console.log('MeetingsManagement#collectMeetings()');
+
+  webex.meetings.syncMeetings()
+    .then(() => new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


packages/node_modules/samples/browser-meetings/app.js
```diff
@@ -83,3 +83,117 @@ function unregister() {
...
+      }
+
+      meetingsListElm.appendChild(
+        generateMeetingsListItem(meeting)
+      );
+    });
+}
+
+function joinMeeting(meetingId) {
+  const meeting = webex.meetings.getAllMeetings()[meetingId];
```
brunogenaro, line 166: If this fails, the code won't be able to access the index of the array [meetindId]. Can we add some conditionals of try/catch to avoid this?

packages/node_modules/samples/browser-meetings/app.js
```diff
@@ -83,3 +83,117 @@ function unregister() {
...
+    moveToResource: false
+    // resourceId: meetingsJoinDeviceElm.value // Add this after the devices section is ready
+  })
+    .then(() => {
+      meetingsCurrentDetailsElm.innerHTML = meeting.destination ||
+        meeting.sipUri ||
+        meeting.id;
+
+      meetingsLeaveElm.onclick = () => leaveMeeting(meeting.id);
+    });
```
brunogenaro, line 180: Do we need to address the failure state adding a `catch` block here?

packages/node_modules/samples/browser-meetings/app.js
```diff
@@ -83,3 +83,117 @@ function unregister() {
...
+  // This method will be filled when generating the Meeting Controls Sources
+  // are implemented.
+}
+
+function leaveMeeting(meetingId) {
+  if (!meetingId) {
+    return;
+  }
+
+  const meeting = webex.meetings.getAllMeetings()[meetingId];
```
brunogenaro, line 193: If this fails, the code won't be able to access the index of the array [meetindId]. Can we add some conditionals of try/catch to avoid this?

packages/node_modules/samples/browser-meetings/app.js
```diff
@@ -83,3 +83,117 @@ function unregister() {
...
+      meetingsListElm.innerHTML = '';
+
+      Object.keys(meetings).forEach(
+        (key) => {
+          meetingsListElm.appendChild(
+            generateMeetingsListItem(meetings[key])
+          );
+        }
+      );
+    });
```
brunogenaro, line 149: Do we need to address the failure state adding a `catch` block here?

packages/node_modules/samples/browser-meetings/app.js
```diff
@@ -83,3 +83,117 @@ function unregister() {
...
+  if (!meetingId) {
+    return;
+  }
+
+  const meeting = webex.meetings.getAllMeetings()[meetingId];
+
+  meeting.leave()
+    .then(() => {
+      meetingsCurrentDetailsElm.innerHTML = 'Not currently in a meeting';
+    });
```
brunogenaro, line 198: Do we need to address the failure state adding a `catch` block here?

packages/node_modules/samples/browser-meetings/app.js
InteractiveTimmy: `getAllMeetings()` always returns an object. I added error handling in cases that the meetingId is invalid or no longer exists.

packages/node_modules/samples/browser-meetings/app.js
InteractiveTimmy: If an error is thrown here, it will present itself in the developer console. This is the intended behavior. This is the plan for all errors within the new meetings sample.

packages/node_modules/samples/browser-meetings/app.js
InteractiveTimmy: I've added error handling to this method to validate the meeting is found. `webex.meetings.getAllMeetings()` always returns an object.

packages/node_modules/samples/browser-meetings/app.js
InteractiveTimmy: This error should properly present itself in the console, which is the intended behavior.

packages/node_modules/samples/browser-meetings/app.js
InteractiveTimmy: We do not, as this is a kitchen-sink style sample, and the developer will need to check the console to inspect the error thrown.

PR #1685, opened 2020-05-27
Add `meeting.shareScreen` function

adamweeks:
# Pull Request Template

## Description

In order to get screen sharing on safari to work, we needed to have less of a promise chain above the actual navigator call. This was causing issues on safari because it lost track that it was called via a user action (security). 

I've added a `meeting.shareScreen` function that reduces the steps to actually sharing a screen.

Samples and documentation have also been updated.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-141969

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/media/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/media/index.js
@@ -320,7 +320,15 @@ Media.getDisplayMedia = (options, config) => {
       });
   }
 
-  return navigator.mediaDevices.getDisplayMedia({audio: options.sendAudio, video: options.sendShare ? shareConstraints : false});
+  const getDisplayMediaParams = {video: options.sendShare ? shareConstraints : false};
+
+  // safari doesn't support audio
+  // https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia
+  if (options.sendAudio && bowser.name.toLowerCase() !== 'safari') {
+    getDisplayMediaParams.audio = options.sendAudio;
+  }
+
+  return navigator.mediaDevices.getDisplayMedia(getDisplayMediaParams);
 };
 
 /**
--- a/packages/node_modules/@webex/plugin-meetings/src/media/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/media/index.js
@@ -276,12 +276,15 @@ Media.updateTransceiver = ({meetingId, remoteQualityLevel}, peerConnection, tran
 
 /**
  * generates share streams
- * @param {object} options parameter
- * @param {Object} options.sendAudio sendAudio: {Boolean} sendAudio constraints
- * @param {Object} options.sendShare sendShare: {Boolean} sendShare constraints
- * @param {Object} options.sharePreferences sharePreferences: {Object} Share constraints and share constraints
+ * @param {Object} options parameter
+ * @param {Boolean} options.sendAudio send audio from the display share
+ * @param {Boolean} options.sendShare send video from the display share
+ * @param {Object} options.sharePreferences
+ * @param {MediaTrackConstraints} options.sharePreferences.shareConstraints constraints to apply to video
+ *   @see {@link https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints}
+ * @param {Boolean} options.sharePreferences.highFrameRate if shareConstraints isn't provided, set default values based off of this boolean
  * @param {Object} config SDK Configuration for meetings plugin
- * @returns {Object} {streams}
+ * @returns {Promise.<MediaStream>}
  */
 Media.getDisplayMedia = (options, config) => {
   const shareConstraints = options.sharePreferences && options.sharePreferences.shareConstraints || {
@@ -322,7 +325,7 @@ Media.getDisplayMedia = (options, config) => {
 
   const getDisplayMediaParams = {video: options.sendShare ? shareConstraints : false};
 
-  // safari doesn't support audio
+  // safari doesn't support sending screen share audio
   // https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia
   if (options.sendAudio && bowser.name.toLowerCase() !== 'safari') {
     getDisplayMediaParams.audio = options.sendAudio;
--- a/packages/node_modules/@webex/plugin-meetings/src/media/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/media/index.js
@@ -10,6 +10,7 @@ import {
   VIDEO_INPUT,
   PEER_CONNECTION_STATE
 } from '../constants';
+import Config from '../config';
 import PeerConnectionManager from '../peer-connection-manager';
 import ReconnectionError from '../common/errors/reconnection';
 import MediaError from '../common/errors/media';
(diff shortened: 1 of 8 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2940,4 +2940,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  * @param {MediaTrackConstraints} options.sharePreferences.shareConstraints constraints to apply to video #see https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints
+  * @param {Boolean} options.sharePreferences.highFrameRate if shareConstraints isn't provided, set default values based off of this boolean
+  * @returns {Promise.<MediaStream>}
+  */
+  shareScreen(options = {sendShare: true, sendAudio: false}) {
+    LoggerProxy.logger.log('Meeting:index#shareScreen --> Getting local share');
+
+    return Media.getDisplayMedia(options).then((shareStream) => {
+      this.updateShare({
+        sendShare: true,
```
mwegman, line 2958: Should this match the parameter passed in through the 'options' object?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -410,26 +410,19 @@ skipInNode(describe)('plugin-meetings', () => {
...
-              console.log('SCREEN SHARE RESPONSE ', JSON.stringify(response));
-            }),
-          testUtils.waitForEvents([{scope: alice.meeting, event: 'media:ready'}])
-            .then((response) => {
-              console.log('MEDIA:READY event ', response[0].result);
-              assert.equal(response[0].result.type === 'localShare', true);
-            })
-        ]))
+      it('alice shares the screen with highFrameRate', () => Promise.all([
+        testUtils.delayedPromise(alice.meeting.shareScreen({sharePreferences: {highFrameRate: true}})),
```
mwegman, line 414: Did you mean to leave out 'sendShare' and 'sendAudio' from your parameter object?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: No. The naming of the params unfortunately collides and the `sendShare` that is sent to `getDisplayMedia` is not the same as the `sendShare` to `updateShare`. `updateShare` tells the meeting that we are about to share the stream passed in as `shareStream`.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
adamweeks: `shareScreen` defaults `sendShare` and `sendAudio` appropriately so it is not needed here.

mwegman: Do you mean 'Media.getDisplayMedia' defaults correctly. Because this object passed into 'shareScreen' will override the one set by default.

adamweeks: Oh, I see what you mean, I need to update the defaults code. Good catch!

adamweeks: @mwegman I've updated the code

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -251,26 +251,8 @@ document.getElementById('share-screen').addEventListener('click', () => {
...
-          console.info('SHARE-SCREEN: Add local share via `updateShare`');
-
-          return activeMeeting.updateShare({
-            sendShare: true,
-            receiveShare: true,
-            stream: localShare
-          });
-        })
+      console.info('SHARE-SCREEN: Sharing screen via `shareScreen()`');
+      activeMeeting.shareScreen()
```
lalli-flores, line 255: So much simpler 🙌🏻

taymoork2:
Pull in master with the ability to test with safari and (besides the general flakiness) everything passed

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: @lalli-flores there's plenty of developer friendly improvements that could be made to the meetings plugin, but outside the scope of this ticket

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -316,7 +318,15 @@ Media.getDisplayMedia = (options) => {
       });
   }
 
-  return navigator.mediaDevices.getDisplayMedia({audio: options.sendAudio, video: options.sendShare ? shareConstraints : false});
+  const getDisplayMediaParams = {video: options.sendShare ? shareConstraints : false};
+
+  // safari doesn't support audio
+  // https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia
+  if (options.sendAudio && bowser.name.toLowerCase() !== 'safari') {
```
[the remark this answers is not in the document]
adamweeks: We already support this, just that safari does not support sharing computer audio.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -534,7 +534,7 @@ describe('plugin-meetings', () => {
       });
       describe('#share', () => {
         it('should have #share', () => {
-          assert.exists(meeting.invite);
+          assert.exists(meeting.share);
```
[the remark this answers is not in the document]
adamweeks: There are tests that verify the functions exist in this file. This one was incorrectly checking the wrong function.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2940,4 +2940,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  /**
+  * @param {Object} options parameter
+  * @param {Boolean} options.sendAudio send audio from the display share
+  * @param {Boolean} options.sendShare send video from the display share
+  * @param {Object} options.sharePreferences
+  * @param {MediaTrackConstraints} options.sharePreferences.shareConstraints constraints to apply to video #see https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints
+  * @param {Boolean} options.sharePreferences.highFrameRate if shareConstraints isn't provided, set default values based off of this boolean
+  * @returns {Promise.<MediaStream>}
+  */
+  shareScreen(options = {}) {
```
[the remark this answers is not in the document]
adamweeks: As @mwegman discovered, it is possible to send an object that wouldn't have either of those like we do in the journey test and then those values would no exist in the options:

```js
alice.meeting.shareScreen({sharePreferences: {highFrameRate: true}})
```

We have no eslint rules regarding defaults or required patterns. Some parts of the code use the `defaults` function from lodash, but I prefer the `object.assign` method.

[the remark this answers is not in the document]
mwegman: If you wanted defaults and except more options you could do the following
```suggestion
  shareScreen({sendShare: true, sendAudio: false, ...args}) {
```
then you can do below:
``` javascript
const getDisplayMediaOptions = {...args, sendShare, sendAudio};
```

packages/node_modules/@webex/plugin-meetings/src/media/index.js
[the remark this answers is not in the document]
lalli-flores: Then let's make sure that's documented somewhere developers can see

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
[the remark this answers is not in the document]
lalli-flores: If they are not needed, then let's get rid of them? And only default the high frame rate.
```suggestion
  shareScreen(options = {sharePreferences: {highFrameRate: true}}) {
```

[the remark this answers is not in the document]
adamweeks: `sendShare` and `sendAudio` are definitely needed. That's why I'm defaulting them. The `sharePreferences` are optional.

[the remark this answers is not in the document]
lalli-flores: Misunderstand your comment:

> t is possible to send an object that wouldn't have either of those like we do in the journey test and then those values would no exist in the options:

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -275,12 +275,15 @@ Media.updateTransceiver = ({meetingId, remoteQualityLevel}, peerConnection, tran
...
+ * @param {Boolean} options.sendShare send video from the display share
+ * @param {Object} options.sharePreferences
+ * @param {MediaTrackConstraints} options.sharePreferences.shareConstraints constraints to apply to video
+ *   @see {@link https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints}
+ * @param {Boolean} options.sharePreferences.highFrameRate if shareConstraints isn't provided, set default values based off of this boolean
  * @param {Object} config SDK Configuration for meetings plugin
- * @returns {Object} {streams}
+ * @returns {Promise.<MediaStream>}
  */
 Media.getDisplayMedia = (options, config) => {
```
taymoork2, line 288: @adamweeks integration tests are failing due to config not having defaults.
In the scenario where **only** `highFrameRate` is passed this function will fail due to config being undefined.
Now setting defaults would be simple, but the problem on top of that would be that both ternary values rely on a `config`. 
the reason this didn't get all before was because `getMediaStreams` was grabbing a bunch of `sharePreferences` props before calling this function (somewhat indirectly).

Also `const shareConstraints = options.sharePreferences && options.sharePreferences.shareConstraints || {` makes no sense and is probably too "busy" to understand on 1st-3rd glance

PR #1686, opened 2020-05-27
feat(meetings): pass sdk config to media functions

adamweeks:
Users were not able to override the default configuration and
the media utilities were taking the hard coded config from config.js

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-145738

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/media/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/media/index.js
@@ -13,7 +13,6 @@ import {
 import PeerConnectionManager from '../peer-connection-manager';
 import ReconnectionError from '../common/errors/reconnection';
 import MediaError from '../common/errors/media';
-import config from '../config';
 
 
 /**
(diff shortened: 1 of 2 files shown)


adamweeks:
Tested by updating the sample to use:

```js
webex = window.webex = Webex.init({
        credentials: {
          access_token: document.getElementById('access-token').value
        },
        config: {
          meetings: {
            resolution: {
              maxWidth: 320,
              maxHeight: 180,
              idealWidth: 320,
              idealHeight: 180
            }
          }
        }
      });
```

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -103,15 +102,16 @@ Media.generateLocalMedias = (mediaId, audioMuted, videoMuted) => {
 /**
  * make a browser call to get the media
  * @param {SendOptions} options
+ * @param {Object} config SDK Configuration for meetings plugin
  * @returns {Promise}
  */
-Media.getLocalMedia = (options) => {
+Media.getLocalMedia = (options, config) => {
```
mwegman, line 108: You could add the 'config' object to the 'options' object instead.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -277,15 +278,16 @@ Media.updateTransceiver = ({meetingId, remoteQualityLevel}, peerConnection, tran
  * @param {Object} options.sendAudio sendAudio: {Boolean} sendAudio constraints
  * @param {Object} options.sendShare sendShare: {Boolean} sendShare constraints
  * @param {Object} options.sharePreferences sharePreferences: {Object} Share constraints and share constraints
+ * @param {Object} config SDK Configuration for meetings plugin
  * @returns {Object} {streams}
  */
-Media.getDisplayMedia = (options) => {
+Media.getDisplayMedia = (options, config) => {
```
mwegman, line 284: Same as above

mwegman:
I'm not going to hold up this PR for my simple changes.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
adamweeks: The media class could stand to be refactored a bit so we don't have to pass the config, but stores it in the class itself. I was hoping having it separate would make it easier to refactor.

PR #1696, opened 2020-06-04
Meetings Upload Log Capabilities

adamweeks:
## Description

Ability to upload SDK logs via the meetings plugin.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
@@ -1,0 +1,44 @@
+import uuid from 'uuid';
+
+import LoggerProxy from './logger-proxy';
+
+/**
+ * @class LogsRequest
+ */
+export default class LogsRequest {
+  /**
+   *
+   * @param {Object} options
+   * @param {Object} options.webex Webex SDK instance
+   * @param {ClientMetrics} metrics
+   */
+  constructor(options) {
+    this.webex = options.webex;
+  }
+
+  /**
+   * Uploads logs to the support server
+   *
+   * @param {Object} [options={}]
+   * @param {String} [options.feedbackId=uuid.v4] ID used for tracking
+   * @param {String} [options.locusId]
+   * @param {Stirng} [options.correlationId]
+   * @returns {Promise.<String>} Feedback ID
+   * @memberof LogsRequest
+   */
+  async uploadLogs(options = {}) {
+    const id = options.feedbackId || uuid.v4();
+    const {locusId, correlationId} = options;
+
+    LoggerProxy.logger.info(`LogsRequest->uploadLogs# uploading user logs for feedbackId: ${id}`);
+
+    try {
+      await this.webex.internal.support.submitLogs({feedbackId: id, locusId, correlationId});
+    }
+    catch (error) {
+      LoggerProxy.logger.error('LogsRequest->uploadLogs# uploading user logs failed', error);
+    }
+
+    return id;
+  }
+}
--- a/packages/node_modules/samples/browser-plugin-meetings/app.js
+++ b/packages/node_modules/samples/browser-plugin-meetings/app.js
@@ -35,6 +35,9 @@ function connect() {
     // eslint-disable-next-line no-multi-assign
     webex = window.webex = Webex.init({
       config: {
+        logger: {
+          level: 'debug'
+        },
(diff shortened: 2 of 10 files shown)


packages/node_modules/samples/browser-plugin-meetings/app.js
```diff
@@ -767,6 +770,11 @@ document.getElementById('resume-recording').addEventListener('click', () => {
   }
 });
 
+document.getElementById('upload-logs').addEventListener('click', () => {
```
brunogenaro, line 773: Should you pass `event` as parameter?
```suggestion
document.getElementById('upload-logs').addEventListener('click', (event) => {
```

packages/node_modules/samples/browser-plugin-meetings/index.html
```diff
@@ -346,6 +346,7 @@ <h2>Then click:</h2>
       <canvas id="myChart2"></canvas>
     </div>
         <h2>Extras</h2>
+        <button id="upload-logs">uploadLogs</button>
```
brunogenaro, line 349: Or you could add `type="button"` to not send the form, instead of `event.preventDefault();`
```suggestion
        <button id="upload-logs" type="button">uploadLogs</button>
```

chrisadubois:
references: https://github.com/webex/webex-js-sdk/compare/master...chrisadubois:logging_meeting_errors?expand=1

packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
```diff
@@ -0,0 +1,44 @@
...
+   * @memberof LogsRequest
+   */
+  async uploadLogs(options = {}) {
+    const id = options.feedbackId || uuid.v4();
+    const {locusId, correlationId} = options;
+
+    LoggerProxy.logger.info(`LogsRequest->uploadLogs# uploading user logs for feedbackId: ${id}`);
+
+    try {
+      await this.webex.internal.support.submitLogs({feedbackId: id, locusId, correlationId});
```
chrisadubois, line 36: this feedbackId *could* also be a correlationId, which would centralize our ids

packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
```diff
@@ -0,0 +1,44 @@
...
+   * @param {Object} [options={}]
+   * @param {String} [options.feedbackId=uuid.v4] ID used for tracking
+   * @param {String} [options.locusId]
+   * @param {Stirng} [options.correlationId]
+   * @returns {Promise.<String>} Feedback ID
+   * @memberof LogsRequest
+   */
+  async uploadLogs(options = {}) {
+    const id = options.feedbackId || uuid.v4();
+    const {locusId, correlationId} = options;
```
mwegman, line 31:
```suggestion
  async uploadLogs({feedbackId: id = uuid.v4(), locusId, correlationId}) {
```

packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
mwegman, line 36: Why did we rename `id` if we were going to rename it back to `feedbackId` for this call?

packages/node_modules/samples/browser-plugin-meetings/index.html
mwegman: `button` and `type=button` are the same thing. `type=button` is nice if we wanted to use a `span` or `div` here.

brunogenaro: @mwegman  In this case is true because we are not inside a `<form>` element, otherwise, the form would submit if we don't specify `type="button"`. @adamweeks can you juts pass the `event` as parameter in order to use it in `event.preventDefault()` then? I'm pretty sure that will break the code since `event` is undefined.

packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
arun3528: ya but a suer can submit logs without correlation ID or user sends logs as well as we send the log on call end

adamweeks: @chrisadubois good call, I'll have the feedbackid set to correlation id when there is one like at the end of a meeting.

PR #1698, opened 2020-06-05
feat(meetings): add call start to logs

adamweeks:
We missed adding the call start time to logs and it is helpful when correlating meetings.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
@@ -469,7 +469,10 @@ export default class Meetings extends WebexPlugin {
     destroy(meeting, reason) {
       if (this.config.autoUploadLogs) {
         LoggerProxy.logger.info('Meetings->index#destroy Meeting destroyed, uploading logs');
+        const callStart = meeting.locusInfo && meeting.locusInfo.fullState ? meeting.locusInfo.fullState.lastActive : undefined;
+
         this.uploadLogs({
+          callStart,
           correlationId: meeting.correlationId,
           feedbackId: meeting.correlationId,
           locusId: meeting.locusId


packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -469,7 +469,10 @@ export default class Meetings extends WebexPlugin {
     destroy(meeting, reason) {
       if (this.config.autoUploadLogs) {
         LoggerProxy.logger.info('Meetings->index#destroy Meeting destroyed, uploading logs');
+        const callStart = meeting.locusInfo ? meeting.locusInfo.fullState.lastActive : undefined;
```
abailey-wwt, line 472: I see that you are protecting against `locusInfo` not being present. Is there a possibility that `fullState` will always be present if `locusInfo` is present? Otherwise logs might not get uploaded due to that error.

adamweeks: Good catch @abailey-wwt we actually always have a `locusInfo` object on the meeting, but the fullstate isn't always there. Will update.

abailey-wwt: Cool. Also we only support latest evergreen browsers correct? Perhaps we can introduce optional chaining? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

mwegman: If I remember correctly, the only reason we don't do optional chaining is because the SDK's version of babel doesn't do it. We plan on doing it in SDK v2.

adamweeks:
Failures due to screen share peerconnection. Merging.

PR #1713, opened 2020-06-09
Ice peer connection events

adamweeks:
In order to stabilize the reconnection logic, we need to disable the auto reconnect based on mercury connection. This is the first step in getting peer connection reconnects working properly.

Additional changes are adding logs for all the peer connection events as well as auto uploading logs after calls at the request of @jpjpjp


arun3528:
lgtm

PR #1716, opened 2020-06-11
Send getUserMedia Failures as Metric

lalli-flores:
A couple refactors removing unnecessary promise-layers.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -41,6 +41,7 @@ import {
   STATS,
   MQA_STATS,
   MEETING_STATE,
+  METRICS_OPERATIONAL_MEASURES,
   FULL_STATE,
   MEETING_REMOVED_REASON,
   SDP,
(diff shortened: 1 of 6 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2206,16 +2207,25 @@ export default class Meeting extends StatelessWebexPlugin {
             audioVideo,
             sharePreferences,
             this.config
-          ))
-        .then((response) => {
-          if (!response[0] && !response[1]) {
-            return Promise.reject(
-              new MediaError('Was not able to retrieve media Stream')
-            );
-          }
```
mwegman, line 2215: So this check is no longer necessary? We won't get a successful return that has no response?

lalli-flores: The error handling flow is a little different and hence why we don't need this check.

This is the call flow for getting the user's media:
`Meeting.getMediaStreams -> Media.getUserMedia -> Media.getLocalMedia -> navigator.mediaDevices.getUserMedia`

And this the error handling change:
In `navigator.mediaDevices.getUserMedia`:
* Catches error
* Logs to logger as error
* Re-throws error as-is

In `Media.getLocalMedia`:
* No handling, returns `navigator.mediaDevices.getUserMedia` (e.g. relies on its `catch`)

In `Media.getUserMedia`:
* No handling, `Media.getLocalMedia` is part of a promise chain, so it it rejects it will propagate

In `Meeting.getMediaStreams`, this is the change:
* Because the in `Media.getUserMedia` error will propagate, we catching here
* Send metric signaling failures to get user media
* Re-throws as custom SDK error `MediaError`

Issue #1467, opened 2020-01-15
Logging Proposal for plugin-logger


chrisadubois:
SDK Logging Proposal

Current State
SDK wraps the console logs, and sends its own logs to a buffer, and sends consumers of the SDK logs to the same buffer. This buffer size is limited by a config value in plugin-logger which defaults to 1000 submissions. The logs get submitted through plugin-support and send to the client logs service.

Requirements
Clients of the SDK can differentiate their logs from SDK logs.

Proposal
A new configuration value, clientBufferSize is added to plugin-logger
A new configuration value, clientLevel is added to plugin-logger
Plugin-logger gets a new series of levels and wrapped console methods, prefixed with ‘client’
(which means no code changes for any consumers of the SDK today except those who want to use the new buffer, i.e., web-client)
Plugin-logger prefixes all SDK logs with ‘wx-js-sdk:’ and prefixes all client logs with ‘client:’ so that eventually we can coalesce the time periods once we have a streaming log service for javascript clients in cisco
Plugin-support or plugin-logger formats the buffers for client log submission “better” – this is an unknown
Plugin-support adds a new configuration value, logSubmission, which can be set to ‘client’, ‘sdk’, or ‘both’
Plugin-support combines the two buffers for submission to clientlogs (atlas) service, noting them as different based on the configuration value

adamweeks:
@chrisadubois proposal is good, feel free to begin work

PR #1717, opened 2020-06-11
feat(logger): logging improvements

glenatwork:
See https://github.com/webex/webex-js-sdk/issues/1467

# Pull Request Template

## Description

Pass logger to network methods, log network http return values

Return headers from request (esp on failure) so client can inspect, log them

Lower a few log levels in storage-adapter-local-storage.  These are very chatty in production without adding much info.

Separate client/SDK log buffers, log levels

Offer log method to only log to buffer

Use iso format for buffer date

See https://github.com/webex/webex-js-sdk/issues/1467

## Type of Change

## Test Coverage

UT's added to logger for new behavior.  Tested in Teams web client, including sending logs to Atlas and downloading

**Test Configuration**:
* SDK Version this
* Node 10 Chrome 83.0.4103.106 Mac
* NPM Version yarn 1.16.0

## Checklist:

- [?] I have made corresponding changes to the documentation


packages/node_modules/@webex/plugin-logger/src/logger.js
```diff
@@ -195,34 +209,94 @@ const Logger = WebexPlugin.extend({
...
 
+/**
+ * Creates a logger method
+ *
+ *
+ * @param {string} level level to create (info, error, warn, etc.)
+ * @param {string} impl the level to use when writing to console
+ * @param {string} type type of log, SDK or client
+ * @param {string} neverPrint function never prints to console
+ * @param {string} alwaysBuffer function always logs to log buffer
```
mwegman, line 271: These are booleans

packages/node_modules/@webex/plugin-logger/src/logger.js
```diff
@@ -195,34 +209,94 @@ const Logger = WebexPlugin.extend({
...
+ * @instance
+ * @memberof Logger
+ * @private
+ * @memberof Logger
+ * @returns {function} logger method with specified params
+ */
+function makeLoggerMethod(level, impl, type, neverPrint = false, alwaysBuffer = false) {
```
mwegman, line 278: I noticed `neverPrint` and `alwaysBuffer` are either both false or both true. Would there ever be a case were this isn't true?

packages/node_modules/@webex/plugin-logger/src/logger.js
glenatwork: thanks.  fixed.

packages/node_modules/@webex/plugin-logger/src/logger.js
glenatwork: I offered neverPrint separately because I figure it should only be used in extreme cases.  In the Teams web client case, we are overriding console methods and sending them to the buffered logs.  In that case, a console print would be a severe problem, since it would cause infinite recursion.  So we want to suppress any possibility of a console log.  Other cases might not want to suppress logging so severely, although, true, I didn't implement such methods.

glenatwork: @mwegman Are you OK with resolving this conversation?

packages/node_modules/@webex/webex-core/src/webex-core.js
```diff
@@ -544,7 +544,8 @@ const WebexCore = AmpState.extend({
         return p;
       })
       .then((...args) => this._uploadPhaseFinalize(options, ...args))
-      .then((res) => res.body);
+      .then((res) => ({...res.body, ...res.headers}))
+      .catch((err) => err);
```
mwegman, line 548: Why are you catching this just to return 'err'? Is this not a real error?

chrisadubois: I think we can just remove the .catch here now that we've added code to handle it.

packages/node_modules/@webex/http-core/src/request/request.shim.js
```diff
@@ -42,6 +45,12 @@ export default function _request(options) {
 
       /* istanbul ignore else */
       if (response) {
+        if (response.statusCode >= 400) {
```
chrisadubois, line 48: We might need to add something here for response.statusCode <= 0

packages/node_modules/@webex/webex-core/src/webex-core.js
glenatwork: ok yeah instead of re-throwing or something, I'll just remove it.

packages/node_modules/@webex/http-core/src/request/request.shim.js
mwegman: I don't think it will be less than zero but it could be zero

glenatwork: This should not be an issue here.  The changing it to zero occurs a few lines below this code in the else:
      else {
        resolve({
          statusCode: 0,
          options,
          headers: options.headers,
          method: options.method,
          url: options.uri,
          body: error
        });
      }
In that case, hopefully we were able to log the error with the warning above this:
      if (error) {
        options.logger.warn(error);
      }
But if we didn't log it there (ie error is not truthy), the thing is, there's really nothing to log, there was no response to log (this if block is checking the existence of response).  My goal here was to log the network level response.

PR #1726, opened 2020-06-15
feat(meetings): change stats analyzer log level

adamweeks:
Starting the collection of logs doesn't need to be "info" level.


mwegman:
Looks good. Getting the tests to pass before merging

PR #1727, opened 2020-06-15
feat(plugin-meetings): add ice disconnect timeout

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request is to add timeout logic to the ice disconnect event tree and remove all current logic from the `meeting.reconnect` method except for general logging.

Fixes # [SPARK-151805](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-151805)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/config.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/config.js
@@ -16,6 +16,8 @@ export default {
     reconnection: {
       enabled: true,
       detection: true,
+      // Timeout duration to wait for ICE to reconnect if a disconnect is received.
+      iceReconnectionTimeout: 10000,
       retry: {
         times: 2,
         backOff: {
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -1909,6 +1909,8 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   reconnect() {
+    LoggerProxy.logger.log('meeting:index#reconnect --> attempting to reconnect');
+
     if (!this.reconnectionManager || !this.reconnectionManager.reconnect) {
       throw new ParameterError('Cannot reconnect, ReconnectionManager must first be defined.');
     }
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
@@ -675,12 +675,20 @@ MeetingUtil.setPeerConnectionEvents = (meeting) => {
         Metrics.postEvent({event: eventType.ICE_END, meeting});
         break;
       case ICE_STATE.CONNECTED:
+        meeting.reconnectionManager.iceReconnected();
         LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE CONNECTED.');
         break;
       case ICE_STATE.CLOSED:
         LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE CLOSED.');
         break;
       case ICE_STATE.DISCONNECTED:
+        meeting.reconnectionManager.waitForIceReconnect()
+          .catch(() => {
+            LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE DISCONNECTED. Automatic Reconnection Timed Out.');
+
+            meeting.reconnect();
+          });
+
         LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE DISCONNECTED.');
         break;
       case ICE_STATE.FAILED:
--- a/packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
@@ -18,6 +18,21 @@ export default class ReconnectionManager {
    */
   constructor(meeting) {
     /**
+     * Stores ICE reconnection state data.
+     *
+     * @instance
+     * @type {Object}
+     * @private
+     * @memberof ReconnectionManager
+     */
+    this.iceState = {
+      disconnected: false,
+      resolve: () => {},
+      timer: undefined,
+      timeoutDuration: meeting.config.reconnection.iceReconnectionTimeout
+    };
+
+    /**
      * @instance
      * @type {String}
      * @private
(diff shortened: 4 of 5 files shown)


packages/node_modules/@webex/plugin-meetings/src/config.js
```diff
@@ -16,6 +16,7 @@ export default {
     reconnection: {
       enabled: true,
       detection: true,
+      iceReconnectionTimeout: 10000,
```
adamweeks, line 19: Could you add a quick comment saying what this does above it?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -22,7 +22,7 @@ import MediaUtil from '../media/util';
 import WebRTCStats from '../stats/index';
 import StatsMetrics from '../stats/metrics';
 import StatsUtil from '../stats/util';
-import ReconnectionError from '../common/errors/reconnection';
+// import ReconnectionError from '../common/errors/reconnection';
```
adamweeks, line 25: Commented out?

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -41,6 +54,50 @@ export default class ReconnectionManager {
...
+
+      this.iceState.waitForReconnection = new Promise((resolve, reject) => {
+        this.iceState.timer = setTimeout(() => {
+          reject(new Error(`ice reconnection did not occur in ${this.iceState.timeoutDuration}ms`));
+        }, this.iceState.timeoutDuration);
+
+        resolution = resolve;
+      });
+
+      this.iceState.waitForReconnection.resolve = resolution;
```
mwegman, line 96: Instead of doing it this way, you could strictly save the 'resolve' function and return the promise.
ICE_STATE.DISCONNECTED could attach the `.catch` to the return value of this function.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
InteractiveTimmy: This is no longer used and creates a linting error [unused vars], I can remove it if needed.

packages/node_modules/@webex/plugin-meetings/src/config.js
InteractiveTimmy: Good idea! will add this comment.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
InteractiveTimmy: Good idea, implementing this now. I will preserve the promise in the `iceState` object to allow deferred resolution within the class as well.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -41,6 +56,62 @@ export default class ReconnectionManager {
...
+
+      return this.iceState.waitForReconnection;
```
mwegman, line 109:
```suggestion
      return new Promise((resolve, reject) => {
        this.iceState.timer = setTimeout(() => {
          reject(new Error(`ice reconnection did not occur in ${this.iceState.timeoutDuration}ms`));
        }, this.iceState.timeoutDuration);

        this.iceState.waitForReconnection.resolve = resolve;
      });
```

mwegman: I was thinking you didn't need to keep a reference to the promise. I believe the promise will stay around because you are saving a reference to the `resolve` function of the promise and the timer has a reference to the `reject` function

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -675,12 +675,20 @@ MeetingUtil.setPeerConnectionEvents = (meeting) => {
...
+        meeting.reconnectionManager.iceReconnected();
         LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE CONNECTED.');
         break;
       case ICE_STATE.CLOSED:
         LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE CLOSED.');
         break;
       case ICE_STATE.DISCONNECTED:
+        meeting.reconnectionManager.iceDisconnected()
+          .catch(() => {
+            LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE FAILED RECONNECT. ATTEMPTING MANUAL RECONNECTION.');
```
adamweeks, line 687: Let's not use `FAILED` here since there is an actual ice state of FAILED. 

```suggestion
            LoggerProxy.logger.log('Meeting:util#setPeerConnectionEvents --> ICE STATE DISCONNECTED. Automatic Reconnection Timed Out.');
```

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -41,6 +56,58 @@ export default class ReconnectionManager {
...
+   */
+  iceDisconnected() {
+    if (!this.iceState.disconnected) {
+      LoggerProxy.logger.log('ReconnectionManager:index#iceDisconnected --> waiting for ice reconnect');
+
+      this.iceState.disconnected = true;
+
+      return new Promise((resolve, reject) => {
+        this.iceState.timer = setTimeout(() => {
+          reject(new Error(`ice reconnection did not occur in ${this.iceState.timeoutDuration}ms`));
```
adamweeks, line 100: Let's check `this.iceState.disconnected` one more time here before rejecting. It could be that we didn't clear the timer before it executed, even though we set `disconnected` to false. Probably an edge case but could help with race conditions.

mwegman: I don't know if it is necessary to check `disconnected`. Javascript isn't multithreaded. We should probably set `disconnected` to false like it is in `iceReconnected`

packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
InteractiveTimmy: Going to make this change in the next push.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
InteractiveTimmy: I can implement a getter on iceState that retrieves the current state if needed, let me know which is preferred @adamweeks and @mwegman

mwegman: You should add `this.iceState.disconnected = false;` after this line in the Promise so your reconnect logic will keep working after the first timer goes off.

InteractiveTimmy: Added.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -41,6 +56,64 @@ export default class ReconnectionManager {
...
+   * @memberof ReconnectionManager
+   */
+  waitForIceReconnect() {
+    if (!this.iceState.disconnected) {
+      LoggerProxy.logger.log('ReconnectionManager:index#waitForIceReconnect --> waiting for ice reconnect');
+
+      this.iceState.disconnected = true;
+
+      return new Promise((resolve, reject) => {
+        this.iceState.timer = setTimeout(() => {
```
adamweeks, line 99: `setTimeout` returns a timer ID which is cleared by `clearTimeout(timerID)`. `timer.clearTimeout()` isn't a function in this case.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -41,6 +56,64 @@ export default class ReconnectionManager {
...
+   */
+  iceReconnected() {
+    if (this.iceState.disconnected) {
+      LoggerProxy.logger.log('ReconnectionManager:index#iceReconnected --> ice has reconnected');
+
+      this.iceState.resolve();
+      this.iceState.resolve = () => {};
+
+      if (this.iceState.timer && this.iceState.timer.clearTimeout) {
+        this.iceState.timer.clearTimeout();
```
mwegman, line 75: clearTimeout(this.iceState.timer);

packages/node_modules/@webex/plugin-meetings/test/unit/spec/reconnection-manager/index.js
```diff
@@ -0,0 +1,143 @@
...
+        it('should return a promise that rejects after a duration', () => {
+          reconnectionManager.iceState.timeoutDuration = 100;
+
+          return assert.isRejected(reconnectionManager.waitForIceReconnect());
+        });
+
+        it('should resolve return a resolved promise when triggered', () => {
+          const promise = reconnectionManager.waitForIceReconnect();
+
+          reconnectionManager.iceState.resolve();
```
mwegman, line 120: `reconnectionManager.iceReconnected`

PR #1732, opened 2020-06-17
fix: update reachability check to not worry about intranet url

holsted:
# Pull Request Template
There is a bug related to a customer escalation that if the intranetInactivityCheckUrl isn't set, then the inNetwork timer never fires. 

## Description

Fixes # (issue)

## Type of Change

## Test Coverage

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* SDK Version
* Node/Browser Version
* NPM Version

## Checklist:


packages/node_modules/@webex/internal-plugin-device/src/device.js
```diff
@@ -528,6 +528,7 @@ const Device = WebexPlugin.extend({
     // Validate if the device has a intranet checking url.
     if (!this.intranetInactivityCheckUrl) {
       this.isInNetwork = false;
+      this.isReachabilityChecked = true;
```
mwegman, line 531: You could move this up out of the `if` statement and delete line 536.

andrewholsted:
@mwegman updated some existing tests. Let me know if you think we should add more here.

packages/node_modules/@webex/internal-plugin-device/test/unit/spec/device.js
```diff
@@ -66,6 +66,7 @@ describe('plugin-device', () => {
 
           it('should call \'checkNetworkReachability()\'', () => {
             assert.called(device.checkNetworkReachability);
+            assert.isTrue(device.isReachabilityChecked);
```
adamweeks, line 69: We try to limit unit test assertions to one per `it` so the assertion matches the description. I know that doesn't seem to be the case all over the sdk, but that is what we are aiming for. It would be better if you put this assertion in its own block, that way, if the unit test fails we know what specifically failed by the name of the block.

```js
it('should set isReachabilityChecked to true', () => {
  assert.isTrue(device.isReachabilityChecked);
});
```

PR #1735, opened 2020-06-18
fix(meetings): sendOperational metrics directly with request

lalli-flores:
This is, instead of using the metrics plugin
For some reason the metrics plugin cannot find its own members.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
@@ -423,7 +423,6 @@ class Metrics {
    * @returns {void}
    */
   sendOperationalMetric(metricName, metricFields = {}, metricTags = {}) {
-    const sendMetric = this.webex.internal.metrics.submitClientMetrics;
     const fields = {
       ...metricFields,
       browser_version: bowser.version,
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
```diff
@@ -433,16 +432,28 @@ class Metrics {
...
-      platform: bowser.name
+      os: bowser.osname
     };
 
     if (!metricName) {
       throw Error('Missing operational metric name. Please provide one');
     }
 
-    sendMetric(metricName, {type: ['operational'], fields, tags});
+    this.webex.request({
```
InteractiveTimmy, line 444: I'm a bit nervous about making this change away from `sendMetric()`. Do we happen to know why we cannot use the `sendMetric()` method? As in, do we have the specific error message being generated? I can attempt to chase this cause a bit in the meantime.

lalli-flores: Is there something specific about this change that makes you nervous?
This is doing basically what the metrics plugin sans the batching.

The issue is that metric's `submitClientMetrics` relies on other functions within the plugin, but once it's imported here, those other functions are not there.

lalli-flores:
@InteractiveTimmy, you're right in treating this as a hot-fix. Long-term we should use the metric's plugin but I just want meeting's metrics to start working right now 🙂

PR #1741, opened 2020-06-22
Reconnect options

adamweeks:


packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -175,60 +175,84 @@ export default class ReconnectionManager {
...
-        if (this.tryCount > meeting.config.reconnection.retry.times) {
-          this.status = RECONNECTION.STATE.FAILURE;
+      LoggerProxy.logger.log('ReconnectionManager:index#executeReconnection --> Media reestablished');
+      this.status = RECONNECTION.STATE.COMPLETE;
 
-          return Promise.reject(new ReconnectionError(err));
-        }
-        // exponential backoff and retry logic
-        setTimeout(() =>
-          this.reconnect(meeting),
```
chrisadubois, line 229: I couldn't view the document for some reason, are we completely removing the reconnect backoff for media and only doing it for joins? (i saw a todo in there for backoff on the join retry)

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -175,60 +175,84 @@ export default class ReconnectionManager {
...
-          .then(() => resolve())
-          .catch((error) => reject(error));
+
+    if (networkChange) {
+      // TODO: Calliope Discovery Needed (Out of band, no need to block execution)
+    }
+
+    if (networkDisconnect) {
+      try {
+        await this.reconnectMercuryWithBackoff();
```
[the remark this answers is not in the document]
adamweeks: img

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
adamweeks: Correct, if the media fails we are falling back to the join request flow.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -53,7 +53,18 @@ export default class ReconnectionManager {
...
     */
     this.webex = meeting.webex;
-    this.configure(meeting);
+    /**
+     * @instance
+     * @type {Meeting}
+     * @private
+     * @memberof ReconnectionManager
+     */
+    this.meeting = meeting;
```
arun3528, line 62: @adamweeks  would this not cause any issues when we change the actual meeting object ?. Also we will need a cleanup logic for meeting object on the parent object destroy !?

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -175,60 +175,84 @@ export default class ReconnectionManager {
...
 
-          return Promise.reject(new ReconnectionError(err));
-        }
-        // exponential backoff and retry logic
-        setTimeout(() =>
-          this.reconnect(meeting),
-        this.backOff *= meeting.config.reconnection.retry.backOff.rate);
-      }));
+      if (rejoinMeeting) {
+        // TODO: Emit event for reshare needed
```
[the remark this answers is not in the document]
adamweeks: No, we were told that if the meeting needed to be rejoined, we can't automatically reshare the screen and that we just need to emit an event to the developer that the user needs to reshare.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
adamweeks: Since it is an object, anything changed on the parent meeting object would be seen in the `this.meeting` reference. Not sure how you want to cleanup since the reconnection manager is a class instance that lives on the meeting object. When the meeting object is destroyed, so will the reconnection manager instance.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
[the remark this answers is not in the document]
chrisadubois: Oh nice that is awesome !! I remember trying that, so good to hear they have stated an answer

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
chrisadubois: Got it. and then if the join fails we are doing exponential backoff with a total cap limit of retries?

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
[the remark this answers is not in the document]
chrisadubois: :D

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
adamweeks: Right now we are going to to just keep attempting to rejoin. At this point, we should have network connectivity, so if we get a failure message from the rejoin with an error code saying we won't be able to rejoin, it will just stop the reconnect attempt.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
InteractiveTimmy: I confirmed that this is created and mounted in only one place: `/meeting/index.js`. There shouldn't be any location that persistently stores `meeting.reconnectionManager` based on my investigation.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -136,99 +139,114 @@ export default class ReconnectionManager {
...
-   * @param {Meeting} meeting
    * @returns {Boolean}
    * @returns {ReconnectionError}
    * @private
    * @memberof ReconnectionManager
    */
-  validate(meeting) {
-    // TODO: re write the validate function
-    if (meeting.config.reconnection.enabled) {
+  validate() {
```
InteractiveTimmy, line 147: API breaking, but only seems to be referenced by `this` object.

chrisadubois: validate is listed as a private function :)

PR #1745, opened 2020-06-23
feat(meetings): move metrics inside reconnection manager

adamweeks:


packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -176,7 +178,26 @@ export default class ReconnectionManager {
     const validatedReconnect = this.validate();
 
     if (validatedReconnect === true) {
```
mwegman, line 180:
```suggestion
    if (validatedReconnect) {
```

adamweeks: `this.validate()` returns true or an error. I feel like this is a bit of an antipattern so I'm going to rework the validate function into one that throws an error.

PR #1753, opened 2020-06-26
Remove reachability in reconnect

adamweeks:


packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -171,76 +171,67 @@ export default class ReconnectionManager {
...
+        throw reconnectError;
+      });
   }
 
   /**
    * @param {Object} reconnectOptions
-   * @param {boolean} [reconnectOptions.networkChange=false] indicates if a network change event happened
    * @param {boolean} [reconnectOptions.networkDisconnect=false] indicates if a network disconnect event happened
    * @returns {Promise}
    * @private
```
InteractiveTimmy, line 227: Does this still return a promise? Should we add JSDoc for the `throws` here?

adamweeks: async function always returns a promise. Will add the throws here.

PR #1766, opened 2020-06-30
feat(meetings): emit triggers on reconnect completion

adamweeks:
Fixing triggers and unit tests around `meeting.reconnect()`.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
@@ -875,41 +875,51 @@ describe('plugin-meetings', () => {
           assert.exists(meeting.reconnect);
         });
         describe('successful reconnect', () => {
-          it('should reconnect and trigger reconnection and return a promise', async () => {
-            meeting.reconnectionManager = new ReconnectionManager({config: {reconnection: {retry: {backOff: 1}}}});
-            meeting.reconnectionManager.reconnect = sinon.stub().returns(Promise.resolve(test1));
+          beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -875,41 +875,59 @@ describe('plugin-meetings', () => {
...
+            }
+            catch (error) {
               assert.calledWith(
                 TriggerProxy.trigger,
                 sinon.match.instanceOf(Meeting),
                 {file: 'meeting/index', function: 'reconnect'},
                 'meeting:reconnectionFailure',
                 {error: sinon.match.any}
               );
+            }
```
mwegman, line 921:
```suggestion
            await expect(meeting.reconnect()).to.be.rejected;
            assert.calledWith(
              TriggerProxy.trigger,
              sinon.match.instanceOf(Meeting),
              {file: 'meeting/index', function: 'reconnect'},
              'meeting:reconnectionFailure',
              {error: sinon.match.any}
            );
```

mwegman: This way it will fail if an exception isn't thrown.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -875,41 +875,59 @@ describe('plugin-meetings', () => {
...
+          });
+
+          it('should reset after an unsuccessful reconnection', async () => {
+            try {
```
mwegman, line 925: Same as above

PR #1770, opened 2020-07-01
fix(plugin-meetings): convo service lookup

InteractiveTimmy:
# Pull Request

## Description

The scope of the changes in this pull request are to address the invalid conversation serviceUrl lookup process used within the meetings plugin when creating a meeting and joining it.

[SPARK-157390](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-157390)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
@@ -30,7 +30,6 @@ import {
   MEET,
   MEET_M,
   HTTPS_PROTOCOL,
-  CONVERSATIONS,
   UUID_REG
 } from '../constants';
 
@@ -126,14 +125,14 @@ MeetingInfoUtil.isPhoneNumber = (phoneNumber) => {
 };
 
 MeetingInfoUtil.getHydraId = (destination) => {
-  const decodedDestination = deconstructHydraId(destination);
+  const {type, id, cluster} = deconstructHydraId(destination);
 
-  if (decodedDestination && UUID_REG.test(decodedDestination.id)) {
-    if (decodedDestination.type === _ROOM_) {
-      return {room: true, destination: decodedDestination.id};
+  if (id && UUID_REG.test(id)) {
+    if (type === _ROOM_) {
+      return {room: true, destination: id, cluster};
     }
-    if (decodedDestination.type === _PEOPLE_) {
-      return {people: true, destination: decodedDestination.id};
+    if (type === _PEOPLE_) {
+      return {people: true, destination: id, cluster};
     }
 
     return {};
@@ -198,7 +197,11 @@ MeetingInfoUtil.generateOptions = async (from) => {
     options.type = _CONVERSATION_URL_;
     try {
       await webex.internal.services.waitForCatalog('postauth');
-      const convoUrl = `${webex.internal.services.get('conversation')}/${CONVERSATIONS}/${hydraId.destination}`;
+
+      const convoUrl = webex.internal.conversation.getUrlFromClusterId({
+        cluster: hydraId.cluster,
+        id: hydraId.destination
+      });
 
       options.destination = convoUrl;
     }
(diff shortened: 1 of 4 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -30,7 +30,6 @@ import {
   MEET,
   MEET_M,
   HTTPS_PROTOCOL,
-  CONVERSATIONS,
```
lalli-flores, line 33: I don't think anything else is using this constant, can you remove if it's no longer needed?

mwegman:
This is one of those strange plugins that says it is a public public but it is using internal services like locus

lalli-flores:
true 🤦🏻‍♀️

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
InteractiveTimmy: Should be removed now.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -126,14 +125,14 @@ MeetingInfoUtil.isPhoneNumber = (phoneNumber) => {
...
-    if (decodedDestination.type === _ROOM_) {
-      return {room: true, destination: decodedDestination.id};
+  if (id && UUID_REG.test(id)) {
+    if (type === _ROOM_) {
+      return {room: true, destination: id, cluster};
     }
-    if (decodedDestination.type === _PEOPLE_) {
-      return {people: true, destination: decodedDestination.id};
+    if (type === _PEOPLE_) {
+      return {people: true, destination: id, cluster};
```
InteractiveTimmy, line 135: Enhanced this method to report cluster.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
lalli-flores: not sure the `constant` file change got pushed?

jpjpjp:
> If we could add a test, that would be good; otherwise, I like it.

Does testing framework support the idea of creating users in different clusters?    A simple test would be two create two users in two different clusters.   User 1 creates a space and add user 2.   User 2 initializes SDK and attempts a meeting with the newly created spaceId.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
lalli-flores: Can you have this refactoring change in a separate commit?

lalli-flores:
> > If we could add a test, that would be good; otherwise, I like it.
> 
> Does testing framework support the idea of creating users in different clusters? A simple test would be two create two users in two different clusters. User 1 creates a space and add user 2. User 2 initializes SDK and attempts a meeting with the newly created spaceId.

I think you're talking about the ["journey" tests](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/plugin-meetings/test/integration/spec/journey.js), this might be a case that's already covered.
For the purposes of this PR, it would be adding unit and integration tests around the utilities changed

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
```diff
@@ -198,7 +197,11 @@ MeetingInfoUtil.generateOptions = async (from) => {
     options.type = _CONVERSATION_URL_;
     try {
       await webex.internal.services.waitForCatalog('postauth');
-      const convoUrl = `${webex.internal.services.get('conversation')}/${CONVERSATIONS}/${hydraId.destination}`;
+
+      const convoUrl = webex.internal.conversation.getUrlFromClusterId({
+        cluster: hydraId.cluster,
+        id: hydraId.destination
+      });
```
InteractiveTimmy, line 204: This logic appears untestable due to a lack of Webex object availability in the tests [common issue we see with testing these util methods]. This prevents stubbing of `getUrlFromClusterId()`.

packages/node_modules/@webex/plugin-meetings/src/meeting-info/util.js
InteractiveTimmy: Fixed and pushed

PR #1781, opened 2020-07-09
Generate and update `Changelog.md` on every publish

taymoork2:
## Description

And now we have a third way for people to ignore what the current version # of the SDK is.
Also now everyone can judge our design decisions to a commit level 😃 

Fixes # (issue)

## Type of Change

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "build:docs": "documentation build --config documentation/config.yml --np --project-name webex --project-version 1.x --format html --output ./docs/api --github ./packages/node_modules/webex/src/index.js ./packages/node_modules/{@ciscospark,@webex}/plugin-*/src/index.js",
     "check-karma-output": "./scripts/analyze-output.sh",
     "build:package": "echo 'deprecated; just use \"npm run --silent build\"'./tooling/build.js",
+    "changelog:generate": "standard-changelog",
     "circleci:local": "circleci local execute -c .circleci/github.config.yml",
     "lint": "npm run --silent lint:js && npm run --silent lint:package-names",
     "lint:commitmsg": "commitlint -e $GIT_PARAMS",
@@ -85,7 +86,8 @@
     "serve:package": "webpack-dev-server --config webpack/webpack.config.js --hot --inline --history-api-fallback",
     "samples:test": "NODE_ENV=test wdio",
     "tooling": "babel-node ./tooling/index.js",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "version": "./tooling/version.sh"
   },
   "dependencies": {
     "@commitlint/cli": "^8.1.0",
--- a/package.json
+++ b/package.json
@@ -222,6 +222,7 @@
     "safe-buffer": "^5.2.0",
     "sdp-transform": "^2.12.0",
     "sinon": "^7.4.1",
+    "standard-changelog": "^2.0.24",
     "supertest": "^3.4.2",
     "terser-webpack-plugin": "^2.3.5",
     "time-grunt": "^1.4.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webex-js-sdk",
-  "version": "0.5.349",
+  "version": "1.80.270",
   "description": "JavaScript SDK for interacting with the Webex API",
   "homepage": "https://github.com/webex/webex-js-sdk",
   "bugs": {
--- a/tooling/version.sh
+++ b/tooling/version.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+echo "Build Packages and Script"
+npm run build
+npm run build:script
+
+echo "Bump all package version"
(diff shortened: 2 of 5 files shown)


tooling/version.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "Build Packages and Script"
+npm run build
+npm run build:script
+
+echo "Bump all package version"
+npm run tooling -- version set "${NPM_PACKAGE_VERSION}" --all
+
+echo "Generate Changelog"
+npx conventional-changelog-cli -p angular -i CHANGELOG.md -s
```
taymoork2, line 11: ignore this npx for now i was just using it to test it out before adding it to the package.json

package.json
```diff
@@ -222,6 +224,7 @@
     "safe-buffer": "^5.2.0",
     "sdp-transform": "^2.12.0",
     "sinon": "^7.4.1",
+    "standard-changelog": "^2.0.24",
```
lalli-flores, line 227: Is there a reason why we are not using `devDependencies`?

tooling/version.sh
```diff
@@ -0,0 +1,13 @@
...
+npm run build
+npm run build:script
+
+echo "Bump all package version"
+npm run tooling -- version set "${npm_package_version}" --all
+
+echo "Generate Changelog"
+npm run changelog:generate
+
+git add packages/node_modules/*/package.json packages/node_modules/@ciscospark/*/README.md packages/node_modules/@webex/*/package.json packages/node_modules/webex/umd/*.js CHANGELOG.md
```
lalli-flores, line 13: Why are you still updating `@ciscospark`?

lalli-flores: Can you move this command to the `Push to upstream/master` step instead?
It's weird (and prone to cause future confusion) that git staging happens here and the other git stuff happens elsewhere. I'd prefer if you can do all the `git` stuff together

package.json
```diff
@@ -85,7 +86,8 @@
     "serve:package": "webpack-dev-server --config webpack/webpack.config.js --hot --inline --history-api-fallback",
     "samples:test": "NODE_ENV=test wdio",
     "tooling": "babel-node ./tooling/index.js",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "version": "./tooling/version.sh"
```
lalli-flores, line 90: According to the docs for [npm version](https://docs.npmjs.com/cli/version), step 3 NPM bumps the version and then executes this script.
Does this not conflict with the script in the CI that already bumps the version?

tooling/version.sh
taymoork2: @ciscospark is a left over from during the rename/new pipeline changes (and doesn't hurt to keep since it's not doing anything)

taymoork2: the `git add` is needed since `npm version` will only commit the `package.json`

https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli#with-npm-version

lalli-flores: Can you add a comment that points to this then?

lalli-flores: > @ciscospark is a left over from during the rename/new pipeline changes (and doesn't hurt to keep since it's not doing anything)
> I could just change it to `git add -A`

Whichever one is fine

package.json
taymoork2: at this point it's a lost cause on the SDK

package.json
taymoork2: it shouldn't since we're explicitly setting it in the circleci config
`npm version ${VERSION} -m 'chore(release): v%s [skip ci]'`

package.json
lalli-flores: Let's not add more code to the bundle though 🙁
or does the SDK not respect that too?

package.json
lalli-flores: Step 3 says:
```
Bump version in package.json as requested
```

So I'm assuming the versioning will work is:
1. Current version [bump CI step](https://github.com/webex/webex-js-sdk/blob/136974e8dc6d45a168dec2a46c2a39a5c0e8c5d3/.circleci/config.yml#L329) bumps version in all `package.json` packages but the [top-level `package.json`](https://github.com/webex/webex-js-sdk/blob/master/package.json)
2. `npm version` will bump top-level [top-level `package.json`](https://github.com/webex/webex-js-sdk/blob/master/package.json)
3. `version` script will get executed (e.g. building + changelog generation)

package.json
taymoork2: It only bundles dependencies it touches, I'm pretty sure sinon isn't gonna show up in the final bundle since that only used in `test/`

taymoork2: Also the SDK doesn't have a `devDependencies` property in the `package.json`
But you're more than welcome to sift through all the packages and figure out what's a devDependency and what's an actual dependency 😅

lalli-flores:
Changelog addressed as part of https://github.com/webex/webex-js-sdk/pull/1787

PR #1784, opened 2020-07-10
Update local samples test documentation

jdaly13:
# Pull Request 

## Description

Update documentation to better inform users how to run end to end tests locally.  In order to create tests users, certain environment variables need to be set that are provided by the Cisco dev team.  Information includes which env variables that need to be set and whom to contact to get these values

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -280,7 +280,8 @@ You can do so by modifying the [wdio.conf.js](./wdio.conf.js) file.
 Simply change the `browserFirefox`'s `desiredCapabilities` object to the same as `browserChrome` (the Chrome instance).
 When you run, you should see two instances of Chrome open.  
 
-To Run end to end tests locally you will need the following env variables  
+To Run end to end tests locally you will need the following env variables 
+These are to create test users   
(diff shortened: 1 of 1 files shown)


jdaly13:
Hi Lalli
On the contributing guide it states 
_You can get these values by registering a new integration on Cisco Webex for Developers_,
I found this to be not true - they need to be specific values - when I registered an integration and tried to use those values given through the development portal - creating test users did not work.  I will demonstrate to you if you like.

In order to run Sauce tests you actually one needs a Sauce Labs account which is not listed at all and it must be a Sauce Labs registered within Cisco because an account registered by a random email (personal email) did not work.
Unless you can specifically quote in the documentation - I don't see anything where you need to register a specific Sauce Labs account, it is implied that functionality is built in.

lalli-flores:
Any integration with the right permission will do. The key here is having the right permissions. 😛
From what I understand the testing service is internal, with no plans on exposing it externally. The client ID and secret is only for our team and not something we share (even internally with other teams) because of the privileges it has.
But otherwise having a client ID/secret should work with the SDK public services.

Not sure on Sauce Labs, maybe @taymoork2 can comment.

PR #1785, opened 2020-07-11
Spark 159014

jpjpjp:
# Pull Request Template

## Description

Fixes and Tests for the issue where IDs were not being properly encoded in the websocket events for the attachmentAction, membership, message and rooms resources.

Fixes # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-159014

## Type of Change

## Test Coverage

- [x] Mocha Tests for the four modules that were changed
- [x] New node based test in the samples directory

Note, two plugin-membership tests were failing.  I believe this was unrelated to this change, it looks like someone modified the tests so that the roomType element of some membership objects were not being tested.  Restored this to include this check.

**Test Configuration**:
Relevant plugin tests all pass

The new webhook/websocket comparison test is in the samples directory.  To run cd to /packages/node_modules/samples/node-compare-sockets-webhooks.  Run `npm install` to load dependencies.   Set environment variables:
 * TOKEN - user token to test -- ideally run this once with token of a user in all 3 clusters 
 * WEBHOOK_URL - url where this app is running, ie local ngrok
 * PORT - port where app should listen for webhooks

This test is configured to use the local build version of webex, but to understand the existing problem it's interesting to modify the following code to include the current npm version of webex:
```js
// Use public webex instance
// const Webex = require('webex');
// Use this builds webex instance
const Webex = require('../../webex');
```
Prior to this PR being accepted and published, the npm version of webex should generate about 20 errors when run wit a token of a user who is not in the ACHM cluster.
Using this branches build should eliminate these errors.

It would be nice to see if there is a way to make running this test part of the build process.

## Checklist:

### DIFF
--- a/packages/node_modules/@webex/common/src/constants.js
+++ b/packages/node_modules/@webex/common/src/constants.js
@@ -67,3 +67,6 @@ export const hydraTypes = {
   ROOM: 'ROOM',
   TEAM: 'TEAM'
 };
+
+export const INTERNAL_US_CLUSTER_NAME = 'urn:TEAM:us-east-2_a';
+export const INTERNAL_US_INTEGRATION_CLUSTER_NAME = 'urn:TEAM:us-east-1_int13';
--- a/packages/node_modules/@webex/common/src/uuid-utils.js
+++ b/packages/node_modules/@webex/common/src/uuid-utils.js
@@ -1,5 +1,5 @@
 import {encode, decode} from './base64';
-import {SDK_EVENT, hydraTypes} from './constants';
+import {SDK_EVENT, hydraTypes, INTERNAL_US_CLUSTER_NAME, INTERNAL_US_INTEGRATION_CLUSTER_NAME} from './constants';
 
 const hydraBaseUrl = 'https://api.ciscospark.com/v1';
 
@@ -23,6 +23,11 @@ export function constructHydraId(
 ) {
   if (!type.toUpperCase) {
     throw Error('"type" must be a string');
+  }
+
+  if ((type === hydraTypes.PEOPLE) || (type === hydraTypes.ORGANIZATION)) {
+    // Cluster is always "us" for people and orgs
+    return encode(`ciscospark://us/${type.toUpperCase()}/${id}`);
   }
 
   return encode(`ciscospark://${cluster}/${type.toUpperCase()}/${id}`);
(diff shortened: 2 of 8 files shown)


packages/node_modules/@webex/common/src/constants.js
```diff
@@ -67,3 +67,5 @@ export const hydraTypes = {
   ROOM: 'ROOM',
   TEAM: 'TEAM'
 };
+
+export const INTERNAL_US_CLUSTER_NAME = 'urn:TEAM:us-east';
```
mwegman, line 71: This will fail if they add a cluster called `urn:TEAM:us-east-something-else`. I don't know how likely this will be.

packages/node_modules/@webex/common/src/uuid-utils.js
```diff
@@ -25,6 +25,11 @@ export function constructHydraId(
     throw Error('"type" must be a string');
   }
 
+  if ((type === hydraTypes.PEOPLE) || (type === hydraTypes.ORGANIZATION)) {
+    // Cluster is always "us" for people and orgs
+    return encode(`ciscospark://us/${type.toUpperCase()}/${id}`);
```
mwegman, line 30: Does the Hydra team agree with this assumption? I don't know another way of getting the cluster id for a person or org without making a call to the server.

jpjpjp: I did get @pritomgogoi  to confirm this logic.   I have also anecdotally validated it by decoding all the people and orgIDs that I could find, including ones that I knew belonged to other clusters.

packages/node_modules/@webex/common/src/constants.js
jpjpjp: @mwegman  how much precision do you recommend?   When we chatted on Thursday I got the impression that sometimes there were "two colons" worth of precision and sometimes three.  I went down this path to protect against NOT properly using the shorthand "us" for the ACHM cluster.    Is there a longer string that you feel will be guaranteed to be returned by the webex.internal.services.getClusterId call?

mwegman: You could compare to `urn:TEAM:us-east-1_int13` and `urn:TEAM:us-east-2_a` to cover both production and integration

jpjpjp: @pritomgogoi is there any chance you could point me to the hydra logic that figures out when to use the hard coded "us" in the cluster?   @mwegman's suggestion protects against future new clusters called `urn:TEAM:us-east-something-else`, but his suggestion also seems exposed to any changes to the internal names of the existing clusters.  (I'm not sure which is more likely!)

mwegman: If you change the names of existing clusters, you are going to have a lot of problems. Any existing Hydra ID will break. Anyone using the current version of the SDK will need to upgrade before we change the default cluster name.

jpjpjp: OK @mwegman, I have added your suggestion.

jpjpjp:
> Can you make the `compare-sockets-webhooks` package a separate PR? Thank you.

Yes, removed from this PR

PR #1786, opened 2020-07-13
fix(plugin-meetings): fix muted by others event

arun3528:
# Pull Request Template

## Description
Small PR to fix muteByOther event not getting triggered when someone else mutes you


arun3528:
@lalli-flores no its does update name so it wont need any change . Its already there . It was just not triggering it

PR #1787, opened 2020-07-13
Use standard-version for SEMVER version bumping

taymoork2:
## Description

Huh, I was wrong, this was pretty simple.

Fixes # (issue)

## Type of Change

## Checklist:


.circleci/config.yml
```diff
@@ -325,38 +325,32 @@ jobs:
             fi
       - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
       - run:
-          name: 'Generate Version Number'
-          command: echo "export VERSION=$(npm run --silent get-next-version | xargs)" >> $BASH_ENV
-      - run: echo ${VERSION}
+          name: 'Store current version number'
+          command: echo "export PREVIOUS_VERSION=$(npm run --silent tooling -- version current | xargs)" >> $BASH_ENV
```
lalli-flores, line 329: I like how the name is "current version" but then the variable name is `PREVIOUS_VERSION` 😂

I know why this is, but I think you should call this `CURRENT_VERSION` for clarity in the future. What is called `CURRENT_VERSION` should be renamed `UPCOMING_VERSION` or something similar, since technically it has not been published yet

.circleci/config.yml
```diff
@@ -325,38 +325,32 @@ jobs:
...
-          name: 'Tag for release'
-          command: git tag "v${VERSION}"
-      - run:
-          name: 'Push tagged release to upstream/master'
-          command: git push upstream v${VERSION}:v${VERSION}
-      - run:
-          name: 'Generate Package Dependencies'
-          command: npm run deps:generate
+          name: 'Push tag to upstream/master'
+          command: git push --tags upstream
```
lalli-flores, line 338: I think this should happen after we publish to NPM. If for whatever reason NPM publishing fails, we would already have pushed the new release in Git, and I think there's more clean up to do then, or may run into situations where we have to make exceptions in the whole flow because nothing but publishing has to happen again.

This is to say that I think is safer/easier to push the new version changes to Git only after publishing is successful. That's how we [do it in the components](https://github.com/webex/components/blob/master/release.config.js) too.

tooling/babel-plugin-inject-package-version.js
```diff
@@ -2,20 +2,9 @@
...
- * object
- * @param {Object} state
- * @private
- * @returns {string}
- */
-function versionFromState(state) {
-  // eslint-disable-next-line global-require
-  return require(pkgUp.sync(state.file.opts.filename)).version;
-}
+const {version} = require('../package.json');
```
lalli-flores, line 7: When in the flow does this script get called?

package.json
```diff
@@ -267,5 +271,11 @@
       "commit-msg": "commitlint --verbose -e $HUSKY_GIT_PARAMS",
       "pre-commit": "npm run --silent srcdist && npm run --silent lint:staged"
     }
+  },
+  "standard-version": {
+    "scripts": {
```
lalli-flores, line 276: What happened to the `prerelease` step?

taymoork2: `npm run build` is where `webex.version` is set and i refactored the babel-plugin to use the package.json instead of pkg-up

tooling/babel-plugin-inject-package-version.js
taymoork2: addressed in prerelease question

.circleci/config.yml
taymoork2: I can do that and it makes sense
Most of these tools suggest you push w/ tags first then publish

CONTRIBUTING.md
```diff
@@ -375,19 +382,6 @@ git rebase master
 
 Finally, open a [new Pull Request](https://github.com/webex/webex-js-sdk/compare) with your changes. Be sure to mention the issues this request addresses in the body of the request. Once your request is opened, a developer will review, comment, and, when approved, merge your changes!
 
-### Pull Request Checklist
```
lalli-flores, line 378: Is there a reason for removing the checklist?

CONTRIBUTING.md
```diff
@@ -307,16 +307,23 @@ If the commit reverts a previous commit, it should begin with `revert:`, followe
...
+> Appending a `!` and/or a `BREAKING CHANGE:` footer to **ANY TYPE** will denote a **BREAKING CHANGE** and will cause a **MAJOR** bump
+
+The following types will _**not**_ cause a version bump:
+
+- **build**: Changes that affect the build system or external dependencies
+- **ci**: Changes to our CI configuration files and scripts
 - **docs**: Documentation only changes
-- **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
-- **refactor**: A code change that neither fixes a bug nor adds a feature
 - **perf**: A code change that improves performance
```
lalli-flores, line 323: Performance fixes won't trigger a patch release?

.circleci/config.yml
lalli-flores: I'm taking it from the suggested flow [from semantic-release](https://semantic-release.gitbook.io/semantic-release/usage/plugins#plugins-declaration-and-execution-order), but I think it makes sense for all the other reasons I mentioned too 😁

.circleci/config.yml
```diff
@@ -325,38 +325,32 @@ jobs:
...
-          command: npm run tooling -- version set ${VERSION} --all
-      - run:
-          name: 'Commit bumped version'
-          command: |
-            git add packages/node_modules/*/package.json packages/node_modules/@ciscospark/*/README.md packages/node_modules/@webex/*/package.json packages/node_modules/webex/umd/*.js
-            git commit -m "chore(release): v${VERSION} [skip ci]"
+          name: 'Run standard-version'
+          command: npm run release
       - run:
           name: 'Push to upstream/master'
```
lalli-flores, line 334: Should this move along with the tag push?

package.json
lalli-flores: OK, I'm ok with that. Let's steps, less bugs! 😁

.circleci/config.yml
taymoork2: Nope, this should be pushed at least, kinda weird to publish to npm first and then github
It's probably easier to publish to npm locally then deal with npm

taymoork2: That being said if a failure happens during npm publish, then the git tag doesnt get pushed which is probably another problem, so i'm more inclined to move git tag back up to here

CONTRIBUTING.md
taymoork2: Guess it wasn't rebased correctly

CONTRIBUTING.md
taymoork2: Im just going off what i've read on https://conventionalcommits.org/

.gitattributes
```diff
@@ -1,3 +1,5 @@
+package-lock.json -diff
```
taymoork2, line 1: https://twitter.com/afschr/status/970816918046330881?lang=en

CONTRIBUTING.md
lalli-flores: Can you add a comment then? Something like `Based on [conventional commits guidance]()`

.circleci/config.yml
lalli-flores: I think is ok for nothing to be pushed to git if the publishing failed. The commits that will get pushed are on all version/publishing changes right?
So if publishing a new version failed we should probably not push those things into git anyway. This way we can re-run the publishing workflow as if the last stuff was never versioned-bumped.

CONTRIBUTING.md
taymoork2: Eh close enough
https://github.com/webex/webex-js-sdk/pull/1787/files#diff-6a3371457528722a734f3c51d9238c13R286

.circleci/config.yml
taymoork2: This is probably a bigger problem if some of the packages get published and some don't

lalli-flores: good point, there are multiple packages to publish. What's the flow if something fails?

taymoork2: I think i'd be fine if the github release and tag are committed before npm
If a failure happens during npm publish we (i.e. npm admin adam) will have to manually push it up

taymoork2: If it ever becomes a problem we can swap to push + commit after npm publish
But i'd rather keep the setup the same~ish and other tools suggest push+tag->publish

lalli-flores: @adamweeks, for the record so you know what @taymoork2 is signing you up for 😆

taymoork2: He already knows, even I dont have access to publish on npm, I think him and Dave are the only ones who can atm

.circleci/config.yml
taymoork2: Aight i googled some more, im seeing more examples that use npm publish -> git push
So i'll switch to that

Issue #1782, opened 2020-07-09
Configurable log file/buffer log level


glenatwork:
## Expected Behavior

Offer a configuration setting, such as "bufferLogLevel" that allows setting what level of logs are sent to the buffer/file logs.

## Current Behavior

The level is currently hard-coded to info:
  shouldBuffer(level) {
    return precedence[level] <= precedence.info;
  },

## Possible Solution (we welcome any logical suggestion)

It should be as simple as something like:

```
    return precedence[level] <= (this.config.bufferLogLevel ? precedence[this.config.bufferLogLevel] : precedence.info);
```

## Environment Details

There is no way to set the buffer log level in any environment.

lalli-flores:
Issue fixed by PR https://github.com/webex/webex-js-sdk/pull/1792. Thank you, @glenatwork

PR #1792, opened 2020-07-15
feat(logging): add config setting for buffer log level

glenatwork:
For debugging issues using the buffered log file, we need the ability to set the logging level, which is currently hard-coded to info.

Issue https://github.com/webex/webex-js-sdk/issues/1782

---


packages/node_modules/@webex/plugin-logger/src/logger.js
```diff
@@ -194,7 +194,7 @@ const Logger = WebexPlugin.extend({
    * @returns {boolean}
    */
   shouldBuffer(level) {
-    return precedence[level] <= precedence.info;
+    return precedence[level] <= (this.config.bufferLogLevel ? precedence[this.config.bufferLogLevel] : precedence.info);
```
lalli-flores, line 197: Is `bufferLogLevel` a new configuration attribute?

taymoork2: @glenatwork once you answer this question and resolve the conversation i can move forward with a merge

glenatwork: @lalli-flores yes bufferLogLevel is a new config setting so as to not interfere with behavior of any existing setting (ie keep everything backwards compatible).  also this way the level can be different for the console vs. the log file.
@taymoork2 hope that answers the question.  thanks for the help.

taymoork2: @glenatwork since the change is small, can you split these commits to be a `feat()` and a `test()` and for me it should be good to merge

glenatwork: @taymoork2 done!

PR #1797, opened 2020-07-17
Standard version tweaks

taymoork2:
Just some tweaks I've observed over the past two days.
`standard-version` isn't as maintained as it seems

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -325,6 +325,17 @@ jobs:
           name: 'Grab Current Version Number'
           command: echo "export CURRENT_VERSION=$(node -p "require('./package.json').version")" >> $BASH_ENV
       - run:
+          name: 'Check if we need to bump'
+          command: |
+            if ! \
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -324,6 +324,17 @@ jobs:
       - run:
           name: 'Grab Current Version Number'
           command: echo "export CURRENT_VERSION=$(node -p "require('./package.json').version")" >> $BASH_ENV
+      - run:
+          name: 'Check if we need to bump'
+          command: |
+            if ! \
+              { git log "$( git describe --tags --abbrev=0 )..HEAD" --format='%s' | cut -d: -f1 | sort -u | sed -e 's/([^)]*)//' | grep -q -i -E '^feat|fix|perf|refactor|revert$' ; } || \
```
lalli-flores, line 331:
```suggestion
              { git log "$( git describe --tags --abbrev=0 )..HEAD" --format='%s' | cut -d: -f1 | sort -u | sed -e 's/([^)]*)//' | grep -q -i -E '^feat|fix|perf|revert$' ; } || \
```

Not sure if a refactoring change should trigger a release. Ideally the ideally the public API should stay the same and its’ just internal rework, e.g. splitting things into files, breakdown helpers for readability, etc.. those changes could wait until next release (because they wouldn't cause a change in the bundle or functionality)

PR #1798, opened 2020-07-21
Fix room update events

jpjpjp:
Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-162146

---


packages/node_modules/@webex/plugin-rooms/src/rooms.js
```diff
@@ -14,7 +14,7 @@ import {
   deconstructHydraId
 } from '@webex/common';
 
-const debug = require('debug')('memberships');
+const debug = require('debug')('rooms');
```
InteractiveTimmy, line 17: Do we know why we're using `require()` here? Not sure why we are using mixed-import methods, if I'm understanding this correctly.

jpjpjp: Can't say.  Its a pattern that is consistent across all the mocha tests that I've ever tried to modify.  My only goal here was to fix the fact that the rooms test was using "memberships" string instead of "rooms" as the debug trigger.

Happy to remove this and leave that error in place if it gets the PR approved.  This was just a side bit of cleanup I was trying to do ;-)

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
```diff
@@ -692,3 +763,26 @@ function validateMembershipEvent(event, membership, actor, creator = actor) {
...
+ */
+function validateRoomsUpdatedEvent(event, room, isLocked) {
+  const {data} = event;
+
+  assert((room.id === data.id),
+    'rooms:updated event did not have the expected roomId');
+  assert((room.type === data.type),
+    'rooms:updated event did not have the expected room type');
+  assert((data.isLocked === isLocked),
+    'rooms:updated did not have expected isLocked value');
```
taymoork2, line 785: Any reason this is formatted like? I think it should be fine on the same line or each param is new lined. This formatting is just weird.

packages/node_modules/@webex/plugin-rooms/src/rooms.js
```diff
@@ -433,38 +436,50 @@ const Rooms = WebexPlugin.extend({
...
       sdkEvent.data.created = activity.published;
       sdkEvent.actorId = buildHydraPersonId(activity.actor.entryUUID, cluster);
+      if (activity.object.id) {
+        sdkEvent.data.id =
+          buildHydraRoomId(activity.object.id, cluster);
+      }
+      else {
+        sdkEvent.data.id =
+          buildHydraRoomId(activity.target.id, cluster);
+      }
```
taymoork2, line 451: Any reason this is formatted like? I think it should be fine on the same line.

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
jpjpjp: The linter complains about lines too long.

packages/node_modules/@webex/plugin-rooms/src/rooms.js
jpjpjp: Responding to linter complaints

jpjpjp: I take it back for these.   These WERE longer when I was using constants and were two lines.  They can be one line now.

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
taymoork2: Eh good enough for me, #FISI

packages/node_modules/@webex/plugin-rooms/src/rooms.js
InteractiveTimmy: Looks good, this is how this is currently implemented throughout the SDK in tests. We will follow this up internally to determine if this is necessary.

PR #1809, opened 2020-07-28
Threading

buckrivard:
This feature adds standard activity thread-ordering logic to conversation plugin to be used by webex teams web client teams and third party developers to abstract threading logic away from each client application instance.

I am eager for feedback in a few main areas:
* ~~babel and babelrc configuration: I had to add some babel plugins to get async* methods to transpile, but they are still not direct source maps. This impacts browser debugging because babel is transpiling into a bit of a messy massive switch statement.~~
* testing: what level of integration and unit testing is expected for a feature like this?
* additional methods: I added additional methods for fetching activity parents and children (I moved these functions from Cantina codebase into SDK)
* listAllChildActivitiesByParentId: I implemented the Page class iteration directly within an SDK method, which I noticed is ordinarily left for callers to do themselves. Is this appropriate?
* error handling: what is expected for public SDK method error handling?


adamweeks:
This concerns me:

>but they are still not direct source maps. This impacts browser debugging because babel is transpiling into a bit of a messy massive switch statement.

We do a LOT of browser based debugging for the meetings plugin and if we can't get source maps, it will make it much more difficult. Is there a way to do this without the async iterator?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -740,6 +752,110 @@ const Conversation = WebexPlugin.extend({
...
+
+    while (page.hasNext()) {
+      // eslint-disable-next-line no-await-in-loop
+      page = await page.next();
+      for (const activity of page) {
+        items.push(activity);
+      }
+    }
+
+    if (last(items).published < items[0].published) {
```
lalli-flores, line 834: When would be the case that we don't want to reserve? Should we set the sort order in the request instead?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -740,6 +752,110 @@ const Conversation = WebexPlugin.extend({
...
+    return items;
+  },
+
+  /**
+   * @param {string} conversationUrl
+   * @param {string} activityParentId
+   * @param {object} query
+   * @returns {Promise}
+   */
+  async getReactionSummaryByParentId(conversationUrl, activityParentId, query) {
```
lalli-flores, line 847: Not sure I understand the value of this function when `listChildActivitiesByParentId` already can filter activity types?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1461,6 +1577,610 @@ const Conversation = WebexPlugin.extend({
...
+   * @returns {object}
+   * returns three functions:
+   *
+   * getOlder - gets older activities than oldest fetched
+   *
+   * getNewer - gets newer activities than newest fetched
+   *
+   * jumpToActivity - gets searched-for activity and surrounding activities
+   */
+  listActivitiesThreadOrdered(options) {
```
lalli-flores, line 1605: Can you keep this consistent with other functions and have them be 2 separate arguments?
```suggestion
  listActivitiesThreadOrdered(conversationUrl, activityId) {
```

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1461,6 +1577,610 @@ const Conversation = WebexPlugin.extend({
...
 
+
+  /**
+   * common interface for facade of generator functions
+   * @typedef {object} IGeneratorResponse
+   * @param {boolean} done whether there is more to fetch
+   * @param {any} value the value yielded or returned by generator
+   */
+
+  /**
```
lalli-flores, line 1588: Can you fill out the rest of the docs for this method?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1461,6 +1577,610 @@ const Conversation = WebexPlugin.extend({
...
+   * @param {string} options.url
+   * @param {?string} options.minActivities how many activities to return in first batch
+   * @param {?string} [options.queryType] one of older, newer, mid. defines which direction to fetch
+   * @param {?object} [options.search] server activity to use as search middle date
+   *
+   * @yields {Activity[]}
+   *
+   * @returns {void}
+   */
+  async* _listActivitiesThreadOrdered(options = {}) {
```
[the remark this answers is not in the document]
holsted: What do you mean? Just adding more util functions at the top of the file? Does it make it more readable that we have to jump up and down in the file? I agree with small/digestible bits of code but this is a complex feature. We're basically put all of the messaging logic into one method. You either get a bunch of utils or one long method. I thought it was well commented and things that were helper/utils were broken out.

holsted:
@lalli-flores 

> My first suggestion would be to break up this into multiple PRs. You've already done so for the babel changes, but it would be good to do the same for the code. It makes it easier to add testing, review and integrate pieces of functionality into the plugin.
For instance, the API calls to get the activity IDs is pretty straight forward. Those can be pulled out into their own PR too, with tests for that functionality only.

This would have been good things to discuss when we did the initial design review. Do you want us opening PRs for functionality that is incomplete or not being used? It's hard to know that a feature fully works until you get all the pieces together. I am 100% a fan of smaller PRs but sometimes you just have a big feature that needs to be reviewed live.

packages/node_modules/@webex/internal-plugin-conversation/src/threading.js
```diff
@@ -0,0 +1,84 @@
...
+  EDIT: 'EDIT',
+  REACTION: 'REACTION',
+  REACTION_SELF: 'REACTION_SELF',
+  ROOT: 'ROOT',
+  CREATE: 'CREATE',
+  TOMBSTONE: 'TOMBSTONE',
+  DELETE: 'DELETE'
+};
+
+// TODO: create type for tombstone edit? could make it easy to ignore older edits...
```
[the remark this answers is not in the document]
buckrivard: Right, that's a good question. The filename itself I would happily change. Perhaps simply `list-activities-thread-ordered.js`? So as to point directly to the method for which it provides helpers and constants? I like the clarity of that name, but it's awfully long... Maybe `activity-thread-ordering.js`?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
buckrivard: This function calls a different endpoint than listChildActivitiesByParentId. I've just spoken to the engineer who pulled reactions into the web client, and he is vouching for this implementation VS a child activities implementation due to the convo API returning both types of reaction summary here, rather than a server round trip to call each one specifically.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
buckrivard: This function accepts more arguments than the two that get destructured at the top: the `baseOptions` variable spreads the options object (minus the destructured properties) and is passed along to the private method

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
buckrivard: I wrote them in-line, just above the wrapper functions that this method returns. Each of the wrapper functions returns this IGeneratorResponse, but their docs say more specifically what's happening there. I'm much more familiar with documenting using TS interfaces and types... My jsdoc-umenting is a little rough

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
buckrivard: I'm not sure, actually. I saw this condition in the `_list` method, and I assumed it had something to do with convo. I see @mwegman gitblame there in the _list method, maybe he can lend some insight on whether this is necessary check?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -740,6 +752,110 @@ const Conversation = WebexPlugin.extend({
...
+   * @param {string} [query.activityType] The type of child activities to return.
+   * The value is one of 'reply', 'edit', 'cardAction', 'reaction', 'reactionSummary', 'reactionSelfSummary'
+   * @param {number} [query.limit] The limit of child activities that can be returned per request
+   * @param {boolean} [query.latestActivityFirst] Sort order for the child activities
+   * @param {boolean} [query.includeParentActivity] Enables the parent activity to be returned in the activity list
+   * @param {string} [sinceDate] Get all child activities after this date
+   * @param {string} [maxDate] Get all child activities before this date
+   * @returns {Promise<Array>}
+   */
+  async listAllChildActivitiesByParentId(options = {}) {
```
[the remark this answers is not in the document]
buckrivard: Sorry, that must be a little unclear. This method doesn't return all children of all types, it returns all children of the passed-in type, as per the jsdoc above. Maybe renaming to `listAllChildActivitiesOfTypeByParentId`? A mouthful, but would clarify intent.

buckrivard:
@lalli-flores 

Here's what I'm thinking: 

One PR for the following methods: `listParentActivityIds`, `listChildActivitiesByParentId`, `getReactionSummaryByParentId`
These methods had been living in cantina codebase with comments about moving them into SDK. All three are used in cantina, and therefore a PR for them specifically is a reasonably tidy logical grouping with a solid reason.

A second PR for the following method: `listAllChildActivitiesByParentId`
This PR builds on the previous `listChildActivitiesByParentId`, and due to its complexity I can see a case for it being a standalone PR and minor version bump.

A third and final PR for the following methods: `_listActivitiesThreadOrdered`, `listActivitiesThreadOrdered`.
This PR is the piece de resistance: common threading logic for client consumption. A public facade and a private method, they work in tandem to abstract threading from client codebases and into a simple, public method that provides coverage for the most common client use cases.

buckrivard:
@lalli-flores 

And now, a defense of the size of the (admittedly) behemoth \_listActivitiesThreadOrdered method:

Yes, it is massive. I had some design decisions to make here, and I (as of yet) stand by them. The first defense is of design consideration, which was the use of a stateful method. This is a stateful method, and for good reason. In order to truly keep client code uncluttered and unconcerned with the business logic of thread ordering, the client _should not be responsible_ for keeping track of their own fetching "contexts". The primary of these being upper- and lower-bound timestamps, or, as in the code itself, `oldestAct` and `newestAct`. These "bookends" are kept alive for the life of the generator instance, and are recalculated after every successful activities fetch. The benefit? Callers may query for activities "relatively", meaning "old-_er_" and "new-_er_" activities than what they have queried previously, with no concern for which specific activity they are interested in being older _than_.

The second defense is code-style. Ordinarily I am an FP fan. I like my functions itty bitty, pure, and well named. In this case, however, I have a stateful function on my hands. Pretty common for generators, actually... Otherwise, yeah. You get it. So I have these states and stateful data structures (see `replyActivityHash` and its ilk). In FP land, I might interact with this object via pure function, with the data structure is passed in, cloned, clone is modified, modified clone is returned. But this _still_ means I have to reassign a global variable for each state change. So why go through the trouble? Not to mention the "one time use" linear-growing helpers list that would require (except in a separate file). So instead I decided to declare multiple helpers within the method itself (think of them like instance methods), and in them I closed over my state variables. What it adds in clutter it makes up for in function-name-specificity and control-flow readability. Once this method reaches its first while loop, where the real action happens, it has an arsenal of well-named, easy-to-reason-about helper functions that interact with the generator state in straightforward ways. The issue with breaking up this function into smaller pieces, then, is that it loses meaningful stateful operations. Any given section of the code, be it the instantiation portion (lines 1721-1817), the execution context (everything that is run upon invocation of the returned "next" method, lines 1817-2181), or the fetching logic (repeated until we have all activities asked for by caller, the lines 1978-2111), depend on outer scope closures that _must work together_ to achieve the desired result.

All this logic in cantina is spread throughout 5+ files, 3 "paradigms" (plain redux, redux thunk, react), and hundreds (thousands maybe, @holsted?) of lines of code.

lalli-flores:
> This would have been good things to discuss when we did the initial design review. Should we open PRs for functionality that is incomplete or not being used? Maybe next time we can discuss how we would slice out the work in a way that makes it easier to review. It's hard to know that a feature fully works until you get all the pieces together. I am 100% a fan of smaller PRs but sometimes you just have a big feature that needs to be reviewed live. Fortunately, this only touches a few files.

@holsted , we discussed this in the design discussion. I specifically asked for "building-block" functionality that as a last step could be put together to create the final outcome we want.
I wouldn't call it "incomplete" either. Is requesting children activity an incomplete feature? That certainly seems like a complete, useful feature on its own. That's certainly something the SDK couldn't do before.
I don't think its feasible to to expect to know upfront how to slice up the work, as it depends on the implementation the engineer is taking, but we do have a disclaimer in [our contributing guide #8](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#contributing) that if we feel a PR is too big, we will ask to be broken down.
I think in the end, we trust that our smart engineers understand when something can add value on its own 🙂, that's the whole premise of [continuous integration](https://martinfowler.com/articles/continuousIntegration.html)

lalli-flores:
> @lalli-flores
> 
> Here's what I'm thinking:
> 
> One PR for the following methods: `listParentActivityIds`, `listChildActivitiesByParentId`, `getReactionSummaryByParentId`
> These methods had been living in cantina codebase with comments about moving them into SDK. All three are used in cantina, and therefore a PR for them specifically is a reasonably tidy logical grouping with a solid reason.
> 
> A second PR for the following method: `listAllChildActivitiesByParentId`
> This PR builds on the previous `listChildActivitiesByParentId`, and due to its complexity I can see a case for it being a standalone PR and minor version bump.
> 
> A third and final PR for the following methods: `_listActivitiesThreadOrdered`, `listActivitiesThreadOrdered`.
> This PR is the piece de resistance: common threading logic for client consumption. A public facade and a private method, they work in tandem to abstract threading from client codebases and into a simple, public method that provides coverage for the most common client use cases.

This is in line with what I was thinking, let's go ahead and do it like this 👍🏼

PR #1811, opened 2020-07-29
build(babel plugins): add support for async generator functions

buckrivard:
Introduces transpilation and syntactical support for async generator functions within codebase


buckrivard:
Closing due to ongoing investigation into source maps issue https://github.com/webex/webex-js-sdk/pull/1809

buckrivard:
Reopening. @adamweeks, it turns out source mapping isn't an issue after all. It _appeared_ to be a problem because I was building the SDK locally and linking it to a local instance of cantina, where source maps _do_ get messed up, but if I build and test by serving up the SDK samples, the source mapping is preserved and works beautifully.

PR #1824, opened 2020-08-05
feat(plugin-memberships): support updating isRoomHidden attribute

jpjpjp:
Applications may now call memberships.update() and set the isRoomHidden attribute of the user's
membership to true or false for a 1-1 space that the user belongs to.
Setting it to false is effectively the same thing as a user choosing the "leave space" option in the
Webex Teams client for a 1-1 space.   The client will no longer display that 1-1 space in the user's
recents list, but if the other user sends a message it becomes "unhidden" and all the previous
conversation history is available.
When a space's hidden status changes for a user, a memberships:updated event is generated for
that user.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-166157


packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
```diff
@@ -628,25 +645,23 @@ describe('plugin-memberships', function () {
...
-                debug('room hidden');
-                debug(r);
+            return webex.memberships.update(membership)
+              .then(async (m) => {
+                debug('memberships.update() request returned OK.  Waiting for memberships:updated event');
                 const event = await updatedEventPromise;
 
-                validateMembershipEvent(event, membership, actor);
-              });
+                assert((m.isRoomHidden === event.data.isRoomHidden === true),
```
adamweeks, line 656: You seem to be combining two assertions into one here. Could you split these into two lines?
```js
assert.isTrue(m.isRoomHidden)
assert.isTrue(event.data.isRoomHidden)
```
That way, if an assertion fails we know which one was the culprit.

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
```diff
@@ -606,18 +606,35 @@ describe('plugin-memberships', function () {
...
+            return webex.memberships.list({
+              roomId,
+              personId: user.id
+            });
+          })
+          .then((memberships) => {
+            // Ensure that membership is the SDK users membership object
+            membership = memberships.items[0];
+            validateMembership(membership);
+            assert((membership.personId === actor.id),
```
adamweeks, line 633: These assertions are not needed since the test is for the `hide` function. These assertions should be covered in the `list` function tests. While they are good for debugging when writing the tests, we have found they cause issues when they fail because the testing framework doesn't link back the failures properly in a `before` block.

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
```diff
@@ -606,18 +606,35 @@ describe('plugin-memberships', function () {
...
+              '#hide test did not find test users membership to hide');
+          })
+          .catch((e) => console.error(e.message)));
```
adamweeks, line 636: We should rethrow the error so the test fails during the setup phase.

packages/node_modules/@webex/plugin-memberships/src/memberships.js
```diff
@@ -549,7 +549,9 @@ const Memberships = WebexPlugin.extend({
   getMembershipEvent(activity, event) {
     try {
       const sdkEvent = cloneDeep(this.eventEnvelope);
-      const cluster = getHydraClusterString(this.webex, activity.target.url);
+      const cluster = (activity.verb !== SDK_EVENT.INTERNAL.ACTIVITY_VERB.HIDE) ?
```
adamweeks, line 552: This change doesn't seem to match the commit message.

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
jpjpjp: OK

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
jpjpjp: Removed

packages/node_modules/@webex/plugin-memberships/test/integration/spec/memberships.js
jpjpjp: Removed the catch from the setup logic.

packages/node_modules/@webex/plugin-memberships/src/memberships.js
jpjpjp: New commit messages, plus a new one associated with additional javadocs

packages/node_modules/@webex/plugin-memberships/src/memberships.js
```diff
@@ -400,6 +401,38 @@ const Memberships = WebexPlugin.extend({
...
+   *   .then((memberships) => {
+   *     membership = memberships.items[0];
+   *     assert.equal(membership.isRoomHidden, false);
+   *     membership.isRoomHidden = true;
+   *     // This will generate a memberships:updated event
+   *     // that will only be seen by this user
+   *     return webex.memberships.update(membership);
+   *   })
+   *   .then(function(membership) {
+   *     var assert = require('assert');
```
adamweeks, line 431: You've already defined this on line 406

jpjpjp: Nice catch.

PR #1826, opened 2020-08-05
feat(conversation): add helper for fetching all child activities

buckrivard:
Create helper method to recurse through the "next" links of convo's */parent/* endpoint.

Part 2 of 3 from suggested PR flow agreed upon [here](https://github.com/webex/webex-js-sdk/pull/1809#issuecomment-666576235)

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
@@ -1477,8 +1477,8 @@ const Conversation = WebexPlugin.extend({
    * Returns a list of _all_ child activities for a given parentId within a given conversation
    * @param {object} [options = {}]
    * @param {string} [options.conversationUrl] targeted conversation URL
-   * @param {string} [activityParentId] parent id of edit activities or thread activities
-   * @param {QueryOptions} [query] object containing query string values to be appended to the url
+   * @param {string} [options.activityParentId] parent id of edit activities or thread activities
+   * @param {QueryOptions} [options.query] object containing query string values to be appended to the url
    * @returns {Promise<Array>}
    */
   async listAllChildActivitiesByParentId(options = {}) {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1473,6 +1473,38 @@ const Conversation = WebexPlugin.extend({
...
+
+    while (page.hasNext()) {
+      // eslint-disable-next-line no-await-in-loop
+      page = await page.next();
+      for (const activity of page) {
+        items.push(activity);
+      }
+    }
+
+    if (last(items).published < items[0].published) {
```
adamweeks, line 1501: What does this do? Could you add a comment as to why it is needed?

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -145,20 +145,78 @@ describe('plugin-conversation', () => {
...
+          .then(() => {
+            assert(spy.calledOnce);
           });
       });
+
+      it('returns children in ascending published order', () => webex.internal.conversation.listAllChildActivitiesByParentId({conversationUrl: convoUrl, activityParentId: '123-abc'})
+        .then((res) => {
+          const firstMessageOlderThanLastMessage = res[0].published < res[res.length - 1].published;
+
+          assert.equal(firstMessageOlderThanLastMessage, true, 'activities out of order');
```
adamweeks, line 218: Nitpick (not required to change but would be more consistent with other tests): use `assert.isTrue` instead

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
adamweeks: I see in the tests that we could get them out of order. Should we do a sort on the array instead?

buckrivard: Perhaps.. I wasn't exactly sure if it _was_ necessary, except that I found it in the _list code and thought that it might be a patch for unexpected server response ordering or something like that. It's simply checking if the first is newer than the last, and if so flipping the list. I'm not sure if it is needed at all.

PR #1830, opened 2020-08-07
feat(conversation): add activity threading fetch methods

buckrivard:
Add method to retrieve activities in thread order. Thread order means that each parent message is zipper-merged with any of its children, rather than being returned in simple chronological order.

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/activity-thread-ordering.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/activity-thread-ordering.js
@@ -2,17 +2,17 @@ import {sortBy, last} from 'lodash';
 
 // use accessors for ease of refactoring underlying implementation
 /**
-       * @param {Map} destination destination map object for data. Currently a Map, but could be refactored to use any type.
-       * @param {string} key
-       * @param {any} value
-       * @returns {Map}
-       */
+ * @param {Map} destination destination map object for data. Currently a Map, but could be refactored to use any type.
+ * @param {string} key
+ * @param {any} value
+ * @returns {Map}
+ */
 export const setValue = (destination, key, value) => destination.set(key, value);
 /**
-             * @param {Map} source source map object to access. Currently expects a Map, but could be refactored to use any type
-             * @param {string} key
-             * @returns {Map}
-             */
+ * @param {Map} source source map object to access. Currently expects a Map, but could be refactored to use any type
+ * @param {string} key
+ * @returns {Map}
+ */
 export const getValue = (source, key) => source.get(key);
 
 export const defaultMinDisplayableActivities = 20;
@@ -36,7 +36,6 @@ export const ACTIVITY_TYPES = {
   DELETE: 'DELETE'
 };
 
-// TODO: create type for tombstone edit? could make it easy to ignore older edits...
 export const getActivityType = (activity) => {
   if (activity.activityType === 'reply') {
     return ACTIVITY_TYPES.REPLY;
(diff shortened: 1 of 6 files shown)


buckrivard:
The notable change between this PR and the [draft threading PR](https://github.com/webex/webex-js-sdk/pull/1809) is the use of the "revealing module pattern", which is a _very_ JS-y way to encapsulate state and expose only necessary logic to interact with the encapsulated state. These functions do not need to be invoked more than once in a given "work loop", or a given batch of activities being fetched, and therefore do not require newing up multiple instances of a class. The benefit is twofold: state encapsulation and method line count shrinkage.

adamweeks:
Could you please add an integration test that does the entire fetch flow? This will give us two things: validating that the flow works as designed and gives a good example of how to use the threading. I know a lot of SDK consumers will look to integration tests for examples of how to use the functionality.

packages/node_modules/@webex/internal-plugin-conversation/src/activity-thread-ordering.js
```diff
@@ -0,0 +1,331 @@
...
+  };
+};
+
+export const getLoopCounterFailsafe = () => {
+  let fetchLoopCount = 0;
+
+  return () => {
+    fetchLoopCount += 1;
+    if (fetchLoopCount > fetchLoopCountMax) {
+      throw new Error('max fetches reached');
```
holsted, line 281: Does throwing here exist the generator gracefully? Changing it to `done` or will it throw an error in the consuming code?

packages/node_modules/@webex/internal-plugin-conversation/src/activity-threading.md
```diff
@@ -0,0 +1,263 @@
...
+  5. edit
+  6. edit
+  7. reply
+  8. reaction
+  9. meeting
+  10. reply
+  
+Client: ...
+
+Convo: ...
```
holsted, line 24: 😂

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1461,109 +1584,490 @@ const Conversation = WebexPlugin.extend({
...
+      // INNER LOOP
+      // responsible for fetching and building our maps of activities
+      // fetch until minActivities is reached, or no more acts to fetch, or we hit our max fetch count
+      // ***********************************************
+
+      const incrementLoopCounter = getLoopCounterFailsafe();
+
+      while (!getNoMoreActs()) {
+        // count loops and throw if we detect infinite loop
+        incrementLoopCounter();
```
holsted, line 1840: 👍

packages/node_modules/@webex/internal-plugin-conversation/src/activity-thread-ordering.js
buckrivard: It will bubble up to caller, where they can catch and respond as they like

packages/node_modules/@webex/internal-plugin-conversation/src/activity-thread-ordering.js
```diff
@@ -0,0 +1,331 @@
...
+             * @param {Map} source source map object to access. Currently expects a Map, but could be refactored to use any type
+             * @param {string} key
+             * @returns {Map}
+             */
+export const getValue = (source, key) => source.get(key);
+
+export const defaultMinDisplayableActivities = 20;
+export const minBatchSize = 10;
+export const fetchLoopCountMax = 100;
+export const batchSizeIncrementCount = 10;
```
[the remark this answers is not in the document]
buckrivard: These are configurable via options object, however, leaving them out of the config will default to these values. Well, I lied. fetchLoopCountMax is not configurable.

buckrivard:
Opened in new PR: https://github.com/webex/webex-js-sdk/pull/1833

PR #1838, opened 2020-08-13
fix(conversation): fix threading iterator done value calculation

buckrivard:
Fixes bug where listActivitiesThreadOredered (wrapper function for _listActivitiesThreadOrdered) determines the "done" value of its iteratator response incorrectly.


buckrivard:
> Also, can your commit message describe what got changed?

Maybe you can give me an example of the sort of commit message you'd like? I don't usually write verbose commit messages

lalli-flores:
> Maybe you can give me an example of the sort of commit message you'd like? I don't usually write verbose commit messages

Thanks for adding the test @buckrivard! Code looks good.
For this PR, can you squash your commits to one? No need for a "refactor" commit since code is not part of the master yet.
As for the commit message, your test description is a good indicator of the change.
Based on your test, this is my commit message suggestion:
```
fix(conversation): return true when no more activities can be fetched

PR #1839, opened 2020-08-17
fix(meeting-plugin): media:stopped event fires correct payload

abailey-wwt:
Event media:stopped does not fire if media was not stopped.
This prevents multiple firing. Also fires correct payload.

Currently the 'media:stopped' event payload is "remote" and not "remoteAudio" or "remoteVideo" as per the docs. The event will fire regardless if media tracks where present or not, leading to the event occurring multiple times.

Fixes #SPARK-167668

---


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1547,13 +1547,16 @@ export default class Meeting extends StatelessWebexPlugin {
...
    * Removes the remote stream on the class instance and triggers an event
-   * to developers
+   * to developers for both remote stream and remote share
    * @returns {undefined}
    * @public
    * @memberof Meeting
    */
   unsetRemoteStream() {
-    this.mediaProperties.unsetRemoteStream();
+    // Unset both remote stream and remote share
```
lalli-flores, line 1556: Thanks for adding the comment, but seems like it's a note for the PR review? Maybe this comment shouldn't be part of the code base. With the JSDoc update you made I think the change should be covered

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1718,18 +1757,20 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   closeLocalStream() {
     return Media.stopTracks(this.mediaProperties.audioTrack)
```
lalli-flores, line 1759: Is there a reason to not use `stopStream` here instead?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1741,18 +1782,21 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    return Media.stopTracks(this.mediaProperties.shareTrack).then(() => {
-      Trigger.trigger(
-        this,
-        {
-          file: 'meeting/index',
-          function: 'closeLocalShare'
-        },
-        EVENT_TRIGGERS.MEDIA_STOPPED, {
-          type: EVENT_TYPES.LOCAL_SHARE
+    return Media.stopTracks(this.mediaProperties.shareTrack)
```
lalli-flores, line 1785: Is there a reason to not use `stopStream`?

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -468,34 +468,38 @@ Media.toggleStream = () => {};
 /**
  * Stop input stream
  * @param {MediaTrack} track A media stream
- * @returns {null}
+ * @returns {Boolean} status Returns true if a media track was stopped.
```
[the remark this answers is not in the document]
abailey-wwt: The entire file uses the convention {Boolean}, wouldn't it make sense to follow convention for consistency and then open another PR to normalize case? Maybe even enforce it if you desire that. I'm hesitant to introduce inconsistencies, would prefer it be addressed in a separate JIRA issue.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -504,14 +508,20 @@ Media.stopStream = (stream) => {
    * 2) Firefox requires the stream to be stopped
    */
   return Promise.resolve().then(() => {
+    let didStopMedia = false;
+
     if (stream.getTracks) {
       stream.getTracks().forEach((track) => {
         track.stop();
+        didStopMedia = true;
```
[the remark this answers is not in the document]
abailey-wwt: Not exactly, the issue is that **"media:stopped"** event is triggered regardless if media was stopped or not. By default `didStopMedia` is **false** meaning nothing was stopped, then set to **true** if any track in the stream was stopped.

packages/node_modules/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -61,6 +61,14 @@ export default class MediaProperties {
   }
 
   setRemoteStream(remoteStream) {
+    const stream = this.remoteStream;
+    const isCompatible = stream && stream.getTracks && stream.addTrack;
```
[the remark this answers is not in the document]
abailey-wwt: This is check is mainly for[ IE and Opera for Android](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/getTracks#Browser_Compatibility) which doesn't support `MediaStream.getTracks()`. In case someone ran Opera on Android the code would still function and not throw errors.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: Appears to be convention in the code file to add comments in parts that later on down the road might raise an eyebrow. If you scroll above there's TODOs and other various comments in the code. Still if you prefer I can remove it.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: Yes, I noticed the original code using `stopTracks`. This is due to the fact that local video, audio and share tracks are not stored in a stream in the `meeting.mediaProperties` instance. They are stored as individual tracks.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: Yes because it's a single track and not a stream. So `.stopStream()` calls `.getTracks()` on the stream to get the individual tracks but this is already a track itself.

abailey-wwt:
> Can you add some unit tests for `stopTracks` and `stopStream`?

...okay, you're not happy I added some new tests and fixed two others calling the wrong methods? Geez! No just kidding, sure I'll add it and update. No problem!

Thanks for poking around the PR, you asked a lot of insightful questions which forces me to explain the whats and the whys. 👍

packages/node_modules/@webex/plugin-meetings/src/media/index.js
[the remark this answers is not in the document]
lalli-flores: Up to you in this case. I'm suggesting based on the standards that we want to use in the future, which will be enforced by tooling. With that said, not sure if we will be porting those to this current SDK to make a difference.
So your call! I prefer consistency too, so if that's what it's being used in the file then it's fine to keep it like that.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
lalli-flores: I think we should remove it, unless like you mentioned, there is a `TODO` item.

packages/node_modules/@webex/plugin-meetings/src/media/properties.js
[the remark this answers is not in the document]
lalli-flores: Right, not questioning the check but rather what is being checked against. Should you be checking on `remoteStream` and not `this.remoteStream`?

packages/node_modules/@webex/plugin-meetings/src/media/index.js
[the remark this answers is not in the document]
lalli-flores: Not sure I agree that's the right approach. We should emit the `media:stopped` event only if we successfully stopped the stream, otherwise we are falsely telling the developer that the streamed has stopped when in case of error it's only partially stopped.

packages/node_modules/@webex/plugin-meetings/src/media/properties.js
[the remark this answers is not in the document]
abailey-wwt: I'm invoking `.getTracks()` from `this.remoteStream`, and I'm also checking to see if a previous stream already exists. I'm getting tracks from one stream and combining it with the other stream. Previously it just replaced the streams which caused a strange bug in that the remote audio never ended.  If you want I can alter the code to:

```suggestion
    const isCompatible = stream && stream.getTracks && remoteStream.addTrack;
```

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: Yes there are other TODOs and comments that already existed, I was just attempting to follow convention. Okay I removed it.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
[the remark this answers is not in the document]
abailey-wwt: Due to the nature of the code, most of the time these streams only contain one track. In fact, it's remoteStream that is the only stream that might contain both audio and video tracks. So we would never be falsely telling the developer. 

The odds of invoking `.getTracks()` having an array of two tracks and looping through and only stopping one of them is very very low.  A `MediaTrack` and `.stop()` method on a track is all native code.

It's really going to be an all or nothing type of deal in reality.

packages/node_modules/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -61,6 +61,14 @@ export default class MediaProperties {
...
+
+    if (isCompatible) {
+      // Combine audio and video tracks into one stream
+      stream.getTracks().forEach((track) => remoteStream.addTrack(track));
```
arun3528, line 69: @anthonywebb why do we need to add tracks here !? you can use the media:ready event to save the stream,

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1160,9 +1163,23 @@ describe('plugin-meetings', () => {
...
+          Media.stopStream = sinon.stub().returns(Promise.resolve(true));
+          await meeting.closeRemoteShare();
+          assert.calledOnce(TriggerProxy.trigger);
+          assert.calledWith(
+            TriggerProxy.trigger,
+            sinon.match.instanceOf(Meeting),
+            {file: 'meeting/index', function: 'closeRemoteShare'},
+            'media:stopped'
+          );
+          assert.notOk(meeting.mediaProperties.shareShare);
```
lalli-flores, line 1177: Can you use a more specific assertion? Like `isNull` or `isFalse`

abailey-wwt:
@arun3528 - Please take a look, I believe I found a middle ground where I applied a lot of changes you requested without making any major changes to the public API or closing streams via PeerConnection which might cause unexpected behavior with reconnection.

- Deprecated public APIs instead of removing them. I assume they can be removed at a later point in time.
- Moved away from concept of streams to tracks per WebRTC direction.
- Altered mediaProperties to store remote tracks while always providing backwards compatible code in case someone invokes the deprecated methods.
- Checking the `readyState` property of the media track to know whether or not to fire the `media:stopped` event.
- Fixes the audio connection not closing.
- Updated breaking tests.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1562,21 +1570,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    LoggerProxy.logger.warn('Meeting:index#closeRemoteStream --> [DEPRECATION WARNING]: closeRemoteStream has been deprecated after v1.89.3');
+    this.closeRemoteTracks();
+  }
+
+  /**
+   * Removes the remote tracks on the class instance and triggers an event
+   * to developers
+   * @returns {undefined}
+   * @memberof Meeting
+   */
```
arun3528, line 1585: we dont need this function , A developer should not close any remote tracks on there side

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1562,21 +1570,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      return Media.stopTracks(track)
+        .then(() => {
+          const isTrackStopped = track && track.readyState === ENDED;
+          const isWrongReadyState = track && !isTrackStopped;
+
+          if (isTrackStopped) {
+            triggerMediaStoppedEvent(type);
+          }
+          else if (isWrongReadyState) {
+            LoggerProxy.logger.warn(`Meeting:index#closeRemoteTracks --> Error: MediaStreamTrack.readyState is ${track.readyState} for ${type}`);
```
arun3528, line 1631: When would this get triggred ?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1717,19 +1778,29 @@ export default class Meeting extends StatelessWebexPlugin {
...
+              file: 'meeting/index',
+              function: 'closeLocalStream'
+            },
+            EVENT_TRIGGERS.MEDIA_STOPPED, {
+              type: EVENT_TYPES.LOCAL
+            }
+          );
+        }
+        else if (audioTrack || videoTrack) {
+          LoggerProxy.logger.warn('Meeting:index#closeLocalStream --> Error: MediaStreamTrack.readyState is incorrect.');
```
arun3528, line 1802: Should we print the readyState here

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: You're right but this function is private. Since we are deprecating the public one,  it's here for backwards compatibility. An effort to minimize any unforeseen issues with removing the public API that's already been exposed.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: This will get triggered if we have a media track but the `readyState` is not equal to the constant `ENDED`. The `else if` is to avoid this getting logged if the function is called but the tracks on `mediaProperties` has already been set to `null`.  

Keep in mind right now `Media.cleanUp` is called at least twice, with it check it you won't get warnings if the tracks have already been cleared.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-wwt: I did at first but it resulted in two log statements (one for audio and video each), I figured one warning log statement per function is probably best to avoid too many logs. I can put it back.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1122,6 +1132,26 @@ describe('plugin-meetings', () => {
...
+        sandbox.stub(meeting.mediaProperties, 'shareTrack').value(fakeMediaTrack());
+        sandbox.stub(meeting.mediaProperties, 'remoteAudioTrack').value(fakeMediaTrack());
+        sandbox.stub(meeting.mediaProperties, 'remoteVideoTrack').value(fakeMediaTrack());
+        sandbox.stub(meeting.mediaProperties, 'remoteShare').value(fakeMediaTrack());
+      });
+      afterEach(() => {
+        sandbox.restore();
+        sandbox = null;
+      });
+
```
abailey-wwt, line 1154: Also something I ran into, a lot of tests do this:

```javascript
someObject.someProperty = sinon.stub().returns(Promise.resolve());
```

which is okay except it doesn't restore the value and it was throwing off other tests further down. I think using sinon sandbox is great way to stub or mock what you need and then restore values after.

Or at least capture the stub to restore it:

```javascript
const somePropertyStub = sinon.stub(someObject, 'someProperty').returns(Promise.resolve());
// ... later on
somePropertyStub.restore();
```

anyhow just my two cents.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: thanks

PR #1840, opened 2020-08-18
fix(conversation): threads iterator 'done' is true when no more activities can be fetched

buckrivard:
Fixes bug where listActivitiesThreadOredered (wrapper function for _listActivitiesThreadOrdered) determines the "done" value of its iteratator response incorrectly.


buckrivard:
Re-opening https://github.com/webex/webex-js-sdk/pull/1838 in squashed branch

Issue #1694, opened 2020-06-04
Updating meeting quality level changes video device


abailey-wwt:
## Expected Behavior
The currently selected video device should not change when updating the meeting quality level.

## Current Behavior
Updating the meeting quality while inside an on-going meeting assumes default media settings, thus activating the default video device and changing the video of the participant. 

## Possible Solution (we welcome any logical suggestion)
Alter the current implementation to allow passing an object identifying the desired video device id to use when getting media streams. If omitted then use defaults.

```
{ video: { deviceId: { exact: ...  } } }
```

## Steps to Reproduce
1. Visit https://js.samples.s4d.io/browser-plugin-meetings/
2. Enter token / credentials.
3. Create a meeting.
4. Join the meeting by clicking “**JOIN WITH MEDIA**”.  _Note:_ The camera you joined the meeting with should be your default camera.
5. Change the video source to a different camera other than the one you joined with.
6. Click **updateVideo** button to see change happen.
7. Update meeting quality to LOW. (_Take notice as the camera feed reverts back to the default camera._)

Same happens when setting local video quality as well. 

## Environment Details
* SDK Version: 1.8.220
* Environment: Mac OS 10.15.4
* Node: 10.20.1
* Browser Version: Chrome Version 83.0.4103.61
* NPM Version: 6.14.4

abailey-wwt:
Resolved by #1846

PR #1846, opened 2020-08-25
fix(meeting-plugin): getMediaStreams uses last used video device

abailey-wwt:
When first reported updating meeting quality level changed the video device if you had two video devices and was using your non-default camera. When recently tested using the samples, the local video stream of the caller did not change but the video displayed to the remote party did change to the other camera. Therefore the caller was unaware of that the remote party sees a different video stream.

If the caller leaves a meeting without closing the browser window and attempts to rejoin then the last camera used will be remembered. To rejoin the caller must getting media streams, click the join button and then click on the add media button.

Caller A - JS SDK (calls into space, must have two cameras)
Remote Party B - Native Desktop client (joins space)

**Steps to reproduce:**
Caller A should start the meeting, update their video to a non-default camera. Then proceed to change the video quality and note that the remote party sees a different video stream from the caller.

Fixes #SPARK-148663
GitHub Issue: #1694 

---

✈️ (wing-wave)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/media/properties.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/media/properties.js
@@ -28,6 +28,7 @@ export default class MediaProperties {
     this.localQualityLevel = options.localQualityLevel || QUALITY_LEVELS.HIGH;
     this.remoteQualityLevel = options.remoteQualityLevel || QUALITY_LEVELS.HIGH;
     this.mediaSettings = {};
+    this.videoDeviceId = null;
 
     // deprecated after v1.89.3, remove when feasible.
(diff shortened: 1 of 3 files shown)


abailey-wwt:
Sure I can add a few tests to this, no problem. Regarding moving the code, tomorrow's my last day. I was hoping to wrap up any loose ends from my side and have a clean/smooth transition.

arun3528:
wanted to get your opinion about it

abailey-wwt:
Gotcha! That makes sense Arun as it also good for separation of concerns. A lot of methods require a meeting object to be created first when you don't necessarily need the meeting object. So yes I think that's a good idea!

PR #1854, opened 2020-08-26
Remote start/stop event

taymoork2:
Did some start sharing the ~~love~~... i mean the screen

Fixes [#SPARK-166067](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-166067)

---


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -888,8 +888,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
           receiveShare: this.mediaProperties.mediaDirection.receiveShare
+        }).then(() => {
+          Trigger.trigger(
+            this,
+            {
+              file: 'meetings/index',
+              function: 'remoteShare'
+            },
+            EVENT_TRIGGERS.MEETING_STARTED_SHARING_REMOTE
+          );
```
mwegman, line 899: This only works if the current user had the share before someone else took it. You want to do the opposite conditional of line 903-904 to trigger this.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -891,6 +891,35 @@ export default class Meeting extends StatelessWebexPlugin {
         });
       }
```
arun3528, line 893: @taymoork2  do you want to add the then block for updateShare and then do the stop as it might cause start then stop local

arun3528: 1) check localStopped is triggering before remote start , ? not. add then block to updatShare and add start remoteL logic there

2) In a 3 party call user A receiving the share events , Remote Share start from user B and Stopped remote share user B then start Remote SHare user C ? add the check for previous and current content ID

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: @mwegman does it look ok now ? or we are still missing the condition you mentioned ?

PR #1855, opened 2020-08-27
fix(activity threading): pass target URL through to threading function when calling jumpToActivity

buckrivard:
This change helps to keep all three threading methods in sync with one another by re-initializing the generator function with the convo URL of the activity that is already being passed in for the jumpToActivity helper (aka search, or midDate query). Rather than forcing the caller to re-initialize all three threading methods with a new URL, this change will automatically update the URL used to make activity requests when the caller passes in an activity to search for.


packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/get.js
```diff
@@ -814,17 +814,72 @@ describe('plugin-conversation', function () {
...
+            })
+            .then((acts) => {
+              actCount = acts.length;
+
+              return jumpToActivity(msg);
+            })
+            .then(() => {
+              assert.isTrue(actCount < minActivities);
+            });
+        });
```
mwegman, line 864:
```suggestion
        it('should return all activities in a space when jumping to an activity in a space with less activities than the asked-for activities limit', () => {
          return webex.internal.conversation.create({participants: [scott.id]})
            .then((c) => {
              return webex.internal.conversation.post(c, {displayName: 'first message'})
                .then((m) => {
                  return webex.internal.conversation.listActivities({conversationUrl: c.url})
                    .then((acts) => {
                      return jumpToActivity(m)
                        .then(() => acts.length);
                    })
                })
            })
            .then((actCount) => {
              assert.isTrue(actCount < minActivities);
            });  
        });
```

mwegman: You don't need to save off variables. You can pass them down or keep them in scope.

PR #1860, opened 2020-09-02
chore(release): v1.92.2 [skip ci]

mwegman:


lalli-flores:
@mwegman, can you add a description in the PR as to why you're manually setting the SDK version?

PR #1867, opened 2020-09-03
fix(plugin-meeting): meeting object not deleted on end

arun3528:
Fixes issue where if recording is not found it does not delete the meeting object from the store 
---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
@@ -41,19 +41,19 @@ ControlsUtils.parse = (controls) => {
 ControlsUtils.getControls = (oldControls, newControls) => {
   const previous = ControlsUtils.parse(oldControls);
   const current = ControlsUtils.parse(newControls);
-  const hasRecord = (controls) => controls && controls.record;
+  const hasRecord = (controls) => controls && !!controls.record;
 
   return {
     previous,
     current,
     updates: {
-      hasRecordingPausedChanged: !isEqual(
+      hasRecordingPausedChanged: hasRecord(current) && !isEqual(
         hasRecord(previous) && previous.record.paused,
-        hasRecord(current) && current.record.paused
+        current.record.paused
       ),
-      hasRecordingChanged: !isEqual(
+      hasRecordingChanged: hasRecord(current) && !isEqual(
         hasRecord(previous) && previous.record.recording,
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -47,11 +47,11 @@ ControlsUtils.getControls = (oldControls, newControls) => {
     previous,
     current,
     updates: {
-      hasRecordingPausedChanged: !isEqual(
+      hasRecordingPausedChanged: current.record && !isEqual(
```
mwegman, line 50: Why do you need to do this check? If it was paused (`previous.record.paused === true`) and now there isn't a `current.record`, wouldn't that be a `recording pause` change?

mwegman: If you need to have the check, perhaps the following change:
```suggestion
      hasRecordingPausedChanged: hasRecord(current) && !isEqual(
```

akoushke: yeah I'm not following this either. @arun3528 let's talk tomorrow about this.

arun3528: When it comes to meeting end or removed the last event which comes in has controls(current) but it does not have record. on it so had to add that check

akoushke: right we are checking that before the comparison
```js
hasRecord(current) && current.record.recording
```

arun3528: sorry i meant to say current.record is not found

akoushke: can we do this instead then? or something similar to this?
```js
 const hasRecord = (controls) => controls && typeof controls.record !== 'undefined';
```

I'm just trying to have the check logic together instead of having it there.

akoushke: actually this works too:

```js
const hasRecord = (controls) => controls && !!controls.record;
```

akoushke (approved):

mwegman (approved):

PR #1872, opened 2020-09-08
fix(ediscovery): removing depreciated summary apis

gannons:
Fixes #[SPARK-160736](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-160736)

On a slave with a bad network connection we started to notice failures in the E2E tests. Looking into the issue there were two outstanding items.

1) A pre-populated cache was not being used
2) A network call received an error it did not retry

This PR is to fix the first issue. Here the problem was simple. There were two caches, in place to support an old API, and the incorrect cache was being used by a decrypt call in the ediscovery module. i.e. It was an oversight.

To address this issue I'm **removing** support for the old _sumary_ APIs. Instead consumer are expected to use the new _container_ APIs. This is already the case with the ediscovery downloader which is the only consumer of this module.


packages/node_modules/@webex/internal-plugin-ediscovery/src/ediscovery.js
```diff
@@ -41,12 +41,6 @@ const EDiscovery = WebexPlugin.extend({
   namespace: 'EDiscovery',
 
   session: {
-    contentSummaryBySpaceIdCache: {
-      type: 'object',
-      default() {
-        return new Map();
-      }
-    },
```
gannons, line 49: The `contentSummaryBySpaceIdCache` holds the results returned from ediscovery's `/summary` API. i.e. Metadata about each _space_ in a report. The cache gives the client faster performance and prevents rate limiting on the server. Previously when the eDiscovery Downloader started a report it would pre-populate this cache.

Later a new API was introduced as eDiscovery expanded to handle meetings as well as spaces. The `/container` API replaces `/summary` and has been adopted by the eDiscovery Downloader. However the old API remains on the server to support legacy clients.

But there was a dependency on `contentSummaryBySpaceIdCache` in `transforms.js` that was missed. When the cache was no longer pre-populated the code used a fallback strategy. It queried the `/summary` API for all spaces not found in the cache.

When the query by space id hits a network error it gives up. This meant all activities for that space cannot be decrypted. Essentially the space will be missing from the final eDiscovery report.

This is being fixed here by removing all references to, and uses of, the `contentSummaryBySpaceIdCache`. That includes the removal of the old `/summary` APIs. Instead the code will use the pre-populated `contentContainerCache`.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -175,34 +175,46 @@ class Transforms {
 
     const promises = [];
 
-    return ctx.webex.internal.ediscovery.getContentSummaryBySpaceId(reportId, activity.targetId)
+    return ctx.webex.internal.ediscovery.getContentContainerByContainerId(reportId, activity.targetId)
```
gannons, line 178: Switch over to the new `/container` API rather than call the, now empty initially, space summary cache.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -175,34 +175,46 @@ class Transforms {
...
           return Promise.resolve(object);
         }
 
-        // add space warning properties to the activity - these will be recorded in the downloader
-        if (space.warning) {
-          activity.spaceWarning = space.warning;
+        // add warning properties to the activity - these will be recorded in the downloader
+        if (container.warning) {
+          activity.spaceWarning = container.warning; // Depreciate this property
+          activity.containerWarning = container.warning;
```
gannons, line 194: Keeping the old property for now as the client still uses it.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -175,34 +175,46 @@ class Transforms {
...
         }
-        else if (space.isOneOnOne && space.participantDisplayNames && space.participantDisplayNames.length === 2) {
+        else if (container.isOneOnOne) {
+          let displayNames = '';
+
+          for (const participant of new Set([...container.participants || [], ...container.formerParticipants || []])) {
+            if (displayNames) {
+              displayNames += ' & ';
+            }
+            displayNames += participant.displayName;
```
gannons, line 209: There was a small change in how `participants` are listed in the new API. Previously they were just UUIDs. Now they're participant objects.

gannons: It strike me that this code would be more appropriate in the client. However I may have argued the opposite in the past as it could duplicate the work if there were multiple clients. Keeping the same approach for now.

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -130,41 +130,37 @@ describe('EDiscovery Transform Tests', () => {
...
+        ctx.webex.internal.ediscovery = {getContentContainerByContainerId: sinon.stub().returns(Promise.resolve({body: contentContainer}))};
       });
 
-      afterEach(() => ctx.webex.internal.ediscovery.getContentSummaryBySpaceId.resetHistory());
+      afterEach(() => ctx.webex.internal.ediscovery.getContentContainerByContainerId.resetHistory());
 
       it('Calls the correct decrypt functions when transforming post activities', () => {
         const result = Transforms.decryptReportContent(ctx, object, reportId)
           .then(() => {
-            assert.callCount(ctx.webex.internal.encryption.decryptText, 1);
```
gannons, line 150: I'm not a fan of this style of testing. It's too concerned with **how** something is done versus **what** it does. I frankly don't care what decryption method is called as long as the results are what I'd expect.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -175,34 +175,46 @@ class Transforms {
...
+          }
```
mwegman, line 210:
```suggestion
          const displayNames = (container.participants || []).concat(container.formerParticipants || []).join(' & ');
```

mwegman (approved):

mwegman (approved):

PR #1877, opened 2020-09-13
fix: export getUserMedia correctly

holsted:
I tried to destructure an object that wasn't exported individually in my previous PR.


adamweeks (approved):
Could you add the `meetings` scope to the git commit so that it will show the change's scope on the changelog? Approved otherwise.

PR #1878, opened 2020-09-14
test(conversation): add test for midDate fetch conditions to include all asked-for activities

buckrivard:
Adds test coverage for an already-merged change. https://github.com/webex/webex-js-sdk/pull/1866

The test ensures that midDate fetches fetch all activities up to the minimum limit when a create activity is not present in the fetched list of activities.

Addresses https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-173370


packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/get.js
```diff
@@ -883,6 +883,30 @@ describe('plugin-conversation', function () {
...
+                .then(() => webex.internal.conversation.post(c, {displayName: 'message 4'})
+                  .then(() => webex.internal.conversation.post(c, {displayName: 'message 5'})
+                    .then(() => webex.internal.conversation.post(c, {displayName: 'message 6'})
+                      .then(() => webex.internal.conversation.post(c, {displayName: 'message 7'})
+                        .then(() => webex.internal.conversation.post(c, {displayName: 'message 8'})
+                          .then(() => webex.internal.conversation.post(c, {displayName: 'message 9'})
+                            .then(() => webex.internal.conversation.post(c, {displayName: 'message 10'})
+                              .then(() => webex.internal.conversation.post(c, {displayName: 'message 11'})
+                                .then(() => webex.internal.conversation.post(c, {displayName: 'message 12'})
+                                  .then(() => webex.internal.conversation.post(c, {displayName: 'message 13'})
```
adamweeks, line 899: Could this be collapsed into a `for` block to send all the messages instead of a giant nesting block? 

```js
  const results = [];
  for (i = 0; i < 13; i++) {
    // Good: all asynchronous operations are immediately started.
    results.push(webex.internal.conversation.post(c, {displayName: `message ${i}`}));
  }
  // Now that all the asynchronous operations are running, here we wait until they all complete.
  return Promise.all(results));
}
```

buckrivard: I have no problem with that. Change incoming.

packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/get.js
```diff
@@ -883,6 +883,26 @@ describe('plugin-conversation', function () {
...
+            const $posts = [];
+
+            // eslint-disable-next-line no-plusplus
+            for (let i = 0; i < 15; i++) {
+              $posts.push(webex.internal.conversation.post(c, {displayName: `message ${i}`}));
+            }
+
+            return Promise.all($posts);
+          })
+          .then((posts) => posts[posts.length - 1]).then((lastPost) => jumpToActivity(lastPost))
```
adamweeks, line 897: Sorry I didn't think about this in my first review, if you want to make sure you get the very last post, sending one final post after your initial batch might make the test more resilient. We could get a promise resolution that takes longer to return for one of the non-last posts that could break the test.

mwegman (approved):

adamweeks (approved):

PR #1879, opened 2020-09-15
feat(samples): add remote share event support

adamweeks:
No JIRA for this one, just added support for the new remote share events to the sample.

---


mwegman (approved):

arun3528 (approved):
Looks good , we decided to move towards streams model because in SDK we never use remote streams for any operation but we need tracks ,

So the developer needs to use the streams from the media:ready event,

Is it possible to change it to use streams from media ready, If possible mention in the event docs for startSharing

arun3528 (changes requested):
need to use the streams from media:ready

adamweeks:
@arun3528 updated

arun3528 (approved):

PR #1882, opened 2020-09-17
fix(meeting-plugin): send share error metrics

arun3528:
Currently getDisplayMedia failure metrics is only being used to track the getDisplayMedia failure

We dont have any for screen share failures and the promise does not reject if share fails

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -3523,18 +3523,19 @@ export default class Meeting extends StatelessWebexPlugin {
     };
 
     return Media.getDisplayMedia(shareConstraints, this.config)
-      .then((shareStream) => {
-        this.updateShare({
-          sendShare: true,
-          receiveShare: this.mediaProperties.mediaDirection.receiveShare,
-          stream: shareStream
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -3523,18 +3523,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        sendShare: true,
+        receiveShare: this.mediaProperties.mediaDirection.receiveShare,
+        stream: shareStream
+      }))
       .catch((error) => {
         // Whenever there is a failure when trying to access a user's display
         // report it as an operational metric
         // This gives visibility into common errors and can help
         // with further troubleshooting
+
```
adamweeks, line 3536: This will now catch an error on both `getDisplayMedia` and `updateShare`, does the code in the catch still apply or does it need to be updated?

arun3528: For now yes we need metrics for the share failures so we are going to use this event and rename it something else laters rather then getDisplayMedia as its a single feature.

adamweeks (approved):

PR #1884, opened 2020-09-18
fix(meeting): trigger correct event at the initiation level

akoushke:
There is no Issue number associated to this work.
problem was that the recording status was not correct when a user joins the meeting on a `paused` recording status. 
the fix was ordering the status priority accordingly.
---


packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -155,6 +160,8 @@ describe('plugin-meetings', () => {
           shareControl: {},
           transcribe: {}
         };
+        // there must be a recording to be puased/resumed
```
adamweeks, line 163:
```suggestion
        // there must be a recording to be paused/resumed
```

adamweeks (approved):
Looks good, thanks for adding the extra tests too. The `if` order is a bit confusing to me since I'm not familiar with it and it seems we're overwriting the `state` value a few times which is probably where my confusion lies. That being said, I see no blockers but would like a review from @arun3528 or @holsted as well before merging.

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -645,17 +645,23 @@ export default class LocusInfo extends EventsScope {
...
+      if (hasRecordingChanged || hasRecordingPausedChanged) {
         let state = null;
 
-        if (updates.hasRecordingChanged) {
+        // Ordered IF Statements!
+        // `idle` > `paused` > `recording` > `resumed`
+        if (hasRecordingChanged) {
           state = current.record.recording ? RECORDING_STATE.RECORDING : RECORDING_STATE.IDLE;
         }
-        else if (updates.hasRecordingPausedChanged) {
```
arun3528, line 656: @akoushke  do you get recording as true when recording and paused is a additional values on top of recording !?

akoushke: yes

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -645,16 +645,21 @@ export default class LocusInfo extends EventsScope {
...
+      if (hasRecordingChanged || hasRecordingPausedChanged) {
         let state = null;
 
-        if (updates.hasRecordingChanged) {
-          state = current.record.recording ? RECORDING_STATE.RECORDING : RECORDING_STATE.IDLE;
+        if (hasRecordingPausedChanged) {
+          if (current.record.paused) {
+            state = RECORDING_STATE.PAUSED;
+          }
+          else {
```
arun3528, line 657: just add a small comment here , rest looks good

arun3528 (approved):
small comment on the If will be helpful

PR #1887, opened 2020-09-18
chore(circleci): use CACHE_VERSION for cache control

adamweeks:
Gives us the ability to clear cache on circle as needed:
https://circleci.com/docs/2.0/caching/#clearing-cache

---


mwegman (approved):

mwegman:
Do you know if there is a way to set a default if the variable isn't defined?

adamweeks:
> Do you know if there is a way to set a default if the variable isn't defined?

@mwegman I do not, but in this case it would just make the cache string `--` instead of `-V1-`

PR #1889, opened 2020-09-21
feat(meeting): initialize recording property from the locus info

akoushke:
SPARK-178087
---
currently we don't which meetings are supporting the `recording` feature by locus. this check will allow us to identify that. if `recording` is null, it means that the meeting does not support recording AKA 1:1 and free trial meetings.


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -619,7 +619,7 @@ export default class Meeting extends StatelessWebexPlugin {
      * @public
      * @memberof Meeting
      */
-    this.recording = null;
+    this.recording = attrs.locus?.controls.record ? {state: RECORDING_STATE.IDLE, modifiedBy: '', lastModified: ''} : null;
```
akoushke, line 622: I do prefer to define classes for different controls with an abstract called `Control` and have classes `Recording` `Locked` etc etc.

arun3528 (changes requested):
This logic needs to be moved in the locusInfo updateControl section , Placing it here wont trigger an event to the developer initially on recording ,

akoushke:
> This logic needs to be moved in the locusInfo updateControl section , Placing it here wont trigger an event to the developer initially on recording ,

and it shouldn't trigger an event at the initial level IMO.  events get triggered after the setup and `recording` should be defined before that the event setup.

akoushke:
I tested this and worked as expected.

akoushke:
1:1 
Screen Shot 2020-09-22 at 6 27 35 PM
meeting
Screen Shot 2020-09-22 at 6 28 26 PM

packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
```diff
@@ -41,19 +41,16 @@ ControlsUtils.parse = (controls) => {
...
-        hasRecord(previous) && previous.record.paused,
-        current.record.paused
+      hasRecordingPausedChanged: current?.record && !isEqual(
+        previous?.record?.paused, current.record.paused
       ),
-      hasRecordingChanged: hasRecord(current) && !isEqual(
-        hasRecord(previous) && previous.record.recording,
-        current.record.recording
+      hasRecordingChanged: current?.record && !isEqual(
+        previous?.record?.recording, current?.record?.recording
```
arun3528, line 53: nice so this will check undefined with false

arun3528 (approved):

PR #1900, opened 2020-10-02
fix(calendar): decrypt organizer email and name

robstax:
calendar service payloads include a `encryptedOrganizer` property, which has `encryptedEmailAddress` and `encryptedName` fields. currently, these properties are not decrypted. this PR fixes that.

if you'd like me to add tests for this change, please let me know where. i was thinking [here](https://github.com/webex/webex-js-sdk/blob/0bd7155d9a614b12cec844596d833580e94db427/packages/node_modules/%40webex/internal-plugin-calendar/test/integration/spec/calendar.js#L183) but i see this test is skipped...


lalli-flores (approved):
Looks good

PR #1902, opened 2020-10-08
feat: add preferTransocoding option

holsted:
This PR adds an optional parameter to the join meeting flows to allow transcoding (Edonus) for 1:1 calls. This fixes a bug in Chrome 86 that breaks remote video. 

The tests were also failing locally so made some resiliency updates to those and suppressed/mocked some errors.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/request.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/request.js
@@ -40,7 +40,7 @@ export default class MeetingRequest extends StatelessWebexPlugin {
    */
   async joinMeeting(options) {
     const {
-      sipUri, deviceUrl, locusUrl, resourceId, correlationId, ensureConversation, moderator, pin, moveToResource, roapMessage
+      sipUri, deviceUrl, locusUrl, resourceId, correlationId, ensureConversation, moderator, pin, moveToResource, roapMessage, preferTranscoding
     } = options;
 
     LoggerProxy.logger.info(
@@ -61,7 +61,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       respOnlySdp: true,
       allowMultiDevice: true,
       ensureConversation: ensureConversation || false,
-      supportsNativeLobby: 1
+      supportsNativeLobby: 1,
+      clientMediaPreferences: {
+        preferTranscoding: preferTranscoding ?? true
+      }
     };
 
     if (moderator !== undefined) {
@@ -311,7 +314,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       usingResource: options.resourceId || null,
       correlationId: options.correlationId,
       respOnlySdp: true,
-      localMedias: options.localMedias
+      localMedias: options.localMedias,
+      clientMediaPreferences: {
+        preferTranscoding: options.preferTranscoding ?? true
+      }
     };
 
     return this.request({
--- a/packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
+++ b/packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
@@ -33,7 +33,6 @@ skipInNode(describe)('plugin-meetings', () => {
       .then(() => Promise.all([testUtils.syncAndEndMeeting(alice),
         testUtils.syncAndEndMeeting(bob)]))
       .catch((error) => {
-        console.log(error);
         throw error;
       }));
 
@@ -93,7 +92,6 @@ skipInNode(describe)('plugin-meetings', () => {
           assert.equal(bob.webex.meetings.getMeetingByType('sipUri', alice.emailAddress), null);
         })
         .catch((err) => {
-          console.log('ERROR JOIN ', err);
           throw err;
         }));
     });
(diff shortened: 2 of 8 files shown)


chrisadubois (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/request.js
```diff
@@ -61,7 +61,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       respOnlySdp: true,
       allowMultiDevice: true,
       ensureConversation: ensureConversation || false,
-      supportsNativeLobby: 1
+      supportsNativeLobby: 1,
+      clientMediaPreferences: {
+        preferTranscoding: preferTranscoding || true
```
mwegman, line 66:
```suggestion
        preferTranscoding: preferTranscoding ?? true
```

packages/node_modules/@webex/plugin-meetings/src/meeting/request.js
```diff
@@ -311,7 +314,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       usingResource: options.resourceId || null,
       correlationId: options.correlationId,
       respOnlySdp: true,
-      localMedias: options.localMedias
+      localMedias: options.localMedias,
+      clientMediaPreferences: {
+        preferTranscoding: options.preferTranscoding || true
```
mwegman, line 319:
```suggestion
        preferTranscoding: options.preferTranscoding ?? true
```

mwegman (changes requested):

packages/node_modules/@webex/plugin-meetings/src/meeting/request.js
holsted: What does that do!?

mwegman: Instead of testing for `falsey`, it tests for `undefined` or `null`

mwegman: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

mwegman (approved):

packages/node_modules/@webex/plugin-meetings/test/integration/spec/testUtils.js
```diff
@@ -134,11 +134,22 @@ const waitForEvents = (scopeEvents) => {
               }
             }
 
-            if (result.length === scopeEvents.length) {
+            if (obj.match) {
+              if (obj.match(value)) {
```
mwegman, line 138:
```suggestion
            if (obj?.match(value)) {
```

packages/node_modules/@webex/plugin-meetings/test/integration/spec/testUtils.js
```diff
@@ -134,11 +134,22 @@ const waitForEvents = (scopeEvents) => {
               }
             }
 
-            if (result.length === scopeEvents.length) {
+            if (obj.match) {
```
mwegman, line 137: Why do the `if` statement? You do the exact same 3 methods either way.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -322,61 +314,54 @@ skipInNode(describe)('plugin-meetings', () => {
...
-          })
-      ])
-        .then(() => {
-          assert.equal(alice.meeting.audio.muted, true);
 
-          return testUtils.waitUntil(4000);
-        }));
+        return Promise.all([
+          testUtils.delayedPromise(alice.meeting.muteAudio()),
+          testUtils.waitForEvents([{scope: bob.meeting.members, event: 'members:update', match: checkEvent}])
```
mwegman, line 323: Above you moved the `waitForEvents` out of this `Promise.all`. Should you do the same thing for the following tests as well?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/testUtils.js
holsted: I left it for backwards compatibility. The previous test cases expected a specific value to be returned and was using `.once` above instead of `.on`. It's possible we can get various member updates (or other meeting updates) in the time the test is running. The idea was to go back at some point and move everything to use `match` since it's a little cleaner in the test itself.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/testUtils.js
holsted: I like it. Forgot I could do the `?.` now

packages/node_modules/@webex/plugin-meetings/test/integration/spec/testUtils.js
```diff
@@ -134,11 +134,20 @@ const waitForEvents = (scopeEvents) => {
...
+              clearTimeout(timer);
+              obj.scope.off(obj.event, handler);
+
+              resolve(result);
+            }
+            else if (result.length === scopeEvents.length) {
               resolve(result);
+              obj.scope.off(obj.event, handler);
               clearTimeout(timer);
             }
```
mwegman, line 147:
```suggestion
            if (obj?.match(value) || (result.length === scopeEvents.length)) {
              clearTimeout(timer);
              obj.scope.off(obj.event, handler);
              resolve(result);
            }
```

PR #1908, opened 2020-10-14
fix(webex-core): fix unit tests

mwegman:
---

### DIFF
--- a/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/token.js
+++ b/packages/node_modules/@webex/webex-core/test/unit/spec/credentials/token.js
@@ -197,7 +197,7 @@ describe('webex-core', () => {
         });
 
         it('computes expires_in and refresh_token_expires_in if not specified', () => {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/webex-core/test/unit/spec/credentials/token.js
```diff
@@ -197,7 +197,7 @@ describe('webex-core', () => {
         });
 
         it('computes expires_in and refresh_token_expires_in if not specified', () => {
-          clock.setSystemTime(0);
+          const now = Date.now();
```
adamweeks, line 200: The `lolex` package is supposed to do this all for us, any idea why it would stop working for this test?

mwegman: The new version of babel does the polyfill differently. It saves off the `Date` object before anything else happens. We can't mock it out in time for it to be effective.

adamweeks (approved):

PR #1915, opened 2020-10-26
fix(plugin-meetings): add MEETING_LEAVE_FAILURE constant

adamweeks:
Currently no jira, but an ad hoc fix to the missing constant that was causing the meetings plugin to crash after leave failure.


taymoork2 (approved):
LGTM

PR #1916, opened 2020-10-28
feat(space-classification): allow user to create a classified space

AMANDUA:
This PR comprises of code changes to allow user to create a classified space

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-176840
---

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
@@ -165,6 +165,9 @@ const Conversation = WebexPlugin.extend({
    * @param {string} params.comment
    * @param {string} params.html
    * @param {Object} params.displayName
+   * @param {string} params.classificationId
+   * @param {string} params.effectiveDate
+   * @param {Boolean} params.isDefaultClassification
    * @param {Object} options
    * @param {Boolean} options.allowPartialCreation
    * @param {Boolean} options.forceGrouped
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -2306,6 +2306,14 @@ const Conversation = WebexPlugin.extend({
       }));
     }
 
+    if (!params.isDefaultClassification && params.classificationId) {
```
adamweeks, line 2309: Please add these additional params to the jsdoc on the params object on the `create` function near line 167.

adamweeks (changes requested):
Thanks for adding this support. Could you please add an additional test on the `create` function that will call with this param?

adamweeks:
@AMANDUA 

> I tried adding test with existing orgs but the integration tests are failing saying org is not having space classifications enabled...

If we aren't able to get any integration orgs for this feature, perhaps you can add a test that expects the error when the feature isn't enabled at least.

AMANDUA:
> @AMANDUA
> 
> > I tried adding test with existing orgs but the integration tests are failing saying org is not having space classifications enabled...
> 
> If we aren't able to get any integration orgs for this feature, perhaps you can add a test that expects the error when the feature isn't enabled at least.

yes that sounds good to me...

AMANDUA:
Added the test @adamweeks  
image

and passed the  CI.

But few integration tests are failing , saying timeout
image

adamweeks (approved):
Merging. The test failures are the same failures as master branch currently experiencing: https://app.circleci.com/pipelines/github/webex/webex-js-sdk/4124/workflows/7d5fae21-b726-48ca-ac1b-dc3714fb90c7/jobs/29488

PR #1918, opened 2020-10-30
fix(authorization): optional chain

chrisadubois:
I could be way out of left field here... but i was trying to add emailhash autosignin functionality to the web client, and this was standing in my way. since the .catch always .resolve anyway, this just makes sure the function doesn't call it on silly params

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/services/services.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/services.js
@@ -8,6 +8,7 @@ import ServiceCatalog from './service-catalog';
 import ServiceRegistry from './service-registry';
 import ServiceState from './service-state';
 
+
 const trailingSlashes = /(?:^\/)|(?:\/$)/;
 
 /* eslint-disable no-underscore-dangle */
@@ -188,7 +189,7 @@ const Services = WebexPlugin.extend({
 
       formattedQuery = {};
 
-      if (queryKey === 'email') {
+      if (queryKey === 'email' && query.email) {
         formattedQuery.emailhash = sha256(query.email.toLowerCase()).toString();
       }
       else {


packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -8,6 +8,7 @@ import ServiceCatalog from './service-catalog';
 import ServiceRegistry from './service-registry';
 import ServiceState from './service-state';
 
+
```
InteractiveTimmy, line 11: This extra new line should be removed.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -189,7 +190,7 @@ const Services = WebexPlugin.extend({
       formattedQuery = {};
 
       if (queryKey === 'email') {
-        formattedQuery.emailhash = sha256(query.email.toLowerCase()).toString();
+        formattedQuery.emailhash = sha256(query.email?.toLowerCase())?.toString();
```
InteractiveTimmy, line 193: Does our babel/node version support this?

InteractiveTimmy (changes requested):
Small quick change request and a question.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
chrisadubois: indeed it does !

packages/node_modules/@webex/webex-core/src/lib/services/services.js
chrisadubois: this was an eslint thing... will remove, after SDK owner review.

packages/node_modules/@webex/webex-core/src/lib/services/services.js
adamweeks: Yes!

InteractiveTimmy (approved):
Approved! Thanks for the clarification!

packages/node_modules/@webex/webex-core/src/lib/services/services.js
adamweeks: eslint doesn't have a problem with it, so it is good to go!

adamweeks (approved):

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -189,7 +190,7 @@ const Services = WebexPlugin.extend({
       formattedQuery = {};
 
       if (queryKey === 'email') {
```
adamweeks, line 192: Would a better fix for this be:

```suggestion
      if (queryKey === 'email' && query.email) {
```

@InteractiveTimmy @chrisadubois

chrisadubois: yes but...optional chaining is finally in the SDK haha

chrisadubois: I'll change it

adamweeks (changes requested):
Thanks for updating the PR. 

Please squash the commits and give a more descriptive commit message for the changelog.

`fix(authorization): add additional email query check` perhaps?

adamweeks (approved):

PR #1920, opened 2020-11-04
chore: update node-jose

mwegman:
SPARK-183544 #done
---


patches/node-jose+2.0.0.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/node_modules/node-jose/lib/index.js b/node_modules/node-jose/lib/index.js
+index b10bec3..baa1f5e 100644
+--- a/node_modules/node-jose/lib/index.js
++++ b/node_modules/node-jose/lib/index.js
+@@ -9,18 +9,6 @@ if (typeof Promise === "undefined") {
+   require("es6-promise").polyfill();
+ }
+ 
+-if (typeof Buffer === "undefined") {
```
lalli-flores, line 9: Can you remind me why is the patch necessary?

mwegman: This section was added to get node-jose to work with React Native. It broke us. We can try to push the change back to a project that won't have a lot of changes or we can add this patch and get it to work.

lalli-flores: Can we do both?
Let's have the patch so we can close our bug, but we should fix it in the right place too in case things change in the future.

lalli-flores (approved):

PR #1923, opened 2020-11-06
feat(meetings): send reconnect failure before rejoin

adamweeks:
This PR updates the reconnect flow and metrics to the updated design document: https://confluence-eng-gpk2.cisco.com/conf/display/WBXT/Reconnect+Logic

The main changes in this is that when a meeting it detected it needs a "rejoin", it will send the client aborted reconnect metric. This allows us to properly measure reconnect successes and failures. Due to the way it was previously setup, the reconnect failure event would not send during a rejoin.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-153797

---


packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -281,32 +313,40 @@ export default class ReconnectionManager {
...
-          LoggerProxy.logger.info('ReconnectionManager:index#executeReconnection --> Unable to rejoin meeting, reconnecting.', joinError);
-          throw (new NeedsRetryError(joinError));
-        }
-        else {
-          LoggerProxy.logger.error('ReconnectionManager:index#executeReconnection --> Unable to rejoin meeting after max attempts.', joinError);
-          this.status = RECONNECTION.STATE.FAILURE;
-          throw joinError;
-        }
-      }
+      throw (new NeedsRejoinError({wasSharing}));
```
arun3528, line 318: nice so it would throw an error and rejoin but it wont be the same as reconnect flow

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -315,33 +355,44 @@ export default class ReconnectionManager {
...
-      LoggerProxy.logger.error('ReconnectionManager:index#executeReconnection --> Media reestablishment failed');
-      this.status = RECONNECTION.STATE.FAILURE;
+    catch (joinError) {
+      this.rejoinAttempts += 1;
+      if (this.rejoinAttempts <= this.maxRejoinAttempts) {
+        LoggerProxy.logger.info(`ReconnectionManager:index#rejoinMeeting --> Unable to rejoin meeting, attempt #${this.rejoinAttempts}, retrying.`, joinError);
+        this.rejoinMeeting();
+      }
+      else {
+        LoggerProxy.logger.error('ReconnectionManager:index#rejoinMeeting --> Unable to rejoin meeting after max attempts.', joinError);
```
arun3528, line 374: I am not sure if we have to mark the rejoin failure as a reconnect failure , As its a totally different flow

arun3528 (commented):
Need to check on the reconnect failure case

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
adamweeks: Correct, as soon as we know we have to rejoin, our "reconnect" has failed and we go into the rejoin flow.

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
adamweeks: It is only a reconnect failure because it is part of the reconnection manager `reconnect` function. This doesn't send reconnect failure metrics, but to the end user, the reconnect has failed.

adamweeks:
> Need to check on the reconnect failure case

Not sure what you mean @arun3528

arun3528:
> Need to check on the reconnect failure case

Looks good was just confused about `this.status = RECONNECTION.STATE.FAILURE;`  should be good now

abailey-dev (approved):
LGTM

arun3528 (approved):

PR #1925, opened 2020-11-09
feat(plugin-meetings): gdm reachability uses orpheus discovery api

abailey-dev:
In order to provide a better user experience for global webex users, the meetings plugin should utilize the Orpheus discovery API to determine the closest media server for the user. This check happens during meetings.register. The results is stored in localStorage locally and used when sending a meeting.join request.

Fixes #SPARK-165138

---


packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -101,12 +104,13 @@ export default class Reachability extends StatelessWebexPlugin {
             resolve({});
           }
 
-          this.peerConnectionMap = localSDPData.peerConnection;
-          ({localSDPList} = localSDPData);
-          this.setRemoteOfferForEachCluster(localSDPList)
-            .then((response) => {
-              resolve(response);
-            });
+          // TODO: Arun why did you comment this out?
```
arun3528, line 107: we currently dont need to set remote SDK so we can remove this , We are just going to track time getting the first ice

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -162,7 +166,13 @@ export default class Reachability extends StatelessWebexPlugin {
                   this.cleanUpPeerConnections(this.peerConnectionMap);
                   resolve({});
                 });
-              peerConnectionManager.setRemoteSessionDetails(pc, ANSWER, jsonObj.sdp);
+
+              pc.setRemoteDescription(
```
arun3528, line 170: I though the server does not send a remote SDK

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -182,6 +192,76 @@ export default class Reachability extends StatelessWebexPlugin {
...
+        return Number.isNaN(percision) ? elapsed : parseFloat(elapsed.toFixed(percision));
+      };
+
+      let timeout = setTimeout(() => {
+        LoggerProxy.logger.error(`Reachability:index#iceGatheringState --> No ice candidate for ${peerConnection.clusterId}.`);
+        resolve();
+      }, waitTime);
+
+      // eslint-disable-next-line no-param-reassign
+      peerConnection.addEventListener('icegatheringstatechange', () => {
```
arun3528, line 222: In short here we are using either the icegatheringState change or onIcecandidate to calculate the time

arun3528 (changes requested):
Need to check on the remoteSDP , need to remove it , And can we document the response as a wiki or in the code , Someone else having a look wont be able to understand , any links will be helpful 

Also the result which gets stored in the localStorage

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -182,6 +181,58 @@ export default class Reachability extends StatelessWebexPlugin {
...
+
+          LoggerProxy.logger.info(`Reachability:index#onIceGatheringStateChange --> Successfully pinged ${peerConnection.key}:`, elapsed);
+          setLatencyAndClose(elapsed);
+          resolve({clusterId: peerConnection.key, elapsed});
+        }
+      };
+
+      peerConnection.onicecandidate = (e) => {
+        if (e.candidate) {
+          const elapsed = getElapsedTime(peerConnection);
```
adamweeks, line 226: `getElapsedTime` doesn't accept any params, does the function need to be updated?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -5,6 +5,7 @@
 import '@webex/internal-plugin-mercury';
 import {forEach} from 'lodash';
 import {WebexPlugin} from '@webex/webex-core';
+import 'webrtc-adapter';
```
adamweeks, line 8: What is the purpose of this import?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -379,6 +380,10 @@ export default class Meetings extends WebexPlugin {
         StaticConfig.set(this.config);
         LoggerConfig.set(this.config.logging);
         LoggerProxy.set(this.webex.logger);
+
+        // SDK need to perform reachability check on app load
+        this.startReachability();
```
adamweeks, line 385: Should we wait to trigger the meetings ready until reachability is completed? @arun3528?

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -49,30 +53,32 @@ export default class Reachability extends StatelessWebexPlugin {
...
-    // Reset reachability result to empty to ensure no stale data
+
+    // Remove stored reachability results to ensure no stale data
+    if (window.localStorage && window.localStorage.removeItem) {
+      window.localStorage.removeItem(REACHABILITY.localStorage);
+    }
+    else {
+      LoggerProxy.logger.error('Reachability:index#gatherReachability --> Error in accessing LocalStorage.');
+
+      return {};
```
adamweeks, line 66: Should we throw the error here instead of returning an empty object?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
abailey-dev: This is required in order for reachability to work on Safari. WebRTC adapter fixes and creates a base line for all browsers. This includes it within webex.js file. It's listed here because it's related to meetings-plugin.

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
abailey-dev: I believe this error is just for debugging purposes later. I think the flow is to just return empty latency results but not prevent or stop the flow from happening. Correct me if I'm wrong @arun3528.

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
abailey-dev: The function is using closure to a reference the **peerConnection**, I could pass it but it would be redundant I feel. I tried to break the code into smaller functions and reduce that long code block that was within the **forEach** loop. I choose names for the functions that self document what's going on. 

This whole function ** iceGatheringState** applies to only one peerConnection within the closure.

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -49,30 +53,32 @@ export default class Reachability extends StatelessWebexPlugin {
    * @public
    * @memberof Reachability
    */
-  gatherReachability = () => {
+  gatherReachability = async () => {
```
adamweeks, line 56: Could you add `@async` to the jsdoc?

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
adamweeks: Ok, that works, should the code not pass in the parameter then like:

```suggestion
          const elapsed = getElapsedTime();
```

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
abailey-dev: Will do thanks!

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
abailey-dev: Thank you for catching that Adam! I'll fix it. I was originally passing it until I decided it was redundant since I can just use the benefits of the closure.

lalli-flores:
@abailey-dev, can you update the PR's description with context and Jira # of this work? Thanks

packages/node_modules/@webex/plugin-meetings/src/webex.js
```diff
@@ -1,36 +0,0 @@
-import '@webex/plugin-authorization';
```
adamweeks, line 1: What are these files used for? Why are we removing them in this PR?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
arun3528: logically it should finish before addMEdia , I am assuming the join meetingInfo should take more then 5 -6 min which is sufficient for the getting reachability data

packages/node_modules/@webex/plugin-meetings/src/webex.js
abailey-dev: The two files below was not being used anywhere and is dead code. This was discovered after troubleshooting why webrtc-adapter was not loading with @arun3528 and @taymoork2. Removing to avoid confusion, because I initially thought it was being used.

@webex/plugin-meetings/src/webex.js
@webex/plugin-meetings/webex.js

arun3528: This file looks like not being used anywhere

arun3528 (approved):

arun3528:
Only samples test were failing on login , Did manually test space meeting and webex meeting

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -30,300 +34,405 @@ export default class Reachability extends StatelessWebexPlugin {
...
+    const resultKey = elapsed === null ? UNREACHABLE : REACHABLE;
+    const intialState = {[REACHABLE]: 0, [UNREACHABLE]: 0};
+
+    this.clusterLatencyResults[key] = this.clusterLatencyResults[key] || intialState;
+    this.clusterLatencyResults[key][resultKey] += 1;
+
+    // Set to null in case this fired from
+    // an event other than onIceCandidate
+    peerConnection.onicecandidate = null;
+    peerConnection.close();
```
akoushke, line 413: Screen Shot 2020-11-17 at 5 25 02 PM

guessing it comes from here.

abailey-dev: @akoushke - Thank you for sharing this, I will update the code. Mind sharing the browser/version?

PR #1926, opened 2020-11-12
Upgrade Webdriver to v6 and switch to latest version of Node LTS

taymoork2:
Fixes not being on latest version of LTS

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-192230
---


packages/node_modules/@webex/http-core/src/lib/detect.js
```diff
@@ -20,10 +20,10 @@ export {detect};
 
 /**
  * Synchronous implementation of {@link detect}
- * @param {Buffer} buffer
- * @returns {string}
+ * @param {Buffer|Uint8Array} buffer
+ * @returns {Promise<string>}
  */
-export function detectSync(buffer) {
+export async function detectSync(buffer) {
```
adamweeks, line 26: Might need to change the function name since this is no longer synchronous

package.json
```diff
@@ -9,8 +9,8 @@
   "license": "MIT",
   "author": "devsupport@webex.com",
   "engines": {
-    "node": "^10.x",
-    "npm": "^6.x"
+    "node": "^14.x",
+    "npm": "6.x"
```
taymoork2, line 13: do we still need engines now that we can use current LTS?

adamweeks: > Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

It is helpful to have so let's keep them in.

adamweeks:
You can rebase this now that mmmagic is in #1934

package.json
```diff
@@ -134,7 +143,7 @@
     "express-session": "^1.16.2",
     "express-validator": "^5.3.1",
     "file-type": "^16.0.1",
-    "firefox-profile": "^1.2.0",
+    "firefox-profile": "^4.0.0",
```
taymoork2, line 146: This is only used in `openh264.js` and at this point i'd rather get rid of it all of the openH264 tooling

karma-ng.conf.js
```diff
@@ -153,6 +153,7 @@ function makeConfig(packageName, argv) {
       public: 'team',
       startConnect: true,
       connectOptions: {
+        logfile: './sauce.log',
```
taymoork2, line 156: It seems like the [sauce-connect-launcher](https://github.com/bermi/sauce-connect-launcher) doesn't do a `mkdir -p` if `reports/` doesn't exist so it just causes testing to hang forever ^This was like a two day issue for me -_-

mwegman (approved):

PR #1927, opened 2020-11-12
fix(plugin-meetings): trigger event based on H264 presence

abailey-dev:
Developers need an event to know that meetings won't work due to H264 missing in the browser, as well as an event once H264 support is ready (after the browser downloads the codec).

Will check for the presence of H264 and triggers either '**media:codec:missing**' event or the '**media:codec:loaded**' event to inform developers. If the codec is missing it will periodically check if the codec got downloaded up to a maximum timeout. If the codec gets downloaded before the timeout it triggers the '**media:codec:loaded**' event.

Fixes #SPARK-187493

---


packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
```diff
@@ -72,4 +76,84 @@ MeetingsUtil.checkForCorrelationId = (deviceUrl, locus) => {
...
+ * @param {String} codec Media codec.
+ * @returns {Boolean}
+ */
+MeetingsUtil.hasMediaCodec = (codec) => {
+  let hasCodec = false;
+  const video = window.document.createElement('video');
+  const PROBABLY = 'probably';
+
+  if (window.MediaSource) {
+    hasCodec = window.MediaSource.isTypeSupported(codec);
```
abailey-dev, line 91: `window.MediaSource.isTypeSupported(codec)` will return `true` or `false` if the codec is supported.

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
```diff
@@ -72,4 +76,84 @@ MeetingsUtil.checkForCorrelationId = (deviceUrl, locus) => {
...
+  }
+  else {
+    hasCodec = video.canPlayType(codec) === PROBABLY;
```
abailey-dev, line 94: Modenizer does this exact check to determine media capabilities of a browser. `canPlayType` will return only three values 'probably', 'maybe', or ''. If the return is not 'probably' then it doesn't support the codec. We use this as a fallback, preferring `isTypeSupported` first.

lak-16: I have a Firefox browser where it fails connecting to call due to H264 issue.  I see same values on browser which supports H264 and on browser which doesn't support H26 too.

window.MediaSource.isTypeSupported('video/mp4; codecs="avc1.42E01E"') - returns true 
video.canPlayType('video/mp4; codecs="avc1.42E01E"') - returns "probably".

abailey-dev: That means the browser you think doesn't support H264 actually does, it supports H.264 Baseline. There's Baseline, Main and High. Let's see which one it doesn't support.

Run this for me, this is for H.264 High:
`window.MediaSource.isTypeSupported('video/mp4; codecs="avc1.64001E"');`

then run this:
`new RTCPeerConnection().createOffer({offerToReceiveVideo: true}).then( offer => console.log(offer.sdp) )`

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
```diff
@@ -27,6 +29,7 @@ MeetingsUtil.extractDestination = (destination, type) => {
 
 // ** This is a workaround to fix Firefox and google android not downloading H264 codec on start
 MeetingsUtil.triggerH264Codec = async () => {
```
adamweeks, line 31: Is this function needed any more? Can we replace the call to it with a call to `checkMediaPlayback`?

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
```diff
@@ -72,4 +75,87 @@ MeetingsUtil.checkForCorrelationId = (deviceUrl, locus) => {
...
+
+/**
+ * Notifies the user whether or not the codec
+ * is present. Will continuously check until
+ * max duration.
+ * @async
+ * @param {Number} firstChecked Timestamp in milliseconds.
+ * @returns {undefined}
+ */
+MeetingsUtil.checkMediaPlayback = async (firstChecked) => {
```
adamweeks, line 112: This function name is a bit ambiguous to me. Let's just call it what it is:
```suggestion
MeetingsUtil.checkH264Support = async (firstChecked) => {
```

adamweeks:
Can you update the README with the new events?

Also, we should remove [this section](https://github.com/webex/webex-js-sdk/blob/530bcd182e9da588096df454c72a38af9786f594/packages/node_modules/@webex/plugin-meetings/internal-README.md#L7) from the internal readme.

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
abailey-dev: Sounds good to me.

abailey-dev: Albeit, there is a function with the same name inside PeerConnectionManager.

packages/node_modules/@webex/plugin-meetings/src/meetings/util.js
abailey-dev: Probably not, just changed it. Thanks!

arun3528:
@adamweeks @anthonywebb i am little scared that this will be a over burden for the developers to check for codec loaded event,
it would be better to put that as part of ready if its loaded.

If its not then send an event for missing codec and loaded , what do you think ?

arun3528:
PR looks good tho

adamweeks:
@abailey-dev after yesterday's meeting, let's go ahead and make this a real PR and get these changes in. 

Also, please update [this section](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/plugin-meetings/README.md#meetings-events) with the new h264 events.

adamweeks (approved):
LGTM!

PR #1934, opened 2020-11-17
Remove mmmagic dep for file-type + update test-helpers

taymoork2:
Fixes libmagic dependency

---


taymoork2:
Let me know if this commit needs to be split out

mwegman (approved):

lalli-flores (changes requested):
Please place the `package.json` changes in a separate `fix` commit from the refactoring

taymoork2:
not ready to be merged yet, testing on the web-client shows it doesnt work

taymoork2:
After some triaging, it seems that the current implementation with mmmagic also returns null/undefined when looking up file-type information
So the fix is to not destructure and just check if the return value isn't undefined

mwegman (approved):

PR #1937, opened 2020-11-17
Reenable 1:1 meeting integration tests

taymoork2:
Fixes [SPARK_185546](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-185546)

---


packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -444,12 +442,13 @@ skipInNode(describe)('plugin-meetings', () => {
...
         testUtils.waitForEvents([
           {scope: bob.meeting, event: 'meeting:startedSharingLocal'}
-        ]),
+        ], 10000),
         testUtils.waitForEvents([
           {scope: alice.meeting, event: 'meeting:startedSharingRemote'}
         ]),
         testUtils.waitForEvents([{scope: alice.meeting.members, event: 'members:update'}])
           .then((response) => {
+            this.retries(3);
```
adamweeks, line 451: Should we add the timeout here instead of retries? What happens when the event comes in while the retry is retrying.

tooling/lib/test/karma.js
```diff
@@ -20,10 +19,6 @@ exports.test = async function test(options, packageName, files) {
 
   const cfg = makeConfig(packageName, options);
 
-  if (packageName === '@webex/plugin-phone' || packageName === '@webex/media-engine-webrtc') {
```
adamweeks, line 23: If we changed this to the meetings plugin, would we get the injected profile on firefox?

taymoork2: I fixed `openh264.js` (not .sh) and got it working, but it's still broken
It only injects mac so extra effort needs to be done to get it working for injecting into windows as well

adamweeks (approved):
This looks good! Thanks for diving in. Can we move this out of draft status?

tooling/lib/test/karma.js
arun3528: once we have anthony fix for H264 we should be fine

arun3528 (approved):

tooling/lib/test/karma.js
taymoork2: The openH264 code is to inject the codec into a profile
This code has been broken for a while now so I'd rather just get rid of it instead of fixing it because it's needs to be refactored to also generate profiles for Windows

PR #1950, opened 2020-11-23
feat(decryption-transform): key rotation edit message error fix

dalipkumar703:
Key rotation edit message space was not working properly. No one was able to load space due to decryption failing for images in that space. it will fix it.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-191140
---


packages/node_modules/@webex/internal-plugin-conversation/src/decryption-transforms.js
```diff
@@ -42,6 +42,10 @@ export const transforms = toArray('inbound', {
     if (object.objectType === 'transcodedContent') {
       return Promise.all(object.files.items.map((item) => ctx.transform('decryptFile', key, item)));
     }
+    // use default encryption keyUrl for images instead of activity encryptionKeyUrl
```
holsted, line 45: Is this comment correct? I thought `key` here was the default key and the one on the object is the specific one to the activity?

dalipkumar703: Oh yeah, I would update comment.  I should comment something ``using object encryption keyUrl for images instead of activity encryptionKeyUrl``

mwegman:
I've been unable to get the tests to pass by rerunning them. I ran the same tests on my machine against master and they pass. TAP tests are also passing. Can you have a look at the integration tests?

dalipkumar703:
> I've been unable to get the tests to pass by rerunning them. I ran the same tests on my machine against master and they pass. TAP tests are also passing. Can you have a look at the integration tests?

yeah I checked too it is always failing at ``updatetypingstatus`` not sure reason for failing. Need to dig deeper to find reason for it.

PR #1960, opened 2020-12-08
Support for tags in conversation create

glenatwork:
Conversation create payload can take a "tags' field.  Essentially a pass-through of the field from the client request is needed, like here: https://github.com/webex/webex-js-sdk/blob/0bec4c49824ab4510d868739f690d6f1cb09f500/packages/node_modules/%40webex/internal-plugin-conversation/src/conversation.js#L2293

[It's not clear to me where I open an issue.  When going to Issues, it offers links to other locations?]

---


adamweeks (changes requested):
Thanks @glenatwork for the PR! Could you update the jsdoc on the `create` function [here](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js#L162) to add that `tags` is available in the `params` object? Also, please add a test near [here](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/internal-plugin-conversation/test/integration/spec/create.js#L187) for creating a convo with tags.

adamweeks (approved):
Looks good, thanks for making those changes! I'm investigating the test failures.

PR #1962, opened 2020-12-08
Add ability to generate test users using Whistler API

taymoork2:
Fixes [SPARK-193921](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-194931)

---

### DIFF
--- a/packages/node_modules/@webex/test-users/src/whistler.js
+++ b/packages/node_modules/@webex/test-users/src/whistler.js
@@ -1,0 +1,155 @@
+import btoa from 'btoa';
+import {request} from '@webex/http-core';
+
+/**
+ * Fetches credentials/access_token to talk to the whistler endpoint
+ *
+ * @param {Object} options
+ * @param {string} options.clientId
+ * @param {string} options.clientSecret
+ * @param {string} options.idbrokerUrl
+ * @param {string} options.orgId
+ * @param {string} options.machineAccount
+ * @param {string} options.machinePassword
+ * @private
+ * @returns {Promise<string>}
+ */
+const getClientCredentials = ({
+  clientId,
+  clientSecret,
+  orgId,
+  idbrokerUrl,
+  machineAccount,
+  machinePassword
+}) => request({
+  method: 'POST',
+  uri: `${idbrokerUrl}/idb/token/${orgId}/v2/actions/GetBearerToken/invoke`,
+  json: true,
+  body: {
+    uid: machineAccount,
+    password: machinePassword
+  }
+})
+  .then((res) => request({
+    method: 'POST',
+    uri: `${idbrokerUrl}/idb/oauth2/v1/access_token`,
+    json: true,
+    form: {
+      assertion: res.body.BearerToken,
+      grant_type: 'urn:ietf:params:oauth:grant-type:saml2-bearer',
+      scope: 'webexsquare:get_conversation webexsquare:admin',
+      self_contained_token: true,
+      client_id: clientId,
+      client_secret: clientSecret
+    },
+    headers: {
+      // Note: we can't request's auth hash here because this endpoint expects
+      // us to send the auth header *without including "Basic "* before the
+      // token string
+      // authorization: `Basic + ${btoa(`${clientId}:${clientSecret}`)}`
+      authorization: btoa(`${clientId}:${clientSecret}`)
+    }
+  }))
+  .then((res) => `${res.body.token_type} ${res.body.access_token}`);
+
+/**
+  * @typedef {Object} TestUserObject
+  * @property {string} password
+  * @property {string} emailAddress
+  * @property {string} displayName
+  * @property {string} token
+  * @property {string} reservationUrl
+  * @property {object} responseMetaData - whistler given properties
+  */
+
+/**
+ * @typedef {Object} CreateUserOptions
+ * @param {Object} [options]
+ * @param {string} [whistlerServiceUrl] defaults to WHISTLER_API_SERVICE_URL
+ * @param {string} [options.clientId] defaults to WEBEX_CLIENT_ID
+ * @param {string} [options.clientSecret] defaults to WEBEX_CLIENT_SECRET
+ * @param {string} [options.idbrokerUrl] defaults to IDBROKER_BASE_URL
+ * @param {string} [options.orgId] organization ID to create the user under
+ * @param {string} [options.machineAccount] defaults to WHISTLER_MACHINE_ACCOUNT
+ * @param {string} [options.machinePassword] defaults to WHISTLER_MACHINE_PASSWORD
+ */
+
+/**
+ * Creates a test user
+ * @param {CreateUserOptions} options
+ * @returns {Promise.<TestUserObject>}
+ */
+export default function createTestUser(options = {}) {
+  const clientId = options.clientId || process.env.WEBEX_CLIENT_ID;
+  const clientSecret = options.clientSecret || process.env.WEBEX_CLIENT_SECRET;
+  const machineAccount = options.machineAccount || process.env.WHISTLER_MACHINE_ACCOUNT;
+  const machinePassword = options.machinePassword || process.env.WHISTLER_MACHINE_PASSWORD;
+  const idbrokerUrl = options.idbrokerUrl || process.env.IDBROKER_BASE_URL;
+  const orgId = options.orgId || process.env.WHISTLER_TEST_ORG_ID;
+  const whistlerServiceUrl = options.whistlerServiceUrl || process.env.WHISTLER_API_SERVICE_URL;
+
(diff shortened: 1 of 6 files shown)


packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,197 @@
...
+      emailAddress: res.body.responseMetaData.name,
+      displayName: res.body.responseMetaData.webExUserName,
+      token: res.body.responseMetaData.ciAccessToken,
+      reservationUrl: res.body.reservationUrl
+    }, res.body.responseMetaData));
+}
+
+/**
+ *
+ * @param {Object} options
```
lalli-flores, line 186: Can you document expected options?

packages/node_modules/@webex/test-users/test/integration/spec/whistler.js
```diff
@@ -0,0 +1,70 @@
...
+
+    it('removes a test user if no access token is available', () => createWhistlerTestUser()
+      .then(prune)
+      .then(async (u) => {
+        const res = await removeWhistlerTestUser(u);
+
+        assert.equal(res.statusCode, 204);
+      }));
+
+    it('removes a test user with an access token set to undefined', () => createWhistlerTestUser()
```
lalli-flores, line 58: If there is no access token, how do you know what user to remove? or what gets removed?

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  if (token.token_type && token.access_token) {
+    token.authorization = `${token.token_type} ${token.access_token}`;
+  }
+
+  return token;
+}
+
+let clientToken;
```
lalli-flores, line 30: Why would you want to cache the user? Isn't the expectation that I will get a new user every time I call "create user"?

packages/node_modules/@webex/test-users/src/whistler.js
taymoork2: It's the entire user object so there's too much to document, also options isnt the best name, but i'm keeping it aligned with what's in `index.js`

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
+import btoa from 'btoa';
+import {request} from '@webex/http-core';
+
+/**
+ * Fetches credentials to talk to the test_users_s endpoint
+ *
+ * Caches result in `clientToken` variable for multiple runs
```
lalli-flores, line 7: I think this is no longer in the code

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
+import btoa from 'btoa';
+import {request} from '@webex/http-core';
+
+/**
+ * Fetches credentials to talk to the test_users_s endpoint
```
lalli-flores, line 5:
```suggestion
 * Fetches credentials to talk to the test users's endpoint
```

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+      // us to send the auth header *without including "Basic "* before the
+      // token string
+      // authorization: `Basic + ${btoa(`${clientId}:${clientSecret}`)}`
+      authorization: btoa(`${clientId}:${clientSecret}`)
+    }
+  }))
+  .then((res) => `${res.body.token_type} ${res.body.access_token}`);
+
+/**
+ * @typedef {Object} AccessTokenObject
```
lalli-flores, line 57: I dont think this object exists anymore

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+ * @property {Number} token.expires_in
+ * @property {string} token.token_type
+ * @property {string} token.refresh_token
+ * @property {Number} token.refresh_token_expires_in
+ * @property {string} token.expires
+ * @property {string} token.refresh_token_expires
+ */
+
+/**
+  * @typedef {Object} TestUserObject
```
lalli-flores, line 68: I think there are a lot more properties now, can you update this to match?

packages/node_modules/@webex/test-users/src/whistler.js
lalli-flores: There is a type definition above for `TestUserObject`, if that is what it's supposed to be then, can you update the correct type?
Maybe rename the argument to `user`

packages/node_modules/@webex/test-users/test/integration/spec/whistler.js
```diff
@@ -0,0 +1,39 @@
...
+
+assert.isTestUser = (user) => {
+  assert.isDefined(user, 'user is defined');
+  assert.isDefined(user.displayName, 'user.displayName is defined');
+  assert.isDefined(user.emailAddress, 'user.emailAddress is defined');
+  assert.isDefined(user.password, 'user.password is defined');
+  assert.isDefined(user.reservationUrl, 'user.reservationUrl is defined');
+};
+
+describe('test-users', () => {
```
lalli-flores, line 22: Can you add tests to show what happens when there's a failure?

packages/node_modules/@webex/test-users/src/whistler.js
adamweeks: This whistler file doesn't talk to the `test_users_s` convo endpoint.

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Creates a test user
+ * @param {CreateUserOptions} options
+ * @returns {Promise.<TestUserObject>}
+ */
+export default function createTestUser(options = {}) {
+  const clientId = options.clientId || process.env.WEBEX_CLIENT_ID;
+  const clientSecret = options.clientSecret || process.env.WEBEX_CLIENT_SECRET;
+  const machineAccount = options.machineAccount || process.env.WHISTLER_MACHINE_ACCOUNT;
```
adamweeks, line 94: This is not in the typedef

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+  const machinePassword = options.idbrokerUrl || process.env.WHISTLER_MACHINE_PASSWORD;
```
adamweeks, line 95:
```suggestion
  const machinePassword = options.machinePassword || process.env.WHISTLER_MACHINE_PASSWORD;
```

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+  * @property {string} emailAddress
+  * @property {string} displayName
+  */
+
+/**
+ * @typedef {Object} CreateUserOptions
+ * @param {Object} [options]
+ * @param {string} [options.clientId] defaults to WEBEX_CLIENT_ID
+ * @param {string} [options.clientSecret] defaults to WEBEX_CLIENT_SECRET
+ * @param {string} [options.conversationServiceUrl] defaults to WEBEX_TEST_USERS_CONVERSATION_SERVICE_URL
```
adamweeks, line 79: Don't need this

adamweeks: Perhaps it should be `WHISTLER_API_SERVICE_URL`

packages/node_modules/@webex/test-users/src/whistler.js
```diff
@@ -0,0 +1,159 @@
...
+    clientId,
+    clientSecret,
+    machineAccount,
+    machinePassword,
+    idbrokerUrl,
+    orgId
+  })
+    .then((authorization) => request({
+      method: 'GET',
+      uri: `${process.env.WHISTLER_API_SERVICE_URL}/reservations/testUser?isAccessTokenRequired=true`,
```
adamweeks, line 132: Can we make this an option like we had with the `options.conversationServiceUrl`?

taymoork2:
@adamweeks writing tests to check if `options.whistler` is working, is becoming more difficult than it should be.
Sinon is unable to stub/spy on methods within the `@webex/test-users` package that is a dependency in `@webex/test-helpers-test-users`, where the test is being created/added. It especially doesn't like trying to stub the exported whistler create user function.
Michael and I tried figuring out a solution, but were both stumped; and I showed Michael the integration tests for whistler were passing.
I don't think we should spend more time on this PR and get to work on getting the other repo setup

adamweeks (approved):

mwegman (approved):

PR #1965, opened 2020-12-10
fix(plugin-meetings): use fetch info for destination

adamweeks:
When using a conversation id from hydra, we were ignoring the work the meeting info utility was doing to decode and detect the destination type. 

---


taymoork2 (approved):
LGTM

PR #1971, opened 2020-12-12
fix(server-interceptor): add additional error codes for failover

holsted:
Adds 502 and 503 support to failover checks

---

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/lib/services/interceptors/server-error.js
+++ b/packages/node_modules/@webex/webex-core/src/lib/services/interceptors/server-error.js
@@ -24,7 +24,7 @@ export default class ServerErrorInterceptor extends Interceptor {
    * @returns {Object}
    */
   onResponseError(options, reason) {
-    if ((reason instanceof WebexHttpError.InternalServerError) && options.uri) {
+    if ((reason instanceof WebexHttpError.InternalServerError || reason instanceof WebexHttpError.BadGateway || reason instanceof WebexHttpError.ServiceUnavailable) && options.uri) {
       const feature = this.webex.internal.device.features.developer.get('web-high-availability');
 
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/webex-core/src/lib/services/interceptors/server-error.js
```diff
@@ -24,7 +24,7 @@ export default class ServerErrorInterceptor extends Interceptor {
    * @returns {Object}
    */
   onResponseError(options, reason) {
-    if ((reason instanceof WebexHttpError.InternalServerError) && options.uri) {
+    if ((reason instanceof WebexHttpError.InternalServerError || reason instanceof WebexHttpError.BadGateway || reason instanceof WebexHttpError.reason instanceof WebexHttpError.ServiceUnavailable) && options.uri) {
```
abailey-dev, line 27: I think you are missing a logical OR here.  Otherwise it's `<boolean> instanceof WebexHttpError.ServiceUnavailable` which will always return `false`.

packages/node_modules/@webex/webex-core/test/unit/spec/services/interceptors/server-error.js
```diff
@@ -119,6 +119,26 @@ describe('webex-core', () => {
               });
           });
 
+          it('should mark a url as failed for a 503', (done) => {
+            reason = new WebexHttpError.ServiceUnavailable({
+              message: 'test message',
+              statusCode: 500,
```
abailey-dev, line 125: Should these code match? You have 503 and 500.

abailey-dev (changes requested):

abailey-dev (approved):
LGTM!

adamweeks (approved):

PR #1973, opened 2020-12-15
test(i-p-conversation): remove mercury logic from typing tests

adamweeks:
No JIRA, lots of false test failures on typing.

If convo 200’s we can assume it does what it needs to. We don’t need to test that the mercury event eventually arrives.


taymoork2 (approved):
FISI

PR #1980, opened 2021-01-05
feat(plugin-meetings): use new correlation on rejoin

adamweeks:
Fixes [SPARK-186101](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-186101)

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -114,6 +114,7 @@ const logRequest = (request, {header = '', success = '', failure = ''}) => {
  * @property {String|Object} [meetingQuality]
  * @property {String} [meetingQuality.local]
  * @property {String} [meetingQuality.remote]
+ * @property {Boolean} [rejoin]
  */
 
 /**
@@ -2333,7 +2334,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   reconnect(options) {
-    LoggerProxy.logger.log('Meeting:index#reconnect --> attempting to reconnect');
+    LoggerProxy.logger.log(`Meeting:index#reconnect --> attempting to reconnect meeting ${this.id}`);
 
     if (!this.reconnectionManager || !this.reconnectionManager.reconnect) {
       throw new ParameterError('Cannot reconnect, ReconnectionManager must first be defined.');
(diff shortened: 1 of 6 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2418,28 +2419,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
-        event: eventType.CALL_INITIATED,
-        meeting: this,
-        data: {pstnAudioType: 'DIAL_IN'}
-      });
+    if (options.rejoin) {
+      LoggerProxy.logger.log(`Meeting:index#join --> Generating a new correlation id for meeting ${this.id}`);
+      LoggerProxy.logger.log(`Meeting:index#join --> Previous correlation id ${this.correlationId}`);
+      this.setCorrelationId(uuid.v4());
+      LoggerProxy.logger.log(`Meeting:index#join --> New correlation id ${this.correlationId}`);
+      this.meetingFiniteStateMachine.reset();
```
arun3528, line 2427: Does this line changes the statemachine where it maintains IDLE -RINGING -JOINED stats !?

packages/node_modules/@webex/plugin-meetings/src/roap/index.js
```diff
@@ -242,7 +243,7 @@ export default class Roap extends StatelessWebexPlugin {
         roapMessage,
         correlationId: meeting.correlationId,
         locusSelfUrl: meeting.selfUrl,
-        mediaId: reconnect ? '' : meeting.mediaId,
+        mediaId: reconnect ? uuid.v4() : meeting.mediaId,
```
arun3528, line 246: I think this should only be in the sendRoapOffer,  When we get a roap answer its assumed there is a media ID already present !?

arun3528 (approved):
One comment on the roap answer function rest looks good,

We will need to do some testing for various scenarios , 1:1 , space , CMR meeting which host joins first and goes to inactive state or 

A guest user joining and going into a rejoin state , We might have to think about it when a guest does to inactive state we should logically kick the user because if the guest goes into a weird state it will keep asking host to admit 

We can put a limit for rejoin that also works

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
adamweeks: Yes, since we haven't joined the meeting yet with a new correlation id, it will reset the state machine back to default for the meeting.

packages/node_modules/@webex/plugin-meetings/src/roap/index.js
adamweeks: I added this because Jeff S requested it:

> Instead of null media id on the rejoin, send a new media id ( that the client generates ), like you do for the initial create media.

PR #1983, opened 2021-01-11
handle peer connection stability on meeting join

taymoork2:
Fixes [SPARK-194866](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-194866)

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -26,35 +26,37 @@ import StatsUtil from '../stats/util';
 import ReconnectionError from '../common/errors/reconnection';
 import ReconnectInProgress from '../common/errors/reconnection-in-progress';
 import {
-  MEETINGS,
+  _BUSY_,
+  _CALL_,
+  _INCOMING_,
+  _JOIN_,
+  CONTENT,
   ENDED,
   EVENT_TRIGGERS,
   EVENT_TYPES,
   EVENTS,
   FLOOR_ACTION,
-  _BUSY_,
-  _INCOMING_,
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2847,11 +2849,22 @@ export default class Meeting extends StatelessWebexPlugin {
       })
         .then((peerConnection) => {
           LoggerProxy.logger.info(`${LOG_HEADER} PeerConnection Received from attachMedia ${peerConnection}`);
+
+          // Check if Peer Connection is STABLE after 30 seconds
+          // If NOT STABLE then BAIL and end meeting
+          setTimeout(() => {
+            if (peerConnection.iceConnectionState !== ICE_STATE.CONNECTED) {
```
adamweeks, line 2856: It's possible that the ice state could go to connected immediately, but then have an issue putting it not in a connected state at exactly 30 seconds. Should we put logic in for when the peer connection goes to connected to cancel this timeout? @arun3528 thoughts?

adamweeks: Ice connection state change handler is [here](https://github.com/webex/webex-js-sdk/blob/6465ed41306a6a3db97a3cbdca7d081af097f3e0/packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js#L5740). It feels like this could go there.

arun3528: +1 we can bail out once it goes to connected state , pc can go to failed and disconnect state with in 30 sec of connection if the network does down or VPN changes etc

PR #1985, opened 2021-01-12
fix(plugin-meeting): unable to mute/unmute with no camera

dalipkumar703:
Currently when calling from Desktop having no camera in Web Client then unable to mute/unmute media. This PR address fix it.

Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-194341

---


packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -400,6 +400,14 @@ Media.getMedia = (audio, video, config) => {
       false,
     fake: process.env.NODE_ENV === 'test' // Special case to get fake media for Firefox browser for testing
   };
+  const {sendAudio, sendVideo} = await Media.getSupportedDevice({sendAudio: mediaConfig.audio, sendVideo: mediaConfig.video});
```
arun3528, line 403: The parent function already does the check for the supported device , So you don't need this

packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -440,7 +448,7 @@ Media.getSupportedDevice = ({sendAudio, sendVideo}) =>
 
       return {
         sendAudio: (supported.audio && sendAudio) || true,
```
arun3528, line 450: Let change this for audio as well , No audio device returned then make it false

arun3528 (changes requested):

packages/node_modules/@webex/plugin-meetings/src/media/index.js
adamweeks: I think this is the source of the bug. We shouldn't be defaulting `send` to `true`.

```suggestion
        sendAudio: supported.audio && sendAudio,
```

adamweeks (changes requested):
Agree with @arun3528's requests.

packages/node_modules/@webex/plugin-meetings/src/media/index.js
dalipkumar703: I tried removing getSupportedDevice function here but it didn't work. Also, it seem getSupportedDevice not exposed to webclient because i got undefined calling this function in webclient. @arun3528 I am not sure how this function would be called from parent function

dalipkumar703: I exposed getSupportedDevice function so we can make call from web client now we don't need function here anymore.

arun3528: https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js#L2675

arun3528 (approved):
Looks good after adams comments

adamweeks (approved):

pacacc (commented):
I cannot add a comment inline, but on lines 430/431 it is still defaulting to true.
Probably it would make sense to change those too?

dalipkumar703:
> I cannot add a comment inline, but on lines 430/431 it is still defaulting to true.
> Probably it would make sense to change those too?

Yeah, make sense. Updated it too.

PR #1996, opened 2021-02-04
fix(upload): add abort function on file upload for webex

arjunkesava:
Fixes [SPARK](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-203178)

On WebEx, when user uploads a file we should check for file limit and restrict the user to upload files within specified limit or 2048 MB by default

---

### DIFF
--- a/packages/node_modules/@webex/webex-core/src/webex-core.js
+++ b/packages/node_modules/@webex/webex-core/src/webex-core.js
@@ -78,6 +78,8 @@ const postInterceptors = [
   'RateLimitInterceptor'
 ];
 
+const MAX_FILE_SIZE_IN_MB = 2048;
+
 /**
  * @class
  */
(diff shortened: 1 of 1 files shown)


mwegman (changes requested):
Code looks good. Please add a test for it.

arjunkesava:
Thank you @mwegman for review.
This PR is to add an ability to abort the upload process for file.
The upload function in the webcore file doesn’t have any test cases until now.
So, if I write a test case for _uploadAbortSession, then I have to write the test cases for the entire upload function which will take a lot more time than expected.
Files team is waiting on the results of this PR. Could you please approve it.
I will add the test cases in my next PR (there is some more work coming on the way)

mwegman (approved):

arun3528:
@mwegman  should we know when the sessions get aborted as we need to know how many times its happens !? @arjunkesava 
 do we have any metrics around it as there will be some meetings we need to triage and if we dont have log file

adamweeks:
@arun3528 the metrics would have to be on the server side to check the DELETE requests. We are not adding client metrics in the SDK outside of the meetings plugin for now.

arun3528:
If someone is going to track the if deletes are spiking then we might have to re visit on how much we log , good for now

adamweeks:
@arjunkesava could you please squash the commits into one? We utilize commit messages to generate the [changelog](https://github.com/webex/webex-js-sdk/blob/master/CHANGELOG.md)

arjunkesava:
@adamweeks @mwegman 

I had rebased it with latest upstream master and also squashed into a single commit. Please do note.

adamweeks (approved):

adamweeks:
@arjunkesava Thanks for making those changes, approved!

PR #2002, opened 2021-02-10
fix(plugin-meetings): prevent success message when log upload fails

abailey-dev:
Currently `meeting:logUpload:success` fires even when logs fail to upload. This ensures the failure is handled correctly.

Fixes #SPARK-210681

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
@@ -37,6 +37,8 @@ export default class LogsRequest {
     }
     catch (error) {
       LoggerProxy.logger.error('Logs:request#uploadLogs --> uploading user logs failed', error);
+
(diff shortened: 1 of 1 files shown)


adamweeks (approved):
Nice catch and quick fix!

packages/node_modules/@webex/plugin-meetings/src/common/logs/request.js
```diff
@@ -37,6 +37,8 @@ export default class LogsRequest {
     }
     catch (error) {
       LoggerProxy.logger.error('Logs:request#uploadLogs --> uploading user logs failed', error);
+
+      return Promise.reject(error);
```
adamweeks, line 41:
```suggestion
      throw error;
```
Since this is an async function, we don't have to use promises here and just throw the error. Not a blocker on the PR though.

mwegman (approved):

PR #2005, opened 2021-02-10
Enable Whistler and CMR tests on Jenkins only

taymoork2:


adamweeks (changes requested):
Could you update the [env var docs](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#advanced-environment-variables) with these new JENKINS and WHISTLER env vars?

adamweeks (approved):

PR #2011, opened 2021-02-15
docs(samples) : Implemented OAuth Flow into the sharescreen sample app

MathewChanda:
In this PR, I have implemented and added the OAuth option to authorized the user in addition to pasting the access token.  

Jira - https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-218451


packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
...
+
+      if (webex.canAuthorize) {
+        // if already authenticated, logout on click
+        webex.logout();
+      }
+    });
+  }
+
+  if (isPaste) {
+    const legend = document.createElement('legend');
```
adamweeks, line 334: Please move this into the `index.html` file and use
`document.getElementById("myDIV").style.display = "none";` to hide/show it.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
   });
 }
 
-// In order to simplify the state management needed to keep track of our button
-// handlers, we'll rely on the current meeting global object and only hook up event
-// handlers once.
+function handleRadio() {
```
adamweeks, line 265: Let's use a more descriptive name on this:

```suggestion
function handleAuthTypeChange() {
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
...
+  const div = document.createElement('div');
+
+  // removes the child node of the form if anything exists
+  while (document.getElementById('authForm').firstChild) {
+    document.getElementById('authForm').removeChild(document.getElementById('authForm').firstChild);
+  }
+
+  // adds the correct form based on enabled radio
+  if (isOAuth) {
+    const legend = document.createElement('legend');
```
adamweeks, line 279: Please move this into the `index.html` file and use
`document.getElementById("myDIV").style.display = "none";` to hide/show it.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
...
+      el.addEventListener('change', (event) => {
+        localStorage.setItem(id, event.target.value);
+      });
+    });
+
+    // add event handlers for webex.init() button
+    document.getElementById('webexInit').addEventListener('click', (event) => {
+      // let's make sure we don't reload the page when we want to initialize webex
+      event.preventDefault();
+
```
adamweeks, line 381: Add a comment about why we are doing this:

```suggestion
// Re-initializing the sdk to authorize with the pasted access token
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
...
+        webex.logout();
+      }
+    });
+  }
+
+  if (isPaste) {
+    const legend = document.createElement('legend');
+    const accessToken = document.createElement('input');
+    const webexInit = document.createElement('input');
+    const register = document.createElement('input');
```
adamweeks, line 337: Don't forget, the register button needs to be used by both authentication types.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,13 +262,172 @@ function joinMeeting(meeting) {
...
+      // again, we don't want to reload when we try to register
+      event.preventDefault();
+
+      connect().then(() => enabledButton());
+    });
+  }
+}
+
+// Enabled buttons after registering or completing the oauth flow
+function enabledButton() {
```
adamweeks, line 412:
```suggestion
function enableMeetingActionButtons() {
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
MathewChanda: Weird this is an old commit. This isn't right

lalli-flores (changes requested):
Have yet to review the code, but please change your commit type, from `feat` to `docs`. `feat` will trigger a new version, which is not necessary for changes in the samples, which are considered documentation.

adamweeks:
image

Once a user has logged in, we need to disable the access token picker.

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
...
+        <form id="typeCredentials">
+          <fieldset>
+            <legend>Pick Method of Getting Access Token</legend>
+              <div>
+                <input type="radio" id="oauth" name="drone" onchange="handleRadio()" value="oauth" checked>
+                <label for="huey">OAuth</label>
+              </div>
+              <div>
+                <input type="radio" id="paste" name="drone" onchange="handleRadio()" value="paste">
+                <label for="huey">Paste Access Token</label>
```
adamweeks, line 57:
```suggestion
                <input type="radio" id="oauthLogin" name="tokenType" onchange="handleRadio()" value="oauth" checked>
                <label for="oauthLogin">OAuth</label>
              </div>
              <div>
                <input type="radio" id="tokenLogin" name="tokenType" onchange="handleRadio()" value="paste">
                <label for="tokenLogin">Paste Access Token</label>
```

Updating the `label` to match the id will allow the user to click the label to select the radio element.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,50 +10,45 @@
...
-  el.value = localStorage.getItem(id);
-  el.addEventListener('change', (event) => {
-    localStorage.setItem(id, event.target.value);
-  });
+let activeMeeting, remoteShareStream;
+let isPaste;
+let isOAuth;
+
+// eslint-disable-next-line no-multi-assign
+let webex = window.webex = Webex.init({
```
lalli-flores, line 29: Why do you initialize the SDK here before you know which way you're authenticating?

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
       <main>
         <h1>Local Screensharing</h1>
         <p>This sample demonstrates sharing your local screen</p>
+
+        <form id="typeCredentials">
```
lalli-flores, line 48: Based on the naming convention of other forms in the page
```suggestion
        <form id="auth-type">
```

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
...
+              <div>
+                <input type="radio" id="paste" name="drone" onchange="handleRadio()" value="paste">
+                <label for="huey">Paste Access Token</label>
+              </div>
+            </div>
+          </fieldset>
+        </form>
+
         <form id="credentials">
+          <fieldset id="oauthForm" style="display: block;">
```
lalli-flores, line 64:
```suggestion
          <fieldset id="oauth" style="display: block;">
```

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
...
+            <legend>Authentication with OAuth </legend>
+              <input id="login" title="login" type="submit" value="login">
+              <input id="logout" title="logout" type="submit" value="logout" disabled="disabled">
+              <p id="login-status">not logged in</p>
+          </fieldset>
+          <fieldset id="pasteForm" style="display: none;">
```
lalli-flores, line 70:
```suggestion
          <fieldset id="token-auth" style="display: none;">
```

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
       <main>
         <h1>Local Screensharing</h1>
         <p>This sample demonstrates sharing your local screen</p>
+
+        <form id="typeCredentials">
+          <fieldset>
+            <legend>Pick Method of Getting Access Token</legend>
+              <div>
+                <input type="radio" id="oauth" name="drone" onchange="handleRadio()" value="oauth" checked>
```
lalli-flores, line 52: Making it more explicit as to what this represents
```suggestion
                <input type="radio" id="oauth-input" name="drone" onchange="handleRadio()" value="oauth" checked>
```

lalli-flores (commented):
Lots of comments from me 😬

This one is kinda picky, but I'm also a bit OCD soooo can you make sure that your commenting is consistent? In the tense but also the style? e.g. starting all comments with capital or not.

Lastly, make sure that where you place the code changes is intentional. What I mean by that is that this is meant to be an example for other developers on how to develop their own code in the SDK. We want to make sure our code has good flow, so I would make sure that the expected execution flow of the code matches the UI that we present, e.g. access type, authentication, registration, dialing, controls, etc.

packages/node_modules/samples/browser-call-with-screenshare/app.js
MathewChanda: Micheal and I decided to initialize the SDK beforehand because we defaulting to OAuth in the sample app. In addition, we had a few initializing problems when switching between OAuth and paste token method, which lead to this solution. Let me know if you need any more clarity on this and I will respond asap!

adamweeks: It needs to be initialized automatically when doing the OAuth flow, especially around when it returns from OAuth. The SDK watches for the token to come across in the browser url hash. 

We could save a cookie on the auth type and only initialize if the cookie exists that the user was doing oauth mode specifically. @lalli-flores thoughts?

packages/node_modules/samples/browser-call-with-screenshare/index.html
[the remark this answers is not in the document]
MathewChanda: Unsure what's this is exactly. I will investigate it. Most likely I think we can remove it.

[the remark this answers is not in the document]
MathewChanda: I remembered what happened here. The reason why both inputs are named drone is because it was temporary fix for selecting only one radio in the form. I will investigate it more to see a better implementation.

packages/node_modules/samples/browser-call-with-screenshare/index.html
MathewChanda: I decided to call the id as oauthForm because I already named the radio's id as oauth.

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
...
+                <label for="huey">OAuth</label>
```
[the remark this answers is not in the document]
MathewChanda: Oops! I think I forgot to change it.

packages/node_modules/samples/browser-call-with-screenshare/index.html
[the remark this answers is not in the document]
lalli-flores: haha, I know that you copied and paste, I actually know exactly where you got it too 😉
I was trying to make a point that as developers we should know exactly what we're putting out. My philosophy is, If you don't understand it, don't add it

packages/node_modules/samples/browser-call-with-screenshare/app.js
lalli-flores: @adamweeks, I think I like that idea better of storing the auth type in a cookie or local storage

packages/node_modules/samples/browser-call-with-screenshare/index.html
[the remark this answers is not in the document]
MathewChanda: Thanks @lalli-flores for the advice and I will continue to follow that philosophy! Also, I figured out that works from this article! 

https://www.w3schools.com/tags/att_input_type_radio.asp

packages/node_modules/samples/browser-call-with-screenshare/app.js
mwegman: Detect if this is being called from a redirect and initialize the SDK.
``` javascript
const params = new URLSearchParams(window.location.search);
if (params.has('code') && params.has('state')) {
  // Initialize SDK
}
```
Most likely if `state` and `code` are in the query string, this was a redirect from CI.

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,16 +44,46 @@
...
+               id="access-token"
+               name="accessToken"
+               placeholder="Your access token"
+               type="text">
+              <input id="webexInit" title="webex.init()" type="submit" value="webex.init()">
+          </fieldset>
+        </form>
+
+        <form id="dialer">
           <fieldset>
```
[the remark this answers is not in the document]
MathewChanda: In this part of the code, we will be enabling one button but disabled another button. If we disabled the fieldset, we would disable both buttons correct? If so, how would the user have one button enabled for login and disabled for log out and vice versa when we disabled the whole thing.

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -65,18 +95,18 @@ <h1>Local Screensharing</h1>
...
 
         <form id="call-controls">
           <fieldset>
             <legend>Call Controls</legend>
-            <button id="hangup" title="hangup" type="button">cancel/hangup</button>
+            <button id="hangup" title="hangup" type="button" disabled>cancel/hangup</button>
 
-            <button id="share-screen" title="share screen" type="button">share screen</button>
-            <button id="stop-screen-share" title="stop screen share" type="button">stop screen share</button>
+            <button id="share-screen" title="share screen" type="button" disabled>share screen</button>
```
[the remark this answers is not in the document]
MathewChanda: I agree with this one. I will go ahead and work on that as well :)

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
-  'access-token',
-  'invitee'
-].forEach((id) => {
-  const el = document.getElementById(id);
 
-  el.value = localStorage.getItem(id);
-  el.addEventListener('change', (event) => {
-    localStorage.setItem(id, event.target.value);
+// OAuth section
+if (window.location.hash) {
```
adamweeks, line 35: Can you move this to the bottom of the file with the call to `handleAuthTypeChange`. I'm afraid this will be missed by consumers trying to use the sample code because it is nested in the function definitions.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
+  document.getElementById('logout').addEventListener('click', (event) => {
+    // Let's make sure we don't reload the page when we submit the form
+    event.preventDefault();
+
+    // Redirects user to sign out page
+    webex.logout();
+  });
+
+  // Immediately enabled all button in the oauth section when Webex is ready to be used
+  webex.once('ready', () => {
```
adamweeks, line 95: When I load the page, this immediately throws an error saying `webex` is not defined.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
+  document.getElementById('token-auth').style.display = 'none';
+  document.getElementById('oauth').style.display = 'block';
+
+  // Now, we adding event handler to login button to initiate login when clicked
+  document.getElementById('login').addEventListener('click', (event) => {
+    // Let's make sure we don't reload the page when we submit the form
+    event.preventDefault();
+
+    // Initialize the Webex
+    initializeWebexOAuthVersion();
```
adamweeks, line 79: We don't need to wait until login is pressed to create the webex oauth instance. That is why the error on line 95 is throwing.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
-// First, let's wire our form fields up to localStorage so we don't have to
-// retype things everytime we reload the page.
+let redirect_uri;
+let activeMeeting;
+let remoteShareStream;
+let isToken;
+let isOAuth;
+let webex;
+
+function getOAuthRedirectURI() {
```
adamweeks, line 21: Let's remove this function and just have this as part of the redirect_uri definition.

```js
let redirect_uri = `${window.location.protocol}//${window.location.host}`;
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
+if (window.location.hash) {
+  // Hash is created which gives us the access token,
+  // which will be used in the SDK initialization process
+  initializeWebexOAuthVersion();
+}
+
+// Initialize Webex to authorize through OAuth
+function initializeWebexOAuthVersion() {
+  // Get's redirect URI for OAuth option
+  getOAuthRedirectURI();
```
adamweeks, line 44: Instead of calling the function that isn't used elsewhere, let's just put the code block here:

```suggestion
  // This is one way you might programattically determine your redirect_uri
  // depending on where you've deployed your app, but you're probably better off
  // having development/staging/production builds and injecting directly from the
  // environment.
  redirect_uri = `${window.location.protocol}//${window.location.host}`;

  if (window.location.pathname) {
    redirect_uri += window.location.pathname;
  }
```

packages/node_modules/samples/browser-call-with-screenshare/index.html
```diff
@@ -44,37 +44,65 @@
...
-            <input
-              id="access-token"
-              name="accessToken"
-              placeholder="Your access token"
-              type="text">
-            <input id="connect" title="connect" type="submit" value="connect">
-            <p id="connection-status">disconnected</p>
+          <fieldset id="oauth" style="display: block;">
+            <legend>Authentication with OAuth </legend>
+              <input id="login" title="login" type="submit" value="Login">
```
adamweeks, line 65: Since we are just adding event listeners on these buttons, and not using the form to do the submit, let's change the type on all the inputs to "button"

```suggestion
              <input id="login" title="login" type="button" value="Login">
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
+        console.error(error);
+
+      // Implement error handling here
+      });
+  });
+
+  // Adding event handler for register button to register the device
+  document.getElementById('register').addEventListener('click', () => {
+    // again, we don't want to reload when we try to register
+    event.preventDefault();
```
adamweeks, line 230: `event` is not defined here.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,224 @@
...
+      .catch((error) => {
+        // Report the error
+        console.error(error);
+
+      // Implement error handling here
+      });
+  });
+
+  // Adding event handler for register button to register the device
+  document.getElementById('register').addEventListener('click', () => {
```
adamweeks, line 228:
```suggestion
  document.getElementById('register').addEventListener('click', (event) => {
```

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -81,30 +238,63 @@ function connect() {
...
-    else {
-      // Device was already connected
-      resolve();
-    }
+  });
+
+  //  Adding event handler for unregister button to unregister the device
+  document.getElementById('unregister').addEventListener('click', () => {
+    // again, we don't want to reload when we try to register
+    event.preventDefault();
```
adamweeks, line 259: `event` is not defined here

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -81,30 +238,63 @@ function connect() {
...
+    disableMeetingActionButtons();
+    document.getElementById('register').disabled = false;
+    document.getElementById('register-status').innerText = 'device is not registered';
+    document.getElementById('auth-type').disabled = false;
+  });
+}
+
+// There's a few different events that'll let us know that we need to start listening
+// for incoming calls, so we'll wrap a few things up in a function.
+function connect() {
```
adamweeks, line 271: Let's rename this for what it is actually doing since `connect` doesn't really apply any more.

```suggestion
function listenForIncomingMeetings() {
```

adamweeks:
In the `app.js` file, let's utilize [JS Regions](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.JavaScriptRegions) to indicate to developers which functions belong to which part of the code.

I can see two main code regions:

* Authentication
  * `handleAuthTypeChange`
  * `initializeWebexOAuthVersion`
  * `createOAuthForm`
  * `initializeWebexTokenVersion`
  * `createTokenForm`
* Meetings
  * `createEventListenerForMediaButtons`
  * `connect` aka `listenForIncomingMeetings`
  * `bindMeetingEvents`
  * `waitForMediaReady`
  * `joinMeeting`
  * `enableMeetingActionButtons`
  * `disableMeetingActionButtons`
  
Since this is a sample app meant to be used by developers to get them into the webex ecosystem, we need to think of a brand new developer seeing this code for the first time and helping them understand what the code layout is. Most likely they won't have the need for changing auth type, so we really want to emphasize where the meeting code is vs the auth handling code is. By grouping the functions together, this will help in making the code easier to digest.

packages/node_modules/samples/browser-call-with-screenshare/test/wdio/spec/normal-dialing.js
```diff
@@ -134,4 +193,266 @@ describe('Call With Screenshare', () => {
...
+      mccoyPasswordInput.waitForExist();
+      mccoyPasswordInput.setValue(mccoy.password);
+      mccoyPasswordButton.click();
+    });
+
+    it('accepts the permissions grant if it appears', () => {
+      const acceptButton = $('button[name="accept"]');
+
+      // Wait upto 2s for grant page to render
+      // Optionally handle the grant page, but don't fail if this is an app that
```
adamweeks, line 290: This doesn't seem to be optional, it will fail if the grant page appears. But since we are always expecting the grant page for this application, let's just remove the comment about it being optional.

packages/node_modules/samples/browser-call-with-screenshare/test/wdio/spec/normal-dialing.js
```diff
@@ -134,4 +193,266 @@ describe('Call With Screenshare', () => {
...
+      // doesn't show the grant page.
+      acceptButton.waitForExist({timeout: 5000});
+      acceptButton.click();
+    });
+
+    it('redirect to sample app for spock and verify auth', () => {
+      browserFirefox.waitUntil(
+        () => browserFirefox.getUrl() === 'http://localhost:8000/browser-call-with-screenshare/',
```
adamweeks, line 298: This url will fail on non localhost. Let's just check for the page title instead.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -263,69 +425,61 @@ function joinMeeting(meeting) {
   });
 }
 
-// In order to simplify the state management needed to keep track of our button
-// handlers, we'll rely on the current meeting global object and only hook up event
-// handlers once.
+// The function is called when selecting the auth type in the form
+// and presenting the correct form associated with the selected auth type
+function handleAuthTypeChange() {
```
adamweeks, line 430: This function is in the middle of meeting action functions. Please move it to the auth section.

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -10,67 +10,197 @@
...
+
+    // Disabled selecting auth options and enabled register button
+    document.getElementById('register').disabled = false;
+    document.getElementById('auth-type').disabled = true;
+  });
+
+  // Add event handlers to the media buttons
+  createEventListenerForMediaButtons();
+}
+
```
adamweeks, line 142: Add a comment here about this being the beginning of meeting functions.

```js
/**
 * The following functions handle the meeting SDK specifically.
 * We are wiring up actions to buttons to perform meeting functions when clicked.
 * We are also listening and responding to events fired by the SDK.
 */
```

adamweeks (approved):
img

lalli-flores (approved):
Everything looks good!
Great job 👍🏼

PR #2023, opened 2021-03-02
fix(plugin-meetings): fix hostId error when host is not present

arun3528:
Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-180894

In scenario where the host joins first then leaves the meeting without assigning a new host , The locus does not send any host info so we make sure to not parse host info it its not present

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
@@ -425,21 +425,27 @@ export default class LocusInfo extends EventsScope {
   }
 
   /**
+   * checks if the host permissions have changed while in the meeting
+   * This would be the case if your role as host or moderator has been updated
    * @returns {undefined}
    * @memberof LocusInfo
    */
   compareAndUpdate() {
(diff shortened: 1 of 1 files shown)


lalli-flores:
@arun3528, can you update the PR description with context on the changes?

abailey-dev (approved):
The explanation @arun3528 gave when providing context to the issue and the fix made perfect sense to me.

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -425,21 +425,27 @@ export default class LocusInfo extends EventsScope {
   }
 
   /**
+   * checks if the host permissions have changed while in the meeting
+   * could be someone assinged you as a host or moderator for meeting got updated
```
adamweeks, line 429:
```suggestion
   * This would be the case if your role as host or moderator has been updated
```

adamweeks (approved):
LGTM

taymoork2 (approved):
LGTM

PR #2029, opened 2021-03-10
fix(meetings): modified locus sequence comparison algorithm in use

abailey-dev:
Prevents the SDK from sending a DESYNC request for every locus delta event.

Fixes #SPARK-209316
---


arun3528 (approved):
thanks @abailey-dev  really nice job understanding and making the logic more cleaner

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,708 @@
...
+      return;
+    }
+
+    const result = Parser.compare(this.workingCopy, newLoci);
+    const lociComparison = extract(result);
+
+    if (this.debug) {
+      // limited debugging, use chrome extension
+      // for full debugging.
+      LoggerProxy.logger.info(`Locus-info:parser#processDeltaEvent --> Locus Debug: ${result}`);
```
adamweeks, line 573: Could we just use log levels to determine if the use wants debug info and change this to:

```suggestion
      LoggerProxy.logger.debug(`Locus-info:parser#processDeltaEvent --> Locus Debug: ${result}`);
```

abailey-dev: Sure np, I removed .info and used .debug and also removed the conditional.

arun3528:
@adamweeks  @taymoork2  can you have a look at it once

arun3528:
@abailey-dev  can you rebase the branch with master

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+      comparison = `${Parser.loci.GT}:SO001`;
+    }
+    // if last working copy sequence is before the earliest incoming sequence
+    else if (current.max < incoming.min) {
+      // choose right side (incoming)
+      comparison = `${Parser.loci.LT}:SO002`;
+    }
+
+    // if no match above, defaults to null
+    return comparison;
```
mwegman, line 65: Shouldn't this be `Parser.loci.EQ`?

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+    PAUSED: 'PAUSED',
+    WORKING: 'WORKING'
+  }
 
-// https://sqbu-github.cisco.com/WebExSquared/cloud-apps/wiki/Locus-Delta-Events
-Parser.updateWithDeltaLocus = (deltaLocus, meeting) => {
-  if (!(deltaLocus.sequence && (deltaLocus.sequence.rangeStart <= deltaLocus.sequence.rangeEnd) && (deltaLocus.sequence.rangeEnd <= min(deltaLocus.sequence.entries)))) {
-    return null;
+  // loci comparison states
+  static loci = {
```
InteractiveTimmy, line 20: I like this use of constants mapped to the class definition.

There is a safe way to make these immutable as well:

**parser.constants.js**
```js
const CONSTANTS = {
  STATUS: {
    IDLE: 'IDLE',
    /* ... */
  },
  LOCI: {
    EQ: 'EQUAL',
    /* ... */
  }
};

export default CONSTANTS;
```

**parser.js**
```js
import CONSTANTS from './parser.constants.js';
/* ... */
export default class Parser {
  /* ... */
  static get status() { // getter to prevent user assigning values.
    return { ...CONSTANTS.STATUS }; // spread operator to generate a new, immutable object.
  }
  /* ... */
}
```

This will prevent a user from changing the values of `Parser.status` values at runtime and isolate constants to a specific file for this class. You could also just place the constants declaration above the class, but can be problematic for some documentation extractors/generators.

We should discuss this type of constant declaration and accessor for future implementations in other areas.

InteractiveTimmy (commented):
Looks good, I like the usage of classes in this PR. Non-blocking comments here.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
abailey-dev: This follows the Loci Comparison Sequence algorithm and returning EQ here would produce unexpected results. The intent is to return `null` so that the next check can be applied.

abailey-dev: I know it seems like it should be EQ but just because the two conditions above don't apply, according to https://sqbu-github.cisco.com/WebExSquared/cloud-apps/wiki/Locus-Sequence-Comparison-Algorithm it doesn't necessarily mean they are equal. Other checks need to be applied. Returning `null` allows the other checks to continue.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
abailey-dev: This class is an internal class that's private, there are no users that would be consuming it. Personally I wanted to keep the constants / "enums" contained in the class.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+  /**
+   * Checks if two sequences have unequal ranges.
+   * Chooses sequence with most larger range.
+   * @param {Types~Locus} current
+   * @param {Types~Locus} incoming
+   * @returns {object} loci comparison
+   */
+
+
+  static checkUnequalRanges(current, incoming) {
```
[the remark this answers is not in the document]
abailey-dev: Not sure I understand the change, can you explain please.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+    }
+    const hasProp = (prop) =>
+      Object.prototype.hasOwnProperty.call(loci.sequence, prop);
+
+    if (hasProp('rangeStart') && hasProp('rangeEnd')) {
+      return true;
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
abailey-dev: The current function always returns a boolean, the suggested change returns `undefined` in certain instances. 

Also I tried to write very clear, often verbose code so that the intention is very apparent to the next dev. If someone wanted to add additional checks for what is a Loci (say Locus team changed structure of a Loci), I feel the current function makes it clear and obvious. 

The PR this replaces was very condensed, I intentionally tried to do the opposite.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+    // nothing to do, queue is empty
+    if (!newLoci) {
+      setStatus(IDLE);
+    }
+
+    // one or both objects are not locus delta events
+    if (!isLoci(this.workingCopy) || !isLoci(newLoci)) {
+      setStatus(IDLE);
+      LoggerProxy.logger.info('Locus-info:parser#processDeltaEvent --> Ignoring non-locus object. workingCopy:', this.workingCopy, 'newLoci:', newLoci);
+    }
```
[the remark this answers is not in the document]
abailey-dev: Thanks for pointing this out. I can actually remove that check and  just leave 

```
if (!isLoci(this.workingCopy) || !isLoci(newLoci)) {
```

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+   */
+  static getNumbersOutOfRange(list, rangeStart, rangeEnd) {
+    const output = [];
+
+    list.forEach((num) => {
+      // if number is outside of specified range
+      if (num < rangeStart || num > rangeEnd) {
+        // add the number
+        output.push(num);
       }
```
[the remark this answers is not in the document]
abailey-dev: I wanted something very clear and verbose but I think your suggestion is clear as well.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
```diff
@@ -1,118 +1,706 @@
...
+
+    const currentHasNoRange = !current.start && !current.end;
+    const incomingHasNoRange = !incoming.start && !incoming.end;
+    const neitherSeqHasRange = currentHasNoRange && incomingHasNoRange;
+
+    const hasUniqOverlap = (list, min, max) => list.some((seq) => min < seq && seq < max);
+    // current unique entries overlap the total range of incoming
+    const currentUniqOverlap = hasUniqOverlap(current.unique, incoming.min, incoming.max);
+    // vice-versa, incoming unique entries overlap the total range of current
+    const incomingUniqOverlap = hasUniqOverlap(incoming.unique, current.min, current.max);
```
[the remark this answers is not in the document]
abailey-dev: Not sure what you mean?  Current can overlap the total range of incoming, and incoming can overlap the total range of current.

See:
https://sqbu-github.cisco.com/WebExSquared/cloud-apps/wiki/Locus-Sequence-Comparison-Algorithm

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
abailey-dev: Great suggestion though thanks, I think in future for classes that aren't private using `getter` makes a lot of sense.

packages/node_modules/@webex/plugin-meetings/src/locus-info/parser.js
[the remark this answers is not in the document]
mwegman: You separate the method from its comment with two blank lines?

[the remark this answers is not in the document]
abailey-dev: Nope lol. Good catch!

arun3528:
@mwegman  @InteractiveTimmy  are we ok to merge this PR and address the non immediate comments which is not impacting the actual fix in the next PR

PR #2030, opened 2021-03-10
fix(meetings): make reason for decline optional

robstax:
when declining a meeting the current SDK defaults to a `reason: 'BUSY'`. the reason part of the request payload is optional, so we should allow people to decline a call with no reason. currently in the web client, declining with a reason of "busy" causes different behavior when rejecting an incoming call compared to the desktop clients. when rejecting the 1:1 on desktop, they don't send a reason.

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/constants.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/constants.js
@@ -72,9 +72,6 @@ export const VIDEO_STATUS = 'videoStatus';
 // Please alphabetize
 export const _ANSWER_ = 'ANSWER';
 export const _ACTIVE_ = 'ACTIVE';
-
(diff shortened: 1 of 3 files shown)


arun3528:
@robstax  can you confirm if the busy text is used in the SDK on the other side, If alice declines a call from bob does bob get a notification that the reject reason as busy 

If you dont see it then its ok to remove it

arun3528:
@robstax can you remove the constant file as well as its not used anywhere

arun3528 (approved):
Just remove the constants rest should be good , I did verify we dont use the busy string anywhere

robstax:
@arun3528 the whole file? this one packages/node_modules/@webex/plugin-meetings/src/constants.js?

taymoork2 (commented):
LGTM

PR #2047, opened 2021-04-01
fix(plugin-meetings): fix various scenarios related to screen sharing

marcin-bazyl:
This commit fixes a number of issues, most of them related to scenarios
where a meeting participant starts sharing while someone else is already
doing a screen share.

I've also introduced a new public field shareStatus that should give the
app full information about the current share status. It is needed
because otherwise if the app relies only on the notifications, it will
not get the meeting:startedSharingRemote event if page is loaded after
someone already started sharing (the SDK sends startedSharingRemote
event before the app registers for it).

Now the isSharing property is deprecated and I've changed the code so
that internally we don't use isSharing anywhere, so it can be easily
removed in the future.

Fixes #SPARK-222423
Fixes #SPARK-222430
Fixes #SPARK-222426
Fixes #SPARK-220647


packages/node_modules/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -68,14 +68,6 @@ MeetingUtil.remoteUpdateAudioVideo = (audioMuted, videoMuted, meeting) => {
   });
 };
 
-MeetingUtil.checkShare = (meeting) => {
```
marcin-bazyl, line 71: this function was not used, so I removed it

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -422,32 +421,18 @@ skipInNode(describe)('plugin-meetings', () => {
...
           }
           assert.equal(alice.meeting.isSharing, true);
+          assert.equal(alice.meeting.shareStatus, 'local_share_active');
+          assert.equal(bob.meeting.shareStatus, 'remote_share_active');
           console.log('SCREEN SHARE PARTICIPANTS ', JSON.stringify(alice.meeting.locusInfo.participants));
 
           return testUtils.waitUntil(10000);
         }));
 
-      it('alice stops sharing ', () => Promise.all([
```
marcin-bazyl, line 430: I've removed this test, because we already do "bob stops sharing" later on so the scenario of stopping sharing was duplicated. Instead I've changed it now so that Bob steals the share from Alice - that's something that wasn't tested at all before

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1021,61 +1032,133 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      // Check if screenshare is being taken over by a remote
-      // If so, terminate the share and allow the remote to be on the floor
-      // Terminating the share allows the user to screenshare again without any issues
       if (
-        this.isSharing &&
-        this.selfId !== contentId &&
-        disposition !== FLOOR_ACTION.RELEASED
+        payload.previous &&
+        contentId === payload.previous.contentId &&
+        disposition === payload.previous.disposition
```
mwegman, line 1038:
```suggestion
        contentId === payload.previous?.contentId &&
        disposition === payload.previous?.disposition
```

mwegman (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: hmm.. ok, this won't be exactly the same as I intended when writing this code, because if payload.previous is undefined and we get contentId and disposition as undefined too, then your proposed condition will evaluate to true, while the current one evaluates to false, but I guess it's OK as we never get disposition as undefined from locus (we get it as null if no one ever shared in the meeting) and also the later code in this function checks disposition against either released or granted in all the conditions, so it should be fine.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1021,61 +1032,132 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            file: 'meetings/index',
-            function: 'remoteShare'
-          },
-          EVENT_TRIGGERS.MEETING_STOPPED_SHARING_REMOTE
-        );
+        newShareStatus = SHARE_STATUS.NO_SHARE;
       }
 
-      this.members.locusMediaSharesUpdate(payload);
+      if (newShareStatus !== this.shareStatus) {
```
arun3528, line 1068: compare the content ID to check if the remote over takes the share and also pass the content ID as part of the event

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1264,7 +1346,7 @@ export default class Meeting extends StatelessWebexPlugin {
             // We need to check what should be the actual direction of media
             sendAudio: this.mediaProperties.mediaDirection?.sendAudio && !this.audio?.isMuted(),
             sendVideo: this.mediaProperties.mediaDirection?.sendVideo && !this.video?.isMuted(),
-            sendShare: this.mediaProperties.mediaDirection?.sendShare && this.isSharing,
```
arun3528, line 1267: check the console logs to see if there are no warning or errors on stats module

marcin-bazyl: I've checked - I haven't seen any warnings or errors in any of the testing I did

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: done.

arun3528 (approved):

mwegman (approved):

PR #2050, opened 2021-04-07
fix(plugin-meetings): prevents screen share from getting out of sync

abailey-dev:
Improved handling of stopping share stream natively, and checks/fixes if screen share is out of sync.

Fixes #SPARK-190274

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -39,6 +39,7 @@ import {
   FLOOR_ACTION,
   FULL_STATE,
   LAYOUT_TYPES,
+  LIVE,
   LOCUSINFO,
   MEETING_REMOVED_REASON,
   MEETING_STATE_MACHINE,
@@ -54,6 +55,7 @@ import {
   RECORDING_STATE,
   ROAP_SEQ_PRE,
   SDP,
+  SENDRECV,
   SHARE_STATUS,
   SHARE_STOPPED_REASON,
   STATS,
@@ -598,6 +600,29 @@ export default class Meeting extends StatelessWebexPlugin {
      * @memberof Meeting
      */
     this.shareStatus = SHARE_STATUS.NO_SHARE;
+    /**
+     * @instance
+     * @type {Boolean}
+     * @readonly
+     * @private
+     * @memberof Meeting
+     */
+    Object.defineProperty(this, 'isLocalShareLive', {
+      get: () => {
+        const {shareTransceiver} = this.mediaProperties.peerConnection;
+        const shareDirection = shareTransceiver?.direction;
+        const trackReadyState = shareTransceiver?.sender?.track?.readyState;
+        const activeShare = trackReadyState === LIVE;
+        const offersToSendData = shareDirection === SENDRECV;
+
+        if (activeShare && offersToSendData) {
+          return true;
+        }
+
+        return false;
+      },
+      configurable: true
+    });
     /**
      * @instance
      * @type {Array}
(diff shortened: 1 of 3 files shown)


mwegman (approved):

arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -598,6 +600,29 @@ export default class Meeting extends StatelessWebexPlugin {
...
+     * @private
+     * @memberof Meeting
+     */
+    Object.defineProperty(this, 'isLocalShareLive', {
+      get: () => {
+        const {shareTransceiver} = this.mediaProperties.peerConnection;
+        const shareDirection = shareTransceiver?.direction;
+        const trackReadyState = shareTransceiver?.sender?.track?.readyState;
+        const activeShare = trackReadyState === LIVE;
+        const offersToSendData = shareDirection === SENDRECV;
```
marcin-bazyl, line 616: shouldn't offersToSendData be also set to true if shareDirection is "sendonly"? it's a minor thing, because in reality most apps would always want to receive shares too, but in theory we could have an app using SDK that doesn't care about remote presentations and use "sendonly"

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -3509,6 +3527,8 @@ export default class Meeting extends StatelessWebexPlugin {
     }
     const previousSendShareStatus = this.mediaProperties.mediaDirection.sendShare;
 
+    this.setLocalShareTrack(stream);
```
marcin-bazyl, line 3530: I see a small problem here: setLocalShareTrack() does more than just setting up onended listener. It also sends the 'media:ready' event to the app, so now we are sending this event before we've done any validation in `MeetingUtil.validateOption()` and before we've called  `MeetingUtil.updateTransceiver()` so if any of these 2 fail the last even that the app receives is 'media:ready', which doesn't seem right. I guess we should extract the setting up of onended out of setLocalShareTrack() and only do that bit here?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -3537,9 +3557,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
-        this.setLocalShareTrack(stream);
         this.mediaProperties.mediaDirection.sendShare = sendShare;
         this.mediaProperties.mediaDirection.receiveShare = receiveShare;
+      })
+      .catch((error) => {
+        this.unsetLocalShareTrack(stream);
+        throw error;
+      })
+      .finally(() => {
+        const delay = 1e3;
```
marcin-bazyl, line 3568: now that we've moved the onended earlier, it's not clear to me why we still need this piece of code here - it feels a bit "hacky" - why delay of 1s? why not 1.5s or any other value? is there maybe some event that we could wait for instead to do this check? the comment says that we check if share was stopped natively, but that would fire the onended callback, so not really sure why we need this block of code.

marcin-bazyl:
Apologies for making some late comments after the PR's been already merged, but I just wasn't aware of this PR and noticed it when updating the webex web app to use latest SDK and it has sparked a few questions in my head.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
abailey-dev: This all came about from a ticket where the screenshare was started and immediately stopped. Although we moved `onended` due to the flow it did not solve or fix the issue. Therefore I added this `.finally` block, feel free to remove and do extra testing with and without the block by starting screenshare and immediately (like 1 sec later) stop screenshare with native button.

Even with the addition of `.shareStatus` the only reliable way of knowing if screenshare started but was stopped immediately was to check the `peerConnection`.  The delay gives the `peerConnection` time to update, 1 sec is the minimum amount of time needed for the update.

The real question is how often / why would someone start screenshare and immediately stop it using the native button?

PR #2053, opened 2021-04-13
chore(meeting): create internal utils to detect a/v silence

taymoork2:
Fixes [SPARK-217149](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-217149)

---


taymoork2:
I messed with the timeout too much, it's returning TRUE all the time now

arun3528:
@marcin-bazyl  @paulwitty , really appreciate your help review this PR

packages/node_modules/@webex/plugin-meetings/src/media/detectSilence/videoSilence.js
```diff
@@ -0,0 +1,48 @@
...
+    0,
+    0,
+    DETECT_VIDEO_SILENCE_SAMPLE_WIDTH,
+    DETECT_VIDEO_SILENCE_SAMPLE_HEIGHT
+  );
+
+  const frameDataWithoutAlpha = frame.data.filter((item, i) => (i + 1) % 4);
+  const max = Math.max(...frameDataWithoutAlpha);
+
+  return max === 0;
```
marcin-bazyl, line 45: this could be optimised quite a lot. Currently we're doing 2 passes over the whole array of data, while we could do it in 1 pass and stop on the first non-zero value (ignoring alpha)

packages/node_modules/@webex/plugin-meetings/src/media/detectSilence/audiocontext.js
```diff
@@ -0,0 +1,88 @@
+/* globals webkitAudioContext, AudioContext */
+/* eslint-disable no-underscore-dangle */
+
+// eslint-disable-next-line no-nested-ternary
+const NativeAudioContext = typeof AudioContext !== 'undefined' ?
```
marcin-bazyl, line 5: I'm not sure if it's worth trying to fallback to webkitAudioContext. According to https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Migrating_from_webkitAudioContext there are many differences between webkitAudioContext and AudioContext and I think all the browsers we care about (Chrome, FF, Safari) support AudioContext already, don't they?

packages/node_modules/@webex/plugin-meetings/src/media/detectSilence/audioSilence.js
```diff
@@ -0,0 +1,61 @@
...
+      return Promise.resolve(true);
+    }
+
+    analyser.getByteTimeDomainData(bttData); // fill the Uint8Array with data returned from getByteTimeDomainData()
+
+    // NOTE(mpatwardhan): An audio MediaStreamTrack can be silent either due to all samples
+    // being equal to 128 or all samples being equal to 0.
+    return bttData.some((sample) => sample !== 128 && sample !== 0) ?
+      Promise.resolve(false) :
+      new Promise((resolve) => setTimeout(resolve, timeout)).then(doDetectSilence);
```
marcin-bazyl, line 48: I'm a bit confused: why are we setting up a timer here with the same timeout as in lines 33-35? Also I think this code is not monitoring the audio data continuously within the specified timeout, it just gets the last 2048 samples once and if the data was all silence then it schedules another run in 250ms but that one won't do anything because by that time timedOut will be equal true? or have I misunderstood something?

packages/node_modules/@webex/plugin-meetings/src/media/detectSilence/videoSilence.js
```diff
@@ -0,0 +1,48 @@
...
+    - https://github.com/twilio/twilio-video.js/blob/master/lib/webaudio/detectsilence.js
+*/
+
+import {
+  DETECT_VIDEO_SILENCE_SAMPLE_HEIGHT,
+  DETECT_VIDEO_SILENCE_SAMPLE_WIDTH
+} from '../../constants';
+
+/**
+ * Check whether the current video frame is silent by selecting a 50x50
```
marcin-bazyl, line 15: this comment is a bit misleading, we're not selecting a 50x50 sample, but downscaling the whole frame to a 50x50 image.

ciaran-deegan-davra:
For inspiration on the audio side of things, I'd like to reference [hark.js, a small library that detects audio silence in a MediaStream](https://github.com/muaz-khan/WebRTC-Experiment/blob/bbcdbc13b3d7864369f0d09a7c089277dec946da/hark/hark.js#L81). The approach is similar to that of this PR.

PR #2069, opened 2021-05-06
fix(plugin-meetings): fix auto rejoin after meeting ended or in left …

arun3528:
Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-199688
---


packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -280,7 +297,11 @@ export default class ReconnectionManager {
         Metrics.postEvent(reconnectMetric);
         if (reconnectError instanceof NeedsRejoinError) {
           // send call aborded event with catogery as expected as we are trying to rejoin
-          return this.rejoinMeeting(reconnectError.wasSharing);
+
+          if (this.autoRejoinEnabled) {
+            return this.rejoinMeeting(reconnectError.wasSharing);
+          }
+          throw new Error('Cannot re connect the meeting , please rejoin.');
```
mwegman, line 304: Do you want to keep the original `reconnectError` object as well?

arun3528: @mwegman  the parent function already throws that error to the developer

mwegman:
```suggestion
          throw new AggregateError([reconnectError], 'Cannot re connect the meeting , please rejoin.');
```
You could use the AggregateError

arun3528: @mwegman i did remove that as the needTOrejoin error should be sufficient

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -318,6 +318,16 @@ const logRequest = (request, {header = '', success = '', failure = ''}) => {
  * @event meeting:self:lobbyWaiting
  * @instance
  * @type {Object}
+ * @property {Object} reason Reason why user left the meeting
```
marcin-bazyl, line 321: this doesn't look right - the reason should be in `meeting:self:left` event and not in `meeting:self:lobbyWaiting`

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -318,6 +318,16 @@ const logRequest = (request, {header = '', success = '', failure = ''}) => {
...
+ * @memberof Meeting
+ */
+
+/**
+ * Meeting Self Left State
+ * Emitted when user leaves the meeting either by own/server kicking off the user/user not active for 40 seconds or more
```
marcin-bazyl, line 327: the comment says that this event is emitted when "user leaves the meeting either by own...." - does this mean that it should be emitted when client calls `leave()`? Currently the code doesn't do that.

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -827,25 +829,18 @@ export default class Meetings extends WebexPlugin {
...
-                  feedbackId: meeting.correlationId,
-                  locusId: meeting.locusId,
-                  meetingId: meeting.locusInfo?.info?.webExMeetingId
-                }).then(() => this.destroy(meeting, MEETING_REMOVED_REASON.NO_MEETINGS_TO_SYNC));
-              }
-              else {
-                this.destroy(meeting, MEETING_REMOVED_REASON.NO_MEETINGS_TO_SYNC);
-              }
+              // destroy function also uploads logs
+              console.log('RECONNECT   destroy ', this.meeting);
```
marcin-bazyl, line 840: console log

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
```diff
@@ -324,10 +344,19 @@ export default class ReconnectionManager {
...
+
+      throw new Error('Unable to rejoin a meeting already ended or inactive .');
+    }
+
     LoggerProxy.logger.info(`ReconnectionManager:index#executeReconnection --> Current state of meeting is ${this.meeting.state}`);
 
     // If the meeting state was left, no longer reconnect media
-    if (this.meeting.state === _LEFT_) {
+    // or the meeting was deleted in process of re connecting
+    if (this.meeting.state === _LEFT_ || this.meeting.state === _INACTIVE_) {
```
marcin-bazyl, line 359: I'm a bit confused by this. I thought the whole point of the new autoRejoin field in config was to rejoin when we are in "inactive" state, but here we don't allow rejoining in "inactive" state.

marcin-bazyl:
This PR has quite a lot of "todo" comments. Are you planning to address them all before merging this PR? In the previous projects I worked on we only allowed todo comments in code if they were accompanied by a Jira reference - this way we made sure that the todos are actually in our backlog and will be eventually done. Otherwise the code base can easily be filled with lots of todos that never get done.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: yes you are correct , I still need to add a JIRA when a user gets kicked off by the server manually

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
arun3528: yep typo

packages/node_modules/@webex/plugin-meetings/src/reconnection-manager/index.js
arun3528: I think this line wont be executed at all . Removing it 

Before line 347 was added reconnect was in progress but the meeting would have already ended before the reconnect promise would finish

arun3528: The rejoin toggle makes sure that you need to trigger the rejoin compared to SDK doing a rejoin for you

arun3528:
@marcin-bazyl  have updated the PR with review comments and also added the JIRA for the refresh on SDk or on the web client

mwegman (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: looks like this PR got merged with this API typo not being fixed

PR #2070, opened 2021-05-07
fix(meetings): don’t set media ID on reconnect to avoid black screen

jdaly13:
This PR is to remove the unique ID sent for `mediaId` property when sending a ROAP request on reconnect. When a unique ID is sent, it prevents the media from streaming properly. We don't fully understand why yet, but we're reverting back how the request was previously sent.

Fixes #[214541](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-214541). See also [WEBEX-180532](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-180532)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/roap/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/roap/index.js
@@ -1,5 +1,4 @@
 import {StatelessWebexPlugin} from '@webex/webex-core';
-import uuid from 'uuid';
 
 import {ROAP} from '../constants';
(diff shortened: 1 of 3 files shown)


arun3528 (approved):

lalli-flores (changes requested):
Thanks @jdaly13 for looking into this bug!
Could you add an integration test to verify that this change works as intended? There may not be an existing test, so you might need to create a new one.

lalli-flores (approved):
Thanks for adding the test!

PR #2080, opened 2021-05-20
fix: prevent updateVideo from updating the audio track

marcin-bazyl:
fixes #SPARK-209898

The problem was that updateVideo was updating also the audio track
reference in the meeting object, while the rtc connection was still
using the old audio track, so then muting would stop working, because
we would be muting the wrong track.

Same applies to updateAudio.


mwegman (approved):

arun3528 (approved):

taymoork2 (approved):
LGTM

PR #2086, opened 2021-05-27
Muting refactor with meeting as argument in all MuteState public methods (aka option 2)

marcin-bazyl:
Same as https://github.com/webex/webex-js-sdk/pull/2073 but without the callbacks and instead passing the meeting object as an argument into the MuteState class methods when needed.


marcin-bazyl:
closing this one as the changes from this PR are now incorporated in PR#2073

PR #2093, opened 2021-06-07
ci(samples): use webpack dev server

martroos:
The samples tests have been failing for a while, and from running them locally it looks like the oauth client redirect uri needs to be https, but the tests run a http server, so idbroker responds with a uri_redirect_mismatch (these tests used to work so perhaps something was changed on idbroker recently). I have replaced the existing node http server, with a webpack dev server (as run when doing `samples:serve`) and this appeared to allow the oauth tests to pass again as this runs an https server.

---

### DIFF
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -3,13 +3,12 @@
 
 const path = require('path');
 const os = require('os');
-const {createServer} = require('http');
 
 const dotenv = require('dotenv');
 const glob = require('glob');
 const uuidv4 = require('uuid/v4');
-const handler = require('serve-handler');
 const webpack = require('webpack');
+const WebpackDevServer = require('webpack-dev-server');
 
 dotenv.config();
 dotenv.config({path: '.env.default'});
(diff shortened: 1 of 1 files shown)


martroos:
In order to run on saucelabs, via the sauce connect proxy, it looks like some extra configuration will be needed, since browsers will not proxy localhost by default. Some workarounds:

- Firefox - set `network.proxy.allow_hijacking_localhost` to `true` in `prefs`
- Chrome - add the `--proxy-bypass-list=<-loopback>` argument in `args` - doesn't work when using a PAC script (see [here](https://source.chromium.org/chromium/chromium/src/+/main:net/docs/proxy.md#Implicit-bypass-rules)) which is the default when running on saucelabs
- Others - need more investigation

This means that running via https will take some more work and investigation. One alternate option is to use an alias host (i.e. add `test.local` or similar to the host file) as this is proxied by the browsers as normal, but not sure how easy that will be to set up on the test machines (and still needs the redirect uri configuring).

The simplest solution for this is to just add (back?) in the `http` versions of the redirect uri, so I am closing this for now, and it can be re-opened if we want to do more work on this.

PR #2095, opened 2021-06-08
fix(meetings): sending meeting locked/unlocked events

marcin-bazyl:
fixes: SPARK-239945

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/infoUtils.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/infoUtils.js
@@ -67,10 +67,7 @@ InfoUtils.getInfos = (oldInfo, newInfo) => {
   return {
     previous,
     current,
-    updates: {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/plugin-meetings/src/locus-info/infoUtils.js
```diff
@@ -67,10 +67,7 @@ InfoUtils.getInfos = (oldInfo, newInfo) => {
   return {
     previous,
     current,
-    updates: {
-      isLocked: InfoUtils.isLocked(current),
```
marcin-bazyl, line 71: this (and next line) never worked, because `isLocked()`/`isUnlocked()` take policy as argument so it should be `current.policy` and not `current`. I've decided to just remove this code, because we seem to have code that has more specific logic for setting the updates a few lines above.

mwegman (approved):

PR #2100, opened 2021-06-15
fix(p-meetings): update metrics to use WEBEX_SDK as client type

mwegman:
[SPARK-226455](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-226455) #done


lalli-flores (changes requested):
@mwegman [This integration test needs to be updated](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/plugin-meetings/test/integration/spec/journey.js#L54).
Also, could you update the commit message to be easier to understand without context from the changelog?
Something like `fix(p-meetings): update metrics to use WEBEX_SDK as client type`

lalli-flores (approved):
Look good!

PR #2104, opened 2021-06-22
feat(metrics): add senderCountryCode to metrics

mwegman:
[SPARK-231618](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-231618) #done

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
@@ -161,6 +161,7 @@ class Metrics {
       originTime: {
         triggered: new Date().toISOString()
       },
+      senderCountryCode: this.webex.meetings.geoHintInfo?.countryCode,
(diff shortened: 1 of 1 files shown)


packages/node_modules/@webex/plugin-meetings/src/metrics/index.js
```diff
@@ -161,6 +161,7 @@ class Metrics {
       originTime: {
         triggered: new Date().toISOString()
       },
+      senderCountryCode: this.webex.meetings.geoHintInfo?.countryCode,
```
taymoork2, line 164: Should this be optional property or not sending undefined if it doesn't exist?

mwegman: I tested it out. If `getHintInfo` is undefined, the field does not exist in the payload to the server.

lalli-flores (approved):

PR #2108, opened 2021-06-28
fix: make sure we only send integers in renderInfo

marcin-bazyl:


arun3528 (approved):
Did the server already pushed there changes to prod ?

marcin-bazyl:
> Did the server already pushed there changes to prod ?

not yet

mwegman (approved):

PR #2110, opened 2021-06-29
fix(meetings): set both streams when muting

mwegman:
- Ensure both video and audio mute state is sent when muting
   either one.

[SPARK-242346](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-242346) #done

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/muteState.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/muteState.js
@@ -173,8 +173,8 @@ class MuteState {
    * @returns {Promise}
    */
   sendLocalMuteRequestToServer(meeting) {
-    const audioMuted = (this.type === AUDIO) ? this.state.client.localMute : undefined;
-    const videoMuted = (this.type === VIDEO) ? this.state.client.localMute : undefined;
+    const audioMuted = (this.type === AUDIO) ? this.state.client.localMute : meeting.isAudioMuted();
+    const videoMuted = (this.type === VIDEO) ? this.state.client.localMute : meeting.isVideoMuted();
 
     LoggerProxy.logger.info(`Meeting:muteState#sendLocalMuteRequestToServer --> ${this.type}: sending local mute (audio=${audioMuted}, video=${videoMuted}) to server`);
 
--- a/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
+++ b/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
@@ -17,6 +17,8 @@ describe('plugin-meetings', () => {
 
   beforeEach(() => {
     meeting = {
+      isAudioMuted: () => false,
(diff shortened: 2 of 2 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/muteState.js
```diff
@@ -173,8 +173,8 @@ class MuteState {
    * @returns {Promise}
    */
   sendLocalMuteRequestToServer(meeting) {
-    const audioMuted = (this.type === AUDIO) ? this.state.client.localMute : undefined;
-    const videoMuted = (this.type === VIDEO) ? this.state.client.localMute : undefined;
+    const audioMuted = (this.type === AUDIO) ? this.state.client.localMute : meeting.isAudioMuted();
```
marcin-bazyl, line 176: shouldn't we push back on this as a backend bug? when client does audio mute we set videoMuted to undefined, so the API request sent to locus /media endpoint only has the audioMuted property set. If as a result of that the video mute status in locus is changed that looks like incorrect Locus behaviour to me....

mwegman:
Bug assigned to Locus for accepting only `audioMuted` or `videoMuted`
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-247638

arun3528:
its a major bug i am not sure how fast the server team will fix it , @marcin-bazyl  should we just release it and then revert the fix later ?

packages/node_modules/@webex/plugin-meetings/src/meeting/muteState.js
marcin-bazyl: also I think this can cause a problem: isAudioMuted() ends up calling MuteState.isMuted() and that one returns true also in the case if audio is only remotely muted but not locally muted. So in a scenario where the app asked to mute video and audio was remotely muted we would end up doing local mute for both audio and video. So we probably want to use here `meeting.audio?.state.client.localMute`.

marcin-bazyl:
ok, I guess we could deliver this is a workaround for now, but I think we need to tweak it a bit - see my other comment.

packages/node_modules/@webex/plugin-meetings/src/meeting/muteState.js
mwegman: If we do `meeting.audio?.state.client.localMute` in the scenario you outlined above, we would unmute the remote audio. Is that what we want?

marcin-bazyl: as discussed offline, Locus has 2 separate APIs: one for local mute and one for remote mute, so we wouldn't affect the remote mute state

marcin-bazyl:
@lalli-flores asked me to raise a new PR for this, because Mike is on holiday, so here it is: https://github.com/webex/webex-js-sdk/pull/2116

lalli-flores:
Placed by https://github.com/webex/webex-js-sdk/pull/2116

PR #2111, opened 2021-07-07
fix(meetings): check current user is still in-meeting before sending meeting:self:unmutedByOthers

marcin-bazyl:
I've found this when testing being removed from a PMR meeting.


mwegman (approved):

lalli-flores (commented):
Thanks for finding this bug and adding a fix!
Can you update your commit message so it's clearer for someone without context on the changelog? Also, don't forget to add a scope of your change.

Something like:
```
fix(meetings): check current user is still in-meeting before unmuting

Scenario found while testing a user being removed from a PMR

lalli-flores (changes requested):
we can merge once commit has been updated

marcin-bazyl:
I've updated the commit message. Thanks.

lalli-flores (approved):

PR #2113, opened 2021-07-08
fix(meetings): remove non-Webex calls after user leaves

marcin-bazyl:
If you use the SDK to join a meeting via a SIP url and that meeting has nothing to do with webex (so it's not a url of a PMR or of any webex space) - for example if you call a meeting hosted by Cisco Meeting Server (CMS): example: `alpha_test2@alphauk.cisco.com` then the meeting object doesn't get destroyed after you leave the meeting and it remains there forever.

This kind of meeting is reported by locus as type = "SIP_BRIDGE". I've checked with Locus team and they confirmed that it's better to treat SIB_BRIDGE types of meetings like CALL ones (and not like MEETING ones)

Fixes [SPARK-248251](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-248251)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/constants.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/constants.js
@@ -115,6 +115,7 @@ export const _RESOURCE_ROOM_ = 'RESOURCE_ROOM';
 export const _RECEIVE_ONLY_ = 'RECVONLY';
 export const _ROOM_ = hydraTypes.ROOM;
 
+export const _SIP_BRIDGE_ = 'SIP_BRIDGE';
 export const _SIP_URI_ = 'SIP_URI';
 export const _SEND_RECEIVE_ = 'SENDRECV';
 export const _SEND_ONLY_ = 'SENDONLY';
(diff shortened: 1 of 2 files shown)


arun3528 (approved):
i am sure the meeting end will work but need to make sure other scenarios are working as well, There are more checks in the code for "CALL" Should we just change the locus event  if they see BRIDGE change that to CALL !?

marcin-bazyl:
yeah, good point, but there are quite a few places where type is checked for "CALL" and we should test all these scenarios before making bigger changes like that. I don't know all these features that well, so I would prefer to merge this PR for now (it fixes the issue that was reported and we're not aware of any other scenarios not working right now) and raise a new Jira to look into all the other scenarios.

lalli-flores (changes requested):
Can you please update your commit message to be more meaningful in the changelog?
Also, please don't forget to add a scope to your commit messages.

Suggestion:
```
fix(meetings): remove non-Webex calls after user leaves

marcin-bazyl:
Jira raised for checking the remaining use cases: SPARK-248639

lalli-flores (approved):
Thanks for the updated commit message!

PR #2115, opened 2021-07-12
feat(webex-core): allow for the x-prelogin-userid header to be set when sending user activation

martroos:
To allow tracking of metrics generated by the webex web client, the x-prelogin-userid header needs to be set when sending the request to activate a user. This PR will allow for this to be provided optionally - if it is omitted, the existing behaviour will continue (it is set to `undefined`).


packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -905,6 +915,47 @@ describe('webex-core', () => {
           assert.isAbove(Object.keys(unauthServices.list(false, 'signin')).length, 0);
           assert.equal(Object.keys(unauthServices.list(false, 'postauth')).length, 0);
         }));
+
+      const getActivationRequest = (requestStub) => {
```
lalli-flores, line 919: Can you move either, at the beginning of the describe block, inside the `beforeEach` or inside the new test (e.g. `it` block)? Just not in the middle of the block

lalli-flores (changes requested):
Please update the [object definition](https://github.com/webex/webex-js-sdk/pull/2115/files#diff-8f9a8aaed0d87dc3441f4a1f87bbd2d470ed4a92efdac5badc45f58f483b3dc8R221) for `ValidateUserPTO` accordingly.

Seems like the commit message is getting truncated as well. Can we make it more concrete?
ex:
```
feat(core): set x-prelogin-userid header if activated user is logged in
```
Sorry if this is not accurate but I don't think I complete understand when the header would be set or when it would not

martroos:
> Please update the [object definition](https://github.com/webex/webex-js-sdk/pull/2115/files#diff-8f9a8aaed0d87dc3441f4a1f87bbd2d470ed4a92efdac5badc45f58f483b3dc8R221) for `ValidateUserPTO` accordingly.
> 
> Seems like the commit message is getting truncated as well. Can we make it more concrete?
> ex:
> 
> ```
> feat(core): set x-prelogin-userid header if activated user is logged in
> ```
> 
> Sorry if this is not accurate but I don't think I complete understand when the header would be set or when it would not

I have shortened the commit message - hopefully it is clear enough now. What is the limit on commit message length so that it doesn't get truncated? I assumed if it passed the husky check it would be acceptable (since this validates the message) but I guess this isn't the case?

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
martroos: I left it just above the tests that use it for context (defining close to the only places it is used), but I have moved it to the start of the describe (after the constants, but before the `beforeEach`). Is there a general style in terms of organising tests I can refer to in future?

lalli-flores: The convention is to add it at the beginning of the describe block. If there are test that have common setups, we usually do a inner describe to group related-tests together

lalli-flores:
> I have shortened the commit message - hopefully it is clear enough now. What is the limit on commit message length so that it doesn't get truncated? I assumed if it passed the husky check it would be acceptable (since this validates the message) but I guess this isn't the case?

It doesn't truncate now 🙂
I don't remember if its' a 100 or 120 characters, but husky should be set to not let you commit if it's over 😮, I thought you skipped it. Maybe something to look into

lalli-flores (approved):

PR #2116, opened 2021-07-12
fix(meetings): set both streams when muting

marcin-bazyl:
Fixes bug where if muting only audio or video, the other would reset to default mode instead of staying on their previous state. This issue is fixed by sending the preferred state of both, audio and video, when requesting the server to mute either one.

Fixes [SPARK-242346](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-242346)


lalli-flores:
Replaced by https://github.com/webex/webex-js-sdk/pull/2118 due to CI issues in this PR.

PR #2122, opened 2021-07-21
feat(internal-plugin-user): update name using PATCH

shnaaz:
webclient needs to update user's givenName and familyName to have more friendly PMR name for free user. User's PMR name format is "givenName familyName's Personal Room" - today we are doing a PATCH on conversation "/users/user" to update display name, we need to also update both "givenName" and "familyName" and for that we need to PATCH on the CI endpoint "/Users/{userid}" instead when setting name during user activation flow.
More details can be found here - SPARK-224067

---


packages/node_modules/@webex/internal-plugin-user/src/user.js
```diff
@@ -232,25 +250,42 @@ const User = WebexPlugin.extend({
...
+      return Promise.reject(new Error('`options.givenName` and `options.familyName` and `options.displayName` are required'));
+    }
+
+    return this._setUser({
+      schemas: ['urn:scim:schemas:core:1.0', 'urn:scim:schemas:extension:cisco:commonidentity:1.0'],
+      name: {givenName: options.givenName, familyName: options.familyName},
+      displayName: options.displayName
+    })
+      .then((res) => res.body);
+  },
```
mwegman, line 287:
```suggestion
  updateName({givenName, familyName, displayName} = {}) {
    if (!givenName || !familyName || !displayName) {
      return Promise.reject(new Error('`options.givenName` and `options.familyName` and `options.displayName` are required'));
    }

    return this._setUser({
      schemas: ['urn:scim:schemas:core:1.0', 'urn:scim:schemas:extension:cisco:commonidentity:1.0'],
      name: {givenName, familyName},
      displayName
    })
      .then((res) => res.body);
  },
```

mwegman (approved):

PR #2128, opened 2021-07-26
test(meetings): fix intermittent failure of 'bob joins the meeting' integration test

marcin-bazyl:
fixes #SPARK-239986

The issue was that we sometimes get an 'members:update' event with some other update so waitForEvents() would resolve and the assert after it would fail. Changed the call to waitForEvents() to wait for the right status


packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -134,11 +134,13 @@ const waitForEvents = (scopeEvents, timeout = max) => {
               }
             }
 
-            if (obj.match && obj.match(value)) {
-              clearTimeout(timer);
-              obj.scope.off(obj.event, handler);
+            if (obj.match) {
```
marcin-bazyl, line 137: this change is needed, because when `match` callback is defined we don't want to resolve when the number of received events equals `scopeEvents.length` - we only want to resolve once the `match()` callback succeeds

arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
arun3528: && operator makes that happen tho !?

marcin-bazyl: no, the problem is that if both conditions are together wrapped with a single "if" then in the case where `obj.match` is defined but `obj.match(value)` returns false we go to the "else if": `else if (result.length === scopeEvents.length) {` and we don't want that to happen in this case.

mwegman (approved):

PR #2149, opened 2021-08-12
fix(encryption): setting onbehalfof to support an edge case

gannons:
This PR will fix [SPARK-257081 - Update webex-js-sdk to support `onBehalfOf` as well as `userId` for scenarios where the CO does an onBehalfOf himself](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-257081). Basically it's an edge case where a Compliance Officer is decrypting content **onBehalfOf** himself. KMS had a bug where, in this scenario, the code flow assumed this is not an **onBehalfOf** request, and as a result fails with a 403.

This issue is being covered by 

 * BEMS01309085 [New] C3 TO BE ASSIGNED CPR COMPANY   691741276  Failed to download Summary Report via ediscovery Download Manager. Hangs at downloading... Indicates 'disconnected-waiting to try' and can't download the data finaly
 * Cisco TAC Case : 691741276

The customer is also complaining about how long it's taken for this issue to be resolved. i.e. It's been 30+ days so far.

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
+++ b/packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
@@ -662,6 +662,7 @@ const KMS = WebexPlugin.extend({
       clientId: originalContext.clientInfo.clientId,
       credential: {
         userId: onBehalfOf,
+        onBehalfOf, // Supports running onBehalfOf self. i.e. A CO which calls onBehalfOf with CO.id.
         bearer: originalContext.clientInfo.credential.bearer
       }
     };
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/internal-plugin-encryption/src/kms.js
```diff
@@ -662,6 +662,7 @@ const KMS = WebexPlugin.extend({
       clientId: originalContext.clientInfo.clientId,
       credential: {
         userId: onBehalfOf,
+        onBehalfOf, // Supports running onBehalfOf self. i.e. A CO which calls onBehalfOf with CO.id.
```
gannons, line 665: Previously this field was depreciated. To resolve this bug KMS are re-using it. There is a bit of redundancy here and the API is less user friendly. But KMS report that the alternative is a more complex item of work, with a higher risk.

taymoork2 (approved):
LGTM

Ran tests locally and the new test cases passed

PR #2150, opened 2021-08-13
Bail if service does not exist in postauth catalog

taymoork2:
Fixes [SPARK-255996](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-255996)

---


arun3528:
Looks good did we confirm if the server is coming as part of the catalog ?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -412,6 +413,7 @@ export default class Meetings extends WebexPlugin {
       }
 
       return Promise.all([
+        this.webex.internal.services.hasRequiredServices(REQUIRED_SERVICES.postauth, 'postauth'),
```
arun3528, line 416: currently all the users are CI registred uses so we will have the utc catalog , Do you know of any type of user not going to get the catalog like guest token etc can you check it

arun3528 (commented):
Can you describe a little bit on pre and post auth i dont know most of us know much about it . 

Also did we test all type of uses ?

PR #2151, opened 2021-08-13
fix(internal-plugin-user): for update name any one name is required

shnaaz:
Fix for updateName- from webclient a user can only provide a givenName so removing the check which required all names to be provided. 
Now check is that at-least  one of givenName, familyName or displayName should be there.

Fixes SPARK-257436

---

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-user/src/user.js
+++ b/packages/node_modules/@webex/internal-plugin-user/src/user.js
@@ -265,15 +265,14 @@ const User = WebexPlugin.extend({
    * Updates a user's name with webex.
    * @instance
    * @memberof User
-   * @param {Object} options
-   * @param {string} options.givenName (required)
-   * @param {string} options.familyName (required)
-   * @param {string} options.displayName (required)
+   * @param {string} givenName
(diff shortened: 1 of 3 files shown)


jonniespratley (approved):
👍

taymoork2 (approved):
LGTM

packages/node_modules/@webex/internal-plugin-user/src/user.js
```diff
@@ -265,15 +265,14 @@ const User = WebexPlugin.extend({
...
+   * @param {string} givenName
+   * @param {string} familyName
+   * @param {string} displayName
    * @returns {Promise<Object>}
    */
   updateName({givenName, familyName, displayName} = {}) {
-    if (!givenName || !familyName || !displayName) {
-      return Promise.reject(new Error('`options.givenName` and `options.familyName` and `options.displayName` are required'));
+    if (!(givenName || familyName || displayName)) {
+      return Promise.reject(new Error('One of `givenName` and `familyName` or `displayName` is required'));
```
pacacc, line 275: the comment and the if statement don't agree here, do they?

Issue #2155, opened 2021-08-23
Support for cross-region boards


pcl:
Currently, webex-js-sdk doesn't properly support cross-region boards. Let's fix that.

Most recently, this manifests in realtime.js:115, which does not propagate the board channel URL (and thus the board channel cluster) to the board registration call, resulting in an attempted registration in the wrong cluster.

https://github.com/webex/webex-js-sdk/blob/097f11497e6befaf7da918ebe925b8be8172ad8c/packages/node_modules/%40webex/internal-plugin-board/src/realtime.js#L97

jbenyovs:
@pcl can you please provide some more details and concrete examples about this issue showing the expected and actual cluster? Can you also share the board channel URL you mentioned? Any steps we can use to reproduce this? I assume it is reproducible with the https://js.samples.s4d.io/browser-plugin-meetings/ demo here, but can you confirm?

pcl:
I haven't tried to reproduce the bug in your sample app, but it's probably reproducible there.

Here are two different board URLs:

https://board-a.wbx2.com/board/api/v1/channels/16321ab0-00d1-11ec-bca4-e56fcbadda24
https://board-k.wbx2.com/board/api/v1/channels/3e9e50a0-ff63-11eb-b573-eda7098f33a9

A user homed in cluster `a` will not be able to call `register()` on the `board-k` URL, since the registration has to happen against `board-k.wbx2.com`.

pcl:
Also, fyi, the API call you're using has been deprecated. I had the board team turn it back on, so that at least the client sorta works, but you should not expect this API to be around forever.

https://sqbu-github.cisco.com/WebExSquared/board/blob/master/server/src/main/java/com/cisco/wx2/board/server/BoardController.java#L81

jbenyovs:
@pcl thanks, I have reached out to engineering to check your commit and reply to this.

pcl:
Great, thanks.

I see that some checks failed, but I can't see why. Can you post details here? I don't intend to give CircleCI access to my private repositories...

pcl:
Oh and also -- the changes in the related PR are working great in our downstream codebase.

jbenyovs:
@pcl we received a feedback from engineering management and according to them this is not a bug, it's an enhancement request. Additionally, board plugin is internal and only meant to be used by the web app team.

pcl:
Ok cool. Be sure to let engineering management know about the bug in the
[...]

PR #2156, opened 2021-08-24
fix: support for connecting to realtime channel for federated boards

pcl:
Federated boards are ones that live in a cluster other than the active
user's default cluster. Registration of a websocket for these boards
must happen through a channel-aware registration API, rather than the
deprecated cluster-wide registration API.

This change uses that approach when the board.useFederatedRegistration
configuration is set to 'true', and uses the current approach when
that config is not set. So, clients wishing for the federation-capable
approach must set the config option in the config tree when constructing
their Webex object.

This could be further developed to use a feature toggle for rollout, but
any such improvements should still respect the configuration value as a
default, so that clients other than the web client can enable the
federated behavior.

This addresses #2155.


packages/node_modules/@webex/internal-plugin-board/src/realtime.js
```diff
@@ -91,19 +91,27 @@ const RealtimeService = WebexPlugin.extend({
...
 
-        return this.realtimeChannels.get(channel.channelId);
+      registrationPromise = this.webex.internal.board.register({bindings: requestBindings})
+        .then(({webSocketUrl, bindings}) => ({socketUrl: webSocketUrl, bindings: bindings[0]}));
+    }
+
+    return registrationPromise.then((registration) => {
+      this.realtimeChannels.add({
+        ...registration,
+        channelId: channel.channelId
```
mwegman, line 110: Do you not want the `channelId` from `registration` to override the `channelId` from `channel`?

pcl: The data structure generated by `registrationPromise` doesn't include a `channelId` field.

packages/node_modules/@webex/internal-plugin-board/src/board.js
```diff
@@ -506,6 +506,22 @@ const Board = WebexPlugin.extend({
...
+  registerToChannel(channel) {
+    return this.webex.request({
+      method: 'POST',
+      api: 'board',
+      uri: `${channel.channelUrl}/register`,
+      body: {action: 'ADD'}
+    })
+      .then((res) => res.body);
+  },
+
```
toretre, line 524: @pcl  I wonder if there should be a corresponding method for removing the binding ? Or could clients use 'unregisterFromSharedMercury' method ?

pcl: Yes, `unregisterFromSharedMercury()` is sufficient from what we've seen.

arun3528:
Can we add some tests here

jonniespratley:
Can you please squash this commit.

jonniespratley (approved):
Also can we get a unit-test for these changes, and they will be good to go.

PR #2163, opened 2021-08-31
fix(internal-plugin-conversation): add hint when uploading space avatars

mjovicic:

Fixes #SPARK-239726


mwegman:
You don't need to make the first 2 changes. They are just suggestions. The last one need to be implemented

mwegman (changes requested):

PR #2168, opened 2021-09-03
docs(samples): fixed remoteShare sync between 2 users

emanuallan:
fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-254773, relates to https://github.com/webex/webex-js-sdk/pull/1879/files

### DIFF
--- a/packages/node_modules/samples/browser-call-with-screenshare/app.js
+++ b/packages/node_modules/samples/browser-call-with-screenshare/app.js
@@ -12,7 +12,6 @@
 
 
 let activeMeeting;
-let remoteShareStream;
 let isToken;
 let isOAuth;
 let webex;
@@ -305,7 +304,6 @@ function bindMeetingEvents(meeting) {
 
   meeting.on('meeting:startedSharingRemote', () => {
     // Set the source of the video element to the previously stored stream
-    document.getElementById('remote-screen').srcObject = remoteShareStream;
     document.getElementById('screenshare-tracks-remote').innerText = 'SHARING';
   });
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -305,7 +304,6 @@ function bindMeetingEvents(meeting) {
 
   meeting.on('meeting:startedSharingRemote', () => {
     // Set the source of the video element to the previously stored stream
```
arun3528, line 306: add a check for remoteShareStream then set it else dont set it , Sometimes a user is already sharing initially so tthe vent will come before the actual remote sttreams

packages/node_modules/samples/browser-call-with-screenshare/app.js
```diff
@@ -330,8 +328,7 @@ function bindMeetingEvents(meeting) {
       document.getElementById('remote-view-audio').srcObject = media.stream;
     }
     if (media.type === 'remoteShare') {
-      // Remote share streams become active immediately on join, even if nothing is being shared
-      remoteShareStream = media.stream;
+      document.getElementById('remote-screen').srcObject = media.stream;
```
arun3528, line 331: condition to set the streams , check meeting.shareShare === 'remotet_share"

emanuallan: isnt this already occuring in line 332: `if (media.type === 'remoteShare')` ?

emanuallan: I also noticed that while stepping, `meeting:startedSharingRemote` is called before the remoteShareStream is set, so remoteShareStream is always undefined.

PR #2169, opened 2021-09-06
docs: add npm bash config to contributing guide

Coread:
Myself and another developer came across this situation on separate occasions on fresh ubuntu work environments. By default npm is using `sh` which means that the distsrc and srcdist commands fail (and consequently the precommit hook as well). 

Adding this to the contributing guide will save someone else the pain of debugging what is a really confusing error.


jonniespratley (changes requested):
Please squash your commits

jonniespratley (approved):
Thank you for your contribution.

PR #2170, opened 2021-09-06
SPARK-238602 Decrypt space information for eDiscovery

decryan1984:
eDiscovery will be handling new activities which contain space information (name, description etc.) that must be decrypted so that the end user will be able to read it.

Fixes #[https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-238602]

---

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
+++ b/packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
@@ -190,30 +190,31 @@ class Transforms {
 
         // add warning properties to the activity - these will be recorded in the downloader
         if (container.warning) {
-          activity.spaceWarning = container.warning; // Depreciate this property
+          activity.spaceWarning = container.warning; // Remove this property once all clients are using the content container model
           activity.containerWarning = container.warning;
         }
 
         // set space name and participants on activity
         if (container.containerName) {
-          activity.spaceName = container.containerName; // Depreciate this property
+          activity.spaceName = container.containerName; // Remove this property once all clients are using the content container model
           activity.containerName = container.containerName;
         }
         else if (container.isOneOnOne) {
           const displayNames = (container.participants || []).concat(container.formerParticipants || []).map((p) => p.displayName).join(' & ');
 
           // One to One spaces have no space name, use participant names as 'Subject' instead
-          activity.spaceName = displayNames; // Depreciate this property
+          activity.spaceName = displayNames; // Remove this property once all clients are using the content container model
           activity.containerName = displayNames;
         }
         else {
-          activity.spaceName = ''; // Depreciate this property
+          activity.spaceName = ''; // Remove this property once all clients are using the content container model
           activity.containerName = '';
         }
 
         // post and share activities have content which needs to be decrypted
-        // as do meeting and recording activities
-        if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp')) {
+        // as do meeting, recording activities, customApp extensions, and space information updates
+        if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp') &&
+          !activity.spaceInfo?.name && !activity.spaceInfo?.description) {
           return Promise.resolve(object);
         }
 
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -475,6 +517,21 @@ class Transforms {
...
+      requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
+        [container.descriptionEncryptionKeyUrl, container.description, {onBehalfOf: container.onBehalfOfUser}])
+        .then((decryptedContainerDescription) => {
+          container.description = decryptedContainerDescription;
+        })
+        .catch((reason) => {
+          ctx.webex.logger.error(`Decrypt container description error for ${container.containerType} container ${container.containerId}: ${reason}`);
+          // add warn property to container info - this warning will be recorded in the downloader
+          container.warning = reason;
+          // don't return, attempt to decrypt the name first
```
decryan1984, line 531: I believe this is the correct thing to do here

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -249,6 +250,47 @@ class Transforms {
...
+              activity.spaceInfo.description = decryptedMessage;
+            })
+            .catch((reason) => {
+              ctx.webex.logger.error(`Decrypt activity.spaceInfo.description error for activity ${activity.activityId} in container ${activity.targetId}: ${reason}`);
+              activity.error = reason;
+
+              return Promise.resolve(object);
+            }));
+        }
+        if (activity.spaceInfo?.previousName && activity.spaceInfo.previousEncryptionKeyUrl) {
```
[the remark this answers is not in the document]
gannons: What happens if both the previousName and current name use the same key? Will this property be set? Or should you default to the main key? I assume `previousEncryptionKeyUrl ` is most relevant in a key rotation environment such as FedRAMP?

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -475,6 +517,21 @@ class Transforms {
       return Promise.resolve(object);
     }
 
+    // decrypt description if present with a descriptionEncryptionKeyUrl
+    if (container.description && container.descriptionEncryptionKeyUrl) {
```
gannons, line 521: The container description uses a different key. This is what I recalled you saying at the scrum. I guess the activity keys are always the same. How come it can't occur for the `name` of the container? How come we don't check the `name` here but do use the `description`?

Actually I'm wrong, below we decrypt the name. Can you elaborate on when the keys would be different? i.e. Unlike with the activities where they are the same.

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -364,6 +364,58 @@ describe('EDiscovery Transform Tests', () => {
 
         return result;
       });
+
+      it('Calls the correct decrypt functions when transforming activity.spaceInfo.name', () => {
+        object.body.verb = 'update';
```
gannons, line 369: It's always going to be an `update`?

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -364,6 +364,58 @@ describe('EDiscovery Transform Tests', () => {
...
+            assert.callCount(ctx.webex.internal.encryption.decryptText, 2);
+            assert.equal(activity.error, undefined);
+          });
+
+        return result;
+      });
+
+      it('Calls the correct decrypt functions when transforming activity.spaceInfo.description', () => {
+        object.body.verb = 'update';
+        object.body.spaceInfo = {description: 'Encrypted Description'};
```
gannons, line 396: Looks like the test does not require `descriptionEncryptionKeyUrl` to be set. That is not a blocker, it simply shows the test can't be too realistic.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
gannons: I don't think it matters. The call to `ctx.webex.internal.encryption.decryptText` _may_ need to retrieve a key. As such the `then` here occurs in the future. i.e. You've already executed the `containerName` code below.

It seems a bit wrong that this code only `returns` the `containerName` decryption promise. i.e. Technically it should return a `Promise.all(.....)` contain every promise here.

But I'm rusty on JavaScript at this point. I think I'm right but there is room for me to be wrong. Also that is not new to this PR.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -190,30 +190,31 @@ class Transforms {
...
+          activity.spaceName = ''; // Deprecate this property
           activity.containerName = '';
         }
 
         // post and share activities have content which needs to be decrypted
-        // as do meeting and recording activities
-        if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp')) {
+        // as do meeting and recording activities and space information updates
+        if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp') &&
+          !activity.spaceInfo?.name && !activity.spaceInfo?.description) {
```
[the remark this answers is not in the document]
decryan1984: Eslint seems to prefer the && on the first line

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -190,30 +190,31 @@ class Transforms {
 
         // add warning properties to the activity - these will be recorded in the downloader
         if (container.warning) {
-          activity.spaceWarning = container.warning; // Depreciate this property
+          activity.spaceWarning = container.warning; // Deprecate this property
```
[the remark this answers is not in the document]
decryan1984: I'd see depreciated as a purely monetary term, whereas I've definitely seen deprecated in spring notation etc.

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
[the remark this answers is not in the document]
decryan1984: Displaying the previous name wasn't a hard requirement, but I said we may as well use it if it's there. Both properties belong to a ConversationPreviousDetails object on the Conversation, so we should be able to use them together reliably, I'd be reluctant to mix and match urls, I'd rather leave out the previous name from the output than risk a decryption error.

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
decryan1984: Yes

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
decryan1984: According to Diego they may end up being different due to key rotation, there may be other reasons, it has apparently always been part of the design: https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?spaceKey=WBXT&title=Conversation%3A+Public+Spaces#Conversation:PublicSpaces-Description

packages/node_modules/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
decryan1984: Since this is a description _activity_, there is no descriptionEncryptionKeyUrl,  the encryptionKeyUrl will be the correct property to decrypt the description. For a _space_ it's different, there will be two urls, one for name and one for description.

gannons (approved):

packages/node_modules/@webex/internal-plugin-ediscovery/src/transforms.js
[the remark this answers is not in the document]
taymoork2: Is there a way to alert/console.warn if a deprecated property as being used?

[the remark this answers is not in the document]
decryan1984: None that I'm aware of @taymoork2 , and after thinking about it some more, I'm not even sure deprecated is the right term to use here, those space properties are actually no longer used in the latest ediscovery clients, so I've changed the comment to read 'Remove this property once all clients are using the content container model' (we plan to do this with a mandatory upgrade of the clients at some point).

jonniespratley (approved):
Please squash your commits. Thank you

decryan1984:
> Please squash your commits. Thank you

Thanks Jonnie, squashed to a single commit

PR #2171, opened 2021-09-07
Pkce code challenge

Coread:
This implements PCKE code challenge as described in https://datatracker.ietf.org/doc/html/rfc7636

Wikis for the idbroker APIs:
https://wiki.cisco.com/display/CIOAUTH/OAuth2+API+to+Get+Authorization+Code
https://wiki.cisco.com/display/CIOAUTH/OAuth2+API+to+Get+Access+Token+by+Authorization+Code

SPARK-204648

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "cookie-parser": "^1.4.4",
     "core-decorators": "^0.20.0",
     "cors": "^2.8.5",
-    "crypto-js": "^3.1.9-1",
+    "crypto-js": "^4.1.1",
     "debug": "^3.2.6",
     "dependency-check": "^4.1.0",
     "detective": "^5.2.0",
--- a/packages/node_modules/@webex/plugin-authorization-browser-first-party/src/authorization.js
+++ b/packages/node_modules/@webex/plugin-authorization-browser-first-party/src/authorization.js
@@ -11,8 +11,15 @@ import {base64, oneFlight, whileInFlight} from '@webex/common';
 import {grantErrors, WebexPlugin} from '@webex/webex-core';
 import {cloneDeep, isEmpty, omit} from 'lodash';
 import uuid from 'uuid';
+import base64url from 'crypto-js/enc-base64url';
+import CryptoJS from 'crypto-js';
+
+// Necessary to require lodash this way in order to stub
+// methods in the unit test
+const lodash = require('lodash');
 
 const OAUTH2_CSRF_TOKEN = 'oauth2-csrf-token';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "cookie-parser": "^1.4.4",
     "core-decorators": "^0.20.0",
     "cors": "^2.8.5",
-    "crypto-js": "^3.1.9-1",
+    "crypto-js": "^4.1.1",
```
Coread, line 116: crypto-js upgrade necessary for URL safe base64 encoder for sha256 toString method

packages/node_modules/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -86,7 +94,7 @@ browserOnly(describe)('plugin-authorization-browser-first-party', () => {
 
           return webex.authorization.when('change:ready')
             .then(() => {
-              assert.calledOnce(webex.request);
+              assert.calledTwice(webex.request);
```
Coread, line 97: This test had been broken since september 2019. The addition of the call to get the catalog first had resulted in two calls to webex.request instead of only one.

lalli-flores (commented):
Code looks good, some clarification needed on the tests

packages/node_modules/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -11,8 +11,13 @@ import {base64, oneFlight, whileInFlight} from '@webex/common';
 import {grantErrors, WebexPlugin} from '@webex/webex-core';
 import {cloneDeep, isEmpty, omit} from 'lodash';
 import uuid from 'uuid';
+import base64url from 'crypto-js/enc-base64url';
+import CryptoJS from 'crypto-js';
+
+const lodash = require('lodash');
```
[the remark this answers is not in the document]
Coread: This allows it to be stubbed. I've commented on the test file about how it is somehow necessary to use this syntax. I'll add a comment describing why this is necessary.

lalli-flores (commented):
Code looks good, but can you split the commits?

* `package/package-lock.json` should be its own commit:
```
build(package): update cyprto-js to v4.1.1
```
* Other is the PKCE changes; scope should be plugin name (truncated in this case because it's so long):
```
feat(plugin-authorization): add PKCE code challenge and verification
*

PR #2172, opened 2021-09-07
fix(internal-plugin-conversation): uploading space avatars

mjovicic:

Fixes #SPARK-239726


jonniespratley (approved):
Please squash your commits and 👍

PR #2177, opened 2021-09-09
fix(internal-plugin-conversation): uploading space avatars 

mjovicic:

Fixes #SPARK-239726

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/share-activity.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/share-activity.js
@@ -78,7 +78,6 @@ const ShareActivity = WebexPlugin.extend({
 
           return url;
         }));
-
       this.hiddenSpaceUrl = Promise.resolve(attrs.conversation._hiddenSpaceUrl || this._retrieveSpaceUrl(`${attrs.conversation.url}/space/hidden`)
         .then((url) => {
           attrs.conversation._hiddenSpaceUrl = url;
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/internal-plugin-conversation/src/share-activity.js
```diff
@@ -251,12 +250,15 @@ const ShareActivity = WebexPlugin.extend({
   /**
    * @param {File} file
    * @param {string} uri
+   * @param {Object} uploadOptions
```
lalli-flores, line 253: Can you outline what the expected format of this object is?

lalli-flores (changes requested):
@mjovivic, don't forget to read [our contribuiton guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md), specifically how to [format git commits](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#git-commit-guidelines). This is important to us because commits translate into our [changelog](https://github.com/webex/webex-js-sdk/blob/master/CHANGELOG.md) that many developers and customers use to reference changes in the SDK. Your commit has to make sense in the context of the changelog.

Please do the following:
1) pull latest master (your branch has to be up-to-date)
2) Rebase your branch on latest master
3) Amend your commit message

Here's a suggested message:
```
feat(internal-conversation): allow space avatars to be uploaded

Uploading an image with "spaceAvatar" role allows us to bypass
File service file-sharing blocking rules
```

packages/node_modules/@webex/internal-plugin-conversation/src/share-activity.js
mjovicic: I have added additional comment there. Thanks

lalli-flores (approved):

PR #2179, opened 2021-09-09
docs(plugin-meetings): updated docs with known limitation about remote sharing

emanuallan:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-254773

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/README.md
+++ b/packages/node_modules/@webex/plugin-meetings/README.md
@@ -1307,6 +1307,22 @@ meeting.on('media:stopped', (media) => {
 }
 ```
 
+### Event Caveats
+##### Remote screen share is not displayed if started before participant joins
+
+When trying to receive screen share in an application, standard plugin-meeting workflow is as follows:
+1. Set event listener for `media:ready`
+2. Call `join()`
+3. Call `addMedia()` with `options.mediaSettings.receiveShare=true`
+4. Wait to receive an event from `media:ready` with payload `type=remoteShare` that contains the remote share media stream
+5. Set `srcObject` of a `video` element in the application to the previously obtained media stream
+(e.g. `document.getElementById('remote-screen').srcObject = media.stream`)
+
+These steps are necessary because you can only register for events like `meeting:startedSharingRemote` on the meeting object,
+however, you can only do that after you've received the `meeting:added` event notification with the meeting payload.
+In the case when a host has already begun sharing before a participant joins, the registration for
+`meeting:startedSharingRemote` happens too late (after the SDK code that would send that event is already executed).
+
 ### Members Events
 
 ```js


emanuallan:
Should i provide background info to the developers? I didnt think theyd be interested so I kept it out

ex. 

"There is a known limitation in the SDK that you can only register for events like `meeting:startedSharingRemote` on the meeting object, 
however, you can only do that after you've received the `meeting:added` event notification with the meeting payload. In the case when a host has already begun sharing before a participant joins, the registration for `meeting:startedSharingRemote` happens too late (after the SDK code that would send that event is already executed)"

lalli-flores:
@emanuallan, yes adding some context on the topic would be helpful to developers. This way they can determine if the use case applies to them or not.
Although, I wouldn't call it "an SDK limitation" that's how eventing works

packages/node_modules/@webex/plugin-meetings/README.md
```diff
@@ -1307,6 +1307,14 @@ meeting.on('media:stopped', (media) => {
 }
 ```
 
+##### Not seeing remote share if host started sharing screen before participant joins
```
lalli-flores, line 1310: Maybe add a section for potential other caveats in the future?
```suggestion
### Event Caveats
#### Not seeing remote share if host started sharing screen before participant joins
```

packages/node_modules/@webex/plugin-meetings/README.md
```diff
@@ -1307,6 +1307,14 @@ meeting.on('media:stopped', (media) => {
...
+
+What apps typically do is call `join()` and then `addMedia()` with `options.mediaSettings.receiveShare=true` 
+and you will get a `media:ready` event with `type='remoteShare'` that contains a stream. 
+Then you set the srcObject of the video element to that stream to be able to see it
```
lalli-flores, line 1314: Explain a little more why this is necessary

emanuallan:
Updated the docs with your suggestions @lalli-flores , feel free to review

packages/node_modules/@webex/plugin-meetings/README.md
```diff
@@ -1307,6 +1307,19 @@ meeting.on('media:stopped', (media) => {
 }
 ```
 
+### Event Caveats
+##### Not seeing remote share if host started sharing screen before participant joins
+
+What apps typically do is call `join()` and then `addMedia()` with `options.mediaSettings.receiveShare=true`
```
lalli-flores, line 1313: I think the context would be easier to follow if it's outlines as a series of steps
```
When trying to receive screen share in an application, standard plugin-meeting workflow is as follow:
1. Set event listener for `media:ready`
2. Call `join()`
3. Call `addMedia()` with `options.mediaSettings.receiveShare=true`
4. Wait to receive an event from `media:ready` with payload `type=remoteShare` that contains the remote share media stream
5. Set `srcObject` of a `video` element in the application to the previously obtained media stream
(e.g. `document.getElementById('remote-screen').srcObject = media.stream`)

lalli-flores (changes requested):
I think this would be easier to follow if you divide the explanation into 3 sections:
#### Context
#### Problem
#### Solution

PR #2183, opened 2021-09-15
fix(webex-core): make waitForService check for preauth catalog instead of limited

martroos:
It appears that the `waitForService` function is waiting for a catalog with the name `'limited'`, which, based on the names available [in the catalog status prop](https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/@webex/webex-core/src/lib/services/service-catalog.js#L22), I believe is incorrect, and should in fact be `'preauth'` (the GET is to `/limited/catalog` which may be the confusion).

The tests I have added fail before the change to the `waitForService` function, and pass after it. Additionally I have checked this when running the webex web client with the fix in it (this is how I first noticed the issue) and it appears to work correctly once the fix is in (previously it would error after 5s with `service 'metrics' was not found after waiting`, as web client attempts to send a metric on first load).


arun3528:
@InteractiveTimmy any thoughts on this , when i try this out i do see the preauth catalog

packages/node_modules/@webex/webex-core/src/lib/services/services.js
```diff
@@ -565,7 +565,7 @@ const Services = WebexPlugin.extend({
     }
 
     return new Promise((resolve, reject) => {
-      const groupsToCheck = ['limited', 'signin', 'postauth'];
+      const groupsToCheck = ['preauth', 'signin', 'postauth'];
```
InteractiveTimmy, line 568: Good catch!

Surprised we went this long without fixing this.

packages/node_modules/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -1028,37 +1028,68 @@ describe('webex-core', () => {
...
 
-          describe('when using the name parameter property', () => {
-            it('should resolve to the appropriate url', () =>
-              Promise.all([
-                services.waitForService({name}),
-                services.initServiceCatalogs()
-              ])
-                .then(([foundUrl]) => assert.equal(foundUrl, url)));
-          });
+          describe('when only the preauth (limited) catalog becomes available', () => {
```
InteractiveTimmy, line 1037: Test look good [seems to have more coverage].

InteractiveTimmy (approved):
LGTM! As long as tests pass.

lalli-flores (approved):

arun3528 (approved):
thanks @InteractiveTimmy

PR #2189, opened 2021-09-22
feat(meetings): invite participants via phone number

jdaly13:
This PR will enable meeting hosts to invite users via an E.164 formatted telephone address to join the current meeting via PSTN, currently only email address types were supported, this will add an extra check to see if the address property is a valid phone number

Fixes #[SPARK-269110]


arun3528 (commented):
Looks good should we change the address to phone number or sip address ,Also check with the locus team if they only support phone number and not sip address

jdaly13:
> Looks good should we change the address to phone number or sip address ,Also check with the locus team if they only support phone number and not sip address

Sounds good - I was going off the current documentation - https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?spaceKey=LOCUS&title=WEBEX-189819+%3A+Host+is+using+VoIP+audio+for+invite+by+phone
I will reach to the locus team to confirm

taymoork2 (approved):
LGTM

jdaly13:
@arun3528 - they currently do not support SIP address and I have changed the address property to phoneNumber as the final step creates the address field in the body of the put anyway - this is more inline with the way email is passed in

lalli-flores (changes requested):
@jdaly13 , can you squash your commits as this is only one new feature change?
Something like
```
feat(meetings): invite participants via phone number
```
Otherwise changes look good

lalli-flores (approved):

lalli-flores (commented):
Actually, @jdaly13 I don't see any tests added for this new feature?

jdaly13:
@lalli-flores there are no tests because it's not new functionality - we are just adding another acceptable parameter, the current tests only check if meeting.invite returns a promise - there are no unit test for members currently

PR #2192, opened 2021-09-27
fix(plugin-meetings): fixed false firing of stop recording event

emanuallan:
this would occur upon joining a meeting that was created using the API

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-255479

Also noticed that there are some issues with the `current` and `previous` state (in ControlUtils) while I was trying to create tests for this pr. `current` is not the same in both cases and i think that is important to look into to properly make tests. will need to get more info on this. 

Top 2 logs are from *api created meeting*, bottom 2 are from *ui created meeting*, first in each section is `previous` and the other is `current`. can ignore `#meeting:recording` just using that to filter through logs 

Screen Shot 2021-09-27 at 12 58 48 PM

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/locus-info/controlsUtils.js
@@ -46,12 +46,13 @@ ControlsUtils.getControls = (oldControls, newControls) => {
     previous,
     current,
     updates: {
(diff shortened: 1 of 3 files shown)


arun3528 (approved):
lgtm

marcin-bazyl (approved):

emanuallan:
Had to close because the merge isnt from fork, continuation: https://github.com/webex/webex-js-sdk/pull/2194

PR #2198, opened 2021-09-29
fix(plugin-meetings): isRecording shows which member started recording

emanuallan:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-189023?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=tr

Previously `this.parsedLocus.controls.modifiedBy` was being used to get the `recordingId`, but that doesn't exist (always `undefined`, and thus the reason why `isRecording` is always `false` when comparing id's). Changed it to the correct reference `this.parsedLocus.controls.record.modifiedBy` and now it correctly gets the id

Screen Shot 2021-09-29 at 12 33 57 PM
Screen Shot 2021-09-29 at 1 55 51 PM

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
@@ -10,7 +10,8 @@ import LocusDeltaParser from '@webex/plugin-meetings/src/locus-info/parser';
 import {
   LOCUSINFO,
   RECORDING_STATE,
-  LOCUSEVENT
+  LOCUSEVENT,
+  EVENTS
 } from '../../../../src/constants';
 
 import {self, selfWithInactivity} from './selfConstant';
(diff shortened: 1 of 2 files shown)


arun3528 (approved):

lalli-flores (changes requested):
@emanuallan can you make your commit message more meaningful for the changelog?
ex:
```
fix(meetings): isRecording shows which member started meeting recording

**Comment by emanuallan** [comment]
@lalli-flores addressed your input, thank you.

File: packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js [hunk 1]
```diff
@@ -267,6 +268,39 @@ describe('plugin-meetings', () => {
         ];
       });
 
+      it('should pass the correct id values', () => {
```
lalli-flores, line 271:
```suggestion
      it('should emit participant update of recordingId ', () => {
```

lalli-flores (approved):

emanuallan:
updated test description to be more clearer

PR #2200, opened 2021-10-04
Chore(deps-dev): Updated 

emanuallan:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-264725

Updated webpack-dev-server to latest version. Also updated, webpack-cli because otherwise it breaks trying to use the --env flag

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3272,15 +3272,9 @@
       }
     },
     "@discoveryjs/json-ext": {
-      "version": "0.5.2",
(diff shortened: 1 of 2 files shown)


emanuallan:
Latest commit included adding `multicast-dns` to the resolutions section of the `package.json` file. This caused not only `multicast-dns` to upgrade but also `dns-packet` (to the jira requested version of v.5.3.0). This solution was adopted from a previous [stack overflow solution](https://stackoverflow.com/questions/67680511/npm-audit-issue-with-dns-packet) that was very similar to our issue.

PR #2208, opened 2021-10-11
fix(plugin-meetings): Handle the roap state for sending answer

arun3528:
There was a race condition  when we make a api for /media the response was coming back after the mercury message so the roap state was getting set later after mercury message so we assuming the request is send as soon the api call is made so that if the mercury message comes in then the state changes correctly if the response fails then state will be reset 

so we move the state change before the request then just log the success

https://sqbu-github.cisco.com/WebExSquared/cloud-apps/wiki/Locus---ROAP-callflow
https://sqbu-github.cisco.com/WebExSquared/linus/blob/master/components/protocols/roap/src/roap_session.cpp#L629
https://sqbu-github.cisco.com/Calliope/documentation/blob/master/apis/roap/api_v2.md

Fixes #[SPARK-272808](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-272808)

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/roap/handler.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/roap/handler.js
@@ -192,8 +192,10 @@ export default class RoapHandler extends StatelessWebexPlugin {
   handleAction(session, action, meeting, correlationId) {
     let signal;
 
+
     switch (action.type) {
       case ROAP.RECEIVE_ROAP_MSG:
+        LoggerProxy.logger.log(`Roap:handler#handleAction --> RECEIVE_ROAP_MSG event captured, reciving a roap message : ${JSON.stringify(action)}`);
         if (compareWithLastRoapMessage(this.lastRoapMessage, action)) {
           LoggerProxy.logger.warn(`Roap:handler#handleAction --> duplicate roap offer from server: ${action.msg.seq}`);
         }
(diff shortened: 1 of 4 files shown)


packages/node_modules/@webex/webex-core/src/interceptors/redirect.js
```diff
@@ -74,6 +74,8 @@ export default class RedirectInterceptor extends Interceptor {
       response.body.errorCode === LOCUS_REDIRECT_ERROR &&
       response.body.location) {
       options = clone(options);
+
+      this.webex.logger.warn('redirect: url redirects needed from', options.url);
```
marcin-bazyl, line 78: this and the other change in this file seem unrelated to the roap issue. This is for the locus redirect issue, isn't it? So just checking if it's here intentionally.
Also, these new logs are logging `options.url` while the code around here is using `options.uri` so just double checking that you want to log `url` and not `uri`

packages/node_modules/@webex/plugin-meetings/src/roap/state.js
```diff
@@ -25,6 +25,10 @@ const shouldStep = (roap, meeting) => {
 };
 
 const handleTransition = (value, signal, meeting) => {
+  LoggerProxy.logger.log('Roap:state#handleTransition --> current roap state', value);
+  LoggerProxy.logger.log('Roap:state#handleTransition --> next roap state', signal);
```
marcin-bazyl, line 29: log text is incorrect - this is not displaying the next state, but the received signal
also a minor thing, but wouldn't it be better to log both in a single log call?

packages/node_modules/@webex/plugin-meetings/src/roap/handler.js
```diff
@@ -192,8 +192,10 @@ export default class RoapHandler extends StatelessWebexPlugin {
   handleAction(session, action, meeting, correlationId) {
     let signal;
 
+
     switch (action.type) {
       case ROAP.RECEIVE_ROAP_MSG:
+        LoggerProxy.logger.log(`Roap:handler#handleAction --> RECEIVE_ROAP_MSG event captured, reciving a roap message : ${action}`);
```
marcin-bazyl, line 198: I think having `${action}` in the log like this will result in just having "[object Object]" in the output log, won't it? I think we need to use `JSON.stringify()` (same for the other log in this file)

PR #2206, opened 2021-10-08
Wbxappapi fixes

marcin-bazyl:
These are 2 commits with fixes for 2 issues I've found today when testing the web app with wbxappapi:
1. SDK fails to get the meeting info for meetings currently in progress when web app is loaded
2. SDK fails to get the meeting info for PMR

TODO:
Tests need to be updated/added.

PR #2219, opened 2021-11-05
fix(plugin-meetings): added missing orgId for v2 get meeting info requests

emanuallan:
Got from: https://github.com/webex/webex-js-sdk/pull/2206, seperated it so it would be it's own trackable pr. 

Ticket responsible for this pr: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-274567

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
@@ -189,10 +189,11 @@ MeetingInfoUtil.getDestinationType = async (from) => {
  * @param {Object} options type and value to fetch meeting info
  * @param {String} options.type One of [SIP_URI, PERSONAL_ROOM, MEETING_ID, CONVERSATION_URL, LOCUS_ID, MEETING_LINK]
  * @param {Object} options.destination ?? value.value
+ * @param {Object} options.webex Webex SDK instance
  * @returns {Object} returns an object with {resource, method}
  */
 MeetingInfoUtil.getRequestBody = (options) => {
-  const {type, destination} = options;
+  const {type, destination, webex} = options;
   const body = {};
 
   switch (type) {
@@ -201,7 +202,7 @@ MeetingInfoUtil.getRequestBody = (options) => {
       break;
     case _PERSONAL_ROOM_:
       body.userId = destination;
-      body.orgId = ''; // TODO: when to use the org iD (fetch others PMR)
+      body.orgId = webex.credentials.getOrgId(); // TODO: when to use the org iD (fetch others PMR)
       break;
     case _MEETING_ID_:
       body.meetingKey = destination;
--- a/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting-info/utilv2.js
+++ b/packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting-info/utilv2.js
@@ -14,6 +14,7 @@ import {
   _MEETING_UUID_
 } from '@webex/plugin-meetings/src/constants';
 import MeetingInfoUtil from '@webex/plugin-meetings/src/meeting-info/utilv2';
+import WebexCore from '@webex/webex-core';
 
 describe('plugin-meetings', () => {
   describe('Meeting Info Utils V2', () => {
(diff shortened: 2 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting-info/utilv2.js
```diff
@@ -72,12 +73,17 @@ describe('plugin-meetings', () => {
 
     describe('#getRequestBody', () => {
       it('for _PERSONAL_ROOM_', () => {
+        const webex = new WebexCore({
+          credentials: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJyZWFsbSI6Im15LXJlYWxtIn0.U16gzUsaRW1VVikJA2VeXRHPX716tG1_B42oxzy1UMk'
```
lalli-flores, line 77: Where did these credentials come from? is this a real value?

emanuallan: I got it from another test `packages/node_modules/@webex/webex-core/test/unit/spec/credentials/credentials.js` line 275

packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
```diff
@@ -192,7 +192,7 @@ MeetingInfoUtil.getDestinationType = async (from) => {
  * @returns {Object} returns an object with {resource, method}
```
lalli-flores, line 192: Can you add a comment that `webex` is now part of the options?

packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
```diff
@@ -201,7 +201,7 @@ MeetingInfoUtil.getRequestBody = (options) => {
       break;
     case _PERSONAL_ROOM_:
       body.userId = destination;
-      body.orgId = ''; // TODO: when to use the org iD (fetch others PMR)
+      body.orgId = webex.credentials.getOrgId(); // TODO: when to use the org iD (fetch others PMR)
```
lalli-flores, line 204: What does this TODO mean?

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting-info/utilv2.js
lalli-flores: I see, seems like an encoded base64 string for a JWT token. can you add a comment for that?

packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
emanuallan: https://github.com/webex/webex-js-sdk/pull/2206#discussion_r744543716

@arun3528 May need your opinion on this.

arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting-info/utilv2.js
emanuallan: @arun3528 Need your opinion on what to do with this TODO Comment. bg info: https://github.com/webex/webex-js-sdk/pull/2206#discussion_r744543716

lalli-flores: I confirmed with the API team that the org Id that we need to send should correspond to that of the user for which we are trying to fetch PMR information. This code won't work when requesting PMR information of external org users, we need a different solution

marcin-bazyl: ok, yes, then this fix only works for checking the logged in user's PMR or PMRs of other users from the same org, so it's still better than nothing and fixes the most common use case for the web app, but yeah a further, better fix is required. How are you planning to proceed with this? deliver this fix and raise a new Jira for the "other org" case? or do a proper fix for all cases in this existing jira/pr?

lalli-flores:
Replaced by https://github.com/webex/webex-js-sdk/pull/2237

PR #2220, opened 2021-11-05
fix(meetings-plugin): missing meetingInfo for meetings already in progress on page load

emanuallan:
Got from: https://github.com/webex/webex-js-sdk/pull/2206, separated it so it would be it's own trackable pr.

Ticket responsible for this pr: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-274564 

at the time of this writing: test fails when using v2 (doesnt fail when using v1) but it is failing for another reason not related to my addition because it fails even without my changes.


arun3528 (approved):
The code looks good , Enabling V2 version needs to be a separate PR and we need to fix the broken test when V2 is enabled

packages/node_modules/@webex/plugin-meetings/test/integration/spec/space-meeting.js
```diff
@@ -44,6 +44,15 @@ skipInNode(describe)('plugin-meetings', () => {
           console.log('CONVERSATION', conversation);
           space = conversation;
         })
+        .then(async () => {
+          const destinationWithType = await alice.webex.meetings.meetingInfo.fetchMeetingInfo(space.url, 'CONVERSATION_URL');
```
marcin-bazyl, line 48: I see new test is added for cases with type===CONVERSATION_URL and type===undefined, but the code changes in this PR are for the case when type===_LOCUS_ID_. Where is the test for that? Is it already covered by some existing test? 
Also while trying to see if there is an existing test I came across tests that check that `MeetingsUtil.extractDestination` has been called (like here: https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/plugin-meetings/test/unit/spec/meetings/index.js#L450 and more cases further in that file) and `MeetingsUtil.extractDestination` doesn't exist anymore as it was moved into src/meeting-info/util.js so I'm not sure how these tests are passing. I would expect that they need updating, or am I missing something?

arun3528: Going to get handled here https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-259981

taymoork2:
Closing in favor of #2231

PR #2227, opened 2021-11-12
fix(meetings): move meetingInfo instantiation to onReady callback

robstax:
Fixes https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-280998
---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
@@ -116,15 +116,7 @@ export default class Meetings extends WebexPlugin {
      */
     constructor(...args) {
       super(...args);
-      /**
-       * The MeetingInfo object to interact with server
-       * @instance
-       * @type {Object}
-       * @private
-       * @memberof Meetings
-       */
 
-      this.meetingInfo = null;
       /**
        * The Meetings request to interact with server
        * @instance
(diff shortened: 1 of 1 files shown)


lalli-flores (changes requested):
@robstax can you update the commit message? It's a little misleading for other consumer of the SDK looking at the changelog, as it seems very web client specific. From the SDK perspective nothing is crashing and seeing the word "crash" could be alarming to some customers.

```
fix(meetings): move meetingInfo to constructor

File: packages/node_modules/@webex/plugin-meetings/src/meetings/index.js [hunk 1]
```diff
@@ -124,7 +124,7 @@ export default class Meetings extends WebexPlugin {
        * @memberof Meetings
        */
 
-      this.meetingInfo = null;
+      this.meetingInfo = this.config.experimental.enableUnifiedMeetings ? new MeetingInfoV2(this.webex) : new MeetingInfo(this.webex);
```
marcin-bazyl, line 127: I've tested this and `this.config` is actually not available at the point when constructor is called, so this fix doesn't work. However, I think I've found the right place to do this. When I've moved this line of code into the callback for `READY` event that's registered in the `onReady()` method, then it works for me.

robstax: i was worried about that... was gonna see if the tests caught it and then investigate it

robstax: thanks for diggin in @marcin-bazyl !

lalli-flores (approved):

PR #2229, opened 2021-11-15
fix(http): params.xhr for all requests

chrisadubois:
Fixes #https://jira-eng-gpk2.cisco.com/jira/browse/ONL-9213

ANY library that instruments XHR requests would run into this same problem, not isolated to logrocket. GET requests were not instrumented because it only happened from bindProgressEvents


packages/node_modules/@webex/http-core/src/request/request.shim.js
```diff
@@ -93,11 +94,23 @@ export default function _request(options) {
    */
   function bindProgressEvents(params, o) {
     if (params.method && ['PATCH', 'POST', 'PUT'].includes(params.method.toUpperCase())) {
-      params.xhr = new XMLHttpRequest();
+      if (!params.xhr) {
```
lalli-flores, line 97: When would there be the case that `xhr` wouldn't exist at this point if the only place this private function is called is right after `setXhr`?

lalli-flores (commented):
Without having much context on this code or what it is that you're trying to do. Can you elaborate on the implications of this change? Why do you need to create an XHR object for all requests and not just the already defined verbs?

packages/node_modules/@webex/http-core/src/request/request.shim.js
chrisadubois: it's for completeness and modularity. Bind progress events OR setXHR could disappear at any time, and we want the function to be complete on it's own.

chrisadubois:
There are no implications of the change beyond being able to instrument (through the use of external third party libraries) all verbs that run through the js-sdk.

taymoork2 (approved):
FISI

PR #2233, opened 2021-11-18
fix(plugin-authorization-browser-first-party): emailHash

shnaaz:
This is a security bug which needs to be fixed for webapp. emailHash should be used in GET request instead of an email. Changed initiateLogin and initialize in plugin-authorization-browser-first-party for the same. Added a unittest to check for emailHash in options.

SPARK-280519
---


packages/node_modules/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -259,6 +259,26 @@ browserOnly(describe)('plugin-authorization-browser-first-party', () => {
...
+        const expectedEmailHash = '73062d872926c2a556f17b36f50e328ddf9bff9d403939bd14b6c3b7f5a33fc2';
+
+        sinon.spy(webex.authorization, 'initiateAuthorizationCodeGrant');
+
+        return webex.authorization.initiateLogin({email: 'test@email.com'})
+          .then(() => {
+            assert.called(webex.authorization.initiateAuthorizationCodeGrant);
+            const grantOptions = webex.authorization.initiateAuthorizationCodeGrant.getCall(0).args[0];
+
+            assert.equal(grantOptions.emailHash, expectedEmailHash);
```
marcin-bazyl, line 278: maybe also worth adding a check that email itself is not set in grantOptions?

marcin-bazyl:
LGTM although I know nothing about this area of the SDK

marcin-bazyl (approved):

arun3528 (approved):

packages/node_modules/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -127,9 +127,14 @@ const Authorization = WebexPlugin.extend({
    * @returns {Promise}
    */
   initiateLogin(options = {}) {
+    if (options.email) {
+      options.emailHash = CryptoJS.SHA256(options.email).toString();
+    }
+    delete options.email;
```
lalli-flores, line 133: You should not modify arguments of the function. You may create a copy of the arguments and modify that but do not introduce side-effects on this function.

lalli-flores (changes requested):

packages/node_modules/@webex/plugin-authorization-browser-first-party/src/authorization.js
martroos: Is it specifically the `delete` you are objecting to for this function, or the way the function works overall? Whilst I agree that in general function arguments should not be modified (unless actually intended), this particular function has always modified `options`, for example adding `state` (and attributes of it) if it didn't exist previously, or more recently adding in the `code_challenge`.

lalli-flores: Overall; I know this was modified recently by Colin and we didn't catch that.
I don't agree that we should keep doing something that we know it's not ok just because that's how it was done in the past. Let's do better with our code changes.
Creating an object copy and modifying that would satisfy my request

lalli-flores (approved):
Thanks for addressing concerns @shnaaz

PR #2252, opened 2021-12-17
WIP test(samples): real device mobile test

taymoork2:
VERY WIP

Known issues:
- Aliasing localhost isn't working properly so being on VPN or going to Sauce won't work because the local server isn't available
  - Something like ngrok or localtunnel could be achievable but are extra steps
  -  It was working at one point, but not anymore, so just grabbing your local DHCP IP is serviceable 
  - Tunneling doesnt become a big issue when using either BrowserStack or LambdaTest since they provide workarounds builtin
- Haven't tested two Androids or two iOS devices, but it's possible
- iOS is broken right now and more investigation is needed into webdriverio to fix, however click should work

Fixes [SPARK-292078](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-292078)

---


jonniespratley (approved):
I was able to follow instructions and run tests for both IOS and ANDROID. The following examples commands I used:

For IOS:

```
$ IOS=true npm run samples:test:mobile -- --spec packages/node_modules/samples/browser-auth-jwt/test/wdio/spec/jwt-authentication.js
```

For ANDRIOD:

```
$ ANDRIOD=true npm run samples:test:mobile -- --spec packages/node_modules/samples/browser-auth-jwt/test/wdio/spec/jwt-authentication.js
```

PR #2256, opened 2022-01-04
fix(plugin-meetings) :add google bitrate adjustment

arun3528:
http://www.rtcbits.com/2016/11/controlling-bandwidth-usage-in-webrtc.html

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-279720)

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/config.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/config.js
@@ -76,7 +76,8 @@ export default {
       // please note, these are the maximum bandwidth values
       // the server supports, minimums have to be tested
       audio: 64000,
-      video: 4000000
+      video: 4000000,
+      startBitrate: 2000
     },
     screenFrameRate: 10,
     videoShareFrameRate: 30,
--- a/packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js
@@ -79,9 +79,18 @@ const setMaxFs = (sdp, level = QUALITY_LEVELS.HIGH) => {
   let replaceSdp = sdp;
   const maxFsLine = `${SDP.MAX_FS}${MAX_FRAMESIZES[level]}`;
 
-  replaceSdp = replaceSdp.replace(/(\na=fmtp:(\d+).*level-asymmetry-allowed=1.*)/gi, `$1;${maxFsLine}`);
+  replaceSdp = replaceSdp.replace(/(\na=fmtp:(\d+).*profile-level-id=.*)/gi, `$1;${maxFsLine}`);
 
   return replaceSdp;
+};
+
+
+const setStartBitrateOnRemoteSdp = (sdp) => {
+  if (StaticConfig.meetings.bandwidth.startBitrate) {
+    sdp = sdp.replace(/(\na=fmtp:(\d+).*profile-level-id=.*)/gi, `$1;x-google-start-bitrate=${StaticConfig.meetings.bandwidth.startBitrate}`);
+  }
+
+  return sdp;
 };
 
 /**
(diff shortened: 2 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js
```diff
@@ -84,6 +84,13 @@ const setMaxFs = (sdp, level = QUALITY_LEVELS.HIGH) => {
   return replaceSdp;
 };
 
+
+const setMaxFsOnRemoteSdp = (sdp) => {
```
marcin-bazyl, line 88: function name is wrong, the code is not setting maxFs but adding x-google-start-bitrate

packages/node_modules/@webex/plugin-meetings/src/config.js
```diff
@@ -76,7 +76,8 @@ export default {
       // please note, these are the maximum bandwidth values
       // the server supports, minimums have to be tested
       audio: 64000,
-      video: 4000000
+      video: 4000000,
+      startBitrate: 2000
```
marcin-bazyl, line 80: just wondering if we actually need a separate value for this.. why not use the existing `video` value (divided by 1000)? also why 2000 and not 4000? what value did they test with?

if we are adding a new value then might be worth supporting a case where it's set to null or undefined and not appending `x-google-start-bitrate` in that case.

packages/node_modules/@webex/plugin-meetings/src/peer-connection-manager/index.js
```diff
@@ -84,6 +84,13 @@ const setMaxFs = (sdp, level = QUALITY_LEVELS.HIGH) => {
...
+  sdp = sdp.replace(/(\na=fmtp:(\d+).*level-asymmetry-allowed=1.*)/gi, `$1;x-google-start-bitrate=${StaticConfig.meetings.bandwidth.startBitrate}`);
```
marcin-bazyl, line 89: I'm not sure if it's the best approach to search for `level-asymmetry-allowed=1` I think we can set `x-google-start-bitrate` even if level asymmetry is not allowed (so value 0 or not present in the fmtp line). Maybe it would be better to search for `profile-level-id=` string?

packages/node_modules/@webex/plugin-meetings/src/config.js
arun3528: The customer tested with 2000 , The audio video values are set in the TIAS line which was used for linus i am not sure if edonus is using it , Anyways those values are for the TIAS line which is to indicate the max bitrateValue supported , 

Its better to check with the edonus team on max bitrate

packages/node_modules/@webex/plugin-meetings/test/unit/spec/peerconnection-manager/index.js
```diff
@@ -0,0 +1,43 @@
...
+import PeerConnectionManager from '@webex/plugin-meetings/src/peer-connection-manager/index';
+import StaticConfig from '@webex/plugin-meetings/src/common/config';
+
+describe('Peerconnection Manager', () => {
+  beforeEach(() => {
+    StaticConfig.set({bandwidth: {audio: 50, video: 500, startBitrate: 2000}});
+  });
+  describe('Methods', () => {
+    describe('setRemoteSessionDetails', () => {
+      it('change the start bitrate on remoteSDP', async () => {
```
marcin-bazyl, line 13: is there also a test that checks that "x-google-start-bitrate" is not added when startBitrate is undefined in config?

marcin-bazyl (approved):

jonniespratley (approved):

packages/node_modules/@webex/plugin-meetings/test/unit/spec/peerconnection-manager/index.js
arun3528: done

PR #2259, opened 2022-01-10
Spark-290101 - Toggle Meeting Info between v1 and v2

mkesavan13:
Hi team,

This PR completes [https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-290101](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-290101). 

- Instead of having two methods enable and disable, we have one single method `_toggleUnifiedMeetings` which when set to true will enable Unified Meetings and when set to false will disable Unified Meetings.
- Unit test cases are written accordingly for base, success & failure cases.
- Samples app for browser-plugins-meetings is also updated with a button that could toggle MeetingInfo between old and new
---


packages/node_modules/samples/browser-plugin-meetings/app.js
```diff
@@ -259,6 +264,13 @@ function leaveMeeting(meetingId) {
     });
 }
 
+function toggleMeetingInfo() {
+  const changeState = !webex.meetings.config.experimental.enableUnifiedMeetings;
```
lalli-flores, line 268: If you do this, the sample app page will always start off with the wrong API, instead of the default one. For instance, right now the default it's disabled, meaning "use v1 of meeting info" but when the app page gets loaded, it will use "v2 meeting info" even though it's not the default API. 
The same thing will happen when we enable v2 by default, it will put user on the wrong version

lalli-flores (commented):
Code looks good, just an issue with the sample app.

Please use rebase/edit the sample commit to address feedback to keep commits properly. Also, sample app changes / documentation / README updates should use `docs` commit type

packages/node_modules/samples/browser-plugin-meetings/app.js
mkesavan13: Hi @lalli-flores 

This method is called only upon clicking the button `webex.meetings._toggleMeetingInfo()` and isn't called on load. 

In the samples app, it's enabled by default on webex.init. Kindly check [here](https://github.com/webex/webex-js-sdk/blob/b7ff9b8699564ea1dabfffa213479c414a003cec/packages/node_modules/samples/browser-plugin-meetings/app.js#L66). Therefore, it would take users to the right API as per the init. 

Kindly let me know if I'm on the same page as you are.

lalli-flores: @mkesavan13, you can try it yourself by going to a temporary link generated by this PR:
https://pr-2259.d3m3l2kee0btzx.amplifyapp.com/browser-plugin-meetings/

mkesavan13: @lalli-flores - I just checked. I think I understood your concern. Although by default the SDK has disabled unified meetings, the sample app enabled it by default during init.

Please check here -> [packages/node_modules/samples/browser-plugin-meetings/app.js#L55](https://github.com/webex/webex-js-sdk/blob/b7ff9b8699564ea1dabfffa213479c414a003cec/packages/node_modules/samples/browser-plugin-meetings/app.js#L55)

That's why, by default we have V2 init. The method I've written toggles between enable and disable based on what state it is currently in. Would you want me to change the init to disableUnifiedMeetings?

lalli-flores: @mkesavan13, thanks for pointing it out, I didn't realize that the flag was enabled for the app

mkesavan13: Ok @lalli-flores 

Will do that.

lalli-flores:
Merging with the following know tests failures:

### Integration tests

- Successful 1:1 meeting (including Guest) bob stops sharing
- Successful 1:1 meeting (including Guest) bob steals the share from alice with desktop share 

Both failing due to the event they were waiting for timed out. These tests are sometimes flaky and fail from time to time, very likely due to network conditions of running tests on VMs

### Samples

Tests for screen sharing & call end fail, theses code changes are unrelated to such tests. They also fail on master, so need to be addressed separately.

packages/node_modules/samples/browser-plugin-meetings/app.js
mkesavan13: https://github.com/webex/webex-js-sdk/pull/2270

Here's the PR for disabling Unified meetings by default in sample app @lalli-flores

PR #2262, opened 2022-01-11
fix(plugin-meetings): allow calling join() again after IntentToJoinError

marcin-bazyl:
When Locus returns one of the errors from the INTENT_TO_JOIN array, the web app is supposed to call join() again with the right pin and moderator flag set. However, this is failing because the meeting state machine is in error state and doesn't allow to call join() in that state.

Fixes #SPARK-290105


marcin-bazyl:
closing this draft, instead I've opened a new SDK jira for this issue and created a new PR: https://github.com/webex/webex-js-sdk/pull/2266 with a better fix

PR #2303, opened 2022-03-14
fix(plugin-meetings): fix first locus message about a meeting is type locus.difference

chburket:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-305669
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-311026

This PR fixes the behavior of plugin-meetings when the first locus message about a meeting is a locus.difference message. Previously in this case, the parser.workingCopy was not being initialized from the first locus.difference message, which caused some meeting update messages to not be handled properly.

From Ask Locus space:  "You should never rely on the event type.  Pretty much any event should be treated as a locus.difference event since it can contain many changes, not just a single one.  My guess is this is related to a change where we now send difference events from the beginning of a meeting instead of waiting until the # of people cross a threshold."

This defect manifested itself in the web client by not removing meeting list items after the end time had passed, and also after the meeting was deleted.

I added a test for this but the plugin-meetings/test/unit/spec/meetings/index.js suite is currently disabled via skipInBrowser, since this suite is only ever tested in the browser. I unskipped, ran the tests, and reskipped (there are a lot of other failures).

I've also manually tested the effects of these through the web client.

< DESCRIBE YOUR CHANGES >

There was logic in handleLocusEvent() that would call meeting.locusInfo.initialSetup() with locus and a meeting or just with locus data, depending on the eventType. I changed this to always call initialSetup(data.locus) when the meeting is new, so that parser.onFullLocus() gets called (which sets workingCopy variable).

I also added an additional assert to two tests and added a variation of the tests, even though these tests are not currently being exectuted due to a skipInBrowser.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manual: See JIRA's for steps to reproduce

Automated:

npm test -- --packages @webex/plugin-meetings

(a few tests were failing before I changed anything; I did not try to fix those)

### I certified that

---


packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -253,14 +253,8 @@ export default class Meetings extends WebexPlugin {
...
-          if (data.eventType === LOCUSEVENT.DIFFERENCE) {
-            // its a delta object and we have a new meeting
-            meeting.locusInfo.initialSetup(data.locus, meeting);
-          }
-          else {
-            // Its a new meeting and have a fresh locus object
-            meeting.locusInfo.initialSetup(data.locus);
-          }
+          // Its a new meeting so initialize with locus data
+          meeting.locusInfo.initialSetup(data.locus);
```
Coread, line 257: After this change the `initialSetup` method in locus-info is never called with a second "meeting" argument. Should the code that handles this be removed from locusInfo?

Coread (approved):

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
arun3528: We will need to pass the meeting object i guess as the logic for delta event is different then initialSetup 

Can you test a meeting with more then 10 or 15 participant , We will see issues with members list there as the delta is like appending and initial-setup is fully replacing it 

You can may be set the first event to initialSetup and reset of the event to delta and that should work

arun3528 (changes requested):
Need to test the large meeting scenario

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
chburket: @Coread Good point. Will simplify intialSetup().

@arun3528 But initialSetup() should only be called in the case where we do not find a matching meeting object, right?  So only the first time, then onDeltaEvent() for the rest.

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -159,21 +159,13 @@ export default class LocusInfo extends EventsScope {
 
   /**
    * @param {Object} locus
-   * @param {Meeting} meeting
    * @returns {undefined}
    * @memberof LocusInfo
    */
-  initialSetup(locus, meeting) {
```
chburket, line 166: As Colin pointed out, this method is only being called without a meeting param now, so I removed the extra logic.

arun3528: @chburket  let have a 1:1 on this i feel that we will miss out something so let make sure we test the large meeting scenario

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
arun3528: we are assuming that the first event will always be a full event

chburket: The first event should be a full event if syncMeetings() does its job (which is another issue).  Even if something goes wrong and syncMeetings does not return a full locus for a meeting, then when the first locus difference event arrives, handleLocusEvent() will create a meeting object with whatever locus info is there (the same as it did before).  The only difference will be that initialSetup() is called with the partial locus info.  I think that's fine, and at the very least, better than what happened before.

I was able to get 12 aMTbot participants to join a scheduled space meeting (according to aMTbot), though I only every saw evidence of 6 "waiting to join" in the web client. I did see mercury events for the participants and getAllMeetings()[0].members.membersCollection.members shows them all.  Not sure if that is enough to address your request.

arun3528:
@chburket Any update on this ticket

chburket:
> @chburket Any update on this ticket

Yes, I replied to the first conversation:

"The first event should be a full event if syncMeetings() does its job (which is another issue). Even if something goes wrong and syncMeetings does not return a full locus for a meeting, then when the first locus difference event arrives, handleLocusEvent() will create a meeting object with whatever locus info is there (the same as it did before). The only difference will be that initialSetup() is called with the partial locus info. I think that's fine, and at the very least, better than what happened before.

I was able to get 12 aMTbot participants to join a scheduled space meeting (according to aMTbot), though I only every saw evidence of 6 "waiting to join" in the web client. I did see mercury events for the participants and getAllMeetings()[0].members.membersCollection.members shows them all. Not sure if that is enough to address your request."

arun3528 (approved):

PR #2340, opened 2022-04-19
Use type for verifying meeting password

Coread:
# COMPLETES # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-323395

Prior to this change the destination type was not being stored on the meeting object.
If the destination type isn't stored on the meeting object, then it cannot be used to correctly fetchMeetingInfo.
Verifying password/captcha was not possible on meetings with a destination type other than SIP_URI

1. Store new parameter destinationType on meeting object if it is passed into the constructor
2. Store the destination on the meeting object
3. Pass destination and destinationType in from meetings
4. Use meeting.destinationType and meeting.destination when fetching meeting info. This means verifyPassword doesn't need pass in destination or destinationType

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -979,7 +978,7 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   verifyPassword(password, captchaCode) {
     return this.fetchMeetingInfo({
-      destination: this.sipUri, type: _SIP_URI_, password, captchaCode
+      destination: this.sipUri, type: this.type, password, captchaCode
```
arun3528, line 981: Thanks for making this changes so going forward for type conversations also the verifyPassword will work

arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -592,7 +591,7 @@ export default class Meeting extends StatelessWebexPlugin {
      * @public
      * @memberof Meeting
      */
-    this.type = null;
+    this.type = attrs.type;
```
marcin-bazyl, line 594: this seems wrong. Currently the type field is either "meeting" or "call" and it's checked against these values in many places, see for example this line here: https://github.com/webex/webex-js-sdk/blob/6ac03087c2156df9fb1f0b95b2089ac131b2f5e3/packages/node_modules/%40webex/plugin-meetings/src/meeting/index.js#L4140 
this change will now break it.
If we need to store the destination type, it should be stored in a new field with appropriate name like `destinationType`

Coread: Changed to add new field

PR #2344, opened 2022-04-21
feat(spark-311364): add random delay to fetchMeetingInfo

nataliadelmar:
## DESCRIBE THE CONTEXT OF THE ISSUE

When a meeting is about to start in 5 minutes, we receive information about it from Locus and SDK creates a meeting object. At this time SDK fetches the meeting info for that object. This will cause a spike of traffic to the wbxappapi service - they have seen this problem with UCF and it has caused an outage of wbxappapi service on multiple occasions.

WbxAppAPI team have asked us to put a random delay for the call to meetinginfo api for OBTPs to avoid the spike in traffic. They've already agreed this solution with UCF team and UCF have already done this implementation - they add a random delay (up to 2-3 minutes) or fetch the meeting info immediately if the user presses the OBTP before the random delay is finished.

## DESCRIBE YOUR CHANGES

1. Change createMeeting() so that it doesn't call meeting.fetchMeetingInfo() immediately if createMeeting() was called as a result of a new meeting notification from Locus (see 
handleLocusEvent()) for a scheduled meeting that hasn't started yet
2. Make sure that some basic properties like sip url, meeting number, conversationUrl, locusUrl, owner are taken from the Locus DTO that came over Mercury connection as we now won't have the meeting info from which we normally take them 
3. Choose a random time between "now" and up to "now + 2 minutes" or "meeting scheduled start time" (whichever is sooner) and fetch meeting info at that random time
4. Allow the web app to trigger fetching meeting info immediately when needed (web app will call it when user presses the OBTP button before the random timer expires) - there is already existing meeting method called fetchMeetingInfo() that does this, but it's marked as "private"
5. Fetch the meeting info if it wasn't fetched before and meeting.join() was called.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -906,6 +906,25 @@ export default class Meeting extends StatelessWebexPlugin {
     this.hasJoinedOnce = false;
   }
 
+  /**
+   * Fetches meeting information for public use.
+   * @public
+   * @memberof Meeting
+   * @returns {Promise}
+   */
+  fetchMeetingInfoForMeeting() {
```
arun3528, line 915: There is already one on the webex.meetings object , I don't think we need one here also

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -804,9 +808,27 @@ export default class Meetings extends WebexPlugin {
       );
 
       this.meetingCollection.set(meeting);
+      const isMeetingActive = destination.fullState.active;
+      const syncMeetingsIsSource = (source === 'syncMeetings');
```
arun3528, line 812: whats the intention of source= syncMeeting here

arun3528 (changes requested):
Need to remove the changes with package json , skip min files in your commit

marcin-bazyl:
I think the "Breaking change" tick box should be ticked here, because we're changing API behaviour here - for some random period of time the meeting info will be unavailable while previously it was always available and now clients need to listen on the new event to know when the info is available. @arun3528 is there some documentation that we also need to update because of this?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: @arun3528 the one on webex.meetings object would not update the meeting info in the correct meeting instance, so Cantina needs to call fetchMeetingInfo on the meeting object and initially when Natalia started working on this the existing fetchMeetingInfo() was taking destination and type as arguments, so she needed a version that doesn't need these args.
However, now after recent Colin's changes the existing `meeting.fetchMeetingInfo()` no longer takes destination and type as arguments, so @nataliadelmar we don't need this new method anymore, you can just move the logic from it into the existing `meeting.fetchMeetingInfo()` method

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -804,9 +808,27 @@ export default class Meetings extends WebexPlugin {
...
 
       try {
-        await meeting.fetchMeetingInfo({});
+        if (!isMeetingActive && !syncMeetingsIsSource) {
```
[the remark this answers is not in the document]
marcin-bazyl: yeah I agree, I think we should check that the source is mercury rather than that it's not syncMeetings

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2388,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  parseMeetingInfo(meetingInfo, destination = null) {
     const webexMeetingInfo = meetingInfo?.body;
+    const locusMeetingObject = destination;
 
     // MeetingInfo will be undefined for 1:1 calls
     if (webexMeetingInfo && !(meetingInfo.errors && meetingInfo.errors.length > 0)) {
-      this.conversationUrl = webexMeetingInfo.conversationUrl || this.conversationUrl;
-      this.locusUrl = webexMeetingInfo.locusUrl || this.locusUrl;
-      this.setSipUri(this.config.experimental.enableUnifiedMeetings ? webexMeetingInfo.sipUrl : webexMeetingInfo.sipMeetingUri || this.sipUri);
+      this.conversationUrl = locusMeetingObject.conversationUrl || webexMeetingInfo.conversationUrl || this.conversationUrl;
```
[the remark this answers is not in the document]
nataliadelmar: This comes from this part of the Jira Ticket that asks to get try to get the info we can from the object we get from Locus Mercury. https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-311364#:~:text=2.%20Make%20sure%20that%20some%20basic%20properties%20like%20sip%20url%2C%20meeting%20number%2C%20conversationUrl%2C%20locusUrl%2C%20owner%20are%20taken%20from%20the%20Locus%20DTO%20that%20came%20over%20Mercury%20connection%20as%20we%20now%20won%27t%20have%20the%20meeting%20info%20from%20which%20we%20normally%20take%20them%C2%A0

@marcin-bazyl

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2388,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      this.locusUrl = locusMeetingObject.url || webexMeetingInfo.locusUrl || this.locusUrl;
```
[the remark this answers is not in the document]
nataliadelmar: same here. This comes from this part of the Jira Ticket that asks to get try to get the info we can from the object we get from Locus Mercury. https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-311364#:~:text=2.%20Make%20sure%20that%20some%20basic%20properties%20like%20sip%20url%2C%20meeting%20number%2C%20conversationUrl%2C%20locusUrl%2C%20owner%20are%20taken%20from%20the%20Locus%20DTO%20that%20came%20over%20Mercury%20connection%20as%20we%20now%20won%27t%20have%20the%20meeting%20info%20from%20which%20we%20normally%20take%20them%C2%A0

@marcin-bazyl

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2388,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      this.setSipUri(this.config.experimental.enableUnifiedMeetings ? webexMeetingInfo.sipUrl : locusMeetingObject.info.sipUri || webexMeetingInfo.sipMeetingUri || this.sipUri);
       if (this.config.experimental.enableUnifiedMeetings) {
-        this.meetingNumber = webexMeetingInfo.meetingNumber;
+        this.meetingNumber = locusMeetingObject.info.webExMeetingId || webexMeetingInfo.meetingNumber;
         this.meetingJoinUrl = webexMeetingInfo.meetingJoinUrl;
       }
-      this.owner = webexMeetingInfo.owner || webexMeetingInfo.hostId || this.owner;
+      this.owner = locusMeetingObject.info.owner || webexMeetingInfo.owner || webexMeetingInfo.hostId || this.owner;
```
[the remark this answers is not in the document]
nataliadelmar: would it help to include a optional chaining operator?
``` this.owner = locusMeetingObject?.info.owner ||  ... ```

File: packages/node_modules/@webex/plugin-meetings/src/meetings/index.js (hunk 2 above)
Comment (nataliadelmar) [reply packages/node_modules/@webex/plugin-meetings/src/meetings/index.js:812]: it is to avoid running the random timer if the meeting bricklet appearance is triggered by syncMeetings (which is when the user loads the page). We just want to add the timer to those meetings that appear as a product of handleLocusMercury (meetings starting in the next 5 mins) when the page is already loaded.
I have change this so that handleLocusMercury sends source='handleLocusMercury'), to make sure only in the above case, the timer runs.

File: packages/node_modules/@webex/plugin-meetings/src/meeting/index.js [hunk 7]
```diff
@@ -918,6 +937,7 @@ export default class Meeting extends StatelessWebexPlugin {
   async fetchMeetingInfo({
     password = null, captchaCode = null
   }) {
+    this.fetchMeetingInfoCalled = true;
```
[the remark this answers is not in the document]
nataliadelmar: gotcha... will add that. 
What if fetchMeetingInfo was called at some point and later, for from another source, it tries to be called again. Shouldn't we keep a flag to avoid fetching the same info more than once?

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2388,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
    * @param {Object} meetingInfo.body.owner
+   * @param {Object} destination locus object with meeting data
    * @returns {undefined}
    * @private
    * @memberof Meeting
    */
-  parseMeetingInfo(meetingInfo) {
+  parseMeetingInfo(meetingInfo, destination = null) {
     const webexMeetingInfo = meetingInfo?.body;
+    const locusMeetingObject = destination;
```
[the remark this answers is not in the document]
nataliadelmar: mmm what if we check if destination is an object. If it is not, we simply have an undefined locusMeetingObject, that will be a falsie and, thus, evaluate the webexMeetingInfo object alternatively
Like in
this.conversationUrl = locusMeetingObject?.conversationUrl || webexMeetingInfo.conversationUrl || this.conversationUrl;

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2392,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
     const webexMeetingInfo = meetingInfo?.body;
+    // We try to use as much info from Locus meeting object, stored in destination
+    const locusMeetingObject = typeof destination === 'object' ? destination : undefined;
 
     // MeetingInfo will be undefined for 1:1 calls
     if (webexMeetingInfo && !(meetingInfo.errors && meetingInfo.errors.length > 0)) {
-      this.conversationUrl = webexMeetingInfo.conversationUrl || this.conversationUrl;
-      this.locusUrl = webexMeetingInfo.locusUrl || this.locusUrl;
-      this.setSipUri(this.config.experimental.enableUnifiedMeetings ? webexMeetingInfo.sipUrl : webexMeetingInfo.sipMeetingUri || this.sipUri);
+      this.conversationUrl = locusMeetingObject?.conversationUrl || webexMeetingInfo.conversationUrl || this.conversationUrl;
```
arun3528, line 2407: Do we have to use both locus.info and meetingInfo or can we skip one as the server are converting !?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -804,9 +806,27 @@ export default class Meetings extends WebexPlugin {
       );
 
       this.meetingCollection.set(meeting);
+      const isMeetingActive = destination.fullState.active;
+      const handleLocusMercuryIsSource = (source === 'handleLocusMercury');
```
arun3528, line 810: Just a suggestion rather then having the function name as a key to decide if we need time cant we just pass a boolean saying useRandomDelayForInfo or something name 

any thoughts !?

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -804,9 +806,27 @@ export default class Meetings extends WebexPlugin {
...
 
       try {
-        await meeting.fetchMeetingInfo({});
+        // if no participant has joined the scheduled meeting (meaning meeting is not active) and we still get a locusEvent,
+        // it means the meeting will start in 5-6 min. In that case, we want to fetchMeetingInfo
+        // between 5 and 2 min (random between 3 minutes) before the meeting starts
+        // to avoid a spike in traffic to the wbxappi service
+        const {startTime} = destination.meeting;
```
arun3528, line 817: Have we tried to see what happens when its a 1:1 call or adhoc meeting do they have startTime all the time

arun3528 (commented):
Over all looks good

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
arun3528: Should we change this to boolean rather then the name 'handleLocusMercury'

@marcin-bazyl

arun3528:
@nataliadelmar  here is the change log https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md

Check for how to add breaking change message

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -654,6 +654,23 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    this.setTimeoutFetchMeetingInfoId = null;
+    /**
+     * @instance
+     * @type {String}
+     * @readonly
+     * @public
+     * @memberof Meeting
+     */
+
+    this.isFetchingMeetingInfo = false;
```
marcin-bazyl, line 666: this is not being used now in SDK, so we can remove it

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -981,6 +1013,10 @@ export default class Meeting extends StatelessWebexPlugin {
         throw (err);
       }
     }
+    finally {
+      this.isFetchingMeetingInfo = false;
+      console.log('fetchMeetingInfo Finshed!');
```
marcin-bazyl, line 1018: console log

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -2356,23 +2392,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    const locusMeetingObject = typeof destination === 'object' ? destination : undefined;
 
     // MeetingInfo will be undefined for 1:1 calls
     if (webexMeetingInfo && !(meetingInfo.errors && meetingInfo.errors.length > 0)) {
-      this.conversationUrl = webexMeetingInfo.conversationUrl || this.conversationUrl;
-      this.locusUrl = webexMeetingInfo.locusUrl || this.locusUrl;
-      this.setSipUri(this.config.experimental.enableUnifiedMeetings ? webexMeetingInfo.sipUrl : webexMeetingInfo.sipMeetingUri || this.sipUri);
+      this.conversationUrl = locusMeetingObject?.conversationUrl || webexMeetingInfo.conversationUrl || this.conversationUrl;
+      this.locusUrl = locusMeetingObject?.url || webexMeetingInfo.locusUrl || this.locusUrl;
+      this.setSipUri(this.config.experimental.enableUnifiedMeetings ? webexMeetingInfo.sipUrl : locusMeetingObject?.info.sipUri || webexMeetingInfo.sipMeetingUri || this.sipUri);
```
marcin-bazyl, line 2409: we need to do this for the case when enableUnifiedMeetings is true too

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -804,9 +806,27 @@ export default class Meetings extends WebexPlugin {
...
+        const startTimeDate = new Date(startTime);
+        const startTimeDatestamp = startTimeDate.getTime();
+        const timeToStart = startTimeDatestamp - Date.now();
+        const maxWaitingTime = Math.max(Math.min(timeToStart, 3 * 60 * 1000), 0);
+        const waitingTime = Math.round(Math.random() * maxWaitingTime);
+
+        if (!isMeetingActive && handleLocusMercuryIsSource && waitingTime > 0) {
```
marcin-bazyl, line 824: we need to check also this.config.experimenta.enableUnifiedMeetings

packages/node_modules/@webex/plugin-meetings/src/meetings/index.js
```diff
@@ -803,7 +805,34 @@ export default class Meetings extends WebexPlugin {
...
+      }
+      let isMeetingActive = false;
+
+      if (destination.fullState) {
+        isMeetingActive = destination.fullState.active;
+      }
+      const {enableUnifiedMeetings} = this.config.experimental;
+
+      if (enableUnifiedMeetings && !isMeetingActive && useRandomDelayForInfo && waitingTime > 0) {
+        meeting.setTimeoutFetchMeetingInfoId = setTimeout(() => meeting.fetchMeetingInfo({}), waitingTime);
```
marcin-bazyl, line 831: I think here was also need to call `meeting.parseMeetingInfo()` so that some basic fields like locusUrl, sipUri get populated (point 2 from the jira description)

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -789,7 +796,41 @@ describe('plugin-meetings', () => {
...
+                active: false
+              }
+            };
+
+            const meeting = await webex.meetings.createMeeting(fakeDestination, 'test type', true);
+
+            assert.instanceOf(meeting, Meeting, 'createMeeting should eventually resolve to a Meeting Object');
+            assert.notCalled(webex.meetings.meetingInfo.fetchMeetingInfo);
+            assert.calledOnce(setTimeoutSpy);
+            clock.tick(FAKE_TIME_TO_START);
```
marcin-bazyl, line 820: before this, we should verify that things like locusUrl, sipUri, etc are set correctly

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -789,7 +796,41 @@ describe('plugin-meetings', () => {
...
+            assert.calledOnce(webex.meetings.meetingInfo.fetchMeetingInfo);
+            assert.calledOnce(MeetingsUtil.getMeetingAddedType);
+            assert.calledTwice(TriggerProxy.trigger);
+            assert.calledWith(webex.meetings.meetingInfo.fetchMeetingInfo, fakeDestination, 'test type');
+            assert.calledWith(MeetingsUtil.getMeetingAddedType, 'test type');
+            assert.calledWith(TriggerProxy.trigger, sinon.match.instanceOf(Meetings), {
```
marcin-bazyl, line 826: this notification should have been sent already before the timer expiry, right? so we should move that check before the call to `clock.tick()` and here instead we should check the "meeting:meetingInfoAvailable" event being sent

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -646,6 +646,18 @@ export default class Meeting extends StatelessWebexPlugin {
      * @memberof Meeting
      */
     this.mediaConnections = null;
+
+    /**
+     * Fetching meeting info can be done randomly 2-5 mins before meeting start
+     * In case it is done before the timer expires, this timeout id is reset to cancel the timer.
+     * @instance
+     * @type {String}
```
marcin-bazyl, line 654: this is not a string, the timeout id is an integer as far as I know

packages/node_modules/@webex/plugin-meetings/src/constants.js
```diff
@@ -40,6 +40,7 @@ export const LOCAL = 'local';
 export const LOCI = 'loci';
 export const LOCUS_URL = 'locusUrl';
 
+export const MAX_RANDOM_RELAY_FOR_MEETING_INFO = 3 * 60 * 1000;
```
marcin-bazyl, line 43: typo: RELAY -> DELAY

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
marcin-bazyl: I still don't see the check for "meeting:meetingInfoAvailable" event - it's not tested anywhere

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -789,7 +798,70 @@ describe('plugin-meetings', () => {
               meeting: sinon.match.instanceOf(Meeting), type: 'test meeting added type'
             });
           });
+
+          it('creates the meeting from a successful meeting info fetch with random delay', async () => {
```
marcin-bazyl, line 802: we should also test that we don't do the random delay in these cases:
- when enableUnifiedMeetings is disabled
- when meeting is already active
- when meeting start time is in the past

marcin-bazyl (approved):

arun3528 (approved):
Thanks for adding a random delay feature , Congrats for your 1st PR

PR #2350, opened 2022-05-10
build: allow node 14

robstax:


arun3528 (changes requested):
Can we do something like "node": ">=14.x<=16.x"

same for npm >= 6.x

arun3528 (approved):

PR #2363, opened 2022-06-08
Imc chat events

robstax:
1. added support for incoming meeting chat activity events
2. do not encrypt payload sent to meeting-container service to create convo (check `created: 'True'`)

### DIFF
--- a/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
+++ b/packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
@@ -1078,6 +1078,16 @@ const Conversation = WebexPlugin.extend({
   },
 
   /**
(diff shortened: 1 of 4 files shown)


Coread (changes requested):
Would be great to have some tests for these

robstax:
> Would be great to have some tests for these

agreed... unfortunately the test coverage for these files and functions is already *really* bad. as far as i can tell the existing `processActivity` (which `processInmeetingchatEvent copies) is essentially untested.
`encryptActivity` is also currently untested.
all that being said, i'll work on adding tests for the cases i've implemented.

InteractiveTimmy (approved):
LGTM!

Coread (approved):

PR #2370, opened 2022-07-05
fix(meetings): loggerProxy error

rajeshtezu90:
# COMPLETES #< [SPARK-263135](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-263135) >

While using LoggerProxy to log different levels of messages, the logger was being called even before being set in the LoggerProxy causing to throw error especially while running the code in node.js runtime instead of browser.

The check for H256 codec and initial metrics setup is done along with Logger setup when webex is in READY state, which were causing the error. 

Tested with hookbuster as well.

LoggerProxy

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


Neeraj-swarnkar (approved):
Looks Good to me.

arun3528 (approved):
Looks good any specific reason why we moved the h264 and metrics logic in onready ?

rajeshtezu90:
> Looks good any specific reason why we moved the h264 and metrics logic in onready ?

The meetings util `hasH264Codec()` tries to access `window.RTCPeerConnection()` which is only accessible in browser so, while trying to execute that code in node runtime it throws error and in catch block we are calling Logger, which by that time does not get initialized because it's initialization is done inside the `onReady()` which is async, as a result it throws error `NO LOGGER DEFINED`.

PR #2375, opened 2022-07-06
fix(internal-plugin-device): fixing the boolean check in device interceptor

sknth:
# COMPLETES - NA 

An issue with overwriting the cisco-device-url header that the webex-js-sdk sends out over the wire.
An attempt to overwrite the cisco-device-url header using the following snippet fails.
```
    return this.webex.request({
      uri: `${URL}`,
      method: HTTP_METHODS.DELETE,
      headers: {
        'cisco-device-url': deviceUrl,
      },
    });  
```
The webex-js-sdk doesn't comply. It sends out the url that is stored in the webex object (webex.internal.device.url).

The issue is located in the condition that checks to see if the cisco-device-url inside the headers object contains a non-empty string.

Added another '!' to the condition to evaluate the expression to a truthy/falsy.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Tested the expression manually and ensured that it evaluates to the correct value for different cisco-device-url values.
   a. 'cisco-device-url' : undefined
   b. 'cisco-device-url' : ''
   c. 'cisco-device-url' : null
   d. 'cisco-device-url' : 'url'
### I certified that

---


arun3528:
@InteractiveTimmy  not sure if there was a typo in the logic before can you check it once

packages/node_modules/@webex/internal-plugin-device/src/interceptors/device-url.js
```diff
@@ -31,7 +31,7 @@ export default class DeviceUrlInterceptor extends Interceptor {
     // Check if header is already set before moving forward
     if (
       !device.url ||
-      (headers && CISCO_DEVICE_URL in headers && !headers[CISCO_DEVICE_URL])
+      (headers && CISCO_DEVICE_URL in headers && !!headers[CISCO_DEVICE_URL])
```
InteractiveTimmy, line 34: if you could, please add a small test case for this that validates that it works correctly. It should be a unit test.

InteractiveTimmy (changes requested):
LGTM, just needs a test.

packages/node_modules/@webex/internal-plugin-device/src/interceptors/device-url.js
InteractiveTimmy: I'm realizing there is a test that already covers this, I will move this to approved.

InteractiveTimmy (approved):
LGTM! test coverage already exists.

PR #2381, opened 2022-07-14
BNR PF - Phase 1 (enableBNR & disableBNR APIs)

mkesavan13:
# COMPLETES 
- [SPARK-325519 - WebEx JS SDK - Enable BNR in meeting](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-325519)
- [SPARK-325520 - WebEx JS SDK - Disable BNR in meeting](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-325520)

## This pull request addresses

This pull request is phase 1 release of BNR where the enable & disable APIs

These are already merged changes from the following pull requests,
- #2373 
- #2379 

## by making the following changes

The change is under plugin-meetings/Meeting where the enableBNR & disableBNR uses WebRTC's Effects/BNR module APIs underneath

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #2383, opened 2022-07-15
BNR Phase 1 PR review comments

mkesavan13:
# COMPLETES 
- [SPARK-345985 - WebEx JS SDK - Incorporate PR comment on BNR Phase 1](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-345985)

The review comments on the pull request #2381 

- updateMedia call's receiveAudio is set from meeting.mediaDirections rather than hardcoding.
- BNR enabled audio track is disabled if meeting audio is muted by the user

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Ensured muted audio track is still muted upon enabling BNR

### I certified that

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -5756,8 +5756,8 @@ export default class Meeting extends StatelessWebexPlugin {
       if (typeof this.mediaProperties === 'undefined' || typeof this.mediaProperties.audioTrack === 'undefined') {
         throw new Error("Meeting doesn't have an audioTrack attached");
       }
-      this.mediaProperties.audioTrack = await WebRTCMedia.Effects.BNR.enableBNR(this.mediaProperties.audioTrack);
-      const audioStream = MediaUtil.createMediaStream([this.mediaProperties.audioTrack]);
+      const bnrEnabledTrack = await WebRTCMedia.Effects.BNR.enableBNR(this.mediaProperties.audioTrack);
+      const audioStream = MediaUtil.createMediaStream([bnrEnabledTrack]);
 
       LoggerProxy.logger.info('Meeting:index#enableBNR. BNR enabled track obtained from WebRTC & sent to updateAudio');
       await this.updateAudio({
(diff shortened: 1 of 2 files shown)


arun3528:
@mkesavan13  can we test this once , I am not sure what exactly happens when we pass a muted track to the BNR and what comes back

arun3528 (approved):

meta-dipanshu-sharma (approved):

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -5757,12 +5757,17 @@ export default class Meeting extends StatelessWebexPlugin {
         throw new Error("Meeting doesn't have an audioTrack attached");
       }
       this.mediaProperties.audioTrack = await WebRTCMedia.Effects.BNR.enableBNR(this.mediaProperties.audioTrack);
+
+      if (this.isAudioMuted()) {
+        this.mediaProperties.audioTrack.enabled = false;
```
marcin-bazyl, line 5762: This way of doing this is coupling BNR with muting logic. This function here should not need to know how the local muting of a track is actually implemented (BNR does not need to know that we're setting enabled=true when muting a track). Instead we should just call `this.audio.applyClientStateLocally(this);`
And actually the setting of `this.mediaProperties.audioTrack` directly in line 5759 is also not great, because it feels like we should instead be calling `setLocalAudioTrack()` which already takes care of the call to `applyClientStateLocally()` for us. Looking at this further, I can see that `setLocalAudioTrack()` is actually called from `updateAudio()` so it seems like none of this code here is needed and it should work all fine. The new audio track should just be passed into the call to `this.updateAudio()` without setting it on `this.mediaProperties.audioTrack`. Can you please check this?

Also we should have unit tests to check the scenario of enabling BNR after muteAudio() was called (and if we had these tests in the first place we would have realised that the code in updateAudio() is already doing what needs to be done for muting and this change is not needed).

arun3528 (changes requested):
lets check on what marcin added

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
mkesavan13: Thanks for the comment @marcin-bazyl 

I understand now. Will clean them up and also go through the meetings plugin for better understanding of what all APIs we have. Thank you. 

Will update here once I am done.

mkesavan13: Hi @marcin-bazyl 

These changes are done and respective test written. Kindly verify.

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -5756,13 +5756,13 @@ export default class Meeting extends StatelessWebexPlugin {
       if (typeof this.mediaProperties === 'undefined' || typeof this.mediaProperties.audioTrack === 'undefined') {
         throw new Error("Meeting doesn't have an audioTrack attached");
       }
-      this.mediaProperties.audioTrack = await WebRTCMedia.Effects.BNR.enableBNR(this.mediaProperties.audioTrack);
```
arun3528, line 5759: thank you

mkesavan13:
Closed this pull request as the changes & comments are incorporated as part of this PR -> #2396

PR #2385, opened 2022-07-19
Revert "use icegatheringstatechange for ice completed"

andriy-andreyev:
Reverting https://github.com/webex/webex-js-sdk/commit/ade28cd3e28d2810ebac0818bf773eae521bc56f

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/common/errors/webex-errors.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/common/errors/webex-errors.js
@@ -111,29 +111,10 @@ WebExMeetingsErrors[UserInLobbyError.CODE] = UserInLobbyError;
 class InvalidSdpError extends WebexMeetingsError {
   static CODE = 30201;
 
-  constructor(message) {
-    super(InvalidSdpError.CODE, message || 'iceConnection: sdp generated is invalid');
+  constructor() {
+    super(InvalidSdpError.CODE, 'iceConnection: sdp generated is invalid');
   }
 }
 
 export {InvalidSdpError};
 WebExMeetingsErrors[InvalidSdpError.CODE] = InvalidSdpError;
-
-
-/**
- * @class IceGatheringFailed
- * @classdesc Raised whenever ice gathering fails.
- * @extends WebexMeetingsError
- * @property {number} code - 30202
- * @property {string} message - 'user failed ice gathering check network/firewall'
- */
-class IceGatheringFailed extends WebexMeetingsError {
-  static CODE = 30202;
-
-  constructor() {
-    super(IceGatheringFailed.CODE, 'iceConnection: gethering ice candidate failed');
-  }
-}
-
-export {IceGatheringFailed};
-WebExMeetingsErrors[IceGatheringFailed.CODE] = IceGatheringFailed;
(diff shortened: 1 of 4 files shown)


arun3528:
@andriy-andreyev  what was the issue you are facing here ?

andriy-andreyev:
> @andriy-andreyev what was the issue you are facing here ?

Hello @arun3528 !

I apologize. I created this PR by mistake. I wanted to create PR for my fork.

The reason is an issue that appears from time to time on old versions iOS mobile platform. I tested calls on iOS 15.0.2 and 14.7.1. 1 of 10 calls fails.

Steps to reproduce:
Make a call between two of my apps that use SDK 2.6. One app runs in the browser on iOS 15.0.2.
Expected behavior:
Normal call
Actual behavior:
Looks like a call is created. Endpoints are connected. IN_MEETING status is for both apps. But no audio and video streams are present. Absolute silence.
Frequency:
10%

I tested calls on different SDK versions and find that it works fine on 1.153.1 versions and older. The problem appears on 1.153.2 and newer versions. I found a commit that makes a change between 1.153.1 and 1.153.2. I reverted it for 2.6 version and tested it. The problem is gone.

Maybe, it is a known issue. And you can advise some another right fix for it because I'm not sure that this revert is a good idea.

arun3528:
@andriy-andreyev Can you create a issues in the github and post this content there ?

Someone from the team will have a look at it ,  I would exactly check why it actually failed before reverting it .
Can you post the console and the har file for the network tab if possible

arun3528:
Correlation ID of the call also works

andriy-andreyev:
Hi @arun3528 !
Unfortunately, I cannot create an issue. It just redirects me to developer doc.
Can you please create an issue? I formatted text for issue creation.


SDK 2.6 was tested on iOS 15.0.2 and 14.7.1. 1 of 10 calls fails.

Steps to reproduce:
Make a call between two of my apps that use SDK 2.6. One app runs in the browser on iOS 15.0.2 (mobile).
Expected behavior:
Normal call
Actual behavior:
Looks like a call is created. Endpoints are connected. IN_MEETING status is for both apps. But no audio and video streams are present. Absolute silence.
Frequency:
10%

It was tested on different SDK versions and was found that 1.153.1 version and older works fine on iOS 15.0.2. The problem appears on SDK 1.153.2 and newer versions. There is a commit that makes a change between 1.153.1 and 1.153.2 (https://github.com/webex/webex-js-sdk/commit/ade28cd3e28d2810ebac0818bf773eae521bc56f). If revert it for SDK 2.6 version the problem is gone. So, reverting of commit https://github.com/webex/webex-js-sdk/commit/ade28cd3e28d2810ebac0818bf773eae521bc56f fixes this problem. 


P.S. I'll try to collect logs and attach them here.

andriy-andreyev:
@arun3528 
Also, I got one more issue.

DESTROY MEETING issue:
JS SDK 2.7.0
**Steps to reproduce**:
Init Webex sdk via OAuth on one side (PC) and via access token on another one (IPhone).
Create Webex room on access token logined side. Join to the room and invite another side there (email should be used to invite)
OAuth logined side receives "meeting:added" event and joins
**Expected behavior**:
All sides are connected to meeting, audio and video streams enabled, meeting works without crushing
**Actual behavior**:
In 15 seconds (approx.) after OAuth side meeting joining this side receives "DESTROY_MEETING" event (meeting.on('all') subscription) and stops receiving "members:update" events. At that moment access token logined side receives "members:update" event with members object where OAuth side status is 'NOT_IN_MEETING'
**Frequency**: 5-10% of calls

PR #2391, opened 2022-07-25
fix(internal-plugin-conversation): fix delete content to accommodate meeting containers

yjlo:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-336425

Changes the delete method on conversation to generate a KMS message when the activity contains a meeting container.

Conversation Plugin --> added conditional to handle meeting container activities
encryption transform --> no-op for meeting container activities, otherwise it stripes the KMS message from the request and the request fails

image

### Change Type

- [x] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/node_modules/@webex/internal-plugin-conversation/src/encryption-transforms.js
```diff
@@ -159,27 +166,25 @@ export const transforms = toArray('outbound', {
         key = get(activity, 'target.defaultActivityEncryptionKeyUrl');
       }
 
-      if (key) {
```
Coread, line 162: question: this section used to be guarded by an `if (key)`. Why did that need to be removed? Is there a danger for normal conversations now that this code will run when it wasn't meant to?

packages/node_modules/@webex/internal-plugin-conversation/src/encryption-transforms.js
```diff
@@ -151,6 +151,13 @@ export const transforms = toArray('outbound', {
 
   prepareActivityKmsMessage(ctx, key, activity) {
     if (activity.kmsMessage) {
+      if (!key && activity.verb === 'delete' && activity?.isMeetingContainerActivity === true) {
+        // Essentially no-op since we don't want to strip the KMS message
+        key = true;
```
Coread, line 156: question: why is key being set here and then returning instantly?

packages/node_modules/@webex/internal-plugin-conversation/src/encryption-transforms.js
```diff
@@ -151,6 +151,13 @@ export const transforms = toArray('outbound', {
 
   prepareActivityKmsMessage(ctx, key, activity) {
     if (activity.kmsMessage) {
+      if (!key && activity.verb === 'delete' && activity?.isMeetingContainerActivity === true) {
```
Coread, line 154: issue: this piece of logic should have a test in spec/encryption-transform.js

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -363,11 +363,23 @@ const Conversation = WebexPlugin.extend({
...
-    })
+      object: pick(object, 'id', 'url', 'objectType'),
+    };
+
+    // Deleting meeting container requires KMS message
+    if (object.object.objectType === 'meetingContainer') {
+      // It's building a string resourceUri + "/authorizations?authId=" + id, where resourceUri is the meeting container's KRO URL, and id is the conversation's KRO URL.
+      request.kmsMessage = {
+        method: 'delete',
+        uri: `${object.object.kmsResourceObjectUrl}/authorizations?${querystring.stringify({authId: convoWithUrl.kmsResourceObjectUrl})}`
```
Coread, line 377: question: Your comment mentions `resourceUri` (and the prepareActivityKmsMessage function uses both uri and resourceUri) but are creating `uri`. Are they both the meant to be same? Is one no longer required?

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
```diff
@@ -1004,6 +1016,10 @@ const Conversation = WebexPlugin.extend({
           actor: this.webex.internal.device.userId
         });
 
+        if (params?.isMeetingContainerActivity) {
```
Coread, line 1019: suggestion: One way to avoid having to add code here is to pass in an activity object with a prepare method. This is done for ShareActivity right now (which is quite a complex method) but it would be much more simple for the meeting container. 

I would try and avoid putting specific hooks for a specific message type in so something like `activity.dontRemoveKmsMessage` could be added in this prepare method if the activity verb was delete.

Then the code in encryption-transform just looks for that and doesn't have to check the verb or the flag for meeting container activity.

packages/node_modules/@webex/internal-plugin-conversation/src/encryption-transforms.js
```diff
@@ -151,6 +151,13 @@ export const transforms = toArray('outbound', {
 
   prepareActivityKmsMessage(ctx, key, activity) {
     if (activity.kmsMessage) {
```
Coread, line 153: issue: I undid the changes in this file and ran the unit tests and they still passed so something is not being covered here

Coread (changes requested):
please see comments and questions

packages/node_modules/@webex/internal-plugin-conversation/src/encryption-transforms.js
yjlo: Good catch I didn't want to delete it.

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
yjlo: I've seen it with either one, however for this case, it has to be uri

packages/node_modules/@webex/internal-plugin-conversation/src/conversation.js
yjlo: Sounds good

robstax (approved):

packages/node_modules/@webex/internal-plugin-conversation/test/unit/spec/conversation.js
```diff
@@ -707,5 +707,67 @@ describe('plugin-conversation', () => {
...
+          url: 'https://example.com/activity1',
+          object: {objectType: 'activity'}
+        },
+        {
+          object:
+          {objectType: 'activity'}
+        });
+
+        return request.then(({request}) => {
+          assert.isUndefined(request.kmsMessage);
```
Coread, line 743: question: Is it worth checking here that the `request.target.kmsResourceObjectUrl` hasn't been modified?

Coread (approved):

arun3528:
Can we rebase and squash the commit

arun3528 (commented):
Can we squash the commit can merge it once done

yjlo:
Redone with correct commit log --> https://github.com/webex/webex-js-sdk/pull/2474

PR #2400, opened 2022-08-03
fix: getting TURN TLS info from Linus

marcin-bazyl:
this is a proof of concept work for integration with Linus team when they work on TURN TLS on port 443.

We shouldn't merge it, we should wait for a better solution where we can get TURN info from them without having to send the the SDP offer.


packages/node_modules/@webex/plugin-meetings/src/media/index.js
```diff
@@ -119,6 +119,147 @@ Media.getLocalMedia = (options, config) => {
...
+  console.log('marcin6: remote SDP:', mc?.mediaConnection?.pc?.remoteDescription?.sdp);
+  const turnTlsInfo = mc.getTurnTlsInfo();
+
+  console.log(`marcin6: got tls info: ${JSON.stringify(turnTlsInfo)}`);
+
+  // cleanup
+  mc.close();
+  meeting.mediaProperties.unsetPeerConnection();
+
+  turnTlsInfo.url = 'turns:linus1.aintm-a-6.int.infra.webex.com:443?transport=tcp'; // todo temp hack;
```
marcin-bazyl, line 259: note, that we have a hardcoded Linus FQDN here

PR #2431, opened 2022-09-14
fix(plugin-meetings): fix some flaky integration tests

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-364475

6 flaky tests:
- "bob joins the meeting"
- "Bob and Alice addsMedia"
- 'alice adds chris as guest to 1:1 meeting'
- 'alice Leaves the meeting'
- 'bob and chris leave meeting'
- 'alice adds chris as guest to space meeting'

 "bob joins the meeting" - the test was starting to wait for expected events only after the join promise was resolved, so it meant that if the member update event arrived before join was resolved, the test would miss that event and sit there forever until it times out. Fixed it by making sure that the waitForEvents call is done from the start

"Bob and Alice addsMedia" - it's using addMedia() test util function, which calls addMedia and after that resolves it would always sit for 10 seconds doing nothing and after that check if expected events arrived. Sometimes that 10s wait would be enough to go over the overall test timeout limit. Improved the utility function so that now it doesn't wait for 10s, instead it just waits for the right events.

The rest of the tests where all doing the same mistake when waiting for a members update with a specific member having a specific status (like "IN_MEETING" or "NOT_IN_MEETING", etc): they were just waiting for the 1st update and checking that member status in only that 1st update. In real life, Locus may send multiple updates and the expected status may be received only in the 2nd or 3rd update, so I've changed these tests to use the `match` argument when calling `waitForEvents()` - it allows you to give a callback that's called for each event and the participant status can be checked in that callback, so now it will work even if the expected status arrives later, not in the 1st update event.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

plugin-meetings integration tests

### I certified that

---


Coread (approved):

packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -191,13 +191,34 @@ const delayedTest = (callback, timeout) => new Promise((resolve) => {
...
+
 const addMedia = (user) => {
-  const mediaReadyEvent = {};
+  const mediaReadyPromises = {
+    local: createControlledPromise(),
+    remoteAudio: createControlledPromise(),
+    remoteVideo: createControlledPromise(),
+  };
   const mediaReady = (media) => {
     if (!media) {
```
maxinteger, line 216: @marcin-bazyl 
just out of curiosity, is this a valid case when the `media:ready` callback gets an undefined media?

maxinteger (approved):

packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
marcin-bazyl: this check was already there so I didn't touch it, but yeah looking at the code that sends `media:ready` it's actually not possible to ever receive no data with the event, it will always have at least `media.type` defined

Coread (approved):

arun3528 (approved):
Thanks for fixing the flaky test

packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -191,13 +191,34 @@ const delayedTest = (callback, timeout) => new Promise((resolve) => {
   }, timeout);
 });
 
+// returns a promise that has resolve() and reject() methods
+// allowing you to resolve/reject it at any time you want
+const createControlledPromise = () => {
```
maxinteger, line 196: This usually called `Defer` and it is already defined in the common package `packages/node_modules/@webex/common/src/defer.js`

packages/node_modules/@webex/plugin-meetings/test/integration/spec/space-meeting.js
```diff
@@ -152,32 +153,22 @@ skipInNode(describe)('plugin-meetings', () => {
...
-      testUtils.waitForEvents([{scope: chris.meeting.members, event: 'members:update'}])
-        .then((response) => {
-          const bobParticipant = response[0].result.delta.updated.find((member) => bob.meeting.members.selfId === member.id);
+    it('bob and chris leave meeting', () =>
+      Promise.all([
+        testUtils.delayedPromise(bob.meeting.leave()),
+        testUtils.waitForEvents([{scope: chris.meeting.members, event: 'members:update', match: testUtils.checkParticipantUpdatedStatus(bob, 'NOT_IN_MEETING')}])
+      ]).then(() => testUtils.waitForStateChange(bob.meeting, 'LEFT'))
+        .then(() => {
+          console.log('CHRIS MEETING', chris.meeting.locusInfo);
```
maxinteger, line 167: Is this console.log necessary?

maxinteger (changes requested):
Added some minor comments, otherwise looks good

maxinteger (approved):

packages/node_modules/@webex/plugin-meetings/test/integration/spec/space-meeting.js
marcin-bazyl: it was there before..., I guess someone put it there thinking it could be useful when looking at logs of a failed test, but actually these logs are not shown in the log by default, so I've removed it now.

packages/node_modules/@webex/plugin-meetings/test/utils/testUtils.js
marcin-bazyl: ah, nice, I wasn't aware that we had that in the SDK. Thanks for pointing it out. I've now removed `createControlledPromise()` and changed the code to use `Defer`

PR #2446, opened 2022-09-27
test: integration test for video quality changes

rajeshtezu90:
# COMPLETES #< [SPARK-351412](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-351412) >

Adds integration tests for video quality changes for **chrome** browser.

**Note** : Setting ```'media.navigator.streams.fake': true``` for firefox is also not working and returning an empty object for `videoStreams.getSettings()`. [Link](https://bugzilla.mozilla.org/show_bug.cgi?id=1537986) to the bug, It's closed but I haven't found any solution. Please let me know if you have any suggestion.

### Screenshot

resolution

Local resolution is getting set to 1080p but remote resolution is not going beyond 720p even after using the fake stream.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


Neeraj-swarnkar (changes requested):
Raise your PR against FHD branch.

arun3528:
can you rebase

meta-dipanshu-sharma:
Have you checked by changing remote video resolution if it's working?

Neeraj-swarnkar:
@rajeshtezu90, We can add some tests for setRemoteQualityLevel , e.g. Bob changes the remote res and checks the incoming from 360 to 1080p. Changing remote video resolution should also have tests.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -298,6 +298,44 @@ skipInNode(describe)('plugin-meetings', () => {
         assert.exists(alice.meeting.members.selfId, 'selfId not present');
       });
 
+      it('alice resolution is 1080p, bob receives 1080p', () => {
+        testUtils.delayedPromise(alice.meeting.mediaProperties.setLocalQualityLevel('1080p'));
```
marcin-bazyl, line 302: I don't see how these tests can work at all, there are multiple issues with them:
1. All calls in all these tests need to be wrapped with `return Promise.all()` otherwise the test runner will think that the tests are completed before the code that is executed in them actually completes.
2. The return value of `testUtils.waitForEvents()` is ignored, so the test doesn't actually check if these events fired and also the checks inside the `then()` after waitForEvents are not actually executed.
3. The tests call meeting.mediaProperties.setLocalQualityLevel() which does nothing except for storing the value - see here: https://github.com/webex/webex-js-sdk/blob/master/packages/node_modules/%40webex/plugin-meetings/src/media/properties.js#L62 - they should instead call `meeting.setLocalVideoQuality()`

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -298,6 +298,44 @@ skipInNode(describe)('plugin-meetings', () => {
...
+
+        testUtils.waitForEvents([
+          {scope: alice.meeting, event: 'media:ready'}
```
marcin-bazyl, line 305: this check should verify also the event payload's type is "local" and the stream contains a track with correct settings and we should use a `match` argument for `waitForEvents()` to do these checks, otherwise this test can be flaky as it will fail if any other event is sent first.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: Yes, I realised that. Updated the tests.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: Actually we don't need to wait for this event as this gets triggered when we call `addMedia()`.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -706,6 +706,89 @@ skipInNode(describe)('plugin-meetings', () => {
...
+          });
+      });
+
+      it('bob changes remote to 480p, check what bob receives', (done) => {
+        testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('720p'))
+          .then(() => {
+            testUtils.waitUntil(2000);
+            testUtils.delayedPromise(bob.meeting.setRemoteQualityLevel('LOW'))
+              .then(() => {
+                testUtils.waitUntil(3000);
```
arun3528, line 764: Are we sure this time is sufficient for the action to happen !?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -706,6 +706,89 @@ skipInNode(describe)('plugin-meetings', () => {
           throw e;
         }));
 
+      it('alice resolution is 1080p, check what bob receives', (done) => {
+        testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('1080p'))
+          .then(() => {
+            testUtils.waitUntil(2000);
```
marcin-bazyl, line 712: what's the purpose of this call here? was it supposed to add a delay before the checks are done? It doesn't do that, because the return value is ignored

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -706,6 +706,89 @@ skipInNode(describe)('plugin-meetings', () => {
...
+            const remoteSettings = bob.meeting.mediaProperties.remoteVideoTrack.getSettings();
+            const localSettings = alice.meeting.mediaProperties.videoTrack.getSettings();
+
+            assert.equal(localSettings.height, '1080');
+            assert.equal(remoteSettings.height, '1080');
+            done();
+          }).catch((error) => {
```
marcin-bazyl, line 719: why are we catching the errors? this will cause the test to always pass even if something (like the asserts) throws, won't it?

marcin-bazyl (changes requested):

rajeshtezu90:
> Have you checked by changing remote video resolution if it's working?

As we tested the functionality, attached the resultant screenshot of test in the description.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: It was not, added an interval.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: My bad, removed. Used interval wrapped in a promise instead.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: Removed.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
           throw e;
         }));
 
+      // FHD - remote video tests - starts
+      it('alice resolution is 1080p, check what bob receives', (done) => {
+        // Firefox bug - https://bugzilla.mozilla.org/show_bug.cgi?id=1537986
+        if (getBrowserName() === 'Firefox') {
```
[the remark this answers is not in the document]
rajeshtezu90: Moved out the condition and kept all the FHD tests inside.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+        }
+
+        testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('1080p'))
+          .then(() => {
+            const heightPromise = new Promise((resolve, reject) => {
+              testInterval = setInterval(() => {
+                const remoteSettings = bob.meeting.mediaProperties.remoteVideoTrack.getSettings();
+                const height = remoteSettings.height && +remoteSettings.height;
+
+                if (height >= 700) {
```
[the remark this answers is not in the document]
rajeshtezu90: Actually should have added `height > 720` check here, if that condition meets it should resolve there itself else whatever value it reaches after allowed duration passes would be provided to assert as change in the remote resolution does not get affected immediately instead it changes gradually and sometimes it also gets stuck on some particular resolution due to environmental factors.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+                  else resolve(height);
+                }
+
+                count += 1;
+              }, WAIT_TIME);
+            });
+
+            heightPromise
+              .then((height) => {
+                assert.isAbove(height, 500);
```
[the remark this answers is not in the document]
rajeshtezu90: Addressed [here](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r985610199)

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+        }
+
+        testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('720p'))
+          .then(() => {
+            const heightPromise = new Promise((resolve, reject) => {
+              testInterval = setInterval(() => {
+                const remoteSettings = bob.meeting.mediaProperties.remoteVideoTrack.getSettings();
+                const height = remoteSettings.height && +remoteSettings.height;
+
+                if (height >= 600 && height <= 700) {
```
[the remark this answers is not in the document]
rajeshtezu90: Fixed.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+                      else resolve(height);
+                    }
+
+                    count += 1;
+                  }, WAIT_TIME);
+                });
+
+                heightPromise.then((height) => {
+                  assert.isAbove(height, 400);
+                  assert.isBelow(height, 721);
```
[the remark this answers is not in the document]
rajeshtezu90: Fixed.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+      it('bob upgrades remote to 720p, check what bob receives', (done) => {
+        if (getBrowserName() === 'Firefox') {
+          done();
+
+          return;
+        }
+
+        testUtils.delayedPromise(bob.meeting.setRemoteQualityLevel('MEDIUM'))
+          .then(() => {
+            const heightPromise = new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
rajeshtezu90: Agree, fixed the duplicate code. Used async/await.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,213 @@ skipInNode(describe)('plugin-meetings', () => {
...
+              testInterval = setInterval(() => {
+                const remoteSettings = bob.meeting.mediaProperties.remoteVideoTrack.getSettings();
+                const height = remoteSettings.height && +remoteSettings.height;
+
+                if (height >= 600 && height <= 700) {
```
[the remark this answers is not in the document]
rajeshtezu90: Fixed.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+
+        it('alice resolution is 1080p, check what bob receives', async () => {
+          await testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('1080p'));
+          const height = await getRemoteHeight(1080);
+
+          assert.isAbove(height, 500);
+          assert.isBelow(height, 1081);
+        });
+
+        it('bob requests 480p, should get resolution around 480p', async () => {
```
marcin-bazyl, line 332: Has this test ever passed for you? I see it's failing in CI and I don't think it will ever pass, because that functionality is broken, I think it can only pass by luck if the network conditions restrict the resolution - see a fix for this done in [this PR](https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/139) in the webrtc-media lib. 
BTW, I don't think we should be fixing this issue now in SDK, we should just wait for SDK v3 which uses the webrtc-media lib that has the fix and until then this test should be disabled.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+          await testUtils.delayedPromise(bob.meeting.setRemoteQualityLevel('HIGH'));
+          const height = await getRemoteHeight(721);
+
+          assert.isAbove(height, 719);
+          assert.isBelow(height, 1081);
+        });
+
+        it('alice resolution is 720p, check what bob receives', async () => {
+          await testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('720p'));
+          const height = await getRemoteHeight(600, 720);
```
marcin-bazyl, line 358: why the mismatch between the values here and in the asserts later?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+
+          assert.isAbove(height, 500);
+          assert.isBelow(height, 721);
+        });
+
+        it('alice downgrades to 480p, check what bob receives', async () => {
+          await testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('480p'));
+          const height = await getRemoteHeight(400, 480);
```
marcin-bazyl, line 366: why the mismatch between the values here and in the asserts later?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
           throw e;
         }));
 
+      // FHD - remote video tests - starts
```
marcin-bazyl, line 300: not sure what "starts" is supposed to mean here

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+
+            count += 1;
+          }, WAIT_TIME);
+        });
+
+        it('alice resolution is 1080p, check what bob receives', async () => {
+          await testUtils.delayedPromise(alice.meeting.setLocalVideoQuality('1080p'));
+          const height = await getRemoteHeight(1080);
+
+          assert.isAbove(height, 500);
```
marcin-bazyl, line 328: why 500 here?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+          await testUtils.delayedPromise(bob.meeting.setRemoteQualityLevel('LOW'));
+          const height = await getRemoteHeight(400, 480);
+
+          assert.isAbove(height, 400);
+          assert.isBelow(height, 481);
+        });
+
+        it('bob upgrades remote to 720p, should get resolution around 720', async () => {
+          await testUtils.delayedPromise(bob.meeting.setRemoteQualityLevel('MEDIUM'));
+          const height = await getRemoteHeight(720, 1079);
```
marcin-bazyl, line 342: why the mismatch between the values here and in the asserts later?

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
```diff
@@ -286,6 +297,80 @@ skipInNode(describe)('plugin-meetings', () => {
...
+      // FHD - remote video tests - starts
+      // Firefox bug - https://bugzilla.mozilla.org/show_bug.cgi?id=1537986
+      if (getBrowserName() !== 'Firefox') {
+        const getRemoteHeight = (minHeight = 360, maxHeight = 1080) => new Promise((resolve, reject) => {
+          testInterval = setInterval(() => {
+            const remoteSettings = bob.meeting.mediaProperties.remoteVideoTrack.getSettings();
+            const height = remoteSettings.height && +remoteSettings.height;
+
+            console.log(`Resolution for ${minHeight} ${maxHeight} : `, count, height);
+            if (height >= minHeight && height <= maxHeight) {
```
marcin-bazyl, line 309: maybe it would be better to change this to check to > and < so that then the arguments passed to getRemoteHeight() could be the same as later used in the asserts. Currently it looks a bit silly and messy that you have to manually add or remove 1 to the values when using the asserts and in many places you're actually using completely different values in the asserts which seems like a mistake, or if it's not a mistake then there should be comments explaining why the values are different

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: Skipped

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: The range provided here are preferred one, we may or may not get the value exactly in that range (Also mentioned [here](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r985610199)). It's to try to get values as close as possible but `getRemoteHeight()` is not guaranteed to return value specifically in that range all the time. console.log is added inside the getRemoteHeight(), you may increase the timeout in the config file and check the  log for longer time and observe value changes gradually and sometimes takes longer than expected.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: [Same reason](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r986433243)

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: This was for reference as file size grows and it becomes difficult to keep track. Removed.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: [Same reason](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r986433243)

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: [Same reason](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r986433243)

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
rajeshtezu90: Even if we use > and <, assert range will differ as mentioned [here](https://github.com/webex/webex-js-sdk/pull/2446#discussion_r986433243). Adding a comment to explain why the values are different.

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
marcin-bazyl: if it's skipped, it should have a comment pointing to a jira describing the issue, so that once that jira is resolved, the test can be enabled.
Having a skipped test with no comment and no jira means that it will be skipped forever, so it will be useless and we may as well just delete it

packages/node_modules/@webex/plugin-meetings/test/integration/spec/journey.js
marcin-bazyl: At the end of the day the test can either pass or fail, there is nothing else in-between. No one will look at any logs for a passing test, so if the test can pass with resolution not actually being changed to the more stringent checks inside the getRemoteHeight() I see no point in doing the different boundary values between the call to getRemoteHeight() and asserts after it. With the code as it is, if it fails to change the resolution, the test will still pass but it will take longer, because getRemoteHeight() will wait for the higher res, but assert will pass because it has less stringent condition - so that's pointless, we're just wasting time waiting for getRemoteHeight() to finish.

PR #2451, opened 2022-09-29
feat(plugin-meetings): added support for TURN discovery (under a config flag)

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-369983

Adding ability to use Linus as a TURN server for TURN TLS connections, so that we can have media flowing in networks behind firewalls that block all UDP and TCP apart from TLS on 443.

Using new Roap messages to get TURN information from Linus and configuring RTCPeerConnection with that TURN server information. Also removing xTLS candidates from remote SDP as Linus now sends them to us when it's pretending to be a TURN server.

All changes are behind an experimental config flag. Added also a checkbox for it to the sample app.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual e2e with a special Linus version

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -26,6 +26,7 @@ const registerElm = document.querySelector('#registration-register');
 const unregisterElm = document.querySelector('#registration-unregister');
 const registrationStatusElm = document.querySelector('#registration-status');
 const integrationEnv = document.getElementById('integration-env');
+const turnDiscoveryCheckbox = document.getElementById('enable-turn-discovery');
 const eventsList = document.getElementById('events-list');
 
 // Disable screenshare on join in Safari patch
@@ -73,6 +74,39 @@ function changeAuthType() {
   }
 }
 
+function generateWebexConfig({credentials}) {
+  return {
+    appName: 'sdk-samples',
+    appPlatform: 'testClient',
+    fedramp: fedRampInput.checked,
+    logger: {
+      level: 'debug'
+    },
+    ...(integrationEnv.checked && {
+      services: {
+        discovery: {
+          u2c: 'https://u2c-intb.ciscospark.com/u2c/api/v1',
+          hydra: 'https://apialpha.ciscospark.com/v1/'
+        }
+      }
+    }),
+    meetings: {
+      reconnection: {
+        enabled: true
+      },
+      enableRtx: true,
+      experimental: {
+        enableMediaNegotiatedEvent: false,
+        enableUnifiedMeetings: true,
+        enableAdhocMeetings: true,
+        enableTurnDiscovery: turnDiscoveryCheckbox.checked,
+      }
+    },
+    credentials,
+    // Any other sdk config we need
+  };
+}
+
 function initOauth() {
   let redirectUri = `${window.location.protocol}//${window.location.host}`;
 
@@ -81,40 +115,13 @@ function initOauth() {
   }
 
   webex = window.webex = Webex.init({
-    config: {
-      appName: 'sdk-samples',
-      appPlatform: 'testClient',
-      fedramp: fedRampInput.checked,
-      logger: {
-        level: 'debug'
-      },
-      ...(integrationEnv.checked && {
-        services: {
-          discovery: {
-            u2c: 'https://u2c-intb.ciscospark.com/u2c/api/v1',
-            hydra: 'https://apialpha.ciscospark.com/v1/'
-          }
-        }
-      }),
-      meetings: {
-        reconnection: {
-          enabled: true
-        },
-        enableRtx: true,
-        experimental: {
-          enableMediaNegotiatedEvent: false,
-          enableUnifiedMeetings: true,
-          enableAdhocMeetings: true
-        }
-
-      },
+    config: generateWebexConfig({
       credentials: {
         client_id: 'C7c3f1143a552d88d40b2afff87600c366c830850231597fb6c1c1e28a5110a4f',
         redirect_uri: redirectUri,
         scope: 'spark:all spark:kms'
       }
-      // Any other sdk config we need
-    }
+    })
   });
 
   webex.once('ready', () => {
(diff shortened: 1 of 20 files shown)


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -73,6 +74,39 @@ function changeAuthType() {
   }
 }
 
+function generateWebexConfig({credentials}) {
```
marcin-bazyl, line 77: created this function, because config was duplicated in 2 places. Apart from this change and the fact that we read the checkbox value for `enableTurnDiscovery` config entry, the rest of the changes in this file are caused by automatic formatting of the code.

packages/node_modules/@webex/plugin-meetings/src/common/logs/logger-proxy.js
```diff
@@ -1,13 +1,14 @@
...
 
 const LoggerProxy = {
   logger: {
-    info: () => { console.error('LoggerProxy->info#NO LOGGER DEFINED'); },
-    log: () => { console.error('LoggerProxy->log#NO LOGGER DEFINED'); },
-    error: () => { console.error('LoggerProxy->error#NO LOGGER DEFINED'); },
-    warn: () => { console.error('LoggerProxy->warn#NO LOGGER DEFINED'); },
-    trace: () => { console.error('LoggerProxy->trace#NO LOGGER DEFINED'); },
-    debug: () => { console.error('LoggerProxy->debug#NO LOGGER DEFINED'); }
+    info: (args) => { console.error('LoggerProxy->info#NO LOGGER DEFINED'); },
```
marcin-bazyl, line 6: I had to add the args because otherwise typescript was complaining when I was using the logger in the newly created ts file

packages/node_modules/@webex/plugin-meetings/src/constants.ts
```diff
@@ -754,9 +754,6 @@ export const PEER_CONNECTION_STATE = {
   FAILED: 'failed'
 };
 
-export const RTC_CONFIGURATION_FIREFOX = {iceServers: [], bundlePolicy: 'max-compat'};
```
arun3528, line 757: was this not getting used !?

packages/node_modules/@webex/plugin-meetings/src/media/util.js
```diff
@@ -2,19 +2,29 @@
...
+  const config = {iceServers: []};
+
+  if (turnServerInfo) {
+    config.iceServers.push({
+      urls: turnServerInfo.url,
+      username: turnServerInfo.username || '',
+      credential: turnServerInfo.password || ''
+    });
+  }
+  if (isBrowser('firefox')) {
```
arun3528, line 21: ok got it you have added here

packages/node_modules/@webex/plugin-meetings/src/roap/index.js
```diff
@@ -245,12 +256,17 @@ export default class Roap extends StatelessWebexPlugin {
       correlationId: meeting.correlationId
     });
 
+    // When reconnecting, it's important that the first roap message being sent out has empty media id.
+    // Normally this is the roap offer, but when TURN discovery is enabled,
+    // then this is the TURN discovery request message
+    const sendEmptyMediaId = reconnect && !meeting.config.experimental.enableTurnDiscovery;
```
arun3528, line 262: So now any kind of reconnect , either media release or inactive the client will not send media ID

packages/node_modules/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -0,0 +1,238 @@
+import {Defer} from '@webex/common';
```
arun3528, line 1: This just looks like a promise , Do you like the name Defer or we need to change it to WaitForPromise or something 

This code is good but is there a general pattern which we can use in the future !?

packages/node_modules/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -0,0 +1,238 @@
...
+
+        LoggerProxy.logger.info('Roap:turnDiscovery#doTurnDiscovery --> TURN discovery completed');
+
+        return this.turnInfo;
+      })
+      .catch((e) => {
+        // we catch any errors and resolve with no turn information so that the normal call join flow can continue without TURN
+        LoggerProxy.logger.info(`Roap:turnDiscovery#doTurnDiscovery --> TURN discovery failed, continuing without TURN: ${e}`);
+
+        Metrics.sendBehavioralMetric(
```
arun3528, line 225: Thanks for adding this

arun3528 (approved):
The code looks good few questions that it

arun3528:
FYI we will need ticket for usage on TLS443 and also any failures after the TUEN discovery is succesful

marcin-bazyl:
> FYI we will need ticket for usage on TLS443 and also any failures after the TUEN discovery is succesful

I've created https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-370008 for updating the 'js_sdk_add_media_success' metric so that we know if the connection is udp, tcp or turn tls. 
Yeah, I agree that it would be useful to also know if there are any users with TURN discovery enabled that still fail to establish any media connection. I think for that we can just update the existing 'js_sdk_add_media_failures' metric to include a flag whether TURN discovery was used or not. I can add this to this PR.

packages/node_modules/@webex/plugin-meetings/src/constants.ts
marcin-bazyl: it's moved into `MediaUtil.createPeerConnection()`

packages/node_modules/@webex/plugin-meetings/src/roap/index.js
marcin-bazyl: Yeah the assumption is that in all cases it will always go through either `ReconnectionManager.setupPeerConnection()` or a new call to `addMedia()` so we will do TURN discovery again. Is that a correct assumption?

packages/node_modules/@webex/plugin-meetings/src/roap/turnDiscovery.ts
marcin-bazyl: Yeah I personally prefer a name like "ControlledPromise", but "defer" seems to be a more commonly known name for this so maybe we should keep it. Defer is also sometimes referred to as an anti-pattern, mainly because of the issue if there is any error happening that prevents the `defer.resolve()`, then you might be stuck forever waiting for it. In this case we have a timeout in `waitForTurnDiscoveryResponse()` so it's not a problem, but it's something to be aware of when using Defer in general.
I think overall the more general pattern to be used in scenarios like this one is a state machine, but because here it's very simple - we only wait for one message, then I thought that just using a single defer here will make the code easier to read than having to define a state machine. It allows me to just write 
```
this.sendRoapTurnDiscoveryRequest(meeting, isReconnecting)
      .then(() => this.waitForTurnDiscoveryResponse())
      .then(() => this.sendRoapOK(meeting))
```
in `doTurnDiscovery()` and it's immediately clear how this whole turn discovery sequence works.

maxinteger (approved):
Look good to me, but I am not an expert of this topic :)

PR #2454, opened 2022-09-30
feat(meetings/meeting): add size-only layout requests

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add the ability to change the layout size without forcing a layout type.

## Links

[JIRA](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-370386).

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
@@ -5156,7 +5156,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
   /**
    * Sends request to change layout type for the current meeting for the specific participant/device only
-   * @param {String} layoutType a layout type that should be available in meeting constants {@link #layout_types}
+   * @param {String} [layoutType] a layout type that should be available in meeting constants {@link #layout_types}
    * @param {Object} renderInfo preferred dimensions for the remote main and content streams (server can ignore it)
    * @param {Object} renderInfo.main preferred dimensions for the remote main video stream
    * @param {Number} renderInfo.main.width preferred width of main video stream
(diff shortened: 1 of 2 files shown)


packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -5182,9 +5182,6 @@ export default class Meeting extends StatelessWebexPlugin {
     if (LAYOUT_TYPES.includes(layoutType)) {
       layoutInfo.layoutType = layoutType;
     }
-    else {
```
marcin-bazyl, line 5185: I think we should still fail if a non-undefined layoutType is passed in that doesn't match any valid value. We should maybe also update the docs for this function to say that layoutType is optional?

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1942,9 +1942,19 @@ describe('plugin-meetings', () => {
             });
           });
 
-          it('doesn\'t have layoutType which exists in the list of allowed layoutTypes should throw an error', async () => {
+          it('doesn\'t have layoutType which exists in the list of allowed layoutTypes should not assign for payload', async () => {
```
marcin-bazyl, line 1945: I think we should leave that test as it was as we don't want to just silently ignore "invalid layout" and instead add a new test that sets the layout to some value and then calls changeVideoLayout with layoutType being undefined and verify that we send the previous value to the server.

marcin-bazyl: might be also worth testing that the scenario where changeVideoLayout is called first time on a meeting object without layoutType - in that case we should use undefined value for it as there is no previous value available

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
InteractiveTimmy: I've updated the source to address this. layoutType is now optional in JSDoc. This conditional needs to stay due to a previously existing layoutInfo validation check later within this function.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: but wouldn't it be useful to have a test that checks the following:
if some non-undefined, valid layout is set and changeLayout() is called without a layout type, then verify that we send a request without layout being specified

packages/node_modules/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: ok, but now we silently ignore any invalid `layoutType` values?

InteractiveTimmy: I believe I understand now. I've made some changes, please re-review.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/meeting/index.js
InteractiveTimmy: I believe I understand now. I've made some changes, please re-review.

marcin-bazyl (approved):

PR #2468, opened 2022-10-17
feat(slido): embedded app eventing

chburket:
# COMPLETES

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-374106

Add eventing for embedded apps data (found at locus.embeddedApps)

- embeddedAppsUtils.js: parse relevant values from locus event
- locus-info/index.js: updateEmbeddedApps() to update locus info and emit event if apps changed
- meeting/index.js: setUpLocusEmbeddedAppsListener() to listen for above event and trigger event for client
- added constants and tests to support above

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- plugin-meetings tests were run (including new tests)
- tested events received by client

### I certified that


arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
```diff
@@ -841,6 +844,27 @@ export default class LocusInfo extends EventsScope {
     }
   }
 
+  /**
+   * @param {Object} embeddedApps
+   * @returns {undefined}
+   * @memberof LocusInfo
+   */
+  updateEmbeddedApps(embeddedApps) {
```
robstax, line 852: i'm a little concerned about performance here especially in locus events. locus has *lots* of events and `isEqual` does deep comparison of everything in `embeddedApps`. so if `embeddedApps` has quite large, nested objects, i think this could get to be a little bit of a bottle neck. is there some other way we do this `!isEqual(this.embeddedApps, embeddedApps)` better? what does each object in `embeddedApp` look like? is it possible to use some other property on them?

packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -828,6 +828,52 @@ describe('plugin-meetings', () => {
...
+        LOCUSINFO.EVENTS.EMBEDDED_APPS_UPDATED,
+        newEmbeddedApps);
+      });
+
+      it('does not emit EMBEDDED_APPS_UPDATED when apps didn\'t change', () => {
+        locusInfo.updateEmbeddedApps(newEmbeddedApps);
+
+        locusInfo.emitScoped = sinon.stub();
+
+        locusInfo.updateEmbeddedApps(newEmbeddedApps);
```
robstax, line 871: i'm not sure this test does exactly what i think it should. in this test we're comparing the same reference to `newEmbeddedApps` so it doesn't fire a new event. would it be a more robust test to have a line between 867 and 871 that is `const clonedApps = deepClone(newEmbeddedApps)` and then pass that to `updateEmbeddedApps`?

robstax (commented):
lgtm! couple questions about comparing apps that i'd like to discuss/think about, but overall should be good quickly.

packages/node_modules/@webex/plugin-meetings/src/locus-info/embeddedAppsUtils.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns {Object} parsedObject - parsed embedded app object
+ */
+EmbeddedAppsUtils.parseApp = (embeddedApp) => {
+  const parsedApp = {...embeddedApp};
+
+  parsedApp.type = EMBEDDED_APP_TYPES.OTHER;
+  const url = parsedApp.instanceInfo?.appInstanceUrl;
+
+  if (url && url.match(SLIDO_REGEX)) {
+    parsedApp.type = EMBEDDED_APP_TYPES.SLIDO;
```
srinijay98, line 19: I don't see the reason why we need to differentiate slido from other apps - isn't it just another app?

srinijay98 (changes requested):

packages/node_modules/@webex/plugin-meetings/src/locus-info/embeddedAppsUtils.js
srinijay98: even in cantina - i don't think we should give special values to slido - if we trigger sdk for an embedded app, we just show them and hide them irrespective of which app is run

chburket: This was specifically requested by @arun3528

chburket: We are going to have to differentiate between app types because integration with each type of app will be different.  Cantina will need to ignore anything other than slido, so Arun wanted the logic to go in the SDK instead of the client app.  More types would be added as we support them.  I'm fine either way, though, so you and Arun can fight it out.  :)

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
chburket: Good point.  I think I can test the length of the embeddedApps array and the state of the first app to determine if it has changed.

packages/node_modules/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
chburket: `updateEmbeddedApps` already makes a clone of the embeddedApps object, so passing in the same `newEmbeddedApps` should test it just fine.  But if someone changes the implementation in the future it might not test it properly, so I think your suggestion is better.

packages/node_modules/@webex/plugin-meetings/src/locus-info/index.js
robstax: doublechecking...
so they are always in the same order?
and we only care about insert/deletes? we don't need to know if something changed within them?

robstax (commented):
couple more questions!

packages/node_modules/@webex/plugin-meetings/src/locus-info/embeddedAppsUtils.js
srinijay98: okie, I will confirm with Arun on this.

srinijay98: Slido alone might have some preference I guess - as we might be showing extra buttons at the bottom

robstax (approved):

PR #2475, opened 2022-10-19
fix(plugin-meetings): Using new reachability timeout values (1s for video mesh, 3s for the rest)

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-364898

With the current design of reachability in JS-SDK, it will always fail if the cluster is on the same network as the client, which means that it will always fail for video mesh nodes if client is on the corporate network and it also always fails for the cisco cloud clusters if you are on cisco network (or cisco vpn). Cisco users always get some video mesh nodes in their cluster list, so that means that reachability always takes 5s for cisco users, because:
- if they're on Cisco VPN, the reachability checks for all clusters will fail, so they wait for the 5s timeout 
- if they're not on VPN, the reachability checks for cisco video mesh nodes fail, because they're only available on cisco network, so again the clients will have to wait for the 5s timeout

According to Orpheus metrics, only 1.7% of web clients report reachability values greater than 3s to any cluster. Therefore, we're reducing the overall timeout from 5s to 3s.

We are working on a solution to reachability issues for nodes running on the same network, but it may take months before the required changes are done in the backend so in the meantime we are reducing the timeout for video mesh to 1s.

Also fixed:
- some incorrect logs,
- typos
- a bug that was causing a log "Attempting to set latency of null on closed peerConnection" being displayed every time when reachability checks are done

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

manually tested with and without vpn

### I certified that

---

### DIFF
--- a/packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
+++ b/packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
@@ -15,6 +15,9 @@ import {
 
 import ReachabilityRequest from './request';
 
+const DEFAULT_TIMEOUT = 5000;
+const VIDEO_MESH_TIMEOUT = 1000;
+
 /**
  * @class Reachability
  * @export
@@ -77,6 +80,8 @@ export default class Reachability {
 
       window.localStorage.setItem(REACHABILITY.localStorage, JSON.stringify(results));
 
+      LoggerProxy.logger.log('Reachability:index#gatherReachability --> Reachability checks completed');
+
       return results;
     }
     catch (getClusterError) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
```diff
@@ -174,7 +179,7 @@ export default class Reachability {
       peerConnection.begin = Date.now();
       peerConnection.setLocalDescription(description);
 
-      return this.iceGatheringState(peerConnection)
+      return this.iceGatheringState(peerConnection, cluster.isVideoMesh ? VIDEO_MESH_TIMEOUT : DEFAULT_TIMEOUT)
```
Coread, line 182: I can't see a test for this

arun3528 (approved):

packages/node_modules/@webex/plugin-meetings/src/reachability/index.js
marcin-bazyl: yeah the reachability code in SDK has no tests at all, I can add it

PR #2479, opened 2022-10-21
Spark 366191 region api timeout added

meta-dipanshu-sharma:
# COMPLETES #< [SPARK-366191](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-366191) >

Sometimes when api is not able to fetch response in a given time then it gives error for eg., this api `https://ds.ciscospark.com/v1/region`. We needed to add a timeout limit after which api calls should be cancelled automatically.

Added a timeout of 5 secs to above api.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/http-core/src/request/index.js
+++ b/packages/@webex/http-core/src/request/index.js
@@ -36,23 +36,27 @@ export default function request(options) {
     const successKey = `on${key}`;
     const errorKey = `on${key}Error`;
 
-    return interceptors.reduce((promise, interceptor) => promise.then(
-      (result) => {
-        interceptor.logOptions(options);
-        if (interceptor[successKey]) {
-          return interceptor[successKey](options, result);
-        }
+    return interceptors.reduce(
+      (promise, interceptor) =>
+        promise.then(
+          (result) => {
+            interceptor.logOptions(options);
+            if (interceptor[successKey]) {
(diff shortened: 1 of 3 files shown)


packages/@webex/http-core/src/request/index.js
```diff
@@ -18,6 +18,10 @@ export default function request(options) {
 
   options.headers = options.headers || {};
 
+  // it will take timeout if passed from request options
+  // else will add default timeout of 5 secs(5000ms)
+  options.timeout = options.timeout || 5000;
```
marcin-bazyl, line 23: 5s is probably too small. For me the request we do to https://webexapis.com/v1/meetingPreferences  always takes somewhere between 4s and 6s. Also the addMedia request that sends Roap offer takes sometimes over 15s when Linus reports 503 (for example if we send Turn Discovery Request and the server has the feature toggle for that disabled).
So, either this default should be higher, something like 30s or specific bigger timeouts need to be set for the 2 requests I've mentioned, but I don't know how many other "slow" requests we have.

arun3528 (commented):
Can you add the test for it

arun3528 (approved):

PR #2491, opened 2022-10-25
SDK v3 multistream api: draft, created only to trigger a deploy to Amplify

marcin-bazyl:
This is just a draft PR created against master branch so that the code is deployed to Amplify.
These changes are meant to be delivered to sdk_v3 branch, not master, but Amplify hosting seems to be enabled only for PRs destined for master.


marcin-bazyl:
This won't work until we deploy a version of the webrtc-media lib that contains the multistream changes

PR #2498, opened 2022-11-01
fix: release to publish correct version number

arun3528:
Fixes the issues with the release broken with yarn workspace, converts all the workspace reference to actual version number

Screenshot 2022-11-01 at 4 12 48 PM

### DIFF
--- a/tooling/dep2.js
+++ b/tooling/dep2.js
@@ -1,0 +1,22 @@
+
+const fs = require('fs');
+
+const glob = require('glob');
+
+// gets reference to all the package.json file
+glob('packages/**/package.json', {}, (er, files) => {
+  files.forEach((file) => {
+    const currentPackageJson = fs.readFileSync(file);
+
+    const parsedPackages = JSON.parse(currentPackageJson);
+
+    let stringified = JSON.stringify(parsedPackages, null, 2);
+
+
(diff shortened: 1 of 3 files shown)


tooling/dep2.js
```diff
@@ -0,0 +1,22 @@
...
+const glob = require('glob');
+
+// gets reference to all the package.json file
+glob('packages/**/package.json', {}, (er, files) => {
+  files.forEach((file) => {
+    const currentPackageJson = fs.readFileSync(file);
+
+    const parsedPackages = JSON.parse(currentPackageJson);
+
+    let stringified = JSON.stringify(parsedPackages, null, 2);
```
maxinteger, line 13: It would be safer to parse the JSON than update the version numbers in the object and stringify it only before you save it

tooling/dep2.js
```diff
@@ -0,0 +1,22 @@
+
```
maxinteger, line 1: Why do you need this script?

`yarn pack` should do it automatically for you https://yarnpkg.com/features/workspaces#publishing-workspaces

maxinteger (changes requested):

maxinteger (approved):
I approve it to un-block others

PR #2513, opened 2022-11-23
feat: added turnDiscoveryEnabled to js_sdk_add_media_failures

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-380079

< DESCRIBE YOUR CHANGES >
Added "turnDiscoveryEnabled" to "js_sdk_add_media_failures" metric

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -4266,7 +4266,8 @@ export default class Meeting extends StatelessWebexPlugin {
               correlation_id: this.correlationId,
               locus_id: this.locusUrl.split('/').pop(),
               reason: error.message,
-              stack: error.stack
+              stack: error.stack,
+              turnDiscoveryEnabled: true
```
marcin-bazyl, line 4270: This is not right. We don't always do turn discovery - see inside `doTurnDiscovery()` function - it doesn't do it for at least 2 possible reasons being checked at the beginning - you're already familiar with that code :-). 
Also, thinking about it a bit more... TURN discovery might fail and then we proceed without a TURN server and I'm wondering what value we should put in metrics for that case... Maybe instead of 'turnDiscoveryEnabled' we need 2 different fields: `turnDiscoverySkippedReason` and `turnServerUsed`:
- `turnDiscoverySkippedReason` would be a string 'config' if TURN discovery is disabled in config, 'reachability' if some clusters were reachable or `undefined` if TURN discovery is not skipped
- `turnServerUsed` - would be true only if TURN discovery resolved with some turnServerInfo (so it wasn't undefined)

cc @arun3528

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.js
shnaaz: yes it doesn't do turn discovery every time but then if that's not attempted then it has not failed as well. Skipping is not failure, does it make sense to have skipping in failure metric

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -961,6 +961,19 @@ describe('plugin-meetings', () => {
           await meeting.addMedia().catch((err) => {
             assert.exists(err);
             assert.isNull(meeting.statsAnalyzer);
+            assert(Metrics.sendBehavioralMetric.calledOnce);
```
marcin-bazyl, line 964: it would be good to have at least 1 test for the case where turnDiscoverySkippedReason is not undefined 

BTW, these existing tests are not very good, because they will pass if addMedia() succeeds and doesn't send any metrics at all as the catch() code will not be executed at all in such case, but don't try to fix that now as I've refactored them on the sdk_v3 branch.

marcin-bazyl (approved):

PR #2515, opened 2022-11-23
Spark 379747 [CDR Compliance] add encrypted text to decryption flows

hokeeffecisco:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-383474

Calling compliance for eDiscovery has to decrypt new content.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-ediscovery/src/transforms.js
+++ b/packages/@webex/internal-plugin-ediscovery/src/transforms.js
@@ -214,7 +214,7 @@ class Transforms {
         // post and share activities have content which needs to be decrypted
         // as do meeting, recording activities, customApp extensions, and space information updates
         if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp') &&
-          !activity.spaceInfo?.name && !activity.spaceInfo?.description) {
+          !activity.spaceInfo?.name && !activity.spaceInfo?.description && !activity.encryptedTextKeyValues) {
           return Promise.resolve(object);
         }
 
@@ -362,8 +362,8 @@ class Transforms {
         }
 
         // Decrypt encrypted text map if present
-        if (activity.encryptedTextKeyValues) {
-          for (let [key, value] of Object.entries(activity.encryptedTextKeyValues)) {
+        if (activity.encryptedTextKeyValues !== undefined) {
+          for (let [value] of activity.encryptedTextKeyValues.values()) {
             promises.push(requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
               [activity.encryptionKeyUrl, value])
               .then((decryptedMessage) => {
--- a/packages/@webex/internal-plugin-ediscovery/src/transforms.js
+++ b/packages/@webex/internal-plugin-ediscovery/src/transforms.js
@@ -361,6 +361,23 @@ class Transforms {
           }
         }
 
+        // Decrypt encrypted text map if present
+        if (activity.encryptedTextKeyValues) {
+          for (let [key, value] of Object.entries(activity.encryptedTextKeyValues)) {
+            promises.push(requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
+              [activity.encryptionKeyUrl, value])
+              .then((decryptedMessage) => {
+                value = decryptedMessage;
+              })
+              .catch((reason) => {
+                ctx.webex.logger.error(`Decrypt activity.encryptedTextKeyValues error for activity ${activity.activityId} in container ${activity.targetId}: ${reason}`);
+                activity.error = reason;
+
+                return Promise.resolve(object);
+              }));
+          }
+        }
+
         // Decrypt meeting title if present
         if (activity?.meeting?.title) {
           promises.push(requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
--- a/packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
+++ b/packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
@@ -416,6 +416,22 @@ describe('EDiscovery Transform Tests', () => {
 
         return result;
       });
+
+      it('Calls the correct decrypt functions when transforming activity.encryptedTextKeyValues', () => {
+        object.body.verb = 'add';
+        object.body.objectDisplayName = undefined;
+        object.body.encryptedTextKeyValues = new Map();
+        object.body.encryptedTextKeyValues.set('CallingLineId', 'Encrypted Phone Number 1');
+        object.body.encryptedTextKeyValues.set('CallingNumber', 'Encrypted Phone Number 2');
+        object.body.encryptedTextKeyValues.set('RedirectingNumber', 'Encrypted Phone Number 3');
+        const result = Transforms.decryptReportContent(ctx, object, reportId)
+          .then(() => {
+            assert.callCount(ctx.webex.internal.encryption.decryptText, object.body.encryptedTextKeyValues.size);
+            assert.equal(activity.error, undefined);
+          });
+
+        return result;
+      });
     });
   });
 });


packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -361,6 +361,23 @@ class Transforms {
           }
         }
 
+        // Decrypt encrypted text map if present
+        if (activity.encryptedTextKeyValues !== undefined) {
+          for (let [value] of activity.encryptedTextKeyValues.values()) {
```
sregger, line 366: My JavaScript is starting to rust. But `[value]` stands out to me. Are you sure it's what you mean to do? e.g.

image

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -361,6 +361,23 @@ class Transforms {
...
+            promises.push(requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
+              [activity.encryptionKeyUrl, value])
+              .then((decryptedMessage) => {
+                value = decryptedMessage;
```
sregger, line 370: I don't think this sets the decrypted data.

image

I think you'd need to set the value directly on the object. e.g.

    activity.encryptedTextKeyValues.values()[0] = decryptedMessage

But please note that `values()[0]` is pseudo code, I can't recall how this object is structured, but it should be simple to modify.

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -416,6 +416,22 @@ describe('EDiscovery Transform Tests', () => {
...
+        object.body.verb = 'add';
+        object.body.objectDisplayName = undefined;
+        object.body.encryptedTextKeyValues = new Map();
+        object.body.encryptedTextKeyValues.set('CallingLineId', 'Encrypted Phone Number 1');
+        object.body.encryptedTextKeyValues.set('CallingNumber', 'Encrypted Phone Number 2');
+        object.body.encryptedTextKeyValues.set('RedirectingNumber', 'Encrypted Phone Number 3');
+        const result = Transforms.decryptReportContent(ctx, object, reportId)
+          .then(() => {
+            assert.callCount(ctx.webex.internal.encryption.decryptText, object.body.encryptedTextKeyValues.size);
+            assert.equal(activity.error, undefined);
```
sregger, line 430: This test is inline with the rest. But it's all too mocked to be realistic. Which might explain some of the comments above.

packages/@webex/internal-plugin-ediscovery/src/transforms.js
sregger: Or if I use objects

image

packages/@webex/internal-plugin-ediscovery/src/transforms.js
hokeeffecisco: Yes, cat is not a reference, it's a new local variable

packages/@webex/internal-plugin-ediscovery/README.md
```diff
@@ -39,6 +39,11 @@ Run the ediscovery from the top level using
 
 > yarn run test --packages @webex/internal-plugin-ediscovery
```
hokeeffecisco, line 41: Added an example of how to run just unit tests, and select one test

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -214,7 +214,7 @@ class Transforms {
         // post and share activities have content which needs to be decrypted
         // as do meeting, recording activities, customApp extensions, and space information updates
         if (!['post', 'share'].includes(activity.verb) && !activity.meeting && !activity.recording && !(activity.extension && activity.extension.extensionType === 'customApp') &&
```
hokeeffecisco, line 216: Every encrypted item has to be checked for null here, or it's optimised out.

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -416,6 +416,25 @@ describe('EDiscovery Transform Tests', () => {
...
+            for (const value of object.body.encryptedTextKeyValues.values()) {
```
hokeeffecisco, line 431: Added a check for encrypted text on the values after the transform function

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -361,6 +361,23 @@ class Transforms {
           }
         }
 
+        // Decrypt encrypted text map if present
+        if (activity.encryptedTextKeyValues !== undefined) {
+          activity.encryptedTextKeyValues.forEach((encryptedData, key) => {
```
hokeeffecisco, line 366: forEach will create a function for every key and value(encryptedData). After the decryption promise has returned we write back to the map by key with the decrypted text.

sregger: Is this meant to be the other way around? i.e.

    (key, encryptedData) =>

sregger: image

This is new, I'm not sure why the web-client is involved here. But we had to get reviews from the webex-js-sdk team in the past. So this is not too surprising. i.e. My review alone will not be enough.

hokeeffecisco: No, it's correct, but it does look wrong. Intelij helps here
image

sregger (approved):

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -361,6 +361,23 @@ class Transforms {
...
+            promises.push(requestWithRetries(ctx.webex.internal.encryption, ctx.webex.internal.encryption.decryptText,
+              [activity.encryptionKeyUrl, encryptedData])
+              .then((decryptedMessage) => {
+                activity.encryptedTextKeyValues.set(key, decryptedMessage);
+              })
+              .catch((reason) => {
+                ctx.webex.logger.error(`Decrypt activity.encryptedTextKeyValues error for activity ${activity.activityId} in container ${activity.targetId}: ${reason}`);
```
Coread, line 373: Need a unit test to check the behavior of the catch.

hokeeffecisco: Sure.

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -416,6 +416,25 @@ describe('EDiscovery Transform Tests', () => {
 
         return result;
       });
+
+      it('Calls the correct decrypt functions when transforming activity.encryptedTextKeyValues', () => {
+        object.body.verb = 'add';
+        object.body.objectDisplayName = undefined;
+        object.body.encryptedTextKeyValues = new Map();
```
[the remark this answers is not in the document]
hokeeffecisco: It comes from the json representation of this [Java Object](https://sqbu-github.cisco.com/WebExSquared/ediscovery/blob/master/dto/src/main/java/com/cisco/wx2/dto/ediscovery/Activity.java#L159)

Coread (changes requested):
please see comment and question

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -225,7 +225,8 @@ class Transforms {
           return Promise.resolve(object);
         }
 
-        if (!container.onBehalfOfUser) {
```
hokeeffecisco, line 228: I've let the flow continue for encryptedTextKeyValues when there is on onBehalfOf present. This is the correct use case for CDR compliance which uses org wide keys.

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -416,6 +416,64 @@ describe('EDiscovery Transform Tests', () => {
...
+            assert.callCount(ctx.webex.internal.encryption.decryptText, object.body.encryptedTextKeyValues.size);
+            assert.equal(activity.error, undefined);
+            for (const value of object.body.encryptedTextKeyValues.values()) {
+              assert.equal(value, decryptedText);
+            }
+          });
+
+        return result;
+      });
+
```
hokeeffecisco, line 438: Added a test with the onBehalfOfUser is empty

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -416,6 +416,64 @@ describe('EDiscovery Transform Tests', () => {
...
```
hokeeffecisco, line 458: Added a test for the decryption fail flow

packages/@webex/internal-plugin-ediscovery/src/transforms.js
```diff
@@ -522,6 +524,30 @@ class Transforms {
           }
         }
 
+        // Decrypt encrypted text map if present
+        if (activity.encryptedTextKeyValues !== undefined) {
+          for (const [key, value] of Object.entries(activity.encryptedTextKeyValues)) {
```
Coread, line 529: suggestion: minor point, but the `forEach` construction when iterating Object.entries is more common in this codebase. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#iterating_through_an_object

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -428,6 +428,67 @@ describe('EDiscovery Transform Tests', () => {
...
+          CallingLineId: 'Encrypted Phone Number 1',
+          CallingNumber: 'Encrypted Phone Number 2',
+          RedirectingNumber: 'Encrypted Phone Number 3',
+        };
+        const result = Transforms.decryptReportContent(ctx, object, reportId)
+          .then(() => {
+            assert.fail('Decrypt did not fail as expected');
+          })
+          .catch(() => {
+            assert.isDefined(activity.error);
```
Coread, line 487: suggestion: you can be more precise here e.g.
```js
assert.deepEqual(activity.error, new Error('@@@@@@'));
```

Coread (approved):

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
hokeeffecisco: Nice idea

packages/@webex/internal-plugin-ediscovery/test/unit/spec/transforms.js
```diff
@@ -289,16 +289,17 @@ describe('EDiscovery Transform Tests', () => {
...
-          .returns(Promise.reject(new Error('@@@@@@')));
+          .returns(Promise.reject(expectedError));
 
         const result = Transforms.decryptReportContent(ctx, object, reportId)
           .then(() => {
             assert.fail('Decrypt did not fail as expected');
           })
           .catch(() => {
-            assert.isDefined(activity.error);
+            assert.deepEqual(activity.error, expectedError);
```
hokeeffecisco, line 302: Nothing to do with my PR, but Adding a more robust assert here for the desired exception, rather than any exception.

PR #2517, opened 2022-11-24
feat(plugin-meetings): added connection type to "js_sdk_add_media_success" metric

marcin-bazyl:
Added connectionType to BEHAVIORAL_METRICS.ADD_MEDIA_SUCCESS metric (this is part of the TURN TLS feature, so that we know how many clients connect using TURN).

jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-370008

There seems to be a bug in Firefox and Safari where they don't set the relayProtocol correctly for the local candidate when connecting using TURN. It works fine in Chrome/Edge.

There is an indentation change in addMedia() in meeting/index.js, because there were some `then()` calls that were unnecessarily nested, so it's best to view the diff with "hide whitespace" checkbox ticked. I've fixed it here, because this is also already fixed on sdk_v3 branch and is causing conflicts every time I merge latest master into sdk_v3, so fixed it now on master to make my life easier.

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/properties.js
+++ b/packages/@webex/plugin-meetings/src/media/properties.js
@@ -1,5 +1,7 @@
 import {
+  ICE_STATE,
   MEETINGS,
+  PC_BAIL_TIMEOUT,
   QUALITY_LEVELS
 } from '../constants';
 import LoggerProxy from '../common/logs/logger-proxy';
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -195,4 +197,99 @@ export default class MediaProperties {
...
+    // we can only get the connection type after ICE connection has been established
+    await this.waitForIceConnectedState();
+
+    const allStatsReports = [];
+
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      const statsResult = await this.peerConnection.getStats();
+
+      statsResult.forEach((report) => allStatsReports.push(report));
```
marcin-bazyl, line 253: putting all statsReports into an array so that it's easier to work with (so I can filter them, etc)

packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -4238,127 +4236,108 @@ export default class Meeting extends StatelessWebexPlugin {
...
-                // TODO: Fix this after the error code pr goes in
-                reject(createMeetingsError(30202, 'Meeting connection failed'));
-              }
-              else {
-                LoggerProxy.logger.info(`${LOG_HEADER} PeerConnection CONNECTED`);
-                resolve(peerConnection);
-              }
-            }, PC_BAIL_TIMEOUT);
-
-            this.once(EVENT_TRIGGERS.MEDIA_READY, () => {
```
marcin-bazyl, line 4330: this existing code on master was wrong, because EVENT_TRIGGERS.MEDIA_READY is triggered from 2 places: when we start receiving remote media and when we set the local tracks before any media connection is even created, so this block of code has always been resolving too early. Now I had to fix it, otherwise the call to getStats() would not work as it was being called too early.

packages/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -195,4 +197,99 @@ export default class MediaProperties {
...
+          resolve();
+        }
+      };
+
+      timer = setTimeout(() => {
+        this.peerConnection.removeEventListener('iceconnectionstatechange', iceListener);
+        reject();
+      }, PC_BAIL_TIMEOUT);
+
+      this.peerConnection.addEventListener('iceconnectionstatechange', iceListener);
```
arun3528, line 234: @marcin-bazyl  i am assuming this wont override the event registred in peer connection manager

packages/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -195,4 +197,99 @@ export default class MediaProperties {
...
+      if (localCandidate === undefined) {
+        LoggerProxy.logger.warn(`Media:properties#getCurrentConnectionType --> failed to find local candidate "${pair.localCandidateId}" in getStats() results`);
+
+        return false;
+      }
+
+      let connectionType;
+
+      if (localCandidate.relayProtocol) {
+        connectionType = `TURN-${localCandidate.relayProtocol.toUpperCase()}`;
```
arun3528, line 278: nice

packages/@webex/plugin-meetings/src/media/properties.js
```diff
@@ -195,4 +197,99 @@ export default class MediaProperties {
...
+      this.peerConnection.addEventListener('iceconnectionstatechange', iceListener);
+    });
+  }
+
+  /**
+   * Returns the type of a connection that has been established
+   *
+   * @returns {Promise<'UDP' | 'TCP' | 'TURN-TLS' | 'TURN-TCP' | 'TURN-UDP' | 'unknown'>}
+   */
+  async getCurrentConnectionType() {
```
arun3528, line 243: could be nicer to move this logic to webrtc-core sometime later

packages/@webex/plugin-meetings/src/media/properties.js
marcin-bazyl: correct, you can call `addEventListener()` multiple times to register multiple listeners

packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -4246,129 +4244,110 @@ export default class Meeting extends StatelessWebexPlugin {
...
+            connectionType
+          }
+        );
+      })
       .catch((error) => {
         // Clean up stats analyzer, peer connection, and turn off listeners
         const stopStatsAnalyzer = (this.statsAnalyzer) ? this.statsAnalyzer.stopAnalyzer() : Promise.resolve();
 
-        stopStatsAnalyzer
+        return stopStatsAnalyzer
```
Coread, line 4350: question: Prior to this change, the error caught on line 4346 would not have been re-thrown. Do we know what kind of errors these are?

packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -4246,129 +4244,110 @@ export default class Meeting extends StatelessWebexPlugin {
...
-          );
+          // When the self state changes to JOINED then request the floor
+          this.floorGrantPending = true;
+        }
 
-          return Promise.resolve();
-        }))
+        return {};
+      })
+      .then(() => this.mediaProperties.getCurrentConnectionType())
```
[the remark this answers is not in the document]
marcin-bazyl: you can enable "hide whitespace" checkbox in github, then you'll see the exact changes:
image
in this file it's the call to `getCurrentConnectionType()` and the call to `waitForIceConnectedState()` that are the main changes

packages/@webex/plugin-meetings/src/media/properties.js
marcin-bazyl: yeah, I was thinking the same, but for now as we're not using the PeerConnection from webrtc-core on master yet, I've put it here

packages/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: correct, previously the code was wrong - it was not throwing any error that happened during this long promise chain in addMedia(). I've found this out now when I added the "should reject if waitForIceConnectedState() rejects" test and it was not passing, because the error that is thrown in line 4318 in my new code (and line 4332 before my changes) was getting lost and not thrown to the client.

marcin-bazyl: it may be many types of errors, I don't know the full list, but whatever they are, they should all be thrown all the way up to the client

packages/@webex/plugin-meetings/src/meeting/index.js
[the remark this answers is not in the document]
arun3528: nice :D

packages/@webex/plugin-meetings/src/media/properties.js
marcin-bazyl: actually, peer connection manager doesn't use `addEventListener()` but just sets `oniceconnectionstatechange` directly. This is also fine - it works, I've just checked this by running the sample app with the code from this PR and I could see the "PeerConnectionManager:index#setPeerConnectionEvents" logs in the console when joining meeting and also later when I blocked and unblocked UDP traffic, so it's fine.

marcin-bazyl: I've checked it on chrome, firefox and safari

Coread (approved):

arun3528 (approved):

PR #2519, opened 2022-11-28
Fix event listeners for closed captions

srinijay98:

Currently the code allows users to register for events multiple times. by adding a flag inside the class, we will register the events only once.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-voicea/src/voicea.ts
+++ b/packages/@webex/internal-plugin-voicea/src/voicea.ts
@@ -53,8 +53,10 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
    * Listen to websocket messages
    */
   private listenToEvents() {
-    this.webex.internal.llm.on('event:relay.event', this.eventProcessor);
-    this.hasSubscribedToEvents = true;
+    if (!this.hasSubscribedToEvents) {
+      this.webex.internal.llm.on('event:relay.event', this.eventProcessor);
+      this.hasSubscribedToEvents = true;
+    }
   }
 
   public deregisterEvents() {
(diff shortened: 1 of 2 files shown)


packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -231,7 +235,9 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
    */
   private sendAnnouncement = (): void => {
     if (this.hasVoiceaJoined || !this.webex.internal.llm.isConnected()) return;
-    this.listenToEvents();
+    if (!this.hasSubscribedToEvents) {
```
marcin-bazyl, line 238: it's better to check that flag inside listenToEvents(), so that in the future if anyone adds another call to this.listenToEvents() the flag will be already checked for them.

packages/@webex/internal-plugin-voicea/test/unit/spec/voicea.js
```diff
@@ -69,6 +69,16 @@ describe('plugin-voicea', () => {
           trackingId: sinon.match.string,
         });
       });
+
+      it('listens to events once', () => {
+        assert.equal(voiceaService.hasSubscribedToEvents, false);
```
marcin-bazyl, line 74: hasSubscribedToEvents is private so we shouldn't be reading it in the test. I think you're only able to do it because this test file is a js file not ts - otherwise typescript would not let you read the private field. Anyway, we don't need to check that private field, checking the spy on listenToEvents is enough. Although you'll need to spy on `webex.internal.llm` now (see my other comment)

srinijay98: Modified

packages/@webex/internal-plugin-voicea/src/voicea.ts
srinijay98: Modified

marcin-bazyl (approved):

PR #2523, opened 2022-12-05
fix: unit tests for Media.createMediaConnection() added

marcin-bazyl:
jira: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-385742

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/internal-media-core-wrapper.ts
+++ b/packages/@webex/plugin-meetings/src/media/internal-media-core-wrapper.ts
@@ -1,0 +1,9 @@
+import {MediaConnection as MC} from '@webex/internal-media-core';
+
+/* We have this wrapper just because otherwise it's impossible to mock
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/media/internal-media-core-wrapper.ts
```diff
@@ -0,0 +1,9 @@
+import {MediaConnection as MC} from '@webex/internal-media-core';
+
+/* We have this wrapper just because otherwise it's impossible to mock
+ * RoapMediaConnection and MultistreamRoapMediaConnection constructors in unit tests,
+ * because they are exported by @webex/internal-media-core as non-writable, non-configurable
+ * properties of MediaConnection and sinon doesn't allow spying on them or stubbing them.
+ */
+export const RoapMediaConnection = MC.RoapMediaConnection;
```
arun3528, line 8: Makes sense thanks for the comment , Jest would be the option but will take more time

PR #2536, opened 2022-12-15
Breakout sessions

Coread:
# COMPLETES SPARK-344982

This implements breakout sessions for attendees.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


webpack.config.js
```diff
@@ -62,11 +62,11 @@ module.exports = (
         // we don't need to manually rebuild after changing code.
         alias[`./packages/${packageName}`] = path.resolve(
           __dirname,
-          `./packages/${packageName}/src/index.js`
+          `./packages/${packageName}/src/index`
```
Coread, line 65: needed so a typescript plugin can be used in the sample apps

packages/@webex/plugin-meetings/src/breakouts/README.md
```diff
@@ -0,0 +1,191 @@
...
+// This will cause the participant to return to the main session
+breakouts.currentBreakoutSession.leave()
+
+// Ask for help when in a breakout session. 10 second cooldown.
+breakout.currentBreakoutSession.askForHelp()
+```
+
+#### Members
+
+The current session members are available on the meeting object as normal
```
torgeadelin, line 56: Maybe rephrase this, hard to understand what you mean.

torgeadelin: I mean the whole paragrah/s? Can't tell

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!mainSession) {
+      throw new Error('Cannot leave, no main session found');
+    }
+
+    return mainSession.join();
+  },
+
+  /**
+   * Sends a help request for the current breakout
+   */
```
torgeadelin, line 88: @returns {Promise}

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -0,0 +1,111 @@
...
+      body: {
+        groupId: this.groupId,
+        sessionId: this.sessionId
+      }
+    });
+  },
+
+  /**
+   * Parses the participants from the locus object
+   * @param locus
```
torgeadelin, line 102: What is locus, object, or url?

packages/@webex/plugin-meetings/src/breakouts/collection.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2015-2023 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import AmpCollection from 'ampersand-collection';
+
+import {MEETINGS} from '../constants';
+
+import Breakout from './breakout';
+
+const BreakoutCollection = AmpCollection.extend({
```
torgeadelin, line 11: What is this `AmpCollection` dependency, and why does it help us here?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  },
+
+  /**
+   * The initial roster lists need to be quried because you don't
+   * get a breakout.roster event when you join the meeting
+   */
+  queryRosters() {
+    this.webex.request(
+      {uri: `${this.url}/roster`, qs: {locusUrl: btoa(this.locusUrl)}
+    }).then((result) => {
```
torgeadelin, line 95: should we handle a .catch ?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    session.parseRoster(locus);
+  },
+
+  /**
+   * Sets up listener for broadcast messages sent to the breakout session
+   */
+  listenToBroadcastMessages() {
+    this.listenTo(this.webex.internal.llm, 'event:breakout.message', (event) => {
```
torgeadelin, line 122: why not just `this.webex.internal.llm.on('event:breakout.message', ...)`

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+      const {data: {senderUserId, sentTime, message}} = event;
+
+      this.trigger(BREAKOUTS.EVENTS.MESSAGE, {
+        senderUserId,
+        sentTime,
+        message,
+        // FIXME: This is only the current sessionId
```
torgeadelin, line 129: is this because the new llm logic you added initially automatically connects to the new dataChannelURL and this could happen half way through this action (receiving the `event:breakout.message` and triggering it further)?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        sessionId: this.currentBreakoutSession.sessionId,
+      });
+    })
+  },
+
+  /**
+   * Sets up a listener for roster messags from mecury
+   */
+  listenToBreakoutRosters() {
+    this.listenTo(this.webex.internal.mercury, 'event:breakout.roster', (event) => {
```
torgeadelin, line 141: Any reason we get the roster events on mercury and not LLM?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    })
+  },
+
+  /**
+   * Updates the information about the current breakout
+   * @param {Object} params
+   * @returns {void}
+   */
+  updateBreakout(params) {
+    this.set(params);
```
torgeadelin, line 153: why is this needed?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        forEach(payload.breakoutSessions[state], (breakout) => {
+          const {sessionId} = breakout;
+
+          if (!breakouts[sessionId]) {
+            breakouts[sessionId] = breakout;
+            breakouts[sessionId].active = false;
+            breakouts[sessionId].assigned = false;
+            breakouts[sessionId].allowed = false;
+            breakouts[sessionId].assignedCurrent = false;
+            breakouts[sessionId].requested = false;
```
torgeadelin, line 189: why are you setting this to false for all states?
doesn't the `breakout` object already have information about its own state?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+          }
+
+          breakouts[sessionId][state] = true;
```
torgeadelin, line 192: oh, I see what you're doing. 
You're updating it here...

Seems a bit confusing?

you could just do something like 
```js
 breakouts[sessionId].allowed = state === 'allowed';
```
 in the if statement above.
 
 Also, would be nice to create constants for all the available states.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1188,6 +1199,49 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    });
+
+    this.breakouts.on(BREAKOUTS.EVENTS.MEMBERS_UPDATE, () => {
+      Trigger.trigger(
+        this,
+        {
+          file: 'meeting/index',
+          function: 'setUpBreakoutsListener'
+        },
+        EVENT_TRIGGERS.MEETING_BREAKOUTS_UPDATE,
```
torgeadelin, line 1242: one says members update(in the breakounts.on), and the other general breakouts update(in the trigger). Is this on purpose?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1712,6 +1766,19 @@ export default class Meeting extends StatelessWebexPlugin {
         }
       });
 
+    this.locusInfo.on(LOCUSINFO.EVENTS.CONTROLS_MEETING_BREAKOUT_UPDATED,
+      ({breakout}) => {
+        this.breakouts.updateBreakout(breakout);
```
torgeadelin, line 1771: do we need to call this on `this.breakouts.on(BREAKOUTS.EVENTS.MEMBERS_UPDATE, () => {` too?

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -112,6 +115,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       }
     };
 
+    if (breakoutsSupported) {
+      body.deviceCapabilities = ['BREAKOUTS_SUPPORTED'];
```
torgeadelin, line 119: make it a constant?

torgeadelin (changes requested):
I've tested manually in the samples app and seems to work ok.
Most of my comments are questions, with small request of changes.

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -120,6 +120,18 @@ MemberUtil.isHandRaised = (participant: any) => {
   return participant.controls?.hand?.raised || false;
 };
 
+/**
+ * @param {Object} participant the locus participant
+ * @returns {Boolean}
+ */
+MemberUtil.isSupportsBreakouts = (participant) => {
```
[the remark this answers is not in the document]
Coread: changed to isBreakoutsSupported

packages/@webex/plugin-meetings/package.json
```diff
@@ -53,6 +53,7 @@
     "@webex/plugin-people": "workspace:^",
```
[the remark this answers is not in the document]
Coread: I reset these to the version from beta but they are still showing up here. Perhaps github is not showing the diff correctly?

packages/@webex/plugin-meetings/src/breakouts/README.md
Coread: I've updated this paragraph. Hopefully it is clearer now

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
Coread: locus object. I'll update the comment

packages/@webex/plugin-meetings/src/breakouts/collection.ts
Coread: We use ampersand state in the SDK. Ampersand state has the concept of collections, groups of models. The collections offer helpful methods for searching, sorting models as well as event propagation. From what I can tell looking at the rest of the SDK, in places where authors didn't use ampersand collections, they had to resort to writing a lot of extra custom cost to do things like a get a model by its ID or to listen to an event within a model in a collection. There's no need for all that extra code, since the framework we already chose does this for us.

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+   * Update the current locus url of the meeting
+   */
+  locusUrlUpdate(locusUrl) {
+    this.set('locusUrl', locusUrl);
+  },
+
+  /**
+   * The initial roster lists need to be quried because you don't
+   * get a breakout.roster event when you join the meeting
+   */
```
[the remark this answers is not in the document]
Coread: fixed

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
Coread: updated

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -0,0 +1,111 @@
...
+        groupId: this.groupId,
+        sessionId: this.sessionId
+      }
+    });
+  },
+
+  /**
+   * Leaves the breakout session to return to the main session
+   * @returns {Promise}
+   */
```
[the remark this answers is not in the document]
Coread: updated

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: This is again where taking advantage of the framework that ampersand state provides means you can worry less and write less code. `listenTo` is designed so that when you destroy the object you automatically stop listening to events. You don't have to worry that you've left the event handler there and need to clean it up manually.

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: Yes. I am happy to leave this edge case at this stage though. It would only be a problem in cases where the application developer was filtering the broadcast messages in their app by sessionId and dropped a message because it wasn't for the current room.

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: I don't know the specific reasoning architecturally. If I had to guess it would be because the main meeting roster list arrives via the locus DTO on normal mercury so this is to be consistent with that.

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: You mean why not call .set directly? In the rest of the sdk we typically don't access built in methods of ampersand state directly. After all, ampersand state is basically an implementation choice and so exposing its API to something outside of the unit might make it more brittle. Typically we define methods like updateWhatever and then other parts of the code don't have to know that set exists, or how it works. Adding a separate method also gives us room to sanitize input in the future

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: I'm initializing as false and then updating the one to be true later

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: The payload arrives as an object with arrays of sessions in different states
e.g.
{
active: [{session1}, {session2}],
allowed: [{session1}]
}

I want to flatten and deduplicate this into a single collection of models where each state is just a flag. Sessions can appear in multiple places, so you can't just assign say allowed based on whether the current bucket is allowed because you may be overwriting a value set in by a previous bucket.

Coread: I'm now using constants for session states

packages/@webex/plugin-meetings/src/meeting/index.ts
Coread: Are you asking why this is the same event being triggered eventually? This is because the result is the same for the application developer. Something has changed in the breakouts and they need to go update their state. There was no point in propagating two distinct events outside of the meetings plugin for this.

packages/@webex/plugin-meetings/src/meeting/index.ts
Coread: no. This is receiving information from locus and updating the breakouts with that information. After doing that it triggers the event so the app consuming this knows something has changed. The other event is listening to changes that have already happened regarding the members of the breakouts. No need to update anything else on the breakouts, simply to inform the consuming app that something has changed.

packages/@webex/plugin-meetings/src/meeting/request.ts
Coread: updated

packages/@webex/plugin-meetings/src/breakouts/collection.ts
torgeadelin: Thanks for the thorough explanation! Makes sense!

packages/@webex/plugin-meetings/src/breakouts/index.ts
torgeadelin: I guess not?

packages/@webex/plugin-meetings/src/breakouts/index.ts
torgeadelin: thank you for the explanation.

packages/@webex/plugin-meetings/src/breakouts/index.ts
torgeadelin: yeah, don't mind it here, was just trying to understand why was needed.

packages/@webex/plugin-meetings/src/breakouts/index.ts
torgeadelin: I was referring to why calling this.set() since it's called below again with more parameters. I guess it would be nice to describe what is `params` or how does the object look like.

packages/@webex/plugin-meetings/src/breakouts/index.ts
torgeadelin: makes sense!!

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: I see, I was just curious why we are listening on "member_update" and trigger "breakout_update" instead of "member_update". Is a bit confusing but since it's the same data then maybe it's fine? Would be nice to be documented.

torgeadelin (approved):
LGTM

packages/@webex/plugin-meetings/src/breakouts/collection.ts
arun3528: we will need the same for members at some point

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: I actually missed this one. I'll take a look. We typically do log an error in this situation.

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: yes, perhaps we could define a type somehow to show what exactly gets set on the object. Ampersand state ignores the params that are not defined anyway.

packages/@webex/plugin-meetings/src/breakouts/index.ts
Coread: I've added a catch here

packages/@webex/plugin-meetings/src/meeting/index.ts
Coread: The current session sends member update events and those will still apply. This is for members of other sessions, sessions the user isn't in which is handled by the consuming app going over the members of those sessions. It felt overkill to submit individual member update events for each other breakout session and then force the consuming app to keep track of all of those listeners, given that they all come from a single roster update event.

arun3528 (approved):

PR #2538, opened 2022-12-16
fix: set unique debugId for media connections

marcin-bazyl:
Changed Media.createMediaConnection() to take a value for debugId that's passed to the roap media connection, so that we can set it to be unique per meeting - this is especially useful when looking at logs from integration test runs where we have multiple test users and meetings happening at the same time in same browser tab.

While I was there, I've reorganised the arguments of Media.createMediaConnection() and put things that are optional in options and things that are not optional as separate args.

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.js
+++ b/packages/@webex/plugin-meetings/src/meeting/index.js
@@ -4436,14 +4436,17 @@ export default class Meeting extends StatelessWebexPlugin {
   };
 
   createMediaConnection(turnServerInfo) {
-    const mc = Media.createMediaConnection(this.mediaProperties, {
-      isMultistream: this.isMultistream,
-      meetingId: this.id,
-      remoteQualityLevel: this.mediaProperties.remoteQualityLevel,
-      enableRtx: this.config.enableRtx,
-      enableExtmap: this.config.enableExtmap,
-      turnServerInfo
-    });
+    const mc = Media.createMediaConnection(
+      this.isMultistream,
+      `MC-${this.id.substring(0, 4)}`,
+      {
+        mediaProperties: this.mediaProperties,
+        remoteQualityLevel: this.mediaProperties.remoteQualityLevel,
+        enableRtx: this.config.enableRtx,
+        enableExtmap: this.config.enableExtmap,
+        turnServerInfo
+      }
+    );
 
     this.mediaProperties.setMediaPeerConnection(mc);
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -4436,14 +4436,17 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      isMultistream: this.isMultistream,
-      meetingId: this.id,
-      remoteQualityLevel: this.mediaProperties.remoteQualityLevel,
-      enableRtx: this.config.enableRtx,
-      enableExtmap: this.config.enableExtmap,
-      turnServerInfo
-    });
+    const mc = Media.createMediaConnection(
+      this.isMultistream,
+      `MC-${this.id.substring(0, 4)}`,
```
marcin-bazyl, line 4441: I wanted to keep it short (hence only 4 chars from the id) and also easy to find if you don't know the meeting id (hence the fixed "MC-" part)

chrisadubois, line 4441: IMO the id should be done as a function, i.e.,

```
.../utils
const generateMeetingDebugId = (prefix = 'MC', id = 'DEFAULT') => return `${prefix}-id.substring(0, 4)`
...
and then call that function here
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1105,7 +1105,7 @@ describe('plugin-meetings', () => {
           assert.calledWith(meeting.roap.doTurnDiscovery, meeting, false);
           assert.calledOnce(meeting.mediaProperties.setMediaDirection);
           assert.calledOnce(Media.createMediaConnection);
-          assert.calledWith(Media.createMediaConnection, sinon.match.any, sinon.match({turnServerInfo: undefined}));
+          assert.calledWith(Media.createMediaConnection, false, `MC-${meeting.id.substring(0, 4)}`, sinon.match({turnServerInfo: undefined}));
```
chrisadubois, line 1108: that way you could just use that function here again

chrisadubois (approved):
Non blocking changes but please review the request

packages/@webex/plugin-meetings/src/meeting/index.js
marcin-bazyl: done

PR #2540, opened 2022-12-19
feat: showing stats results in logs on CTRL+click on a video pane in the sample app

marcin-bazyl:
Added a "Get stats" button and also showing stats when user does a CTRL+click on a video pane.

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -34,6 +34,7 @@ const integrationEnv = document.getElementById('integration-env');
 const turnDiscoveryCheckbox = document.getElementById('enable-turn-discovery');
 const eventsList = document.getElementById('events-list');
 const multistreamLayoutElm = document.querySelector('#multistream-layout');
+const getStatsButton = document.getElementById('get-stats');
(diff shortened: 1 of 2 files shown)


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1825,12 +1838,50 @@ function updateMultistreamVideoLayout() {
   }
 }
 
+async function getStatsForVideoPane(meeting, videoPane) {
+  const {remoteMedia} = videoPane;
```
arun3528, line 1842: are there plans to have this function inside the SDK or WCF ?

arun3528 (approved):
For now its fine but i feel that we should move this logic inside the media connection logic

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: yeah, I think longer term we should have WCME API to get the stats for a specific receive slot and then we can expose that API through SDK/WCF. Right now it's just a hack and as I'm accessing private property I can only do it from js, not typescript so I've put it in the sample app

PR #2549, opened 2023-01-05
refactor(plugin-meetings): renamed methods for requesting/releasing share floor

marcin-bazyl:
Renamed the following private methods from Meeting class:
- stopFloorRequest() renamed to releaseScreenShareFloor()
- share() renamed to requestScreenShareFloor()

because the names were confusing, especially the share() method name which was very vague and generic.

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1502,7 +1502,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
       // If user moved to a JOINED state and there is a pending floor grant trigger it
       if (this.floorGrantPending && payload.newSelf.state === MEETING_STATE.STATES.JOINED) {
-        this.share()
+        this.requestScreenShareFloor()
           .then(() => { this.floorGrantPending = false; });
       }
     });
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1792,7 +1792,7 @@ export default class Meeting extends StatelessWebexPlugin {
               this,
               {
                 file: 'meeting/index',
-                function: 'stopFloorRequest'
+                function: 'localShare'
```
marcin-bazyl, line 1795: this in theory should be a function name, but the all the code around this area is not putting the function name, so I've put "localShare" here to be consistent with the surrounding code

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3457,22 +3457,34 @@ describe('plugin-meetings', () => {
...
+          meeting.selfId = 'some self id';
+          meeting.locusInfo.mediaShares = [{name: 'content', url: url1, floor: {beneficiary: {id: meeting.selfId}}}];
           meeting.locusInfo.self = {url: url2};
+          meeting.mediaProperties = {mediaDirection: {sendShare: true}};
           meeting.meetingRequest.changeMeetingFloor = sinon.stub().returns(Promise.resolve());
         });
-        it('should call change meeting floor', async () => {
-          const share = meeting.share();
+        it('should call changeMeetingFloor()', async () => {
+          const share = meeting.releaseScreenShareFloor();
```
marcin-bazyl, line 3472: this test has always been wrong - it was calling share() instead of stopFloorRequest(), so was never testing what it was supposed to test. After changing this line I had to add a couple more things to beforeEach() (see above) to make it work.

srinijay98 (approved):

PR #2550, opened 2023-01-05
feat(plugin-meetings): automatically connect llm

Coread:
# COMPLETES # SPARK-394463

As we introduce more meeting features that require LLM we should have a way to automatically connect the data channel when available.

I've added an ensureLLMConnected method that will check the currently connected channel's locus URL and disconnect/reconnect as necessary.

ensureLLMConnected is called initially on join, whenever the locus info changes and on meeting cleanup.

I've added a new meetings config parameter `enableAutomaticLLM` so users of the SDK can opt in to automatically connecting LLM.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/package.json
+++ b/packages/@webex/plugin-meetings/package.json
@@ -45,6 +45,7 @@
     "@webex/internal-media-core": "^0.0.17-beta",
     "@webex/internal-plugin-conversation": "workspace:^",
     "@webex/internal-plugin-device": "workspace:^",
+    "@webex/internal-plugin-llm": "workspace:^",
     "@webex/internal-plugin-mercury": "workspace:^",
     "@webex/internal-plugin-metrics": "workspace:^",
     "@webex/internal-plugin-support": "workspace:^",
--- a/packages/@webex/plugin-meetings/src/config.ts
+++ b/packages/@webex/plugin-meetings/src/config.ts
@@ -88,6 +88,7 @@ export default {
     enableRtx: true,
     receiveTranscription: false,
     enableExtmap: false,
+    enableAutomaticLLM: false,
     experimental: {
       enableMediaNegotiatedEvent: false,
       enableUnifiedMeetings: false,
--- a/packages/@webex/plugin-meetings/src/locus-info/infoUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/infoUtils.ts
@@ -29,6 +29,10 @@ InfoUtils.parse = (info, roles, isJoined = true) => {
     parsed.meetingNumber = info.meetingId;
   }
 
+  if (info.datachannelUrl) {
+    parsed.datachannelUrl = info.datachannelUrl;
+  }
+
   return parsed;
 };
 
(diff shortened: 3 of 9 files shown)


packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -1944,10 +1944,27 @@ export default class Meeting extends StatelessWebexPlugin {
             this.inMeetingActions.get()
           );
         }
+
+        this.handleDataChannelUrlChange(payload.info.datachannelUrl);
```
arun3528, line 1948: would recommend to move this logic to join as we only need llm when meeting is actively joined meeting

arun3528 (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.js
Coread: This needs to be here to disconnect/reconnect the data channel when we move in and out of breakout sessions. However I'll put a check in ensureLLMConnected to only connect if we are joined

packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -3557,7 +3574,11 @@ export default class Meeting extends StatelessWebexPlugin {
         );
 
         return join;
-      }).then(async (join) => {
+      })
+      .then(async () => {
+        await this.ensureLLMConnected();
+      })
+      .then(async (join) => {
```
marcin-bazyl, line 3581: the join argument will always be empty now. It used to be returned be the previous then clause, now the newly added then() doesn't return it

arun3528 (approved):

packages/@webex/plugin-meetings/src/meeting/util.js
```diff
@@ -116,6 +116,8 @@ MeetingUtil.joinMeeting = (meeting, options) => {
 };
 
 MeetingUtil.cleanUp = (meeting) => {
+  meeting.ensureLLMConnected();
```
marcin-bazyl, line 119: should this call maybe be part of the promise chain returned by this function? Otherwise cleanUp() may resolve before LLM is disconnected

packages/@webex/plugin-meetings/src/meeting/index.js
```diff
@@ -3616,6 +3637,30 @@ export default class Meeting extends StatelessWebexPlugin {
       });
   }
 
+  /**
+   * Connects to low latency mercury and reconnects if the address has changed
+   * @param {String} datachannelUrl
+   * @returns {Promise}
+   */
+  async ensureLLMConnected() {
```
marcin-bazyl, line 3645: the name is a bit misleading because this function also just disconnects LLM if we are not joined, but not sure what a better name would be without making it very long. Maybe at least mention the disconnection in the comments?

packages/@webex/plugin-meetings/src/meeting/util.js
Coread: I've changed this to be part of the promise chain

arun3528:
can you also not push the samples changes they will auto pushed after the PR is merged

PR #2574, opened 2023-01-18
feat(plugin-meetings): send MQEs for multistream connections

mccarthytyler:
# COMPLETES [SPARK-377700](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-377700)

Allowing MQEs to be sent from a multistream connection as well as transcoded streams.

- Sending multiple transmit and receiver objects with their respective steams in a single MQE interval.
- Fixing a bug where we could not ascertain a network type on a multistream connection.
Note: relies on PR from Internal-media-core: https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/156

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


.lintstagedrc
```diff
@@ -1,5 +1,5 @@
 {
-  "*": [
+  "*.{js,ts}": [
```
arun3528, line 2: looks good

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -76,54 +102,7 @@ export class StatsAnalyzer extends EventsScope {
...
-          latency: [],
-          bitRate: [],
-          frameRate: [],
-          resolutionWidth: [],
-          resolutionHeight: [],
-          requestedKeyFrame: [],
-          receivedKeyFrame: [],
-        },
-      },
+      resolutions: {},
```
arun3528, line 105: i feel that some of these needs to be defined , else when testing it out you might see errors seeing recv or undefined 

can you test it and confirm

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -416,6 +388,73 @@ export class StatsAnalyzer extends EventsScope {
       return;
     }
 
+    // Generate empty stats results
+    if (!this.statsResults[type]) {
```
arun3528, line 392: looks like you are setting it here , Do you think this is ok then creating a const and attaching it !?

arun3528 (commented):
@mccarthytyler  is there some har which which tells how the date is generated ?

Can we also check with the WAP team to validate the date , they can tell if any values are missing after the change 

The PR is little big , But i can understand its hard to have one with MQE

packages/@webex/plugin-meetings/src/media/properties.ts
```diff
@@ -259,10 +259,13 @@ export default class MediaProperties {
     const allStatsReports = [];
 
     try {
-      // eslint-disable-next-line no-await-in-loop
       const statsResult = await this.webrtcMediaConnection.getStats();
 
-      statsResult.forEach((report) => allStatsReports.push(report));
+      if (typeof statsResult.forEach === 'function') {
```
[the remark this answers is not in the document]
mccarthytyler: This is because depending on the type of media connection we get the stats reports in different formats, either a js object or a RTCStatsReport object.  This chunk of code is to unify them.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: We generate the previous mqa items dynamically, so instead of `this.lastMqaDataSent['audio']`  we have `this.lastMqaDataSent['audio-recv-1']`, `this.lastMqaDataSent['video-send-2']` etc.  I have tested it and it is able to handle the comparisons between last and current.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Updated for a better approach

mccarthytyler:
> @mccarthytyler is there some har which which tells how the date is generated ?
> 
> Can we also check with the WAP team to validate the date , they can tell if any values are missing after the change
> 
> The PR is little big , But i can understand its hard to have one with MQE

Not sure what you mean here.  Isn't the date generated by the utility that sends the MQE?

packages/@webex/plugin-meetings/src/media/properties.ts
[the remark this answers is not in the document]
marcin-bazyl: I think I would prefer to unify them at WCME or internal-media-core level. Longer term we want the RoapMediaConnection and MultistreamRoapMediaConnection classes to have the same API interface. We're calling getStats() here and I think it's reasonable to expect that the result has a forEach() method (like the normal webrtc getStats() result does), so can we change WCME to return an object that has forEach() from getStats()?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -237,82 +182,125 @@ export class StatsAnalyzer extends EventsScope {
...
-      name: MEDIA_DEVICES.SPEAKER,
+    newMqa.intervalMetadata.peripherals = [];
+
+    let audioTrackLabel = _UNKNOWN_;
+    let videoTrackLabel = _UNKNOWN_;
+
+    // Find track labels
+    Object.keys(this.statsResults).forEach((mediaType) => {
+      if (mediaType.includes('video-send') && this.statsResults[mediaType].trackLabel) {
+        videoTrackLabel = this.statsResults[mediaType].trackLabel;
```
marcin-bazyl, line 283: hmm not sure about this... in reality we are only ever going to have a single sender for audio and video. Even WCME is hardcoded for that right now (it doesn't support publishing more than 1 track of same type at the same time). This code here is written in a generic way to support many senders (it iterates over the statsResults keys) but the output is just a single audio/video track label - so it just uses the last one it finds, so it won't work correctly if we do have multiple senders anyway. Maybe we should just make it clear that we only have one sender and not bother with "video-send-${idx}" but just have "video-send" and not have the loop here but just access this.statsResults['video-send'].trackLabel?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -556,253 +576,262 @@ export class StatsAnalyzer extends EventsScope {
...
-        } else {
           if (
-            currentStats.framesEncoded === previousStats.framesEncoded ||
-            currentStats.framesEncoded === 0
+            currentStats.totalPacketsReceived === previousStats.totalPacketsReceived ||
+            currentStats.totalPacketsReceived === 0
           ) {
             LoggerProxy.logger.info(
-              `StatsAnalyzer:index#compareLastStatsResult --> No ${mediaType} Frames Encoded`
+              `StatsAnalyzer:index#compareLastStatsResult --> No ${simplifiedMediaType} RTP packets received`
```
marcin-bazyl, line 635: I think it would be more useful for debugging to have the mediaType in the logs instead of the simplified one (applies to all the logs in `compareLastStatsResult()`)

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -556,253 +576,262 @@ export class StatsAnalyzer extends EventsScope {
...
-            this.statsResults.resolutions[mediaType].send.framesSent ===
-              this.lastStatsResults.resolutions[mediaType].send.framesSent ||
-            this.statsResults.resolutions[mediaType].send.framesSent === 0
-          ) {
-            LoggerProxy.logger.info(
-              `StatsAnalyzer:index#compareLastStatsResult --> No ${mediaType} frames sent`
-            );
-          }
+          this.emitStartStopEvents(
+            simplifiedMediaType,
```
marcin-bazyl, line 742: I'm not sure if using `simplifiedMediaType` here will work correctly - let's have a quick call to discuss this.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -832,47 +861,30 @@ export class StatsAnalyzer extends EventsScope {
     LoggerProxy.logger.trace('StatsAnalyzer:index#getStatsAndParse --> Collecting Stats');
 
     return this.mediaConnection.getTransceiverStats().then((transceiverStats) => {
-      this.filterAndParseGetStatsResults(
-        transceiverStats.video.sender,
-        STATS.VIDEO_CORRELATE,
-        true
+      transceiverStats.video.senders.forEach((sender, i) =>
```
marcin-bazyl, line 864: see my previous comment about local tracks - I think here we should just throw if we have more than 1 sender and hardcode everything for now to work with just 1 video sender (so use mediaType as "video-send")
and same for audio

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -37,26 +37,6 @@ export const EVENTS = {
...
- */
-const getSimplifiedMediaType = (mediaType) => {
-  if (mediaType.includes('share')) {
-    return STATS.SHARE_CORRELATE;
-  }
-  if (mediaType.includes('audio')) {
-    return STATS.AUDIO_CORRELATE;
-  }
-  if (mediaType.includes('video')) {
-    return STATS.VIDEO_CORRELATE;
```
marcin-bazyl: so now I think we shouldn't be using these `STATS.*_CORRELATE` constants anywhere anymore, right? When I search the code I see that `STATS.AUDIO_CORRELATE` is still used in a few places, can you check them?

packages/@webex/plugin-meetings/src/media/properties.ts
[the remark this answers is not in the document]
mccarthytyler: I talked to the WCME team and in WCME they break down and inject csi and mid into the stats, that's why it is a js object instead of a `RTCStatsReport`.  Since we need csi in the MQE but don't have it yet, should we try to do the same thing for RoapMediaConnection in internal-media-core?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Right.  I removed the `STATS.*_CORRELATE` constants and adjusted the references to `STATS.AUDIO_CORRELATE`.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -556,253 +494,260 @@ export class StatsAnalyzer extends EventsScope {
...
-          LoggerProxy.logger.info(
-            `StatsAnalyzer:index#compareLastStatsResult --> No ${mediaType} RTP packets received`
-          );
-        } else if (
-          currentStats.totalSamplesReceived === previousStats.totalSamplesReceived ||
-          currentStats.totalSamplesReceived === 0
-        ) {
-          LoggerProxy.logger.info(
-            `StatsAnalyzer:index#compareLastStatsResult --> No ${mediaType} samples received`
+          this.emitStartStopEvents(
```
marcin-bazyl, line 533: in this call (and all the other ones to emitStartStopEvents()) the mediaType will now be different to what it used to be even for transcoded meetings and as a result we will send call analyzer metrics "client.media.rx.stop" and "client.media.rx.start" with a different mediaType and that is not allowed according to this: https://sqbu-github.cisco.com/WebExSquared/event-dictionary/blob/ba25c9ff3c7440bb423a75ea28c6d0b023833a24/diagnostic-events.raml#L145

packages/@webex/plugin-meetings/test/unit/spec/networkQualityMonitor/index.js
```diff
@@ -42,12 +42,12 @@ describe('plugin-meetings', () => {
         roundTripTimeMeasurements: 14,
       },
       statsAnalyzerCurrentStats: {
-        audio: {
+        'video-send': {
```
marcin-bazyl, line 45: here "audio" was changed to "video-send" and few line further "video" to "audio-send" - can you double check if this is correct?

mccarthytyler: I changed it back to maintain consistency. Tests still pass as the values changing weren't enough to change the test result.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: updated to use a cumulative approach on media types.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -493,261 +489,303 @@ export class StatsAnalyzer extends EventsScope {
...
+        Object.keys(this.statsResults)
+          .filter((keys) => keys.includes(keyPrefix))
+          .reduce((prev, cur) => prev + (this.statsResults[cur].recv[value] || 0), 0);
+
+      const getPreviousStatsTotals = (keyPrefix: string, value: string): number =>
+        Object.keys(this.statsResults)
+          .filter((keys) => keys.includes(keyPrefix))
+          .reduce((prev, cur) => prev + (this.lastStatsResults[cur].recv[value] || 0), 0);
+
+      const getCurrentResolutionsStatsTotals = (keyPrefix: string, value: string): number =>
```
marcin-bazyl, line 502: this function is exactly the same like `getCurrentStatsTotals()` or am I missing something?
same for `getPreviousResolutionsStatsTotals()` being same as `getPreviousStatsTotals()`

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -493,261 +489,303 @@ export class StatsAnalyzer extends EventsScope {
...
-          return;
-        }
-
-        if (this.meetingMediaStatus.expected.sendAudio && mediaType.includes('audio-send')) {
-          currentStats = this.statsResults[mediaType].send;
-          previousStats = this.lastStatsResults[mediaType].send;
-
+      const getCurrentStatsTotals = (keyPrefix: string, value: string): number =>
+        Object.keys(this.statsResults)
+          .filter((keys) => keys.includes(keyPrefix))
```
marcin-bazyl, line 494: should we use startsWith() instead of includes() here? and also the variable name should be `key` not `keys`, right?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -493,261 +489,303 @@ export class StatsAnalyzer extends EventsScope {
...
+          'totalPacketsReceived'
+        );
+        const currentPacketsSent = getCurrentStatsTotals('video-share-recv', 'totalPacketsSent');
+        const previousPacketsReceived = getPreviousStatsTotals(
+          'video-share-recv',
+          'totalPacketsReceived'
+        );
+        const currentFramesReceived = getCurrentResolutionsStatsTotals(
+          'video-share-recv',
+          'totalFramesReceived'
```
marcin-bazyl, line 727: question: I don't see 'totalFramesReceived' referenced anywhere else in the code, so how/where is it being set?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -493,261 +489,303 @@ export class StatsAnalyzer extends EventsScope {
...
-            currentStats.totalPacketsReceived === previousStats.totalPacketsReceived ||
-            currentStats.totalPacketsSent === 0
-          ) {
+        const senderFramesDecoded = Object.keys(this.statsResults)
+          .filter((keys) => keys.includes('video-share-recv'))
+          .reduce(
+            (prev, cur) =>
+              prev +
+              ((this.statsResults.resolutions[cur].send &&
+                this.statsResults.resolutions[cur].send.framesDecoded) ||
```
marcin-bazyl, line 756: `send.framesDecoded` - this doesn't seem right, "send" and "frames decoded"? if we're sending then we're encoding not decoding frames, and the filter in line 751 is for "video-share-recv" while this whole block is for setting a value of `senderFramesDecoded` which is confusing me further

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Yes this was missing `.resolutions`

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Yes good call, updated.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Updated to `framesReceived` along with a couple other instances of this.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Looking at it again I realized the old version was a mistake and the check should have been `this.statsResults.resolutions[mediaType].recv.framesDecoded === 0`.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -598,203 +552,219 @@ export class StatsAnalyzer extends EventsScope {
...
+                0),
+            0
+          );
+
+        if (currentPacketsReceived === previousPacketsReceived || currentPacketsSent === 0) {
```
marcin-bazyl, line 748: this doesn't look right - "currentPacketsSent" doesn't belong here, should be `currentPacketsReceived` right?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
marcin-bazyl: overall there has been quite a few various mismatches in this compareLastStatsResult() function found in this PR, so looks like we don't have any tests for this. I guess, it's because most of this stuff just results in some logs, but maybe it would be worth adding some tests or at least use the sample app and check if the logs this function produces are correct and we don't log "no frame received" or something like that when video is clearly received.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: Correct, updated.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
mccarthytyler: The current tests do test `compareLastStatsResult` but only to trigger the start and stopped events.  I tested with the sample app to make sure the stats were coming in correctly, led to my last updates.

marcin-bazyl (approved):

arun3528 (approved):

PR #2581, opened 2023-01-24
feat(recording): add recording streaming service

chrisadubois:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-396426

Recording streaming service is required for the converged meeting backend. Web was left out of this conversation and we missed it. This implements a conditional split for old locus style recording, and new converged backend recording streaming service style recordings.

conditional split, abstraction and make into an object

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added unit tests for branch paths, did not change E2e tests, and tested manually with both converged and non converged meeting spaces and received the recordings.

### I certified that

---


packages/@webex/plugin-meetings/src/recording-controller/index.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * @param {MeetingRequest} request
+   * @param {LocusInfo} info
+   * @returns {void}
+   * @private
+   * @memberof RecordingController
+   */
+  private initialize(request: MeetingRequest, info: LocusInfo) {
```
arun3528, line 81: why do we need to save the LocusInfo here ? can we just update the specific info rather then the whole locusInfo object !?

chrisadubois: Yeah we could do it that way as well, or the recording class can parse the locusInfo, it's sort of 6 and one half dozen in this case because it's an internal implementation and is not exposed as an API

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -497,33 +497,19 @@ export default class MeetingRequest extends StatelessWebexPlugin {
...
-  }) {
-    const uri = `${options.locusUrl}/${CONTROLS}`;
-    const body = {
-      record: {
-        recording: options.recording,
-        paused: options.paused,
-      },
-    };
-
+  recordMeeting(options: {uri: string; body: object; method: string}): Promise<any> {
```
marcin-bazyl, line 507: This method is redundant now, it adds no value and has nothing specific about recording in the implementation. You could call it with a URL and body that would do something completely different like sending an SDP and it would also work, so I think we should just remove it.

packages/@webex/plugin-meetings/src/recording-controller/index.ts
```diff
@@ -0,0 +1,214 @@
...
+    this.request = request;
+    this.info = info;
```
marcin-bazyl, line 83: info is being save here, but not in the set() method. We should either save it in both places or not save it at all. I prefer to not save it at all.

packages/@webex/plugin-meetings/test/unit/spec/recording-controller/index.js
```diff
@@ -0,0 +1,284 @@
...
+                        fullState: {
+                            sessionId: 'testId',
+                        },
+                        services: {
+                            record: {
+                                url: 'test'
+                            }
+                        }
+                    })
+                    assert(controller.serviceUrl, 'test');
```
marcin-bazyl, line 24: we're reading private properties of the class here - I would say this is not ideal, but I leave it up to you, it will be a bit more work to test this if you can't read these properties. If this file was a ts file not js, you would have typescript errors here. Anyway, also worth considering if maybe the tests should be ts files? Anyway, up to you.

marcin-bazyl (commented):
Overall looks good - not bad for a manager ;-)

packages/@webex/plugin-meetings/src/recording-controller/index.ts
marcin-bazyl: I also think it's better not to save the whole info, it will save a bit of memory (locus info can grow quite large, especially for big meetings) and it's also confusing when debugging things and seeing same locus info duplicated in many places.
Anyway, you're already extracting the bits you need in the extract() method, so why save the whole info?

chrisadubois: Cool. I'm on the same page here and I can update the PR for this

packages/@webex/plugin-meetings/test/unit/spec/recording-controller/index.js
chrisadubois: yeah, I'll add getters and test using it that way

packages/@webex/plugin-meetings/src/recording-controller/index.ts
chrisadubois: yeah (as above) we won't use locusInfo anylonger.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -906,7 +906,7 @@ export default class Meeting extends StatelessWebexPlugin {
      * @public
      * @memberof Meeting
      */
-    this.recordingController = new RecordingController(this.meetingRequest, this.locusInfo);
+    this.recordingController = new RecordingController(this.meetingRequest);
```
marcin-bazyl, line 909: question: previously we were setting the locusInfo here, now we're relying on the various LOCUSINFO.EVENTS.* to populate all the things we need. Are these events guaranteed to fire at least once? I mean can there be cases when we got all locus info at the point of meeting object creation and there are no changes to it later so these events don't trigger?

packages/@webex/plugin-meetings/src/recording-controller/index.ts
```diff
@@ -78,36 +76,127 @@ export default class RecordingController {
...
   /**
-   * @param {LocusInfo} info
+   * @param {Object} options
    * @returns {void}
    * @public
    * @memberof RecordingController
    */
-  public set(info: LocusInfo) {
-    this.extract(info);
+  public set(options: {
```
marcin-bazyl, line 89: is this function called from anywhere now? it seems to be not used

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: This is a good question. I had the same question. I don't think so because the locusInfo is *always* instantiated with "nothing" and then it slowly gets populated with various DTOs, but it's a reasonable thing to build in elegance for.

What i *could* do to solve this problem is instantiate it with the various objects it requires from locusInfo here during the constructor phase and then still leave the event listeners, and that would handle both cases (in case locus changes their implementation I suppose)

As of right now, this code will work based on Locus, but *could* be hardened a bit more.

packages/@webex/plugin-meetings/src/recording-controller/index.ts
chrisadubois: it's not used, except for in the tests and just internally calls other functions, however, I think with your question above, I could use it to make the code simpler in meeting/index and solve the "hardening" issue with timing of events.

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: as an example, the *services* links DTO will *always* come in later and therefore must *always* trigger an event, and that's the logic we rely on. I know for sure the fullState is the same. locusUrl I guess they could change at any time?

marcin-bazyl: Yeah I think it's best if we just call this.recordingController.set() here with the current values from locusInfo and keep the listeners for any later changes. Then we are covered for all possible cases.

packages/@webex/plugin-meetings/src/recording-controller/index.ts
marcin-bazyl: yeah, let's do that

packages/@webex/plugin-meetings/src/meeting/index.ts
arun3528: From what i have seen in most of the cases the values comes during the join as part of locus (Meeting related property and persmissions )

But those values do change based on what host does 

in some cases it differs if the suer is guest or in lobby so we need to be careful when we test the scenarios out

arun3528: LocusUrl could change if there is a locus migration happening but the locusInfo wont as the meeting remains the same

chrisadubois: adjusted based on all these comments ! tested and looking good so far

packages/@webex/plugin-meetings/src/recording-controller/index.ts
```diff
@@ -0,0 +1,315 @@
...
+    return this.displayHints;
+  }
+
+  /**
+   * @param {Object} options
+   * @returns {void}
+   * @private
+   * @memberof RecordingController
+   */
+  private extract(options?: {
```
marcin-bazyl, line 200: a very minor thing, but this method feels kind of redundant now. It's only called from `set()` and its body could just be inside `set()`

marcin-bazyl (approved):

PR #2587, opened 2023-01-26
fix: workaround for edonus bug when handling Roap glare

marcin-bazyl:
# COMPLETES #
This is a workaround for https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-308749

Edonus sometimes doesn't handle Roap glare correctly. This happens especially when we are stealing screen share from someone. In that case we send an SDP offer to Edonus and also request floor from Locus. Locus giving us floor triggers an offer from Edonus. Edonus should not send us that offer if it already has an outstanding offer from us, but it does it and it sends it with an increased seq, which is a breach of the Roap protocol and as a result Linus doesn't correctly detect the glare condition and even though we reply with Error-Conflict to Edonus' offer, Linus keeps resending us that offer until eventually Edonus times out and call is destroyed.

Changing the code so that we only send the floor request to Locus after Roap offer-answer-ok sequence started by us is fully completed. This alignes our behaviour on beta more closely with how this has been done on master. On master floor request was sent after we received http response from Locus for our Roap offer (so it wasn't waiting for the answer and OK to be completed) and that was causing this issue to happen only roughly 1 in 10 times when stealing screen share. Without this fix, we are hitting the Edonus bug about 90% of the time when stealing screen share.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manual e2e with samples app, also running integration tests in circle ci multiple times

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -130,6 +130,7 @@ export const MEDIA_UPDATE_TYPE = {
   AUDIO: 'AUDIO',
   VIDEO: 'VIDEO',
   SHARE: 'SHARE',
+  LAMBDA: 'LAMBDA',
 };
 
 /**
@@ -5351,7 +5352,14 @@ export default class Meeting extends StatelessWebexPlugin {
    * @private
    * @memberof Meeting
    */
-  private enqueueMediaUpdate(mediaUpdateType: string, options: object) {
+  private enqueueMediaUpdate(mediaUpdateType: string, options: any) {
+    if (mediaUpdateType === MEDIA_UPDATE_TYPE.LAMBDA && typeof options?.lambda !== 'function') {
+      return Promise.reject(
+        new Error('lambda must be specified when enqueuing MEDIA_UPDATE_TYPE.LAMBDA')
+      );
+    }
+    const canUpdateMediaNow = this.canUpdateMedia();
+
     return new Promise((resolve, reject) => {
       const queueItem = {
         pendingPromiseResolve: resolve,
@@ -5364,6 +5372,10 @@ export default class Meeting extends StatelessWebexPlugin {
         `Meeting:index#enqueueMediaUpdate --> enqueuing media update type=${mediaUpdateType}`
       );
       this.queuedMediaUpdates.push(queueItem);
+
+      if (canUpdateMediaNow) {
+        this.processNextQueuedMediaUpdate();
+      }
     });
   }
 
@@ -5415,6 +5427,9 @@ export default class Meeting extends StatelessWebexPlugin {
           break;
         case MEDIA_UPDATE_TYPE.SHARE:
           this.updateShare(options).then(pendingPromiseResolve, pendingPromiseReject);
+          break;
+        case MEDIA_UPDATE_TYPE.LAMBDA:
+          options.lambda().then(pendingPromiseResolve, pendingPromiseReject);
           break;
         default:
           LoggerProxy.logger.error(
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5706,13 +5730,17 @@ export default class Meeting extends StatelessWebexPlugin {
...
+          .then(() =>
+            this.enqueueMediaUpdate(MEDIA_UPDATE_TYPE.LAMBDA, {
+              lambda: async () => {
+                if (startShare) {
+                  return this.requestScreenShareFloor();
+                }
 
-            return Promise.resolve();
-          })
+                return undefined;
```
chrisadubois, line 5740: should this return promise.reject? if we return undefined maybe we should write a log? especially since this is kind of hacky non-roap protocol

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2209,6 +2209,190 @@ describe('plugin-meetings', () => {
           );
           assert.isTrue(myPromiseResolved);
         });
+
+        it('should request floor only after roap transaction is completed', async () => {
```
chrisadubois, line 2213: jeez what a complicated test :(

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5706,13 +5730,17 @@ export default class Meeting extends StatelessWebexPlugin {
               remoteQualityLevel: this.mediaProperties.remoteQualityLevel,
             },
           })
-          .then(() => {
-            if (startShare) {
-              return this.requestScreenShareFloor();
-            }
+          .then(() =>
+            this.enqueueMediaUpdate(MEDIA_UPDATE_TYPE.LAMBDA, {
+              lambda: async () => {
```
chrisadubois, line 5735: what is the reason for the name lambda here?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5364,6 +5372,10 @@ export default class Meeting extends StatelessWebexPlugin {
         `Meeting:index#enqueueMediaUpdate --> enqueuing media update type=${mediaUpdateType}`
       );
       this.queuedMediaUpdates.push(queueItem);
+
+      if (canUpdateMediaNow) {
```
chrisadubois, line 5376: else, write a log ?

chrisadubois (approved):
all very small comments and questions

arun3528 (approved):

arun3528:
Thanks @marcin-bazyl for making it happen

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: if startShare is not true, it just means that we're not starting a share, because maybe we're just stopping a share or just changing the share stream, so it's not an error condition, it's just that we have nothing to do, and typescript was complaining when I had no `return` or an empty `return` so I'm returning undefined

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: other media update types are for specific predefined operations like updateAudio, updateVideo, updateMedia, this one is a generic one that just takes a function that can run anything, usually it's an anonymous function, so a "lambda function"

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: The test is simple, just the setup of the mocks is complicated. In the tests for updateShare I've separated the setup into `beforeEach()` so the 2 tests there are simpler. I haven't done it here, because that setup is only used in that 1 single test, although it is very similar to the setup done in the updateShare tests, so maybe this all could have been refactored further, but on the other hand the updateShare/updateMedia functions are all going to be deleted when we move to webrtc-core so I didn't want to spend too much time on these tests.

PR #2598, opened 2023-01-31
fix(plugin-meetings): set correct local audio state in localsdp 

mkesavan13:
# COMPLETES [SPARK-390306](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-390306)

When the user joins remotely muted (meeting demands users to join muted), then the user unmutes themselves, the locus wasn't unmuting the user. This is because, on join locus thinks we are locally muted as well as remotely muted as we send localSdp based on both local mute state and remote mute state.

We have introduced a new audio method to know if we are locally muted which is sent on the /media request on join. This way, locus is aware we are remotely muted but locally un mute. This way the ROAP OFFER / ANSWER / OK cycle is corrected.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/muteState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/muteState.ts
@@ -330,6 +330,17 @@ class MuteState {
   }
 
   /**
+   * Returns true if the user is locally muted
+   *
+   * @public
+   * @memberof MuteState
+   * @returns {Boolean}
+   */
+  public isLocallyMuted() {
+    return this.state.client.localMute || this.state.server.localMute;
+  }
+
+  /**
    * Returns true if the user is muted as a result of the client request (and not remotely muted)
    *
    * @public
--- a/packages/@webex/plugin-meetings/src/roap/index.ts
+++ b/packages/@webex/plugin-meetings/src/roap/index.ts
@@ -170,8 +170,8 @@ export default class Roap extends StatelessWebexPlugin {
           locusSelfUrl: meeting.selfUrl,
           mediaId: options.mediaId,
           correlationId: options.correlationId,
-          audioMuted: meeting.isAudioMuted(),
-          videoMuted: meeting.isVideoMuted(),
+          audioMuted: meeting.audio?.isLocallyMuted(),
+          videoMuted: meeting.video?.isLocallyMuted(),
           meetingId: meeting.id,
         })
         .then(() => {
@@ -222,8 +222,8 @@ export default class Roap extends StatelessWebexPlugin {
         locusSelfUrl: meeting.selfUrl,
         mediaId: options.mediaId,
         correlationId: options.correlationId,
-        audioMuted: options.audioMuted,
-        videoMuted: options.videoMuted,
+        audioMuted: meeting.audio?.isLocallyMuted(),
+        videoMuted: meeting.video?.isLocallyMuted(),
         meetingId: meeting.id,
       })
       .then(() => {
@@ -295,8 +295,8 @@ export default class Roap extends StatelessWebexPlugin {
         correlationId: meeting.correlationId,
         locusSelfUrl: meeting.selfUrl,
         mediaId: sendEmptyMediaId ? '' : meeting.mediaId,
-        audioMuted: meeting.isAudioMuted(),
-        videoMuted: meeting.isVideoMuted(),
+        audioMuted: meeting.audio?.isLocallyMuted(),
+        videoMuted: meeting.video?.isLocallyMuted(),
         meetingId: meeting.id,
       })
       .then(({locus, mediaConnections}) => {
@@ -353,8 +353,8 @@ export default class Roap extends StatelessWebexPlugin {
               roapMessage,
               // eslint-disable-next-line no-warning-comments
               // TODO: check whats the need for video and audiomute
-              audioMuted: !!meeting.isAudioMuted(),
-              videoMuted: !!meeting.isVideoMuted(),
+              audioMuted: meeting.audio?.isLocallyMuted(),
+              videoMuted: meeting.video?.isLocallyMuted(),
             })
           ),
           // mediaId: meeting.mediaId
--- a/packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
+++ b/packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
@@ -161,6 +161,17 @@ describe('plugin-meetings', () => {
       assert.isFalse(audio.isSelf());
     });
 
+    describe('#isLocallyMuted()', () => {
+      it('does not consider remote mute status for audio', async () => {
+        // simulate being already remote muted
+        meeting.remoteMuted = true;
+        // create a new MuteState intance
+        audio = createMuteState(AUDIO, meeting, {sendAudio: true});
+
+        assert.isFalse(audio.isLocallyMuted());
+      });
+    });
+
     describe('#handleClientRequest', () => {
       it('disables/enables the local audio track when audio is muted/unmuted', async () => {
         // mute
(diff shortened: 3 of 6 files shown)


packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -329,6 +329,17 @@ class MuteState {
...
   }
 
+  /**
+   * Returns true if the user is locally muted
+   *
+   * @public
+   * @memberof MuteState
+   * @returns {Boolean}
+   */
+  public isLocallyMuted() {
```
arun3528, line 339: Looks good to me as we are going have only one state for server or local 

@marcin-bazyl  any thoughts on this

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -170,7 +170,7 @@ export default class Roap extends StatelessWebexPlugin {
           locusSelfUrl: meeting.selfUrl,
           mediaId: options.mediaId,
           correlationId: options.correlationId,
-          audioMuted: meeting.isAudioMuted(),
+          audioMuted: meeting.audio.isLocallyMuted(),
```
marcin-bazyl, line 173: `meeting.audio` may be undefined

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -68,6 +68,15 @@ describe('plugin-meetings', () => {
       assert.isFalse(audio.isSelf());
     });
 
+    it('does not consider remote mute status - isLocallyMuted', async () => {
```
marcin-bazyl, line 71: minor thing, but to follow the convention used in the existing tests, it would be better to wrap this test with a `describe('#isLocallyMuted')`

marcin-bazyl (changes requested):
The fix is good, but I see it's only done for sending Roap OK. The same has to be done for all Roap messages (offer, answer and turn discovery messages)

packages/@webex/plugin-meetings/src/meeting/muteState.ts
marcin-bazyl: looks good, when we send Roap messages to Locus we need to indicate only the local mute, not the remote one.

arun3528: ok so if someone does any action and remote is muted then it will change the local to muted also 

@mkesavan13  can you check the media request what values it is using

mkesavan13: Noted @arun3528 & @marcin-bazyl 

Will keep an eye for this when we move to webrtc-core lib. Thank you.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -161,6 +161,17 @@ describe('plugin-meetings', () => {
       assert.isFalse(audio.isSelf());
     });
 
+    describe.only('#isLocallyMuted()', () => {
```
marcin-bazyl, line 164: `.only` needs to be removed

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -161,6 +161,17 @@ describe('plugin-meetings', () => {
...
+      it('does not consider remote mute status - isLocallyMuted', async () => {
```
marcin-bazyl, line 165: the bit "- isLocallyMuted" can be removed now

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -170,7 +170,7 @@ export default class Roap extends StatelessWebexPlugin {
           locusSelfUrl: meeting.selfUrl,
           mediaId: options.mediaId,
           correlationId: options.correlationId,
-          audioMuted: meeting.isAudioMuted(),
+          audioMuted: meeting.audio && meeting.audio.isLocallyMuted(),
```
marcin-bazyl, line 173: minor thing, but doing `meeting.audio?.isLocallyMuted()` would be shorter and work just as well right?

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: Sorry. I missed to remove it while pushing. Thanks for noticing. Changed now.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: Removed :-)

packages/@webex/plugin-meetings/src/roap/index.ts
mkesavan13: Changed everywhere Marcin. Thanks :-)

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -170,7 +170,7 @@ export default class Roap extends StatelessWebexPlugin {
           locusSelfUrl: meeting.selfUrl,
           mediaId: options.mediaId,
           correlationId: options.correlationId,
-          audioMuted: meeting.isAudioMuted(),
```
marcin-bazyl, line 173: we should really do the same for video too. It's the same MuteState class that manages video mute and it's the same Locus API that we use, so the same bug exists for video, although I'm not sure if you can have video mute on  entry in webex, but anyway, we should have the same code for audio and video

mkesavan13: I've made these changes @marcin-bazyl 

Kindly check. However this may not be the case until in the future we get a case for Remote Video Mute & On Join Video Mute. We have neither use cases right now in the Meetings.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -161,6 +161,19 @@ describe('plugin-meetings', () => {
       assert.isFalse(audio.isSelf());
     });
 
+    describe('#isLocallyMuted()', () => {
+      it('does not consider remote mute status', async () => {
+        // simulate being already remote muted
+        meeting.remoteMuted = true;
```
marcin-bazyl, line 167: this only applies to audio, not video, but I think we don't support remote video mute right now, so I guess it's fine (maybe worth saying this in the comment?)

packages/@webex/plugin-meetings/test/unit/spec/roap/index.ts
```diff
@@ -47,8 +47,12 @@ describe('Roap', () => {
       correlationId: 'correlation id',
       selfUrl: 'self url',
       mediaId: 'media id',
-      isAudioMuted: () => true,
-      isVideoMuted: () => false,
+      audio:{
+        isLocallyMuted: () => true,
+      },
+      video:{
+        isLocallyMuted: () => true,
```
marcin-bazyl, line 54: previously the test was setup for isVideoMuted to return false, probably so that we check that the video and audio are not mixed up, might be worth keeping it that way

marcin-bazyl (approved):

arun3528 (approved):
@mkesavan13  there is one comment from marcin on test rest looks good

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: Agreed @marcin-bazyl. Didn't realise it. Thanks for pointing. Have removed video related code for now. Lets revisit if time comes we have remote video mute by default.

packages/@webex/plugin-meetings/test/unit/spec/roap/index.ts
mkesavan13: Sorry I missed it out. Corrected now. Thanks @marcin-bazyl

Neeraj-swarnkar (approved):
LGTM

PR #2600, opened 2023-02-02
test(plugin-meetings): converged meetings testing users

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add the necessary test to validate if a converged organization operates as expected when used within the `meetings` plugin. The changes include source changes amending the `roapConnectionService` as a key to the `meeting.mediaProperties.webrtcMediaConnection` object for reference in both tests, debugging, and conditional reactions.

**NOTE** - The ENV variable `WEBEX_CONVERGED_ORG_ID` must be set in order for this tests to operate as expected.

## Links

[JIRA](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-385748).

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/util.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/util.ts
@@ -59,6 +59,22 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
         };
 
         meeting.mediaProperties.webrtcMediaConnection.roapMessageReceived(roapMessage);
+
+        let roapConnectionService;
+
+        try {
+          roapConnectionService = roapMessage.sdp
+            .split('\r\n')
+            .find((line) => line.includes('o='))
+            .split(' ')
+            .shift()
+            .replace('o=', '');
+        } catch {
+          roapConnectionService = undefined;
+        }
+
+        meeting.mediaProperties.webrtcMediaConnection.roapConnectionService = roapConnectionService;
+        meeting.mediaProperties.webrtcMediaConnection.roapMessage = roapMessage;
       }
     }
   }
--- a/packages/@webex/plugin-meetings/src/meetings/util.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/util.ts
@@ -60,20 +60,21 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
 
         meeting.mediaProperties.webrtcMediaConnection.roapMessageReceived(roapMessage);
 
-        let roapConnectionService;
+        let mediaServer;
 
+        // Attempt to collect the media server from the roap message.
         try {
-          roapConnectionService = roapMessage.sdp
+          mediaServer = roapMessage.sdp
             .split('\r\n')
-            .find((line) => line.includes('o='))
+            .find((line) => line.startsWith('o='))
             .split(' ')
             .shift()
             .replace('o=', '');
         } catch {
-          roapConnectionService = undefined;
+          mediaServer = undefined;
         }
 
-        meeting.mediaProperties.webrtcMediaConnection.roapConnectionService = roapConnectionService;
+        meeting.mediaProperties.webrtcMediaConnection.mediaServer = mediaServer;
         meeting.mediaProperties.webrtcMediaConnection.roapMessage = roapMessage;
       }
     }
--- a/packages/@webex/plugin-meetings/src/meetings/util.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/util.ts
@@ -58,27 +58,34 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
           errorCause,
         };
 
+        const mediaServer = MeetingsUtil.getMediaServer(roapMessage.sdp);
+
         meeting.mediaProperties.webrtcMediaConnection.roapMessageReceived(roapMessage);
 
-        let mediaServer;
-
-        // Attempt to collect the media server from the roap message.
-        try {
-          mediaServer = roapMessage.sdp
-            .split('\r\n')
-            .find((line) => line.startsWith('o='))
-            .split(' ')
-            .shift()
-            .replace('o=', '');
-        } catch {
-          mediaServer = undefined;
+        if (mediaServer) {
+          meeting.mediaProperties.webrtcMediaConnection.mediaServer = mediaServer;
         }
-
-        meeting.mediaProperties.webrtcMediaConnection.mediaServer = mediaServer;
-        meeting.mediaProperties.webrtcMediaConnection.roapMessage = roapMessage;
       }
     }
   }
+};
+
+MeetingsUtil.getMediaServer = (sdp) => {
+  let mediaServer;
+
+  // Attempt to collect the media server from the roap message.
+  try {
+    mediaServer = sdp
+      .split('\r\n')
+      .find((line) => line.startsWith('o='))
+      .split(' ')
+      .shift()
+      .replace('o=', '');
+  } catch {
+    mediaServer = undefined;
+  }
+
+  return mediaServer;
 };
 
 MeetingsUtil.checkForCorrelationId = (deviceUrl, locus) => {
--- a/packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
+++ b/packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
@@ -1,0 +1,126 @@
+import { config } from 'dotenv';
+
+import {assert} from '@webex/test-helper-chai';
+import {skipInNode} from '@webex/test-helper-mocha';
+
+import testUtils from '../../utils/testUtils';
+import webexTestUsers from '../../utils/webex-test-users';
+
+config();
+
+skipInNode(describe)('plugin-meetings', () => {
+  describe('converged-space-meeting', () => {
+    let users, alice, bob, chris;
+    let meeting = null;
+    let space = null;
+
+    before(() => webexTestUsers.generateConvergedTestUsers({
+        count: 3,
+        whistler: process.env.WHISTLER || process.env.JENKINS,
+      }).then((userSet) => {
+        users = userSet;
+        alice = users[0];
+        bob = users[1];
+        chris = users[2];
+        alice.name = 'alice';
+        bob.name = 'bob';
+        chris.name = 'chris';
+      }).then(() => Promise.all([
+        testUtils.syncAndEndMeeting(alice),
+        testUtils.syncAndEndMeeting(bob),
+      ])).catch((error) => {
+        console.log(error);
+      })
+    );
+
+    it('user "alice" starts a space', () =>
+      alice.webex.internal.conversation.create({
+        participants: [bob, chris],
+      }).then((conversation) => {
+        assert.lengthOf(conversation.participants.items, 3);
+        assert.lengthOf(conversation.activities.items, 1);
+        space = conversation;
+      }).then(() => Promise.all([
+        alice.webex.meetings.meetingInfo.fetchMeetingInfo(space.url, 'CONVERSATION_URL'),
+        alice.webex.meetings.meetingInfo.fetchMeetingInfo(space.url),
+      ])).then(([destinationWithType, destinationWithoutType]) => {
+        assert.exists(destinationWithoutType);
+        assert.exists(destinationWithType);
+        assert.exists(destinationWithoutType.body.meetingNumber);
+        assert.exists(destinationWithType.body.meetingNumber);
+      }),
+    );
+
+    it('user "alice" starts a meeting', () =>
+      Promise.all([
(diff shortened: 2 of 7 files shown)


packages/@webex/plugin-meetings/src/meetings/util.ts
```diff
@@ -59,6 +59,22 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
         };
 
         meeting.mediaProperties.webrtcMediaConnection.roapMessageReceived(roapMessage);
+
+        let roapConnectionService;
+
+        try {
+          roapConnectionService = roapMessage.sdp
```
chrisadubois, line 66: can you write a comment on this? seems kinda whacky but i know it's to check which backend this is going to. 

I'm not sure there is really roapConnectionservice though... name might better be mediaServer?

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
```diff
@@ -0,0 +1,126 @@
...
+          assert.isTrue(bob.meeting.mediaProperties.mediaDirection.receiveVideo);
+          assert.isTrue(chris.meeting.mediaProperties.mediaDirection.sendAudio);
+          assert.isTrue(chris.meeting.mediaProperties.mediaDirection.sendVideo);
+          assert.isTrue(chris.meeting.mediaProperties.mediaDirection.receiveAudio);
+          assert.isTrue(chris.meeting.mediaProperties.mediaDirection.receiveVideo);
+        })
+    );
+
+    it('user "alice", "bob", and "chris" should be using the "homer" roap connection service', () => {
+        assert.equal(alice.meeting.mediaProperties.webrtcMediaConnection.roapConnectionService, 'homer');
```
chrisadubois, line 121: homer should be a constant

packages/@webex/plugin-meetings/src/meetings/util.ts
marcin-bazyl: I would rather avoid having this code here and have that logic just in the test. You can check the SDP directly from the test by accessing `meeting.mediaProperties.webrtcMediaConnection.multistreamConnection.pc.getRemoteDescription().sdp` (it is a bit hacky because you will be accessing some private properties, but this is temporary solution anyway - we will not need it for long)

and I agree with Chris about the name, "mediaServer" is probably better

packages/@webex/plugin-meetings/src/meetings/util.ts
```diff
@@ -59,6 +59,22 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
...
+            .split('\r\n')
+            .find((line) => line.includes('o='))
```
marcin-bazyl, line 68: better to use `line.startsWith()`

packages/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -195,12 +195,19 @@ const delayedTest = (callback, timeout) =>
...
   });
 
-const addMedia = (user) => {
-  const mediaReadyPromises = {
-    local: new Defer(),
-    remoteAudio: new Defer(),
-    remoteVideo: new Defer(),
-  };
+const addMedia = (user, options = {}) => {
+  const mediaReadyPromises = Array.isArray(options.streams)
```
marcin-bazyl, line 199: maybe we should rename `streams` to `expectedMediaReadyTypes` or something like that to reflect the purpose of it better

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meetings/util.ts
InteractiveTimmy: Renaming this.

InteractiveTimmy: I've added test constants to `./test/utils/constants`.

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
InteractiveTimmy: Pending the correct event name, `media:negotiated` does not trigger for any user. @marcin-bazyl could you provide the correct event here?

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
```diff
@@ -0,0 +1,126 @@
...
+        chris.name = 'chris';
+      }).then(() => Promise.all([
+        testUtils.syncAndEndMeeting(alice),
+        testUtils.syncAndEndMeeting(bob),
+      ])).catch((error) => {
+        console.log(error);
+      })
+    );
+
+    it('user "alice" starts a space', () =>
```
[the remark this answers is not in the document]
InteractiveTimmy: moved to async/await. I ran into an issue in which the tests would take too long unless we initialized each promise prior to awaiting them all. We should consider this when deciding on whether to use async/await vs promises at "ends" of the application.

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
```diff
@@ -0,0 +1,126 @@
...
+        alice.webex.meetings.meetingInfo.fetchMeetingInfo(space.url),
+      ])).then(([destinationWithType, destinationWithoutType]) => {
+        assert.exists(destinationWithoutType);
+        assert.exists(destinationWithType);
+        assert.exists(destinationWithoutType.body.meetingNumber);
+        assert.exists(destinationWithType.body.meetingNumber);
+      }),
+    );
+
+    it('user "alice" starts a meeting', () =>
```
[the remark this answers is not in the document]
InteractiveTimmy: I've created a local serial workflow that bails/aborts whenever a single test in a chain fails. This will be elevated later. But should work here for now.

packages/@webex/plugin-meetings/test/utils/webex-test-users.js
```diff
@@ -53,6 +51,23 @@ const generateTestUsers = (options) =>
       console.error('#generateTestUsers=>ERROR', error);
     });
 
+    /**
+     * Generate test users using a converged test user organization.
+     *
+     * @param {Object} options - generateTestUser options param.
+     * @returns - An array of test users.
+     */
+const generateConvergedTestUsers = (options = {}) => {
```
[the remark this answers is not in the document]
InteractiveTimmy: An investigation can be performed later to determine if merging these two orgs into one is possible. For now, we should separate these to allow the testing workflows to proceed without further delay.

packages/@webex/plugin-meetings/src/meetings/util.ts
InteractiveTimmy: Changed.

packages/@webex/plugin-meetings/test/utils/testUtils.js
InteractiveTimmy: Changed.

packages/@webex/plugin-meetings/test/utils/constants.js
```diff
@@ -0,0 +1,9 @@
+// MOVE TO TEST CONSTANTS
+export const MEDIA_SERVERS = {
+  // The homer media server for converged meetings.
```
marcin-bazyl, line 3: it's better to change "converged" to "converged multistream"

packages/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -195,12 +195,19 @@ const delayedTest = (callback, timeout) =>
...
-const addMedia = (user) => {
-  const mediaReadyPromises = {
-    local: new Defer(),
-    remoteAudio: new Defer(),
-    remoteVideo: new Defer(),
-  };
+const addMedia = (user, options = {}) => {
+  const mediaReadyPromises = Array.isArray(options.expectedMediaReadyTypes)
+    ? options.expectedMediaReadyTypes.reduce((output, expectedMediaReadyType) => {
+      if (typeof stream !== 'string') {
```
marcin-bazyl, line 201: `stream` is always undefined now, it should be `expectedMediaReadyType`, right?

packages/@webex/plugin-meetings/src/meetings/util.ts
```diff
@@ -59,6 +59,23 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
         };
 
         meeting.mediaProperties.webrtcMediaConnection.roapMessageReceived(roapMessage);
+
+        let mediaServer;
+
+        // Attempt to collect the media server from the roap message.
+        try {
+          mediaServer = roapMessage.sdp
```
marcin-bazyl, line 67: I've checked and unfortunately looks like the browser is removing the username from the origin line in the remote SDP, so we cannot read it later from the test and we have to parse it out from the incoming message like you're doing it here. So for now this code needs to stay as it is here, but can you please wrap it in a function like `extractMediaServer()` or something like this and just call the function from here, so it's nicely contained and easier to remove in the future.

packages/@webex/plugin-meetings/src/meetings/util.ts
```diff
@@ -59,6 +59,23 @@ MeetingsUtil.handleRoapMercury = (envelope, meetingCollection) => {
...
+            .split('\r\n')
+            .find((line) => line.startsWith('o='))
+            .split(' ')
+            .shift()
+            .replace('o=', '');
+        } catch {
+          mediaServer = undefined;
+        }
+
+        meeting.mediaProperties.webrtcMediaConnection.mediaServer = mediaServer;
```
marcin-bazyl, line 77: `sdp` is undefined for most roap messages, so you end up with the catch and overwriting `meeting.mediaProperties.webrtcMediaConnection.mediaServer` with `undefined`. It's better to check roapMessage.messageType to be ROAP.ROAP_TYPES.ANSWER or ROAP.ROAP_TYPES.OFFER

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
```diff
@@ -0,0 +1,163 @@
...
+      assert.isTrue(bob.meeting.mediaProperties.mediaDirection.receiveAudio);
+      assert.isTrue(bob.meeting.mediaProperties.mediaDirection.receiveVideo);
+      assert.isTrue(chris.meeting.mediaProperties.mediaDirection.sendAudio);
+      assert.isTrue(chris.meeting.mediaProperties.mediaDirection.sendVideo);
+      assert.isTrue(chris.meeting.mediaProperties.mediaDirection.receiveAudio);
+      assert.isTrue(chris.meeting.mediaProperties.mediaDirection.receiveVideo);
+    });
+
+    it(`user "alice", "bob", and "chris" should be using the "${MEDIA_SERVERS.HOMER}" roap connection service`, async () => {
+      // await testUtils.waitForEvents([ // PENDING CORRECT EVENT NAMES
```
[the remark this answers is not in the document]
InteractiveTimmy: The `media:negotiated` event does not fire in this context, and the test times out. Do you happen to know the correct event to listen for in this context?

packages/@webex/plugin-meetings/test/utils/webex-test-users.js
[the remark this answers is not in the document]
InteractiveTimmy: Move all tests to use the new organization.

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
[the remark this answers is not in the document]
marcin-bazyl: that is the correct event. It should be emitted when SDP exchange is done. If it doesn't get emitted, then it's a bug that we need to investigate and fix. Have you checked if you have `config.experimental.enableMediaNegotiatedEvent` enabled? It looks like it's disabled by default

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
```diff
@@ -0,0 +1,163 @@
...
+      await bobJoin;
+      await chrisJoin;
+      await bobJoined;
+      await chrisJoined;
+
+      assert.exists(bob.meeting.joinedWith);
+      assert.exists(chris.meeting.joinedWith);
+    });
+
+    it('user "alice", "bob", and "chris" adds media', async () => {
```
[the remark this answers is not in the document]
InteractiveTimmy: missed the `adds` => `add` part. fixing.

packages/@webex/plugin-meetings/test/integration/spec/converged-space-meetings.js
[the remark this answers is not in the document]
InteractiveTimmy: image

Thanks for finding this. I've made adjustments and now it works as expected.

marcin-bazyl (approved):

marcin-bazyl (approved):

arun3528 (approved):

PR #2548, opened 2023-01-04
chore: fix prettier errors on the SDK

arun3528:
# COMPLETES # [SPARK-396392](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-396392)

## This pull request addresses

Adds eslint rules for both ts and js file , have upgraded the eslint to more recent standards, 

Most of the changes are automatically added from eslint and prettier 

disabled some of the files using decorators for eslint 

confirmed if the unit test are failing checked against the master and they are the same 

https://app.circleci.com/pipelines/github/webex/webex-js-sdk/9335/workflows/a8dfd7a8-12f5-4e9f-a5d5-dc006fb46361/jobs/56256 

https://github.com/webex/webex-js-sdk/pull/2554

## by making the following changes

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

### I certified that

- [ ] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [ ] I discussed changes with code owners prior to submitting this pull request

- [ ] I have not skipped any automated checks
- [ ] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #2611, opened 2023-02-10
fix: call disconnecting after calling updateMedia

arun3528:

@marcin-bazyl 

the roap bug from the "Participants being kicked out of real world converged meeting" space is a regression on SDK master that was introduced in the [prettier PR](https://github.com/webex/webex-js-sdk/pull/2548) - every time anyone calls Meeting.updateMedia() SDK sends the new Offer twice (increasing the seq each time), so as a result you will be dropped from the call.
This is because updateMedia() calls sendRoapMediaRequest() twice: the block between lines 4888 and 4913 should be removed - it's a duplication of the block in the next lines - looks like some merge that has gone wrong

removed the merged code

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4905,32 +4905,6 @@ export default class Meeting extends StatelessWebexPlugin {
             // we actually get a OFFER from the server and a GLAR condition happens
             if (startShare) {
               // We are assuming that the clients are connected when doing an update
-              // @ts-ignore
-              return this.share();
-            }
-
-            return Promise.resolve();
-          })
-          .then(() =>
-            logRequest(
-              this.roap.sendRoapMediaRequest({
-                sdp: this.mediaProperties.peerConnection.sdp,
-                roapSeq: this.roapSeq,
-                meeting: this, // or can pass meeting ID
-              }),
-              {
-                header: `${LOG_HEADER} sendRoapMediaRequest being sent`,
-                success: `${LOG_HEADER} sendRoadMediaRequest successful`,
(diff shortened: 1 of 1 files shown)


marcin-bazyl (approved):
`Meeting.updateMedia()` has almost no unit tests, so this kind of regression could easily slip through. Might be worth adding a basic test, although `updateMedia()` is going to be deleted on beta branch soon anyway.

torgeadelin (approved):
Same as marcin, unit test would be good.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4900,32 +4900,6 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            // This is a special case if we do an /floor grant followed by /media
-            // we actually get a OFFER from the server and a GLAR condition happens
-            if (startShare) {
-              // We are assuming that the clients are connected when doing an update
-              // @ts-ignore
-              return this.share();
-            }
-
-            return Promise.resolve();
-          })
```
torgeadelin, line 4913: was this bit removed on purpose too?
i see that the second .then is a duplicate.

arun3528: we changed the name of the function which cause a merge conflict , accidentally we kept both the code causing the issue , so basically they are the same

arun3528:
> `Meeting.updateMedia()` has almost no unit tests, so this kind of regression could easily slip through. Might be worth adding a basic test, although `updateMedia()` is going to be deleted on beta branch soon anyway.

yes marcin , i am setting up a plan for the beta, going to skip the test for now as we are going to freeze the master for now only major bug fixes

PR #2613, opened 2023-02-10
fix(meetings): types distributable

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to fix the output directory for `*.d.ts` files within the `@webex/plugin-meetings` package. This moves the types to distribute in parallel to their built source distributable. This solves the issue of types being moved between branches when switching branches. This also solves the build issues that generate from the aforementioned issue.

Screenshot 2023-02-10 at 12 02 02 PM


packages/@webex/plugin-meetings/package.json
```diff
@@ -12,7 +12,7 @@
     "Taymoor Khan <taykhan@cisco.com>"
   ],
   "main": "dist/index.js",
-  "types": "types/index.d.ts",
```
arun3528, line 15: so both the types and code are in the same folder we dont need seperate types folder !?

arun3528 (approved):

PR #2615, opened 2023-02-13
feat(breakout): add broadcast, askAllReturn methods

othellashao:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-400365

Adds broadcast and askAllReturn for Breakout (Host/Cohost part)

add broadcast into breakout module
add askAllReturn into breakout module

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2694,6 +2694,39 @@ function viewBreakouts(event) {
...
+    button.onclick = () => {
+      breakoutSession.askForHelp();
+    };
+
+    return button;
+  };
+  const createAskAllReturnButton = (breakoutSession) => {
+    const button = document.createElement('button');
+
+    button.innerText = 'Return';
```
Coread, line 2711: suggestion: I think something like "Ask to return" would be clearer than just "Return" here as that sounds very similar to Leave.

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -105,6 +106,34 @@ const Breakout = WebexPlugin.extend({
...
+      id: this.groupId,
+      recipientRoles: roles.length ? roles : undefined,
+      sessions: [
+        {
+          id: this.sessionId,
+        },
+      ],
+    };
+
+    return this.webex.request({
```
Coread, line 128: issue: When there is nobody in a breakout session, the API errors with something like the following:
```json
{
    "errorCode": 201409036,
    "message": "Breakout message no valid recipients found",
    "errors": [
        {
            "errorCode": 201409036,
            "description": "Breakout message no valid recipients found"
        }
    ],
    "trackingId": "webex-js-sdk_36e33064-32fb-418e-a202-0bddd1e3513f_22"
}
```

We have two possibilities here
1) Catch the error and throw a suitable Error, one that can be caught by the consuming app.
2) Catch and hide this error here. i.e. if the error is a 429 and the error code is 201409036, catch and ignore.

Given it is a reasonably common scenario, and that the desktop clients also suppress this error, I suggest we do option 2.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2694,6 +2694,39 @@ function viewBreakouts(event) {
...
+    };
+
+    return button;
+  };
+  const createBroadcastButton = (breakoutSession) => {
+    const button = document.createElement('button');
+
+    button.innerText = 'broadcast';
+
+    button.onclick = () => {
```
Coread, line 2724: suggestion: It would be nice if it were possible to test the cohost/presenter logic in the sample app. Perhaps tickboxes for those options?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -170,6 +174,13 @@ const Breakouts = WebexPlugin.extend({
    */
   updateBreakout(params) {
     this.set(params);
+    if (params.sessionType === BREAKOUTS.SESSION_TYPES.MAIN) {
```
Coread, line 177: issue: It is not necessary to store the main session ID or main group ID separately. Either the user is in the main session, in which case the sessionId, groupId are on this object or they are in a session, in which case you can filter the breakouts by isMain to determine which groupId, sessionId to use.

A function similar to this could be used to retrieve the information:

```ts
  getMainSession() {
    if (this.isInMainSession) {
      return this;
    }

    const mainSession = this.parent.breakouts.filter((breakout) => breakout.isMain)[0];

    return mainSession;
  },
```

There is similar logic used in the leave function in breakout,ts so this function could possibly be shared.

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -170,6 +174,13 @@ const Breakouts = WebexPlugin.extend({
...
+      this.set({
+        mainGroupId: params.groupId,
+        mainSessionId: params.sessionId,
+        breakoutGroupId: params.groups && params.groups.length ? params.groups[0].id : '',
```
Coread, line 181: question: When I looked at this value, it seems to be the same as the mainGroupId. Are you sure we actually need breakoutGroupId?

Coread (changes requested):
Please see comments regarding 429 error handling and storing the group/session ID for the main session.

packages/@webex/plugin-meetings/src/breakouts/index.ts
othellashao: need the breakoutGroupId (should not same with mainGroupId) when the host in bo session, and broadcast to all. I'm going to debug it again and try to get the groupId through the breakout collection instead of adding a new prop.

docs/samples/browser-plugin-meetings/app.js
othellashao: Done for it

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
othellashao: Got it. Catch the error 201409036 & 409 -> add a log and ignore

packages/@webex/plugin-meetings/src/breakouts/index.ts
othellashao: Done. Please help to check the PR again

packages/@webex/plugin-meetings/src/breakouts/index.ts
othellashao: I'd removed these code and get the main groupId and breakout group Id from current session and breakouts collection

docs/samples/browser-plugin-meetings/app.js
Coread: breakouts_dropdown

At the moment the broadcast button just sends a broadcast message to all the sessions and all the participants. If you see the above screenshot, we are unable to test sending to presenters/cohosts/hosts in the different combinations from the sample app. We are also unable to test sending a message to a particular breakout session

Coread (changes requested):
Please see comment about unit tests for error handling. Please also see suggestion regarding factoring out the broadcast method to a separate stateless plugin to remove duplication in both the code and the tests you will need to write

docs/samples/browser-plugin-meetings/app.js
othellashao: Got it. 👍

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -105,6 +107,42 @@ const Breakout = WebexPlugin.extend({
...
     this.members.locusParticipantsUpdate(locus);
   },
+
+  /**
+   * Broadcast message to this breakout session's participants
+   * @param {String} message
+   * @param {Object} options
+   * @returns {Promise}
+   */
+  broadcast(message, options) {
```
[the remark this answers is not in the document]
othellashao: Great. I'd change follow this suggestion

Coread (approved):

mickelr (approved):

arun3528 (approved):

PR #2622, opened 2023-02-17
feat(legacy): elevate babel and eslint

InteractiveTimmy:
# DO NOT MERGE --POC

# Description

The scope of the changes in this pull request are to demonstrate how various configurations and tooling can be elevated into packages to be consumed in an independent way. This allows for various packages to have explicit build tooling based on their requirements.


arun3528 (commented):
can we use the yarn 4.0 rc version not sure if it makes too much difference

InteractiveTimmy:
> can we use the yarn 4.0 rc version not sure if it makes too much difference

We don't benefit from any of the features of the 4.0 RC at this time, we should use the most recent, stable, version of yarn instead of a pre-release version for stability.

.gitignore
```diff
@@ -11,7 +11,7 @@ tmp*.*
 
 _site
 .coverage
-/node_modules
+**/node_modules
```
arun3528, line 14: there should not be an node_modules in the packages , do you see any ?

packages/@webex/internal-plugin-flag/package.json
```diff
@@ -29,5 +33,12 @@
     "@webex/test-helper-test-users": "workspace:^",
     "@webex/webex-core": "workspace:^",
     "lodash": "^4.17.21"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.17.10",
+    "@webex/babel-config-legacy": "workspace:^",
+    "@webex/eslint-config-legacy": "workspace:^",
+    "eslint": "^8.24.0",
+    "prettier": "^2.7.1"
```
arun3528, line 42: does it throw error for prettier , shouldn't the parent module take care of it

packages/legacy/babel/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "packageManager": "yarn@3.4.1",
+  "engines": {
+    "node": ">=18.0.0",
+    "npm": ">=8.0.0",
+    "yarn": ">=3.0.0"
+  },
+  "type": "commonjs",
+  "main": "./static/index.js",
+  "exports": {
+    ".": "./static/index.js",
```
arun3528, line 13: so this package will take care of babel config and version bump 
Will build legacy be a different one ?

packages/legacy/babel/static/plugins/inject-package-version.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+module.exports = function injectPackageVersion() {
+  return {
+    visitor: {
+      /**
+       * Adds a version property to all SparkPlugins
+       * @param {Object} path
+       * @returns {void}
+       */
+      CallExpression(path) {
```
arun3528, line 22: when there is a release for each package it points to the path

packages/legacy/babel/static/plugins/inject-package-version.js
```diff
@@ -0,0 +1,55 @@
...
+        if (t.isMemberExpression(path.get('callee'))) {
+          if (
+            path.node.callee.object.name === 'WebexPlugin' &&
+            path.node.callee.property.name === 'extend'
+          ) {
+            const def = path.node.arguments[0];
+            const visited = def.properties.reduce(
+              (acc, p) => acc || t.isObjectProperty(p, {key: 'version'}),
```
arun3528, line 30: Can we add comments on what it exactly does , little confused with the names

arun3528 (approved):

.gitignore
InteractiveTimmy: This is needed in the case that a user adds a new dependency that conflicts with the global workspace dependencies. In this case, it will add a `./node_modules` folder within the scope of the package. This prevents those from being committed.

packages/@webex/internal-plugin-flag/package.json
InteractiveTimmy: It does not, we *can* safely remove it here. I will do this.

packages/legacy/babel/package.json
InteractiveTimmy: Yes, this babel config is a global config that is used by ESLint and out build workflows. This needs to be a separate package for this reason.

* `./legacy/build` => consumes `./legacy/babel`
* `./legacy/eslint` => consumes `./legacy/babel`

packages/legacy/babel/static/plugins/inject-package-version.js
InteractiveTimmy: This can be updated as a part of the `build` task in JIRA. This config was placed here specifically for ESLint to use.

packages/legacy/babel/static/plugins/inject-package-version.js
InteractiveTimmy: This was copy-pasted from the root config, I do not have any idea what this does at the moment, but, it was necessary as a part of the core config mounting flow for ESLint.

PR #2625, opened 2023-02-21
fix(plugin-meetings): update degradation logic to use live sources

brycetham:
# COMPLETES [#SPARK-403056](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-403056)

Updating the automatic max-fs degradation logic to use the number of live sources instead of number of sources requested.

When retrieving the degraded client requests, only consider sources with that are in the "live" state. Then, whenever a source state has been updated, resend the client requests.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Manually tested joining with both unified web client and unified native client, muting and unmuting video on both clients.
2. Added unit test for media request with non-live source states.

### I certified that

---


packages/@webex/plugin-meetings/src/multistream/receiveSlotManager.ts
```diff
@@ -71,6 +71,15 @@ export class ReceiveSlotManager {
       (csi: CSI) => this.meeting.members.findMemberByCsi(csi)?.id
     );
 
+    // update media requests on source update
+    if (getMediaFamily(mediaType) === MediaFamily.Video) {
+      receiveSlot.on(ReceiveSlotEvents.SourceUpdate, () => {
+        this.meeting.mediaRequestManagers[
```
marcin-bazyl, line 77: hmm... it's not ideal that ReceiveSlotManager now has a dependency on MediaRequestManager... maybe it would be better the other way around? ReceiveSlotManager doesn't know anything about media requests, but MediaRequestManager is already aware of receive slots (because they're part of media requests) so it would make more sense for MediaRequestManager to depend on ReceiveSlotManager and just listen on some event, maybe even MediaRequestManager can just listen for SourceUpdate events directly on the receive slots that are used in the media requests it manages?
Also I would rather avoid having the logic of doing this only for videos here in ReceiveSlotManager. I think that logic belongs in MediaRequestManager or could be configured for the right instance of MediaRequestManager by the meeting object.

brycetham: Yeah, I think it makes sense inside of `MediaRequestManager`. I'm thinking maybe I can listen to the event in `addRequest` and remove the listener in `cancelRequest`. The degradation logic already only degrades video requests, so I can probably remove the logic to check for the media family.

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -684,6 +687,30 @@ describe('MediaRequestManager', () => {
     assert.calledOnce(sendMediaRequestsCallback);
   });
 
+  it('should not degrade max-fs if receive slot sources are not live', () => {
```
marcin-bazyl, line 690: it would be good to add a test that checks that source update triggers new degradation evaluation and new media request

packages/@webex/plugin-meetings/src/multistream/receiveSlotManager.ts
marcin-bazyl: yeah it's much better now, but I'm worried a bit about us sending excessive number of media requests now... I'll 1-1 with you about the details

marcin-bazyl (approved):

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
brycetham: Ok, will do it in a later PR, since we want to get this in today.

arun3528 (approved):

PR #2635, opened 2023-02-28
feat(reachability): add clientMediaIP

torgeadelin:
# COMPLETES  [SPARK-382440](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-382440)

-add clientMediaIP to reachability result

- collect public apis from ice candidates and store them, and eventually add them in the reachability result
 
### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- unit tests, manual test
- 
### I certified that

---


packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -430,6 +456,29 @@ export default class Reachability {
...
+    if (peerConnection.connectionState === CLOSED) {
+      LoggerProxy.logger.log(
+        `Reachability:index#setClientMediaIPs --> Attempting to set clientMediaIP of ${clientMediaIP} on closed peerConnection.`
+      );
+    }
+
+    // Set to null in case this fired from
+    // an event other than onIceCandidate
+    peerConnection.onicecandidate = null;
+    peerConnection.close();
```
marcin-bazyl, line 477: this is not needed here, it's done in setLatencyAndClose because setLatencyAndClose() is supposed to be the last thing that is called and closes the connection

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -430,6 +456,29 @@ export default class Reachability {
...
+    // @ts-ignore
+    peerConnection.clientMediaIP = clientMediaIP;
```
marcin-bazyl, line 479: there can be multiple of these IPs  (if the computer has multiple network adapters connected to different networks), Orpheus wants to know all of them. So it probably also makes sense to rename setClientMediaIPs() to addClientMediaIPs() or actually addPublicIP() (see my other comment)

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -14,6 +14,7 @@ import ReachabilityRequest from './request';
 const DEFAULT_TIMEOUT = 3000;
 const VIDEO_MESH_TIMEOUT = 1000;
 
+type IceCandidateResult = {clusterId: string; elapsed?: string; clientMediaIP?: string};
```
marcin-bazyl, line 17: I think I would prefer a name like "publicIP" or "publicMediaIp" so it's more clear to people reading the code that don't know much about ICE, webrtc etc

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -319,12 +323,19 @@ export default class Reachability {
...
-            return true;
+          switch (property) {
+            case ELAPSED:
+              // @ts-ignore
+              resolve({clusterId: peerConnection.key, elapsed: value});
+
+              return true;
+            case CLIENT_MEDIA_IP:
+              // @ts-ignore
+              resolve({clusterId: peerConnection.key, clientMediaIP: value});
```
marcin-bazyl, line 334: I don't think this is needed. The promise is resolved when the "elapsed" property is set, so with the current way you have the code calling setClientMediaIPs() after setLatencyAndClose() then this line will do nothing, because the promise is already resolved. I think you can just leave this code block as it was and only resolve the promise when "elapsed" is set and don't do anything specials for CLIENT_MEDIA_IP and just make sure that setLatencyAndClose() is the last thing that is called.

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -344,6 +355,7 @@ export default class Reachability {
         // Close any open peerConnections
         if (peerConnectionProxy.connectionState !== CLOSED) {
           this.setLatencyAndClose(peerConnectionProxy, null);
+          this.setClientMediaIPs(peerConnectionProxy, null);
```
marcin-bazyl, line 358: better to call this before `this.setLatencyAndClose()` - see my other comment why, or maybe we don't even need to call this at all here

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -291,6 +292,7 @@ export default class Reachability {
           elapsed
         );
         this.setLatencyAndClose(peerConnection, elapsed);
+        this.setClientMediaIPs(peerConnection, e.candidate.address);
```
marcin-bazyl, line 295: it's better to call this before `this.setLatencyAndClose()` - see my other comments why

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: removed

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: renamed to `publicIP`

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: reverted, and made sure `setLatencyAndClose ` is last

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: switched the order.

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: switched the order

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: ooh i seeee. i've kind of tried to add the "multiple IPs" logic in the `parseIceResultsToReachabilityResults`, but it was wrong. Makes more sense here now, I was really confused yesterday..

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -391,6 +403,20 @@ export default class Reachability {
...
       };
+
+      if (clientMediaIP) {
+        if (!reachabilityMap[clusterId].udp.clientMediaIPs) {
+          reachabilityMap[clusterId].udp.clientMediaIPs = [clientMediaIP];
+        } else {
+          reachabilityMap[clusterId].udp.clientMediaIPs.push(clientMediaIP);
+        }
+
+        if (!reachabilityMap[clusterId].tcp.clientMediaIPs) {
```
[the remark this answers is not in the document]
torgeadelin: removed

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: but how will the value for clientMediaIP get through the promise in the end? if i revert this back, it won't be part of the iceResults and won't be able to add it to the reachabilityResult.

marcin-bazyl: you need to store it somehow and then include it in the resolve() call that's done in this line: https://github.com/webex/webex-js-sdk/blob/master/packages/%40webex/plugin-meetings/src/reachability/index.ts#L325
the current code stores the elapsed property on the peer connection object itself and I see you're doing the same storing it as `publicIPs` - I guess it's fine - I'm not a big fan of that approach but it's consistent with the existing code in this module.

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -440,6 +452,36 @@ export default class Reachability {
...
+    const {CLOSED} = CONNECTION_STATE;
+
+    if (peerConnection.connectionState === CLOSED) {
+      LoggerProxy.logger.log(
+        `Reachability:index#addPublicIPs --> Attempting to set publicIP of ${publicIP} on closed peerConnection.`
+      );
+    }
+
+    if (publicIP) {
+      // @ts-ignore
```
marcin-bazyl, line 471: maybe instead of so many ts-ignore we could cast the peerConnection to an object that has `publicIPs` property once at the beginning?

marcin-bazyl (approved):

PR #2636, opened 2023-03-01
feat(package): add cli tooling

InteractiveTimmy:
# **DRAFT -- DO NOT MERGE!**

# Description

The scope of the changes in this pull request are to set up the foundations for core JS SDK configuration packages for modern implementation. This also includes the mounting of a new `@webex/cli-tools` package that will be consumed when setting up deploy workflows in CI/CD.


chrisadubois (commented):
my only prob with this PR is that jasmine is now in conjunction with mocha/chai/jest

else than that, looks good and mostly cleans the configs up and structures them into packages

packages/config/eslint/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+const config = {
+  root: true,
+  env: {
+    node: true,
+  },
+  extends: ['@webex/eslint-config/core'],
```
arun3528, line 6: this will include all the default rules we had and additional rules

arun3528: what is the difference between the legacy eslint and the new eslint package ?

packages/tools/cli/api-extractor.config.js
```diff
@@ -0,0 +1,3 @@
+const build = require('@webex/api-extractor-config');
+
+build();
```
arun3528, line 3: do we still need build or can we just use the cli command ?

arun3528 (approved):
thanks tim

PR #2649, opened 2023-03-06
feat(config): update configurations

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to update the core configurations within the JS SDK's sub-package: config.

This includes various updates to linters, builders, and the addition of ESBuild as an option for rapid building workflows in modern packages.


packages/config/esbuild/static/cli.js
```diff
@@ -0,0 +1,14 @@
+const cli = {
+  bundle: true,
+  entryPoints: [
+    './src/main.ts',
```
arun3528, line 4: shouldn't this be index,js?

packages/config/esbuild/static/cli.js
```diff
@@ -0,0 +1,14 @@
...
+  ],
+  format: 'cjs',
+  minify: true,
+  outfile: './dist/cli/index.js',
+  platform: 'node',
```
arun3528, line 9: is node here because  of esm version

packages/config/esbuild/static/cli.js
InteractiveTimmy: Not for `CLI`, cli needs a different entry point because it's an IEF [immediately executing file].

packages/config/esbuild/static/cli.js
InteractiveTimmy: Since this is for building a CLI execution app, which can only occur in node, node is used.

arun3528 (approved):

PR #2653, opened 2023-03-07
feat: provide size hint remote media

Coread:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-385727

Adds size hint API to multistream

The width and height of a video element can be given to the remote media object which causes a request to be sent respecting the frame size calculated from the width and height.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -1934,7 +1934,7 @@ async function getStatsForVideoPane(meeting, videoPane) {
   return result;
 }
 
-const remoteMediaIds = {};
+let remoteMediaIds = {};
 
 function setSizeHint() {
   const sizeHintValue =  document.getElementById('size-hint-input').value;
@@ -1954,6 +1954,10 @@ function addRemoteMediaOption(remoteMedia) {
   option.text = remoteMedia.id;
 
   select.add(option);
+}
+
+function clearRemoteMedia() {
+  remoteMediaIds = {};
 }
 
 function processNewVideoPane(meeting, paneGroupId, paneId, remoteMedia) {
@@ -2053,6 +2057,7 @@ function setupMultistreamEventListeners(meeting) {
     console.log('memberVideoPanes:', memberVideoPanes);
     console.log('screenShareVideo:', screenShareVideo);
 
+    clearRemoteMedia();
     currentVideoPaneList.length = 0;
     clearAllMultistreamVideoElements();
     createVideoElementsForLayout(layoutId);
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -1934,8 +1934,32 @@ async function getStatsForVideoPane(meeting, videoPane) {
   return result;
 }
 
+const remoteMediaIds = {};
+
+function setSizeHint() {
+  const sizeHintValue =  document.getElementById('size-hint-input').value;
+  const remoteMediaId = document.getElementById('remote-media-selector').value;
+
+  let [width, height] = sizeHintValue.split(',');
+
+  remoteMediaIds[remoteMediaId].setSizeHint(parseInt(width), parseInt(height))
+}
+
+function addRemoteMediaOption(remoteMedia) {
+  remoteMediaIds[remoteMedia.id] = remoteMedia;
+  const select = document.getElementById('remote-media-selector');
+
+  const option = document.createElement('option');
+  option.value = remoteMedia.id;
+  option.text = remoteMedia.id;
+
+  select.add(option);
+}
+
 function processNewVideoPane(meeting, paneGroupId, paneId, remoteMedia) {
   const videoPane = allVideoPanes[paneGroupId][paneId];
+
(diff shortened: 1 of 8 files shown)


packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -56,6 +57,8 @@ const CODEC_DEFAULTS = {
   },
 };
 
+const DEBOUNCED_SOURCE_UPDATE_TIME = 10;
```
marcin-bazyl, line 60: we probably need a bit bigger value than 10ms. The app will call size hint API a lot when the user resizes the app window, so maybe we should have 1000ms here?

packages/@webex/plugin-meetings/src/multistream/remoteMedia.ts
```diff
@@ -102,6 +102,29 @@ export class RemoteMedia extends EventsScope {
...
+   * @param height height of the video element
+   */
+  public setSizeHint(width, height) {
+    // only base on height for now
+    let fs: number;
+
+    if (height < 135) {
+      fs = 60;
+    } else if (height < 270) {
+      fs = 225;
```
marcin-bazyl, line 118: let's use same values we have in getMaxFs(), so 240 here

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -213,6 +223,10 @@ export class MediaRequestManager {
 
     mediaRequest.receiveSlots.forEach((rs) => {
       rs.on(ReceiveSlotEvents.SourceUpdate, this.sourceUpdateListener);
+      rs.on(ReceiveSlotEvents.MaxFsUpdate, ({maxFs}) => {
+        mediaRequest.preferredMaxFs = maxFs;
```
marcin-bazyl, line 227: a media request will have a number of receive slots, so here we're just storing the maxFs we got from only one of them (the last one that emitted the event). Shouldn't we store the minimum? Or is this assuming that all receive slots within a group will have same size hints?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1934,9 +1934,33 @@ async function getStatsForVideoPane(meeting, videoPane) {
   return result;
 }
 
+const remoteMediaIds = {};
```
marcin-bazyl, line 1937: we need to clear this on layout change

packages/@webex/plugin-meetings/src/multistream/remoteMedia.ts
```diff
@@ -102,6 +102,29 @@ export class RemoteMedia extends EventsScope {
...
+    } else if (height < 540) {
+      fs = 920;
+    } else {
+      fs = 3600;
+    }
+
+    this.receiveSlot.setMaxFs(fs);
```
marcin-bazyl, line 125: this.receiveSlot can be undefined.

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
Coread: I'm assuming that all receive slots in a group will have the same size hints. It is a necessary assumption as the alternative of storing the minimum would mean that you would degrade a large video within a group because of the presence of a small video. Basically you can't have a group with different sized panes in it.

docs/samples/browser-plugin-meetings/app.js
Coread: I'm now clearing this

packages/@webex/plugin-meetings/src/multistream/remoteMedia.ts
Coread: changed to use 240

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
Coread: Changed to be 1000ms

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMedia.ts
```diff
@@ -222,4 +224,24 @@ describe('RemoteMedia', () => {
...
+        {height: 269, fs: 240},
+        {height: 270, fs: 920},
+        {height: 539, fs: 920},
+        {height: 540, fs: 3600},
+      ],
+      ({height, fs}) => {
+        it(`sets the max fs to ${fs} correctly when height is ${height}`, () => {
+          remoteMedia.setSizeHint(100, height);
+
+          assert.calledOnceWithExactly(fakeReceiveSlot.setMaxFs, fs);
```
shnaaz, line 242: Do you think there should be a test when receiveSlot is undefined, that is then it is not called

Coread: I've added a test for this

shnaaz (approved):

PR #2658, opened 2023-03-09
feat: Host/Cohost - Admit user in lobby to main meeting from Breakout Session SPARK-400367

BBigLee:
# COMPLETES #
[SDK - Host/Cohost - Admit user in lobby to main meeting from Breakout](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-400367)

The Control Admit API allows a host or cohost to admit users who are in the lobby to the main meeting. The host or cohost may be in main or in a Breakout session. If the host or cohost is in a Breakout session, then they must specify the authorizingLocusUrl of the Breakout that the host or cohost is in. The API accepts one or more participantIds to admit from the lobby.
- Admit API accepts one or more participantIds to admit from the lobby to main.
- Admit API may be invoked by host/cohost in main or by host/cohost in a Breakout session.
- If the host/cohost is in a Breakout session, they must specify the authorizingLocusUrl which will be the Locus URL of the Breakout.
- Regardless of host/cohost location, the Locus Id (lid) in the path should be the Locus Id of the main.

< DESCRIBE YOUR CHANGES >
1. Added logics for Admit API to allow host/cohost from breakout session admits participants to main meeting
2. Added unit tests and updated existed unit tests
3. Logic optimization
4. Added examples in sample page
5. Added a new prop in object Breakout: mainLocusUrl

image
image
image
image

### Change Type

- [✓] New feature (non-breaking change which adds functionality)
- [✓] Breaking change (fix or feature that would cause existing functionality to change)
- [✓] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

- [✓] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [✓] I discussed changes with code owners prior to submitting this pull request
- [✓] I have not skipped any automated checks
- [✓] All existing and new tests passed
- [✓] I have updated the documentation accordingly

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -2679,14 +2679,32 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * @param {Object} sessionLocusUrls: {authorizingLocusUrl, mainLocusUrl}
    * @returns {Promise} see #members.admitMembers
    * @public
    * @memberof Meeting
    */
-  public admit(memberIds: Array<any>) {
-    return this.members.admitMembers(memberIds);
+  public admit(
+    memberIds: Array<any>,
+    sessionLocusUrls?: {authorizingLocusUrl: string; mainLocusUrl: string}
```
Coread, line 2697: issue: I am worried about the additional complexity here on the admit method. As a user of the SDK I shouldn't need to worry about the details i.e. what the authorizingUrl is. Is it possible to work these parameters out so the user can still call admit(memberIds)?

BBigLee: Yes, I agree. Actually, our users can ignore this argument, and the SDK should handle the details. As you can see, sessionLocusUrls is optional, so it can be omitted. And if this argument is not passed in, we have logics to cover the case. So, yes, it is fine to work these parameters out so the user can still call admit. But in case of some special requirements or situations, we allow users to pass sessionLocusUrls argument to fit their businesses.

BBigLee: On the sample page, the admitting action  just calls  meeting.admit([participantID]), and it works well
image

Coread (approved):

arun3528 (approved):

PR #2668, opened 2023-03-13
feat: assign attendees to bo session

lhy2143:
# COMPLETES 
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-400360

host/cohost can assign attendees to a breakout session

add a new function named assign and add two buttons(assign/move) on sample page to test the function
add unit test for assign function

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

join meeting from sample page (with support breakout session checkbox checked) -> create bo session -> start bo session ->  click assign button (will see popover in native)-> click join (will join the bo session) -> click move button on breakout session 1 (will move from session 1 to session 2)

### I certified that

---


packages/@webex/plugin-meetings/src/breakouts/breakout.ts
```diff
@@ -111,6 +111,31 @@ const Breakout = WebexPlugin.extend({
   },
 
   /**
+   * assign participants to breakout session
+   * @param {Array} sessions
+   * @param {object} groupPayload
+   * @returns {void}
+   */
+  assign(sessions: any[], groupPayload: any = {}) {
```
Coread, line 119: issue: Given that this is the public API, since you have used it in the sample app, there is too much state passed in or hard coded.

The breakout.assign method should only take a member (or members). The other attributes necessary to control the groups should already be stored on the breakout object and should not be changed during assigning.

i.e. allowBackToMain, allowToJoinLater etc, all of these should not be hardcoded. 

Additionally, the groupPayload should not be passed in. Only members should be passed in. The breakout object is a 1-1 mapping of a session. It has a sessionId already. You shouldn't need to pass in sessions either.

Coread (changes requested):
Please see comment regarding the design of this as a public API.

packages/@webex/plugin-meetings/src/breakouts/breakout.ts
Coread, line 1: There is still a problem with the way this is done. The user of this API should not have to care about what a session is. At the moment they have to pass in sessions, when each breakout already represents one session.

The assign api for a breakout should take an array of memberIds and the session object should be created in the assign function.
i.e.
```ts
assign(memberIds: string[]) {
  // only one session, because each session can be set separately
  const sessions = [
    {
      id: this.sessionId,
      assigned: memberIds,
    },
  ];
   return this.request({
     method: HTTP_VERBS.PUT,
     uri: this.url,
     body: {
        groups: [
          {
            id: this.groupId,
            sessions,
          },
        ],
      },
  });
}
```

note that we don't attempt to set the name of the session or the anyoneCanJoin parameter because we aren't changing them, only changing who is assigned.

Coread (changes requested):
please see comment

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -506,6 +506,30 @@ const Breakouts = WebexPlugin.extend({
 
     return breakout;
   },
+
+  /**
+   * assign participants to breakout session
+   * @param {Object} params
```
Coread, line 512: params needs a better description. These doc strings should let the user know how to call the function

packages/@webex/plugin-meetings/src/breakouts/README.md
```diff
@@ -134,6 +130,10 @@ breakout.rename(newSessionName)
 // Start breakout sessions with necessary params
 breakouts.start(params)
 
+// Assign a particpant/participants to breakouts/a breakout
+// sessions should be like: [{id:'xx',memberIds:['111'],emails:['111@ss.com']},...]
```
Coread, line 134: would be nice to add "where id is the id of the breakout"

packages/@webex/plugin-meetings/test/unit/spec/breakouts/index.ts
```diff
@@ -703,5 +736,15 @@ describe('plugin-meetings', () => {
         );
       });
     });
+
+    describe('#assign', () => {
+      it('assign attendee to a breakout session', async () => {
+        breakouts.assign = sinon.stub().returns(Promise.resolve('ASSIGN_RETURN_VALUE'));
+        const params = {assigned: []};
```
Coread, line 743: there needs to be a tests for the following:
1) emails but no member ids
2) member ids but no emails

Coread (approved):

PR #2683, opened 2023-03-17
fix(remote-video-mute): fixed remote video mute

chburket:
# COMPLETES #

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-413032

Really not sure how I pushed a reverted version of a line of code.  Only thing I can think of is that I did something wrong during merge/rebase.

Fixed the event that is being emitted.


InteractiveTimmy (approved):
LGTM!

PR #2684, opened 2023-03-17
fix: throw error when sync meetings fails

brycetham:
# COMPLETES #[SPARK-393171](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-393171)

Incorrectly leaving the meeting when failing to sync meetings during client media failover.

Throwing an error when sync meetings fails so that, when caught, the reconnection is retried.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Manually tested disconnecting from the network for 15 seconds, then reconnecting. During the reconnection process, disconnect from the network again while syncing meetings, and reconnect again.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1108,33 +1108,41 @@ export default class Meetings extends WebexPlugin {
    * @memberof Meetings
    */
   public syncMeetings() {
-    return this.request.getActiveMeetings().then((locusArray) => {
-      const activeLocusUrl = [];
+    return this.request
+      .getActiveMeetings()
+      .then((locusArray) => {
+        const activeLocusUrl = [];
 
-      if (locusArray?.loci && locusArray.loci.length > 0) {
-        locusArray.loci.forEach((locus) => {
-          activeLocusUrl.push(locus.url);
-          this.handleLocusEvent({
-            locus,
-            locusUrl: locus.url,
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
brycetham, line 1: The only change here is adding the "catch" block. Everything else is just indentation.

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1108,33 +1108,41 @@ export default class Meetings extends WebexPlugin {
...
           }
         }
-      }
-    });
+      })
+      .catch((error) => {
+        LoggerProxy.logger.error(
+          `Meetings:index#syncMeetings --> failed to sync meetings, ${error}`
+        );
+        throw new Error('Failed to sync meetings');
```
marcin-bazyl, line 1144: wouldn't it be better to throw the original `error` - it may contain some more useful detailed information that we'll lose by creating a new one.

arun3528: +1

arun3528: if i am not wrong the catch was not here because , some logic was calling syncMeeting before it started a new meeting 

if sync fails we should not stop the user to make calls was the intention I believe

marcin-bazyl (approved):

arun3528 (approved):

PR #2693, opened 2023-03-22
feat (Beta): get pre-assignments info for breakout session

csshz:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-400364

get pre-assignments info for breakout session

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


Coread (changes requested):
Can you add a test for the meeting util please.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -2577,6 +2577,12 @@ export default class Meeting extends StatelessWebexPlugin {
       );
     });
 
+    // We need to reinitialize  when user upgrades to host or cohost
+    this.locusInfo.on(LOCUSINFO.EVENTS.SELF_MODERATOR_OR_COHOST_UPGRADE, (payload) => {
+      this.breakouts.queryPreAssignments(payload);
+      // ...
```
Coread, line 2583: question: Does the ... mean there is more to come here? What else will happen in this event handler?

csshz: At present, this method handles just follow this process. when the attendee is upgraded to cohost or moderator, he can obtain the pre-assign list and operate the breakout session with high authority.

csshz:
> Can you add a test for the meeting util please.



> Can you add a test for the meeting util please.



> Can you add a test for the meeting util please.

Lost this part, now done, thanks ~

csshz:
the pre-assign feature added three users cases
	1.	Moderator join meeting directly.
	2.	Attendee joined the meeting and then upgrade to moderator & cohost.
	3.	breakout session enable switch from disabled to enabled.
All cases are non-repeatable and can only trigger pre-assign once.

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
```diff
@@ -70,6 +71,7 @@ SelfUtils.parse = (self: any, deviceId: string) => {
 };
 
 SelfUtils.getBreakouts = (self) => self?.controls?.breakout?.sessions;
+SelfUtils.isModerator = (self) => !!self?.moderator;
```
Coread, line 74: Question: What is this for? I can't see it being used elsewhere

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -38,7 +39,7 @@ const Breakouts = WebexPlugin.extend({
     mainLocusUrl: 'string', // the locus url of the main session
     groups: 'array', // appears when create breakouts
   },
-
+  isCallPreassignments: false,
```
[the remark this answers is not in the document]
csshz: This parameter is mainly to control the pre-assign API to only call once. No matter how complicated the user behavior , we only need to call once after the user’s identity is clarified. These cases can be triggered multiple times. For example, the user upgrades and downgrades repeatedly , switch the current breakout enable or disable, so the life cycle of this parameter is the same as the current breakout session. When the breakout session is reset, then we can call it again.
The main purpose is to reduce the useless API on the network. This list is set when the meeting was scheduled. We only need to take it when it is time to take it, and take it once.

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -41,6 +42,16 @@ MeetingUtil.parseLocusJoin = (response) => {
...
   });
 
+  // We need to confirm whether to call queryPreAssignments
+  if (
+    parsed.moderator &&
+    parsed.controls.breakout &&
+    parsed.controls.breakout.enableBreakoutSession &&
+    parsed.controls.breakout.hasBreakoutPreAssignments
+  ) {
+    meeting.breakouts.queryPreAssignments();
```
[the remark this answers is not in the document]
csshz: good suggestions, I will do later.

packages/@webex/plugin-meetings/src/breakouts/index.ts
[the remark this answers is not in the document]
Coread: Ok, thanks for the explanation. Can you change it to use the model attributes rather than add an attribute to the class? i.e. can you put the variable in to props? I also think a name like `shouldFetchPreassignments` would be a better name to describe what it is doing.

packages/@webex/plugin-meetings/src/meeting/util.ts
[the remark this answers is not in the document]
csshz: now above code have been removed, according to the native client logic, we just cover two cases for pre-assign logic. so we ignore the join meeting process.

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -38,7 +39,7 @@ const Breakouts = WebexPlugin.extend({
     mainLocusUrl: 'string', // the locus url of the main session
     groups: 'array', // appears when create breakouts
   },
-
+  isCallPreassignments: false, // Controlling the lifecycle of the pre-assign API
```
Coread, line 42: issue: I think you misunderstood the request here. It is better to put the attributes of a breakouts object in the props so they can be get/set correctly, rather than using a flag on the class itself i.e.
```ts
    mainLocusUrl: 'string', // the locus url of the main session
    groups: 'array', // appears when create breakouts
    shouldFetchPreassignments: 'boolean', // Controlling the lifecycle of the pre-assign API
```

then later
```ts
this.set('shouldFetchPreassignments', true);
```

csshz: Yes, I lost your mention before, let's follow this rule which object in the props so they can be get/set correctly.

Coread (approved):

PR #2694, opened 2023-03-22
fix: initial changes for using webrtc-core in multistream api

marcin-bazyl:
SDK multistream API to be using webrtc-core tracks

unit tests updated so they're passing, still need to update "converged-space-meeting" integration tests.

There are many todo comments referencing SPARK-399695 - this will be the next thing I'll work on after this gets merged. I tried to avoid doing too much refactoring in this PR to keep it as small as possible and reduce the risk of breaking transcoded meetings, more refactoring will be done in SPARK-399695

spark-399694


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1222,8 +1283,8 @@ function toggleBNR() {
   console.log('BNR not supported');
 }
 
-let publishedLocalShareAudioTrack = null; // todo: stop and unset these on "unpublished" event (SPARK-399694)
-let publishedLocalShareVideoTrack = null;
+let publishedLocalShareAudioTrack;
```
marcin-bazyl, line 1286: todo: move these 2 into localMedia object

packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * @returns true if user is allowed to unmute the track, false otherwise
+   */
+  isUnmuteAllowed() {
+    return this.unmuteAllowed;
+  }
+
+  setMuted(muted: boolean): Promise<void> {
```
marcin-bazyl, line 39: webrtc-core's setMuted returns void. I've changed it here to return a promise, because it might fail if you're not allowed to unmute, so I think it's nicer that it rejects or resolves now, but the consequence is that many things become async now (see other changes in this PR like Meeting.preMedia() etc) so I'm not 100% sure if it's a good idea to return a promise here. The alternative is to return void, but throw if you're not allowed to unmute.

packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable @typescript-eslint/no-misused-new */
+/* eslint-disable valid-jsdoc */
+/* eslint-disable require-jsdoc */
+import {
+  AudioDeviceConstraints,
+  createCameraTrack as wcmeCreateCameraTrack,
```
arun3528, line 6: why does it have to be named as wcmeCreateCameraTrack

there is no change in the functionality

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6740,7 +6771,10 @@ export default class Meeting extends StatelessWebexPlugin {
           error
         );
       } finally {
-        this.setLocalShareTrack(null);
+        // todo: once SPARK-399695 is done, we will be able to just call this.setLocalShareTrack(null); here instead of the next 2 lines:
+        this.mediaProperties.shareTrack?.off(LocalTrackEvents.Ended, this.handleShareTrackEnded);
```
arun3528, line 6775: nice

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7225,23 +7270,45 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     if (tracks.microphone) {
-      this.setLocalAudioTrack(tracks.microphone);
+      const oldTrack = this.mediaProperties.audioTrack;
```
arun3528, line 7273: would have been nicer if change track happened on the track level we don't have to do on and off

packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -63,6 +78,68 @@ class MuteState {
     this.pendingPromiseReject = null;
   }
 
+  /**
+   * Starts the mute state machine. Needs to be called after a new MuteState instance is created.
+   *
+   * @param {Object} meeting - the meeting object
+   * @returns {Promise<void>}
+   */
+  public async init(meeting: any) {
```
arun3528, line 87: can we sync on this tomrrow ?

packages/@webex/media-helpers/src/webrtc-core.ts
marcin-bazyl: wcmeCreateCameraTrack is a generic function, it takes a track class as a parameter so that it can instantiate the right class. SDK is calling it with the SDK's LocalCameraTrack class passed as that parameter

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: maybe, but even when you do that change in webrtc-core and change underlying track when changing microphone, at API level you still need to allow people to just call publish with a completely new LocalMicrophoneTrack, so then you still have to do on and off.

packages/@webex/media-helpers/src/webrtc-core.ts
marcin-bazyl: I've decided to change this back to return void (not a promise) and just throw when unmuting is not allowed. This simplifies the code in many other places.

packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -63,6 +74,69 @@ class MuteState {
...
+  /**
+   * Mutes/unmutes local track
+   *
+   * @param {Object} meeting - the meeting object
+   * @param {Boolean} mute - true to mute the track, false to unmute it
+   * @param {ServerMuteReason} reason - reason for muting/unmuting
+   * @returns {void}
+   */
+  private muteLocalTrack(meeting: any, mute: boolean, reason: ServerMuteReason) {
+    this.ignoreMuteStateChange = true;
```
shnaaz, line 131: Why is this done?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -922,7 +922,25 @@ function sendDtmfTones() {
...
+  '360p': '360p',
+  '480p': '480p',
+  '720p': '720p',
+  '1080p': '1080p'
+};
+const localMedia = {
+  microphoneTrack: undefined,
+  cameraTrack: undefined,
+  videoConstraints: {
+    [localVideoQuality["360p"]]: { width: 640, height: 360 },
```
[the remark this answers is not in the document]
arun3528: not sure if marcin did it to fit the sample app

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7225,23 +7265,45 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      const localTrack = tracks.microphone;
+
+      oldTrack?.off(LocalTrackEvents.Muted, this.localAudioTrackMuteStateHandler);
```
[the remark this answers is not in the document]
arun3528: track ended for all the track are the same as its on the different class so name could be the same

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7225,23 +7265,45 @@ export default class Meeting extends StatelessWebexPlugin {
...
         this.mediaProperties.audioTrack
       );
     }
 
     if (tracks.camera) {
-      this.setLocalVideoTrack(tracks.camera);
+      const oldTrack = this.mediaProperties.videoTrack;
+      const localTrack = tracks.camera;
+
+      oldTrack?.off(LocalTrackEvents.Muted, this.localVideoTrackMuteStateHandler);
```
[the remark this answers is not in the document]
arun3528: same the track is a custom track so they can have common name

packages/@webex/plugin-meetings/src/meeting/muteState.ts
arun3528: when doing local mute if there is a server mute we need to check ignoreMuteStateChange

shnaaz (commented):
https://github.com/webex/webex-js-sdk/pull/2739 has unittests added hence supersedes this PR

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
marcin-bazyl: no, this is correct, for screen share we only care about Ended event and for mic and camera we only care about Muted events.
Reasoning:
- for screen share - when user stops screen share by pressing the "stop" button in browser's UI, the Ended event is fired, there is no concept of muting screen share audio or video, even if the app called SetMuted() on screen share tracks, there is no equivalent Locus API that we would need to call when that happens, so we just don't listen on Muted events
- for mic and camera - we don't care about ended event, we don't listen for it, we only care about Mute event - it's fired when the app mutes/unmutes mic/camera

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
marcin-bazyl: same answer as above

packages/@webex/plugin-meetings/src/meeting/muteState.ts
marcin-bazyl: yeah, whenever SDK needs to mute the local webrtc-core track (because of remote mute), SDK calls setMuted() on it and that triggers the Muted event, which we need to ignore, because we only want to react to it if it's triggered by the app calling setMuted()

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1196,6 +1243,15 @@ function toggleSendVideo() {
     return;
   }
 
+  if (isMultistream && localMedia.cameraTrack) {
+    const newMuteValue = !localMedia.cameraTrack.muted;
+
+    localMedia.cameraTrack.setMuted(newMuteValue);
+
+    console.log(`MeetingControls#toggleSendAudio() :: Successfully ${newMuteValue ? 'muted': 'unmuted'} audio!`);
```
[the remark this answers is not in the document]
marcin-bazyl: done

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -922,7 +922,25 @@ function sendDtmfTones() {
   }
 }
 
-function getMediaStreams(mediaSettings = getMediaSettings(), audioVideoInputDevices = {}) {
+const localVideoQuality = {
```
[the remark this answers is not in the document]
marcin-bazyl: this stuff is actually already exported as AVAILABLE_RESOLUTIONS, but the sample app can't import things

marcin-bazyl:
I've changed the destination branch from beta to beta_bnr - this is now our feature branch for BNR, because there are multiple people working on this right now. Hoping to merge this into beta in the next few days (as soon as unit and integration tests are fixed in SDK and corresponding Cantina beta_bnr branch is ready for merging to Cantina master too)

torgeadelin (approved):

arun3528 (approved):

PR #2722, opened 2023-04-06
feat(plugin-meetings): duplicate-SCRs - remove duplicates temp fix

torgeadelin:
# COMPLETES [SPARK-417740](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-417740)

We decided to temporarily fix this in the SDK.  Please not that we've discussed this with Brian Baldino and he agreed to have his team fix this in the JMP layer (@webex/json-multistream), but that might take a while and has some other implications.
This is a temp fix to avoid sending duplicate SCRs AND to avoid delays. 

- add logic to compare previous requests to new ones

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

- unit tests, manual test

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
+++ b/packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
@@ -552,28 +552,28 @@ describe('MediaRequestManager', () => {
     ]);
   });
 
-  it('clears all the requests on reset()', () => {
+  it('avoids sending duplicate requests and clears all the requests on reset()', () => {
     // send some requests and commit them one by one
-    addReceiverSelectedRequest(1500, fakeReceiveSlots[0], RECEIVER_SELECTED_MAX_FS, true);
-    addReceiverSelectedRequest(1501, fakeReceiveSlots[1], RECEIVER_SELECTED_MAX_FS, true);
+    addReceiverSelectedRequest(1500, fakeReceiveSlots[0], RECEIVER_SELECTED_MAX_FS, false);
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -552,28 +552,28 @@ describe('MediaRequestManager', () => {
     ]);
   });
 
-  it('clears all the requests on reset()', () => {
+  it('avoids sending duplicate requests and clears all the requests on reset()', () => {
```
torgeadelin, line 555: I'm not entirely sure this is how it's supposed to work.
The `addRequest` method has a flag (`commit`) which forces the method to also send the request immediately after adding it. So this test was initially testing that, after manually triggering commit after each addRequest (see how the last param used to be `true` and now is `false`), and calling commit again, it will re-send all the requests again (in a batch). 

Is this behaviour necessary still? I.e forcing the requests to be send on `addRequest` and then sending them again as a batch? At the moment this is no longer the case, because it detects them as duplicates.

It's also why i've slightly changed the meaning of this test.

arun3528: yes the behavior is needed @marcin-bazyl

torgeadelin: @Coread

Coread (approved):

PR #2744, opened 2023-04-17
fix: added unitttests

shnaaz:
Part of https://github.com/webex/webex-js-sdk/pull/2694
Added unittests for muteState.ts and webrtc-core.js

### DIFF
--- a/packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
+++ b/packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
@@ -1,45 +1,130 @@
-import {assert} from '@webex/test-helper-chai';
+import {assert, expect} from '@webex/test-helper-chai';
 import sinon from 'sinon';
-import {LocalCameraTrack, LocalMicrophoneTrack} from '@webex/media-helpers';
+import {LocalCameraTrack, LocalMicrophoneTrack, LocalMicrophoneTrackEvents, LocalCameraTrackEvents, LocalDisplayTrack, createCameraTrack, createMicrophoneTrack, createDisplayTrack} from '@webex/media-helpers';
+import * as wcmetracks from '@webex/internal-media-core';
 
 describe('media-helpers', () => {
   describe('webrtc-core', () => {
 
     const classesToTest = [
-      {className: LocalCameraTrack, title: 'LocalCameraTrack'},
-      {className: LocalMicrophoneTrack, title: 'LocalMicrophoneTrack'},
+      {className: LocalCameraTrack, title: 'LocalCameraTrack', event: LocalCameraTrackEvents, createFn: createCameraTrack, spyFn: 'createCameraTrack'},
+      {className: LocalMicrophoneTrack, title: 'LocalMicrophoneTrack', event: LocalMicrophoneTrackEvents, createFn: createMicrophoneTrack, spyFn: 'createMicrophoneTrack'},
     ];
 
-    classesToTest.forEach(({className, title}) =>
-      describe(title, () => {
-        const fakeStream = {
-          getTracks: sinon.stub().returns([{
-            label: 'fake track',
-            id: 'fake track id',
-          }])
-        };
-        const track = new className(fakeStream);
+    classesToTest.forEach(({className, title, event, createFn, spyFn}) =>
+    describe(title, () => {
+      const fakeStream = {
+        getTracks: sinon.stub().returns([{
+          label: 'fake track',
+          id: 'fake track id',
+        }])
+      };
+      const track = new className(fakeStream);
 
-        it('by default allows unmuting', async () => {
-          assert.equal(track.isUnmuteAllowed(), true);
-          await track.setMuted(false);
-        })
+      afterEach(() => {
+        sinon.restore();
+      });
 
-        it('rejects setMute(false) if unmute is not allowed', async () => {
-          track.setUnmuteAllowed(false);
-
-          assert.equal(track.isUnmuteAllowed(), false);
-          await assert.isRejected(track.setMuted(false), Error, 'Unmute is not allowed');
-        });
-
-        it('resolves setMute(false) if unmute is allowed', async () => {
-          track.setUnmuteAllowed(true);
-
-          assert.equal(track.isUnmuteAllowed(), true);
-          await track.setMuted(false);
-        })
-
+      it('by default allows unmuting', async () => {
+        assert.equal(track.isUnmuteAllowed(), true);
+        await track.setMuted(false);
       })
-    )
+
+      it('rejects setMute(false) if unmute is not allowed', async () => {
+        track.setUnmuteAllowed(false);
+
+        assert.equal(track.isUnmuteAllowed(), false);
+        const fn = () => track.setMuted(false);
+        expect(fn).to.throw(/Unmute is not allowed/);
+      });
+
+      it('resolves setMute(false) if unmute is allowed', async () => {
+        track.setUnmuteAllowed(true);
+
+        assert.equal(track.isUnmuteAllowed(), true);
+        track.setMuted(false);
+      });
+
+      it('tests setServerMuted- true to false', async () => {
+        track.underlyingTrack.enabled = true; //start with unmuted
+
+        track.setMuted(true); //this.muted is true, call setServerMuted with false
+        assert.equal(track.muted, true);
+
+        const spy = sinon.stub(track, 'emit');
+        track.setServerMuted(false, 'remotelyMuted');
+
+        assert.equal(track.muted, false);
+
+        const spyCall = spy.getCall(1);
+        assert.equal(spyCall.calledWithExactly(event.ServerMuted, {muted: false, reason: 'remotelyMuted' }), true);
+      });
+
+      it('tests setServerMuted- false to true', async () => {
+        track.underlyingTrack.enabled = false; //start with muted
+
+        track.setMuted(false); //this.muted is false, call setServerMuted with true
+        assert.equal(track.muted, false);
+
+        const spy = sinon.stub(track, 'emit');
+        track.setServerMuted(true, 'remotelyMuted');
+
+        assert.equal(track.muted, true);
+
+        const spyCall = spy.getCall(1);
+        assert.equal(spyCall.calledWithExactly(event.ServerMuted, {muted: true, reason: 'remotelyMuted' }), true);
+      });
+
+      it('tests setServerMuted - true to true', async () => {
+        track.underlyingTrack.enabled = true; //start with unmuted
+
+        track.setMuted(true); //this.muted is true, call setServerMuted with true
+        assert.equal(track.muted, true);
+
+        const spy = sinon.stub(track, 'emit');
+        track.setServerMuted(true, 'remotelyMuted');
+
+        assert.equal(track.muted, true);
+
+        const spyCall = spy.getCall(1);
+        assert.equal(spyCall, null);
+      });
+
+      it('tests setServerMuted- false to false', async () => {
+        track.underlyingTrack.enabled = false; //muted
+
+        track.setMuted(false); //this.muted is false, call setServerMuted with false
+        assert.equal(track.muted, false);
+
+        const spy = sinon.stub(track, 'emit');
+        track.setServerMuted(false, 'remotelyMuted');
+
+        assert.equal(track.muted, false);
+
+        const spyCall = spy.getCall(1);
+        assert.equal(spyCall, null);
+      });
+
+      it('checks creating tracks', async () => {
+        const constraints = {devideId: 'abc'};
+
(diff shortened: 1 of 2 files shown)


packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,130 @@
...
-          assert.equal(track.isUnmuteAllowed(), false);
-          await assert.isRejected(track.setMuted(false), Error, 'Unmute is not allowed');
-        });
+      it('rejects setMute(false) if unmute is not allowed', async () => {
+        track.setUnmuteAllowed(false);
 
-        it('resolves setMute(false) if unmute is allowed', async () => {
-          track.setUnmuteAllowed(true);
+        assert.equal(track.isUnmuteAllowed(), false);
+        const fn = () => track.setMuted(false);
```
marcin-bazyl, line 37: why this change?

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,130 @@
...
-          await track.setMuted(false);
-        })
+      it('resolves setMute(false) if unmute is allowed', async () => {
+        track.setUnmuteAllowed(true);
+
+        assert.equal(track.isUnmuteAllowed(), true);
+        track.setMuted(false);
+      });
+
+      it('tests setServerMuted- true to false', async () => {
```
marcin-bazyl, line 48: all these setServerMuted tests could be just wrapped with a describe() and then no need to repeat "tests setServerMuted" in each test title

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,130 @@
...
+        track.underlyingTrack.enabled = true; //start with unmuted
```
marcin-bazyl, line 49: why do we need this? isn't the call to setMuted() enough to set the right state?

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,130 @@
...
+
+        track.setMuted(true); //this.muted is true, call setServerMuted with false
+        assert.equal(track.muted, true);
+
+        const spy = sinon.stub(track, 'emit');
```
marcin-bazyl, line 54: I think it would be better to just register an event handler for the event we are expecting and check that it gets called rather than stubbing the emit() method

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,130 @@
...
+        track.setServerMuted(true, 'remotelyMuted');
+
+        assert.equal(track.muted, true);
+
+        const spyCall = spy.getCall(1);
+        assert.equal(spyCall, null);
+      });
+
+      it('tests setServerMuted- false to false', async () => {
```
marcin-bazyl, line 93: this test and 3 previous ones are almost exactly the same, they could be refactored to re-use code from same helper test function

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+
+  tests.forEach(({mediaType, title}) =>
+    describe(title, () => {
+
+      afterEach(() => {
+        MeetingUtil.remoteUpdateAudioVideo = originalRemoteUpdateAudioVideo;
+      });
+
+
+      it('tests handleLocalTrackChange', async () => {
```
marcin-bazyl, line 583: title should say "handleLocalTrackChange calls init()" or even better it should be:
```
describe("#handleLocalTrackChange", () => {
  it('calls init()', () => ....);
})
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+    const direction = mediaType === AUDIO ? {sendAudio: true} : {sendVideo: true};
+    muteState = createMuteState(mediaType, meeting, direction, false);
+
+    originalRemoteUpdateAudioVideo = MeetingUtil.remoteUpdateAudioVideo;
+
+    MeetingUtil.remoteUpdateAudioVideo = sinon.stub().resolves(fakeLocus);
+  }
+
+   const setupSpies = (muteState, mediaType) => {
+    applyUnmuteAllowedToTrackSpy = sinon.spy(muteState, 'applyUnmuteAllowedToTrack');
```
marcin-bazyl, line 562: applyUnmuteAllowedToTrack, muteLocalTrack and applyClientStateToServer are all private functions. I don't think we should be spying on them - this makes the tests very closely coupled with the implementation, instead we should be mocking/spying the external methods/functions that MuteState calls, so for example instead of muteLocalTrack we should spy on the setServerMuted() method of the right track, instead of applyUnmuteAllowedToTrack we should spy on setUnmuteAllowed() on the right track, instead of applyClientStateToServer we should be checking that the right requests are sent to the server

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+
+
+      it('tests handleLocalTrackChange', async () => {
+        setup(mediaType);
+        const spy = sinon.spy(muteState, 'init');
+        muteState.handleLocalTrackChange(meeting);
+        assert.calledOnceWithExactly(spy, meeting);
+      });
+
+      it('tests init when track is undefined', async () => {
```
marcin-bazyl, line 590: all tests that start with "tests init" should be wrapped with describe('#init')

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+        muteState.init(meeting);
+
+        assert.calledOnceWithExactly(applyUnmuteAllowedToTrackSpy, meeting);
+        assert.notCalled(muteLocalTrackSpy);
+        assert.notCalled(setServerMutedSpy);
+        assert.calledOnceWithExactly(applyClientStateToServerSpy, meeting);
+        assert.isFalse(muteState.state.client.localMute);
+      });
+
+      it('#muteLocalTrack', async () => {
```
marcin-bazyl, line 628: muteLocalTrack is a private function, I don't think we should be testing it, but the test is actually calling init() instead, so I'm confused, what is being tested here? same for the next test

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+          muteState.state.client.localMute = false;
+
+          const spy = sinon.spy(muteState, 'applyClientStateToServer');
+          muteState.handleLocalTrackMuteStateChange(meeting, true);
+          assert.equal(muteState.state.client.localMute, true);
+          assert.calledOnceWithExactly(spy, meeting)
+        });
+      });
+
+      describe('#applyClientStateLocally', () => {
```
marcin-bazyl, line 692: it's not clear to me what is the purpose of the tests inside this describe from the their titles. Are you sure you meant "ignoreMuteStateChange" in the titles and not sdkOwnsLocalTrack? Can you make them more descriptive?

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,221 @@ describe('plugin-meetings', () => {
...
+          sinon.restore();
+        });
+
+        it('checks when ignoreMuteStateChange is false', () => {
+          setup(mediaType);
+          setupSpies(muteState, mediaType);
+          muteState.sdkOwnsLocalTrack= false;
+
+          muteState.applyClientStateLocally(meeting, 'somereason');
+          assert.calledOnceWithExactly(muteLocalTrackSpy, meeting, muteState.state.client.localMute, 'somereason');
```
marcin-bazyl, line 704: we should be checking that setMuted was called on the track instead of a private muteLocalTrack

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
shnaaz: I am trying to specifically test test muteLocalTrack by calling it through muteState.init(meeting) since it is a private method.

shnaaz: If we do not want specific its here then I can add checks for serverMted in tests above

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
shnaaz: Because the test was failing

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
shnaaz: Here's  the reason
image

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
marcin-bazyl: ah, yes, sorry I forgot to update it when I changed setMuted to not return a promise

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
```diff
@@ -1,45 +1,140 @@
-import {assert} from '@webex/test-helper-chai';
+import {assert, expect} from '@webex/test-helper-chai';
 import sinon from 'sinon';
-import {LocalCameraTrack, LocalMicrophoneTrack} from '@webex/media-helpers';
+import {LocalCameraTrack, LocalMicrophoneTrack, LocalMicrophoneTrackEvents, LocalCameraTrackEvents, LocalDisplayTrack, createCameraTrack, createMicrophoneTrack, createDisplayTrack} from '@webex/media-helpers';
+import * as wcmetracks from '@webex/internal-media-core';
 
-describe('media-helpers', () => {
+describe.only('media-helpers', () => {
```
marcin-bazyl, line 6: another .only

packages/@webex/media-helpers/test/unit/spec/webrtc-core.js
shnaaz: setMuted calls super.setMuted which as you see in screenshot does something conditionally, hence setMuted isn't doing anything as this condition is not satisfied.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,202 @@ describe('plugin-meetings', () => {
...
+          sinon.restore();
+        });
+
+        it('checks when sdkOwnsLocalTrack is false', () => {
+          setup(mediaType);
+          setupSpies(muteState, mediaType);
+          muteState.sdkOwnsLocalTrack= false;
+
+          muteState.applyClientStateLocally(meeting, 'somereason');
+          assert.calledOnceWithExactly(setServerMutedSpy, muteState.state.client.localMute, 'somereason');
```
marcin-bazyl, line 686: it would be good to check that setMuted was not called here

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -509,3 +509,202 @@ describe('plugin-meetings', () => {
...
+  
+        it('tests when remoteMuted is true', async () => {
+          // testing that muteLocalTrack is called
+          setup(mediaType, true);
+          setupSpies(muteState, mediaType);
+  
+          muteState.init(meeting);
+  
+          const serverMutedSpyCall1 = setServerMutedSpy.getCall(0);
+          const serverMutedSpyCall2 = setServerMutedSpy.getCall(1);
```
marcin-bazyl, line 633: why are we expecting 2 calls to setServerMuted? the init() should trigger just 1 call to it.
I've checked this and it's because the first call is triggered from init() called from createMuteState() that's called by setup(), so I think we should call setServerMutedSpy.resetHistory() in setup() after calling createMuteState() (or inside setupSpies()) so that we have a clean slate for the test. We should probably do it for all the spies.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
shnaaz: done

marcin-bazyl (approved):

PR #2755, opened 2023-04-18
fix: testUtils

shnaaz:
Updated converged-space-meetings test

### DIFF
--- a/packages/@webex/plugin-meetings/test/utils/testUtils.js
+++ b/packages/@webex/plugin-meetings/test/utils/testUtils.js
@@ -197,40 +197,42 @@ const delayedTest = (callback, timeout) =>
   });
 
 const addMedia = async (user, options = {}) => {
-  const mediaReadyPromises = Array.isArray(options.expectedMediaReadyTypes)
-    ? options.expectedMediaReadyTypes.reduce((output, expectedMediaReadyType) => {
-      if (typeof expectedMediaReadyType !== 'string') {
-        return output;
-      }
-
-      output[expectedMediaReadyType] = new Defer();
-
-      return output;
-    }, {})
-    : {local: new Defer(), remoteAudio: new Defer(), remoteVideo: new Defer()};
-
-  const mediaReady = (media) => {
-    if (!media) {
-      return;
-    }
-    if (mediaReadyPromises[media.type]) {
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/test/utils/testUtils.js
```diff
@@ -196,41 +197,42 @@ const delayedTest = (callback, timeout) =>
...
+    sendShare: false,
+  });
 
   if (options.multistream) {
     await user.meeting.addMedia({});
+    await user.meeting.publishTracks({microphone: new LocalMicrophoneTrack(new MediaStream([localStream.getAudioTracks()?.[0]])), camera: new LocalCameraTrack(new MediaStream([localStream.getVideoTracks()?.[0]]))});
 
-    await user.meeting.publishTracks({microphone: localStream.getAudioTracks()[0], camera: localStream.getVideoTracks()[0]})
   } else {
+    const mediaReadyPromises = Array.isArray(options.expectedMediaReadyTypes)
```
marcin-bazyl, line 213: I think these changes of moving mediaReadyPromises into the non-multistream case are not really needed - it should be enough that the test calls addMedia without any `expectedMediaReadyTypes` and addMedia should simply not wait for any media:ready events in that case, right?

marcin-bazyl: actually I think existing code assumes that by default it waits for all media:ready events, so in our case we just need to call it from the test with `expectedMediaReadyTypes: []` to explicitly say that we don't want to wait for any media:ready events

martroos: Is there a good reason not to move the code? Unless we add a check to make sure that event never gets called then I can't see it being beneficial to add the event handler to it.

If we don't want to move the code, then we should probably force `expectedMediaReadyTypes` (or `mediaReadyPromises`) to be `[]` when `multistream: true` is set as an option - otherwise everywhere we use the multstream option will also need the expected ready types to change to `[]`. For completeness as a test util we should probably also error if any non-empty array is passed in as `expectedMediaReadyTypes` when `multistream` is `true`.

marcin-bazyl (approved):

PR #2756, opened 2023-04-18
fix: syncmeeting made to wait for handleLocusEvent

mkesavan13:
# COMPLETES [SPARK-412189](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-412189)

The ExtCare customer is trying to join the meeting without having a gap for the media to be attached to the locus and that caused both of the users to be stuck in the waiting room.

This PR fixes the above issue by waiting for the locus event to be complete and then proceed with syncMeetings. 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
@@ -267,6 +267,7 @@ SelfUtils.getStatus = (status) => ({
 SelfUtils.wasMediaInactiveOrReleased = (oldSelf: any = {}, changedSelf: any) =>
   oldSelf.joinedWith &&
   oldSelf.joinedWith.state === _JOINED_ &&
+  changedSelf.joinedWith &&
   changedSelf.joinedWith.state === _LEFT_ &&
   (changedSelf.joinedWith.reason === MEETING_END_REASON.INACTIVE ||
     changedSelf.joinedWith.reason === MEETING_END_REASON.MEDIA_RELEASED);


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1082,9 +1088,9 @@ export default class Meetings extends WebexPlugin {
       const activeLocusUrl = [];
 
       if (locusArray?.loci && locusArray.loci.length > 0) {
-        locusArray.loci.forEach((locus) => {
```
arun3528, line 1085: @mkesavan13  do you think we need just this change as there are no aync calls , to do you need all the changes mentioned in the PR ?

mkesavan13: @arun3528 - You are right. Removed the unwanted changes. Sent build to customer and tested as well. Kindly review again.

arun3528 (approved):

arun3528:
Looks good , we might have to update the beta branch also

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -335,8 +336,13 @@ export default class Meetings extends WebexPlugin {
...
         });
     } else {
-      meeting.locusInfo.parse(meeting, data);
+      return new Promise((resolve) => {
+        meeting.locusInfo.parse(meeting, data);
+        resolve(undefined);
+      });
     }
+
+    return Promise.resolve(undefined);
```
marcin-bazyl, line 345: as you've added the `async` keyword to the function, then it automatically always returns a promise resolved with whatever return value was returned, so you don't need to convert all the return statements in the function to return a promise - the function `handleLocusEvent()` will always return a promise anyway

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1082,9 +1088,9 @@ export default class Meetings extends WebexPlugin {
       const activeLocusUrl = [];
 
       if (locusArray?.loci && locusArray.loci.length > 0) {
-        locusArray.loci.forEach((locus) => {
+        locusArray.loci.forEach(async (locus) => {
```
marcin-bazyl, line 1091: `forEach` does not wait for each of the functions invoked on each array element to resolve
see https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

packages/@webex/plugin-meetings/src/meetings/index.ts
mkesavan13: Thanks @marcin-bazyl 

The eslint actually threw [consistent-return](https://eslint.org/docs/latest/rules/consistent-return) error. That's why I made this change. Turns out, it accepts `return undefined`

Made this change.

packages/@webex/plugin-meetings/src/meetings/index.ts
mkesavan13: Makes sense @marcin-bazyl 

Refactored it to the `for..of` loop. Thank you.

marcin-bazyl (approved):
info for others reading this PR: the changes to wait for all handleLocusEvent() calls to resolve in syncMeetings were not actually needed, so in the end the fix is just in selfUtils.ts

arun3528:
> 

@marcin-bazyl  yes the issue will get fixed if we just add a check on the selfUtil but joinedWith should not be null But the root cause was that the promise was not resolved properly which was causing the response parsed after join

PR #2806, opened 2023-05-11
Feat/breakout sort main breakout locus

othellashao:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-424011

Cover case that same account already join into a breakout session with another device(native client), then open the web client with same account, will list two meeting items at the top of conversation list. Should show one meeting item.

Sort the locus list from syncMeeting api:
1. if no meeting object for current meeting, create it with main locus first, after meeting object created, then handle the breakout locus of the meeting (try to drop the main locus, use the active breakout locus, will miss some data)
2. if already have meeting object, just update them.

Change to use shallow merge data into main session cache:
1. Fix issue that when host end the breakouts, still keep started status. Caused by in such case received DTO won't have `startTime`, `status` in `controls.breakout`, will use the old value if use deep merge
2. check locus parse logic in our code, only the `participants` will do special delta merge

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -262,6 +262,7 @@ const Breakouts = WebexPlugin.extend({
     this.set(params);
     this.set('groups', params.groups);
     this.set('startTime', params.startTime);
+    this.set('status', params.status);
```
Coread, line 265: question: Why is this (and the groups, startTime) needed? Line 262 sets all params to the breakouts, manually setting part of params to something with the same name shouldn't be necessary

Coread: I've worked this out myself. This is so that status, groups, startTime will be overwritten when they aren't provided in the params object. This isn't very obvious, maybe work a comment here.

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1280,6 +1295,72 @@ export default class Meetings extends WebexPlugin {
...
   }
 
+  /**
+   * sort out locus array for initial creating
+   * @param {Array} loci original locus array
+   * @returns {undefined}
+   * @public
+   * @memberof Meetings
+   */
+  sortLocusArrayToUpdate(loci: any) {
```
Coread, line 1305: issue: presumably the type here is at least of an array i.e. not any but at least `any[]`

Coread (approved):

PR #2807, opened 2023-05-11
feat: allow for extra params to be provided when fetching meeting info

martroos:

In order to fix https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-421937 there is a need to be able to pass in extra parameters when fetching meeting info, specifically when creating a new meeting. 

This PR adds the ability to provide arbitrary additional params when creating and when fetching meeting info in general, filtered to ensure both key and value are truthy (this can be changed later of course).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
+++ b/packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
@@ -1,3 +1,4 @@
+import {pickBy} from 'lodash';
 import url from 'url';
 
 import {
@@ -218,6 +219,15 @@ MeetingInfoUtil.getDestinationType = async (from) => {
 };
 
 /**
+ * Predicate function to be used when filtering request params.
+ * Currently this simply checks for truthy values but in future could restrict keys and values.
+ * @param {Any} value the request param value
+ * @param {String} key the request param key
+ * @returns {Boolean} if the key and value are valid as a request param
+ */
+MeetingInfoUtil.isValidRequestParam = (value: any, key: string): boolean => !!(key && value);
+
+/**
  * Helper function to build up a correct locus url depending on the value passed
  * @param {Object} options type and value to fetch meeting info
  * @param {String} options.type One of [SIP_URI, PERSONAL_ROOM, MEETING_ID, CONVERSATION_URL, LOCUS_ID, MEETING_LINK]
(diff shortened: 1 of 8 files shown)


packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
```diff
@@ -217,6 +218,15 @@ MeetingInfoUtil.getDestinationType = async (from) => {
...
 };
 
+/**
+ * Predicate function to be used when filtering request params.
+ * Currently this simply checks for truthy values but in future could restrict keys and values.
+ * @param {Any} value the request param value
+ * @param {String} key the request param key
+ * @returns {Boolean} if the key and value are valid as a request param
+ */
+MeetingInfoUtil.isValidRequestParam = (value: any, key: string): boolean => !!(key && value);
```
marcin-bazyl, line 228: question: I'm not sure what's the benefit of using this predicate in its current form. What if someone wants to add an extra param that has a false-y value? Can't we just spread all the extraParams without the call to pickBy?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1130,12 +1141,12 @@ export default class Meetings extends WebexPlugin {
...
       if (enableUnifiedMeetings && !isMeetingActive && useRandomDelayForInfo && waitingTime > 0) {
         meeting.fetchMeetingInfoTimeoutId = setTimeout(
-          () => meeting.fetchMeetingInfo({}),
+          () => meeting.fetchMeetingInfo({extraParams: infoExtraParams}),
           waitingTime
         );
         meeting.parseMeetingInfo(undefined, destination);
       } else {
-        await meeting.fetchMeetingInfo({});
+        await meeting.fetchMeetingInfo({extraParams: infoExtraParams});
```
marcin-bazyl, line 1149: question: do we need to provide these extra params in other places when meeting info is fetched, for example in Meeting.verifyPassword?

martroos: The use case I'm currently trying to solve was for meeting create. I can add this to `verifyPassword` also, but I couldn't think of a use case right now for it (the create case needs the extra params to provide `mtid` / `joinTXId` which are instead of a password / captcha).

If you'd like this added let me know and I'll make the change.

packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
marcin-bazyl: > My reasoning for adding the predicate was actually to eventually remove the explicit installedOrgID and locusId.... 

Sounds like a good idea, but realistically I don't think this will ever get done unless you do it now in this PR, so I would either do it fully now or remove the predicate.

My 2nd concern is that sometimes we actually need false-y values to be sent to server (like empty string), so that's why I feel we shouldn't be filtering out them, we can filter out just false-y keys

packages/@webex/plugin-meetings/src/meetings/index.ts
marcin-bazyl: ok, looks like it is called only from `verifyPassword()` and from `PersonalMeetingRoom.get()` which I guess both don't actually need the mtid so it's OK to keep it as it is

packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
martroos: As discussed, I have removed the predicate.

https://github.com/webex/webex-js-sdk/pull/2807/commits/0bf745d36489c2b395d6fa1aeb97166be27cdd9d

marcin-bazyl (approved):

PR #2812, opened 2023-05-15
feat(CA): fixed CA & BO issues for the base functions

csshz:
<ASSOCIATED ISSUE NUMBER/TRACKING LINK (REQUIRED)>
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-422176
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-400364
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-429048

SPARK-422176: append CA event for the case of Many times users abandon meetings by just closing the browser tab.
The SDK should support the post event method for third parties to meet the requirements of Call Analyzer.

SPARK-400364: get pre-assignments info for breakout session.
one issue overwrites the original 'groups', so change to new nodes as 'preAssignments'.

SPARK-429048: When we join a meeting from the page in "Join via browser", the correlationid will be released in the sdk, and the crosslaunch case should keep the correlationid and send it as CA data.

background: correlationID is used as the unique index for background business logic analysis within Meera side, and all related events are analysed in a unified method to determine the life cycle of a user, whether it is JMS, and what causes JMF if it occurs.

< DESCRIBE YOUR CHANGES >
those issues  for the base functions to thrid-part.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1683,10 +1683,6 @@ export default class Meeting extends StatelessWebexPlugin {
         identifiers.trackingId = trackingId;
       }
 
-      if (options.correlationId) {
-        this.setCorrelationId(options.correlationId);
-      }
-
       let payload = {};
 
       const joinRespRxStartAudio = this.getSetupDelayDuration('audio');
@@ -4317,6 +4313,10 @@ export default class Meeting extends StatelessWebexPlugin {
       );
       this.setCorrelationId(uuid.v4());
       LoggerProxy.logger.log(`Meeting:index#join --> New correlation id ${this.correlationId}`);
+    }
+
+    if (options.correlationId) {
+      this.setCorrelationId(options.correlationId);
     }
 
     if (options.rejoin) {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4306,7 +4306,7 @@ export default class Meeting extends StatelessWebexPlugin {
     if (options.correlationId) {
       this.setCorrelationId(options.correlationId);
       LoggerProxy.logger.log(
-        `Meeting:index#join --> Using a new correlation from app ${this.correlationId}`
+        `Meeting:index#join --> Using a new correlation id from app ${this.correlationId}`
       );
     }
 
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1683,6 +1683,10 @@ export default class Meeting extends StatelessWebexPlugin {
         identifiers.trackingId = trackingId;
       }
 
+      if (options.correlationId) {
+        this.setCorrelationId(options.correlationId);
+      }
+
       let payload = {};
 
       const joinRespRxStartAudio = this.getSetupDelayDuration('audio');
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -190,6 +190,7 @@ export const MEDIA_UPDATE_TYPE = {
  * @property {String} [meetingQuality.remote]
  * @property {Boolean} [rejoin]
  * @property {Boolean} [enableMultistream]
+ * @property {String} [correlationId]
  */
 
 /**
@@ -1400,12 +1401,12 @@ export default class Meeting extends StatelessWebexPlugin {
   }
 
   /**
-   * Post metrics for events, provide the ability to send Call Analyzer for the request of third-party.
+   * Posts metrics event for this meeting. Allows the app to send Call Analyzer events.
+   * @param {String} eventName - event name is used to identify the type sent by the call analyzer
    * @public
    * @memberof Meeting
    * @returns {Promise}
    */
-  // eslint-disable-next-line valid-jsdoc
   public postMetrics(eventName) {
     Metrics.postEvent({
       event: eventName,
@@ -4302,6 +4303,10 @@ export default class Meeting extends StatelessWebexPlugin {
       joinSuccess = resolve;
     });
 
+    if (options.correlationId) {
+      this.setCorrelationId(options.correlationId);
+    }
+
     if (!this.hasJoinedOnce) {
       this.hasJoinedOnce = true;
     } else {
@@ -4313,10 +4318,6 @@ export default class Meeting extends StatelessWebexPlugin {
       );
       this.setCorrelationId(uuid.v4());
       LoggerProxy.logger.log(`Meeting:index#join --> New correlation id ${this.correlationId}`);
-    }
-
-    if (options.correlationId) {
-      this.setCorrelationId(options.correlationId);
     }
 
     if (options.rejoin) {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1402,7 +1402,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
   /**
    * Posts metrics event for this meeting. Allows the app to send Call Analyzer events.
-   * @param {String} eventName - event name is used to identify the type sent by the call analyzer
+   * @param {String} eventName - Call Analyzer event, see eventType in src/metrics/config.ts for possible values
    * @public
    * @memberof Meeting
    * @returns {Promise}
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -444,6 +444,7 @@ export default class Meeting extends StatelessWebexPlugin {
   annotation: any;
   conversationUrl: string;
   correlationId: string;
+  caCorrelationId: string;
```
marcin-bazyl, line 447: There is no need to have a separate new correlation id field - we should use existing `correlationId` field. Correlation id is used for identifying calls in logs, metrics, etc. so it's important that all the code always uses the same value for correlation id everywhere for a given meeting join. 
If you need the web app to specify the correlationId, then that should be done through an argument to Meeting.join() method - if the app specifies it, SDK should use it, otherwise SDK will generate a new random value (that's currently done by existing SDK code).

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1400,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
   }
 
+  /**
+   * post Event, provide the ability to send Call Analyzer for the request of third-party
+   * @public
+   * @memberof Meeting
+   * @returns {Promise}
+   */
+  // eslint-disable-next-line valid-jsdoc
+  public postEvent(eventName) {
```
marcin-bazyl, line 1410: the name postEvent() is too generic, the method name should be postMetric(). BTW, I don't see the method being called from anywhere, so is it really needed? Also we should specify the type for the eventName argument and the docs are missing for it

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1400,30 @@ export default class Meeting extends StatelessWebexPlugin {
     );
   }
 
+  /**
+   * post Event, provide the ability to send Call Analyzer for the request of third-party
```
marcin-bazyl, line 1404: This description doesn't seem grammatically correct, also it should be a sentence starting with upper case letter, ending with a dot.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1400,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    });
+  }
+
+  /**
+   * setCaCorrelationId, In the cross launch case, the correlationId sent by the page needs to be passed to prepare for the ucf CA report in the analyzer system.
+   * @public
+   * @memberof Meeting
+   */
+  // eslint-disable-next-line valid-jsdoc
+  public setCaCorrelationId(correlationId) {
```
marcin-bazyl, line 1423: this method is not needed, we already have `setCorrelationId()` (see my other comment)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3497,6 +3497,28 @@ describe('plugin-meetings', () => {
...
+        });
+
+        it('should trigger `postEvent`', async () => {
+          await meeting.postEvent(eventType.LEAVE);
+          assert.calledWithMatch(Metrics.postEvent, {event: eventType.LEAVE});
+        });
+      });
+
+      describe('#setCaCorrelationId', () => {
+        it('should have #postEvent', () => {
```
marcin-bazyl, line 3512: test title is incorrect

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: ah, sorry I've realised now that this method is meant to be called by the app, so it's ok that it's not called from within SDK, but that makes it even more important for the docs and types to be fully populated and correct.

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: yes, the original idea was that the correlationId of cross launch would overwrite the original correlationId, which would cause inconsistency in business logic. According to your suggestion, I will directly pass the correlationId through the meeting.join interface, which is a good way.

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: Yes, This is just an exposed port, which is convenient for the app to pass some Call analyzer data. We know that some post Mertris behaviors must be performed on many page UI terminals (login page, interstitial page etc.) to make basic data for the JMF & JMS report analysis.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
csshz: this test case should be removed by your suggestion.

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: got, updated as your mentioned.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4301,6 +4315,10 @@ export default class Meeting extends StatelessWebexPlugin {
       LoggerProxy.logger.log(`Meeting:index#join --> New correlation id ${this.correlationId}`);
     }
 
+    if (options.correlationId) {
```
marcin-bazyl, line 4318: 1. the docs for JoinOptions need to updated to contain correlationId
2. this seems like the wrong place to do it, because we will end up with the log from line 4315 printing the wrong correlation id and also unnecessary call to setCorrelationId in line 4314,

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1399,20 @@ export default class Meeting extends StatelessWebexPlugin {
     );
   }
 
+  /**
+   * Post metrics for events, provide the ability to send Call Analyzer for the request of third-party.
```
marcin-bazyl, line 1403: please change this to: "Posts metrics event for this meeting. Allows the app to send Call Analyzer events."

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1399,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * @public
+   * @memberof Meeting
+   * @returns {Promise}
+   */
+  // eslint-disable-next-line valid-jsdoc
```
marcin-bazyl, line 1408: why do we need this? I think the "@param" string is missing for eventName

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1399,6 +1399,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  public postMetrics(eventName) {
```
marcin-bazyl, line 1409: type is missing for eventName

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: added.

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: 1. done.
2. reset for correct position.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: I meant that the type should be declared in the code and not as a string but the actual possible values of CA events, but I see now that `eventType` is actually just a const object, so this will be problematic. We should change it to be a proper typescript type at some point, but we don't have to do this in this PR.

Maybe in this PR for now at least we could do:
`public postMetrics(eventName: string) {`
and in the jsdoc it should say something like:
`@param {String} eventName - Call Analyzer event, see eventType in src/metrics/config.ts for possible values`

csshz: that's clean code.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4288,9 +4303,16 @@ export default class Meeting extends StatelessWebexPlugin {
       joinSuccess = resolve;
     });
 
+    if (options.correlationId) {
+      this.setCorrelationId(options.correlationId);
+      LoggerProxy.logger.log(
+        `Meeting:index#join --> Using a new correlation from app ${this.correlationId}`
```
marcin-bazyl, line 4309: very minor thing, but to be consistent with other logs it would be better to say "correlation id" instead of just "correlation"

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
csshz: fixed.

PR #2820, opened 2023-05-18
feat(meeting-info): add CA events

torgeadelin:
# COMPLETES [SPARK-351223](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-351223)

- add CA events for meetinginfo request and response
- refactor fetchMeetingInfo method inside meeting-info-v2 to include a version that sends events when fetchMeetingId initiated from a meeting object / or the call is associated with a meeting. This is needed as this is a public API and 3rd party users can fetch meeting info without actually joining the meeting. The requirement from CA is that this events are only send at the beginning of a join phase. (Sending duplicate error cases is fine)
- Also, because of the same requirement of only sending the events once at the beginning of the join phase, inside the join method on the meeting object, we're checking wether the meetingInfo object is defined, in which case we assume there was a success response from the meeting info api.
- added logic to handle service error codes from Webex API, parse them and send them to CA

- add new functionality
- refactor

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- unit tests, manual tests

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4333,6 +4342,21 @@ export default class Meeting extends StatelessWebexPlugin {
       data: {trigger: trigger.USER_INTERACTION, isRoapCallEnabled: true},
     });
 
+    if (!isEmpty(this.meetingInfo)) {
+      Metrics.postEvent({
+        event: eventType.MEETING_INFO_REQUEST,
+        meeting: this,
+      });
+
+      Metrics.postEvent({
```
marcin-bazyl, line 4351: so we're sending the metric for request and response together at the time of join and not when we actually fetch meeting info? (I just want to double check that CA is OK with that)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -996,6 +996,52 @@ describe('plugin-meetings', () => {
...
+
+            assert.calledOnce(MeetingUtil.joinMeeting);
+            assert.calledOnce(meeting.setLocus);
+            assert.equal(result, joinMeetingResult);
+
+            assert.calledThrice(Metrics.postEvent)
+
+            assert.equal(Metrics.postEvent.getCall(0).args[0].event, 'client.call.initiated');
+            assert.equal(Metrics.postEvent.getCall(1).args[0].event, 'client.meetinginfo.request');
+            assert.equal(Metrics.postEvent.getCall(2).args[0].event, 'client.meetinginfo.response');
```
marcin-bazyl, line 1021: should we be checking also the data being sent with the event?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4333,6 +4342,21 @@ export default class Meeting extends StatelessWebexPlugin {
       data: {trigger: trigger.USER_INTERACTION, isRoapCallEnabled: true},
     });
 
+    if (!isEmpty(this.meetingInfo)) {
+      Metrics.postEvent({
+        event: eventType.MEETING_INFO_REQUEST,
```
marcin-bazyl, line 4347: just to double check: meetingLookupUrl is not required in this CA event and only required in the event for the response?

packages/@webex/plugin-meetings/src/metrics/index.ts
```diff
@@ -452,6 +453,29 @@ class Metrics {
...
+   *
+   * @param {object} err
+   * @param {boolean} showToUser
+   * @returns {any}
+   */
+  parseWebexApiError(err: any, showToUser: boolean) {
+    const serviceErrorCode = err?.body?.code;
+    const clientCodeError = WebexAPIServiceErrorCodes[serviceErrorCode];
+
+    return this.generateErrorPayload(clientCodeError, showToUser, error.name.OTHER, err);
```
marcin-bazyl, line 467: clientCodeError may be undefined - I think the code in generateErrorPayload() will handle that, but maybe better to not call it at all in that case or define the argument types in generateErrorPayload() to say that errorCode can be undefined?

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -238,8 +239,14 @@ export default class MeetingInfoV2 {
     } = null,
     installedOrgID = null,
     locusId = null,
-    extraParams = {}
+    extraParams: {meetingId?: string; [key: string]: string} = {}
   ) {
+    let meetingId: string;
+    if (extraParams.meetingId) {
+      meetingId = extraParams.meetingId;
+      delete extraParams.meetingId;
```
marcin-bazyl, line 247: I think the purpose of extraParams was that these are all params passed in the request to wbxappapi, although not sure if this is documented, so I'm not sure if it's a great idea to be putting the meetingId into them and then stripping it so that it doesn't get sent out in the request. Maybe it should be passed as a separate parameter. You can check with Martin who introduced the extraParams recently.

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: yes, that's correct

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: Yes, I've confirmed with Meera before doing this, and we only want to send the meeting info events once at the beginning of the meeting, but we're doing fetch meeting info several times and in several different contexts, hence we can't send them all the time.

This was just one solution Arun and I came up with and CA team said it works for them too.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1641,7 +1646,7 @@ export default class Meeting extends StatelessWebexPlugin {
       | any
   ) {
     if (options) {
-      const {event, trackingId, mediaConnections} = options;
+      const {event, trackingId, mediaConnections, meetingLookupUrl} = options;
```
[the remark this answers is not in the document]
torgeadelin: this will significatly improve once we have the CA events in typescript (which we will soon)

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
martroos: The intention of `extraParams` was to provide extra parameters to the API call, so if the user adds a parameter to this object, they know it will end up being sent to the service. It isn't well documented as you say, but the use of the phrase "params" should be enough to know that they aren't just "options", which is what you effectively change them into. If we make this change, this also opens the door to similar additions in future, which will further change the meaning of params.

Personally, I'd prefer for this to be an extra argument to the function, instead of overloading the `extraParams` argument.

torgeadelin: addressed

marcin-bazyl (approved):

PR #2826, opened 2023-05-19
test: skipping flakey integrationtest

marcin-bazyl:
"alice video unmute" test very often fails, we haven't seen any issues with video unmuting with the web app and this test has been flakey for a very long time, also on master even before beta branch was created.

For now I'm skipping it (and the previous test so that the video remains unmuted - to not affect other later tests). This test will be rewritten in SPARK-399695 anyway so then I'll re-enable it.


InteractiveTimmy (approved):
LGTM!

PR #2832, opened 2023-05-25
fix: refactor code for sending /media requests to Locus

marcin-bazyl:
# COMPLETES [SPARK-429638](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-429638), [SPARK-396696](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-396696)

3 main issues:
1. When tracks are published, there are 2 separate state machines for audio and video and they trigger local mute updates in Locus independently. If these happen quickly one after another Locus may handle them out of order resulting in incorrect mute state.
2. Roap messages can happen at any time and they also end up calling /media API in Locus (same as local mute state updates), so they can happen while we are sending a local mute state update - this can cause problems and incorrect mute state in Locus.
3. If client mutes/unmutes very quickly at the beginning when joining a meeting and we try to send local mute update on /media API to Locus before we've sent an SDP offer, Locus fails, because there is no confluence in the server

Created a new class LocusMediaRequest that makes sure that we only ever have 1 /media request to Locus "in flight" and also avoids redundant calls to /media, for example, if we need to send a roap message and do a local mute update, these 2 requests can be combined into a single /media request or if there are multiple local mute state updates queued, they can all be combined into a single request with the latest required audioMuted and videoMuted values.
LocusMediaRequest also makes sure we don't call /media with just local mute updates before we've sent a Roap offer.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

Manual e2e testing with the web app + unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/muteState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/muteState.ts
@@ -34,8 +34,10 @@ export const createMuteState = (type, meeting, mediaDirection, sdkOwnsLocalTrack
    the last requested state by the client.
 
    More info about Locus muting API: https://sqbu-github.cisco.com/pages/WebExSquared/locus/guides/mute.html#
+
+   This class is exported only for unit tests. It should never be instantiated directly with new MuteState(), instead createMuteState() should be called
 */
-class MuteState {
+export class MuteState {
   pendingPromiseReject: any;
   pendingPromiseResolve: any;
   state: any;
@@ -62,7 +64,7 @@ class MuteState {
         localMute: false,
       },
       server: {
-        localMute: false,
+        localMute: true,
         // because remoteVideoMuted and unmuteVideoAllowed are updated seperately, they might be undefined
         remoteMute: type === AUDIO ? meeting.remoteMuted : meeting.remoteVideoMuted ?? false,
         unmuteAllowed: type === AUDIO ? meeting.unmuteAllowed : meeting.unmuteVideoAllowed ?? true,
@@ -95,7 +97,7 @@ class MuteState {
           : meeting.mediaProperties.videoTrack?.muted;
 
       LoggerProxy.logger.info(
-        `Meeting:muteState#start --> ${this.type}: local track initial mute state: ${initialMute}`
+        `Meeting:muteState#init --> ${this.type}: local track initial mute state: ${initialMute}`
       );
 
       if (initialMute !== undefined) {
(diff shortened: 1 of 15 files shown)


packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -34,8 +34,10 @@ export const createMuteState = (type, meeting, mediaDirection, sdkOwnsLocalTrack
    the last requested state by the client.
 
    More info about Locus muting API: https://sqbu-github.cisco.com/pages/WebExSquared/locus/guides/mute.html#
+
+   This class is exported only for unit tests. It should never be instantiated directly with new MuteState(), instead createMuteState() should be called
```
torgeadelin, line 38: you can add @internal tag

packages/@webex/plugin-meetings/test/unit/spec/meeting/locusMediaRequest.ts
```diff
@@ -0,0 +1,319 @@
...
+      webexRequestStub.resetHistory();
+
+      // at this point the request should be sent out and "in progress",
+      // so any further calls should be queued
+      results.push(sendRoapMessage('OK'));
+
+      // OK should not be sent out yet, only queued
+      assert.notCalled(webexRequestStub);
+
+      // not simulate the first locus request (offer) to resolve,
```
torgeadelin, line 250: now

marcin-bazyl: fixed

packages/@webex/plugin-meetings/src/meeting/muteState.ts
marcin-bazyl: I don't think it will help. "internal" tag hides the method from type declaration, so it can't be easily called from other packages, but it won't stop someone from calling new MuteState() from within plugin-meetings

marcin-bazyl: actually I'm wrong, "internal" can help even within plugin-meetings, but here we want the class just don't want to instantiate it with "new" so I guess we would have to set "internal" for the constructor? I tried it but then the class doesn't have the constructor at all in the d.ts file, is that ok? feels a bit weird...

torgeadelin: i think that's fine? that's the whole purpose to hide the instantiating i think

torgeadelin (approved):

packages/@webex/plugin-meetings/src/meeting/muteState.ts
torgeadelin: anyway, not a big deal honestly

PR #2845, opened 2023-06-01
Feat/ca metrics/refactor

torgeadelin:

The purpose of this PR is to showcase how the metrics pipeline in the SDK will be refactored.
The aim is to centralise submitting metrics in a single place and have different classes to handle the different required events that we need to send(CA, MQE, behavioural, operational etc).

Some definitions:
CA = Call Analyzer 
CE =  ClientEvent
MQE = Media Quality Event
CD = Call Diagnostic (which includes CA, MQE, and FeatureEvents)

The idea is to have a singleton class called `Metrics` which can afterwords initialise the sub classes that are part of it.
Some categories of metrics (CD) requires the webex object and meetingsCollection. At the moment, they are only stored inside the CallDiagnosticMetrics class, but I think it could potentially live in the Metrics singleton, and the rest could just access it from there. 

### Before:
**plugin-meetings:**
- Metrics class singleton only handled CD and behavioural (inside plugin meetings)
- singleton was handling CD events initialization
- singleton was handling CD gathering payload
- CD events would call public method meeting object, which would also do some payload computation
- eventually call plugin-metrics

**internal-plugin-metrics:**
- sending the request for CD and behavioural metric via batchers
- initialize behavioural metric
- create batcher classes

### After:
**plugin-meetings:**
- plugin meetings will not be loaded with metrics related logic
- only submit metrics inside the plugin-meetings

**internal-plugin-metrics:**
- handle all types of metrics (including operational, CD, behavioural etc)
- have separate classes for each type of metric group, and handle all logic in there
- create bathers for each metrics group/category

## Latencies
To calculate joinTimes, I've created a class to store and calculate certain latencies. This will store any event (not just CD), it will save event name and timestamp. 

Latency = T0 - T1, where T0 is point 0 (start), and T1 is point 1 (end). these are just timestamps at equivalent T0, T1 locations. There are defined latencies required for Call Analyzer.
JoinTimes = Multiple latencies added together, or subtracted by following formula from CA.

For example: meeting info request latency:
T0 = when we send the meeting info request
T1 = when we receive the meeting info response
In this case, both latencies are CD events, but some of them might overlap with other existing behavioural events that we have, hence I'm saving all the timestamps, for each event.

There is also a CD events batcher where we intercept requests, and append more metadata.
This is where we will also add the joinTimes.

Here, we're also doing JSON schema validation, but JSON validation will only be done in dev mode or testing environment.

THIS WORK ONLY FOCUSES ON CD EVENTS (SPECIFICALLY CLIENT EVENTS ONLY). I have added code for MQE too even tho is not in spec, but because the new metrics pipeline handles things in groups (i.e. CD), i decided to include it as well).

Small diagram showing how the flow will work.
https://www.tldraw.com/s/v2_c_-NgNy6pCqhBaFwN8wOBuA?viewport=318%2C-225%2C2206%2C1236&page=page%3At4yKpEa53fx1UreQZCcwq

image

# Potential Merge strategy.
Since this is a big pile of code, and because no unit tests are yet included and is already 35 files, I will be following the following strategy to getting this in.

1. Will not touch any previous metric code, and will add the new code exclusively in new files (suffixed by `-new`).
This will reduce the number of files changed and will keep the existing functionality. (We can't risk sending bad events to CA, as they will complain).

2. I will initially add all the boilerplate, and the functionality without using it, so any calls to the new metrics will be done in a separate PR.

3. Cleanup, once we have the new metrics and submitting the new metrics in, I will remove the old code.

4. All PRs will have equivalent unit tests for each.

PS. The big typescript types and the JSON schemas are automatically generated, and will be published on an private artifactory where we can consume it directly. This automation will be on the https://sqbu-github.cisco.com/WebExSquared/event-dictionary repo, and will be part of the CI. 
For the POC, I just copy pasted the types in here, but these will be consumed via a package.

- will update later

### Change Type

- [ ] Documentation update
- [x] Tooling change
- [x] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/ca-metrics-latencies.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /**
+   * Total JMT
+   * @returns - latency
+   */
+  public getTotalJMT() {
+    const clickToInterstitial = this.getClickToInterstitial();
+    const interstitialToJoinOk = this.getInterstitialToJoinOK();
+
+    if (clickToInterstitial && interstitialToJoinOk) {
```
maxinteger, line 244: is it OK to return with `undefined` when one or both of the values are `0`

packages/@webex/internal-plugin-metrics/src/ca-metrics-latencies.ts
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   * Total Media JMT
+   * @returns - latency
+   */
+  public getTotalMediaJMT() {
+    const clickToInterstitial = this.getClickToInterstitial();
+    const interstitialToJoinOk = this.getInterstitialToJoinOK();
+    const joinConfJMT = this.getJoinConfJMT();
+
+    if (clickToInterstitial && interstitialToJoinOk && joinConfJMT) {
```
maxinteger, line 275: is it OK to return with undefined when one or all of the values are 0

packages/@webex/internal-plugin-metrics/src/ca-metrics-latencies.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /**
+   * Client JMT
+   * @returns - latency
+   */
+  public getClientJMT() {
+    const interstitialToJoinOk = this.getInterstitialToJoinOK();
+    const joinConfJMT = this.getJoinConfJMT();
+
+    if (interstitialToJoinOk && joinConfJMT) {
```
maxinteger, line 290: is it OK to return with undefined when one or both of the values are 0

packages/@webex/internal-plugin-metrics/src/ca-metrics.ts
```diff
@@ -0,0 +1,312 @@
...
+   */
+  constructor() {
+    if (!CallAnalyzerMetrics.instance) {
+      CallAnalyzerMetrics.instance = this;
+      this.meetingCollection = null;
+      this.latencies = new CallAnalyzerLatencies();
+    }
+
+    // eslint-disable-next-line no-constructor-return
+    return CallAnalyzerMetrics.instance;
```
maxinteger, line 63: surprisingly this work as it is expected. Although you export a single instance, this is useful to prevent creating more instances via `new instance.constructor`

packages/@webex/internal-plugin-metrics/src/ca-metrics.ts
```diff
@@ -0,0 +1,312 @@
...
+  /**
+   * Submit Client Event CA event
+   * @param event - event key
+   * @param payload - payload for the event
+   * @param options - payload
+   * @returns
+   */
+  public submitClientEvent({
+    name,
+    // additional payload to be merged with default payload
```
maxinteger, line 247: this comment should be in the JSDoc

packages/@webex/internal-plugin-metrics/src/ca-metrics.util.ts
```diff
@@ -0,0 +1,69 @@
...
+      'webex-js-sdk/%s %s',
+      `${process.env.NODE_ENV}-${webexVersion}`,
+      userAgentOption
+    );
+  }
+
+  return util.format('webex-js-sdk/%s', `${process.env.NODE_ENV}-${webexVersion}`);
+};
+
+export const clearEmpty = (obj) => {
```
maxinteger, line 52: Pleas add some JSDoc and maybe change the name of the function, because it is not obvious what happening here

packages/@webex/internal-plugin-metrics/src/ca-metrics.util.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Check if the object is empty
+  if (Object.keys(obj).length === 0) {
+    return;
+  }
+
+  Object.keys(obj).forEach((key) => {
+    if ((typeof obj[key] === 'object' || Array.isArray(obj[key])) && isEmpty(obj[key])) {
+      delete obj[key];
```
maxinteger, line 60: in general `delete` is not a good idea to use because it ban mass up the JS engine runtime optimisation.

Since this function called frequently I recommend to create a shallow clone of the object when it is necessary

maxinteger (changes requested):

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -0,0 +1,178 @@
...
+    // eslint-disable-next-line no-constructor-return
+    return Metrics.instance;
+  }
+
+  /**
+   * Initialize Call diagnostic class
+   * @param meetingCollection
+   * @param webex
+   */
+  initialSetupCallDiagnosticMetrics(meetingCollection: any, webex: object) {
```
torgeadelin, line 61: Since this Metrics class extends WebexPlugin, I'm not entirely sure if it's ok to consider it as a Singleton?
Also, at the moment i'm storing webex object and meetingCollection only on the `CallDiagnosticMetrics` class, but I'm thinking wether it makes sense now.. since this (Metrics) class is a WebexPlugin, maybe I already have access to both?


Not sure how these 2 things work, so will need some imput on this one.

torgeadelin: CC: @maxinteger @marcin-bazyl @arun3528

packages/@webex/internal-plugin-metrics/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+interface GenericMetric {
+  name: string;
+  payload?: any;
+  options?: any;
+}
+
+export interface ClientEvent extends GenericMetric {
+  name: RawClientEvent['name'];
+  payload?: RawClientEvent;
+  options: SubmitClientEventOptions;
```
maxinteger, line 21: ideally you should have separated definition for each event:
```ts
interface ClientFeatureAudioNoiseRemoval {
  name:"client.feature.audio.noise.removal",
  payload: ClientFeatureAudioNoiseRemovalPayload
}

interface ClientFeatureEmbedObjectInfo {
  name:"client.feature.embedded-object-info",
  payload: ClientFeatureEmbedObjectInfoPayload
}

interface ClientFeatureGestureRecognition {
  name:"client.feature.embedded-object-info",
  options: ClientFeatureGestureRecognitionOptions
}

// ...


type ClientEvent = 
  | ClientFeatureAudioNoiseRemoval 
  | ClientFeatureEmbedObjectInfo 
  | ClientFeatureGestureRecognition
  | ...
```
some for the other events

packages/@webex/internal-plugin-metrics/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export interface BehavioralEvent extends GenericMetric {
+  name: 'client.interstitial.joinButton.clicked' | 'client.meeting.initialize';
+  payload?: {some: 'string'};
+  options?: never;
+}
+
+export interface OperationalEvent extends GenericMetric {
+  name: 'a' | 'b' | 'c'; // TODO: not implemented
```
maxinteger, line 31: if it is not ready, the either remove it for now or use `name: never`

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
maxinteger: you do not export the `Metrics` class only one instance, that should cover the Singleton  part.

Yes, you can use `this.webex` in `Metrics`, but probably better to store it in `CallDiagnosticMetrics` as well

maxinteger (changes requested):
I found few issues

packages/@webex/internal-plugin-metrics/src/call-diagnostic/types/ClientEvent.json
```diff
@@ -0,0 +1,5428 @@
+{
+  "$ref": "#/definitions/ClientEvent",
```
arun3528, line 2: does this cover all the call analyzer events ?

torgeadelin: Call Analyzer team defined multiple types of events in the https://sqbu-github.cisco.com/WebExSquared/event-dictionary repo.

the main categories are:
- ClientEvent (used to determine JMF, JMS, JMT, FFD, share results, call drops etc)
- MQE (Media Quality Events)
- FeatureUsageEvents (tracking feature usage on UCF, data that gets processed and vizualised in webex oneview.

All those types are defined in RAML file, but we managed to get them converted to typescript and JSON schema, so when we build the events we can assure the structure (types) and semantics (JSON schema validation).

packages/@webex/internal-plugin-metrics/src/index.ts
```diff
@@ -7,11 +7,16 @@ import '@webex/internal-plugin-device';
...
+import {ClientEvent} from './call-diagnostic/types/ClientEvent';
 
 registerInternalPlugin('metrics', Metrics, {
   config,
 });
 
+// eslint-disable-next-line no-restricted-exports
 export {default, getOSNameInternal} from './metrics';
-export {config};
+export {config, NewMetrics, userAgentToString};
```
marcin-bazyl, line 21: just to confirm, once you do all these changes, we will be exporting only the NewMetrics and not the old ones, right?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -0,0 +1,299 @@
...
+    this.latencyTimestamps.set(key, value);
+  }
+
+  /**
+   * Helper to calculate end - start
+   * @param a start
+   * @param b end
+   * @returns latency
+   */
+  private getDiffBetweenLatencies(a: MetricEventNames, b: MetricEventNames) {
```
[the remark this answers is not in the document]
torgeadelin: refactored in new pr

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -0,0 +1,299 @@
...
+    return undefined;
+  }
+
+  /**
+   * Total Media JMT
+   * @returns - latency
+   */
+  public getTotalMediaJMT() {
+    const clickToInterstitial = this.getClickToInterstitial();
+    const interstitialToJoinOk = this.getInterstitialToJoinOK();
```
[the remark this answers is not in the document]
torgeadelin: this formula is from CA, I didn't create it :/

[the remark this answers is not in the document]
torgeadelin: https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=342335050

packages/@webex/internal-plugin-metrics/src/index.ts
torgeadelin: yes

PR #2853, opened 2023-06-06
feat: transcoded meetings APIs using webrtc-core local tracks

marcin-bazyl:
# COMPLETES #SPARK-399695

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-399695

SDK APIs for transcoded meetings are still using streams instead of local tracks from webrtc-core.

- change all media related transcoded APIs to use local tracks so that they are the same for multistream and transcoded meetings.
- remove old APIs like updateAudio, updateVideo, etc.

More details on this wiki:
https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/SDK+changes+required+for+moving+to+new+webrtc-core

Corresponding web app PR: https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/4429

This is an API breaking change, but I'm not ticking the box in the section below, because we don't want a new major version being released. This PR is for beta branch and beta as a whole is a breaking change.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

manually tested end-2-end with the sample app and web app + unit tests and integration tests

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3130,66 +3126,6 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    return !!this.audio;
-  }
-
-  /**
-   * Convenience function to tell whether a meeting is muted
-   * @returns {Boolean} if meeting audio muted or not
-   * @public
-   * @memberof Meeting
-   */
-  public isAudioMuted() {
```
arun3528, line 3149: yep this wont be needed any more

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3419,274 +3340,110 @@ export default class Meeting extends StatelessWebexPlugin {
...
-
-    if (rawAudioTrack) {
-      const settings = rawAudioTrack.getSettings();
+  private setLocalAudioTrack(localTrack?: LocalMicrophoneTrack) {
+    const oldTrack = this.mediaProperties.audioTrack;
 
-      const localMicrophoneTrack = new LocalMicrophoneTrack(
-        MediaUtil.createMediaStream([rawAudioTrack])
-      );
+    oldTrack?.off(LocalTrackEvents.Muted, this.localAudioTrackMuteStateHandler);
```
arun3528, line 3352: in a normal scenario this wont happen if we have the device change logic in the local track

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3784,285 +3544,73 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      return Promise.reject(new UserNotJoinedError());
-    }
-
-    // @ts-ignore
-    if (!this.mediaId) {
-      // Happens when addMedia and mute are triggered in succession
-      return Promise.reject(new NoMediaEstablishedYetError());
-    }
+  public joinWithMedia(
+    options: {
```
arun3528, line 3563: do you know if lobby scenario we are able to send media , any thoughts on join and joinWithMedia 

2

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1190,84 +1130,45 @@ function setAudioOutputDevice() {
     });
 }
 
+leet = audioEnabled = true;
```
meta-dipanshu-sharma, line 1133: is this a typo for `let audioEnabled`?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1329,25 +1223,19 @@ async function stopScreenShare() {
...
+    if (localMedia.screenShare.audio) {
+      tracksToUnpublish.push(localMedia.screenShare.audio);
     }
-    else {
-      await meeting.stopShare();
+    if (localMedia.screenShare.video) {
+      tracksToUnpublish.push(localMedia.screenShare.video);
     }
+
+    if (tracksToUnpublish.length > 0) {
```
meta-dipanshu-sharma, line 1235: small thing, only `if(tracksToUnpublish.length)` will work

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: if we do the webrtc-core changes to allow device changing within same local track instance, then this code that listens on the underlying track change will handle it and we need the underlying track change anyway for effects to work too

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: done

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: yeah, typo and anyway `audioEnabled` was not used, so removed it

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5574,24 +4951,27 @@ export default class Meeting extends StatelessWebexPlugin {
...
-   * Specify joining via audio (option: pstn), video, screenshare
+   * Creates a media connection to the server. Media connection is required for sending or receiving any audio/video.
+   * In order to send audio/video, call publishTracks() after addMedia() resolves.
+   *
    * @param {Object} options A configurable options object for joining a meeting
-   * @param {Object} options.resourceId pass the deviceId
-   * @param {MediaDirection} options.mediaSettings pass media options
-   * @param {MediaStream} options.localStream
-   * @param {MediaStream} options.localShare
+   * @param {Object} options.audioEnabled set it to false if you don't want to send or receive any audio (default is true)
```
Coread, line 4958: is the `@param` here really Object for audioEnabled/videoEnabled/receiveShare?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5574,24 +4951,27 @@ export default class Meeting extends StatelessWebexPlugin {
   }
 
   /**
-   * Specify joining via audio (option: pstn), video, screenshare
+   * Creates a media connection to the server. Media connection is required for sending or receiving any audio/video.
+   * In order to send audio/video, call publishTracks() after addMedia() resolves.
+   *
    * @param {Object} options A configurable options object for joining a meeting
```
Coread, line 4957: you could update this to be `AddMediaOptions`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5633,34 +5017,59 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            },
-            preferTranscoding: !this.isMultistream,
-          },
-          {
-            // @ts-ignore
-            parent: this.webex,
-          }
-        );
-      })
+    // we set sendAudio & sendVideo to true even before any tracks are published
```
Coread, line 5020: The comment suggests that it will always set sendAudio/sendVideo to true, but in reality it just defaults to true. It could still be passed in as false.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5734,7 +5143,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
             // eslint-disable-next-line func-names
             // eslint-disable-next-line prefer-arrow-callback
-            if (this.type === _CALL_) {
+            if (this.type === _CALL_ || this.meetingState === FULL_STATE.ACTIVE) {
```
Coread, line 5146: what is FULL_STATE.ACTIVE?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5927,406 +5332,102 @@ export default class Meeting extends StatelessWebexPlugin {
...
-          break;
-        case MEDIA_UPDATE_TYPE.VIDEO:
-          this.updateVideo(options).then(pendingPromiseResolve, pendingPromiseReject);
-          break;
-        case MEDIA_UPDATE_TYPE.SHARE:
-          this.updateShare(options).then(pendingPromiseResolve, pendingPromiseReject);
+        case MEDIA_UPDATE_TYPE.TRANSCODED_MEDIA_CONNECTION:
+          mediaUpdate = this.updateTranscodedMediaConnection();
           break;
         case MEDIA_UPDATE_TYPE.LAMBDA:
```
Coread, line 5341: suggestion: LAMBDA is used twice, and both times it is used to call requestScreenShareFloor. This suggests that there isn't a need for a arbitrary type LAMBDA for the media update queue and instead could be a SHARE type.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6568,55 +5669,67 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+            return Promise.reject(error);
           });
+      }
 
-          return Promise.reject(error);
-        });
+      return Promise.reject(new ParameterError('Cannot share without content.'));
     }
+    this.floorGrantPending = true;
```
Coread, line 5716: question: Is this meant to set the floorGrantPending outside of the if JOINED?

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yeah, this just saves 1s on addMedia() call because previously we were always starting the joiningTimer (line 5149) and checking the state after 1s, but the state might be already ACTIVE so no need to start the timer in that case. This is especially true in all unit tests - all addMedia() tests used to take at least 1s, now they all finish in a few milliseconds.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yes, the logic hasn't changed - it covers the case when we don't receive the Locus DTO with meeting state JOINED before someone calls addMedia with a screen share track. Previously the check `if (this.state === MEETING_STATE.STATES.JOINED)` and setting of `this.floorGrantPending = true;` was done outside the call to `requestScreenShareFloor()` in just 1 place, I moved it inside `requestScreenShareFloor()` so that the check is always done.

packages/@webex/plugin-meetings/src/meeting/index.ts

packages/@webex/plugin-meetings/src/meeting/index.ts
Coread: It can wait until the next PR

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -143,12 +138,29 @@ const logRequest = (request: any, {logText = ''}) => {
...
 };
 
+export type LocalTracks = {
+  microphone?: LocalMicrophoneTrack;
+  camera?: LocalCameraTrack;
+  screenShare?: {
+    audio?: LocalTrack; // todo: for now screen share audio is not supported (will be done in SPARK-399690)
+    video?: LocalDisplayTrack;
+  };
+  annotationInfo?: AnnotationInfo;
```
marcin-bazyl, line 148: When merging latest master into my branch I've found that this annotation info has been recently added to publishTracks() API (in this PR:https://github.com/webex/webex-js-sdk/pull/2894). I think it doesn't belong here, it should probably be a property on the LocalDisplayTrack itself, but for now I've kept it and will discuss this with the Marmot team next week after their public holidays.

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl, line 1: the changes in the sample app are just a minimum set of changes to make it work. With the new "local tracks" approach, the whole UI of the sample app really needs to be redesigned to allow tracks to live outside of meetings and to be re-used between meetings, etc. This is covered by SPARK-436949.

Also the readme file and other JS-SDK docs need to be updated, this will be done in SPARK-436950

Coread (approved):

arun3528 (approved):

meta-dipanshu-sharma (approved):

PR #2859, opened 2023-06-07
feat(plugin-meetings): preferLiveVideo

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-434080
We need the ability to change the preferLiveVideo remoteMediaManager config option from Cantina.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -151,6 +151,11 @@ export class RemoteMediaGroup {
     throw new Error(`remote media object ${remoteMedia.id} not found in the group`);
   }
 
+  public setPreferLiveVideo(preferLiveVideo: boolean) {
```
shnaaz, line 154: May be rename this to updatePreferLiveVideoOption

marcin-bazyl: I think I prefer existing name `setPreferLiveVideo`

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -151,6 +151,11 @@ export class RemoteMediaGroup {
...
+    this.options.preferLiveVideo = preferLiveVideo;
+    this.sendActiveSpeakerMediaRequest(false);
```
marcin-bazyl, line 156: hardcoded commit=false means that nothing will happen, the change will not take effect. We need `commit: boolean` to be added as an argument to `setPreferLiveVideo()`.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -490,6 +490,16 @@ export class RemoteMediaManager extends EventsScope {
     return this.currentLayoutId;
   }
 
+  /**
+   * sets the preferLiveVideo
+   */
+  public setPreferLiveVideo(preferLiveVideo: boolean) {
+    this.config.video.preferLiveVideo = preferLiveVideo;
```
marcin-bazyl, line 497: I think it would be useful to add a log here so that we know when this API is called - this may be relevant when debugging issues

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -490,6 +490,16 @@ export class RemoteMediaManager extends EventsScope {
...
+    Object.values(this.media.video.activeSpeakerGroups).forEach((activeSpeakerGroup) => {
+      activeSpeakerGroup.setPreferLiveVideo(preferLiveVideo);
```
marcin-bazyl, line 499: once you add the commit parameter to `RemoteMediaGroup.setPreferLiveVideo()` we will need to call it with commit being `false` here and after the loop call `this.mediaRequestManagers.video.commit();`

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
```diff
@@ -97,6 +97,68 @@ describe('RemoteMediaGroup', () => {
...
+      const group = new RemoteMediaGroup(fakeMediaRequestManager, fakeReceiveSlots, 255, true, {
+        resolution: 'medium',
+        preferLiveVideo: false,
+      });
+
+      group.setPreferLiveVideo(true);
+
+      assert.calledOnce(fakeMediaRequestManager.cancelRequest);
+
+      assert.calledTwice(fakeMediaRequestManager.addRequest);
```
marcin-bazyl, line 112: why twice? I guess the first one is from the constructor call in line 103. It would be better to call `fakeMediaRequestManager.addRequest.resetHistory()` before calling `setPreferLiveVideo()` so that here we would expect only 1 call - that would make the test more clear and precise.

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
```diff
@@ -97,6 +97,68 @@ describe('RemoteMediaGroup', () => {
...
+          receiveSlots: fakeReceiveSlots,
+          codecInfo: sinon.match({
+            codec: 'h264',
+            maxFs: 3600,
+          }),
+        })
+      );
+    });
+
+    it('sets to false initial option was true', () => {
```
marcin-bazyl, line 131: I don't feel there is much value in having 2 same tests with just the value changed true vs false - for me 1 would be enough, but there is another more interesting case, so maybe we should have these 2 tests: 
1. "setPreferLiveVideo() called with different value",
2. "setPreferLiveVideo() called with same value" - in that case we could detect it and avoid doing anything
Or if you prefer to test all boolean values then we should have tests for all 4 cases (all combinations of initial value and new value) and the tests should be parametrized to avoid duplication

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -663,6 +663,56 @@ describe('RemoteMediaManager', () => {
...
+  
+      await remoteMediaManager.start();
+  
+      resetHistory();
+      remoteMediaManager.setPreferLiveVideo(true);
+
+      expect(config.video.preferLiveVideo).to.equal(true);
+
+      Object.values(remoteMediaManager.media.video.activeSpeakerGroups).forEach((activeSpeakerGroup: any) => {
+        expect(activeSpeakerGroup.options.preferLiveVideo).to.equal(true);
```
marcin-bazyl, line 688: it would be better to check that RemoteMediaGroup.setPreferLiveVideo() was called rather than accessing private options of RemoteMediaGroup, because that's an implementation detail of RemoteMediaGroup that RemoteMediaManager and its tests should not know about

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -663,6 +663,56 @@ describe('RemoteMediaManager', () => {
...
+      );
+  
+      await remoteMediaManager.start();
+  
+      resetHistory();
+      remoteMediaManager.setPreferLiveVideo(true);
+
+      expect(config.video.preferLiveVideo).to.equal(true);
+
+      Object.values(remoteMediaManager.media.video.activeSpeakerGroups).forEach((activeSpeakerGroup: any) => {
```
marcin-bazyl, line 687: We shouldn't be accessing `remoteMediaManager.media` - that's private. Typescript normally wouldn't even allow you to do that, but it seems it's somehow misconfigured in JS-SDK and doesn't throw the error. 
Instead we should iterate all active speaker groups that client gets notified with via Event.VideoLayoutChanged - see how it's done in other tests in this file

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -663,6 +663,56 @@ describe('RemoteMediaManager', () => {
...
+        expect(activeSpeakerGroup.options.preferLiveVideo).to.equal(true);
+      });
+    });
+
+    it('sets preferLiveVideo to false', async () => {
```
marcin-bazyl, line 692: same as with RemoteMediaGroup tests -  I don't see much value in doing same test with true and false, but if you do then please make it a parametrized one to avoid duplication. More useful is to test if new value is same or different than the old one

marcin-bazyl (changes requested):
We also need to update the SDK sample app to have a UI for toggling "prefer live video"

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -663,6 +663,36 @@ describe('RemoteMediaManager', () => {
...
+      await remoteMediaManager.start();
+  
+      resetHistory();
+      remoteMediaManager.setPreferLiveVideo(true);
+
+      expect(config.video.preferLiveVideo).to.equal(true);
+      
+      remoteMediaManager.on(Event.VideoLayoutChanged, (layoutInfo: VideoLayoutChangedEventData) => {
+        Object.values(layoutInfo.activeSpeakerVideoPanes).forEach((group) =>
+          assert.calledOnceWithExactly(group.setPreferLiveVideo, true, false)
```
marcin-bazyl, line 689: I don't think this assert is ever executed. You need to register the listener for `Event.VideoLayoutChanged` before calling `start()`, because that's when that event is emitted.

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -663,6 +663,40 @@ describe('RemoteMediaManager', () => {
...
+      remoteMediaManager.on(Event.VideoLayoutChanged, (layoutInfo: VideoLayoutChangedEventData) => {
+        Object.values(layoutInfo.activeSpeakerVideoPanes).forEach((group) => stubs.push(sinon.stub(group, 'setPreferLiveVideo')));
+
+      });
+      await remoteMediaManager.start();
+      resetHistory();
+      await remoteMediaManager.setPreferLiveVideo(true);
+
+
+      stubs.forEach((stub) => {
```
marcin-bazyl, line 690: nitpick: we could also add an assert that `stubs.length > 0`, because if for some weird reason we would not get any activeSpeakerVideoPanes or no VideoLayoutChanged event, then the test would still pass

marcin-bazyl (approved):

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -34,6 +34,7 @@ const integrationEnv = document.getElementById('integration-env');
 const turnDiscoveryCheckbox = document.getElementById('enable-turn-discovery');
 const eventsList = document.getElementById('events-list');
 const multistreamLayoutElm = document.querySelector('#multistream-layout');
+const preferLiveVideoElm = document.quesrySelector('#prefer-live-video')
```
Coread, line 37: typo

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1965,6 +1966,28 @@ function updateMultistreamVideoLayout() {
...
+  let preferLiveVideo = false;
+  const meeting = getCurrentMeeting();
+
+  if (!meeting) {
+    return;
+  }
+  if (!meeting.mediaProperties.webrtcMediaConnection) {
+    return;
+  }
+  const value = document.getElementById("prefer-live-video").value;
```
marcin-bazyl, line 1979: once you fix `preferLiveVideoElm` then we should use it here

PR #2860, opened 2023-06-07
feat(plugin-meetings): fallback to getting preferred site from user

Coread:

Sometimes the request to fetch the preferred site information from hydra fails. In that case fall back to getting it from the user info.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -953,6 +953,8 @@ export default class Meetings extends WebexPlugin {
               user?.userPreferences?.userPreferencesItems?.preferredWebExSite;
             if (preferredWebexSite) {
               this.preferredWebexSite = preferredWebexSite;
+            } else {
+              throw new Error('site not found');
             }
           })
           .catch(() => {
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -942,6 +942,27 @@ export default class Meetings extends WebexPlugin {
...
+
+      // fall back to getting the preferred site from the user information
+      if (!this.preferredWebexSite) {
+        // @ts-ignore
+        return this.webex.internal.user
+          .get()
+          .then((user) => {
+            const preferredWebexSite =
+              user?.userPreferences?.userPreferencesItems?.preferredWebExSite;
+            if (preferredWebexSite) {
```
maxinteger, line 954: from the result point of view when `preferredWebexSite` is `undefined`, is the same as failing fetch. We should log this case as well.

You can and an else branch and throw an error so the catch will log both cases.

maxinteger (approved):
just one suggestion, otherwise looks good

packages/@webex/plugin-meetings/src/meetings/index.ts
Coread: I've now added this

PR #2874, opened 2023-06-12
feat: add media effects in plugin-meetings

meta-dipanshu-sharma:
# COMPLETES #<[ SPARK-430464 ](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-430464)>

This PR covers the changes required to import the media helper in the plugin meetings and creating a wrapper around `NoiseReductionEffect` and `VirtualBackgroundEffect` and export it from plugin-meetings.

Created a wrapper for `NoiseReductionEffect` and `VirtualBackgroundEffect` and exported it from plugin-meetings.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -59,6 +59,7 @@ import CaptchaError from '../common/errors/captcha-error';
 import MeetingCollection from './collection';
 import MeetingsUtil from './util';
 import PermissionError from '../common/errors/permission';
+import {INoiseReductionEffect, IVirtualBackgroundEffect} from './meetings.types';
 
 let mediaLogger;
 
@@ -810,12 +811,12 @@ export default class Meetings extends WebexPlugin {
   /**
    * Creates a noise reduction effect
    *
-   * @param {Object} options optional custom effect options
+   * @param {INoiseReductionEffect} options optional custom effect options
    * @returns {Promise<effect>} noise reduction effect.
    * @public
    * @memberof Meetings
    */
-  createNoiseReductionEffect = async (options = {}) => {
+  createNoiseReductionEffect = async (options?: INoiseReductionEffect) => {
     // @ts-ignore
     const authToken = this.webex.credentials.supertoken.access_token;
 
@@ -825,12 +826,12 @@ export default class Meetings extends WebexPlugin {
   /**
    * Creates a virtual background effect
    *
-   * @param {Object} options optional custom effect options
+   * @param {IVirtualBackgroundEffect} options optional custom effect options
    * @returns {Promise<effect>} virtual background effect.
    * @public
    * @memberof Meetings
    */
-  createVirtualBackgroundEffect = async (options = {}) => {
+  createVirtualBackgroundEffect = async (options?: IVirtualBackgroundEffect) => {
     // @ts-ignore
     const authToken = this.webex.credentials.supertoken.access_token;
 
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -808,6 +808,36 @@ export default class Meetings extends WebexPlugin {
   }
 
   /**
+   * Creates a noise reduction effect
+   *
+   * @param {Object} options optional custom effect options
+   * @returns {Promise<effect>} noise reduction effect.
+   * @public
+   * @memberof Meetings
+   */
+  createNoiseReductionEffect = async (options = {}) => {
+    // @ts-ignore
+    const authToken = this.webex.credentials.supertoken.access_token;
+
+    return new mediaHelpersModule.NoiseReductionEffect({authToken, ...options});
+  };
+
+  /**
+   * Creates a virtual background effect
+   *
+   * @param {Object} options optional custom effect options
+   * @returns {Promise<effect>} virtual background effect.
+   * @public
+   * @memberof Meetings
+   */
+  createVirtualBackgroundEffect = async (options = {}) => {
+    // @ts-ignore
+    const authToken = this.webex.credentials.supertoken.access_token;
+
+    return new mediaHelpersModule.VirtualBackgroundEffect({authToken, ...options});
+  };
+
+  /**
    * Uploads logs to the webex services for tracking
    * @param {Object} [options={}]
    * @param {String} [options.callStart] Call Start Time
--- a/packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
+++ b/packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
@@ -34,6 +34,7 @@ import CaptchaError from '@webex/plugin-meetings/src/common/errors/captcha-error
 import { forEach } from 'lodash';
 import PasswordError from '@webex/plugin-meetings/src/common/errors/password-error';
 import PermissionError from '@webex/plugin-meetings/src/common/errors/permission';
+import {NoiseReductionEffect,VirtualBackgroundEffect} from '@webex/media-helpers';
 
 describe('plugin-meetings', () => {
   const logger = {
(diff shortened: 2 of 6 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -807,6 +807,36 @@ export default class Meetings extends WebexPlugin {
     );
   }
 
+  /**
+   * Creates a noise reduction effect
+   *
+   * @param {Object} options optional custom effect options
```
arun3528, line 813: would import the types from effect and remove the access token from it , we will need to be more specific when it comes to types 

same for noise

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -807,6 +807,36 @@ export default class Meetings extends WebexPlugin {
...
+   * @returns {Promise<effect>} noise reduction effect.
+   * @public
+   * @memberof Meetings
+   */
+  createNoiseReductionEffect = async (options = {}) => {
+    // @ts-ignore
```
sreenara, line 819: What complaint from ts are we ignoring here?

sreenara (commented):
I see that there are functions to return the effect to the developer. But how does this effect get applied in the ongoing meeting object?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -807,6 +807,36 @@ export default class Meetings extends WebexPlugin {
...
 
+  /**
+   * Creates a noise reduction effect
+   *
+   * @param {Object} options optional custom effect options
+   * @returns {Promise<effect>} noise reduction effect.
+   * @public
+   * @memberof Meetings
+   */
+  createNoiseReductionEffect = async (options = {}) => {
```
marcin-bazyl, line 818: wouldn't it be better for these "create effects" functions be all in a new separate ts file? We currently have createMicrophoneTrack, createCameraTrack, createDisplayTrack in media-helpers and they are just re-exported in plugin-meetings. Shouldn't we do the same with effects? They could be in media-helpers, just in a separate file, not webrtc-core.ts

packages/@webex/plugin-meetings/src/meetings/index.ts
marcin-bazyl: +1 from me on this one, any new API we add has to be fully typed and I think we should also do something about jsdocs, because right now they require specifying types which is a duplication of the types already declared in typescript and it's a potential source of problems if there is a mismatch

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -1601,9 +1693,13 @@ describe('plugin-meetings', () => {
       });
 
       it('if newLocus replaceAt time is expired, then return false', () => {
-        sinon.stub(webex.meetings.meetingCollection, 'getActiveBreakoutLocus').returns({joinedWith: {replaces: [{
+        sinon.stub(webex.meetings.meetingCollection, 'getActiveBreakoutLocus').returns({
```
marcin-bazyl, line 1696: why are we doing formatting changes? Please avoid any unnecessary formatting changes as I have a bit PR coming up for local tracks for transcoded meetings and these changes will cause a lot of conflicts

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -352,6 +352,98 @@ describe('plugin-meetings', () => {
         });
       });
 
+      describe('virtual background effect', () => {
```
marcin-bazyl, line 355: none of the tests check that the right constructor was called

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meetings/index.ts
meta-dipanshu-sharma: 1. added the types for both the methods by removing `authToken` which we're explicitly passing to effects constructor.

2. for `jsdocs`, should we start not adding jsdocs to typescript files as anyway ts methods will annotate types on its own? It is recommended to use **jsdocs with js** and **typescript alone** but here we're doing **jsdocs with typescript** which is anyways not recommended.

we can discuss more if required on this topic.

packages/@webex/plugin-meetings/src/meetings/index.ts
meta-dipanshu-sharma: 1. previously I wrote a method where we were **creating the effect** and was **applying it on the passed audio/video track** in the same method and also **enabling it** at the end of the method.

But after discussion with Arun/Marcin we came to a conclusion that this is not the approach we planned to take instead just **return the effect** and let developers use it on their side of code. We will create a detailed wiki explaining the steps to use these steps on the tracks. I updated the code as per the discussed approach.

2. for using @ts-ignore, we are ignoring the warning which is `Property 'webex' does not exist on type 'Meetings'`

packages/@webex/plugin-meetings/src/meetings/index.ts
meta-dipanshu-sharma: Yes this could be an approach we could take but what I discussed with Arun is that we were planning to rename `media-helpers` package to `internal-media-helpers` as the effects imported and exported from media-helpers package won't run without an access_token and we want developers to use it with `meetings` plugin once this PR is merged as we have created a wrapper here which doesn't require token from user.

Also if we move effects wrapper logic there then developers using `meetings` package have to import `media-helpers` package also but with the current approach they can directly access it on `meetings` object.

I'm open for more discussion if we want to move this logic on media-helpers package.

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
meta-dipanshu-sharma: I'll remove formatting changes.

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
meta-dipanshu-sharma: In these tests we're not stubbing the constructor and using the real one so without stub we can't test whether it's called or not. Do we want me to mock the effects constructor and then test it?

packages/@webex/plugin-meetings/src/meetings/index.ts
marcin-bazyl: developers don't have to import media-helpers, you can export stuff from media-helpers in plugin-meetings - that's what we're already doing for createMicrophoneTrack, createCameraTrack, createDisplayTrack etc

packages/@webex/plugin-meetings/src/meetings/index.ts
meta-dipanshu-sharma: Agreed. Will discuss it with the team how we can proceed further on this. Will have a separate Jira for this.

packages/@webex/plugin-meetings/src/meetings/index.ts
meta-dipanshu-sharma: `media-helpers` is just plain plugin for exporting tracks and effects related things and we won't have access to `webex` object there. That's why `createEffect` logic is in plugin-meetings where we have access to webex object.

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
meta-dipanshu-sharma: Added the `instanceOf` check in test cases.

packages/@webex/plugin-meetings/src/meetings/index.ts
marcin-bazyl: ok, fine let's keep in in meetings

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -807,6 +808,36 @@ export default class Meetings extends WebexPlugin {
...
+
+  /**
+   * Creates a virtual background effect
+   *
+   * @param {Partial<IVirtualBackgroundEffect>} options optional custom effect options
+   * @returns {Promise<effect>} virtual background effect.
+   * @public
+   * @memberof Meetings
+   */
+  createVirtualBackgroundEffect = async (options?: Partial<IVirtualBackgroundEffect>) => {
```
marcin-bazyl, line 834: why is `Partial` used here but not in `createNoiseReductionEffect()`? I guess you're using partial because you don't want the authToken. I think it's better to use [typescript's omit](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) to define the new type without the authToken

meta-dipanshu-sharma: Yes I'm already using `Omit` syntax to remove authToken but some of the properties were required in NoiseReductionEffect `constructor` apart from authToken so that's why used Partial as well. But just checked on the constructor of NoiseReductionEffect, `web-media-effects` library already using Partial so we don't have to pass.

Removing it from here then.

PR #2878, opened 2023-06-14
feat(calling): added new calling class

Kesari3008:

Created global Calling class which will take care of below:
1. If Webex object is provided then store it inside SDK connector. Proceed with creating client objects based on the config received.
2. If Webex is not present and webexConfig is shared by developer, we will follow below:

- Initialize the Webex object, emit 'calling:ready' event once webex is ready. 
- On receiving the 'calling:ready' event, developer can trigger register() API under calling class.
- register() API will trigger WDM registration, mercury connection.
- Once above actions are completed, we will create client objects based on the calling config received.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

### I certified that

---


packages/calling/calling/src/Calling/Calling.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  public callSettingsClient!: ICallSettings;
+
+  public voicemailClient!: IVoicemail;
+
+  /**
+   *
+   * @param callingConfig
+   * @param webex
+   * @param webexConfig
```
meta-dipanshu-sharma, line 45: going forward we should stop adding jsdocs in ts file.

meta-dipanshu-sharma:
I'm curious about the folder structure. Why we have `calling/calling/*` type of path?

packages/calling/calling/src/Calling/Calling.ts
Kesari3008: Ohh I didn't add this..seems to have added on its own when i ran lint --fix command. Let me fix it

packages/webex/src/calling.js
```diff
@@ -0,0 +1,49 @@
+import {
```
sreenara, line 1: What's this calling.js file that is created here?

packages/calling/calling/src/Calling/Calling.ts
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable tsdoc/syntax */
```
sreenara, line 1: Why is this required?

packages/calling/calling/src/Calling/Calling.ts
```diff
@@ -0,0 +1,153 @@
...
+        SDKConnector.setWebex(this.webex);
+      }
+
+      this.initializeClients(callingConfig);
+    }
+    log.setLogger(callingConfig.logger.level);
+  }
+
+  /**
+   *
```
sreenara, line 73: Can we add a one line description above the function so that this gets added to the tsdoc automatically?

packages/calling/calling/src/Calling/Calling.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  public register() {
+    const logContext = {
+      file: CALLING_FILE,
```
sreenara, line 77: This constant should be within a constants file in the Calling folder. I don't think we should import it from CallingClient.

packages/calling/calling/src/Calling/Calling.ts
```diff
@@ -0,0 +1,153 @@
...
+      this.voicemailClient = createVoicemailClient(logger);
+    }
+  }
+
+  /**
+   * Function to return SDKConnector object.
+   *
+   * @returns SDKConnector.
+   */
+  public getSDKConnector(): ISDKConnector {
```
sreenara, line 138: This doesn't seem to be used anywhere?

packages/calling/calling/src/Calling/types.ts
```diff
@@ -0,0 +1,37 @@
...
+import {ICallSettings} from '../CallSettings/types';
+import {CallingClientConfig, ICallingClient} from '../CallingClient/types';
+import {IContacts} from '../Contacts/types';
+import {LoggerConfig} from '../Logger/types';
+import {ISDKConnector} from '../SDKConnector/types';
+import {IVoicemail} from '../Voicemail/types';
+import {Eventing} from '../Events/impl';
+
+interface ClientConfig {
+  calling?: boolean;
```
sreenara, line 12: Does boolean make sense here? You're setting it as optional, which means that this field doesn't need to be present in the object. So then setting this to false doesn't make sense right?

packages/calling/calling/src/Calling/types.ts
```diff
@@ -0,0 +1,37 @@
...
+  contact?: boolean;
+  history?: boolean;
+  settings?: boolean;
+  voicemail?: boolean;
+}
+
+export interface WebexConfig {
+  access_token: string;
```
sreenara, line 20: Are we not going to add any other members in the WebexConfig?
We should be giving the developer the ability to set additional parameters in the webex config right?

packages/calling/calling/src/Calling/types.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export interface CallingConfig {
+  clientConfig: ClientConfig;
+  callingClientConfig: CallingClientConfig;
+  logger: LoggerConfig;
+}
+
+export interface ICalling extends Eventing<CallingEventTypes> {
+  register: () => Promise<void>;
```
sreenara, line 30: Please set the members first followed by the methods.

sreenara (changes requested):
Please see the comments.
Also, please list out the files that are changed as part of this PR. The number of files in the PR is really large.

Kesari3008:
New PR raised: https://github.com/webex/webex-js-sdk/pull/2908

PR #2887, opened 2023-06-15
fix: changes required to work with new internal media core api

marcin-bazyl:
internal-media-core PR with the API changes: https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/183

This PR has a minimum set of changes in JS-SDK for it to work with internal-media-core from above PR. There should be no change in behavior here. 

It's a step towards having webrtc-core local tracks in transcoded meetings APIs

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/index.ts
+++ b/packages/@webex/plugin-meetings/src/media/index.ts
@@ -77,6 +77,22 @@ Media.getLocalMedia = (options: any, config: object) => {
   return Promise.resolve(undefined);
 };
 
+Media.getDirection = (receive: boolean, send: boolean) => {
+  if (!receive && !send) {
+    return 'inactive';
+  }
+
+  if (receive && send) {
+    return 'sendrecv';
+  }
+
+  if (receive) {
+    return 'recvonly';
+  }
+
+  return 'sendonly';
+};
+
 /**
  * creates a webrtc media connection with provided tracks and mediaDirection configuration
  *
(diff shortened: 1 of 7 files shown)


arun3528 (approved):

arun3528:
the changes looks good

marcin-bazyl:
interestingly the integration tests pass even though the PR is missing the update to package.json to use new internal-media-core (because internal-media-core changes haven't been merged yet). I think it's because now we're passing a string so any non-empty string gets evaluated as `true` so it works with the "old" API that was taking booleans

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -77,6 +77,22 @@ Media.getLocalMedia = (options: any, config: object) => {
...
+
+  if (receive && send) {
+    return 'sendrecv';
+  }
+
+  if (receive) {
+    return 'recvonly';
+  }
+
+  return 'sendonly';
```
torgeadelin, line 93: is it ok to assumed that the default will always be sendOnly?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6007,9 +6007,18 @@ export default class Meeting extends StatelessWebexPlugin {
                 : null,
             },
             receive: {
```
torgeadelin, line 6009: i think this property needs to be called direction?

torgeadelin (approved):

PR #2893, opened 2023-06-19
feat(ca-metrics): switch to new metrics (client events)

torgeadelin:

- switch from old postEvent to submitClientEvent
- it only includes the switching + unit tests
- it doesn't remove old code, that will be done separately
- it also includes some bug fixing (some events were not making sense or missing fields).

- refactor

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

-unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3844,10 +3868,10 @@ export default class Meeting extends StatelessWebexPlugin {
...
         .then(() => {
           MeetingUtil.handleAudioLogging(this.mediaProperties.audioTrack);
-          Metrics.postEvent({
-            event: eventType.MUTED,
-            meeting: this,
-            data: {trigger: trigger.USER_INTERACTION, mediaType: MetricsMediaType.AUDIO},
+          NewMetrics.submitClientEvent({
+            name: 'client.muted',
+            payload: {trigger: 'user-interaction', mediaType: 'audio'},
+            options: {meetingId: this.id},
```
marcin-bazyl, line 3874: just a thought: meetingId for all these events seems like something that's mandatory, so not sure if it should be in "options" - options sound like things that are "optional". Shouldn't meeting id be in payload.identifiers?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5310,14 +5330,15 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            errors: [
-              Metrics.generateErrorPayload(2004, false, MetricsError.name.MEDIA_ENGINE, undefined),
-            ],
+        NewMetrics.submitClientEvent({
+          name: 'client.ice.end',
+          payload: {canProceed: false},
+          options: {
+            meetingId: this.id,
+            error: {
+              // TODO: adapt this code to work in separate JIRA
```
marcin-bazyl, line 5339: does the jira exist? can you create it so that it doesn't get forgotten?

packages/@webex/plugin-meetings/test/unit/spec/breakouts/events.ts
```diff
@@ -30,17 +29,19 @@ describe('plugin-meetings', () => {
...
-            breakoutMoveId: 'breakoutMoveId',
-            breakoutSessionId: 'sessionId',
-            breakoutGroupId: 'groupId',
+        breakoutEvent.postMoveCallAnalyzer('client.breakout-session.join.response', eventInfo);
+        assert.calledWithMatch(NewMetrics.submitClientEvent, {
+          name: 'client.breakout-session.join.response',
+          payload: {
+            identifiers: {
+              breakoutMoveId: 'breakoutMoveId',
+              breakoutSessionId: undefined,
```
marcin-bazyl, line 40: why undefined values here? these are breakout unit tests so this doesn't feel right

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -1900,5 +1906,113 @@ describe('plugin-meetings', () => {
         );
       });
     });
+
+    describe('#isMeetingActive', () => {
+      it('sends client event correctly for state = inactive', () => {
```
marcin-bazyl, line 1911: nitpick: feels like these tests are all very similar and could be parametrized, but they're also quite short anyway so feel free to ignore this

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -55,7 +55,10 @@ const MeetingUtil = {
       );
     }
 
-    Metrics.postEvent({event: eventType.MEDIA_REQUEST, meeting});
+    NewMetrics.submitClientEvent({
```
marcin-bazyl, line 58: we should probably move the code for sending 'client.locus.media.request' and 'client.locus.media.response' into LocusMediaRequest.sendHttpRequest() function because that's when the /media request is actually being sent out (we can do it in another PR)

packages/@webex/plugin-meetings/src/roap/request.ts
```diff
@@ -96,7 +97,12 @@ export default class RoapRequest extends StatelessWebexPlugin {
       `Roap:request#sendRoap --> ${locusSelfUrl} \n ${roapMessage.messageType} \n seq:${roapMessage.seq}`
     );
 
-    Metrics.postEvent({event: eventType.MEDIA_REQUEST, meetingId});
+    NewMetrics.submitClientEvent({
```
marcin-bazyl, line 100: same like my other comment - we should move this into LocusMediaRequest.sendHttpRequest()

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -827,6 +830,38 @@ describe('plugin-meetings', () => {
               },
             });
           });
+
+          //TODO: can't seem to make this pass
```
marcin-bazyl, line 834: This test is failing because the assert is checked before the finally() code is executed. It's because `handleLocusEvent()` doesn't return the promise from the call to `this.create()`

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -430,10 +428,13 @@ export default class LocusInfo extends EventsScope {
         LoggerProxy.logger.warn(
           'Locus-info:index#isMeetingActive --> Call Ended, locus state is inactive.'
         );
-        Metrics.postEvent({
-          event: eventType.REMOTE_ENDED,
-          meetingId: this.meetingId,
+        NewMetrics.submitClientEvent({
+          name: 'client.call.remote-ended',
```
marcin-bazyl, line 432: we should also remove all the const values from eventType defined in src/metrics/config.ts (like REMOTE_ENDED in this case), because they're not used anymore

packages/@webex/plugin-meetings/src/breakouts/events.ts
```diff
@@ -1,37 +1,40 @@
...
-    event: events,
-    meetingId: eventInfo.meeting.id,
-    data: {
-      breakoutMoveId: eventInfo.breakoutMoveId,
-      breakoutSessionId: eventInfo?.currentSession?.sessionId,
-      breakoutGroupId: eventInfo?.currentSession?.groupId,
-    },
-  });
+const breakoutEvent: {
+  onBreakoutMoveRequest: any;
```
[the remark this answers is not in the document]
torgeadelin: i mean i still don't know what eventInfo holds, so replacing it with (eventInfo: any) => void; doesn't provide too much value either, but added it anyways.

packages/@webex/plugin-meetings/src/breakouts/events.ts
```diff
@@ -1,37 +1,40 @@
...
+  onBreakoutMoveRequest: (eventInfo) => {
+    breakoutEvent.postMoveCallAnalyzer('client.breakout-session.move.request', eventInfo);
+  },
+  onBreakoutMoveResponse: (eventInfo) => {
+    breakoutEvent.postMoveCallAnalyzer('client.breakout-session.move.response', eventInfo);
+  },
+  onBreakoutJoinResponse: (eventInfo) => {
+    breakoutEvent.postMoveCallAnalyzer('client.breakout-session.join.response', eventInfo);
+  },
+  postMoveCallAnalyzer: (event: ClientEvent['name'], eventInfo: any) => {
```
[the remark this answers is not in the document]
torgeadelin: I don't see any type created in the breakouts for it.. I'll leave it as it is.

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: The `payload` object contains "true" payload for the event, and by "true" I mean it follows the expected type/structure for the client event. It is not all of the payload tho, some of it is computed inside the new metrics class and we merge them together to form the final payload. The reason why meetingId is needed here is in order to compute the extra payload inside the class(locus related identifiers, correlationId, etc).

There will be events which won't require meetingId (as they will happen before the meeting exists), which is why it's optional.

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: yes, https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-435138

packages/@webex/plugin-meetings/test/unit/spec/breakouts/events.ts
torgeadelin: hmm, this is an actual bug! Thanks for catching it!
The previous implementation was broke, the parameters were not grabbed correctly from the eventInfo inside the source code.

torgeadelin: fixed now

packages/@webex/plugin-meetings/src/meeting/util.ts
torgeadelin: Have created a jira for this: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-440552

packages/@webex/plugin-meetings/src/roap/request.ts
torgeadelin: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-440552

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
torgeadelin: how can I test this then?

packages/@webex/plugin-meetings/src/locus-info/index.ts
torgeadelin: i have a separate JIRA to do cleanup

torgeadelin: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-440534

packages/@webex/plugin-meetings/src/breakouts/events.ts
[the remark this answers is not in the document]
marcin-bazyl: maybe ask Marmot team to add the types - they should know them

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
marcin-bazyl: you could change `handleLocusEvent()` to return a promise - the production code ignores its return value anyway, but just adding `await testUtils.flushPromises();` in the test before the assert also fixes the test and doesn't affect the production code, so probably better to just do that

marcin-bazyl (approved):

mkesavan13 (approved):
Looks good to me. Just a few questions out of curiosity - 

- In the new metrics, why are we sending only the meeting Id instead of the whole meeting object? Won't it be less information on the dashboard for us?
- Why are we removing all the constants and replacing them with plain strings?

torgeadelin:
@mkesavan13 
- meetingId is used to gather more information about the event inside the new metrics classes, it is not send over. Previously you would pass in either the meetingId or the whole meeting object. (but again, they were only used to grab other information about that specific event). that logic still applies, but the only difference is that im always passing down the id, and grab the meeting object internally.
- the new metrics classes are strongly typed, and therefore we don't need to store additional constants.

mkesavan13:
@torgeadelin That makes sense. Thank you.

PR #2897, opened 2023-06-21
fix(locus-info): adding lastModified field when handling requestedToU…

shivani1211:
# COMPLETES #SPARK-316107

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-316107

the issue that event `meeting:self:requestedToUnmute` is generated only once even if locus sends multiple requests with requestedToUnmute=true. This results in UI showing the 'requested to unmute' banner only once even when multiple requests to unmute were made in the same meeting. 

- reading the timeStamp that locus sends with `requestedToUnmute` call: `lastModifiedRequestToUnmuteTime` 
- storing `lastModifiedRequestToUnmuteTime` and comparing it with older request's timeStamp to correctly generate and fire the event `meeting:self:requestedToUnmute` as many time as requestedToUnmute=true.
- All the above changes result in UI correctly displaying the requested to unmute banner as many times as the request was made. 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

wrote unit tests in @webex/plugin-meetings/test/unit/locus-info

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
@@ -37,6 +37,7 @@ SelfUtils.parse = (self: any, deviceId: string) => {
       remoteMuted: SelfUtils.getRemoteMuted(self),
       unmuteAllowed: SelfUtils.getUnmuteAllowed(self),
       localAudioUnmuteRequested: SelfUtils.getLocalAudioUnmuteRequested(self),
+      localAudioUnmuteRequestedTimeStamp: SelfUtils.getLocalAudioUnmuteRequestedTimeStamp(self),
       localAudioUnmuteRequired: SelfUtils.getLocalAudioUnmuteRequired(self),
       lastModified: SelfUtils.getLastModified(self),
       modifiedBy: SelfUtils.getModifiedBy(self),
@@ -270,6 +271,10 @@ SelfUtils.getRemoteMuted = (self: any) => {
 
 SelfUtils.getLocalAudioUnmuteRequested = (self) => !!self?.controls?.audio?.requestedToUnmute;
 
+// requestedToUnmute timestamp
+SelfUtils.getLocalAudioUnmuteRequestedTimeStamp = (self) =>
+  Date.parse(self?.controls?.audio?.lastModifiedRequestedToUnmute) || 0;
+
 SelfUtils.getUnmuteAllowed = (self) => {
   if (!self || !self.controls || !self.controls.audio) {
     return null;
@@ -443,7 +448,10 @@ SelfUtils.localAudioUnmuteRequestedByServer = (oldSelf: any = {}, changedSelf: a
     );
   }
 
-  return changedSelf.localAudioUnmuteRequested && !oldSelf.localAudioUnmuteRequested;
+  return (
+    changedSelf.localAudioUnmuteRequested &&
+    changedSelf.localAudioUnmuteRequestedTimeStamp > oldSelf.localAudioUnmuteRequestedTimeStamp
+  );
 };
 
 SelfUtils.localAudioUnmuteRequiredByServer = (oldSelf: any = {}, changedSelf: any) => {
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
```diff
@@ -270,6 +271,9 @@ SelfUtils.getRemoteMuted = (self: any) => {
 
 SelfUtils.getLocalAudioUnmuteRequested = (self) => !!self?.controls?.audio?.requestedToUnmute;
 
+SelfUtils.getLocalAudioUnmuteRequestedTimeStamp = (self) =>
+  self?.controls?.audio?.lastModifiedRequestedToUnmute || '0';
```
robstax, line 275: why `'0'` vs `0`

packages/@webex/plugin-meetings/test/unit/spec/locus-info/selfConstant.js
```diff
@@ -116,6 +116,7 @@ export const self = {
     audio: {
       muted: false,
       requestedToUnmute: false,
+      lastModifiedRequestedToUnmute: '2023-06-16T18:25:04.369Z',
```
robstax, line 119: if the data we are receiving from the backend is ISO strings, i think we should consider converting them to a format that can be more easily compared.
```
Date.parse('2023-06-16T19:25:04.369Z')
```
outputs `1686943504369`

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
```diff
@@ -443,7 +447,10 @@ SelfUtils.localAudioUnmuteRequestedByServer = (oldSelf: any = {}, changedSelf: a
     );
   }
 
-  return changedSelf.localAudioUnmuteRequested && !oldSelf.localAudioUnmuteRequested;
+  return (
+    changedSelf.localAudioUnmuteRequested &&
+    changedSelf.localAudioUnmuteRequestedTimeStamp > oldSelf.localAudioUnmuteRequestedTimeStamp
```
robstax, line 452: see my below comment. we don't want to compare ISO strings this way. instead let's convert them to ms with `Date.parse`. that way the `<` will work as expected since it will be comparing numbers vs. numbers instead of strings vs. strings

packages/@webex/plugin-meetings/test/unit/spec/locus-info/selfUtils.js
```diff
@@ -251,6 +251,38 @@ describe('plugin-meetings', () => {
           assert.equal(updates.canNotViewTheParticipantListChanged, false);
         });
       });
+
+      describe('localAudioUnmuteRequestedByServer', () => {
+        it('should return localAudioUnmuteRequestedByServer = true when requestedToUnmute = true and lastModifiedRequestedToUnmute timestamps are different', () => {
```
robstax, line 256: i think we should not only be checking that they are different, but that the newer one is also `>` than the existing, so we should add test cases for that.

robstax (changes requested):
really good so far! main issue is that we should convert the ISO strings to ms so that we can compare them. once that's handled, i think this should be good to go

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
shivani1211: changed it now that I am using Date.Parser

packages/@webex/plugin-meetings/test/unit/spec/locus-info/selfUtils.js
```diff
@@ -251,6 +251,55 @@ describe('plugin-meetings', () => {
...
+
+          assert.equal(updates.localAudioUnmuteRequestedByServer, false);
+        });
+
+        it('should return localAudioUnmuteRequestedByServer = true when first request is made with requestedToUnmute = true', () => {
+          const clonedSelf = cloneDeep(self);
+
+          //request to unmute with timestamp
+          clonedSelf.controls.audio.requestedToUnmute = true;
+          clonedSelf.controls.audio.lastModifiedRequestedToUnmute = '2023-06-16T18:25:04.369Z';
```
robstax, line 269: since we're now storing these values as ints should the tests also be ints (ms since epoch) to better reflect the actual code?

robstax (approved):

Neeraj-swarnkar (approved):
Looks Good. SOme of the checks are failing please check once..

PR #2899, opened 2023-06-21
feat(plugin-meetings): added setActiveSpeakerCsis() API

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-436886

Added fixCsis to be called with an array so that we avoid sending too many requests to pin/unpin

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
+++ b/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
@@ -72,7 +72,7 @@ export class RemoteMediaGroup {
    *
    */
   public fixCsis(
-    remoteMediaCsis: {remoteMedia: RemoteMedia; csi: number | undefined}[],
+    remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number | undefined}[],
     commit = true
   ): void {
     forEach(remoteMediaCsis, ({csi, remoteMedia}) => {
--- a/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
+++ b/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
@@ -70,6 +70,7 @@ export class RemoteMediaGroup {
   /**
    * setActiveSpeakerCsis - sets csis for remoteMedia
    * pins when csis is there and unpins when csi is undefined
+   * @internal
    */
   public setActiveSpeakerCsis(
     remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number}[],
@@ -196,6 +197,10 @@ export class RemoteMediaGroup {
     throw new Error(`remote media object ${remoteMedia.id} not found in the group`);
   }
 
+  /**
+   * setPreferLiveVideo - sets preferLiveVideo to true/false
+   * @internal
+   */
   public setPreferLiveVideo(preferLiveVideo: boolean, commit: boolean) {
     if (this.options.preferLiveVideo !== preferLiveVideo) {
       this.options.preferLiveVideo = preferLiveVideo;
--- a/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
+++ b/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
@@ -68,8 +68,10 @@ export class RemoteMediaGroup {
   }
 
   /**
-   * setActiveSpeakerCsis - sets csis for remoteMedia
-   * pins when csis is there and unpins when csi is undefined
+   * Sets CSIs for multiple RemoteMedia instances belonging to this RemoteMediaGroup.
+   * For each entry in the remoteMediaCsis array:
+   * - if csi is specified, the RemoteMedia instance is pinned to that CSI
+   * - if csi is undefined, the RemoteMedia instance is unpinned
    * @internal
    */
   public setActiveSpeakerCsis(
--- a/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
+++ b/packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
@@ -68,26 +68,35 @@ export class RemoteMediaGroup {
   }
 
   /**
-   * fixCsis
+   * setCsis - sets csis for remoteMedia
    *
    */
-  public fixCsis(
-    remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number | undefined}[],
-    commit = true
-  ): void {
+  public setCsis(remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number}[], commit = true): void {
     forEach(remoteMediaCsis, ({csi, remoteMedia}) => {
       if (csi) {
         if (!(this.pinnedRemoteMedia.indexOf(remoteMedia) >= 0)) {
           const unpinId = this.unpinnedRemoteMedia.indexOf(remoteMedia);
-          this.unpinnedRemoteMedia.splice(unpinId, 1);
-          this.pinnedRemoteMedia.push(remoteMedia);
+          if (unpinId >= 0) {
+            this.unpinnedRemoteMedia.splice(unpinId, 1);
+            this.pinnedRemoteMedia.push(remoteMedia);
+          } else {
+            throw new Error(
+              `failed to pin a remote media object ${remoteMedia.id}, because it is not found in this remote media group`
+            );
+          }
         }
         remoteMedia.sendMediaRequest(csi, false);
       } else {
         if (!(this.unpinnedRemoteMedia.indexOf(remoteMedia) >= 0)) {
           const pinId = this.pinnedRemoteMedia.indexOf(remoteMedia);
-          this.pinnedRemoteMedia.splice(pinId, 1);
-          this.unpinnedRemoteMedia.push(remoteMedia);
+          if (pinId >= 0) {
+            this.pinnedRemoteMedia.splice(pinId, 1);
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,38 @@ export class RemoteMediaGroup {
     return [...this.unpinnedRemoteMedia, ...this.pinnedRemoteMedia];
   }
 
+  /**
+   * fixCsis
+   *
+   */
+  public fixCsis(
+    remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number | undefined}[],
```
marcin-bazyl, line 75: you don't need the `| undefined` bit in the csi type because you already have the `?`

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,38 @@ export class RemoteMediaGroup {
...
+    commit = true
+  ): void {
+    forEach(remoteMediaCsis, ({csi, remoteMedia}) => {
+      if (csi) {
+        if (!(this.pinnedRemoteMedia.indexOf(remoteMedia) >= 0)) {
+          const unpinId = this.unpinnedRemoteMedia.indexOf(remoteMedia);
```
marcin-bazyl, line 81: we need to handle a case where the remoteMedia is not in either pinned or unpinned list - basically it's not in the group

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,38 @@ export class RemoteMediaGroup {
...
+          this.unpinnedRemoteMedia.splice(unpinId, 1);
+          this.pinnedRemoteMedia.push(remoteMedia);
+        }
+        remoteMedia.sendMediaRequest(csi, false);
+      } else {
+        if (!(this.unpinnedRemoteMedia.indexOf(remoteMedia) >= 0)) {
```
marcin-bazyl, line 87: maybe it's just me, but I feel that having a condition `x === -1` or `x < 0` would be clearer than having a `!(x >= 0)`

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,38 @@ export class RemoteMediaGroup {
...
+          const pinId = this.pinnedRemoteMedia.indexOf(remoteMedia);
```
marcin-bazyl, line 88: same as previous comment - we need to take care of a case when remoteMedia is not in either pinned or unpinned list

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -505,6 +505,21 @@ export class RemoteMediaManager extends EventsScope {
     this.mediaRequestManagers.video.commit();
   }
 
+  /**
+   * fixCsis
+   */
+  public fixCsis(remoteMediaCsis: {remoteMedia: RemoteMedia; csi: number | undefined}[]) {
```
marcin-bazyl, line 511: we could have `csi?: number` instead of `csi: number | undefined`

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
```diff
@@ -144,6 +144,200 @@ describe('RemoteMediaGroup', () => {
...
+      assert.strictEqual(group.getRemoteMedia('unpinned').length, NUM_SLOTS);
+      assert.strictEqual(group.getRemoteMedia('pinned').length, 0);
+
+      const remoteMedia = group.getRemoteMedia('all')[PINNED_INDEX];
+
+      resetHistory();
+
+      const remoteMedia2 = group.getRemoteMedia('all')[PINNED_INDEX2];
+
+      const remoteMedisCsis = [{remoteMedia, csi: CSI}, {remoteMedia: remoteMedia2, csi: CSI2}, {remoteMedia}];
```
marcin-bazyl, line 298: hmm I don't know..... it feels a bit weird to be pinning and unpinning the same remoteMedia instance in a single call. Is this on purpose? Wouldn't it be better to have a 3rd remoteMedia instance in this test?

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,38 @@ export class RemoteMediaGroup {
     return [...this.unpinnedRemoteMedia, ...this.pinnedRemoteMedia];
   }
 
+  /**
+   * fixCsis
```
marcin-bazyl, line 71: Comments need to be a bit better as this is a public JS-SDK API that will be used potentially by other apps.
Also I'm not a big fan of the name "fixCsis" - the word "fix" can have multiple meanings (like fixing something that's broken). Maybe we can name it "setRemoteMediaCsis" or just "setCsis"

marcin-bazyl: actually we already have a `setRemoteVideoCsi()` API on RemoteMediaManager, but that one allows only calling it on "memberPanes" - so on the RemoteMedia that are by design "recever selected" on the current layout. Our new API is supposed to be called only on RemoteMedia objects that are part of an active speaker group, maybe this needs to be reflected in the API name somehow?

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
shnaaz: Was intentional, wanted to pin 2 remoteMedia and unpin one of them to check that one is still pinned. I can change this to pin 2 remoteMedias in one call and unpin in another call.

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
marcin-bazyl: a client is free to call the RemoteMediaGroup directly, so the check needs to be there too.

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
```diff
@@ -332,10 +341,26 @@ describe('RemoteMediaGroup', () => {
 
       const remoteMedisCsis = [{remoteMedia, csi: CSI}, {remoteMedia: remoteMedia2, csi: CSI2}, {remoteMedia}];
 
-      group.fixCsis(remoteMedisCsis, true);
+      group.setCsis(remoteMedisCsis, true);
 
       assert.calledOnce(fakeMediaRequestManager.commit);
     });
+
+    it.only('throws when remoteMedia id is not in unpinned and pinned array - csi is there', () => {
```
marcin-bazyl, line 349: `.only` in this test and the next one

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
marcin-bazyl: thanks, yeah that better reflects real usage of the API

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,50 @@ export class RemoteMediaGroup {
...
 
+  /**
+   * setActiveSpeakerCsis - sets csis for remoteMedia
+   * pins when csis is there and unpins when csi is undefined
+   */
+  public setActiveSpeakerCsis(
+    remoteMediaCsis: {remoteMedia: RemoteMedia; csi?: number}[],
+    commit = true
+  ): void {
+    forEach(remoteMediaCsis, ({csi, remoteMedia}) => {
```
marcin-bazyl, line 78: any reason why we're using `forEach` from lodash instead of javascript's Array.forEach() - so just `remoteMediaCsis.forEach()`?

shnaaz: Reads better for me

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
shnaaz: For me this reads better as I read it as if it is not in unpinned/pinned array

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -66,6 +67,50 @@ export class RemoteMediaGroup {
     return [...this.unpinnedRemoteMedia, ...this.pinnedRemoteMedia];
   }
 
+  /**
+   * setActiveSpeakerCsis - sets csis for remoteMedia
```
marcin-bazyl, line 71: Please update this to:
```
/**
   * Sets CSIs for multiple RemoteMedia instances belonging to this RemoteMediaGroup. 
   * For each entry in the remoteMediaCsis array:
   * - if csi is specified, the RemoteMedia instance is pinned to that CSI
   * - if csi is undefined, the RemoteMedia instance is unpinned
   *
   */
```

and the same for the comments in RemoteMediaManager

marcin-bazyl (approved):

PR #2902, opened 2023-06-22
chore(webpack): separate library mentions for each entry point

mkesavan13:
# No Spark - Adhoc PR

the Calling class in the calling SDK `packages/webex/calling.js` not importing correctly for the umd builds.

in the webpack config to indicate which entry point will have what library name

### Change Type

- [ ] Documentation update
- [x] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/webex/src/calling.js
+++ b/packages/webex/src/calling.js
@@ -47,4 +47,4 @@ class Calling {
   }
 }
 
-export {Calling};
+export default Calling;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -21,17 +21,35 @@ dotenv.config({path: '.env.default'});
  */
 module.exports = (env = {NODE_ENV: process.env.NODE_ENV || 'production'}) => ({
   entry: {
-    webex:
-      env && env.NODE_ENV === 'development'
-        ? `${path.resolve(__dirname)}/packages/webex/src/index.js`
-        : './packages/webex',
-    meetings: `${path.resolve(__dirname)}/packages/webex/src/meetings.js`,
-    calling: `${path.resolve(__dirname)}/packages/webex/src/calling.js`,
+    webex: {
+      import:
+        env && env.NODE_ENV === 'development'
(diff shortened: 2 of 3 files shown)


arun3528 (approved):

webpack.config.js
```diff
@@ -21,17 +21,35 @@ dotenv.config({path: '.env.default'});
...
+        name: 'Webex',
+        type: 'umd',
+      },
+    },
+    calling: {
+      import: `${path.resolve(__dirname)}/packages/webex/src/calling.js`,
+      library: {
+        name: 'Calling',
+        type: 'umd',
+        export: 'default',
```
Kesari3008, line 46: why do we need to add export here specifically for calling only

packages/webex/src/calling.js
```diff
@@ -47,4 +47,4 @@ class Calling {
   }
 }
 
-export {Calling};
+export default Calling;
```
Kesari3008, line 50: is this also mandatory ?

Kesari3008 (approved):

webpack.config.js
mkesavan13: So, since the calling.js is es6 module, we cannot do `export Calling` - this will throw syntax error.

What we could do is `export default Calling` which equals to `module.exports.default = Calling`

In this scenario, if we do not add this config, webpack will export the Calling class in the global object inside the library `Calling`. So you would have to access it like `new Calling.Calling()`

Whereas, when you add this, whatever is exported as default will be assigned to the library `Calling` itself.

- Read more about webpack library config here -> [webpack:output.library](https://webpack.js.org/configuration/output/#outputlibrary)
- Read more about the webpack library export config here -> [webpack:output.library.export](https://webpack.js.org/configuration/output/#outputlibraryexport)

packages/webex/src/calling.js
mkesavan13: I think [this comment](https://github.com/webex/webex-js-sdk/pull/2902#discussion_r1239333384) will clarify this question too

PR #2906, opened 2023-06-26
refactor(voicea): export announce and status

sundayu:
# COMPLETES # [SPARK-439057](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-439057) [SPARK-440574](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-440574)

1. export announce API, when closed caption is active, user no more needs to request turn on caption, just announce is enough.
2. export announce status
3. export caption enable status

1. add announce api. 
2. add announce status.
3. add caption enable status.
4. llm isConnected now use connected property instead of registered.

### Change Type

- [x] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -30,6 +38,12 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
 
   private vmcDeviceId?: string;
 
+  private announceStatus: string;
+
+  private captionStatus: string;
+
+  private turnOnCaptionsWaitingOnlinePromise: Promise<void> | null;
```
Coread, line 45: question: What does turnOnCaptionsWaitingOnlinePromise do? I can't see it being used anywhere

sundayu: Removed.
Thanks!

Coread (approved):

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -327,11 +345,21 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
...
 
   /**
-   * Turn on Captions
+   * Binding llm online event
+   * @param {function} callback
+   * @returns {void}
+   */
+  private onceLLMOnline = (callback) => {
+    // @ts-ignore
+    this.webex.internal.llm.once(LLM_EVENTS.ONLINE, callback);
```
arun3528, line 354: can llm go online and offline mid meeting ?

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -341,20 +369,107 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
...
+
+  /**
+   * bind llm online event once to turn on captions
+   * @returns {void}
+   */
+  private turnOnCaptionsAfterLLMOnline = (): Promise<void> => {
+    this.captionStatus = TURN_ON_CAPTION_STATUS.WAITING_LLM_ONLINE;
+
+    return new Promise((resolve) => {
+      this.onceLLMOnline(() => {
```
arun3528, line 438: why dont we move this logic inside the onceLLMonline -> 

```
        if (this.captionStatus === TURN_ON_CAPTION_STATUS.WAITING_LLM_ONLINE) {
          this.captionStatus = TURN_ON_CAPTION_STATUS.IDLE;
          resolve(this.turnOnCaptions());
        }
        ```

arun3528 (commented):
i feel that the onceLLMOnline is in the wrong place .

The registration to the llm needs to happen ahead of time and have to maintain the state 

you might miss out on events , it should happen on register

packages/@webex/internal-plugin-voicea/src/voicea.ts
sundayu: The problem to be solved here is that when the user clicks to turn on the captions, if the connection of llm is not yet completed, it will fail to turn on. We have added a failure message on the UI, but for guest users, the failure rate of turning on captions for the first time is almost 100%.
So when the user calls the turn-on api, if the llm is not yet connected, it will finish sending the turn-on message when the llm is online. And this message only needs to be sent once, and if the llm reconnects later, it does not need to send the open message again.

sundayu:
> i feel that the onceLLMOnline is in the wrong place .
> 
> The registration to the llm needs to happen ahead of time and have to maintain the state
> 
> you might miss out on events , it should happen on register

onceLLMOnline is a workaround.
Because all connections in the meeting are asynchronous, the complete presentation of the meeting UI is not required to wait for llm to be connected.
Yes, some events will be missed, but nothing will be missed in terms of the voicea function in the meeting. because captions is a dynamically enabled feature that can be turned on at any time during the meeting. There is no events until it is turned on.

packages/@webex/internal-plugin-voicea/src/voicea.ts
sundayu: onceLLMOnline is a utility method that is also used in announce

sundayu:
> i feel that the onceLLMOnline is in the wrong place .
> 
> The registration to the llm needs to happen ahead of time and have to maintain the state
> 
> you might miss out on events , it should happen on register

Agree, I have removed onceLLMOnline and decided to take the long view.

mickelr (approved):
LGTM

PR #2914, opened 2023-06-30
feat: media ready event

shnaaz:
[SPARK-440553](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-440553)

Added icePhase to client.ice.end metrics event.
Sending media-engine.ready event at the end of a successful Meeting.addMedia() 

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4930,7 +4930,7 @@ export default class Meeting extends StatelessWebexPlugin {
         this.reconnect({networkDisconnect: true});
         NewMetrics.submitClientEvent({
           name: 'client.ice.end',
-          payload: {canProceed: false},
+          payload: {canProceed: false, icePhase: 'JOIN_MEETING_RETRY'},
```
marcin-bazyl, line 4933: this code is executed when we lose ICE connection in-meeting, after we initially established it successfully during join, so the icePhase should be 'IN_MEETING'

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4966,6 +4966,7 @@ export default class Meeting extends StatelessWebexPlugin {
         case ConnectionState.Connected:
           NewMetrics.submitClientEvent({
             name: 'client.ice.end',
+            payload: {icePhase: 'IN_MEETING'},
```
marcin-bazyl, line 4969: this code gets executed both during join and in-meeting, it's not going to be straightforward to determine which case is it (we'll probably have to add some flag to the meeting object to know when we connect successfully for the first time), so maybe we can just skip sending icePhase here? @torgeadelin what do you think? does CA team care about this? 
if we want do set it properly, it needs to be 'JOIN_MEETING_FINAL' if this is the first time we are "connected" for the current meeting join attempt and 'IN_MEETING' for any other case

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5425,8 +5439,16 @@ export default class Meeting extends StatelessWebexPlugin {
           connectionType,
           isMultistream: this.isMultistream,
         });
+        // send media-engine.ready event at the end of a successful Meeting.addMedia()
```
marcin-bazyl, line 5442: yup this is good

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5370,6 +5371,8 @@ export default class Meeting extends StatelessWebexPlugin {
         }
       })
       .catch((error) => {
+        // Do we have to send any event
+        // send media-engine.ready with an error - UCF uses errorCode 2004 ?
```
marcin-bazyl, line 5375: probably better to just send it in the final catch at the end instead of here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5425,8 +5439,16 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        // send media-engine.ready event at the end of a successful Meeting.addMedia()
+        NewMetrics.submitClientEvent({
+          name: 'media-engine.ready',
+          options: {
+            meetingId: this.id,
+          },
+        });
       })
       .catch((error) => {
+        // send media-engine.ready with an error - UCF uses errorCode 2004
```
marcin-bazyl, line 5451: yes, but not sure if we should send it in the case if we've already sent the 'client.ice.end' from line 5409

shnaaz: hmm, I think we should not as we are sending client.ice.end

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: Not sure if you mean skipping sending ice.end entirely? I know that we should always send ice.start and ice.end.

packages/@webex/plugin-meetings/src/meeting/index.ts
torgeadelin: yeah, i think it's fine if we don't

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: I think Marcin mean not sending icePhase in this particular case

torgeadelin: what does this mean tho? was there a client.ice.end sent previously if we reach this case?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5425,8 +5436,16 @@ export default class Meeting extends StatelessWebexPlugin {
           connectionType,
           isMultistream: this.isMultistream,
         });
+        // send media-engine.ready event at the end of a successful Meeting.addMedia()
```
marcin-bazyl, line 5439: nitpick: this comment is redundant, it doesn't add any value

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5425,8 +5436,16 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        NewMetrics.submitClientEvent({
+          name: 'media-engine.ready',
+          options: {
+            meetingId: this.id,
+          },
+        });
       })
       .catch((error) => {
+        // remove send media-engine.ready with an error - UCF uses errorCode 2004
```
marcin-bazyl, line 5448: I don't understand this comment, I guess it's not needed?

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: > was there a client.ice.end sent previously if we reach this case?

no, not necessarily. This same piece of code is executed when we successfully connect the first time or at any later point after a brief network loss

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5403,6 +5403,17 @@ export default class Meeting extends StatelessWebexPlugin {
       )
       .then(() =>
         this.mediaProperties.waitForMediaConnectionConnected().catch(() => {
+          NewMetrics.submitClientEvent({
```
torgeadelin, line 5406: this now needs to be `this.webex.internal.newMetrics.submitClientEvent` instead

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5425,6 +5436,12 @@ export default class Meeting extends StatelessWebexPlugin {
           connectionType,
           isMultistream: this.isMultistream,
         });
+        NewMetrics.submitClientEvent({
```
torgeadelin, line 5439: this now needs to be `this.webex.internal.newMetrics.submitClientEvent` instead

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1216,6 +1216,22 @@ describe('plugin-meetings', () => {
               errorThrown = true;
             });
 
+          assert.calledTwice(NewMetrics.submitClientEvent);
+
+          assert.calledWithMatch(NewMetrics.submitClientEvent, {
```
torgeadelin, line 1221: same for the tests

torgeadelin (changes requested):

torgeadelin (approved):

PR #2919, opened 2023-07-04
fix(breakout): skip moved state main locus

othellashao:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-442495

to fix issue that sometimes can't get the right roster list when join into breakout session. The root cause is that sometimes will receive the moved main locus (the sequence is bigger one) first, then receive the breakout locus (the sequence is smaller) later, will drop the breakout locus by sequence comparing. Should skip the main locus that resource is moved on this device (follow native client's new logic on it)

add resource is moved on this device in `isNeedHandleMainLocus`, if resource is moved which mean already join into bo, drop this main locus 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -261,6 +261,9 @@ export default class Meetings extends WebexPlugin {
     const isSelfMoved = newLocus?.self?.state === _LEFT_ && newLocus?.self?.reason === _MOVED_;
     // @ts-ignore
     const deviceFromNewLocus = MeetingsUtil.getThisDevice(newLocus, this.webex.internal.device.url);
+    const isResourceMovedOnThisDevice =
+      deviceFromNewLocus?.state === _LEFT_ && deviceFromNewLocus?.reason === _MOVED_;
+
     const isNewLocusJoinThisDevice = MeetingsUtil.joinedOnThisDevice(
       meeting,
       newLocus,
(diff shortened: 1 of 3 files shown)


mickelr (approved):
LGTM

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -261,6 +261,9 @@ export default class Meetings extends WebexPlugin {
     const isSelfMoved = newLocus?.self?.state === _LEFT_ && newLocus?.self?.reason === _MOVED_;
     // @ts-ignore
     const deviceFromNewLocus = MeetingsUtil.getThisDevice(newLocus, this.webex.internal.device.url);
+    const isResourceMovedOnThisDevice =
```
arun3528, line 264: Next time let's try handling the locus chang logic in the locus info file . The web client and the sdk team handle all the logic there .

I can have a session to explain how it works 

Meetings file should only handle meeting collection and routing events to the actual meeting object which then does further processing

othellashao: @arun3528 Yes, But for this case we need to check `isNeedHandleLocus` at the beginning of receiving locus DTO to cover the meeting object haven't create case. For example: user use native client start a meeting and join into bo, then user open web client with same account, web client may receive locusDTOs of Main and Breakout at same time, if not check the `isNeedHandleLocus` in meetings/index, may will create two meeting objects for these DTOs, will caused wrong behavior and issues. So I need to put the logic at the beginning position.

arun3528: got it thanks or the explanation

arun3528 (approved):

PR #2920, opened 2023-07-04
test: improve integration test stability

marcin-bazyl:

Integration tests in meetings plugin often fail at the start - just failing to create a space, so after that everything else will fail too, because all other tests depend on having that space.

This is because the test org has "basicMessage" entitlement, but when we create the test users we don't ask for that entitlement, so the entitlement is assigned to the users automatically from the org, but that takes some time - it requires multiple backend services to talk to each other, so if the test tries to create a space before that entitlement is given to the user, then it will fail - a classic race condition.

I'm fixing this by explicitly asking for the "basicMessage" entitlement when creating the test users. This way they have it from the start and we don't need to worry about waiting for the backend components to talk to each other and assign it in time before we try to create the space.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Running tests in the pipeline - I've run them 8 times at different times of the day, they all passed

### I certified that

---

### DIFF
--- a/packages/@webex/test-users/src/index.js
+++ b/packages/@webex/test-users/src/index.js
@@ -190,6 +190,7 @@ export function createTestUser(options = {}) {
       'squaredRoomModeration',
       'squaredInviter',
       'webExSquared',
+      'basicMessage',
     ],
     idbrokerUrl,
     machineType: options.machineType,


torgeadelin (approved):

sundayu (approved):

arun3528 (approved):
thanks @marcin-bazyl

PR #2928, opened 2023-07-06
fix: add a timeout option to dss requests

maxinteger:
# COMPLETES #SPARK-398014

Add a timeout option to DSS requests

Previously, DSS requests never failed
when mercury service had not responded.
This can lead to bad user experience and issues
with restoring connection after a network problem.

- Added a new restartable Timer
- Added DSS timeout error
- Added timeout option to all DSS API
- Used the timer to fail the request when the timeout expired.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-dss/src/config.ts
```diff
@@ -4,6 +4,12 @@
 
 export default {
   dss: {
+    /**
+     * Timeout before DDS request fails.
```
lismith2-cisco, line 8: typo DDS, and please add unit

packages/@webex/internal-plugin-dss/src/config.ts
```diff
@@ -4,6 +4,12 @@
...
+     * @type {Number}
+     */
+    requestTimeout: 6000,
```
lismith2-cisco, line 11: Did you ask DSS team about this?

packages/@webex/internal-plugin-dss/src/dss.ts
```diff
@@ -163,17 +166,23 @@ const DSS = WebexPlugin.extend({
...
     const result = {};
-    let expectedSeqNums;
-    let notFoundArray;
+    let expectedSeqNums: string[];
+    let notFoundArray: unknown[];
+
+    return new Promise((resolve, reject) => {
+      const timer = new Timer(() => {
+        this.stopListening(this, eventName);
+        reject(new DssTimeoutError({requestId, timeout, resource, params}));
```
lismith2-cisco, line 181: the timeout in the error should also default to the config one, just tried it and got "webex-web-client Module:DSS Error:The KMS did not respond within undefined ms."

packages/@webex/internal-plugin-dss/src/config.ts
maxinteger: @lismith2-cisco I got it from the KMS plugin

lismith2-cisco: I've added you to DSS comms space and asked Willie

packages/@webex/internal-plugin-dss/src/dss.ts
```diff
@@ -355,7 +370,7 @@ const DSS = WebexPlugin.extend({
    * @returns {Promise} Resolves with an array of entities found
```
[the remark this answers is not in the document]
maxinteger: timeout applied for all single and batch requests. I updated the JSDocs

packages/@webex/internal-plugin-dss/src/config.ts
maxinteger: Looks like maximum request time is around 100ms. I set the default timeout to `1000` it give us enough room for network slowness and it should not be noticeable for the user. Let me know if you disagree.

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -239,6 +246,61 @@ describe('plugin-dss', () => {
...
+
+      it('fails with default timeout when mercury does not respond and the timeout option is not set', async () => {
+        const {promise} = await testMakeRequest({
+          method: 'lookupDetail',
+          resource: '/lookup/orgid/userOrgId/identity/test id/detail',
+          params: {id: 'test id'},
+          bodyParams: {},
+        });
+
+        clock.tick(webex.config.dss.requestTimeout);
```
lismith2-cisco, line 258: I'd prefer explicit time (1000) rather than assuming the config is set correctly.

lismith2-cisco, line 258: Did this ought to use tickAsync? Mind you if it works, it's probably fine.

packages/@webex/internal-plugin-dss/src/config.ts
```diff
@@ -4,6 +4,12 @@
 
 export default {
   dss: {
+    /**
+     * Timeout before DSS request fails, in milliseconds.
+     * @type {Number}
+     */
+    requestTimeout: 1000,
```
marcin-bazyl, line 11: I don't know much about DSS, maybe it's very quick, but 1s feels like a very short timeout from my experience with Webex backend components

packages/@webex/internal-plugin-dss/src/dss.ts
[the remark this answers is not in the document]
marcin-bazyl: I guess we have 3 options:
1. don't apply timeout for batched requests at all
2. apply it to the first item in the batcher
3. apply it to the last item in the batcher
I don't mind which one we go for, current implementation is like point 2 and I guess it's fine as long as this is explained in the jsdocs

packages/@webex/internal-plugin-dss/src/config.ts
maxinteger: @marcin-bazyl based on the metrics the average DSS response time is below 100ms: https://metrics-noram.wbx2.com/goto/5vwQxSCVz I thought 10 times more would be sufficient for timeout

packages/@webex/common-timers/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.startTimer();
+    this.state = 'running';
+  }
+
+  /**
+   * Clear the current timer and start a new one
+   * @returns {void}
+   */
+  reset() {
+    if (this.state !== 'running') {
```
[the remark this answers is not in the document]
maxinteger: My idea was, when the timer is done then it is done. It feels wrong to reset a finished timer, for example you can not change promise state either to `pending` after it fulfilled.

packages/@webex/internal-plugin-dss/src/dss.ts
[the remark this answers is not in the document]
maxinteger: @marcin-bazyl correct me if I am wrong, but the point with batch request is collect all the requests in a specific timeframe and do only one call to the server. Since the the server call falls back to the `_request` function from `dss.ts` the whole batch will fail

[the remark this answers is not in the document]
lismith2-cisco: @maxinteger the point is that each request in a batch is made (potentially) with a different timeout, but only one timeout is applied to the batch, so which one?

packages/@webex/internal-plugin-dss/src/config.ts
lismith2-cisco: Doesn't include network time at any rate, so need to add on maximum expected network delay (which we presumably have no idea), maybe go back to your first number of 6000, since it isn't causing any problems we know of for kms (I know I suggested changing it in the first place)? I really don't know though :/ but would hate to cause this to start failing for no good reason

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -1252,6 +1259,78 @@ describe('plugin-dss', () => {
         expect(Batcher.prototype.request.getCall(1).args).to.deep.equal(['id2']);
         expect(result).to.equal(response2);
       });
+
+      it('fails when mercury does not response but only the affected request, not all if them ', async () => {
```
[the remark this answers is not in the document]
lismith2-cisco: Had a look, I'll comment here instead. This test is just checking this particular function, so it may not need 4 requests or to wait 1000ms. It does need to check the timeout is passed through to the batcher correctly. You also need to add higher-level unit tests to '#lookup' to check that the timeout actually works properly (with testMakeBatchedRequests)

packages/@webex/internal-plugin-dss/src/dss.ts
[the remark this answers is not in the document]
maxinteger: @lismith2-cisco We store the timeout in the batcher instance, which is created for each resources only once. So the first call on `_batchedLookup` will determine the timeout for all subsequent request on the same resource.

If we drop the batcher instance after the queued items finished then we can set a new timeout for the next batch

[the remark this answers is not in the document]
lismith2-cisco: Sorry, I was incorrect, yes it's per batcher not batch. I don't think dropping the batcher after each batch would be better, and I think it would be extremely hard to do. To apply timeouts properly, create a new batcher based on resource/timeout combination, instead of just resource. Or don't make timeouts configurable for batched requests, and document it. Or don't make timeouts configurable at all (which I would vote for tbh), the global config can be set by the app, and I don't see a use case for each request choosing a timeout.

packages/@webex/internal-plugin-dss/src/config.ts
maxinteger: true, the network time was not part of it. KMS use 6000ms as well so it could be a good staring point. We can adjust it later when necessary.

packages/@webex/internal-plugin-dss/src/dss.ts
```diff
@@ -154,26 +157,35 @@ const DSS = WebexPlugin.extend({
...
-   * @param {string} options.foundPath the path to get the lookups of the found data (optional)
-   * @param {string} options.notFoundPath the path to get the lookups of the not found data (optional)
+   * @param {string} [options.foundPath] the path to get the lookups of the found data
+   * @param {string} [options.notFoundPath] the path to get the lookups of the not found data
+   * @param {number} [options.timeout] connection timeout in milliseconds
    * @returns {Promise<Object>} result Resolves with an object
    * @returns {Array} result.resultArray an array of entities found
    * @returns {Array} result.foundArray an array of the lookups of the found entities (if foundPath provided)
    * @returns {Array} result.notFoundArray an array of the lookups of the not found entities (if notFoundPath provided)
+   * @throws {DssTimeoutError} when server does not response in the specified timeframe
```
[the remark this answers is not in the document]
marcin-bazyl: there is still 1 instace left in the test title: 'fails when mercury does not response but only the affected request'

packages/@webex/internal-plugin-dss/src/dss.ts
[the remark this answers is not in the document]
maxinteger: @lismith2-cisco I agree, I dropped the timeout parameter and it always use the default timeout

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -921,6 +1157,79 @@ describe('plugin-dss', () => {
...
+          ],
+        });
+
+        // only 1 response
+        mercuryCallbacks['event:directory.lookup'](
+          createData('id 1', 0, true, 'lookupResult', {entitiesNotFound: ['id1', 'id13']})
+        );
+
+        // Timeout
+        await clock.tickAsync(6000);
```
lismith2-cisco, line 1196: Not sure exactly what you're trying to test, but if you do this waiting after the second batch, the first batch will still be in the batcher

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -921,6 +1157,79 @@ describe('plugin-dss', () => {
...
+
+        // Next batch
+        uuidStub.onCall(requests.length).returns('req-id-2');
```
lismith2-cisco, line 1199: why not use testMakeBatchedRequests again?

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -921,6 +1157,79 @@ describe('plugin-dss', () => {
...
+            },
+            {
+              method: 'lookup',
+              params: {id: 'id3', shouldBatch: true},
+            },
+          ],
+        });
+
+        // only 1 response
+        mercuryCallbacks['event:directory.lookup'](
```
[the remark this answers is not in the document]
maxinteger: I wanted to make it more realistic. 

This proves timeout is still ticking even when a request finished.

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
maxinteger: I try to test 2 things 
1. batch fails with timeout
2. batch after the timed out batch finish without any problem.

What do you mean by "the first batch will still be in the batcher"? At the end of the test I am checking the result of the requests and all the request from the first brach are rejected

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
maxinteger: fixed

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
lismith2-cisco: Ok, the description sounded like you wanted to test having two batches in progress at the same time, one of which failed. Maybe change the description to be more clear, like 'fails when mercury does not respond, later batches can still pass ok'. 
You could check that the failure doesn't afffect any other in-progress batches if you wanted, by setting off batch 1, then batch 2, then waiting for batch 1 to timeout, then sending mercury message for batch 2, and chacking only batch 1 promises are rejected. Probably not necessary though.

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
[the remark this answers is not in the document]
lismith2-cisco: This mercury response doesn't finish any requests. If you meant to do 
createData('req-id-1', 0, true, 'lookupResult', {entitiesNotFound: ['id1', 'id3']})
instead of 
createData('id 1', 0, true, 'lookupResult', {entitiesNotFound: ['id1', 'id13']})
that would be an invalid mercury message.
if you meant to do 
createData('req-id-1', 0, false, 'lookupResult', {entitiesFound: ['id1', 'id3'], entities:['data1', 'data3']})
that would be a valid mercury message, though it wouldn't get back to the batcher, it's handled in the dss request code.

packages/@webex/internal-plugin-dss/test/unit/spec/dss.ts
```diff
@@ -921,6 +1156,95 @@ describe('plugin-dss', () => {
...
+          ],
+          calls: [
+            {
+              method: 'lookup',
+              params: {id: 'id4', shouldBatch: true},
+            },
+          ],
+        });
+
+        // Batch 1 - only 1 response out of 3
```
lismith2-cisco, line 1207: "// Batch 1 - only 1 mercury response out of 2 received"
(I would say 2 responses out of 3, but it doesn't really count as a response until we have the full sequence)

lismith2-cisco (approved):

shnaaz (approved):

arun3528 (approved):

PR #2932, opened 2023-07-10
feat(ca-metrics): fix meeting info req latency

torgeadelin:

- fix meeting info request latency (it was broken because both the req and res were sent one after the other in the join method -- which is what we still need, but we need the correct latency)
- because of the way we're sending the meeting info events (req and res), we need additional events to measure the latency of when these events are supposed to happen. In order to calculate it, i added 2 internal events that replicate the correct behaviour of the meeting info events. 
- the reason we can't send the real CA events inside the actual meeting info request code, is that meeting info is used in a lot of other places not just the join phase of the meeting (which is when we need to send it for CA exclusively)
### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

-unit test, manual tesst

### I certified that

---


packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -336,16 +336,29 @@ export default class MeetingInfoV2 {
...
+    if (meetingId) {
+      this.webex.internal.newMetrics.submitInternalEvent({
+        name: 'internal.client.meetinginfo.request',
+      });
+    }
+
     return this.webex
       .request(requestOptions)
       .then((response) => {
+        if (meetingId) {
```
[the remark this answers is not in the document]
torgeadelin: hm, i think it's important we cover all cases.
Could you point me to where in the code this scenario happens?

arun3528 (approved):

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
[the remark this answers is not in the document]
torgeadelin: oh that's great. Yeah I think we discussed about this when I implemented the meeting info events CA a while back, and I remember we made sure it covers all cases.

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -336,16 +336,29 @@ export default class MeetingInfoV2 {
       requestOptions.resource = 'meetingInfo';
     }
 
+    if (meetingId) {
```
marcin-bazyl, line 339: what about non-USM meetings (v1)? We still have customers using them, they will have getMeetingInfoReqResp() returning undefined

marcin-bazyl (approved):
looks good, my only concern is about v1 meetings, but this can be addressed in another PR

packages/@webex/plugin-meetings/src/meeting-info/index.ts
```diff
@@ -113,21 +141,39 @@ export default class MeetingInfo {
...
+  ) {
     if (type === _PERSONAL_ROOM_ && !destination) {
       destination = this.webex.internal.device.userId;
     }
 
     return this.fetchInfoOptions(MeetingInfoUtil.extractDestination(destination, type), type).then(
-      (options) =>
+      (infoOptions) =>
         // fetch meeting info
         this.requestFetchInfo(options).catch((error) => {
```
marcin-bazyl, line 169: before this change the call to `this.requestFetchInfo()` was passing the options that were the result of the call to `this.fetchInfoOptions()`, now we're passing the options which are just the `meetingId` - this doesn't seem right, we should pass `{...infoOptions, meetingId: options.meetingId}`, right?

BTW, I would expect that there is some existing unit or integration test that should catch this. Was it caught by any test?

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting-info/index.ts
torgeadelin: oh, i missed that there are 2 calls to this.requestFetchInfo, will add it to the first call as well.
I don't think it was caught, I have a green build..

torgeadelin: but there were no tests for meeting info v1, i had to create a new file, but only tested the new changes i introduced

torgeadelin: fixed it and added a test

packages/@webex/plugin-meetings/test/unit/spec/meeting-info/index.js
```diff
@@ -7,6 +7,8 @@ import MeetingInfo from '@webex/plugin-meetings/src/meeting-info/index';
 import MeetingInfoUtil from '@webex/plugin-meetings/src/meeting-info/util';
 import MeetingInfoRequest from '@webex/plugin-meetings/src/meeting-info/request';
 
+const flushPromises = () => new Promise(setImmediate);
```
marcin-bazyl, line 10: there is already flushPromises in packages/@webex/plugin-meetings/test/utils/testUtils.js

packages/@webex/plugin-meetings/test/unit/spec/meeting-info/index.js
```diff
@@ -88,7 +90,57 @@ describe('plugin-meetings', () => {
         } catch (err) {
           const submitInternalEventCalls = webex.internal.newMetrics.submitInternalEvent.getCalls();
 
-          assert(submitInternalEventCalls[0].args, {
+          assert.deepEqual(submitInternalEventCalls[0].args[0], {
```
marcin-bazyl, line 93: we should add an `assert.fail()` call at the end of the try clause

marcin-bazyl (approved):

PR #2938, opened 2023-07-11
feat(plugin-meetings): add screenshare audio

prsaroha:
# COMPLETES #SPARK-337410
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-337410

Adds handling for screen share audio in multistream connections only.

- add tracks for screenshare audio for local and remote share
- add separate handling for screenshare audio and video to allow either to be shared without the other

More details on this can be found here: https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/SPARK-337409%3A+Share+System+Audio
Corresponding web app PR: Coming soon!

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3482,7 +3489,37 @@ export default class Meeting extends StatelessWebexPlugin {
...
       await this.unpublishTrack(oldTrack);
     }
-    await this.publishTrack(this.mediaProperties.shareTrack);
+    await this.publishTrack(this.mediaProperties.shareVideoTrack);
+  }
+
+  /**
+   * Stores the reference to a new screen share audio track, sets up the required event listeners
+   * on it, cleans up previous track, etc.
+   * It also sends the floor grant/release request.
```
marcin-bazyl, line 3498: this part of the comment is not true, the function doesn't send the floor grant/release request. I realize you copied this from the existing setLocalShareVideoTrack() comments - they're incorrect too, can you fix them as well, please?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3482,7 +3489,37 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+    this.mediaProperties.setLocalShareAudioTrack(localSystemAudioTrack);
+
+    localSystemAudioTrack?.on(LocalTrackEvents.Ended, this.handleShareAudioTrackEnded);
+    localSystemAudioTrack?.on(
+      LocalTrackEvents.UnderlyingTrackChange,
+      this.underlyingLocalTrackChangeHandler
+    );
+
+    this.mediaProperties.mediaDirection.sendShare = !!localSystemAudioTrack;
```
marcin-bazyl, line 3517: this should be `!!this.mediaProperties.shareVideoTrack || !!this.mediaProperties.shareAudioTrack` and same in setLocalShareVideoTrack()

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6117,37 +6176,65 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  /**
+   * Functionality for when a share video is ended.
+   * @private
+   * @memberof Meeting
+   * @param {MediaStream} localShare
+   * @returns {undefined}
+   */
+  private handleShareVideoTrackEnded = async () => {
+    if (this.wirelessShare) {
+      this.leave({reason: MEETING_REMOVED_REASON.USER_ENDED_SHARE_STREAMS});
```
marcin-bazyl, line 6210: This is incorrect. When a share with audio is stopped via browser UI, we will get handleShareVideoTrackEnded() and handleShareAudioTrackEnded() called one after another, so we will call this.leave() on the first one and we should be calling it on the 2nd call (only if both audio and video share tracks ended)

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6756,7 +6843,13 @@ export default class Meeting extends StatelessWebexPlugin {
     let floorRequestNeeded = false;
 
     if (tracks.screenShare?.video) {
-      await this.setLocalShareTrack(tracks.screenShare?.video);
+      await this.setLocalShareVideoTrack(tracks.screenShare?.video);
+
+      floorRequestNeeded = true;
```
marcin-bazyl, line 6848: this will result in SDK sending a floor request twice if someone calls publishTracks() first with just a video share track and then again with just audio share track

packages/@webex/plugin-meetings/README.md
```diff
@@ -1023,8 +1026,11 @@ meeting.on('media:stopped', (media) => {
   if (media.type === 'remoteAudio') {
     document.getElementById('remoteaudio').srcObject = null;
   }
-  if (media.type === 'localShare') {
-    document.getElementById('localshare').srcObject = null;
+  if (media.type === 'localShareVideo') {
```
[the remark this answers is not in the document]
prsaroha: removed

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: oops, fixed both

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: fixed this

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: moved this to triggerStoppedSharing() which I'm calling from both handlers, it checks for both share tracks to have ended

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: changed this to !this.isJoined
two publishTracks() calls in very quick succession could still result in two calls for the request depending on how long it takes for share to start, but adding more proper state management for floorGrantPending or a new variable is extra work that I haven't scoped yet

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6202,27 +6197,31 @@ export default class Meeting extends StatelessWebexPlugin {
...
 
+  /**
+   * Emits meeting:stoppedSharingLocal
+   * @private
+   * @returns {undefined}
+   * @memberof Meeting
+   */
   private triggerStoppedSharing = () => {
     if (!this.mediaProperties.shareAudioTrack && !this.mediaProperties.shareVideoTrack) {
+      if (this.wirelessShare) {
```
marcin-bazyl, line 6222: if we do this handling of wirelessShare now here, then the name and comments of the `triggerStoppedSharing()` function need to be updated, because they don't match what the function is doing anymore.
Also previously we would only call `unpublishTracks()` in the `handleShareTrackEnded()` if `this.wirelessShare` was false, now we always call it and then do the leave, this may cause a problem especially for transcoded meetings because unpublish will result in a new SDP exchange and we will leave before that exchange is finished which can cause some undesired error metrics for the backend teams.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yes, but I think it's a fairly simple fix we need here - just change the boolean `this.isSharing` into an enum with 3 values (also rename it because current name is not very precise and can be confusing), so if we have: `this.shareFloorState: 'released' | 'pending' | 'granted' then inside `publishTracks()` you could do `floorRequestNeeded = (this.shareFloorState === 'released`)`, right?

prsaroha: fair point. added the enum screenShareFloorState

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: moved it back to the handleShareTrackEnded methods and added a check for both share tracks to have ended before leaving

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5788,6 +5802,7 @@ export default class Meeting extends StatelessWebexPlugin {
           })
           .then(() => {
             this.isSharing = true;
```
marcin-bazyl, line 5804: Please remove the `isSharing` flag, it's redundant now. Also there are 2 places where `isSharing` is used and `screenShareFloorState` should be used instead now:
1. in clearMeetingData()
2. in moveTo() in the listener for LOCUSINFO.EVENTS.SELF_OBSERVING

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6182,13 +6200,17 @@ export default class Meeting extends StatelessWebexPlugin {
...
    */
   private handleShareAudioTrackEnded = async () => {
-    try {
-      await this.unpublishTracks([this.mediaProperties.shareAudioTrack]);
-    } catch (error) {
-      LoggerProxy.logger.log(
-        'Meeting:index#handleShareAudioTrackEnded --> Error stopping share: ',
-        error
-      );
+    if (this.wirelessShare && !this.mediaProperties.hasLocalShareTrack()) {
```
marcin-bazyl, line 6203: this condition will always be false - at this point the audio or video share track is still set (same for handleShareVideoTrackEnded)

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6866,6 +6889,8 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     if (floorRequestNeeded) {
+      this.screenShareFloorState = ScreenShareFloorStatus.PENDING;
```
marcin-bazyl, line 6892: there is 1 more place where we enqueue SHARE_FLOOR_REQUEST, so we need to set that flag there too (probably best to create a helper function enqueueShareFloorRequest() and also we need to set `this.screenShareFloorState` to `ScreenShareFloorStatus.PENDING` in `requestScreenShareFloor()` itself, because it gets called from `requestScreenShareFloorIfPending()` without the queuing.

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: removed `isSharing` and added `screenShareFloorState` where it was being used

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: oh you're right. I changed this to check for the other track - so in handleShareAudioTrackEnded I'm checking for shareVideoTrack to be undefined. In case we had both tracks, whichever handler gets called second will trigger the call to `leave()`. Otherwise, if we only have one track, that track's handler will call it as the other is undefined already

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -961,24 +968,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
-     * @deprecated after v1.118.13
      */
-    this.isSharing = false;
+    this.shareStatus = SHARE_STATUS.NO_SHARE;
     /**
      * @instance
-     * @type {string}
-     * @readonly
+     * @type {ScreenShareFloorStatus}
      * @public
```
marcin-bazyl, line 981: I'm not sure if `screenShareFloorState` needs to be public, I know that previous isSharing was public, but for now let's keep it private and we'll only make it public when we find it's required

packages/@webex/plugin-meetings/src/media/properties.ts
```diff
@@ -149,6 +160,15 @@ export default class MediaProperties {
     this.unsetRemoteShare();
   }
 
+  /**
+   * returns if we have at least one local share track
```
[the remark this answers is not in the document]
prsaroha: changed this

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3333,34 +3342,64 @@ export default class Meeting extends StatelessWebexPlugin {
...
 
-    localDisplayTrack?.on(LocalTrackEvents.Ended, this.handleShareTrackEnded);
+    localDisplayTrack?.on(LocalTrackEvents.Ended, this.handleShareVideoTrackEnded);
     localDisplayTrack?.on(
       LocalTrackEvents.UnderlyingTrackChange,
       this.underlyingLocalTrackChangeHandler
     );
 
-    this.mediaProperties.mediaDirection.sendShare = !!localDisplayTrack;
+    this.mediaProperties.mediaDirection.sendShare =
```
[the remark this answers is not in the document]
prsaroha: true, changed as suggested

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3333,34 +3342,64 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+    this.mediaProperties.setLocalShareAudioTrack(localSystemAudioTrack);
+
+    localSystemAudioTrack?.on(LocalTrackEvents.Ended, this.handleShareAudioTrackEnded);
+    localSystemAudioTrack?.on(
+      LocalTrackEvents.UnderlyingTrackChange,
+      this.underlyingLocalTrackChangeHandler
+    );
+
+    this.mediaProperties.mediaDirection.sendShare =
```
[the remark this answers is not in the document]
prsaroha: did this

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5346,6 +5402,7 @@ export default class Meeting extends StatelessWebexPlugin {
       )
       .then(() => {
         if (localTracks?.screenShare?.video) {
+          this.screenShareFloorState = ScreenShareFloorStatus.PENDING;
```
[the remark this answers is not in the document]
prsaroha: added

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5939,6 +6001,7 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   private requestScreenShareFloorIfPending() {
     if (this.floorGrantPending && this.state === MEETING_STATE.STATES.JOINED) {
+      this.screenShareFloorState = ScreenShareFloorStatus.PENDING;
```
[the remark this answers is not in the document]
prsaroha: moved to the method

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5874,11 +5931,16 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    if (!this.mediaProperties.shareTrack || !this.mediaProperties.mediaDirection.sendShare) {
+    if (
+      !this.mediaProperties.hasLocalShareTrack() ||
+      !this.mediaProperties.mediaDirection.sendShare
+    ) {
       LoggerProxy.logger.log(
-        `Meeting:index#requestScreenShareFloor --> NOT requesting floor, because we don't have the share track anymore (shareTrack=${
-          this.mediaProperties.shareTrack ? 'yes' : 'no'
-        }, sendShare=${this.mediaProperties.mediaDirection.sendShare})`
+        `Meeting:index#requestScreenShareFloor --> NOT requesting floor, because we don't have the share track anymore (shareVideoTrack=${
```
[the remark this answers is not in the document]
prsaroha: did this

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6302,37 +6365,75 @@ export default class Meeting extends StatelessWebexPlugin {
...
    * @memberof Meeting
-   * @param {MediaStream} localShare
    * @returns {undefined}
    */
-  private handleShareTrackEnded = async () => {
-    if (this.wirelessShare) {
+  private handleShareAudioTrackEnded = async () => {
+    // current share audio track has ended, but there might be an active
+    // share video track. we only leave from wireless share if share has
+    // completely ended, which means no share audio or vidoe tracks active
```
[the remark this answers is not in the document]
prsaroha: oops

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6773,6 +6880,8 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     if (floorRequestNeeded) {
+      this.screenShareFloorState = ScreenShareFloorStatus.PENDING;
```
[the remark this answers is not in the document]
prsaroha: replaced with the new function

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3712,6 +3735,22 @@ describe('plugin-meetings', () => {
...
           });
+          
+          it('requests screen share floor and publishes the screen share audio track', async () => {
+            await meeting.publishTracks({screenShare: {audio: audioShareTrack}});
+            
+            assert.calledOnce(meeting.mediaProperties.webrtcMediaConnection.publishTrack);
+            checkScreenShareAudioPublished(audioShareTrack);
+          });
+          
+          it('does not request screen share floor when publishing share track if already sharing', async () => {
```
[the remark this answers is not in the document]
prsaroha: added new test

packages/@webex/plugin-meetings/src/meeting/index.ts
prsaroha: changed to private

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5920,6 +5995,8 @@ export default class Meeting extends StatelessWebexPlugin {
               stack: error.stack,
             });
 
+            this.screenShareFloorState = ScreenShareFloorStatus.RELEASED;
```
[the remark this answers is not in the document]
marcin-bazyl: This is not done as I asked in this comment. We're setting it to RELEASED irrespective of the `content` so also in the case if `content` is `true` and `this.shareStatus` is `SHARE_STATUS.LOCAL_SHARE_ACTIVE` - that's the case when we are sharing and already have the floor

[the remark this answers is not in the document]
prsaroha: updated

marcin-bazyl (approved):

arun3528 (approved):

PR #2958, opened 2023-07-18
fix(plugin-meetings): isVideoMuted should reflect remote video mute

chburket:
# COMPLETES
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-443363

The member.isVideoMuted flag was not taking remote mute into account.

Changed MemberUtil.isVideoMuted() to use participant.controls.video.muted for mute status before falling back to participant.status.videoStatue.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added unit tests and tested manually with Cantina.

### I certified that

---


packages/@webex/plugin-meetings/test/unit/spec/member/util.js
```diff
@@ -58,12 +59,16 @@ describe('plugin-meetings', () => {
...
               {type: 'COHOST', hasRole: true},
               {type: 'MODERATOR', hasRole: true},
-            ]
-          }
-        }
-      }
+            ],
+          },
+        },
+      };
```
chburket, line 65: eslint/prettier changes, skip to bottom for new tests

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -164,11 +164,18 @@ MemberUtil.isAudioMuted = (participant: any) => {
...
  */
-MemberUtil.isVideoMuted = (participant: any) => {
+MemberUtil.isVideoMuted = (participant: any): boolean => {
   if (!participant) {
     throw new ParameterError('Video could not be processed, participant is undefined.');
   }
 
+  // This seems to be the best way to determine isVideoMuted,
+  // especially when remote video muting is involved.
+  if (participant.controls?.video?.muted !== undefined) {
```
arun3528, line 174: cc @meta-dipanshu-sharma

arun3528 (approved):

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -164,11 +164,18 @@ MemberUtil.isAudioMuted = (participant: any) => {
  * @param {Object} participant the locus participant
  * @returns {Boolean}
  */
-MemberUtil.isVideoMuted = (participant: any) => {
+MemberUtil.isVideoMuted = (participant: any): boolean => {
```
marcin-bazyl, line 167: These changes don't look right to me. The implementation doesn't match `MemberUtil.isAudioMuted`:
If `participant.controls.video.muted === false` we will return `false`, completely ignoring the local mute status, while `isAudioMuted` takes local mute into account.
Also I think the logic for audio and video should be the same, so it would be better for all the logic for local and remote mute to be done inside the common `MemberUtil.isMuted` function and `isVideoMuted` and `isAudioMuted` should just call it with the right parameters for video and audio

packages/@webex/plugin-meetings/test/unit/spec/member/util.js
```diff
@@ -408,18 +377,60 @@ describe('plugin-meetings', () => {
...
+
+    it('returns true when controls is there and video muted', () => {
+      const participant = {
+        controls: {video: {muted: true}},
+      };
+
+      assert.isTrue(MemberUtil.isVideoMuted(participant));
+    });
+
+    it('returns false when controls is there and video is not muted', () => {
```
marcin-bazyl, line 410: this test and the previous one don't take into account the local mute state

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/member/util.ts
chburket: Yep, you are correct.  Fixed.

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -226,21 +218,28 @@ MemberUtil.isLiveAnnotationSupported = (participant) => {
...
+    return true;
   }
 
-  return null;
+  // check local mute
+  const localStatus = participant?.status?.[statusAccessor];
+  if (localStatus === _RECEIVE_ONLY_) {
+    return true;
+  }
+  if (localStatus === _SEND_RECEIVE_) {
```
marcin-bazyl, line 238: not strictly related to this PR, but I wonder why the existing code is only checking for _RECEIVE_ONLY_ and _SEND_RECEIVE_, why is it ignoring the other 2 possible values: _SEND_ONLY_ and _INACTIVE_, @arun3528 do you know?

packages/@webex/plugin-meetings/test/unit/spec/member/util.js
```diff
@@ -408,18 +377,111 @@ describe('plugin-meetings', () => {
...
   });
+
+  const getMuteStatus = (muted) => {
+    if (muted === undefined) {
+      return undefined;
+    }
+    return muted ? _RECEIVE_ONLY_ : _SEND_RECEIVE_;
+  };
+
+  describe.only('MemberUtil.isAudioMuted', () => {
```
marcin-bazyl, line 402: .only

packages/@webex/plugin-meetings/test/unit/spec/member/util.js
```diff
@@ -408,18 +377,111 @@ describe('plugin-meetings', () => {
...
+
+    it('returns defined status when the other is undefined', () => {
+      testResult(undefined, true, true);
+      testResult(undefined, false, false);
+      testResult(true, undefined, true);
+      testResult(false, undefined, false);
+    });
+  });
+
+  describe.only('MemberUtil.isVideoMuted', () => {
```
marcin-bazyl, line 445: another .only

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -226,21 +218,28 @@ MemberUtil.isLiveAnnotationSupported = (participant) => {
...
+    return false;
+  }
+
+  return remoteMute;
```
marcin-bazyl, line 242: Looking at this code it seems that `remoteMute` might be undefined and it would be better to always return a boolean here, but I see you have explicit unit tests for this case that check if we return a boolean, so I must be missing something.

chburket: I was trying to maintain the original behavior.  Not sure why returning undefined was desirable but didn't want to mess with it.  There is code in cantina that assumes muted if undefined.

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/member/util.ts
arun3528: recently someone was looking into it , @meta-dipanshu-sharma  was it you ?

PR #2961, opened 2023-07-19
fix(samples): video resolution issue fix

meta-dipanshu-sharma:
# COMPLETES #< [SPARK-389444](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-389444) >

Resolution and frame rate of the video streams is not displayed correctly on samples when we join another meeting after leaving the first meeting.

**UPDATED REQUIREMENT**:

Also we need to separate local video and remote video resolution check interval because local video can be created without joining a meeting. And remote video resolution interval check should be cleared when we leave a meeting but local video resolution interval check should be when we stop the local video.

 Did split meetingsResolutionCheckInterval method in 2 d/f methods which are `localVideoResolutionCheckInterval` and `remoteVideoResolutionCheckInterval`. Former is called when we do `getMediaStreams()` and stopped when we stop local video while later is called after `addMedia()` and stopped when leave a meeting.

`meetingsResolutionCheckInterval()` is called during register. Changed it to be called on `addMedia()` function.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -181,7 +181,6 @@ function register() {
       toggleUnifiedMeetings.removeAttribute('disabled');
       unregisterElm.disabled = false;
       unregisterElm.classList.add('btn--red');
-      meetingsResolutionCheckInterval();
     })
     .catch((error) => {
       console.warn('Authentication#register() :: error registering', error);
@@ -529,6 +528,7 @@ function leaveMeeting(meetingId) {
       passwordCaptchaStatusElm.style.backgroundColor = 'white';
       meetingsJoinPinElm.value = '';
       meetingsJoinCaptchaElm.value = '';
+      clearVideoResolutionCheckInterval(remoteVideoResElm, remoteVideoResolutionInterval);
     });
 }
 
@@ -661,6 +661,10 @@ function cleanUpMedia(mediaElements) {
       elem.srcObject.getTracks().forEach((track) => track.stop());
       // eslint-disable-next-line no-param-reassign
       elem.srcObject = null;
+      
+      if(elem.id === "local-video") {
+        clearVideoResolutionCheckInterval(localVideoResElm, localVideoResolutionInterval);
+      }
     }
   });
 }
@@ -912,6 +916,8 @@ function getMediaStreams(mediaSettings = getMediaSettings(), audioVideoInputDevi
         meetingStreamsLocalAudio.srcObject = new MediaStream(localStream.getAudioTracks());
       }
 
+      localVideoResolutionCheckInterval();
+      
       return {localStream};
     })
     .catch((error) => {
@@ -1261,9 +1267,11 @@ function clearMediaDeviceList() {
 }
 
 function getLocalMediaSettings() {
-  const meeting = getCurrentMeeting();
-  if(meeting && meeting.mediaProperties.videoTrack) {
-    const videoSettings = meeting.mediaProperties.videoTrack.getSettings();
+  const [localStream] = currentMediaStreams;
+  const localVideoTrack = localStream.getVideoTracks()[0];
+
+  if (localVideoTrack) {
+    const videoSettings = localVideoTrack.getSettings();
     const {frameRate, height} = videoSettings;
     localVideoResElm.innerText = `${height}p ${Math.round(frameRate)}fps`;
   }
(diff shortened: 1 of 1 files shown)


arun3528 (approved):

meta-dipanshu-sharma:
> Approving, because this change is making things better, but it's not a full fix. We should also move the call to `clearMeetingsResolutionCheckInterval()` - right now it's done on the first "media:stopped" event which kind of works, but probably better to do it when we leave the meeting. Also the meetingsResolutionCheckInterval() kicks of an interval that gets settings for local and remote tracks - this should be separated because we can display the settings for local tracks even before we join the meeting and call addMedia()

1. Yes we should move `clearMeetingsResolutionCheckInterval()` after we call leaveMeeting() instead of media:stopped event. Will update the PR.

2. Yes, instead of calling `meetingsResolutionCheckInterval()` after addMedia(), we can move it to after we do `getMediaStreams()`. Now regarding separating `getLocalMediaSettings()` and `getRemoteMediaSettings()` we have 2 options here:
  2.1. Either we should not separate them because anyways we are checking if remoteVideoTrack exists or not before fetching 
         the settings for it so it will return directly without throwing an error.
  2.2 Or we can separate just it will be a slight complex logic then.  
  
What do you suggest? @marcin-bazyl @arun3528

marcin-bazyl:
> > Approving, because this change is making things better, but it's not a full fix. We should also move the call to `clearMeetingsResolutionCheckInterval()` - right now it's done on the first "media:stopped" event which kind of works, but probably better to do it when we leave the meeting. Also the meetingsResolutionCheckInterval() kicks of an interval that gets settings for local and remote tracks - this should be separated because we can display the settings for local tracks even before we join the meeting and call addMedia()
> 
> 1. Yes we should move `clearMeetingsResolutionCheckInterval()` after we call leaveMeeting() instead of media:stopped event. Will update the PR.
> 2. Yes, instead of calling `meetingsResolutionCheckInterval()` after addMedia(), we can move it to after we do `getMediaStreams()`. Now regarding separating `getLocalMediaSettings()` and `getRemoteMediaSettings()` we have 2 options here:
>    2.1. Either we should not separate them because anyways we are checking if remoteVideoTrack exists or not before fetching the settings for it so it will return directly without throwing an error.
>    2.2 Or we can separate just it will be a slight complex logic then.
> 
> What do you suggest? @marcin-bazyl @arun3528

I think we should separate them, because they have different lifetime. The one for local video should start when we get the camera stream and stop when we stop the local camera stream - this can happen without even having any meeting. The remote video one is only valid after we call addMedia and should be stopped when we leave the meeting.

meta-dipanshu-sharma:
> > > Approving, because this change is making things better, but it's not a full fix. We should also move the call to `clearMeetingsResolutionCheckInterval()` - right now it's done on the first "media:stopped" event which kind of works, but probably better to do it when we leave the meeting. Also the meetingsResolutionCheckInterval() kicks of an interval that gets settings for local and remote tracks - this should be separated because we can display the settings for local tracks even before we join the meeting and call addMedia()
> > 
> > 
> > 
> > 1. Yes we should move `clearMeetingsResolutionCheckInterval()` after we call leaveMeeting() instead of media:stopped event. Will update the PR.
> > 2. Yes, instead of calling `meetingsResolutionCheckInterval()` after addMedia(), we can move it to after we do `getMediaStreams()`. Now regarding separating `getLocalMediaSettings()` and `getRemoteMediaSettings()` we have 2 options here:
> >    2.1. Either we should not separate them because anyways we are checking if remoteVideoTrack exists or not before fetching the settings for it so it will return directly without throwing an error.
> >    2.2 Or we can separate just it will be a slight complex logic then.
> > 
> > What do you suggest? @marcin-bazyl @arun3528
> 
> I think we should separate them, because they have different lifetime. The one for local video should start when we get the camera stream and stop when we stop the local camera stream - this can happen without even having any meeting. The remote video one is only valid after we call addMedia and should be stopped when we leave the meeting.

Updated the code. Created 2 separate functions for local and remote video settings check. Called and stopped them at correct time as discussed above.

marcin-bazyl (approved):
Looks good, are you going to port this to the beta branch too?

meta-dipanshu-sharma:
> Looks good, are you going to port this to the beta branch too?

Yes

PR #2969, opened 2023-07-21
fix(samples): video resolution issue fix

meta-dipanshu-sharma:
NOTE: This is PR for beta where master branch [PR](https://github.com/webex/webex-js-sdk/pull/2961) is already merged.

# COMPLETES #< [SPARK-389444](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-389444) >

Resolution and frame rate of the video streams is not displayed correctly on samples when we join another meeting after leaving the first meeting.

**UPDATED REQUIREMENT**:

Also we need to separate local video and remote video resolution check interval because local video can be created without joining a meeting. And remote video resolution interval check should be cleared when we leave a meeting but local video resolution interval check should be when we stop the local video.

 Did split meetingsResolutionCheckInterval method in 2 d/f methods which are `localVideoResolutionCheckInterval` and `remoteVideoResolutionCheckInterval`. Former is called when we do `getUserMedia()` and stopped when we stop local video while later is called after `addMedia()` and stopped when leave a meeting.

`meetingsResolutionCheckInterval()` is called during register. Changed it to be called on `addMedia()` function.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -190,7 +190,6 @@ function register() {
       toggleUnifiedMeetings.removeAttribute('disabled');
       unregisterElm.disabled = false;
       unregisterElm.classList.add('btn--red');
-      meetingsResolutionCheckInterval();
     })
     .catch((error) => {
       console.warn('Authentication#register() :: error registering', error);
@@ -556,6 +555,7 @@ function leaveMeeting(meetingId) {
       enableMultistreamControls(false);
       breakoutHostOperation.innerHTML = '';
       breakoutTable.innerHTML = '';
+      clearVideoResolutionCheckInterval(remoteVideoResElm, remoteVideoResolutionInterval);
     });
 }
 
@@ -960,6 +960,15 @@ function handleTrackUnpublished(track, mediaElement, debugString) {
     if (!isPublished) {
       console.log(`MeetingControls#getUserMedia() :: ${debugString} track unpublished, stopping it`);
       track.stop();
+
+      if(mediaElement.id === "local-video") {
+        clearVideoResolutionCheckInterval(localVideoResElm, localVideoResolutionInterval);
+        meetingStreamsLocalVideo.srcObject= null;
+      }
+
+      if(mediaElement.id === "local-audio") {
+        meetingStreamsLocalAudio.srcObject = null;
+      }
     }
   });
 }
@@ -1009,6 +1018,8 @@ async function getUserMedia(constraints = {audio: true, video: true}) {
     meetingStreamsLocalVideo.srcObject = localMedia.cameraTrack.underlyingStream;
   }
 
+  localVideoResolutionCheckInterval();
+  
   console.log('MeetingControls#getUserMedia() :: got following local tracks:', localMedia.microphoneTrack, localMedia.cameraTrack);
   return;
 }
@@ -1313,12 +1324,9 @@ function clearMediaDeviceList() {
 }
 
 function getLocalMediaSettings() {
-  const meeting = getCurrentMeeting();
-
-  if (meeting && meeting.mediaProperties.videoTrack) {
-    const videoSettings = meeting.mediaProperties.videoTrack?.getSettings();
+  if (localMedia.cameraTrack) {
+    const videoSettings = localMedia.cameraTrack.getSettings();
     const {frameRate, height} = videoSettings;
-
     localVideoResElm.innerText = `${height}p ${Math.round(frameRate)}fps`;
   }
 }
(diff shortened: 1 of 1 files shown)


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -703,6 +703,8 @@ function cleanUpMedia() {
   // the local tracks get stopped when we get the unpublished event for them, so here we only need to lose the reference to them
   localMedia.microphoneTrack = undefined;
   localMedia.cameraTrack = undefined;
+  meetingStreamsLocalAudio.srcObject = null;
```
marcin-bazyl, line 706: maybe it would be better to do this inside the handler for when the track is unpublished (in `handleTrackUnpublished()`) - that's also where we stop it.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1329,26 +1334,29 @@ function getRemoteMediaSettings() {
...
 
-let resolutionInterval;
+let localVideoResolutionInterval;
+let remoteVideoResolutionInterval;
 const INTERVAL_TIME = 3000;
 
-function meetingsResolutionCheckInterval() {
-  resolutionInterval = setInterval(() => {
+function localVideoResolutionCheckInterval() {
+  localVideoResolutionInterval = setInterval(() => {
```
marcin-bazyl, line 1346: I think localVideoResolutionCheckInterval() can be called multiple times (it's called from getUserMedia() which is called for example every time user changes the camera), so we need to clear the previous interval before setting the new one

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2144,6 +2152,8 @@ function addMedia() {
       forceScreenShareViewLayout(meeting);
     }
 
+    remoteVideoResolutionCheckInterval();
```
marcin-bazyl, line 2155: `remoteVideoResolutionCheckInterval()` in the current implementation only works for transcoded meetings, so we should call it here only if `!isMultistream`, or you can make it work for multistream too

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1329,26 +1334,29 @@ function getRemoteMediaSettings() {
...
     getLocalMediaSettings();
-    getRemoteMediaSettings();
   }, INTERVAL_TIME);
 }
 
-function clearMeetingsResolutionCheckInterval() {
-  localVideoResElm.innerText = '';
-  remoteVideoResElm.innerText = '';
+function remoteVideoResolutionCheckInterval() {
+  remoteVideoResolutionInterval = setInterval(() => {
```
marcin-bazyl, line 1352: remoteVideoResolutionCheckInterval() is called only once per meeting, so doesn't necessarily need the same logic to clear the previous interval like I commented on localVideoResolutionCheckInterval(), but maybe it would be good to add it here anyway for completeness?

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Yeah updated the code for both `meetingStreamsLocalAudio` and `meetingStreamsLocalVideo`

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Yes, added a check to clear interval if exist before starting a new one for both `localVideoResolutionCheckInterval` and `remoteVideoResolutionCheckInterval `

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Removed the function call for multistream meetings as of now. For multistream, we can take this as a separate jira as it will require some logic to add resolution checker for each video pane.

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Done

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: thanks, but I can see that neither `localVideoResolutionInterval` nor `remoteVideoResolutionInterval` is cleared when  `clearVideoResolutionCheckInterval()` is called, so we may end up calling `clearInterval` on an interval that's already cleared. Not sure if this is big problem, I guess `clearInterval()` will just not do anything in that case? it would be cleaner to reset these variables to undefined when the interval is cleared.

marcin-bazyl (approved):

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Yes, that's correct. `clearInterval` just clears the interval associated with that particular id returned by setInterval. Also there's nothing clearInterval will do if interval is already cleared but yeah just for code cleanliness we can set those variables to null.

meta-dipanshu-sharma: Updated the code

arun3528 (approved):
looks good thanks dipanshu

arun3528:
merging it as its a sample change

PR #2973, opened 2023-07-24
refactor: remove package version numbers

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to remove all version numbers from `package.json` files within this project. This is a necessary step in creating and using the new deployment pipeline.

## Links

[SPARK-446314](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-446314)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,7 +4,7 @@ run-name: ${{ github.actor }} is running Deploy CD
 on:
   push:
     branches: # White-list of deployable tags and branches.
-      - latest
+      - next
 
 env:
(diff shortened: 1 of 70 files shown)


mkesavan13 (commented):
Thanks for the PR Tim.

Looks good. However, if we remove the version number from everywhere, where are we going to keep track of what is the last SDK version that we published?

Unless we have a reference to that, we cannot bump up the version when we publish the SDK. Right?

mkesavan13 (approved):
Never mind about the previous comment. I went through the increment command and understood that we will obtain the latest version number of each package from npm itself using the npm view command

**Approving PR**

arun3528 (approved):

PR #2979, opened 2023-07-27
feat(ca-metrics): do not subtract stay in lobby for mediaJMT

torgeadelin:
# COMPLETES [SPARK-422888](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-422888)
[SPARK-449067](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-449067)

- since we're creating the media connection in lobby now, it doesn't make sense to subtract it from totalMediaJMT

- refactor

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

-unit test

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
@@ -311,10 +311,9 @@ export default class CallDiagnosticLatencies {
     const clickToInterstitial = this.getClickToInterstitial();
     const interstitialToJoinOk = this.getInterstitialToJoinOK();
     const joinConfJMT = this.getJoinConfJMT();
-    const stayLobbyTime = this.getStayLobbyTime() || 0;
 
     if (clickToInterstitial && interstitialToJoinOk && joinConfJMT) {
-      return clickToInterstitial + interstitialToJoinOk + joinConfJMT - stayLobbyTime;
+      return clickToInterstitial + interstitialToJoinOk + joinConfJMT;
     }
 
     return undefined;
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
@@ -1,6 +1,6 @@
 /* eslint-disable class-methods-use-this */
 /* eslint-disable valid-jsdoc */
-import {StatelessWebexPlugin} from '@webex/webex-core';
+import {WebexPlugin} from '@webex/webex-core';
 
 import {MetricEventNames} from '../metrics.types';
 
@@ -11,11 +11,11 @@ import {MetricEventNames} from '../metrics.types';
  * @exports
  * @class CallDiagnosticLatencies
  */
-export default class CallDiagnosticLatencies extends StatelessWebexPlugin {
+export default class CallDiagnosticLatencies extends WebexPlugin {
   latencyTimestamps: Map<MetricEventNames, number>;
   precomputedLatencies: Map<string, number>;
   // meetingId that the current latencies are for
-  meetingId?: string;
+  private meetingId?: string;
 
   /**
    * @constructor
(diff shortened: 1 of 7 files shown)


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -311,10 +311,9 @@ export default class CallDiagnosticLatencies {
     const clickToInterstitial = this.getClickToInterstitial();
     const interstitialToJoinOk = this.getInterstitialToJoinOK();
     const joinConfJMT = this.getJoinConfJMT();
-    const stayLobbyTime = this.getStayLobbyTime() || 0;
 
     if (clickToInterstitial && interstitialToJoinOk && joinConfJMT) {
-      return clickToInterstitial + interstitialToJoinOk + joinConfJMT - stayLobbyTime;
```
marcin-bazyl, line 317: we should do this or not based on the value of `allowMediaInLobby` passed in to `addMedia()`

maxinteger (approved):

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -184,6 +194,30 @@ describe("internal-plugin-metrics", () => {
...
+      cdl.saveTimestamp('internal.client.interstitial-window.click.joinbutton', 10);
+      cdl.saveTimestamp('client.locus.join.request', 12);
+      cdl.saveTimestamp('client.locus.join.response', 20);
+      cdl.saveTimestamp('internal.host.meeting.participant.admitted', 24)
+      cdl.saveTimestamp('client.ice.start', 30);
+      cdl.saveTimestamp('client.ice.end', 40);
+      assert.deepEqual(cdl.getTotalMediaJMT(), 27);
+    })
+
+    it('calculates getTotalMediaJMT correctly with allowMediaInLobby true', () => {
```
marcin-bazyl, line 209: it's probably worth keeping the call to saveTimestamp with 'internal.host.meeting.participant.admitted' in this test too, because that's what will happen in reality and to really test that the time when participant is admitted doesn't affect the TotalMediaJMT

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -54,6 +52,8 @@ class Metrics extends WebexPlugin {
     this.webex.once('ready', () => {
       // @ts-ignore
       this.callDiagnosticMetrics = new CallDiagnosticMetrics({}, {parent: this.webex});
+      // @ts-ignore
+      this.callDiagnosticLatencies = new CallDiagnosticLatencies({webex: this.webex});
```
marcin-bazyl, line 56: seeing that the constructor for `CallDiagnosticMetrics` takes webex instance as a `parent` and CallDiagnosticLatencies takes it as `webex` makes me wonder if maybe CallDiagnosticLatencies class should also extend a `WebexPlugin`?

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -160,6 +160,10 @@ class Metrics extends WebexPlugin {
     options: SubmitClientEventOptions;
   }) {
     this.callDiagnosticLatencies.saveTimestamp(name);
+    // save the meetingId so we can use the meeting object in latency calculations if needed
+    if (options?.meetingId && !this.callDiagnosticLatencies.meetingId) {
```
marcin-bazyl, line 164: just wondering... wouldn't it be better for `saveTimestamp()` to just take meetingId as an optional parameter and have this logic of storing meetingId inside CallDiagnosticLatencies? Then you don't need the new public `setMeetingId()` method - currently it feels a bit weird that we have a setter for a public property meetingId - it's public so you could just set it without the setter, or maybe it should be private, but then you need a getter....

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
torgeadelin: that's a good point, added

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
torgeadelin: i've refactored to extend the plugin

marcin-bazyl (approved):

arun3528 (approved):

PR #2993, opened 2023-08-07
fix: added error for transcription not starting when guest joins first

Kesari3008:
# COMPLETES #[SPARK-432028](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-432028)

Added error message to be thrown if guest join the meeting before host and tries to start the transcription.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3760,6 +3761,14 @@ export default class Meeting extends StatelessWebexPlugin {
         opened LLM web socket connection successfully.`
       );
 
+      Object.values(this.members.membersCollection.members).forEach((member: Member) => {
```
arun3528, line 3764: i am not sure if we should throw an error, rather auto-subscribe and start receiving the events once the host starts 

we can do that with voicia plugin migration too whats do you think @Kesari3008 , The problem here I am not aware of any events which indicate that the transcript it enabled by host later in the meeting , we will need to stream line the flow

sreenara: @arun3528 what's the scope of the work to move this entire logic to voicea though? We would like to be able to complete this story with the bug fix and work on moving to voicea in the ticket I pinged you yesterday [SPARK-424739](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-424739)

sreenara: Also I think we are only logging the error so the developer is aware of why the transcription failed.

Kesari3008: @arun3528 if Webex assistant is enabled for the guest, then the guest will start receiving the transcription once the host starts it. This error message is only to indicate why it didn't start for the guest immediately when they subscribed for it.

meta-dipanshu-sharma:
AFAIK, we don't commit `*.min.js` or `*.min.js.map` files and also these are present in `gitignore` file. Any reason why we have committed them here?

packages/@webex/plugin-meetings/src/meeting/index.ts
arun3528: sorry did not see the message , The for loop here is not needed ,we should be able to check controls in the locusinfo to decide if need to throw and error on not

Neeraj-swarnkar: @Kesari3008 you can use `this.locusInfo.host` to have the host details and add your condition for the guest. Also `this.locusInfo.controls.transcribe` will have Transcription details.  @arun3528

Neeraj-swarnkar (changes requested):
Added comment for checking the host, As @meta-dipanshu-sharma also suggested, we should not commit min.js files. Any reason for adding those files ???

packages/@webex/plugin-meetings/src/meeting/index.ts
Kesari3008: Using isClosedCaptionsActive property now. This is set to false until any licensed user enables transcription. Once they enable it, transcription starts working for the guest user irrespective of the isWebexAssistantActive property which is only set to true if the host has it enabled.

sreenara (approved):
Code change looks good. Please remove the min.js files in the commit and update the changes in the PR description.

Kesari3008:
> Code change looks good. Please remove the min.js files in the commit and update the changes in the PR description.

Update the PR description

packages/@webex/plugin-meetings/src/meeting/index.ts
Kesari3008: @arun3528 Please check now

arun3528 (approved):

arun3528:
@Kesari3008  lets do this on beta too looks good thank you

PR #3015, opened 2023-08-21
feat(call-leave): allow metric options to be prepared ahead of browser close

chburket:
## COMPLETES

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-451786

We need a way to prepare the options needed to submit a new metric via fetch.  This will allow us to call the function from cantina and then simply call fetch with the prepared options when the browser/tab is closed.

Adds a new function equivalent to newMetric.submitClientEvent that prepares all the options required to submit a metric via fetch, but does not call fetch.

Rearranges some code into utils so it can be called by both submitClientEvent and prepareMetricFetchOptions.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Unit tests and manual tests via updated cantina code.

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-batcher.ts
```diff
@@ -1,116 +1,19 @@
...
-    if (
-      webClientDomain?.includes('localhost') ||
-      webClientDomain?.includes('127.0.0.1') ||
-      process.env.NODE_ENV !== 'production'
-    ) {
-      return 'test';
-    }
-
-    return process.env.NODE_ENV === 'production' ? 'prod' : 'test';
-  },
```
chburket, line 30: moved to util file

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-batcher.ts
```diff
@@ -1,116 +1,19 @@
...
-
-    if (!isEmpty(audioSetupDelay)) {
-      item.eventPayload.event = merge(item.eventPayload.event, {audioSetupDelay});
-    }
-
-    if (!isEmpty(videoSetupDelay)) {
-      item.eventPayload.event = merge(item.eventPayload.event, {videoSetupDelay});
-    }
-
-    item.eventPayload.origin = Object.assign(origin, item.eventPayload.origin);
```
chburket, line 111: moved to util file

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -145,6 +145,36 @@ class Metrics extends WebexPlugin {
...
+  }: {
+    name: ClientEvent['name'];
+    payload?: RecursivePartial<ClientEvent['payload']>;
+    options: SubmitClientEventOptions;
+  }) {
+    // TODO: what to do with this?
+    // this.callDiagnosticLatencies.saveTimestamp({
+    //   key: name,
+    //   options: {meetingId: options?.meetingId},
+    // });
```
chburket, line 170: We can't save a timestamp here, because it would be the time that the options were prepared, not when the fetch happens.  The fetch will happen in cantina, though, and I'm not sure I could/should call saveTimestamp from there.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -525,4 +549,42 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
...
+  /**
+   * Prepare Client Event CA event.
+   * @param arg - submit params
+   * @param arg.event - event key
+   * @param arg.payload - additional payload to be merged with default payload
+   * @param arg.options - payload
+   * @returns {Promise<any>}
+   * @throws
+   */
+  public async prepareMetricFetchOptions({
```
arun3528, line 562: @chburket  can you explain more on how this work , isn't there a way we could send this metic on the SDK side too? we would be happy to help it

cc @sreenara  @mkesavan13

chburket: @arun3528, I used a similar strategy for CSAT surveys when the browser is closed.

There is a browser window event that we can listen for (` beforeUnload`) that fires when the browser tab is refreshed or closed, and when the browser is closed. Requests can be made from here but the `fetch` API must be used because it has a `keepalive` option that allows the request to outlive the tab.

In addition, we've found that there is only a small window for making these requests. Calling async functions, dispatching actions, etc. will often not complete when the browser itself is closed (otherwise we could call `webex.internal.newMetrics.submitClientEvent(args)` directly here).

So the idea is to pre-gather all the information/options needed for the request(s), and then simply and quickly fire the fetch(es) when `beforeUnload` is triggered.

We will do this by calling `webex.internal.newMetrics.prepareFetchOptions(args)` from a hook in Cantina when a meeting is active, and generate a function to "fetch" the metric data for immediate use when `beforeUnload` fires. In addition, this hook will add/remove the `beforeUnload` event handler from `window` as needed (e.g. add immediately after meeting join, and remove when user leaves).

I assumed the `window.on('beforeUnload')` logic was a better fit in Cantina, but please let me know if that is not the case. The SDK `prepareFetchOptions()` function could be made to return the function to call when browser is closing (e.g. `prepareFetchFunction`), but it's useful to be able to tell if the options changed so we know when to update the `beforeUnload` listener. We could definitely create an SDK function to do the actual `fetch` but that is one line of code.

There *will* be other code that will need to be run immediately before the fetch is called (updating the `originTime` values) so it would make some sense for that logic to be in the SDK, but it's pretty simple code and we can refactor later, as needed.

So anyway, that's the general idea. I'm happy to hear better ideas! As I said, this strategy is already in use for CSAT surveys, although more is done in Cantina because more of the old metrics code is in Cantina (vs the SDK).

chburket: @torgeadelin and I are going to discuss offline

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -129,3 +129,9 @@ export type SubmitMQE = (args: {
   payload: SubmitMQEPayload;
   options: any;
 }) => void;
+
+export type PrepareMetricFetchOptions = (args: {
```
torgeadelin, line 133: Also rename this accordingly.

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -145,6 +145,36 @@ class Metrics extends WebexPlugin {
...
+  /**
+   * Returns a promise that will resolve to fetch options for submitting a metric.
+   * @public
+   * @param {Object} arg
+   * @param {String} arg.name - event name
+   * @param {Object} arg.payload - event payload
+   * @param {Object} arg.options - other options
+   * @returns {Promise} promise that resolves to options to be used with fetch
+   */
+  public async prepareMetricFetchOptions({
```
torgeadelin, line 157: Also rename this accordingly.

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
torgeadelin: You can remove this, we won't use it in this case.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
torgeadelin: Also important to note that the event won't have the correct triggerTime and needs appending at the time of fetch. Worth mentioning in the comment as well.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-batcher.ts
```diff
@@ -1,116 +1,19 @@
...
 
-import {isEmpty, merge} from 'lodash';
 import Batcher from '../batcher';
-import {
-  ClientEvent,
-  MetricEventNames,
-  MediaQualityEventAudioSetupDelayPayload,
-  MediaQualityEventVideoSetupDelayPayload,
-} from '../metrics.types';
+import {prepareDiagnosticMetricItem} from './call-diagnostic-metrics.util';
```
torgeadelin, line 5: I would move this to call-diagnostic-metrics-batcher.util.ts

chburket: It's being called from `call-diagnostic-metrics` and `call-diagnostic-metrics-batcher`, so I think it is fine where it is.  If you feel strongly, I will create a new `call-diagnostic-metrics-batcher.util.ts` file.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -474,14 +475,15 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
...
+   * Prepare Client Event CA event.
    * @param arg - submit params
    * @param arg.event - event key
    * @param arg.payload - additional payload to be merged with default payload
    * @param arg.options - payload
+   * @returns {any} options to be with fetch
    * @throws
    */
-  public submitClientEvent({
+  private prepareClientEvent({
```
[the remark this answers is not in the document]
chburket: This was not renamed, the diff just makes it look that way.  `submitClientEvent` now calls the new `prepareClientEvent` and then submits.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/generated-types-temp/ClientEvent.ts
```diff
@@ -491,187 +498,188 @@ export interface ClientEvent {
...
+    | 'client.daemon.pending-on-existing-upgrade-job'
+    | 'client.daemon.query-site-info'
+    | 'client.daemon.upgrade-and-join'
+    | 'client.daemon.wakeup-meeting-client'
+    | 'client.device.resource-room.joined'
+    | 'client.e2ee.join.response'
+    | 'client.embedded-object-info'
+    | 'client.entering-background'
+    | 'client.entering-foreground'
+    | 'client.exit.app'
```
chburket, line 541: This is the only real change

packages/@webex/internal-plugin-metrics/src/call-diagnostic/generated-types-temp/Event.ts
```diff
@@ -714,187 +720,188 @@ export interface Event {
...
+          | 'client.daemon.pending-on-existing-upgrade-job'
+          | 'client.daemon.query-site-info'
+          | 'client.daemon.upgrade-and-join'
+          | 'client.daemon.wakeup-meeting-client'
+          | 'client.device.resource-room.joined'
+          | 'client.e2ee.join.response'
+          | 'client.embedded-object-info'
+          | 'client.entering-background'
+          | 'client.entering-foreground'
+          | 'client.exit.app'
```
chburket, line 763: this is the only real change

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
chburket: Done

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
chburket: Added comments

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -158,14 +159,65 @@ class Metrics extends WebexPlugin {
     name: ClientEvent['name'];
     payload?: RecursivePartial<ClientEvent['payload']>;
     options: SubmitClientEventOptions;
-  }): Promise<any> {
+  }) {
```
torgeadelin, line 162: Why was this removed? I added it last week.

packages/@webex/http-core/src/index.js
```diff
@@ -68,6 +68,36 @@ export const protoprepareFetchOptions = curry(function protoprepareFetchOptions(
...
+ * Sets the $timings value(s) before the request/fetch.
+ * This function is only useful if you are about to send a request
+ * using prepared fetch options; normally it is done in webex.request();
+ *
+ * @param {any} options
+ * @returns {any} the updated options object
+ */
+const setRequestTimings = (options) => {
+  const now = new Date().getTime();
+  options.$timings = options.$timings || {};
```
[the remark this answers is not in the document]
chburket: Because that's how the NetworkTimingInterceptor named it.  I didn't make up new properties, just used the ones that were already being set elsewhere.

torgeadelin (approved):

arun3528 (approved):

PR #3038, opened 2023-08-31
Stream classes

marcin-bazyl:
draft PR for WCME 3.0 support created from stream-classes branch so that it's easier to see the whole diff in 1 place and add comments


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1206,53 +1171,38 @@ async function startScreenShare() {
...
-      localMedia.screenShare.audio = localShareAudioTrack;
-      localMedia.screenShare.audio.on('published-state-update', ({isPublished}) => {
-        if (!isPublished) {
-          console.log('MeetingControls#startScreenShare() :: local share audio track unpublished, stopping it');
+      localMedia.screenShare.audio = undefined;
+      isAudioShareEnded = true;
 
-          localMedia.screenShare.audio.stop();
-          localMedia.screenShare.audio = undefined;
+      if(isAudioShareEnded && isVideoShareEnded) meetingStreamsLocalShare.srcObject = null;
```
marcin-bazyl, line 1196: why do we need `isAudioShareEnded` and `isVideoShareEnded`? we can do this by just checking `localMedia.screenShare.audio` and `localMedia.screenShare.video`, right?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1325,8 +1278,8 @@ function clearMediaDeviceList() {
 }
 
 function getLocalMediaSettings() {
-  if (localMedia.cameraTrack) {
-    const videoSettings = localMedia.cameraTrack.getSettings();
+  if (localMedia.cameraStream) {
+    const videoSettings = localMedia.cameraStream.getTracks()[0]?.getSettings();
```
marcin-bazyl, line 1282: LocalStream class has a `getSettings()` method that calls getSettings() on the inputTrack, so we should probably use that here, right? BTW, I can't find a `getTracks()` method in the webrtc-core stream classes, so does this existing code work at all?

packages/@webex/plugin-meetings/src/media/properties.ts
```diff
@@ -23,18 +24,18 @@ export type MediaDirection = {
...
   mediaSettings: any;
   webrtcMediaConnection: any;
-  remoteAudioTrack: any;
+  remoteAudioStream: RemoteStream;
   remoteQualityLevel: any;
-  remoteShare: any;
-  remoteVideoTrack: any;
-  shareVideoTrack?: LocalDisplayTrack;
-  shareAudioTrack?: LocalSystemAudioTrack;
+  remoteShare: RemoteStream;
```
marcin-bazyl, line 33: maybe this should be renamed to `remoteShareStream` for consistency?

packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -29,18 +31,22 @@ export type ServerMuteReason =
...
-export enum LocalCameraTrackEvents {
+export enum LocalCameraStreamEventNames {
   ServerMuted = 'muted:byServer',
 }
 
-export class LocalMicrophoneTrack extends WcmeLocalMicrophoneTrack {
+export class LocalMicrophoneStream extends WcmeLocalMicrophoneStream {
   private unmuteAllowed = true;
 
+  [LocalMicrophoneStreamEventNames.ServerMuted] = new TypedEvent<
```
marcin-bazyl, line 46: don't we need to also use `AddEvents` and export the type created with `WithEventsDummyType`? similarily like it's done in webrtc-core when classes emitting events extend other classes emitting events? see here: https://github.com/webex/webrtc-core/blob/787e429d070ac40aa8a817eeb2eee21a6fe1e72c/src/media/local-stream.ts#L193

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -18,7 +19,13 @@ import BrowserDetection from '../common/browser-detection';
 
 const {isBrowser} = BrowserDetection();
 
-type MultistreamConnectionConfig = ConstructorParameters<typeof MultistreamRoapMediaConnection>[0];
+// TODO: in internal-media-core add appropriate types in MultistreamConnectionConfig for enable main audio and video
```
marcin-bazyl, line 22: I don't understand this todo comment, but anyway, enableMainAudio and enableMainVideo have been removed from MultistreamConnection constructor parameters in WCME 3.0, so any references to them should be removed also from internal-media-core (I can see they're still there) and from here and from Media.createMediaConnection function

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -201,11 +209,13 @@ Media.createMediaConnection = (
       },
     },
     {
+      // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now
```
marcin-bazyl, line 212: it would be better to reference a jira in any TODOs that we put in the code

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -201,11 +209,13 @@ Media.createMediaConnection = (
...
       localTracks: {
-        audio: audioTrack?.underlyingTrack,
-        video: videoTrack?.underlyingTrack,
-        screenShareVideo: shareVideoTrack?.underlyingTrack,
-      },
+        audio: audioStream?.outputStream.getTracks()[0],
+        video: videoStream?.outputStream.getTracks()[0],
+        screenShareVideo: shareVideoStream?.outputStream.getTracks()[0],
+        screenShareAudio: shareAudioStream?.outputStream.getTracks()[0], // TODO: add type for screenShareAudio in internal-media-core
```
marcin-bazyl, line 217: again, TODO without a jira - this one should reference https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-446923 and same todo comment is twice in 2 consecutive lines, shouldn't one be enough?

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -370,21 +380,23 @@ Media.toggleStream = () => {};
...
   return Promise.resolve().then(() => {
-    if (track && track.stop) {
+    if (stream && stream.stop) {
       try {
-        track.stop();
+        stream.stop();
       } catch (e) {
         LoggerProxy.logger.error(
-          `Media:index#stopTracks --> Unable to stop the track with state ${track.readyState}, error: ${e}`
+          `Media:index#stopTracks --> Unable to stop the track with state ${
```
marcin-bazyl, line 397: not sure if this log makes sense..... it should say "stream" not "track" and why is it logging the readyState of only the output stream's track? it should probably log it for both inputTrack and outputTrack

docs/samples/browser-plugin-meetings/app.js
mkesavan13: I think this is a mistake. Will update it.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5225,18 +5212,25 @@ export default class Meeting extends StatelessWebexPlugin {
     this.mediaProperties.setMediaPeerConnection(mc);
     this.setupMediaConnectionListeners();
 
-    // publish the tracks
-    if (this.mediaProperties.audioTrack) {
-      await this.publishTrack(this.mediaProperties.audioTrack);
+    if (this.isMultistream) {
+      this.sendSlotManager.createSlot(mc, MediaType.VideoMain, false);
```
marcin-bazyl, line 5216: don't we need to set the `active` parameter value based on the values of `audioEnabled`/`videoEnabled` parameters that client passed to `addMedia()`? also not sure about the hardcoded `false` for slides...

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: @marcin-bazyl - This was added after the tests failed because of this. I think particularly this test [here](https://github.com/webex/webex-js-sdk/commit/9a313e115ed59d2baacb7e4b4c935c142136c5e0#diff-826c5d0df79137a1fb8dba26f75550d10afa643de27ee68684256200af661a17R226).

Looks like it checks if MultistreamRoapMediaConnection's constructor is called with `enableMainAudio` and `enableMainVideo`

So, I thought those are missed out when creating types. Do we want to remove that check from the test instead?

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Looks like instead of moving it to next line, I've copy pasted. Will remove one of them and add the Jira

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: @marcin-bazyl - Do you remember? We decided that we will createSlot for all four Media types here but it will be disabled by default and later updateMedia shall enable it as per the need. 

That is why we do a `createSlot` here with an inactive state for each sendSlot.

docs/samples/browser-plugin-meetings/app.js
mkesavan13: I agree. I introduced that so, both events will update their respective variables and one of them would make the srcObject to null but it makes sense to use `localMedia` reference instead. Will update it.

packages/@webex/plugin-meetings/src/media/properties.ts
mkesavan13: I agree. Actually, there's a `shareAudioStream` and a `shareVideoStream` already. We might as well remove it if its not used anywhere.

@brycetham - Any reason why we have this here?

packages/@webex/media-helpers/src/webrtc-core.ts
mkesavan13: I am still not very familiar with Typed events. I'll leave it to Bryce to answer.

CC - @brycetham

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: I'll create a Jira for it and add. Thanks

@marcin-bazyl / @sreenara - We should actually create this story to take up after the v3 release. Is that correct?

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -370,21 +380,23 @@ Media.toggleStream = () => {};
 
 /**
  * Stop input stream
- * @param {MediaTrack} track A media stream
+ * @param {LocalStream} stream A local stream
  * @returns {null}
  */
-Media.stopTracks = (track: any) => {
-  if (!track) {
+Media.stopStreams = (stream: LocalStream) => {
```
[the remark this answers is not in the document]
mkesavan13: Will change

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -370,21 +380,23 @@ Media.toggleStream = () => {};
...
+  if (!stream) {
     return Promise.resolve();
   }
 
   return Promise.resolve().then(() => {
-    if (track && track.stop) {
+    if (stream && stream.stop) {
```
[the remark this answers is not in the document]
mkesavan13: Will update

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Ok. Will add a log for inputStream as well and update the track to stream. Probably a missed one while renaming.

packages/@webex/plugin-meetings/src/media/index.ts
sreenara: @mkesavan13 yes we don't have a plan to do that right now. We can come back to it at a later time if there is a use-case for it.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -956,36 +955,6 @@ const localMedia = {
   }
 }
 
-function handleTrackUnpublished(track, mediaElement, debugString) {
```
marcin-bazyl, line 959: I guess the published state change handler will be added back once the "published-state-update" event is added to the meeting object?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3348,193 +3351,177 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    localTrack?.on(LocalTrackEvents.Muted, this.localAudioTrackMuteStateHandler);
-    localTrack?.on(LocalTrackEvents.UnderlyingTrackChange, this.underlyingLocalTrackChangeHandler);
+    localStream?.on(StreamEventNames.MuteStateChange, this.localAudioStreamMuteStateHandler);
+    localStream?.on(LocalStreamEventNames.OutputTrackChange, this.localOutputTrackChangeHandler);
 
-    if (!this.isMultistream || !localTrack) {
-      // for multistream WCME automatically un-publishes the old track when we publish a new one
-      await this.unpublishTrack(oldTrack);
+    if (!localStream) {
+      await this.unpublishStream(MediaType.AudioMain);
```
marcin-bazyl, line 3397: as far as I can see, WCME does a replaceTrack() on a transceiver so it effectively unpublishes the old stream, so the comment from the old code (before this PR) still stands and we shouldn't be calling unpublishStream() on the send slot when we want to publish a different stream, just calling publishStream() with the new one should be enough

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5225,18 +5212,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      await this.publishStream(this.mediaProperties.videoStream, MediaType.VideoMain);
     }
-    if (this.mediaProperties.shareVideoTrack) {
-      await this.publishTrack(this.mediaProperties.shareVideoTrack);
+    if (this.mediaProperties.shareVideoStream) {
+      await this.publishStream(this.mediaProperties.shareVideoStream, MediaType.VideoSlides);
     }
-    if (this.isMultistream && this.mediaProperties.shareAudioTrack) {
-      await this.publishTrack(this.mediaProperties.shareAudioTrack);
+    if (this.mediaProperties.shareAudioStream) {
```
marcin-bazyl, line 5232: is share audio now supported for transcoded meetings? I thought it wasn't, hence we had the `(this.isMultistream)` condition here and now it's removed, is that correct?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3348,193 +3351,177 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      LocalTrackEvents.UnderlyingTrackChange,
-      this.underlyingLocalTrackChangeHandler
+    localDisplayStream?.on(StreamEventNames.Ended, this.handleShareVideoStreamEnded);
+    localDisplayStream?.on(
+      LocalStreamEventNames.OutputTrackChange,
+      this.localOutputTrackChangeHandler
     );
 
-    this.mediaProperties.mediaDirection.sendShare = this.mediaProperties.hasLocalShareTrack();
+    this.mediaProperties.mediaDirection.sendShare = !!localDisplayStream;
```
marcin-bazyl, line 3451: why are we not using hasLocalShareStream() here?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5507,8 +5501,8 @@ export default class Meeting extends StatelessWebexPlugin {
         })
       )
       .then(() => {
-        if (this.mediaProperties.hasLocalShareTrack()) {
-          return this.enqueueScreenShareFloorRequest();
+        if (localStreams?.screenShare?.video) {
```
marcin-bazyl, line 5504: this looks wrong, is this a merge gone wrong? we should be using `hasLocalShareStream()` and `enqueueScreenShareFloorRequest()` here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6038,16 +6026,11 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   private requestScreenShareFloor() {
-    if (
-      !this.mediaProperties.hasLocalShareTrack() ||
-      !this.mediaProperties.mediaDirection.sendShare
-    ) {
+    if (!this.mediaProperties.shareVideoStream || !this.mediaProperties.mediaDirection.sendShare) {
```
marcin-bazyl, line 6029: we should be using hasLocalShareStream() here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6865,10 +6847,14 @@ export default class Meeting extends StatelessWebexPlugin {
 
     return this.mediaProperties.webrtcMediaConnection
       .update({
+        // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now
         localTracks: {
-          audio: this.mediaProperties.audioTrack?.underlyingTrack || null,
-          video: this.mediaProperties.videoTrack?.underlyingTrack || null,
-          screenShareVideo: this.mediaProperties.shareVideoTrack?.underlyingTrack || null,
+          audio: this.mediaProperties.audioStream?.outputStream.getTracks()[0] || null,
```
marcin-bazyl, line 6852: I think it would be better to use `outputTrack` instead of `outputStream.getTracks()[0]`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6907,84 +6893,69 @@ export default class Meeting extends StatelessWebexPlugin {
...
-   * Publishes a track.
+   * Publishes a stream.
    *
-   * @param {LocalTrack} track to publish
+   * @param {LocalStream} stream to publish
+   * @param {MediaType} mediaType of the stream
    * @returns {Promise}
    */
-  private async publishTrack(track?: LocalTrack) {
-    if (!track) {
```
marcin-bazyl, line 6916: there are many possible cases where publishStream() will be called with stream being undefined, that's why the old code had that `if (!track)` check, same check for stream is needed in the new code (and type in function arguments also needs updating)

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6907,84 +6893,69 @@ export default class Meeting extends StatelessWebexPlugin {
...
       return;
     }
 
     let floorRequestNeeded = false;
 
-    if (tracks.screenShare?.video) {
-      await this.setLocalShareVideoTrack(tracks.screenShare?.video);
-
-      floorRequestNeeded = this.screenShareFloorState === ScreenShareFloorStatus.RELEASED;
+    if (streams.screenShare?.audio) {
```
marcin-bazyl, line 6943: this doesn't look right, see my other comments about share audio on transcoded meetings

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7000,31 +6971,40 @@ export default class Meeting extends StatelessWebexPlugin {
...
-   * @param {Array<MediaStreamTrack>} tracks
+   * @param {Array<LocalStream>} streams
    * @returns {Promise}
    */
-  async unpublishTracks(tracks: LocalTrack[]): Promise<void> {
+  async unpublishStreams(streams: LocalStream[]): Promise<void> {
     this.checkMediaConnection();
 
     const promises = [];
+    let isAudioReleased = false;
```
marcin-bazyl, line 6983: this should be named `isShareAudioReleased`, but overall the code using this variable doesn't seem right. What if share audio is in the `streams` list behind share video? or if the app calls unpublishStreams separately with just share audio and just share video?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    return slot;
+  }
+
+  /**
+   * This method is used to retrieve the sendSlot for the given mediaType
+   * @param {MediaType} mediaType of which the slot needs to be retrieved
+   * @returns {SendSlot}
+   */
+  public getSlot(mediaType: MediaType): SendSlot {
```
marcin-bazyl, line 48: this is not called from anywhere, so maybe not needed?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3348,193 +3351,177 @@ export default class Meeting extends StatelessWebexPlugin {
...
     this.mediaProperties.mediaDirection.sendVideo = false;
     this.mediaProperties.mediaDirection.sendShare = false;
 
-    // WCME doesn't unpublish tracks when multistream connection is closed, so we do it here
-    // (we have to do it for transcoded meetings anyway, so we might as well do for multistream too)
-    audioTrack?.setPublished(false);
-    videoTrack?.setPublished(false);
-    shareVideoTrack?.setPublished(false);
-    shareAudioTrack?.setPublished(false);
+    this.unpublishStreams([audioStream, videoStream, shareAudioStream, shareVideoStream]);
```
marcin-bazyl, line 3524: this won't work, because the tracks are already unset on the mediaProperties, so unpublishStreams won't do anything

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -0,0 +1,162 @@
...
+   * This method publishes the given stream to the sendSlot for the given mediaType
+   * @param {MediaType} mediaType MediaType of the sendSlot to which a stream needs to be published (AUDIO_MAIN/VIDEO_MAIN/AUDIO_SLIDES/VIDEO_SLIDES)
+   * @param {LocalStream} stream LocalStream to be published
+   * @returns {Promise<void>}
+   */
+  public async publishStream(mediaType: MediaType, stream: LocalStream): Promise<void> {
+    const slot = this.slots.get(mediaType);
+
+    if (!slot) {
+      throw new Error(`Slot for ${mediaType} does not exist`);
```
marcin-bazyl, line 68: I think we can end up with this error being thrown in many normal situations, for example see what happens in `Meeting.setLocalAudioStream()` when it's called with undefined from unpublishStreams().

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -544,14 +545,15 @@ export default class Meeting extends StatelessWebexPlugin {
...
-  underlyingLocalTrackChangeHandler: () => void;
+  localAudioStreamMuteStateHandler: (muted: boolean) => void;
+  localVideoStreamMuteStateHandler: (muted: boolean) => void;
+  localOutputTrackChangeHandler: () => void;
   roles: any[];
   environment: string;
   namespace = MEETINGS;
   annotationInfo: AnnotationInfo;
   allowMediaInLobby: boolean;
+  sendSlotManager: SendSlotManager = new SendSlotManager(LoggerProxy);
```
marcin-bazyl, line 556: I don't think we need to expose the sendSlotManager to the app, so this should probably be private

packages/@webex/plugin-meetings/src/media/properties.ts
marcin-bazyl: `shareAudioStream` and `shareVideoStream` are **local** webrtc-core streams, while `remoteShare` is for **remote** share stream - it's only used for transcoded meetings (for multistream we use receive slots and remote streams are managed by RemoteMediaManager), so because for now we don't support share audio in transcoded meetings, we only have 1 share remote stream - that's the remote video share

packages/@webex/plugin-meetings/src/media/index.ts
marcin-bazyl: Yes, the test needs to be updated. As you can se [here](https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/blob/8705925929ce9084b43416ad4682e23f1e14284e/src/MediaConnection/config.ts#L48) the MultistreamRoapMediaConnection's constructor takes MultistreamConnectionConfig that doesn't have these fields anymore (you've removed them in this [PR](https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/192/) )

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yeah I remember we decided that we'll always create the send slots, but the initial "active" value for them should be set correctly based on what the app passes to `addMedia()`. There is no reason to require the app to call `updateMedia()` later. The app will only call `updateMedia()` when it wants to change the values for audioEnabled/videoEnabled to somehing different than what was in the `addMedia()` call

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    await slot.publishStream(stream);
+
+    this.LoggerProxy.logger.info(
+      `SendSlotsManager->publishStream#Published stream for ${mediaType} and stream ${stream}`
```
marcin-bazyl, line 74: in the log the `${stream}` bit ends up as "[object Object]" which is not very useful, we should try to make it better and show inputTrack's id and label maybe? enabled and muted properties might also be useful

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: During my published-state-update event PR, I noticed these errors while testing in the samples app. I have updated some of these code as per the error and also checked the discussion above. Below is the PR with the changes: https://github.com/webex/webex-js-sdk/pull/3052

Will comment the PR link wherever I have already rectified it.

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6907,84 +6893,69 @@ export default class Meeting extends StatelessWebexPlugin {
...
-    }
-
-    if (this.mediaProperties.webrtcMediaConnection) {
-      if (this.isMultistream) {
-        await this.mediaProperties.webrtcMediaConnection.publishTrack(track);
-      } else {
-        track.setPublished(true); // for multistream, this call is done by WCME
-      }
+  private async publishStream(stream: LocalStream, mediaType: MediaType) {
+    if (this.isMultistream && this.mediaProperties.webrtcMediaConnection) {
```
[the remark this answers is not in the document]
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Sorry. These were cherry-picked changes from Bryce and so missed to notice. Thanks. Will update thet tests accordingly

packages/@webex/media-helpers/src/webrtc-core.ts
brycetham: Oh, that's a good question. Yeah it seems like we should, based on https://github.com/webex/ts-events#methodology.

Perhaps @bbaldino can explain better why this is necessary.

bbaldino: Yeah if you want the 'js-style' syntax (`class.on('eventName', handler)`) you'll want to declare the events as laid out in the ts-events README.  Using a raw `TypedEvent` gets you something like `class.event.on(handler)`

docs/samples/browser-plugin-meetings/app.js
mkesavan13: Thanks @meta-dipanshu-sharma

docs/samples/browser-plugin-meetings/app.js
mkesavan13: Thanks @meta-dipanshu-sharma

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/media/index.ts
marcin-bazyl: @sreenara so there is no plan to move calling SDK to use webrtc-core tracks at all?

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: in #3052 I can see the handler is back but it only calls stop(). Do we need to also call `clearVideoResolutionCheckInterval()` when main video is unpublished and reset srcObject properties on the audio/video elements? or is that done sowhere else now?

meta-dipanshu-sharma: For remote video, `clearVideoResolutionCheckInterval()` is called after meeting.leave() [here](https://github.com/webex/webex-js-sdk/blob/stream-classes/docs/samples/browser-plugin-meetings/app.js#L558). srcObjects are cleared during cleanUpMedia().

For local video, as we will use existing audio/video/share streams in another meeting so we don't have to call either of above methods. I can see local video and audio stream's srcObject is set to undefined but I will remove it once my older PR for effects will be applied on the new code.

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: Done: https://github.com/webex/webex-js-sdk/pull/3052

packages/@webex/plugin-meetings/src/media/index.ts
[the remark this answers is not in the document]
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/media/index.ts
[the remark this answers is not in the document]
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Ok. Makes sense. Will update.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Sorry. Must''ve come from [this commit](https://github.com/brycetham/webex-js-sdk/blob/26d52f3bdc66538b20bc7a2956deb291dbd1fadb/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3445) while solving merge conflicts.

Updated this in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: I agree. So, we should remove the following lines from cleanUpMedia as unpublish will do it for us. Right?

```
this.mediaProperties.setLocalAudioStream(undefined);
this.mediaProperties.setLocalVideoStream(undefined);
this.mediaProperties.setLocalShareAudioStream(undefined);
this.mediaProperties.setLocalShareVideoStream(undefined);
```

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Looks like a merge gone wrong. Sorry about that. Done in #3053

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
mkesavan13: We use this in tests now extensively as we do not stub the sendSlotManager, but rather the sendSlot itself. 

That said, we decided to have it there just in case in the future if we need to check if a slot is active or not, we shall getSlot and then check its `active` property. 

If this really needs to go away, we can make it go away with the `isActive(MediaType)` method but I'd like to take that as a TODO. We will have to fix tests as well if we are removing this now.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
mkesavan13: In that case, it calls `meeting.publishStream` which returns if there is no stream. So, I think we won't hit that situation. Anything else that would cause this?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
mkesavan13: I agree. Have added just stream's label and muted property as inputStream & inputTrack is protected and can't access from here.

Done in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Updated this and also added a comment above like this,

> //Screenshare Audio is supported only in multi stream. So we check for screenshare audio presence only if it's a multi stream meeting

Done in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done the same in #3053 

Kindly let me know if I have understood it incorrectly

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: > No @marcin-bazyl. I don't think we had that condition here

What do you mean we didn't have it? It's clearly visible in the diff in this PR:
image


and also in beta branch the publishTracks() method check for multisteam when processing share audio track, see [here](https://github.com/webex/webex-js-sdk/blob/beta/packages/%40webex/plugin-meetings/src/meeting/index.ts#L7064)

About the 2nd point of publish vs activating sendSlot:
Publication of a stream and the active state of a sendSlot are 2 different things. Publish track determines what track is set on the sender of a transceiver, while active flag determines the direction of that transceiver (sendrecv or inactive). So the way it worked with tracks was that you could publish a track (for example main audio), but if you called updateMedia({enableAudio: false}) then audio was not sent and not received even though the audio track's state was "published". The `enableAudio` flag in addMedia and updateMedia overrides the publish state - it's a more high level thing that says whether audio can be sent and received at all. Later if you call updateMedia({enableAudio: true}) then if you had any published audio track, it would start to be sent again. I think we should keep that same behavior with streams.

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: > For remote video, clearVideoResolutionCheckInterval() is called after meeting.leave() [here](https://github.com/webex/webex-js-sdk/blob/stream-classes/docs/samples/browser-plugin-meetings/app.js#L558). srcObjects are cleared during cleanUpMedia().

I'm only talking about local steams here, publishing is only for local streams not remote

> For local video, as we will use existing audio/video/share streams in another meeting

sorry I'm confused: in PR#3052 code was added to call stop() when stream is unpublished. You can't use a stopped stream for another meeting or anything else later. Once it's stopped, it cannot be resumed. 
Yes, in general with the webrtc-core track or stream classes you can reuse the same track/stream later, but that's not how the sample app works on beta branch right now. I agree it would be good to change this and re-use them, but that's a separate piece of work, let's not do it now as part of wcme 3.0 transition, because that in itself is big enough already.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: it's done incorrectly there, but I'll mark this comment here as resolved and we can continue the discusion in #3053

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
marcin-bazyl: ok, that's fine we can keep it

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: The changes in #3053 for this are not right. By calling unpublishStreams() you will potentially invoke code that does things like updating MuteState and releasing screen share floor, which we don't want to do. This function cleanupLocalTracks() is doing cleanup on meeting leave/destroy and just needs to clear all the references and not send any new requests to the backend. All you need to do is just put the code for cleanupLocalTracks() back to how it is on beta branch and rename all "Track" strings to "Stream". Why do you keep doing other changes?

marcin-bazyl: ah, wait, ok, there is one thing you'll have to change in cleanupLocalTracks() apart from renaming: the set of calls to `setPublished(false)` at the end, because we don't have that method anymore on the streams, so these will have to be replaced by code that emits the "meeting:streamPublishStateChanged" event with isPublished=false

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
marcin-bazyl: ok, yes, you're right, so I think it should be fine as it is.

packages/@webex/media-helpers/src/webrtc-core.ts
mkesavan13: So, do we need to change this to a Typed event?

@brycetham @marcin-bazyl

marcin-bazyl: we don't want to use just a raw TypedEvent, we want to be able to do class.on(event, handler), see README from ts-events lib and webrtc-core source code for an example how to do it, look at `_LocalStream` class - it extends Stream class and add more events to it, we are doing the same here, we're extending webrtc-core classes and add events to them.

packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -125,37 +125,38 @@ export class MuteState {
...
+   * Mutes/unmutes local stream
    *
    * @param {Object} meeting - the meeting object
-   * @param {Boolean} mute - true to mute the track, false to unmute it
+   * @param {Boolean} mute - true to mute the stream, false to unmute it
    * @param {ServerMuteReason} reason - reason for muting/unmuting
    * @returns {void}
    */
-  private muteLocalTrack(meeting: any, mute: boolean, reason: ServerMuteReason) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
marcin-bazyl, line 135: this doesn't seem to be needed for anything.... all vars are used in this method, aren't they? Also having this eslint comment here causes a warning about jsdocs missing.

mkesavan13: Done in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Oops. My bad. For unpublish, I misunderstood between a client unpublish and server I believe as we had this comment earlier,
> // WCME doesn't unpublish tracks when multistream connection is closed, so we do it here
> // (we have to do it for transcoded meetings anyway, so we might as well do for multistream too)

Fixed it in #3053

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: @marcin-bazyl - Understood. What do we do in the Video and Audio Share in that case? I see the mediaDirection for share is set to false by default. Should we create a disabled SendSlot for Share in that case?

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: > so these will have to be replaced by code that emits the "meeting:streamPublishStateChanged" event with isPublished=false

When you said so, do you mean I have to trigger an event like this?

```
Trigger.trigger(
      this,
      {
        file: 'meeting/index',
        function: 'publishStream',
      },
      EVENT_TRIGGERS.MEETING_STREAM_PUBLISH_STATE_CHANGED,
      {
        isPublished: false,
        mediaType,
        stream,
     }
);

File: docs/samples/browser-plugin-meetings/app.js (hunk 12 above)
Comment (mkesavan13) [reply docs/samples/browser-plugin-meetings/app.js:959]: Done in #3053

File: packages/@webex/media-helpers/src/webrtc-core.ts (hunk 4 above)
Comment (mkesavan13) [reply packages/@webex/media-helpers/src/webrtc-core.ts:46]: Done in #3053

File: packages/@webex/plugin-meetings/src/media/index.ts (hunk 6 above)
Comment (marcin-bazyl) [reply packages/@webex/plugin-meetings/src/media/index.ts:212]: ok, great, but if you are planning to do it, surely you have a jira :-) can you reference it here in the todo comment?

File: packages/@webex/plugin-meetings/src/meeting/index.ts (hunk 23 above)
Comment (marcin-bazyl) [reply packages/@webex/plugin-meetings/src/meeting/index.ts:3528]: yes, send such event for all streams that were set (they were not undefined)

File: packages/@webex/plugin-meetings/src/media/index.ts (hunk 6 above)
Comment (sreenara) [reply packages/@webex/plugin-meetings/src/media/index.ts:212]: Created [SPARK-460530](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-460530) to do this in the future once we get some bandwidth.

File: packages/@webex/plugin-meetings/src/meeting/index.ts (hunk 23 above)
Comment (mkesavan13) [reply packages/@webex/plugin-meetings/src/meeting/index.ts:3528]: Done in #3053 

Have written [a separate method to emit this event](https://github.com/mkesavan13/webex-js-sdk/blob/wcme3-update-importExports-variablesMethods/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3497) since the code was repetitive and also updated the changes from Dipanshu to use this method.

File: packages/@webex/plugin-meetings/src/media/index.ts (hunk 6 above)
Comment (marcin-bazyl) [reply packages/@webex/plugin-meetings/src/media/index.ts:212]: thanks, @mkesavan13 can you reference [SPARK-460530](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-460530) in this todo comment, please?

Comment (mkesavan13) [reply packages/@webex/plugin-meetings/src/media/index.ts:212]: Done in #3053

File: packages/@webex/plugin-meetings/src/meeting/index.ts (hunk 14 above)
Comment (mkesavan13) [reply packages/@webex/plugin-meetings/src/meeting/index.ts:5236]: As per discussions on Webex, made this change so there's no more `receiveShare` in `AddMediaOptions`

Instead, we have introduced `enableShareAudio` and `enableShareVideo` 

`mediaDirection.receiveShare` will be true if one of the above is `true`

Done in #3053

File: packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js [hunk 29]
```diff
@@ -1623,438 +1655,421 @@ describe('plugin-meetings', () => {
...
-                  preferTranscoding: !meeting.isMultistream,
-                  ipver: 0,
+          assert.calledWith(locusMediaRequestStub, {
+            method: 'PUT',
+            uri: `${meeting.selfUrl}/media`,
+            body: {
+              device: {
+                url: meeting.deviceUrl,
+                deviceType: meeting.config.deviceType,
+                regionCode: undefined,
```
marcin-bazyl, line 1818: why are regionCode and countryCode expected values undefined now? (here and in lines 1845-1846)

mkesavan13: Done @marcin-bazyl

arun3528:
merged it after taking with the team

PR #3053, opened 2023-09-05
fix(plugin-meetings): stream-classes refactor and tests

mkesavan13:
# COMPLETES [SPARK-455627](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-455627)

The fix for the tests that are broken due to stream-classes refactor

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/index.ts
+++ b/packages/@webex/plugin-meetings/src/media/index.ts
@@ -19,13 +19,8 @@ import BrowserDetection from '../common/browser-detection';
 
 const {isBrowser} = BrowserDetection();
 
-// TODO: in internal-media-core add appropriate types in MultistreamConnectionConfig for enable main audio and video
-type MultistreamConnectionConfig = ConstructorParameters<
-  typeof MultistreamRoapMediaConnection
->[0] & {
-  enableMainAudio: boolean;
-  enableMainVideo: boolean;
-};
+// TODO: Export MultiStreamConnectionConfig and update here: SPARK-458871
+type MultistreamConnectionConfig = ConstructorParameters<typeof MultistreamRoapMediaConnection>[0];
 
 export type BundlePolicy = ConstructorParameters<
   typeof MultistreamRoapMediaConnection
@@ -170,10 +165,6 @@ Media.createMediaConnection = (
   if (isMultistream) {
     const config: MultistreamConnectionConfig = {
       iceServers,
-      enableMainAudio:
-        mediaProperties.mediaDirection?.sendAudio || mediaProperties.mediaDirection?.receiveAudio,
-      enableMainVideo:
-        mediaProperties.mediaDirection?.sendVideo || mediaProperties.mediaDirection?.receiveVideo,
     };
 
     if (bundlePolicy) {
@@ -209,13 +200,13 @@ Media.createMediaConnection = (
       },
     },
     {
-      // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now
+      // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now SPARK-460530
       localTracks: {
         audio: audioStream?.outputTrack,
         video: videoStream?.outputTrack,
         screenShareVideo: shareVideoStream?.outputTrack,
-        screenShareAudio: shareAudioStream?.outputTrack, // TODO: add type for screenShareAudio in internal-media-core
-      } as unknown, // TODO: add type for screenShareAudio in internal-media-core
+        screenShareAudio: shareAudioStream?.outputTrack, // TODO: add type for screenShareAudio in internal-media-core SPARK-446923
+      } as unknown,
       direction: {
         audio: Media.getDirection(true, mediaDirection.receiveAudio, mediaDirection.sendAudio),
         video: Media.getDirection(true, mediaDirection.receiveVideo, mediaDirection.sendVideo),
@@ -383,20 +374,18 @@ Media.toggleStream = () => {};
  * @param {LocalStream} stream A local stream
  * @returns {null}
  */
-Media.stopStreams = (stream: LocalStream) => {
+Media.stopStream = (stream: LocalStream) => {
   if (!stream) {
     return Promise.resolve();
   }
 
   return Promise.resolve().then(() => {
-    if (stream && stream.stop) {
-      try {
-        stream.stop();
-      } catch (e) {
-        LoggerProxy.logger.error(
-          `Media:index#stopTracks --> Unable to stop the track with state ${stream.outputTrack.readyState}, error: ${e}`
-        );
-      }
+    try {
+      stream.stop();
+    } catch (e) {
+      LoggerProxy.logger.error(
+        `Media:index#stopStream --> Unable to stop the stream with ready state of the output track => ${stream.outputTrack.readyState} & input track => ${stream.inputTrack.readyState}, error: ${e}`
+      );
     }
   });
 };
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -158,7 +158,8 @@ export type AddMediaOptions = {
   localStreams?: LocalStreams;
   audioEnabled?: boolean; // if not specified, default value true is used
   videoEnabled?: boolean; // if not specified, default value true is used
-  receiveShare?: boolean; // if not specified, default value true is used
+  enableShareAudio?: boolean; // if not specified, default value true is used
+  enableShareVideo?: boolean; // if not specified, default value true is used
   remoteMediaManagerConfig?: RemoteMediaManagerConfiguration; // applies only to multistream meetings
   bundlePolicy?: BundlePolicy; // applies only to multistream meetings
   allowMediaInLobby?: boolean; // allows adding media when in the lobby
@@ -553,7 +554,7 @@ export default class Meeting extends StatelessWebexPlugin {
   namespace = MEETINGS;
   annotationInfo: AnnotationInfo;
   allowMediaInLobby: boolean;
-  sendSlotManager: SendSlotManager = new SendSlotManager(LoggerProxy);
+  private sendSlotManager: SendSlotManager = new SendSlotManager(LoggerProxy);
 
   /**
    * @param {Object} attrs
@@ -3326,7 +3327,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   closeRemoteStreams() {
-    const {remoteAudioStream, remoteVideoStream, remoteShare, shareAudioStream} =
+    const {remoteAudioStream, remoteVideoStream, remoteShareStream, shareAudioStream} =
       this.mediaProperties;
 
     /**
@@ -3359,7 +3360,7 @@ export default class Meeting extends StatelessWebexPlugin {
      */
     // eslint-disable-next-line arrow-body-style
     const stopStream = (stream: RemoteStream, type: string) => {
-      return Media.stopStreams(stream).then(() => {
+      return Media.stopStream(stream).then(() => {
         triggerMediaStoppedEvent(type);
       });
     };
@@ -3367,7 +3368,7 @@ export default class Meeting extends StatelessWebexPlugin {
     return Promise.all([
       stopStream(remoteAudioStream, EVENT_TYPES.REMOTE_AUDIO),
       stopStream(remoteVideoStream, EVENT_TYPES.REMOTE_VIDEO),
-      stopStream(remoteShare, EVENT_TYPES.REMOTE_SHARE),
+      stopStream(remoteShareStream, EVENT_TYPES.REMOTE_SHARE),
       stopStream(shareAudioStream, EVENT_TYPES.REMOTE_SHARE_AUDIO),
     ]);
   }
@@ -3450,7 +3451,7 @@ export default class Meeting extends StatelessWebexPlugin {
       this.localOutputTrackChangeHandler
     );
 
-    this.mediaProperties.mediaDirection.sendShare = !!localDisplayStream;
+    this.mediaProperties.mediaDirection.sendShare = this.mediaProperties.hasLocalShareStream();
 
     if (!this.isMultistream || !localDisplayStream) {
       // for multistream WCME automatically un-publishes the old stream when we publish a new one
@@ -3469,12 +3470,12 @@ export default class Meeting extends StatelessWebexPlugin {
   private async setLocalShareAudioStream(localSystemAudioStream?: LocalSystemAudioStream) {
     const oldStream = this.mediaProperties.shareAudioStream;
 
-    oldStream?.off(LocalStreamEventNames.Ended, this.handleShareAudioStreamEnded);
+    oldStream?.off(StreamEventNames.Ended, this.handleShareAudioStreamEnded);
     oldStream?.off(LocalStreamEventNames.OutputTrackChange, this.localOutputTrackChangeHandler);
 
-    this.mediaProperties.setLocalShareAudioStream(LocalSystemAudioStream);
+    this.mediaProperties.setLocalShareAudioStream(localSystemAudioStream);
 
-    localSystemAudioStream?.on(LocalStreamEventNames.Ended, this.handleShareAudioStreamEnded);
+    localSystemAudioStream?.on(StreamEventNames.Ended, this.handleShareAudioStreamEnded);
     localSystemAudioStream?.on(
       LocalStreamEventNames.OutputTrackChange,
       this.localOutputTrackChangeHandler
@@ -3487,6 +3488,34 @@ export default class Meeting extends StatelessWebexPlugin {
       await this.unpublishStream(MediaType.AudioSlides, oldStream);
     }
     await this.publishStream(MediaType.AudioSlides, this.mediaProperties.shareAudioStream);
+  }
+
+  /**
+   * Handles the local audio stream publish state change event
+   * @internal
+   * @param {Object} options parameters functionName, isPublished, mediaType and stream needed to trigger event
+   * @returns {undefined}
+   */
+  private emitPublishStateChangeEvent(options: {
+    functionName: string;
+    isPublished: boolean;
+    mediaType: MediaType;
+    stream: MediaStream;
+  }) {
+    const {functionName, isPublished, mediaType, stream} = options;
+    Trigger.trigger(
+      this,
+      {
+        file: 'meeting/index',
+        function: functionName,
+      },
+      EVENT_TRIGGERS.MEETING_STREAM_PUBLISH_STATE_CHANGED,
+      {
+        isPublished,
+        mediaType,
+        stream,
+      }
+    );
   }
 
   /**
@@ -3525,7 +3554,38 @@ export default class Meeting extends StatelessWebexPlugin {
     this.mediaProperties.mediaDirection.sendVideo = false;
     this.mediaProperties.mediaDirection.sendShare = false;
 
-    this.unpublishStreams([audioStream, videoStream, shareAudioStream, shareVideoStream]);
+    if (audioStream) {
+      this.emitPublishStateChangeEvent({
+        functionName: 'cleanupLocalStreams',
+        isPublished: false,
+        mediaType: MediaType.AudioMain,
+        stream: audioStream,
+      });
+    }
+    if (videoStream) {
+      this.emitPublishStateChangeEvent({
+        functionName: 'cleanupLocalStreams',
+        isPublished: false,
+        mediaType: MediaType.VideoMain,
+        stream: videoStream,
+      });
+    }
+    if (shareVideoStream) {
+      this.emitPublishStateChangeEvent({
+        functionName: 'cleanupLocalStreams',
+        isPublished: false,
+        mediaType: MediaType.VideoSlides,
+        stream: shareVideoStream,
+      });
+    }
+    if (shareAudioStream) {
+      this.emitPublishStateChangeEvent({
+        functionName: 'cleanupLocalStreams',
+        isPublished: false,
+        mediaType: MediaType.AudioSlides,
+        stream: shareAudioStream,
+      });
+    }
   }
 
   /**
@@ -4932,7 +4992,7 @@ export default class Meeting extends StatelessWebexPlugin {
             break;
           case RemoteTrackType.SCREENSHARE_VIDEO:
             eventType = EVENT_TYPES.REMOTE_SHARE;
-            this.mediaProperties.setRemoteShare(remoteStream);
+            this.mediaProperties.setRemoteShareStream(remoteStream);
             break;
           default: {
             LoggerProxy.logger.log(
@@ -5198,6 +5258,7 @@ export default class Meeting extends StatelessWebexPlugin {
    *
    * @param {Object} turnServerInfo TURN server information
    * @param {BundlePolicy} [bundlePolicy] Bundle policy settings
+   * @param {AddMediaOptions} [options] Options for enabling/disabling audio/video
    * @returns {RoapMediaConnection | MultistreamRoapMediaConnection}
    */
   private async createMediaConnection(turnServerInfo, bundlePolicy?: BundlePolicy) {
@@ -5217,10 +5278,19 @@ export default class Meeting extends StatelessWebexPlugin {
     this.setupMediaConnectionListeners();
 
     if (this.isMultistream) {
-      this.sendSlotManager.createSlot(mc, MediaType.VideoMain, false);
-      this.sendSlotManager.createSlot(mc, MediaType.AudioMain, false);
-      this.sendSlotManager.createSlot(mc, MediaType.VideoSlides, false);
-      this.sendSlotManager.createSlot(mc, MediaType.AudioSlides, false);
+      const [audioEnabled, videoEnabled, shareEnabled] = [
+        this.mediaProperties.mediaDirection.sendAudio ||
+          this.mediaProperties.mediaDirection.receiveAudio,
+        this.mediaProperties.mediaDirection.sendVideo ||
+          this.mediaProperties.mediaDirection.receiveVideo,
+        this.mediaProperties.mediaDirection.sendShare ||
+          this.mediaProperties.mediaDirection.receiveShare,
+      ];
+
+      this.sendSlotManager.createSlot(mc, MediaType.VideoMain, audioEnabled);
+      this.sendSlotManager.createSlot(mc, MediaType.AudioMain, videoEnabled);
+      this.sendSlotManager.createSlot(mc, MediaType.VideoSlides, shareEnabled);
+      this.sendSlotManager.createSlot(mc, MediaType.AudioSlides, shareEnabled);
     }
 
     // publish the streams
@@ -5233,7 +5303,7 @@ export default class Meeting extends StatelessWebexPlugin {
     if (this.mediaProperties.shareVideoStream) {
       await this.publishStream(MediaType.VideoSlides, this.mediaProperties.shareVideoStream);
     }
-    if (this.mediaProperties.shareAudioStream) {
+    if (this.isMultistream && this.mediaProperties.shareAudioStream) {
       await this.publishStream(MediaType.AudioSlides, this.mediaProperties.shareAudioStream);
     }
 
@@ -5291,7 +5361,8 @@ export default class Meeting extends StatelessWebexPlugin {
       localStreams,
       audioEnabled = true,
       videoEnabled = true,
-      receiveShare = true,
+      enableShareAudio = true,
+      enableShareVideo = true,
       remoteMediaManagerConfig,
       bundlePolicy,
       allowMediaInLobby,
@@ -5337,7 +5408,7 @@ export default class Meeting extends StatelessWebexPlugin {
       sendShare: false,
       receiveAudio: audioEnabled,
       receiveVideo: videoEnabled,
-      receiveShare,
+      receiveShare: enableShareAudio || enableShareVideo,
     });
 
     this.locusMediaRequest = new LocusMediaRequest(
@@ -5505,8 +5576,8 @@ export default class Meeting extends StatelessWebexPlugin {
         })
       )
       .then(() => {
-        if (localStreams?.screenShare?.video) {
-          this.enqueueMediaUpdate(MEDIA_UPDATE_TYPE.SHARE_FLOOR_REQUEST);
+        if (this.mediaProperties.hasLocalShareStream()) {
+          return this.enqueueScreenShareFloorRequest();
         }
 
         return Promise.resolve();
@@ -5712,11 +5783,12 @@ export default class Meeting extends StatelessWebexPlugin {
   public async updateMedia(options: {
     audioEnabled?: boolean;
     videoEnabled?: boolean;
-    receiveShare?: boolean;
+    shareAudioEnabled?: boolean;
+    shareVideoEnabled?: boolean;
   }) {
     this.checkMediaConnection();
 
-    const {audioEnabled, videoEnabled, receiveShare} = options;
+    const {audioEnabled, videoEnabled, shareAudioEnabled, shareVideoEnabled} = options;
 
     LoggerProxy.logger.log(
       `Meeting:index#updateMedia --> called with options=${JSON.stringify(options)}`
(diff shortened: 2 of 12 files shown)


packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1647,6 +1648,10 @@ describe('plugin-meetings', () => {
             .stub()
             .resolves({turnServerInfo: {}, turnDiscoverySkippedReason: 'reachability'});
 
+          meeting.sendSlotManager = new SendSlotManager({logger: {info: () => {}}});
```
marcin-bazyl, line 1651: this code is inside the describe "addMedia/updateMedia semi-integration tests", these are not typical unit tests where everything is mocked. In these tests only external interfaces like internal-media-core are mocked, so we shouldn't be mocking the SDK's SendSlotManager, but instead mock only the createSendSlot() method of MultistreamRoapMediaConnection to return fake send slots and the tests should check that we call the right methods on these WCME send slots

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6942,6 +6942,8 @@ export default class Meeting extends StatelessWebexPlugin {
 
     if (streams.screenShare?.audio) {
       await this.setLocalShareAudioStream(streams.screenShare.audio);
+
+      floorRequestNeeded = this.screenShareFloorState === ScreenShareFloorStatus.RELEASED;
```
marcin-bazyl, line 6946: good that tests have caught that, I've also seen this issue and commented about it in https://github.com/webex/webex-js-sdk/pull/3038

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2029,16 +2064,25 @@ describe('plugin-meetings', () => {
                   muted: false,
                   underlyingTrack: webrtcAudioTrack2
                 };
+                const fakeMicrophoneStream = {
```
marcin-bazyl, line 2067: this needs to be renamed, because it overshadows `fakeMicrophoneStream` that you've already added above. The existing code in this test is calling first publishTracks with `fakeMicrophoneTrack` and then calls it again with a different track `fakeMicrophoneTrack2 `. When you change it to streams, you have to keep that logic, so you need `fakeMicrophoneStream2`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2029,16 +2064,25 @@ describe('plugin-meetings', () => {
...
+                  off: sinon.stub(),
+                  outputStream:{
+                    getTracks: () => {
+                      return [fakeMicrophoneTrack2]
+                    }
+                  }
+                }
 
-                await meeting.publishTracks({microphone: fakeMicrophoneTrack2});
+                await meeting.publishStreams({microphone: fakeMicrophoneStream});
```
marcin-bazyl, line 2077: wrong stream is used here, see my previous comment

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2029,16 +2064,25 @@ describe('plugin-meetings', () => {
...
                 await stableState();
 
                 // only the roap media connection should be updated
                 if (isMultistream) {
-                  assert.calledOnceWithExactly(fakeMultistreamRoapMediaConnection.publishTrack, fakeMicrophoneTrack2);
+                  assert.calledOnceWithExactly(meeting.sendSlotManager.publishStream, MediaType.AudioMain, fakeMicrophoneStream);
```
marcin-bazyl, line 2082: in these semi-integration tests we should be checking that correct methods on fakeMultistreamRoapMediaConnection or fake send slots are called, not on the SDK's sendSlotManager

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2029,16 +2064,25 @@ describe('plugin-meetings', () => {
                   muted: false,
                   underlyingTrack: webrtcAudioTrack2
                 };
+                const fakeMicrophoneStream = {
+                  on: sinon.stub(),
+                  off: sinon.stub(),
+                  outputStream:{
+                    getTracks: () => {
+                      return [fakeMicrophoneTrack2]
```
marcin-bazyl, line 2072: fakeMicrophoneTrack2 is being used here as a webrtc raw track, so I think it would be better to rename it so that it's clear and also we should update it to reflect the interface of MediaStreamTrack, so for example methods like on/off, setUnmuteAllowed, setPublished etc should be removed from it

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2052,23 +2096,23 @@ describe('plugin-meetings', () => {
...
                 resetHistory();
 
-                await meeting.unpublishTracks([fakeMicrophoneTrack]);
+                await meeting.unpublishStreams([fakeMicrophoneStream]);
                 await stableState();
 
                 // the roap media connection should be updated
                 if (isMultistream) {
-                  assert.calledOnceWithExactly(fakeMultistreamRoapMediaConnection.unpublishTrack, fakeMicrophoneTrack);
+                  assert.calledOnceWithExactly(meeting.sendSlotManager.unpublishStream, MediaType.AudioMain);
```
marcin-bazyl, line 2112: ideally we should be checking that unpublishStream on the right sendSlot is called here, not that sendSlotManager is called

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2091,19 +2135,19 @@ describe('plugin-meetings', () => {
...
+            await meeting.addMedia({audioEnabled: enableMedia, localStreams: {microphone: stream}});
             await simulateRoapOffer();
 
             resetHistory();
           }
 
           const checkAudioEnabled = (expectedTrack, expectedDirection) => {
             if (isMultistream) {
-              assert.calledOnceWithExactly(fakeMultistreamRoapMediaConnection.enableMultistreamAudio, expectedDirection !== 'inactive');
+              assert.calledOnceWithExactly(meeting.sendSlotManager.setActive, MediaType.AudioMain, expectedDirection !== 'inactive');
```
marcin-bazyl, line 2147: same as above, would be better to check setActive being called on the right sendSlot

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2481,6 +2528,13 @@ describe('plugin-meetings', () => {
             receiveVideo: true,
             receiveShare: true,
           }
+          meeting.sendSlotManager = new SendSlotManager({
```
marcin-bazyl, line 2531: why is this needed? doesn't the meeting already have a SendSlotManager instance? it should be created when the meeting object is created, right?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3802,281 +3906,292 @@ describe('plugin-meetings', () => {
...
+              info: () => undefined
+            }
           });
+          fakeMultistreamRoapMediaConnection = {
+            createSendSlot: sinon.stub().returns({
+              publishStream: sinon.stub(),
+              unpublishStream: sinon.stub(),
+            })
+          }
+          meeting.sendSlotManager.createSlot(fakeMultistreamRoapMediaConnection, MediaType.VideoSlides);
```
marcin-bazyl, line 3923: this doesn't feel right..... in all the unit tests in this file we're using same `meeting` object and creating slots here will impact all other tests. we should probably use sinon to just mock the right wcme send slots and then restore those mocks or if this turns out to be difficult than at least undo the creation of the slot in `afterEach()`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3802,281 +3906,292 @@ describe('plugin-meetings', () => {
...
+          LocalDisplayStreamConstructorStub = sinon
+            .stub(InternalMediaCoreModule, 'LocalDisplayStream')
             .callsFake((stream) => {
-              fakeLocalDisplayTrack = createFakeLocalTrack(stream.getTracks()[0]);
-              return fakeLocalDisplayTrack;
+              return stream
             });
-          LocalMicrophoneTrackConstructorStub = sinon
-            .stub(InternalMediaCoreModule, 'LocalMicrophoneTrack')
+          LocalSystemAudioStreamConstructorStub = sinon
```
marcin-bazyl, line 3935: I'm slightly confused, this is something I don't understand: why this PR shows that you had to **add** things for system audio stream? is the stream-classes branch based on SDK beta version that already has share audio? If yes, then in this case these tests should already take share audio into account and you would only need to **modify** them not add new ones.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4086,10 +4201,15 @@ describe('plugin-meetings', () => {
     describe('#enableMusicMode', () => {
       beforeEach(() => {
         meeting.isMultistream = true;
-          meeting.mediaProperties.webrtcMediaConnection = {
-            setCodecParameters: sinon.stub().resolves({}),
-            deleteCodecParameters: sinon.stub().resolves({}),
-          };
+        meeting.sendSlotManager = new SendSlotManager({logger: { info: () => {}}});
+        meeting.sendSlotManager.createSlot({
```
marcin-bazyl, line 4205: again we're creating a send slot - see my previous comment about making sure this doesn't impact other tests

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -17,11 +17,17 @@ describe('plugin-meetings', () => {
...
   const createFakeLocalStream = (id, muted) => {
     return {
-      id,
+      setServerMuted: sinon.spy(),
+      setUnmuteAllowed: sinon.spy(),
       setMuted: sinon.stub(),
-      setServerMuted: sinon.stub(),
-      setUnmuteAllowed: sinon.stub(),
-      muted,
+      outputStream: {
```
marcin-bazyl, line 23: I'm not sure if this is correct. MuteState reads the muted property of the stream in the MuteState.init() method and webrtc-core implementation for "muted" getter reads the "enabled" property of the inputTrack while you're setting it as the "muted" property of the outputTrack, so both the object and property seem to be wrong. This affects the setup of the tests, so once this is fixed maybe you will not need some of the other true->false and false->true changes that you did in the tests

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -19,13 +19,7 @@ import BrowserDetection from '../common/browser-detection';
...
 const {isBrowser} = BrowserDetection();
 
-// TODO: in internal-media-core add appropriate types in MultistreamConnectionConfig for enable main audio and video
-type MultistreamConnectionConfig = ConstructorParameters<
-  typeof MultistreamRoapMediaConnection
->[0] & {
-  enableMainAudio: boolean;
-  enableMainVideo: boolean;
-};
+type MultistreamConnectionConfig = ConstructorParameters<typeof MultistreamRoapMediaConnection>[0];
```
marcin-bazyl, line 22: ideally we should just export MultistreamConnectionConfig from internal-media-core so that we wouldn't have to do this hack here

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Thanks @marcin-bazyl 
I just noticed your comment. But yes, identified and fixed this yes. Thanks again :-)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: @marcin-bazyl - We do have [tests to fake the sendSlots in the SendSlotManager tests](https://github.com/webex/webex-js-sdk/blob/stream-classes/packages/%40webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts) anyway. That is why we just mock the sendSlotManager. Should we still do this?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1613,7 +1614,7 @@ describe('plugin-meetings', () => {
           getSettings: sinon.stub().returns({deviceId: 'fake device id for audio track'}),
         };
 
-        let fakeMicrophoneTrack;
+        let fakeMicrophoneTrack, fakeMicrophoneStream;
```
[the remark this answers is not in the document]
mkesavan13: Done. Removed and adjusted tests accordingly. Was having it as a reference to test Transcoded meetings but not necessary. We can extract track from the stream itself. Thanks

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: As mentioned in one of the previous comments, have removed `fakeMicrophoneTrack2` & `fakeMicrophoneTrack` and instead set a sample track with ID since the test's aim is to check if the track from the stream is properly sent to the transcoded meeting `RoapMediaConnection.update`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: We do check if the unpublishStream is called with the right MediaType in this line itself.

As already mentioned in one of the previous comments, sendSlotManger has its own tests to know if the methods of the right sendSlot is being called. Do we still need to stub one step deeper?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Commented above

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Makes sense. Removed all creation of SendSlotManager. 

In fact we also have tests to ensure that the SendSlotManager is created as part of meeting creation. Thanks for pointing out.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Agreed. Added `afterEach` routine with each `createSendSlot` in `beforeEach`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: In the beta, I do not see a stub for the Share Audio track: https://github.com/webex/webex-js-sdk/blob/35b946c2ef436248073043eb73c79e3af6e25d81/packages/%40webex/plugin-meetings/test/unit/spec/meeting/index.js

Am I missing something?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Addressed by making changes to this and responding to previous comment as well

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -120,6 +126,8 @@ describe('plugin-meetings', () => {
 
       // now simulate server requiring us to locally unmute
       audio.handleServerLocalUnmuteRequired(meeting);
+      meeting.mediaProperties.audioStream.setServerMuted(false,'localUnmuteRequired');
```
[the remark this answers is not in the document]
mkesavan13: Sorry. I've forgotten to remove this while cleaning up. I should have added that to test something and missed to remove. It was before I understood it was my stub mistake in muteState and not because we flipped the default mute value. Thanks for noticing it.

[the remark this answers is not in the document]
mkesavan13: Removed this now

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: I agree. Turns out the muteState reads the `muted` of the fakeLocalTrack created which is not actual webrtc-core object. So, the getter wasn't invoked. 

Removing the outputStream felt right. But do we want to make this createFakeLocalTrack to return an actual LocalTrack?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: yes, I would rather not mock sendSlotManager in these tests.

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: I agree. Have added a Todo along with a Jira to track it. Will update.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Have commented about it [here](https://github.com/webex/webex-js-sdk/pull/3053#discussion_r1319522326).

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2481,6 +2528,13 @@ describe('plugin-meetings', () => {
             receiveVideo: true,
             receiveShare: true,
           }
+          meeting.sendSlotManager = new SendSlotManager({
+            logger:{
+              info: () => undefined
+            }
+          });
+          meeting.sendSlotManager.createSlot({createSendSlot: sinon.stub().returns({active: true})},MediaType.AudioMain);
+          meeting.sendSlotManager.setActive = sinon.stub();
```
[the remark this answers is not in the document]
mkesavan13: Understood. Updated all the stubs to be created as `sinon.stub(object,'methodName')`

Thanks

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Done. Removed mocking of sendSlotManager and mocked Slots everywhere. Please check.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: looking at this a bit more, I think these constructor stubs are actually not needed at all, because these tests are using fake track (or now stream) objects like audioTrack, videoTrack, audioShareTrack, videoShareTrack created at the beginning of beforeEach(), so the constructors are never called. If I remove these constructor stubs in beta branch, these tests still pass, so you can just remove them

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
marcin-bazyl: do you need an actual LocalTrack for anything? I don't mind it much, so do what is easier. Usually with a fake one it's easier to control how it behaves, but you can use a real one and just stub specific methods on it too. If you use a real one then the name "createFake..." should probably be changed, because it won't be fake anymore

mkesavan13: I agree. We actually don't need an actual LocalTrack. So let this be like this. We will revisit in future if required

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -383,20 +374,21 @@ Media.toggleStream = () => {};
...
-        );
-      }
+    try {
+      stream.stop();
+    } catch (e) {
+      LoggerProxy.logger.error(
+        `Media:index#stopTracks --> Unable to stop the output stream with state ${stream.outputTrack.readyState}, error: ${e}`
+      );
+      LoggerProxy.logger.error(
+        `Media:index#stopTracks --> Unable to stop the input stream with state ${stream.inputTrack.readyState}, error: ${e}`
```
marcin-bazyl, line 390: very minor, but it would be nicer for the 2 logs to say which one is for inputTrack and which one for outputTrack, or it could be even all in a single log

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6030,7 +6025,10 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   private requestScreenShareFloor() {
-    if (!this.mediaProperties.shareVideoStream || !this.mediaProperties.mediaDirection.sendShare) {
+    if (
+      this.mediaProperties.hasLocalShareStream() ||
```
marcin-bazyl, line 6029: this is the opposite of what it should be, it should be `!this.mediaProperties.hasLocalShareStream()` - this should be caught by unit tests for requestScreenShareFloor like this one [here](https://github.com/webex/webex-js-sdk/blob/b58650c19f200e888239d496f2557d443b945f0b/packages/%40webex/plugin-meetings/test/unit/spec/meeting/index.js#L2557). Is it passing? if it is, can you check why?

mkesavan13: You're right! The test did fail. Have fixed it now. Thanks.

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: It is already separated out as `input stream` and `output stream` but will make it one log yes.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Done. Removed and tests are passing. Thanks.

packages/@webex/plugin-meetings/src/media/index.ts
marcin-bazyl: ah, sorry, yeah I didn't notice that. Thanks for combining it into a single log

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -361,23 +364,23 @@ describe('plugin-meetings', () => {
 
       it('does not send remote mute for video', async () => {
         // mute
-        await simulateVideoMuteChange(true);
+        await simulateVideoMuteChange(false);
```
marcin-bazyl, line 367: why are the values still reversed in all these tests in this PR? now that you've fixed the mock stream I don't think you need to change them, so it would be better to not touch them in this PR and keep them as they were, because for example you didn't update all the comments (like in the line above and other places).

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -569,7 +571,7 @@ describe('plugin-meetings', () => {
               await setup(mediaType, false, false);
               muteState.ignoreMuteStateChange = true;
 
-              muteState.handleLocalStreamMuteStateChange(meeting, true);
+              muteState.handleLocalStreamMuteStateChange(meeting, false);
```
marcin-bazyl, line 574: same here, this value should not have been reversed, it should remain as true. With value false, this test makes no sense, because we're not actually changing the mute value (it's initialized to false), so if we comment out the code from line 572, the test still passes, which defeats the point of this test that's supposed to check that when `ignoreMuteStateChange` is true then we don't call `MeetingUtil.remoteUpdateAudioVideo`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1606,14 +1605,9 @@ describe('plugin-meetings', () => {
          to @webex/internal-media-core when addMedia, updateMedia, publishTracks, unpublishTracks are called
          in various combinations.
       */
-      [true,false].forEach((isMultistream) =>
+      [true/false].forEach((isMultistream) =>
```
marcin-bazyl, line 1608: why this change? `true/false` actually evaluates to infinity, which is probably not what we want here....

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: actually I think I was wrong, because the meeting object is created in the beforeEach() so a new one will be created for each test, so these sendSlots won't affect other tests.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1996,10 +2012,10 @@ describe('plugin-meetings', () => {
                   assert.notCalled(locusMediaRequestStub);
                 }
                 if (isMultistream) {
-                  assert.calledOnceWithExactly(fakeMultistreamRoapMediaConnection.publishTrack, fakeMicrophoneTrack);
+                  assert.calledOnceWithExactly(meeting.sendSlotManager.getSlot(MediaType.AudioMain).publishStream, fakeMicrophoneStream);
                 } else {
                   assert.calledOnceWithExactly(fakeRoapMediaConnection.update, {
-                    localTracks: { audio: webrtcAudioTrack, video: null, screenShareVideo: null },
+                    localStreams: { audio: fakeMicrophoneStream, video: null, screenShareVideo: null },
```
marcin-bazyl, line 2018: this doesn't look right. RoapMediaConnection API takes webrtc tracks not streams (same in line 2060)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: I see you've now changed it all to use SendSlotManager.getSlot() and check the right method is called on it, I think that's fine, so marking this as resolved.

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
```diff
@@ -13,27 +13,31 @@ describe('createMediaConnection', () => {
     id: 'any fake track'
   }
   const fakeAudioStream = {
+    outputTrack: fakeTrack,
     outputStream:{
```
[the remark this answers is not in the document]
mkesavan13: Done.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -71,7 +71,7 @@ export default class SendSlotManager {
     await slot.publishStream(stream);
 
     this.LoggerProxy.logger.info(
-      `SendSlotsManager->publishStream#Published stream for ${mediaType} and stream ${stream}`
+      `SendSlotsManager->publishStream#Published stream for ${mediaType} and stream with label ${stream.label} and muteed ${stream.muted}`
```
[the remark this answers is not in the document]
mkesavan13: Done.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
```diff
@@ -18,9 +18,9 @@ describe('plugin-meetings', () => {
   const createFakeLocalStream = (id, muted) => {
     return {
       id,
+      setServerMuted: sinon.spy(),
```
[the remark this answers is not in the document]
mkesavan13: Not sure why it changed. Must've pasted it wrongly from somewhere. Changed to stub now. 

Done.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: Done. Reverted these changes wherever required.

packages/@webex/plugin-meetings/test/unit/spec/meeting/muteState.js
mkesavan13: Done.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Once again, it should have been a typo from my side. Fixed it and also fixed the tests that failed related to this. 

Done.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5217,8 +5225,8 @@ export default class Meeting extends StatelessWebexPlugin {
     this.setupMediaConnectionListeners();
 
     if (this.isMultistream) {
-      this.sendSlotManager.createSlot(mc, MediaType.VideoMain, false);
-      this.sendSlotManager.createSlot(mc, MediaType.AudioMain, false);
+      this.sendSlotManager.createSlot(mc, MediaType.VideoMain, options.audioEnabled);
```
marcin-bazyl, line 5228: when `audioEnabled` is not specified by the client calling `addMedia()` the default value used is `true` (see code inside addMedia), here the type says that it can be undefined and in that case we will set slot to inactive, which is the opposite of what we want. This could be fixed by passing `audioEnabled` instead of `options.audioEnabled` in line 5398, but you should also fix the type that createMediaConnection() takes to reflect the fact that `audioEnabled` cannot be undefined. I think you can do it by using `Required<AddMediaOptions>` but it's also a good practice to define a type that only contains things you really need and use. In this case you're not using most of the things from AddMediaOptions, so we probably should use Pick<> or actually it would be easier to just define a type with the 4 props you actually need (audioEnabled, videoEnabled, etc)

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3489,6 +3489,27 @@ export default class Meeting extends StatelessWebexPlugin {
     await this.publishStream(MediaType.AudioSlides, this.mediaProperties.shareAudioStream);
   }
 
+  /**
+   * Handles the local audio stream publish state change event
+   * @internal
+   * @returns {undefined}
+   */
+  private emitPublishStateChangeEvent({functionName, isPublished, mediaType, stream}) {
```
marcin-bazyl, line 3497: types are missing for all the parameters

mkesavan13: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: We do not need the AddMediaOptions for transcoded meetings. In that case, we cannot mark it as a required parameter, right?

If so, is it okay if I check for those options inside `if(this.isMultiStream)`?

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: mediaType should be MediaType not string

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: As per Webex discussions, used mediaDirections instead of AddMediaOptions.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Sorry. Thanks for noticing. Done.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5750,8 +5825,8 @@ export default class Meeting extends StatelessWebexPlugin {
       }
     }
 
-    if (receiveShare !== undefined) {
-      this.mediaProperties.mediaDirection.receiveShare = receiveShare;
+    if (shareAudioEnabled !== undefined || shareVideoEnabled !== undefined) {
+      this.mediaProperties.mediaDirection.receiveShare = shareAudioEnabled || shareVideoEnabled;
```
marcin-bazyl, line 5829: it's probably worth wrapping it with `!!` like this: `!!(shareAudioEnabled || shareVideoEnabled)`, because one of them might be undefined, for example if  shareAudioEnabled is false and shareVideoEnabled is undefined, the existing code would set receiveShare to undefined instead of false

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5726,7 +5798,10 @@ export default class Meeting extends StatelessWebexPlugin {
       return this.enqueueMediaUpdate(MEDIA_UPDATE_TYPE.UPDATE_MEDIA, options);
     }
 
-    if (this.isMultistream && receiveShare !== undefined) {
+    if (
+      this.isMultistream &&
+      (shareAudioEnabled !== undefined || shareVideoEnabled !== undefined)
+    ) {
       throw new Error(
         'toggling receiveShare in a multistream meeting is not supported, to control receiving screen share call meeting.remoteMediaManager.setLayout() with appropriate layout'
```
marcin-bazyl, line 5806: the error message needs to be updated - it still references non non-existent `receiveShare` from options argument

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Agreed. Done.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done.

marcin-bazyl (approved):

marcin-bazyl:
approving the PR, although there are still some issues to investigate or fix, but they can be done in other PRs

PR #3035, opened 2023-08-30
test(calling): unit test add and refactor

meta-dipanshu-sharma:
# COMPLETES #< [SPARK-422965](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-422965) >

## This pull request addresses

Unit test added and refactored for below files:
1. line.test.ts
2. register.test.ts
3. callingClient.test.ts
4. metrics.test.ts

Refactor work done for following files:
1. line.ts
2. register.ts
3. callingClient.ts
4. metrics.ts
5. testUtils.ts

## by making the following changes

Added and refactored unit tests. Also done refactor work.

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

### I certified that

- [ ] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [ ] I discussed changes with code owners prior to submitting this pull request

- [ ] I have not skipped any automated checks
- [ ] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3059, opened 2023-09-06
test(calling): test refactored and added

meta-dipanshu-sharma:
NOTE: This PR is duplicate of [PR](https://github.com/webex/webex-js-sdk/pull/3035) which got messed up because of rebasing to next.

# COMPLETES #< [SPARK-422965](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-422965) >

Unit test added and refactored for below files:
1. line.test.ts
2. register.test.ts
3. callingClient.test.ts
4. metrics.test.ts

Refactor work done for following files:
1. line.ts
2. register.ts
3. callingClient.ts
4. metrics.ts
5. testUtils.ts

Added and refactored unit tests. Also done refactor work.

Screenshot 2023-09-12 at 1 36 36 AM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -53,6 +53,7 @@ module.exports = {
         '@typescript-eslint/no-explicit-any': 'off',
         'lines-between-class-members': ['error', 'always', {exceptAfterSingleLine: true}],
         'tsdoc/syntax': 'off', // Todo: remove this before merging pr
+        'prefer-destructuring': ['error', {object: false, array: false}],
       },
     },
     {
--- a/packages/calling/src/CallingClient/CallingClient.test.ts
+++ b/packages/calling/src/CallingClient/CallingClient.test.ts
@@ -1,15 +1,8 @@
-import {Mutex} from 'async-mutex';
 import {LOGGER} from '../Logger/types';
-import {
-  getMockDeviceInfo,
-  getTestUtilsWebex,
-  getMobiusDiscoveryResponse,
-  getMockRequestTemplate,
-} from '../common/testUtil';
+import {getTestUtilsWebex, getMockRequestTemplate, getMockDeviceInfo} from '../common/testUtil';
 import {
   CallDirection,
   CallType,
-  MobiusServers,
   MobiusStatus,
   ServiceIndicator,
   WebexRequestPayload,
@@ -17,26 +10,51 @@ import {
 /* eslint-disable dot-notation */
 import {CallSessionEvent, EVENT_KEYS, MOBIUS_EVENT_KEYS} from '../Events/types';
 import log from '../Logger';
-import {CallingClient, createClient} from './CallingClient';
+import {createClient} from './CallingClient';
 import {ICallingClient} from './types';
 import * as utils from '../common/Utils';
 import {getCallManager} from './calling/callManager';
-import {CALLING_CLIENT_FILE, URL_ENDPOINT} from './constants';
+import {
+  CALLING_CLIENT_FILE,
+  DISCOVERY_URL,
+  NETWORK_CHANGE_DETECTION_UTIL,
+  NETWORK_FLAP_TIMEOUT,
+  REGISTRATION_FILE,
+  SPARK_USER_AGENT,
+} from './constants';
 import {MOCK_MULTIPLE_SESSIONS_EVENT, MOCK_SESSION_EVENT} from './callRecordFixtures';
-import * as regUtils from './registration/register';
+import {ILine} from './line/types';
+import {
+  ipPayload,
+  regionBody,
+  regionPayload,
+  primaryUrl,
+  discoveryPayload,
+  registrationPayload,
+  uri,
+  myIP,
+} from './callingClientFixtures';
 
 describe('CallingClient Tests', () => {
   // Common initializers
 
-  const handleErrorSpy = jest.spyOn(utils, 'handleRegistrationErrors');
+  const handleErrorSpy = jest.spyOn(utils, 'handleCallingClientErrors');
   const webex = getTestUtilsWebex();
   const defaultServiceIndicator = ServiceIndicator.CALLING;
   const callManager = getCallManager(webex, defaultServiceIndicator);
 
+  const logSpy = jest.spyOn(log, 'info');
+  const warnSpy = jest.spyOn(log, 'warn');
+
+  const originalProcessNextTick = process.nextTick;
+  function flushPromises() {
+    return new Promise((resolve) => {
+      originalProcessNextTick(resolve);
+    });
+  }
+
   describe('ServiceData tests', () => {
     let callingClient: ICallingClient | undefined;
-    const defaultServiceData = {indicator: ServiceIndicator.CALLING, domain: ''};
-    const createRegistrationSpy = jest.spyOn(regUtils, 'createRegistration');
 
     afterEach(() => {
       callManager.removeAllListeners();
@@ -48,35 +66,10 @@ describe('CallingClient Tests', () => {
      * calling and empty domain value.
      */
     it('Verify valid calling serviceData with no input sdk config', async () => {
-      expect(() => {
-        callingClient = createClient(webex);
+      expect(async () => {
+        callingClient = await createClient(webex);
+        expect(callingClient).toBeTruthy();
       }).not.toThrow(Error);
-      expect(callingClient).toBeTruthy();
-      expect(createRegistrationSpy).toBeCalledWith(
-        webex,
-        defaultServiceData,
-        expect.any(Mutex),
-        expect.anything(),
-        expect.any(String)
-      );
-    });
-
-    /**
-     * Input sdk config to callingClient but no serviceData in it, it should default to
-     * service type calling and empty domain value.
-     */
-    it('Verify valid calling serviceData with no input sdk config', async () => {
-      expect(() => {
-        callingClient = createClient(webex, {logger: {level: LOGGER.INFO}});
-      }).not.toThrow(Error);
-      expect(callingClient).toBeTruthy();
-      expect(createRegistrationSpy).toBeCalledWith(
-        webex,
-        defaultServiceData,
-        expect.any(Mutex),
-        expect.anything(),
-        expect.any(String)
-      );
     });
 
     /**
@@ -90,36 +83,10 @@ describe('CallingClient Tests', () => {
       /* eslint-disable @typescript-eslint/no-explicit-any */
       const serviceDataObj: any = {indicator: '', domain: ''};
 
-      expect(() => {
-        callingClient = createClient(webex, {serviceData: serviceDataObj});
+      expect(async () => {
+        callingClient = await createClient(webex, {serviceData: serviceDataObj});
+        expect(callingClient).toBeTruthy();
       }).not.toThrow(Error);
-      expect(callingClient).toBeTruthy();
-      expect(createRegistrationSpy).toBeCalledWith(
-        webex,
-        defaultServiceData,
-        expect.any(Mutex),
-        expect.anything(),
-        expect.any(String)
-      );
-    });
-
-    /**
-     * Input sdk config to callingClient with serviceData carrying invalid value for indicator
-     * and empty string for domain field in it.
-     *
-     * It should throw error and abort execution as indicator value is invalid.
-     *
-     * INDICATOR value MUST MATCH with one of the values in ServiceIndicator enum.
-     */
-    it('Verify invalid service indicator, empty domain', async () => {
-      /* eslint-disable @typescript-eslint/no-explicit-any */
-      const serviceDataObj: any = {indicator: 'test', domain: ''};
-
-      expect(() => {
-        callingClient = createClient(webex, {serviceData: serviceDataObj});
-      }).toThrow(Error);
-      expect(callingClient).toBe(undefined);
-      expect(createRegistrationSpy).not.toHaveBeenCalled();
     });
 
     /**
@@ -134,11 +101,14 @@ describe('CallingClient Tests', () => {
       /* eslint-disable @typescript-eslint/no-explicit-any */
       const serviceDataObj: any = {indicator: 'test', domain: 'test.example.com'};
 
-      expect(() => {
-        callingClient = createClient(webex, {serviceData: serviceDataObj});
-      }).toThrow(Error);
-      expect(callingClient).toBe(undefined);
-      expect(createRegistrationSpy).not.toHaveBeenCalled();
+      try {
+        callingClient = await createClient(webex, {serviceData: serviceDataObj});
+      } catch (e) {
+        expect(e.message).toEqual(
+          'Invalid service indicator, Allowed values are: calling,contactcenter'
+        );
+      }
+      expect.assertions(1);
     });
 
     /**
@@ -153,11 +123,12 @@ describe('CallingClient Tests', () => {
     it('CALLING: verify invalid service domain', async () => {
       const serviceDataObj = {indicator: ServiceIndicator.CALLING, domain: 'test'};
 
-      expect(() => {
-        callingClient = createClient(webex, {serviceData: serviceDataObj});
-      }).toThrow(Error);
-      expect(callingClient).toBe(undefined);
-      expect(createRegistrationSpy).not.toHaveBeenCalled();
+      try {
+        callingClient = await createClient(webex, {serviceData: serviceDataObj});
+      } catch (e) {
+        expect(e.message).toEqual('Invalid service domain.');
+      }
+      expect.assertions(1);
     });
 
     /**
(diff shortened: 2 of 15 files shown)


packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        webex,
+        defaultServiceData,
+        expect.any(Mutex),
+        expect.anything(),
+        LOGGER.INFO
+      );
+      expect(line.getRegistrationStatus()).toEqual(MobiusStatus.DEFAULT);
+      await line.register();
+
+      expect(webex.request).toBeCalledTimes(1);
```
Kesari3008, line 95: Can we change this to use toBeCalledOnceWith here? Applies everywhere where we are using toBeCalledTimes(1)

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      /** trigger sendKeepAlive and verify whether keepalive request is sent or not */
+      webex.request.mockClear();
+
+      /* specify keepalive interval as 30 seconds and advance timers by 30 */
+      line.sendKeepAlive({...mockRegistrationBody, keepaliveInterval: 30});
+
+      jest.advanceTimersByTime(30 * 1000);
+      await Promise.resolve();
+
+      expect(webex.request).toBeCalledWith({
```
Kesari3008, line 112: Same comment here as well. Instead of toBeCalledWith use toBeCalledOnceWith, applies everywhere

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      webex.request.mockImplementation(() => {
+        throw new Error();
+      });
+
+      expect(line.getRegistrationStatus()).toBe(MobiusStatus.DEFAULT);
+      line.register();
+      await utils.waitForMsecs(20);
+
+      expect(line.getRegistrationStatus()).toBe(MobiusStatus.DEFAULT);
+      expect(handleErrorSpy).toBeCalled();
```
Kesari3008, line 171: When errorSpy is called we should compare what it is being called with. Here as well use toBeCalledOnceWith()

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      );
+
+      expect(createRegistrationSpy).toBeCalledWith(
+        webex,
+        defaultServiceData,
+        expect.any(Mutex),
+        expect.anything(),
+        LOGGER.WARN
+      );
+      expect(line.getRegistrationStatus()).toEqual(MobiusStatus.DEFAULT);
```
Kesari3008, line 143: Just noticed that the registration status is reflecting MobiusStatus enum, I want to discuss because we also have LineStatus created now

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -153,11 +156,12 @@ describe('CallingClient Tests', () => {
     it('CALLING: verify invalid service domain', async () => {
```
Kesari3008, line 156: I feel some of the test cases here might not be required, let's discuss the ones which needs to be removed

Kesari3008 (changes requested):
Minor comments

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: For this line, here it's been called 8 times so we can't use `toBeCalledOnceWith` or `toBeCalledWith`. At other places where it's called only once, we have used `calledWith`. I have added `toBeCalledTimes(8)` at above line.

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -48,35 +66,21 @@ describe('CallingClient Tests', () => {
...
-        expect.anything(),
-        expect.any(String)
-      );
     });
 
     /**
      * Input sdk config to callingClient but no serviceData in it, it should default to
      * service type calling and empty domain value.
      */
     it('Verify valid calling serviceData with no input sdk config', async () => {
```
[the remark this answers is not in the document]
meta-dipanshu-sharma: As per our discussion, I have removed duplicate test cases.

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -48,35 +66,21 @@ describe('CallingClient Tests', () => {
...
-      expect(() => {
-        callingClient = createClient(webex, {logger: {level: LOGGER.INFO}});
+      expect(async () => {
+        callingClient = await createClient(webex, {logger: {level: LOGGER.INFO}});
```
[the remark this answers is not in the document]
meta-dipanshu-sharma: We have separate describe blocks where we are testing `getMobiusServers` and `createLine`. If we have to check if these are getting invoked then we have to mock them. Do you want to test them in this describe block also? Or is there any other way to test if these are getting invoked or not without mocking them?

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: Done at all other places and all other files as well.

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: Done at all places.

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: We can discuss but imo registration status will be related to Mobius only.

packages/calling/src/CallingClient/CallingClient.test.ts
meta-dipanshu-sharma: Yeah removed as per our discussion.

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  const primaryMobiusUris = jest.fn(() => mobiusUris.primary);
+  const backupMobiusUris = jest.fn(() => mobiusUris.backup);
+  const userId = webex.internal.device.userId;
+  const clientDeviceUri = webex.internal.device.url;
+
+  const handleErrorSpy = jest.spyOn(utils, 'handleRegistrationErrors');
+
+  jest.clearAllMocks();
+
+  describe('Registration tests', () => {
```
BhargavSatya, line 42: Looks like `case LINE_EVENTS.ERROR:` lineEmitter method is never hit. Can we write a testcase for that?

BhargavSatya, line 42:
```suggestion
  describe('Line tests', () => {
```

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -70,7 +68,6 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
     mutex: Mutex,
     primaryMobiusUris: string[],
     backupMobiusUris: string[],
-    callingClientEmitter: (event: EVENT_KEYS, clientError?: CallingClientError) => void,
     logLevel: LOGGER,
```
BhargavSatya, line 71: Apologies for noticing this now, the below method in this file looks like a dead code with no references. 
`keepaliveTimer` is  started automatically in register.ts file once a registration is successful and developers need not manually send/refresh the timer manually.
Earlier this was a public method in CallingClient but was never invoked by any client. 

@Kesari3008  Shall we remove this? Lets remove any declarations in any interfaces as well (line and CallingClient)

`134 Line:   private async sendKeepAlive(deviceInfo: IDeviceInfo): Promise<void> {`

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+      expect(webex.request).toBeCalledOnceWith({
+        ...getMockRequestTemplate(),
+        uri: `${mockRegistrationBody.device.uri}/status`,
+        method: 'POST',
+      });
+      jest.useRealTimers();
+    });
+
+    it('verify post registration api call when default mobius url is used', async () => {
```
BhargavSatya, line 129: I feel it does not matter to line class which url is being used.
Ideally CallingClient would take care of deciding the primaryUrl and backupUrl's and adding a default url incase of discovery failure. This should have been a test case in CallingClient. Even registration class would be unaware of presence of a default Url. So this is purely a CallingClient's test case.

packages/calling/src/CallingClient/line/line.test.ts
BhargavSatya: @Kesari3008  isn't it confusing to have both `LineStatus` and `MobiusStatus` enums in SDK since they fundamentally mean the same thing (at least thats how they are being used interchangeably)?

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -234,254 +162,394 @@ describe('CallingClient Tests', () => {
...
+      jest.advanceTimersByTime(NETWORK_FLAP_TIMEOUT + 500);
+
+      await flushPromises();
+
+      /* We should be detecting the network recovery */
+      expect(logSpy).nthCalledWith(
+        7,
+        'Mercury connection is up again, Re-registering with Mobius',
+        {file: REGISTRATION_FILE, method: 'handleConnectionRestoration'}
       );
```
BhargavSatya, line 386: You may add these lines as well since at the beginning of test case: `line.getRegistrationStatus() was ACTIVE`

```suggestion
      );
       expect(deRegSpy).toBeCalledWith();
      expect(restoreSpy).toBeCalledWith('handleConnectionRestoration');
      expect(restartRegisterSpy).toBeCalledWith('handleConnectionRestoration');
      expect(webex.request).toBeCalledTimes(6);
      expect(registerSpy).toBeCalledWith('handleConnectionRestoration', [reg.getActiveMobiusUrl()]);
      expect(registerSpy).lastCalledWith('handleConnectionRestoration', [primaryUrl]);
```

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -3,7 +3,7 @@
 /* eslint-disable @typescript-eslint/no-shadow */
 import * as Media from '@webex/internal-media-core';
 import {Mutex} from 'async-mutex';
-import {filterMobiusUris, handleRegistrationErrors, validateServiceData} from '../common/Utils';
+import {filterMobiusUris, handleCallingClientErrors, validateServiceData} from '../common/Utils';
```
BhargavSatya, line 6: No testcase covers the scenario where there was a registration reconnect during active calls.
Following lines to be covered in a testcase.
Method at Line 351: ` private callsClearedHandler = async ()`

BhargavSatya (changes requested):
Few comments.

Can you manually attach a snapshot of coverage report in Description for future reference since we don't have it yet on the new pipeline yet.

It can be generated using below command for calling, specifically.
`yarn workspace @webex/calling run test `
and can be found at -> `packages/calling/coverage/lcov-report/index.html`

packages/calling/src/CallingClient/line/line.test.ts
BhargavSatya: 💡 Tip: we can always use `mock.calls` or `nthCalledWith` in such scenarios. 

However, not needed in this case.
Although I am wondering how come there are 8 such invocations.

packages/calling/src/CallingClient/line/line.test.ts
Kesari3008: LineStatus and MobiusStatus are nothing but enums created in calling SDK and reflect the same thing. This is redundant so we can remove one of the enums and use the left one. 
This change is not necessary to be done in this PR, can be taken care later

packages/calling/src/CallingClient/line/line.test.ts
Kesari3008: registration failure cases should be hitting the LINE_EVENTS.ERROR only

packages/calling/src/CallingClient/line/index.ts
Kesari3008: @BhargavSatya  Thanks for bringing this to notice, I missed this actually. @meta-dipanshu-sharma  You can remove this method completely, Bhargav has renamed the method sendKeepalive to startKeepaliveTimer and it exists in the register.ts file. It's a private function so we can keep it private as it is only going to be invoked internally.

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: Not sure how other test was affecting number of mock calls in this test as we're clearing all mocks after each test but after removing the default mobius url test case now this is called only once and I have added calledOnceWith here.

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: As suggested, not making this change as part of this test case PR.

packages/calling/src/CallingClient/line/index.ts
meta-dipanshu-sharma: Removed `sendKeepAlive` wherever used.

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: Yeah removed it from line test and default mobius url block is already covered in existing calling client tests.

packages/calling/src/CallingClient/CallingClient.test.ts
meta-dipanshu-sharma: Added

packages/calling/src/CallingClient/CallingClient.ts
meta-dipanshu-sharma: It will be covered as part of call related ticket on which Shreyas is currently working on.

.eslintrc.js
```diff
@@ -92,6 +93,7 @@ module.exports = {
         'no-multi-assign': 0,
         'no-promise-executor-return': 0,
         'no-restricted-exports': 0,
+        'import/no-unresolved': [2, {caseSensitive: false}],
```
Kesari3008, line 96: Why did we need to add this

packages/calling/src/CallingClient/line/line.test.ts
BhargavSatya: Thanks!

packages/calling/src/CallingClient/line/line.test.ts
BhargavSatya: Works!

packages/calling/src/CallingClient/CallingClient.ts
BhargavSatya: Thanks!

BhargavSatya (approved):
Approving this with a comment on one minor test case to check.

https://github.com/webex/webex-js-sdk/pull/3059#discussion_r1321521428

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -289,7 +288,7 @@ describe('Registration Tests', () => {
 
     it('verify 429 error with failback to primary after initial registration with backup: Restore failure', async () => {
       // delete should be successful
-      global.fetch = jest.fn(() => Promise.resolve({json: () => mockDeleteResponse})) as jest.Mock;
+      global.fetch = jest.fn(() => Promise.resolve({json: () => mockDeleteResponse})) as any;
```
Kesari3008, line 291: why can't we use jest.Mock here ?

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -171,53 +142,10 @@ describe('CallingClient Tests', () => {
...
-
-    /**
-     * Input sdk config to callingClient with serviceData carrying valid value for indicator
-     * 'contactcenter', but an empty string for domain field in it.
-     *
-     * It should throw error and abort execution as domain value is invalid.
-     *
-     * DOMAIN field for service type 'contactcenter' must carry a non-empty valid domain type string.
-     */
-    it('ContactCenter: verify empty invalid service domain', async () => {
```
Kesari3008, line 195: we can keep this testcase

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -234,254 +162,401 @@ describe('CallingClient Tests', () => {
...
+      await flushPromises();
+
+      /* We should be detecting the network recovery */
+      expect(logSpy).not.toBeCalledWith(
+        'Mercury connection is up again, Re-registering with Mobius',
+        {file: REGISTRATION_FILE, method: 'handleConnectionRestoration'}
+      );
+
+      expect(registerSpy).not.toBeCalledWith(true);
+      expect(registerSpy).toBeCalledTimes(0);
```
Kesari3008, line 520: first expect statement should suffice, we don't need toBeCalledTimes(0)

Kesari3008 (approved):
Approving with minor comments

packages/calling/src/CallingClient/line/line.test.ts
meta-dipanshu-sharma: Taken care of this line. Please check. Updated the coverage report screenshot as well.

packages/calling/src/CallingClient/registration/register.test.ts
meta-dipanshu-sharma: We can use. Updated it.

packages/calling/src/CallingClient/CallingClient.test.ts
meta-dipanshu-sharma: Pulled it back.

packages/calling/src/CallingClient/CallingClient.test.ts
meta-dipanshu-sharma: Added by mistake while creating this new PR. Corrected it.

meta-dipanshu-sharma:
@BhargavSatya @Kesari3008 I have taken care of all the minor comments given. Please verify and I'll merge the PR after that.

PR #3041, opened 2023-09-01
fix(meeting): added noFramesSent noVideoEncoded events

Shreyas281299:
# COMPLETES #[SPARK-452740](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-452740)

## This pull request addresses

Currently, we don't send any metric when frames are not being sent for various reasons on mobile devices. If device is not sending videoFrames or video is not encoded, we are not triggering any event for that.

## by making the following changes

As part of this PR:
1) added an EVENT when video is encoded but frames are not sent in statsAnalyzer
2) added an EVENT when video is not being encoded in statsAnalyzer
3) Emitted two new EVENTS for the above scenarios
4) sending behavioral metric to Amplitude for above scenarios

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested
Test the following scenarios:
- when frames sent is 0
- when frames sent value is not changing
- when frames sent is 0 but video is encoded
- when video is not being encoded 

## Screenshot of samples page for the new events
Screenshot 2023-09-06 at 5 04 32 PM

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3017, opened 2023-08-22
fix(plugin-meetings): no-frames-received-sent-stats-error

Shreyas281299:
# COMPLETES # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-251243

## This pull request addresses
The logs were filled with 'No video frames sent/received/decoded' and 'No audio samples sent/received/decoded' logs. 

Root cause:
RTCStatsReport used to send statistics for type = 'track' which had the details about the track like frames sent, frames height etc. Now this information is being sent in the type inbound-rtp and outbound-rtp statistics

Fixed the StatsAnalyzer to correctly populate the track results.

## by making the following changes
Calling the processTrackResult function from the  processInboundRTPResult and processOutboundRTPResult functions.

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3066, opened 2023-09-10
fix(plugin-meetings): no-frames-received-sent-stats-error

Shreyas281299:
# COMPLETES #[SPARK-452740](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-452740) and [SPARK-251243](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-251243)

Currently, we don't send any metric when frames are not being sent for various reasons on mobile devices. If device is not sending videoFrames or video is not encoded, we are not triggering any event for that.

To trigger the event we need information about the framesSent and framesEncoded. The logs were filled with 'No video frames sent/received/decoded' and 'No audio samples sent/received/decoded' logs.

Root cause:
RTCStatsReport used to send statistics for type = 'track' which had the details about the track like frames sent, frames height etc. Now this information is being sent in the type inbound-rtp and outbound-rtp statistics

Fixed the StatsAnalyzer to correctly populate the track results.

As part of this PR:
1) added an EVENT when video is encoded but frames are not sent in statsAnalyzer
2) added an EVENT when video is not being encoded in statsAnalyzer
3) Emitted two new EVENTS for the above scenarios
4) sending behavioral metric to Amplitude for above scenarios

##Reference PR in master
[fix(plugin-meetings): no-frames-received-sent-stats-error](https://github.com/webex/webex-js-sdk/pull/3017) 
[ fix(meeting): added noFramesSent noVideoEncoded events](https://github.com/webex/webex-js-sdk/pull/3041)
### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

For testing each event the particular stats are made zero manually and tested that event is triggered after every 5 seconds and metric is sent to amplitude ONLY once.

Following are the events and related stat.
- for NO_FRAMES_ENCODED , framesEncoded is set to zero.
- for NO_FRAMES_SENT, framesSent is set to zero.
- for NO_FRAMES_DECODED, framesDecoded is set to zero
- for NO_FRAMES_RECEVIED, framesRecevied is set to zero
- for Audio, totalAudioEnergy, audioLevel, packetsReceived and samples received were set to zero.

### I certified that

---


packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -623,8 +646,28 @@ export class StatsAnalyzer extends EventsScope {
               this.statsResults.resolutions['video-send'].send.framesSent
             );
           }
-        }
 
+          // Video is encoded but frames are not sent
+          if (
+            currentStats.framesEncoded !== previousStats.framesEncoded &&
+            (currentStats.framesSent === previousStats.framesSent || currentStats.framesSent === 0)
```
marcin-bazyl, line 653: in here we're reading framesSent from currentStats/previousStats, while the code in lines 640-648 is using `this.statsResults.resolutions['video-send'].send.framesSent` and `this.lastStatsResults.resolutions['video-send'].send.framesSent` - why the difference? are these always the same? this is confusing....

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -103,7 +105,20 @@ export class StatsAnalyzer extends EventsScope {
     this.correlationId = config.correlationId;
     this.mqaSentCount = -1;
     this.lastMqaDataSent = {};
-    this.lastEmittedStartStopEvent = {};
+    this.lastEmittedStartStopEvent = {
+      audio: {
```
marcin-bazyl, line 109: if we do this initialization here now, then we don't need the code in lines 466-468

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -602,15 +614,26 @@ export class StatsAnalyzer extends EventsScope {
             `StatsAnalyzer:index#compareLastStatsResult --> No video RTP packets sent`,
             currentStats.totalPacketsSent
           );
-        } else {
+        } else if (this.lastEmittedStartStopEvent.video.local !== EVENTS.LOCAL_MEDIA_STOPPED) {
```
marcin-bazyl, line 617: previously the code would output the log "No video Frames Encode" every 5s for as long as we were not encoding the frames and now it won't do that (because even if you set lastEmittedStartStopEvent.video.local to EVENTS.NO_VIDEO_ENCODED it will get overwritten to LOCAL_MEDIA_STOPPED by emitStartStopEvents that gets called later)  - this is not good, having those logs is useful

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -602,15 +614,26 @@ export class StatsAnalyzer extends EventsScope {
...
           if (
             currentStats.framesEncoded === previousStats.framesEncoded ||
             currentStats.framesEncoded === 0
           ) {
+            this.lastEmittedStartStopEvent.video.local = EVENTS.NO_VIDEO_ENCODED;
```
marcin-bazyl, line 622: I really don't like how we're now setting lastEmittedStartStopEvent to something else than MEDIA_STARTED or MEDIA_STOPPED. The name lastEmittedStartStopEvent clearly says that it should be only started/stopped event and also lastEmittedStartStopEvent  should only ever by updated inside emitStartStopEvents(), Otherwise it's easy to introduce bugs. The whole point of emitStartStopEvents and lastEmittedStartStopEvent is to make sure that we never send STOPPED without STARTED or never send STARTED or STOPPED twice in a row etc. With the changes in this PR, the code becomes more convoluted and I'm not sure if it's guaranteed to still work correctly.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -699,15 +742,26 @@ export class StatsAnalyzer extends EventsScope {
             `StatsAnalyzer:index#compareLastStatsResult --> No share RTP packets sent`,
             currentStats.totalPacketsSent
           );
-        } else {
+        } else if (this.lastEmittedStartStopEvent.share.local !== EVENTS.LOCAL_MEDIA_STOPPED) {
```
marcin-bazyl, line 745: same as above

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -699,15 +742,26 @@ export class StatsAnalyzer extends EventsScope {
...
           if (
             currentStats.framesEncoded === previousStats.framesEncoded ||
             currentStats.framesEncoded === 0
           ) {
+            this.lastEmittedStartStopEvent.share.local = EVENTS.NO_VIDEO_ENCODED;
```
marcin-bazyl, line 750: same as above

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -720,6 +774,27 @@ export class StatsAnalyzer extends EventsScope {
               this.statsResults.resolutions['video-share-send'].send.framesSent
             );
           }
+
+          // Share video is encoded but frames are not sent
+          if (
+            currentStats.framesEncoded !== previousStats.framesEncoded &&
+            (currentStats.framesSent === previousStats.framesSent || currentStats.framesSent === 0)
```
marcin-bazyl, line 781: same as above

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -874,10 +949,11 @@ export class StatsAnalyzer extends EventsScope {
   private processOutboundRTPResult(result: any, mediaType: any) {
     const sendrecvType = STATS.SEND_DIRECTION;
 
+    this.processTrackResult(result, mediaType, sendrecvType);
```
marcin-bazyl, line 952: processTrackResult is now called to process inbound or outbound rtp result and not track result, so it should be renamed

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -874,10 +949,11 @@ export class StatsAnalyzer extends EventsScope {
...
     if (result.bytesSent) {
       let kilobytes = 0;
 
-      if (result.frameWidth && result.frameHeight) {
+      if (result.frameWidth && result.frameHeight && !mediaType.includes('audio')) {
```
marcin-bazyl, line 956: why do we need this now? the code has been working fine so far without it, right? also if you add it here, then why not also for framesEncoded and keyFramesEncoded a few lines below (and more of it further below)

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -955,14 +1031,15 @@ export class StatsAnalyzer extends EventsScope {
...
+    this.processTrackResult(result, mediaType, sendrecvType);
     if (result.bytesReceived) {
       let kilobytes = 0;
       const receiveSlot = this.receiveSlotCallback(result.ssrc);
       const idAndCsi = receiveSlot
         ? `id: "${receiveSlot.id || ''}"${receiveSlot.csi ? ` and csi: ${receiveSlot.csi}` : ''}`
         : '';
 
-      if (result.frameWidth && result.frameHeight) {
+      if (result.frameWidth && result.frameHeight && !mediaType.includes('audio')) {
```
marcin-bazyl, line 1042: same as above, there are many more video specific things further below (like pliCount)

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed line 466-468

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed the condition from here. This was done to avoid sending multiple metrics to amplitude and added the condition before sending the event. So now logs will be printed after every 5s and the event will only be sent once.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed it . This was just an intermediate state for lastEmittedStartStopEvent.video.local.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Moved this condition just above the event so that logs are printed every 5s and event is emitted only once.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -463,7 +475,7 @@ export class StatsAnalyzer extends EventsScope {
 
     if (currentValue - previousValue > 0) {
       newEvent = isLocal ? EVENTS.LOCAL_MEDIA_STARTED : EVENTS.REMOTE_MEDIA_STARTED;
-    } else if (currentValue === previousValue && currentValue > 0) {
+    } else if (currentValue === previousValue && currentValue >= 0) {
```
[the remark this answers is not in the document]
Shreyas281299: Yes this will cause us to emit MEDIA_STOPPED before MEDIA_STARTED. Removed this change.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: These values are same. Changed to `this.statsResults.resolutions['video-send'].send.framesSent` and `this.lastStatsResults.resolutions['video-send'].send.framesSent` to keep it uniform.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Changed to `this.statsResults.resolutions['video-send'].send.framesSent` and `this.lastStatsResults.resolutions['video-send'].send.framesSent`

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed this.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Removed this.

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -320,6 +320,8 @@ export const EVENT_TRIGGERS = {
   MEETING_INTERPRETATION_UPDATE: 'meeting:interpretation:update',
   MEETING_INTERPRETATION_SUPPORT_LANGUAGES_UPDATE: 'meeting:interpretation:supportLanguagesUpdate',
   MEETING_INTERPRETATION_HANDOFF_REQUESTS_ARRIVED: 'meeting:interpretation:handoffRequestsArrived',
+  MEETING_NO_FRAMES_SENT: 'meeting:noFramesSent',
```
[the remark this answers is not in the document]
shreyas-28: Added NO_AUDIO_SENT,NO_AUDIO_RECEIVED for audio and NO_FRAMES_RECEVIED, NO_FRAMES_DECODED for video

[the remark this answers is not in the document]
Shreyas281299: Added NO_AUDIO_SEND and NO_AUDIO_RECEIVED events for audio. Added NO_FRAMES_DECODED and NO_FRAMES_RECEIVED for received video.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -1171,24 +1353,34 @@ export class StatsAnalyzer extends EventsScope {
...
+    if (result.type !== 'inbound-rtp' && result.type !== 'outbound-rtp') {
+      return;
     }
-
-    if (result.trackIdentifier && !mediaType.includes('audio')) {
+    if (result.frameWidth && result.frameHeight) {
+      this.statsResults.resolutions[mediaType][sendrecvType].width = result.frameWidth;
+      this.statsResults.resolutions[mediaType][sendrecvType].height = result.frameHeight;
+    }
+    if (!mediaType.includes('audio')) {
```
Shreyas281299, line 1371: This is an optimization to the resolutions object as this was added redundant stats to the resolutions.audio object.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -541,12 +565,36 @@ export class StatsAnalyzer extends EventsScope {
...
+                }
+              );
+            }
           }
 
           if (currentStats.audioLevel === 0) {
             LoggerProxy.logger.info(
               `StatsAnalyzer:index#compareLastStatsResult --> audio level is 0 for the user`
             );
+            if (this.lastEmittedStartStopEvent.audio.remote === EVENTS.REMOTE_MEDIA_STARTED) {
```
marcin-bazyl, line 586: this doesn't look right - why are we checking for remote media here? is `currentStats.audioLevel` for the audio level of the local user's audio or for the incoming remote audio? I thought it was for local audio and also the event that we're sending in line 587 is `NO_AUDIO_SENT`, so what does this have to do with remove audio?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -541,12 +565,36 @@ export class StatsAnalyzer extends EventsScope {
               `StatsAnalyzer:index#compareLastStatsResult --> No audio Energy present`,
               currentStats.totalAudioEnergy
             );
+            if (this.lastEmittedStartStopEvent.audio.remote === EVENTS.LOCAL_MEDIA_STARTED) {
```
marcin-bazyl, line 568: same like my other comment - `currentStats` is set to `this.statsResults['audio-send'].send` so it's about local stuff not remote

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -532,6 +544,18 @@ export class StatsAnalyzer extends EventsScope {
             `StatsAnalyzer:index#compareLastStatsResult --> No audio RTP packets sent`,
             currentStats.totalPacketsSent
           );
+          if (this.lastEmittedStartStopEvent.audio.local === EVENTS.LOCAL_MEDIA_STARTED) {
```
marcin-bazyl, line 547: the first time this gets called, this condition is checked before `emitStartStopEvents()` is called, so `lastEmittedStartStopEvent.audio.local` will always be undefined the first time and so we will only emit the event the second time this function is called - I think that's 5 seconds later. Is that expected? It doesn't feel right.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -602,29 +671,63 @@ export class StatsAnalyzer extends EventsScope {
...
+        ) {
+          LoggerProxy.logger.info(
+            `StatsAnalyzer:index#compareLastStatsResult --> No video Frames sent`,
+            this.statsResults.resolutions['video-send'].send.framesSent
+          );
+        }
+
+        // Video is encoded but frames are not sent
+        if (
+          currentStats.framesEncoded !== previousStats.framesEncoded &&
```
marcin-bazyl, line 710: suggestion: we could avoid duplicating this big condition in this "if" with code from above by just rearranging the conditions, something like this:
```
if (no frames sent) {
   if (no frames encoded) {
    // no frames encoded (so obviously also no frames sent)
   } else {
    // frames encoded but not sent
  }
} 
```

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -699,25 +825,58 @@ export class StatsAnalyzer extends EventsScope {
...
-          if (
-            this.statsResults.resolutions['video-share-send'].send.framesSent ===
-              this.lastStatsResults.resolutions['video-share-send'].send.framesSent ||
-            this.statsResults.resolutions['video-share-send'].send.framesSent === 0
-          ) {
-            LoggerProxy.logger.info(
-              `StatsAnalyzer:index#compareLastStatsResult --> No share frames sent`,
-              this.statsResults.resolutions['video-share-send'].send.framesSent
+        // Share video is encoded but frames are not sent
+        if (
```
marcin-bazyl, line 861: same as above

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -699,25 +825,58 @@ export class StatsAnalyzer extends EventsScope {
...
-        } else {
-          if (
-            currentStats.framesEncoded === previousStats.framesEncoded ||
-            currentStats.framesEncoded === 0
-          ) {
-            LoggerProxy.logger.info(
-              `StatsAnalyzer:index#compareLastStatsResult --> No share frames getting encoded`,
-              currentStats.framesEncoded
+        } else if (
+          currentStats.framesEncoded === previousStats.framesEncoded ||
```
marcin-bazyl, line 829: does this work? do we ever get to here if share frames are not encoded at all? aren't we going to fall into the condition for `totalPacketsSent` above?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
marcin-bazyl: there are still many places where we use `currentStats`/`previousStats` and `this.statsResults.resolutions` / `this.lastStatsResults.resolutions`, are these values all the same for sure? can we use just `currentStats`/`previousStats` everywhere? I know some of this is existing code, but since you're changing it, you might as well try to improve it.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
marcin-bazyl: actually I misread the code I thought this was checking for "stopped" event, but I see it's checking for "started" - I guess this is to prevent the event firing more than once, so that's OK, it's a bit confusing but it will work, except for the case when we don't send anything from the start, then I think lastEmittedStartStopEvent remains undefined (it's never set to STARTED if we never send anything), so we won't trigger the event, am I right?

Shreyas281299: Yes. This is to prevent sending metric to amplitude multiple times. And if we dont send anything at the start the lastEmittedEvent will be undefined so we wont trigger any event.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Yes, this should check for local event not remote. Fixed this.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Changed.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Revamped the code.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Revamped the code.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: changed to 'ProcessResolutionsStats'.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: I checked. These values are the same. Changed to current and previous stats

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -764,13 +895,36 @@ export class StatsAnalyzer extends EventsScope {
...
+              );
+            }
           }
-
           if (currentFramesDecoded === previousFramesDecoded || currentFramesDecoded === 0) {
             LoggerProxy.logger.info(
               `StatsAnalyzer:index#compareLastStatsResult --> No share frames decoded`,
               currentFramesDecoded
             );
+            if (this.lastEmittedStartStopEvent.video.remote === EVENTS.REMOTE_MEDIA_STARTED) {
```
marcin-bazyl, line 916: this should be "share" not "video"

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -764,13 +895,36 @@ export class StatsAnalyzer extends EventsScope {
...
+                },
+                EVENTS.NO_FRAMES_RECEIVED,
+                {
+                  mediaType: 'share',
+                }
+              );
+            }
           }
-
           if (currentFramesDecoded === previousFramesDecoded || currentFramesDecoded === 0) {
```
marcin-bazyl, line 911: same as above

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
marcin-bazyl: why mistakes like that are not caught by unit tests?

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -764,13 +895,36 @@ export class StatsAnalyzer extends EventsScope {
               `StatsAnalyzer:index#compareLastStatsResult --> No share frames received`,
               currentFramesReceived
             );
+            if (this.lastEmittedStartStopEvent.share.remote === EVENTS.REMOTE_MEDIA_STARTED) {
```
marcin-bazyl, line 898: for share this will never work, because SDK currently doesn't call emitStartStopEvents() for share, also existing SDK code is incorrectly checking for `this.meetingMediaStatus.expected.sendShare` when checking receive share stats. To make this work properly for share, it requires more work to set `this.meetingMediaStatus.expected.sendShare` and `this.meetingMediaStatus.expected.receiveShare` correctly

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -665,13 +754,36 @@ export class StatsAnalyzer extends EventsScope {
...
+                },
+                EVENTS.NO_FRAMES_RECEIVED,
+                {
+                  mediaType: 'video',
+                }
+              );
+            }
           }
-
           if (currentFramesDecoded === previousFramesDecoded || currentFramesDecoded === 0) {
```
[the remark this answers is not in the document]
Shreyas281299: Changed to an 'else if' so that only one event is triggered that is NO_FRAMES_RECEIVED.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Changed.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: Fixed this. The UTs were failing.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
Shreyas281299: We can create a JIRA to fix this. In this PR we are only aiming to add all the necessary events.

marcin-bazyl: what is the link to that Jira? Also I don't think we should be adding new API events that don't work correctly from the start. If you are actually planning to go ahead with this, then at the very least this should be documented somewhere in the readme or next to the event definitions that these events don't work for share and a jira should be referenced.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
marcin-bazyl: > And if we dont send anything at the start the lastEmittedEvent will be undefined so we wont trigger any event.

but we should send the NO_AUDIO_SENT event in that case, shouldn't we?

Shreyas281299:
Closing the PR as it is not required anymore. Beta branch is merged into next. Also there has been alot of changes in stats-analyser. PRs in master has been merged already.

PR #3072, opened 2023-09-11
feat(call-analyzer): fix invalid client.call.leave leaveReasons

chburket:
## COMPLETES

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-458903

Currently we are using one of `MEETING_REMOVED_REASON` as the CA event leaveReason.  These are not valid values for the CA event, which needs to be one of 'paired-leave' | 'one-to-one' | 'ended-by-locus'.

We should not conflate locus leave reason with CA event leave reason.

I added a new prop/arg to meeting.leave() (`clientEventLeaveReason`) that can be used to pass in one of the valid CA reasons. By default this property will be undefined (it is optional).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Updated unit tests and tested manually.

### I certified that


arun3528 (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5910,8 +5914,15 @@ export default class Meeting extends StatelessWebexPlugin {
    * @public
    * @memberof Meeting
    */
-  public leave(options: {resourceId?: string; reason?: any} = {} as any) {
+  public leave(
+    options: {
+      resourceId?: string;
+      reason?: any;
```
marcin-bazyl, line 5920: Looks like `reason` is now not used for anything? its value is copied into `leaveReason`, but then `leaveReason` is not passed anywhere further, only logged in 1 place: in line 5978 and we should probably log `options.clientEventLeaveReason` there instead, right? 
Or am I missing something?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5910,8 +5914,15 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      clientEventLeaveReason?: ClientEventLeaveReason;
```
marcin-bazyl, line 5921: how are client apps supposed to know the difference between `reason` and `clientEventLeaveReason` and which one they should set? Do we really need 2 of them? Also jsdoc needs updating

marcin-bazyl (changes requested):

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
chburket: The "reason" arg is now only used for logging. Several places that call meeting.leave() pass in a specific reason, and two other places use the reason from locus events (EVENTS.DESTROY_MEETING, EVENTS.REMOTE_RESPONSE).

The "clientEventLeaveReason" is used for the CA event leaveReason. It must be one of three values. With this change, the "clientEventLeaveReason" will not be anything other than undefined (which is perfectly fine, it's optional), but I will open another Jira to see if we should make use of this arg from some of our meeting.leave() calls

packages/@webex/plugin-meetings/src/meeting/index.ts
chburket: Updated js doc

chburket:
opening a new branch

PR #3074, opened 2023-09-12
feat(spark-454223): CA error handling for pre join events

nataliadelmar:
# COMPLETES #< [SPARK-454223](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-454223)>

The pre-join phase in Webex can result in multiple errors coming from the multiple services it consumes (wbxappapi MeetingInfo request, locus, login, media request errors). Therefore, a study of the errors that could come from this process has been made so to be able to map them to valid Call Analyzer errors found [here](https://sqbu-github.cisco.com/WebExSquared/event-dictionary/wiki/Error-codes-for-metric-events).

1. Adding new possible error codes and names to map to CA errors
2. Fix the default logic inside generateClientEventErrorPayload
3. Add tests for all changes made

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -347,7 +350,18 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
...
-    const serviceErrorCode = rawError?.body?.errorCode || rawError?.body?.code;
+    if (rawError.name) {
+      const clientErrorCode = SERVICE_ERROR_NAME_TO_CLIENT_ERROR_CODES_MAP[rawError.name];
+      if (clientErrorCode) {
+        return this.getErrorPayloadForClientErrorCode({
+          serviceErrorCode: undefined,
+          clientErrorCode,
+        });
+      }
+    }
```
torgeadelin, line 361: we could extract this to a separate function

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -361,15 +375,22 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
           serviceErrorCode,
         });
       }
+    }
 
-      // otherwise return meeting info
+    // MeetingInfo errors sometime has body.data.meetingInfo object
+    // MeetingInfo errors come with a wbxappapi url
+    if (rawError.body?.data?.meetingInfo || rawError.body?.url?.includes(WBX_APP_API_URL)) {
```
torgeadelin, line 382: create a separate function for this similar to `isLocusError`

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -492,15 +508,25 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
     payload?: ClientEventPayload;
     options?: SubmitClientEventOptions;
   }) {
-    const {meetingId, correlationId} = options;
+    const {meetingId, correlationId, rawError} = options;
     let clientEventObject: ClientEvent['payload'];
 
+    // check if we need to generate errors
+    const errors: ClientEvent['payload']['errors'] = [];
```
torgeadelin, line 515: Could we create a type ClientEventErrors in `metrics.types.ts` and use that everywhere

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -6,7 +6,32 @@ import {ClientEventError} from '../metrics.types';
 
 export const NEW_LOCUS_ERROR_CLIENT_CODE = 4008;
 export const MEETING_INFO_LOOKUP_ERROR_CLIENT_CODE = 4100;
+export const UNKNOWN_ERROR = 9999;
```
torgeadelin, line 9: add comment to explain what it is

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -6,7 +6,32 @@ import {ClientEventError} from '../metrics.types';
...
 export const ICE_FAILURE_CLIENT_CODE = 2004;
+export const WBX_APP_API_URL = 'wbxappapi';
+
+// Found in https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
+const MEDIA_ERROR_NAMES = {
```
torgeadelin, line 14: Add some more context in the comment

torgeadelin, line 14: call it `BROWSER_MEDIA_ERROR_NAMES`

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -6,7 +6,32 @@ import {ClientEventError} from '../metrics.types';
...
+  NOT_ALLOWED_ERROR: 'NotAllowedError',
+  NOT_READABLE_ERROR: 'NotReadableError',
+  ABORT_ERROR: 'AbortError',
+  NOT_FOUND_ERROR: 'NotFoundError',
+  OVERCONSTRAINED_ERROR: 'OverconstrainedError',
+  SECURITY_ERROR: 'SecurityError',
+  TYPE_ERROR: 'TypeError',
+};
+
+export const SERVICE_ERROR_NAME_TO_CLIENT_ERROR_CODES_MAP = {
```
torgeadelin, line 25: should be renamed to `BROWSER_MEDIA_ERROR_NAME_TO_CLIENT_ERROR_CODES_MAP`

torgeadelin (changes requested):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
nataliadelmar: if added ClientEventPayloadError bc ClientEventErrors already existed as a type

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -5,12 +5,13 @@ import {
...
   getBuildType,
   isLocusServiceErrorCode,
+  isMeetingInfoServiceError,
   prepareDiagnosticMetricItem,
   setMetricTimings,
 } from '../../../../src/call-diagnostic/call-diagnostic-metrics.util';
 import CallDiagnosticLatencies from '../../../../src/call-diagnostic/call-diagnostic-metrics-latencies';
 
-describe('internal-plugin-metrics', () => {
+describe.only('internal-plugin-metrics', () => {
```
torgeadelin, line 14: remove .only

torgeadelin (changes requested):
some more small changes

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -375,30 +380,48 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
    * @param rawError
    */
   generateClientEventErrorPayload(rawError: any) {
+    if (rawError.name) {
+      if (isBrowserMediaErrorName(rawError.name)) {
+        return this.getErrorPayloadForClientErrorCode({
+          serviceErrorCode: undefined,
+          clientErrorCode: BROWSER_MEDIA_ERROR_NAME_TO_CLIENT_ERROR_CODES_MAP[rawError.name],
```
marcin-bazyl, line 387: the mapping of `rawError.name` to `clientErrorCode` is not 1 tot 1 it's N to 1, so we're losing some information here. It would be good to still send the `rawError.name` to CA, not sure if there is some field we can use for that?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -6,7 +6,32 @@ import {ClientEventError} from '../metrics.types';
...
+  NOT_ALLOWED_ERROR: 'NotAllowedError',
+  NOT_READABLE_ERROR: 'NotReadableError',
+  ABORT_ERROR: 'AbortError',
+  NOT_FOUND_ERROR: 'NotFoundError',
+  OVERCONSTRAINED_ERROR: 'OverconstrainedError',
+  SECURITY_ERROR: 'SecurityError',
+  TYPE_ERROR: 'TypeError',
+};
+
+export const BROWSER_MEDIA_ERROR_NAME_TO_CLIENT_ERROR_CODES_MAP = {
```
marcin-bazyl, line 25: all these errors prevent us from using a camera or a microphone, but they are not fatal, the users will see in the UI that they can't unmute audio or video, but they can still join a meeting, so I think we need to make sure that the errors we map to are all the kind of errors that CA will treat as non-critical and will not mark the call as failed. 

I see some values mapped to 2001 - that's "FailedToConnectMedia" error, which doesn't feel right. The fact whether we have camera or mic doesn't affect if we can connect media or not. We will be able to connect "the rest" of the media - so for example users will still see remote video, share, hear audio etc - media connection will be successful.

The other value used: 2729 I guess it's fine, because I see it's in the "expected" category

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -375,30 +380,48 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
...
       if (isLocusServiceErrorCode(serviceErrorCode)) {
         return this.getErrorPayloadForClientErrorCode({
           clientErrorCode: NEW_LOCUS_ERROR_CLIENT_CODE,
           serviceErrorCode,
         });
       }
+    }
 
-      // otherwise return meeting info
+    if (isMeetingInfoServiceError(rawError)) {
```
marcin-bazyl, line 413: in the code block above (lines 398-411) we check if the error is in `SERVICE_ERROR_CODES_TO_CLIENT_ERROR_CODES_MAP` and I see there are many error codes from WbxAppApi in that mapping, so I think if we get an error from WbxAppApi meeting info, we will execute the return from line 401, right? and we won't get to this code that seems to be written specifically to handle meeting info errors from WbxAppApi. Or am I misunderstanding something?

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -721,61 +779,133 @@ describe('internal-plugin-metrics', () => {
...
+        checkMeetingInfoError({body: {data: {meetingInfo: 'something'}}}, true);
       });
 
-      it('should return undefined if no error code provided', () => {
-        const res = cd.generateClientEventErrorPayload({body: {errorCode: undefined}});
-        assert.deepEqual(res, undefined);
+      it('should return default meeting info lookup error payload if body.url contains wbxappapi', () => {
+        checkMeetingInfoError({body: {url: '1234567-wbxappapiabcdefg'}}, true);
+      });
+
```
marcin-bazyl, line 892: should we also have a test case "should return default meeting info lookup error payload if the error code is one of WbxAppAPI ones? or do we not care about the actual error code and only rely on data.meetingInfo or the url containing "wbxappapi" string? (this comment is related to my previous one [here](https://github.com/webex/webex-js-sdk/pull/3074/files#r1334594631))

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
nataliadelmar: ohh oki, what about using CA's non-fatal error 'StreamErrorNoMedia'

https://sqbu-github.cisco.com/WebExSquared/event-dictionary/wiki/Error-codes-for-metric-events#:~:text=3013,false

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
nataliadelmar: You said 'I think if we get an error from WbxAppApi meeting info, we will execute the return from line 401'. Actually only if we get any error THAT is in our SERVICE_ERROR_CODES_TO_CLIENT_ERROR_CODES_MAP, then we return line 401. But if we get an error that is not in that map for any chance, at least we can detect if it is a Locus error (lines 404-411) or WbxAppApi error (lines 413-418)

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.ts
nataliadelmar: The case in which the error code is one of the WbxAppAPI ones is covered in tests in line 823-837. 
image

Although the test is using code 2409005 which is a Locus code, it applies the same for WbxAppAPI bc these known codes are written in the same SERVICE_ERROR_CODES_TO_CLIENT_ERROR_CODES_MAP.  

If this map does not capture the error, then we at least try to see if we can identify it as a WbxAppAPI Meeting Info error and catalogue it as a general MEETING_INFO_LOOKUP_ERROR_CLIENT_CODE (error 4100).  That is what tests in line 885 - 896 are about.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
nataliadelmar: that is true, Ill add it to the code

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -364,6 +366,7 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
         error = merge(
           {fatal: true, shownToUser: false, name: 'other', category: 'other'}, // default values
           {errorCode: clientErrorCode},
+          {errorData: {errorName: serviceErrorName}},
```
torgeadelin, line 369: we should check that if serviceErrorName is not provided, we don't send 
```
{
  errorData :{
    errorName: undefiend
  }
}
```

torgeadelin: maybe add a condition after, to add the error.errorData.errorName

torgeadelin: ok, just checked the tests and indeed it does get sent like that. 
I'm not sure if `clearEmptyKeysRecursively` will remove it or not. Check the network logs, and if you don't see errorData at all, then leave it as is, but if you do see it, then add the condition i mentioned above

torgeadelin (changes requested):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
marcin-bazyl: ok, yeah, I misunderstood the intention of this change, after speaking offline I understand that this code is added just as a fallback if `SERVICE_ERROR_CODES_TO_CLIENT_ERROR_CODES_MAP` mapping doesn't catch the code.

marcin-bazyl (approved):

marcin-bazyl:
force merged as Catalin is on PTO and his comments were addressed

PR #3083, opened 2023-09-14
feat(): byok revoked error handling

shivani1211:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-459735

Error handling when an encryption key is revoked and kms throws 405 error codes

Added an event listener that gets triggered as soon as the specific error codes start coming in from revoking the key.

Screenshot 2023-09-07 at 10 04 04 AM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-encryption/src/constants.js
+++ b/packages/@webex/internal-plugin-encryption/src/constants.js
@@ -1,0 +1,3 @@
+export const KMS_KEY_REVOKE_FAILURE = 'event:kms:key:revoke:encryption:failure';
+export const KMS_KEY_REVOKE_ERROR_STATUS = 405;
+export const KMS_KEY_REVOKE_ERROR_CODES = [405005, 405006];
--- a/packages/@webex/internal-plugin-encryption/test/unit/spec/kms-errors.js
+++ b/packages/@webex/internal-plugin-encryption/test/unit/spec/kms-errors.js
@@ -1,0 +1,70 @@
+import {assert} from '@webex/test-helper-chai';
+import {handleKmsKeyRevokedEncryptionFailure} from '../../../src/kms-errors'
+import sinon from 'sinon';
+
+describe('handleKmsKeyRevokedEncryptionFailure', () => {
+
+    it('triggers `event:kms:key:revoke:encryption:failure` event when correct error code is detected', () => {
+        const webex = {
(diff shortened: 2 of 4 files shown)


packages/@webex/internal-plugin-encryption/src/constants.js
```diff
@@ -0,0 +1,3 @@
+export const KMS_KEY_REVOKE_FAILURE = 'event:kms:key:revoke:encryption:failure';
+export const KMS_KEY_REVOKE_ERROR_STATUS = 405;
+export const KMS_KEY_REVOKE_ERROR_CODE = 405005 || 405006;
```
chrisadubois, line 3: this variable should be a list... 

right now this is saying 

if (405005 || 405006) { KMS_KEY_REVOKE_ERROR_CODE = 405005 } and thus it will always be equal to 405005. 

instead we'd want a list to check if VALID_ERROR_CODES.includes(errorCode)
or something of the like

chrisadubois (changes requested):
blocking: needs some tests.

packages/@webex/internal-plugin-encryption/src/constants.js
shivani1211: ah I see, got it. Let me change that

packages/@webex/internal-plugin-encryption/test/unit/spec/kms-errors.js
```diff
@@ -0,0 +1,51 @@
...
+            body: {
+                errorCode: 403009,
+            }
+        }
+
+        handleKmsKeyRevokedEncryptionFailure(item2, webex);
+         
+         assert.notCalled(webex.internal.encryption.trigger);
+       });
+});
```
chrisadubois, line 51: It might be worth considering actually having a listener set up in another test that *actually* gets the event that the SDK triggered, i.e.,

webex.internal.encryption.on('event', () => {})

runAllTicks()

expect(funcToBeCalled)

but this might be overkill and i won't reject it based on this

chrisadubois (changes requested):

chrisadubois (approved):

robstax (approved):
lgtm! nj!

PR #3088, opened 2023-09-15
chore(docs): changes in deploy cd to push docs folder changes to docu…

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-455628 

Added steps in the deploy CD pipeline to build the docs folder and push changes to the documentation branch after SDK deployment.

Added GitHub action in deploy.yml file under .gihtub/worfflows folder for publishing of docs folder to the documentation branch.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


.github/workflows/deploy.yml
```diff
@@ -195,5 +197,47 @@ jobs:
...
+          path: '**/node_modules'
+          key: node-modules-${{ hashFiles('./yarn.lock') }}
+
+      - name: Uncache Distributables
+        uses: actions/cache@v3
+        with:
+          path: '**/dist'
+          key: dist-${{ env.rid }}
+
+      - name: Synchronize Packages
```
sreenara, line 225: So we need to do all these actions again is it? The folder on actions environment doesn't have everything set up between the stages?

arun3528: +1 you dont need all the steps

.github/workflows/deploy.yml
```diff
@@ -195,5 +197,47 @@ jobs:
...
+
+      - name: Set Git Identity
+        run: |
+          git config user.email "${GIT_AUTHOR_EMAIL}"
+          git config user.name "${GIT_AUTHOR_NAME}"
+
+      - name: Publish Docs Folder
+        run: |
+          git add docs/
+          git commit -m "chore(docs): update docs"
```
arun3528, line 242: i would add the versoin number if possible, Not sure if we want to do it here but only if webex package  changes then only we need to deploy the documentation we could optimise it later too

arun3528 (changes requested):

.github/workflows/deploy.yml
Kesari3008: yes tested it without steps so will remove it

mkesavan13 (approved):
Looks good to me. Approving.

However, just curious what does `github.actor` contain?

Neeraj-swarnkar:
> github.actor

@mkesavan13, 
The username of the user that triggered the initial workflow run. If the workflow run is a re-run, this value may differ from github.triggering_actor. Any workflow re-runs will use the privileges of github.actor, even if the actor initiating the re-run (github.triggering_actor) has different privileges.

https://docs.github.com/en/actions/learn-github-actions/contexts

.github/workflows/deploy.yml
Kesari3008: We are not publishing the Webex package yet so we will add this later

PR #3108, opened 2023-09-25
docs(samples): fix multistream sample app bugs

mkesavan13:
# COMPLETES [SPARK-459123](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-459123)

Issues in the Webex Web SDK Sample app related to Multistream meetings

### DIFF
--- a/docs/samples/browser-plugin-meetings/index.html
+++ b/docs/samples/browser-plugin-meetings/index.html
@@ -363,7 +363,10 @@
                     <input type="checkbox" name="ts-media-direction" value="videoEnabled" checked="">videoEnabled
                   </label>
                   <label>
-                    <input type="checkbox" name="ts-media-direction" value="receiveShare" checked="">receiveShare
+                    <input type="checkbox" name="ts-media-direction" value="shareAudioEnabled" checked="">shareAudioEnabled
+                  </label>
+                  <label>
+                    <input type="checkbox" name="ts-media-direction" value="shareVideoEnabled" checked="">shareVideoEnabled
                   </label>
                 </div>
               </div>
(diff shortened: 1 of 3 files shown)


sreenara (approved):

docs/samples/browser-plugin-meetings/index.html
```diff
@@ -363,7 +363,10 @@ <h2 class="collapsible">Meeting Management</h2>
                     <input type="checkbox" name="ts-media-direction" value="videoEnabled" checked="">videoEnabled
                   </label>
                   <label>
-                    <input type="checkbox" name="ts-media-direction" value="receiveShare" checked="">receiveShare
+                    <input type="checkbox" name="ts-media-direction" value="enableShareAudio" checked="">enableShareAudio
```
marcin-bazyl, line 366: This doesn't look right. the value is used in app.js as the name of the property that's then passed into calls to addMedia() and updateMedia(). So here the name is `enableShareAudio` and it matches what addMedia() takes, but updateMedia takes `shareAudioEnabled`. 
I think we should rename the prop that addMedia() takes to `shareAudioEnabled`, because that will fit more with the other props from addMedia like `audioEnabled` and `videoEnabled` and then also rename it here to `shareAudioEnabled`.
Same applies to share video.

mkesavan13: My bad. I thought they both were `enableShareAudio`

Will update and raise a PR. Thanks for noticing it @marcin-bazyl

mkesavan13: Done @marcin-bazyl

marcin-bazyl (approved):

PR #3142, opened 2023-10-09
fix(calling): detectNetwork change

karthikraji2020:
# COMPLETES #[< SPARK-458913 >](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-458913)

We’re calling detectNetworkChange method from calling the client’s [init function](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/CallingClient.ts#L123) after we create line dictionary by initialising line instance. [detectNetworkChange()](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/CallingClient.ts#L159) checks if we have any connection or mercury flap and then calls handleConnectionRestoration() once mercury is up again which internally tries to restore registration or trigger a new registration if not done already.

 

Now, What if the external user of the line hasn’t triggered the line.register() already and mercury connection flaps before it. detectNetworkChange will eventually try to [create the registration](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/CallingClient.ts#L184) even when registration didn’t happen for the first time.
 
Acceptance criteria:

Registration should not happen after the mercury connection is up again when it was not triggered even once.

We have introduced a new state: IDLE. Lines will be in IDLE state initially and will change to Active once registered. It will only be set back to IDLE if deregistration is done on purpose. In the detetctNetworkChange function, we have added a condition to check for line status. For cases where the line is in an IDLE state, handleConnectionRestorion will not be triggered and we also set the retry flag back to false.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Testcase 1: Triggered mercury flap before initial registration. registered the device after the connection was back up.
Testcase 2: Triggered mercury flap after the device is in registered state and restoration takes place.
Testcase 3: Triggered deregistration manually, mercury flap occurs and restoration flow isn't triggered.

Logs: [Mercury_Flap_After_Registration.log](https://github.com/webex/webex-js-sdk/files/13236960/Mercury_Flap_After_Registration.log)
[Mercury_Flap_Before_Initial_Registration.log](https://github.com/webex/webex-js-sdk/files/13236962/Mercury_Flap_Before_Initial_Registration.log)
[Uploading Mercury_Flap_After_Deregistration.log…]()

### I certified that

---


packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -142,7 +142,7 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
         retry = true;
       }
 
-      if (retry && this.webex.internal.mercury.connected) {
+      if (retry && this.webex.internal.mercury.connected && line.mobiusDeviceId) {
```
Kesari3008, line 145: Is there a possibility to add a UT for this by simulating this scenario? Also there maybe a possibility to test this manually also, have you tried to test it?

Kesari3008 (changes requested):

packages/calling/src/CallingClient/CallingClient.ts
karthikraji2020: ack: updated

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -509,6 +509,48 @@ describe('CallingClient Tests', () => {
       expect(registerSpy).not.toBeCalled();
     });
 
+    it('Simulate a network flap before initial registration is done should not trigger re-verify registration: Restore Failure', async () => {
+      expect(line.getRegistrationStatus()).toEqual(MobiusStatus.ACTIVE);
+
+      reg.deregister();
+      line.registration.deregister();
```
sreenara, line 516: Isn't reg = line.registration ? Why are we called line.registration.deregister() after this?
Also, why are we not doing line.deregister() directly?

sreenara (changes requested):
Have a comment in the UT

packages/calling/src/CallingClient/CallingClient.test.ts
karthikraji2020: Ref:[/CallingClient/line/index.ts#L148]  (https://github.com/webex/webex-js-sdk/blob/a57a4decebcf721da6495c398b566e9b803fb6c3/packages/calling/src/CallingClient/line/index.ts#L148)
 within the line Object we have `registration` module..
 
 [the way we are accessing the registration modules method] (https://github.com/webex/webex-js-sdk/blob/a57a4decebcf721da6495c398b566e9b803fb6c3/packages/calling/src/CallingClient/CallingClient.test.ts#L377)

karthikraji2020: > Isn't reg = line.registration ? Why are we called line.registration.deregister() after this? Also, why are we not doing line.deregister() directly?

Ref:[/CallingClient/line/index.ts#L148] (https://github.com/webex/webex-js-sdk/blob/a57a4decebcf721da6495c398b566e9b803fb6c3/packages/calling/src/CallingClient/line/index.ts#L148C1-L150C4)
 within the line Object we have `registration` module..
 
 [the way we are accessing the registration modules method](https://github.com/webex/webex-js-sdk/blob/a57a4decebcf721da6495c398b566e9b803fb6c3/packages/calling/src/CallingClient/CallingClient.test.ts#L377)

sreenara: I still don't understand why we are doing it this way.
Why can't we just have 1 line saying `line.deregister()`?

meta-dipanshu-sharma: @karthikraji2020

> so to access the deregister() we are using line.registration.deregister()

here, `reg == line.registration` as Sreekanth mentioned above so no need to separately call `line.registration.deregister()` to invoke deregister(). `reg.deregister()` will invoke the same function.

Use either of below not both,
1. `reg.deregister()`
2. `line.registration.deregister()`

karthikraji2020: ack: updated

sreenara (approved):

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -509,6 +509,47 @@ describe('CallingClient Tests', () => {
...
+      webex.internal.mercury.connected = true;
+      jest.advanceTimersByTime(NETWORK_FLAP_TIMEOUT + 500);
+
+      await flushPromises();
+
+      /* We should be detecting the network recovery */
+
+      expect(logSpy).not.toBeCalledWith(
+        'Mercury connection is up again, Re-registering with Mobius',
+        {file: REGISTRATION_FILE, method: 'handleConnectionRestoration'}
```
Kesari3008, line 541: Is it possible to spy on the handleConnectionRestoration method and check that it not getting invoked?

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -509,6 +509,47 @@ describe('CallingClient Tests', () => {
       expect(registerSpy).not.toBeCalled();
     });
 
+    it('Simulate a network flap before initial registration is done should not trigger re-verify registration: Restore Failure', async () => {
+      expect(line.getRegistrationStatus()).toEqual(MobiusStatus.ACTIVE);
+
+      reg.deregister();
+      line.mobiusDeviceId = undefined;
```
Kesari3008, line 516: We are specifically setting mobiusDeviceId as undefined and deregistering the line. This scenario should be where there were no registrations at all

Kesari3008 (approved):
Approving but UT apporach needs discussing

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -151,6 +150,7 @@ export class Registration implements IRegistration {
       },
     });
 
+    this.setStatus(RegistrationStatus.INACTIVE);
```
sreenara, line 153: Wouldn't it be better to move this to IDLE? We are deleting the registration on purpose here, so ideally we (developer) already know that the registration is inactive.
I feel INACTIVE is generally a state to be used when the registration has been deleted from the mobius side or we receive an error in registration.

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -592,7 +592,7 @@ export class Registration implements IRegistration {
     for (const url of servers) {
       try {
         abort = false;
-        this.registrationStatus = MobiusStatus.DEFAULT;
+        this.registrationStatus = RegistrationStatus.INACTIVE;
```
sreenara, line 595: Should this be INACTIVE or IDLE?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -760,7 +760,7 @@ export class Registration implements IRegistration {
     }
 
     this.clearKeepaliveTimer();
-    this.setStatus(MobiusStatus.DEFAULT);
+    this.setStatus(RegistrationStatus.INACTIVE);
```
sreenara, line 763: Same question here as in deleteRegistration

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -162,7 +163,11 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
         retry = true;
       }
 
-      if (retry && this.webex.internal.mercury.connected) {
+      if (
+        retry &&
+        this.webex.internal.mercury.connected &&
+        line.status !== RegistrationStatus.IDLE
+      ) {
         retry = await line.registration.handleConnectionRestoration(retry);
```
sreenara, line 171: One comment for the log inside handleConnectionRestoration. Can we change it to "Mercury connection is up again, re-registering with Webex Calling if needed."

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -118,7 +118,7 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
       logLevel
     );
 
-    this.registration.setStatus(MobiusStatus.DEFAULT);
+    this.registration.setStatus(RegistrationStatus.INACTIVE);
```
sreenara, line 121: Shouldn't this be IDLE?

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -189,6 +190,7 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
       case LINE_EVENTS.UNREGISTERED:
       case LINE_EVENTS.RECONNECTED:
       case LINE_EVENTS.RECONNECTING:
+        this.status = this.registration.getStatus();
```
sreenara, line 193: Why did we need this?

sreenara (changes requested):
Please see the comments. Tests also will require changes.

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: As per our discussion, IDLE status is removed for now. We only have Active and Inactive.

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Same comment as above

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Here it needed to be INACTIVE only

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Updated

packages/calling/src/CallingClient/line/index.ts
Kesari3008: Removed the status variable itself from the line object as discussed

meta-dipanshu-sharma:
@Kesari3008 I'm reviewing the PR but just have one question that what was the issue with the earlier fix when we added check for mobiusDeviceId on line when connection is up again?

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -131,7 +127,9 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
    */
   public async register() {
     await this.#mutex.runExclusive(async () => {
-      this.registration.setStatus(MobiusStatus.DEFAULT);
+      if (this.registration.getStatus() !== RegistrationStatus.IDLE) {
```
sreenara, line 130: let's remove this code and check what happens

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -118,7 +114,7 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
       logLevel
     );
 
-    this.registration.setStatus(MobiusStatus.DEFAULT);
+    this.registration.setStatus(RegistrationStatus.IDLE);
```
sreenara, line 117: this shouldn't be needed

sreenara (changes requested):
Please check the comments

packages/calling/src/CallingClient/registration/register.ts
sreenara: Clarified now.

packages/calling/src/CallingClient/registration/register.ts
sreenara: We can leave this as is

packages/calling/src/CallingClient/registration/register.ts
sreenara: As discussed, it needs to be IDLE here for a purposeful deregistration by the developer

sreenara (changes requested):
Resolved all comments that we discussed where no changes are needed.
I've only left the comments where the changes are required.

Also, for the failure registration cases, I think we should mark the status as INACTIVE in [this condition](https://github.com/webex/webex-js-sdk/blob/d0744a57a8d61eb2795097474aa2d528414665cb/packages/calling/src/CallingClient/registration/register.ts#L655). What do you think?

Kesari3008:
> @Kesari3008 I'm reviewing the PR but just have one question that what was the issue with the earlier fix when we added check for mobiusDeviceId on line when connection is up again? PR description still explains the old fix so asking here.

@meta-dipanshu-sharma I have updated the description please check. The reason why we cannot use deviceID in this condition because once we trigger a manual registration after the mercury flap, we could see restoration flow also kicking in and also device is never deleted from SDK. So deviceID will also be present in scenarios where we do not want to trigger restoration.

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: As we discussed yesterday, we cannot change the state to IDLE in deleteRegistration as this gets invoked in restoration and failback cases also

packages/calling/src/CallingClient/line/index.ts
Kesari3008: this is removed

packages/calling/src/CallingClient/line/index.ts
Kesari3008: removed..sorry missed this comment

sreenara (approved):

PR #3148, opened 2023-10-13
Stream classes add media fix

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-466805

1. Browser hangs when SDK tries to log the options passed into addMedia() when camera track has blur effect applied
2. Cannot join multistream meetings

1. Added toJSON method to camera and mic stream classes so that we only log useful fields
2. Updated internal-media-core to bring a fix for handling multiple negotiation needed events that can come before first offer is created

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

tested manually with the web app

### I certified that

---

### DIFF
--- a/packages/@webex/media-helpers/src/webrtc-core.ts
+++ b/packages/@webex/media-helpers/src/webrtc-core.ts
@@ -93,6 +93,16 @@ class _LocalMicrophoneStream extends WcmeLocalMicrophoneStream {
       this[LocalMicrophoneStreamEventNames.ServerMuted].emit(muted, reason);
     }
   }
+
+  toJSON() {
+    return {
+      id: this.id,
+      enabled: this.inputTrack?.enabled,
+      label: this.label,
+      readyState: this.readyState,
+      numEnabledEffects: this.getAllEffects().filter((item) => item.effect.isEnabled).length,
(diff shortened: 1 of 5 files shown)


packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -93,6 +93,16 @@ class _LocalMicrophoneStream extends WcmeLocalMicrophoneStream {
       this[LocalMicrophoneStreamEventNames.ServerMuted].emit(muted, reason);
     }
   }
+
+  toJSON() {
```
marcin-bazyl, line 97: I think long term we should have toJSON in webrtc-core LocalStream class, but until that's done, we can do it here. I've only done if to mic and camera (not share), because web app only does effects on these 2 and this is a temporary fix.

arun3528: what exactly are we loggin , cant the logging go inside the SDK ?

arun3528 (approved):

arun3528 (approved):

PR #3150, opened 2023-10-13
fix(samples): correct mute state on device change

meta-dipanshu-sharma:
# COMPLETES #< [SPARK-466442](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-466442) >

When we change the microphone device, if audio is muted, then with the next device, it starts with audio unmuted. Same for camera device.

We're setting previous mute state to new stream created before publishing the streams in the deviceChange api in samples app.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -1117,10 +1117,14 @@ function setVideoInputDevice() {
   const {video} = getAudioVideoInput();
 
   if (meeting) {
+    const isMuted = localMedia.cameraStream?.muted;
     localMedia.cameraStream?.stop();
 
     return getUserMedia({video})
-      .then(() => meeting.publishStreams({camera: localMedia.cameraStream}));
+      .then(() => {
+        localMedia.cameraStream.setMuted(!!isMuted);
+        meeting.publishStreams({camera: localMedia.cameraStream});
+      });
   }
   else {
     console.log('MeetingControls#setVideoInputDevice() :: no valid meeting object!');
@@ -1132,10 +1136,14 @@ function setAudioInputDevice() {
   const {audio} = getAudioVideoInput();
 
   if (meeting) {
+    const isMuted = localMedia.microphoneStream?.muted;
     localMedia.microphoneStream?.stop();
 
     return getUserMedia({audio})
-      .then(() => meeting.publishStreams({microphone: localMedia.microphoneStream}));
+      .then(() => {
+        localMedia.microphoneStream.setMuted(!!isMuted);
+        meeting.publishStreams({microphone: localMedia.microphoneStream});
+      });
   }
   else {
     console.log('MeetingControls#setAudioInputDevice() :: no valid meeting object!');


packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -74,9 +74,13 @@ export class MuteState {
   public init(meeting: any) {
     this.applyUnmuteAllowedToStream(meeting);
 
-    // if we are remotely muted, we need to apply that to the local stream now (mute on-entry)
-    if (this.state.server.remoteMute) {
-      this.muteLocalStream(meeting, this.state.server.remoteMute, 'remotelyMuted');
+    // if we are locally/remotely muted, we need to apply that to the local stream now (mute on-entry)
+    if (this.state.server.remoteMute || this.state.server.localMute) {
```
marcin-bazyl, line 78: This is not the right fix. `state.server.localMute` is supposed to reflect what server knows about the local mute state - it is driven by the muted property of the local webrtc-core track. The client needs to always make sure that server's localMute matches the client's local mute, not the other way around. 
Most likely the root cause of the problem from the jira is that when switching microphones, a new LocalMicrophoneStream is created by the sample app and it is always unmuted instead of being set to the same mute state as the old microphone (by calling setMuted() after it's created and before it's published). Once that's done, then the code in this function in lines 86-102 will read the new microphone's local mute state and will apply it to the server (in this case actually it won't send any request to the server, because local mute will already match the server's local mute state and that's fine)

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting/muteState.ts
arun3528: @marcin-bazyl  At this point of code the old steam is already gone so we don't know what was the mute state 

we only have server.localmute which should match localMute+muteonJOin from server I assume 

Or is that value server.remoteMute ?

what would be solution here ? before the actual switch happens during publish we update the mute state then publish internally

marcin-bazyl: @arun3528  and @meta-dipanshu-sharma, with the new webrtc-core classes it is the app's responsibility to set the right local mute state of the stream that they are publishing, in this case it's the sample app. The sample app needs to maintain the state of what the user wants to do - if user clicks the mute button, it means the user wants to be muted and if the user then changes the mic they most likely still want to be muted, so the app when creating a new LocalMicrophoneStream needs to call setMuted(true) on it so that user remains muted on the new mic. So when the new LocalMicrophoneStream is published SDK can read its local mute state and inform Locus about it.

marcin-bazyl (approved):

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1117,10 +1117,14 @@ function setVideoInputDevice() {
   const {video} = getAudioVideoInput();
 
   if (meeting) {
+    const isMuted = localMedia.cameraStream?.muted;
```
marcin-bazyl, line 1120: if the old stream was only remotely muted, `localMedia.cameraStream?.muted` will return true, so then we will do a local mute on the new stream, but I guess it's OK because either way server cannot unmute a participant if he/she is locally or remotely muted, so it won't make a difference

PR #3156, opened 2023-10-23
fix(plugin-meetings): do not sync meetings if unverified guest

Coread:

Guest users failed to reconnect because reconnect attempted to syncMeetings

Only syncMeetings during reconnect if user is not an unverified guest

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
+++ b/packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
@@ -443,17 +443,19 @@ export default class ReconnectionManager {
       }
     }
 
-    try {
-      LoggerProxy.logger.info(
-        'ReconnectionManager:index#executeReconnection --> Updating meeting data from server.'
-      );
-      await this.webex.meetings.syncMeetings();
-    } catch (syncError) {
-      LoggerProxy.logger.info(
-        'ReconnectionManager:index#executeReconnection --> Unable to sync meetings, reconnecting.',
-        syncError
-      );
-      throw new NeedsRetryError(syncError);
+    if (!this.webex.credentials.isUnverifiedGuest) {
+      try {
+        LoggerProxy.logger.info(
+          'ReconnectionManager:index#executeReconnection --> Updating meeting data from server.'
+        );
+        await this.webex.meetings.syncMeetings();
+      } catch (syncError) {
+        LoggerProxy.logger.info(
+          'ReconnectionManager:index#executeReconnection --> Unable to sync meetings, reconnecting.',
+          syncError
+        );
+        throw new NeedsRetryError(syncError);
+      }
     }
 
     // TODO: try to improve this logic as the reconnection manager saves the instance of deleted meeting object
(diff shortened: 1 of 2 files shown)


jjkusowski (approved):

packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
```diff
@@ -443,17 +443,19 @@ export default class ReconnectionManager {
...
-        'ReconnectionManager:index#executeReconnection --> Updating meeting data from server.'
-      );
-      await this.webex.meetings.syncMeetings();
-    } catch (syncError) {
-      LoggerProxy.logger.info(
-        'ReconnectionManager:index#executeReconnection --> Unable to sync meetings, reconnecting.',
-        syncError
-      );
-      throw new NeedsRetryError(syncError);
+    if (!this.webex.credentials.isUnverifiedGuest) {
```
chrisadubois, line 446: suggestion: I know this function is already kind of messy in general and does like 5 things in sequence, but as it's growing here and now there is conditional logic, would it not make sense to create a separate function, something like, `reconnectionSync(this.webex.credentials)`

where reconnectionSync looks like

`const reconnectionSync = (options: {isUnverifiedGuest: boolean, overrideSync, etc:}`

and it handles all of this type of stuff, that way this function can slowly become sort of a dispatcher and then the new function could be tested in isolation.

Obviously it's fine as is because we have access to webex.credentials, but just thinking about moving towards cleaning this function up.

packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
```diff
@@ -443,17 +443,19 @@ export default class ReconnectionManager {
...
+      try {
+        LoggerProxy.logger.info(
+          'ReconnectionManager:index#executeReconnection --> Updating meeting data from server.'
+        );
+        await this.webex.meetings.syncMeetings();
```
chrisadubois, line 451: question: Or instead of that, could we perhaps enhance syncMeetings such that if it fails it doesn't completely bork the rest of the flow? probably not due to the catch below, but a thought that maybe sync knows when it can and cannot sync meetings instead.

chrisadubois (approved):
Approved with a suggestion/question

packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
Coread: Possibly, but hard to do that without also changing the surrounding logging in this function. You don't want the logs to say "Updating meeting data from server." when syncMeetings is actually a no-op.

packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
Coread: At some point yes it would be nice to have this function call other functions rather than doing all this extra work, but given the lack of existing unit tests I feel more comfortable introducing minimal changes here rather than splitting up as part of this bug fix.

PR #2782, opened 2023-04-26
fix: reconnection fails when TURN discovery is skipped because of rea…

marcin-bazyl:
# COMPLETES #[SPARK-424966](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-424966)

## This pull request addresses

When TURN discovery was enabled in config and reachability returned some successful connections SDK was not sending the Roap message with empty mediaId on reconnection. This resulted in Locus not recreating a confluence and as a result Homer would not accept a new DTLS connection, so the reconnection was stuck in "connecting" forever.

## by making the following changes

Made sure that when Roap checks if TURN discovery is being done or not it checks all the conditions and not just the config.

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

manually tested end2end with sample app + unit tests

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3160, opened 2023-10-25
fix(plugin-meetings): cherry-picked-PR-2782-from-beta

Shreyas281299:
# COMPLETES #[SPARK-441781](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-441781)

While in a meeting if the internet is disconnected for some amount of time the remote video and audio stops. In samples page remote video shows "undefined 0fps". While the internet is disconnected if the connection status fails the newly created connection never reaches the 'connected' state it gets stuck in connecting. 

This is caused because while reconnection locus doesn't expect the media ID to be passed. We only take into account the config value for enableTurnDiscovery and doesn't take into account that TURN discovery is skipped if reachability succeeds. 

Made sure that when Roap checks if TURN discovery is being done or not it checks all the conditions and not just the config.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios  tested
After all the following scenarios, the meeting is successfully joined with two-way audio and video.

- While in a meeting VPN was turned on.
- While in a meeting VPN was turned off

With VPN
- Internet was disconnected and reconnected in a short period of time, here the connection state changes to disconnect.
- Internet was disconnected and reconnected in a period of time where the connection state  changes to failed..
- Internet was disconnected for a longer period of time where the meetting was disconnected and a rejoin happened.

Without VPN
- Internet was disconnected and reconnected in a short period of time, here the connection state changes to disconnect.
- Internet was disconnected and reconnected in a period of time where the connection state  changes to failed..
- Internet was disconnected for a longer period of time where the meetting was disconnected and a rejoin happened.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/roap/index.ts
+++ b/packages/@webex/plugin-meetings/src/roap/index.ts
@@ -287,18 +287,22 @@ export default class Roap extends StatelessWebexPlugin {
     // When reconnecting, it's important that the first roap message being sent out has empty media id.
     // Normally this is the roap offer, but when TURN discovery is enabled,
     // then this is the TURN discovery request message
-    const sendEmptyMediaId = reconnect && !meeting.config.experimental.enableTurnDiscovery;
+    return this.turnDiscovery
+      .isSkipped(meeting)
+      .then((isTurnDiscoverySkipped) => {
+        const sendEmptyMediaId = reconnect && isTurnDiscoverySkipped;
 
-    return this.roapRequest
-      .sendRoap({
-        roapMessage,
-        correlationId: meeting.correlationId,
-        locusSelfUrl: meeting.selfUrl,
-        mediaId: sendEmptyMediaId ? '' : meeting.mediaId,
-        audioMuted: meeting.audio?.isLocallyMuted(),
-        videoMuted: meeting.video?.isLocallyMuted(),
-        meetingId: meeting.id,
+        return this.roapRequest.sendRoap({
+          roapMessage,
+          correlationId: meeting.correlationId,
+          locusSelfUrl: meeting.selfUrl,
+          mediaId: sendEmptyMediaId ? '' : meeting.mediaId,
+          audioMuted: meeting.audio?.isLocallyMuted(),
+          videoMuted: meeting.video?.isLocallyMuted(),
+          meetingId: meeting.id,
+        });
       })
+
       .then(({locus, mediaConnections}) => {
         this.roapHandler.submit({
           type: ROAP.SEND_ROAP_MSG_SUCCESS,
--- a/packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
+++ b/packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
@@ -230,28 +230,14 @@ export default class TurnDiscovery {
    *                                 media connection just after a reconnection
    * @returns {Promise}
    */
-  doTurnDiscovery(meeting: Meeting, isReconnecting?: boolean) {
-    // @ts-ignore - fix type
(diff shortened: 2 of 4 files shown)


packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -287,14 +287,13 @@ export default class Roap extends StatelessWebexPlugin {
     // When reconnecting, it's important that the first roap message being sent out has empty media id.
     // Normally this is the roap offer, but when TURN discovery is enabled,
     // then this is the TURN discovery request message
-    const sendEmptyMediaId = reconnect && !meeting.config.experimental.enableTurnDiscovery;
```
sreenara, line 290: In the description, it is mentioned that the locus doesn't expect the media id during reconnection, which is causing the problem.
However, in the comment above, it's specified that the media id should NOT be sent. So it seems like this is expected behavior from the locus. This leads to the fact that this is a bug on our SDK.

Can we know why we had the check for enableTurnDiscovery and why we are removing that?

sreenara (changes requested):
Have a doubt about the fix

packages/@webex/plugin-meetings/src/roap/index.ts
marcin-bazyl: this fix is wrong, like the comment in the code says we need to send empty mediaId with the first Roap message that we send on reconnection - tha is the OFFER when we are not using TURN TLS or the TURN_DISCOVEEY_MESSAGE if we are using TURN TLS. This existing code on master is not correct, because it only takes into account the config value for `enableTurnDiscovery` and doesn't take into account that TURN discovery is skipped if reachability succeeds. This was fixed on beta branch in this PR: https://github.com/webex/webex-js-sdk/pull/2782 It is possible that you're hitting the same issue in SPARK-441781 (the logs attached there show that TURN discovery was enabled but skipped), so you need to port that fix or just start using beta branch.

marcin-bazyl (changes requested):
this is not the right fix, this would break scenarios when TURN TLS is used.

packages/@webex/plugin-meetings/src/roap/index.ts
sreenara: Thanks @marcin-bazyl . @Shreyas281299 I suggest we cherry pick the fix from beta and apply it here.

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -287,18 +287,23 @@ export default class Roap extends StatelessWebexPlugin {
...
-        meetingId: meeting.id,
+        return this.roapRequest.sendRoap({
+          roapMessage,
+          correlationId: meeting.correlationId,
+          locusSelfUrl: meeting.selfUrl,
+          mediaId: sendEmptyMediaId ? '' : meeting.mediaId,
+          audioMuted: meeting.audio?.isLocallyMuted(),
+          videoMuted: meeting.video?.isLocallyMuted(),
+          meetingId: meeting.id,
+          preferTranscoding: !meeting.isMultistream,
```
sreenara, line 303: multistream is not supported on `master` branch. Why is this parameter needed?

sreenara (changes requested):
Just one question regarding the change.

packages/@webex/plugin-meetings/src/roap/index.ts
Shreyas281299: I dont think this is required. Removed.

sreenara (approved):

PR #3163, opened 2023-10-25
fix(plugin-meetings): media settings toggles fixed

meta-dipanshu-sharma:
# COMPLETES # [SPARK-466438](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-466438)

Media settings toggle for audioEnabled is not working as expected while we toggle it and calls updateMedia().

1. Changed samples app logic to call updateMedia() only with the toggles which are changed by user and not all settings.
2. Added error logic in updateMedia() when shareAudioEnabled is toggled for a transcoded meeting.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -701,6 +701,7 @@ function updateMultistreamUI() {
   }
 }
 
+// NOTE: remember to set currentMediaSettings after promises resolves in the case when below method is used
 function getMediaSettings(compareLastSettings = false) {
   const settings = {};
 
@@ -708,7 +709,6 @@ function getMediaSettings(compareLastSettings = false) {
     if (compareLastSettings) {
       if (currentMediaSettings[options.value] !== options.checked) {
         settings[options.value] = options.checked;
-        currentMediaSettings[options.value] = options.checked;
       }
     } else {
       settings[options.value] = options.checked;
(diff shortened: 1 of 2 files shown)


mkesavan13 (approved):
Looks good to me

Neeraj-swarnkar (approved):
Looks Good.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2158,14 +2161,16 @@ function addMedia() {
     preferLiveVideoElm.value = 'Enable';
   }
 
+  currentMediaSettings = getMediaSettings();
```
marcin-bazyl, line 2164: this works, but I think it would be better to update `currentMediaSettings` inside `getMediaSettings()`, because otherwise in the future someone might add another call to `getMediaSettings()` and forget to update `currentMediaSettings`, in fact this is already happening, you have another call to `getMediaSettings()` in joinMeeting() where currentMediaSettings don't get updated and it works fine just by luck, because joinMeeting() then also calls addMedia(). If `currentMediaSettings` are supposed to always reflect the last applied media settings, it would be better to set them inside `getMediaSettings()`

marcin-bazyl: actually there is 1 more problem: if user toggles a setting that causes updateMedia() to fail, for example the audio share in a transcoded meeting, then the currentMediaSettings have the wrong value and it causes weird behaviour, because it doesn't take into account the fact that it failed so maybe the best thing to do is to set currentMediaSettings only in the then() clause everywhere after calling addMedia() or updateMedia()

meta-dipanshu-sharma: @marcin-bazyl Made changes as per suggestion. Also, created a method `setLastMediaSettingToggles()` which will also make the media settings toggles to previous state in case of failure.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -697,6 +705,24 @@ function getMediaSettings() {
   return settings;
 }
 
+// in case add/update media function fails then we need to restore the previous media setting toggles
+// reset will make all media toggles as default checked(in case of addMedia() failure as it is the initial case)
+function setLastMediaSettingToggles(reset, previousMediaSettings) {
```
marcin-bazyl, line 710: to me it feels like the code would be more readable if there were 2 separate functions:
1. resetMediaSettingsToDefaults() that does what calling `setLastMediaSettingToggles(true)` does
2. revertMediaSettings(changedMediaSettings) that does what calling `setLastMediaSettingToggles(false, changedSettings)` does

marcin-bazyl (approved):

docs/samples/browser-plugin-meetings/app.js
meta-dipanshu-sharma: Done

arun3528 (approved):

PR #3170, opened 2023-10-27
feat(ca-metrics): CA Share Error Metric

johnsoter13:
# COMPLETES #< [https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-410661](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-410661)
 >
- Issue where no errors are being reported by SDK for CA Share
- Because of this lack of reporting, we've been unaware when Sharing issues happen. There may be more changes that need to be made but this is a good first start.
- Added function to detect common sharing failures that web client/sdk sees
- Created error codes to define these new errors
- Updated meetings plugin to detect and report sharing errors

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


chburket (changes requested):
See comments.

chburket (approved):
looks good to me

arun3528 (approved):
Remove .only from the code rest looks good

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2557,20 +2557,62 @@ describe('plugin-meetings', () => {
...
+        let sandbox;
+
         beforeEach(() => {
           meeting.locusInfo.mediaShares = [{name: 'content', url: url1}];
           meeting.locusInfo.self = {url: url1};
           meeting.meetingRequest.changeMeetingFloor = sinon.stub().returns(Promise.resolve());
           meeting.mediaProperties.shareVideoTrack = {};
           meeting.mediaProperties.mediaDirection.sendShare = true;
           meeting.state = 'JOINED';
+          sandbox = sinon.createSandbox();
```
marcin-bazyl, line 2569: why do we need a new sandbox? sinon has a default sandbox, you can call `sinon.restore()` on it in afterEach(), then you don't need a separate sandbox

chburket: I was just following other examples. Will try without.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6088,6 +6088,17 @@ export default class Meeting extends StatelessWebexPlugin {
         body.resourceToken = resourceToken;
       }
 
+      // @ts-ignore
+      this.webex.internal.newMetrics.submitClientEvent({
```
marcin-bazyl, line 6092: just to double check: are we supposed to send this to CA before we are actually making the request and at that point we don't know if that request succeeded or not and I don't see anything being sent from the catch clause

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6224,6 +6246,18 @@ export default class Meeting extends StatelessWebexPlugin {
               stack: error.stack,
             });
 
+            // @ts-ignore
+            this.webex.internal.newMetrics.submitClientEvent({
+              name: 'client.share.floor-granted.local',
```
marcin-bazyl, line 6251: before sending the request we send "client.share.floor-grant.request" and here when it failed we send "client.share.floor-granted.local" - is that the right event? shouldn't it be "client.share.floor-grant.request" or something like "client.share.floor-grant.response"?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6088,6 +6088,17 @@ export default class Meeting extends StatelessWebexPlugin {
         body.resourceToken = resourceToken;
       }
 
+      // @ts-ignore
+      this.webex.internal.newMetrics.submitClientEvent({
+        name: 'client.share.floor-grant.request',
```
marcin-bazyl, line 6093: in stopWhiteboardShare() we are sending "client.share.stopped", but here we're sending "client.share.floor-grant.request" - the names of the events  don't seem to be very consistent, so just wanted to double check that this is what Call Analyzer wants, is there a spec somewhere that confirms this?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6201,6 +6212,17 @@ export default class Meeting extends StatelessWebexPlugin {
           options: {meetingId: this.id},
         });
 
+        // @ts-ignore
+        this.webex.internal.newMetrics.submitClientEvent({
```
marcin-bazyl, line 6216: this is sent basically at the exact same time as "client.share.initiated" from line 6207 above and with the same data, so it feels a bit suspicious if this is really the correct thing to do. Feels like maybe "client.share.initiated" should be sent somewhere earlier, like when the app calls publishTracks() with a screen share track or maybe should be sent from the web app when it actually calls getDisplayMedia()?

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: To my understanding, this metric specifically marks the instance that a request is sent, but doesn't indicate the success or failure of it. Something should probably be sent in the catch, but whiteboard failures are different from share failures. To make this more clear I will remove this metric for now since we were planning on handling whiteboard in a different epic.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: I think those examples might be old code from the times when sinon did require explicitly creating a sandbox

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: I was told that the result of this request should send `client.share.floor-granted.local`. This can either pass, or in the catch block, fail.

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?spaceKey=MS&title=Share+Analyzer+Details

This is the wiki I'm going off of. Are we certain that `client.share.stopped` should be sent the way it is in `stopWhiteboardShare()`? To me it seems like that would be the response of a successful request, in that case we probably should send `client.share.floor-grant.request` before the request. Again, will remove so we can discuss for a different time.

chburket (changes requested):
Currently reviewing some values with John

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: `client.share.initiated` should probably be sent from somewhere earlier.

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -691,6 +692,102 @@ const MeetingUtil = {
...
+
+    // return unknown error
+    return [
+      {
+        ...errorPayload,
+        fatal: true,
+        category: 'expected',
+        errorCode: 1100,
+      },
+    ];
```
chburket, line 789: Are all these errors actually "expected" and "fatal"?  For example, how can this unknown error be expected?

chburket (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
chburket: Worth noting that the 'client.share.initiated' submit was already there before John's changes. @marcin-bazyl do you have a preferred location between your two suggestions?

packages/@webex/plugin-meetings/src/meeting/index.ts
chburket: @johnsoter13 , told by Meera?

johnsoter13: Yes

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: I would say probably better to put it in publishTracks(), because then it will be sent for all other SDK clients too and we won't send it if user clicked "start share" button but then clicked cancel in the browser UI, but ultimately it's the CA team that should tell us when to send it

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: > This is the wiki I'm going off of.

cool, that wiki looks very useful. A few things I've noticed from it:
1. we need to generate the `shareInstanceId` and send it with all the share related events, without this, all of them will be ignored by CA. I don't see `shareInstanceId` being generated or sent in the current code or in this PR. Is the plan to do it in another PR or should it be added here?
2.  wiki has a client event `client.share.floor-accept.notification` - again don't see it in the code or in this PR, should it be added here or in another PR? actually I don't understand what's the difference between that event and `client.share.floor-granted.local` which we already send, maybe we don't need to send `client.share.floor-accept.notification`? Can you confirm with Meera?
3. wiki mentions also `client.media.tx.start` and `client.media.tx.stop` events. We currently don't send them for share - see the calls to emitStartStopEvents() in statsAnalyzer/index.ts - I think we will need to get that fixed, maybe should be a separate PR so that this one doesn't grow too big, please raise a Jira for this and add to the epic

> Again, will remove so we can discuss for a different time.
ok, but then do we have a jira for handling metrics for whiteboard? I guess it should still be in the same "share" epic

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: just to add more info for you when you talk to CA team about this, the flow is like this:
1. user clicks "start share"
2. app calls browser API getDisplayMedia()
3. browser displays UI to the user to choose what they want to share - they can also tick a box about audio sharing and they can also just click cancel at this point and the flow ends then
4. app calls SDK's Meeting.publishTracks() with a screen share track
5. SDK gives the track to WCME - WCME will tell Homer that we have a share track for sending
6. SDK asks Locus for the floor via an http request
7. Locus tells us that floor is granted via a Mercury message

If we put 'client.share.initiated' in publishTracks() then that's point 4 in the flow above

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: ok, sounds good. For the success case there is already existing code that sends `client.share.floor-granted.local` - see setUpLocusMediaSharesListener(). This place here in the catch looks good for that event. Not sure if we need to send the event with some error also in the case if changeMeetingFloor() succeeds but later we don't get the floor granted confirmation via Mercury

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: 1. This wasn't discussed with Meera but you're totally right, I will ask about this.
2. `client.share.floor-accept.notification` is mostly used for whiteboard, I was told. So we can come back to that for the future epic
3. `client.media.tx.start` and `client.media.tx.stop` are being sent in `setupStatsAnalyzerEventHandlers()`, should they be sent from somewhere else instead?

johnsoter13: I will create follow up PR with shareinstanceid work

marcin-bazyl: > client.media.tx.start and client.media.tx.stop are being sent in setupStatsAnalyzerEventHandlers(), should they be sent from somewhere else instead?

yes, but if you follow the code, you'll see that they're sent when stats analyzer emits LOCAL_MEDIA_STARTED/LOCAL_MEDIA_STOPPED events and these are only sent from emitStartStopEvents() and that function is only called for main audio and video, but not for share

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7119,15 +7110,22 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     let floorRequestNeeded = false;
+    if (tracks.screenShare?.video || (this.isMultistream && tracks.screenShare?.audio)) {
+      // @ts-ignore
+      this.webex.internal.newMetrics.submitClientEvent({
```
marcin-bazyl, line 7115: in theory someone can call publishTracks() twice, first time just with the video share track, second time with audio share track - in that case we will send 'client.share.initiate' twice and we probably want to sent it just once. I guess we just want to send it when floorRequestNeeded is true

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: Updated https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-471989 which is in the follow up epic

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5245,7 +5263,7 @@ export default class Meeting extends StatelessWebexPlugin {
       // @ts-ignore
       this.webex.internal.newMetrics.submitClientEvent({
         name: 'client.media.rx.start',
-        payload: {mediaType: data.type},
+        payload: {mediaType: data.type, shareInstanceId: this.shareInstanceId},
```
marcin-bazyl, line 5266: we shouldn't be sending the event with shareInstanceId for main audio and main video, only for share, same for "client.media.rx.stop"

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5216,7 +5230,9 @@ export default class Meeting extends StatelessWebexPlugin {
       // @ts-ignore
       this.webex.internal.newMetrics.submitClientEvent({
         name: 'client.media.tx.start',
-        payload: {mediaType: data.type},
+        payload: {
```
marcin-bazyl, line 5233: aren't we supposed to add `shareInstanceId` for "client.media.tx.start" and "client.media.tx.stop" too? (but only for "share" mediaType

packages/@webex/plugin-meetings/src/meeting/util.ts
chburket: Still wondering about this.

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: For the case where we don't get confirmation from Mercury, we can handle that case in a later PR.

johnsoter13: Open to discussion on this!

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -692,6 +693,102 @@ const MeetingUtil = {
...
+          errorCode: 1100,
+        },
+      ];
+    }
+    if (reason.includes(LOCAL_SHARE_ERRORS.DEVICE_NOT_JOINED)) {
+      return [
+        {
+          ...errorPayload,
+          fatal: true,
+          category: 'expected',
```
marcin-bazyl, line 726: why is it expected?

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -692,6 +693,102 @@ const MeetingUtil = {
...
+          errorCode: 2048,
+        },
+      ];
+    }
+    if (reason.includes(LOCAL_SHARE_ERRORS.NO_CONFLUENCE_ID)) {
+      return [
+        {
+          ...errorPayload,
+          fatal: true,
+          category: 'expected',
```
marcin-bazyl, line 746: why is this expected?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl, line 1: I can't review this file, because of all the formatting changes. Where's the actual change?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -2474,6 +2483,8 @@ export default class Meeting extends StatelessWebexPlugin {
         switch (newShareStatus) {
           case SHARE_STATUS.REMOTE_SHARE_ACTIVE: {
             const sendStartedSharingRemote = () => {
+              this.shareInstanceId = contentShare.shareInstanceId;
```
marcin-bazyl, line 2486: I'm not convinced if it's a good idea to be setting `Meeting.shareInstanceId` here as we're setting it to the remote share instance. I'm worried that there may be some race conditions if someone starts sharing at the same time as we are and this may overwrite the value that we set in `publishStreams()`. After all I think we only use `Meeting.shareInstanceId` for sending events about our share so I don't think we need to store the remote share instance id at all in `Meeting.shareInstanceId`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -2563,6 +2575,8 @@ export default class Meeting extends StatelessWebexPlugin {
       } else if (newShareStatus === SHARE_STATUS.REMOTE_SHARE_ACTIVE) {
         // if we got here, then some remote participant has stolen
         // the presentation from another remote participant
+        this.shareInstanceId = contentShare.shareInstanceId;
```
marcin-bazyl, line 2578: same as above

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1232,6 +1233,14 @@ export default class Meeting extends StatelessWebexPlugin {
...
     this.keepAliveTimerId = null;
 
+    /**
+     * id for tracking Share instances in Call Analyzer
+     * @instance
+     * @type {String}
+     * @private
+     * @memberof Meeting
+     */
+    this.shareInstanceId = null;
```
marcin-bazyl, line 1243: the annotation code added by Marmot team also seems to be using shareInstanceId. I'm not sure where they get it from, but it might be worth to just double check that to avoid some duplication and to make sure that we use the same shareInstanceId in our CA events as they do in annotation code.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5672,7 +5686,10 @@ export default class Meeting extends StatelessWebexPlugin {
       // @ts-ignore
       this.webex.internal.newMetrics.submitClientEvent({
         name: 'client.media.tx.start',
-        payload: {mediaType: data.type},
+        payload: {
+          mediaType: data.type,
+          shareInstanceId: data.type === 'share' && this.shareInstanceId,
```
marcin-bazyl, line 5691: nitpick: this will put `shareInstanceId: false` in the event for types other than "share" (so for main audio and video). Are we sure that Call Analyzer will be OK with it? I guess they just ignore it, but maybe it would be better to set it to undefined for those cases?

packages/@webex/plugin-meetings/src/meeting/util.ts
johnsoter13: for this error, we aren't entirely sure what the category should be. In this case, should 'signaling' be used?

packages/@webex/plugin-meetings/src/meeting/util.ts
johnsoter13: I'm assuming this error occurs when a user attempts to share before their device connects? I'm assuming it's expected that this could happen?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
johnsoter13: Yeah prettier formatted this whole file, I'll try to fix it for review

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: What do you suggest we do for `client.media.rx.start` and `client.media.rx.stop` then? We would want the contentshare.instanceId in that case right? Not a super big problem for now, since this metric isn't turned on for share.

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: Looks like annotation is getting it based in from the client, which is getting it from selectShareInstanceId (from the activeMeeting). This shareinstanceId matches the one we get from contentShare.shareInstanceId.

Maybe a solution could be having transmitShareInstanceId and a remoteShareInstanceId, that way we won't have to worry about any race conditions.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
johnsoter13: Wasn't able to do it, I'll just post here the tests that were changed

requestScreenShareFloor
publishStreams

packages/@webex/plugin-meetings/src/meeting/util.ts
marcin-bazyl: the web app should make sure that you can only share after you've joined, so this is not expected - if this happens it's a bug in the web app (or sdk) that needs fixing

packages/@webex/plugin-meetings/src/meeting/util.ts
marcin-bazyl: yeah ok, I'm not sure either, but definitely not expected, signaling is fine

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yes, you're right we need it in that case, but it would be better if we store it as a separate `remoteShareInstanceId` for that, or maybe we don't even need to to store it separately, because I think this stuff is already stored - see MediaSharesUtils.getShareInstanceId()

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yeah, I agree, happy with localShareInstanceId and remoteShareInstanceId, or just localShareInstanceId and existing code like MediaSharesUtils.getShareInstanceId() used for the remote one

marcin-bazyl (approved):
Approving, so that you don't have to wait for me another day due to timezones, but please change the category from "expected" to "signaling" for DEVICE_NOT_JOINED error

chburket (approved):

PR #3171, opened 2023-10-27
fix: added metric for succesful log upload

marcin-bazyl:
We often see logs missing in mats.webex.com, we have a metric for log upload failure, but not for success.

Also, noticed the metric for failure was always sending meetingId=undefined


chrisadubois (approved):
https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/5039

shnaaz (approved):

PR #3172, opened 2023-10-30
fix: improved logging of receive slots, muted local audio by default in the sample app

marcin-bazyl:

Logs from RemoteMediaManager for initial layout didn't include the SSRC values for the receive slots, because they are only possible to obtain from WCME after initial SDP offer-answer is exchanged.

In the sample app, by default local audio is always played and this causes unpleasant audio feedback, most of the time you don't really want to play your local audio.

- improved logging
- muted local audio in sample app (same fix was already done on SDK master by Arun a few months ago)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

manually used the sample app to check logs

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5601,6 +5601,11 @@ export default class Meeting extends StatelessWebexPlugin {
             meetingId: this.id,
           },
         });
+        LoggerProxy.logger.info(
+          `${LOG_HEADER} successfully established media connection, type=${connectionType}`
+        );
+
+        this.remoteMediaManager?.logAllReceiveSlots();
       })
       .catch((error) => {
         Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.ADD_MEDIA_FAILURE, {
@@ -5642,10 +5647,7 @@ export default class Meeting extends StatelessWebexPlugin {
             this.unsetPeerConnections();
           }
 
-          LoggerProxy.logger.error(
-            `${LOG_HEADER} Error adding media failed to initiate PC and send request, `,
-            error
-          );
+          LoggerProxy.logger.error(`${LOG_HEADER} failed to establish media connection: `, error);
 
           // Upload logs on error while adding media
           Trigger.trigger(
--- a/packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
+++ b/packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
@@ -473,6 +473,8 @@ export class RemoteMediaManager extends EventsScope {
     if (!this.started) {
       throw new Error('setLayout() called before start()');
     }
+    LoggerProxy.logger.log(`RemoteMediaManager#setLayout --> new layout selected: ${layoutId}`);
+
     this.currentLayoutId = layoutId;
     this.currentLayout = cloneDeep(this.config.video.layouts[this.currentLayoutId]);
 
@@ -725,10 +727,12 @@ export class RemoteMediaManager extends EventsScope {
   /**
    * Logs the state of the receive slots
    */
-  private logReceieveSlots() {
+  private logMainVideoReceiveSlots() {
     let logMessage = '';
     forEach(this.receiveSlotAllocations.activeSpeaker, (group, groupName) => {
-      logMessage += `group: ${groupName}\n${group.slots.map((slot) => slot.logString).join(' ')}`;
+      logMessage += `\ngroup: ${groupName}\n${group.slots
+        .map((slot) => slot.logString)
+        .join(', ')}`;
     });
 
     logMessage += '\nreceiverSelected:\n';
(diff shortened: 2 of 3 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5642,10 +5647,7 @@ export default class Meeting extends StatelessWebexPlugin {
             this.unsetPeerConnections();
           }
 
-          LoggerProxy.logger.error(
-            `${LOG_HEADER} Error adding media failed to initiate PC and send request, `,
-            error
-          );
+          LoggerProxy.logger.error(`${LOG_HEADER} failed to establish media connection: `, error);
```
chrisadubois, line 5650: We should probably move this logger up above before the metrics call, or right after, but at the very least, before the stop analyzer and stop media connection stuff

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5601,6 +5601,11 @@ export default class Meeting extends StatelessWebexPlugin {
             meetingId: this.id,
           },
         });
+        LoggerProxy.logger.info(
+          `${LOG_HEADER} successfully established media connection, type=${connectionType}`
```
chrisadubois, line 5605: nice

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -473,6 +473,8 @@ export class RemoteMediaManager extends EventsScope {
     if (!this.started) {
       throw new Error('setLayout() called before start()');
     }
+    LoggerProxy.logger.log(`RemoteMediaManager#setLayout --> new layout selected: ${layoutId}`);
```
chrisadubois, line 476: nice

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -725,10 +727,12 @@ export class RemoteMediaManager extends EventsScope {
   /**
    * Logs the state of the receive slots
    */
-  private logReceieveSlots() {
+  private logMainVideoReceiveSlots() {
     let logMessage = '';
     forEach(this.receiveSlotAllocations.activeSpeaker, (group, groupName) => {
-      logMessage += `group: ${groupName}\n${group.slots.map((slot) => slot.logString).join(' ')}`;
+      logMessage += `\ngroup: ${groupName}\n${group.slots
```
chrisadubois, line 733: i kinda hate that we add new lines into it. every log should be a singular line IMO as this makes the formatting much easier. instead, using a series of other delimiters would be good, but this is obviously outside the scope of this PR...

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -737,7 +741,28 @@ export class RemoteMediaManager extends EventsScope {
...
 
     LoggerProxy.logger.log(
-      `RemoteMediaManager#updateVideoReceiveSlots --> receive slots updated: unused=${this.slots.video.unused.length}, activeSpeaker=${this.slots.video.activeSpeaker.length}, receiverSelected=${this.slots.video.receiverSelected.length}\n${logMessage}`
+      `RemoteMediaManager#logMainVideoReceiveSlots --> MAIN VIDEO receive slots: unused=${this.slots.video.unused.length}, activeSpeaker=${this.slots.video.activeSpeaker.length}, receiverSelected=${this.slots.video.receiverSelected.length}${logMessage}`
+    );
+  }
+
+  /** Logs all current receive slots */
+  public logAllReceiveSlots() {
+    LoggerProxy.logger.log(
```
chrisadubois, line 750: Shouldn't we define a series of other functions while we're in here, called 

`logMainAudioReceiveSlots` and `logSlidesVideoReceiveSlots` and `logSlidesAudioReceiveSlots` and have this function just sort of call each of those (would be nice if it could be done in parallel but oh well) 

that way we'd have some symmetry, and a singular point for each that could be used, as logMainVideoReceiveSlots has?

chrisadubois (approved):
None of my comments are blocking, but please take a look

Coread (approved):

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: yeah first time I saw these multiline logs I also thought that's a bad idea, but then it actually makes life easier, they're very easy to read this way so I'm not 100% convinced on this, we can discuss within the wider team and update in another PR

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: log moved to the top of the catch

PR #3174, opened 2023-10-30
spark(470922): add missing meetingInfo.request metric on fetchMeetingInfo

nataliadelmar:
# COMPLETES #

## [SPARK-470922](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-470922)

meetingInfoV2 public method fetchMeetingInfo() called the client metrics `client.meetinginfo.response` on erroring. It forgot to call the client metric `client.meetinginfo.request` before fetching/requesting meeting info. 

- Adding submitClientEvent('client.meetinginfo.request') to fetchMeetingInfo method
- - Adding submitClientEvent('client.meetinginfo.response') to fetchMeetingInfo method when the request comes back successful

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/test/unit/spec/meeting-info/meetinginfov2.js
```diff
@@ -374,11 +374,11 @@ describe('plugin-meetings', () => {
...
-          {errorCode: 403104},
-          {errorCode: 403103},
-          {errorCode: 403048},
-          {errorCode: 403102},
-          {errorCode: 403101},
+          // {errorCode: 403104},
+          // {errorCode: 403103},
+          // {errorCode: 403048},
+          // {errorCode: 403102},
+          // {errorCode: 403101},
```
shnaaz, line 381: Should uncomment these

shnaaz (approved):

packages/@webex/plugin-meetings/test/unit/spec/meeting-info/meetinginfov2.js
```diff
@@ -95,7 +95,7 @@ describe('plugin-meetings', () => {
       meetingInfo = new MeetingInfo(webex);
     });
 
-    describe('#fetchMeetingInfo', () => {
+    describe.only('#fetchMeetingInfo', () => {
```
marcin-bazyl, line 98: .only

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -340,6 +340,12 @@ export default class MeetingInfoV2 {
       this.webex.internal.newMetrics.submitInternalEvent({
         name: 'internal.client.meetinginfo.request',
       });
+      this.webex.internal.newMetrics.submitClientEvent({
```
marcin-bazyl, line 343: what about v1 meeting info? I think we need a similar code change there too - see requestFetchInfo() in meeting-info/index.ts

nataliadelmar: do we still use v1 meeting info? in which cases?

shnaaz: If meeting is non converged then we will have v1 meeting info

marcin-bazyl: it's not about being converged, it's about the "web-meeting-wbxappapi-meetinginfo-enabled-v2" feature toggle - some customers still have it disabled and also other SDK clients that may still use `enableUnifiedMeetings=false` in sdk config.

packages/@webex/plugin-meetings/src/meeting-info/index.ts
```diff
@@ -71,20 +71,32 @@ export default class MeetingInfo {
    * @memberof MeetingInfo
    */
   private requestFetchInfo(options: any) {
-    const {meetingId} = options;
-    if (meetingId) {
+    const {meetingId, hasPrejoinStarted} = options;
+    if (meetingId && hasPrejoinStarted) {
```
marcin-bazyl, line 75: are we sure that we want to send both 'internal.client.meetinginfo.request' and 'client.meetinginfo.request' only if `hasPrejoinStarted` is true? If that's the case, then it looks like we are sending these 2 events in all the same places so why do we need the `internal.client.meetinginfo.request` at all? We could just update getMeetingInfoReqResp() to use 'client.meetinginfo.request' and 'client.meetinginfo.response' instead of the internal events.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1287,10 +1287,12 @@ export default class Meeting extends StatelessWebexPlugin {
     password = null,
     captchaCode = null,
     extraParams = {},
+    hasPrejoinStarted = false,
```
marcin-bazyl, line 1290: we need to update the jsdoc above with the new param, but this also makes me wonder what should we say in those docs for all the other apps that use JS-SDK, maybe we could avoid adding this param to the public API and know that we're in "pre-join" phase?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1219,12 +1219,19 @@ export default class Meetings extends WebexPlugin {
 
       if (enableUnifiedMeetings && !isMeetingActive && useRandomDelayForInfo && waitingTime > 0) {
         meeting.fetchMeetingInfoTimeoutId = setTimeout(
-          () => meeting.fetchMeetingInfo({extraParams: infoExtraParams}),
+          () =>
+            meeting.fetchMeetingInfo({
+              extraParams: infoExtraParams,
+              hasPrejoinStarted: !!correlationId, // to determine if we should submit Call Analyzer events
```
marcin-bazyl, line 1225: currently web app doesn't pass `correlationId` to create() when user presses "share on device" button when paired with an endpoint. I'm not sure if we want to send the meeting info events to call analyzer in this case too, or not. Can you check this with the CA team, please?

packages/@webex/plugin-meetings/src/meeting-info/index.ts
nataliadelmar: changing to sending internal.client.meetinginfo.request as before: if there is meeting.id

nataliadelmar: internal.client.meetinginfo.request tracks latencies of meeting info req/res whether in prejoin or in join phase. So it should always be sent as long as there is a meeting.id

client.meetinginfo.request is only for prejoin, so should only be sent if there is a correlation id and a meeting.id

packages/@webex/plugin-meetings/src/meetings/index.ts
nataliadelmar: CA team answered: "I believe so [that we do not need to send non-meeting content sharing to CA]. If there is a sample flow that we can identify in production we can check if share analyzer is picking these up"

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -741,7 +741,7 @@ describe('plugin-meetings', () => {
           );
         });
       });
-      describe('#join', () => {
+      describe.skip('#join', () => {
```
marcin-bazyl, line 744: .skip

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -828,7 +828,7 @@ describe('plugin-meetings', () => {
           it('should send Meeting Info CA events if meetingInfo is not empty', async () => {
```
marcin-bazyl, line 828: maybe this test title needs updating now

packages/@webex/plugin-meetings/src/meetings/index.ts
nataliadelmar: Answer from CA team: given share on device is a separate join on its own and it sens join CA events (client.call.initiated, client.locus.join.request, client.locus.join.response, client.call.leave), then correlationId should be generated when the user initiates share on device action. And if meetingInfo request response occurs at this time, those events should contain the correlationId that was generated for this user action

marcin-bazyl (approved):
approving, but I think we will now be sending "client.meetinginfo.request", "client.meetinginfo.response" twice, because of the code in Meeting.join(). Let's check with @torgeadelin on Monday and this can be addressed in another PR

packages/@webex/plugin-meetings/src/meeting-info/index.ts
```diff
@@ -71,11 +71,19 @@ export default class MeetingInfo {
    * @memberof MeetingInfo
    */
   private requestFetchInfo(options: any) {
-    const {meetingId} = options;
+    const {meetingId, sendCAevents} = options;
     if (meetingId) {
```
torgeadelin, line 75: in this case, check if sendCAevents is true and send both, removing the meetingId

packages/@webex/plugin-meetings/src/meeting-info/index.ts
```diff
@@ -96,21 +112,25 @@ export default class MeetingInfo {
...
-            identifiers: {
-              meetingLookupUrl: error?.url,
-            },
-          },
-          options: {
-            meetingId,
-            rawError: error,
-          },
-        });
+        if (meetingId) {
```
torgeadelin, line 115: same as above

torgeadelin: for both internal and real response

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -340,6 +340,14 @@ export default class MeetingInfoV2 {
       this.webex.internal.newMetrics.submitInternalEvent({
         name: 'internal.client.meetinginfo.request',
       });
+      if (sendCAevents) {
```
torgeadelin, line 343: same as above

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -349,6 +357,14 @@ export default class MeetingInfoV2 {
           this.webex.internal.newMetrics.submitInternalEvent({
             name: 'internal.client.meetinginfo.response',
           });
+          if (sendCAevents) {
```
torgeadelin, line 360: same as above

packages/@webex/plugin-meetings/src/meeting-info/index.ts
```diff
@@ -85,6 +93,14 @@ export default class MeetingInfo {
           this.webex.internal.newMetrics.submitInternalEvent({
             name: 'internal.client.meetinginfo.response',
           });
+          if (sendCAevents) {
+            this.webex.internal.newMetrics.submitClientEvent({
+              name: 'client.meetinginfo.response',
```
torgeadelin, line 98: needs meetingLookupUrl

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1219,12 +1219,19 @@ export default class Meetings extends WebexPlugin {
 
       if (enableUnifiedMeetings && !isMeetingActive && useRandomDelayForInfo && waitingTime > 0) {
         meeting.fetchMeetingInfoTimeoutId = setTimeout(
-          () => meeting.fetchMeetingInfo({extraParams: infoExtraParams}),
+          () =>
+            meeting.fetchMeetingInfo({
+              extraParams: infoExtraParams,
+              sendCAevents: !!correlationId, // to determine if we should submit Call Analyzer events
```
torgeadelin, line 1225: a more explicit comment would be nice

torgeadelin (changes requested):
small changes required

packages/@webex/plugin-meetings/src/meetings/index.ts
torgeadelin: you forgot to add better comment

torgeadelin (approved):

PR #3176, opened 2023-10-31
fix(plugin-meetings): start transcription when guest user is admitted

meta-dipanshu-sharma:
# COMPLETES #< [SPARK-466845](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-466845) >

Transcriptions do not work if user get placed in lobby and then is admitted into the meeting

Issue was happening because transcription method requires a `datachannelUrl` to generate a `webSocketUrl` which is then used to create transcription object. Until guest user is not admitted into the meeting, `datachannelUrl` is not generated and so the method fails to start the transcription.

As a part of this issue fix, we have an event triggered when guest is admitted into the meeting which is "SELF_ADMITTED_GUEST" so added the logic there to retry for transcription if not started earlier.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

After this fix, following 2 scenarios will start working which were not working previously. Below scenarios are tested on web SDK samples app.
1. When guest user wants to join meeting with transcription:
  1.1 guest user registers after putting access_token
  1.2 guest user clicks on `startReceivingTranscription()` in the samples before joining a meeting or sends 
       `receiveTranscription` as true in join options
  1.3 guest user is placed in the lobby until host admits guest in the meeting
  1.4 if Webex Assistant is enabled for the meeting, guest user will start receiving transcription once admitted in the 
        meeting which was not the case before this fix.

2. When guest user wants to start transcription after joining a meeting:
  2.1 guest user registers after putting access_token. For this case, user initialises webex with a config object where he 
       needs to add a property which is `receiveTranscription: true`. User clicks on join meeting
  2.2 guest user is placed in the lobby until host admits guest in the meeting
  2.3 host admits the user in the meeting. At this point, webex assistant is not enabled so user won't receive any 
        transcription. Once host enables webex assistant, user will start receiving transcription which was not the case 
        previously

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -462,6 +462,7 @@ export default class Meeting extends StatelessWebexPlugin {
   shareStatus: string;
   statsAnalyzer: StatsAnalyzer;
   transcription: Transcription;
+  receiveTranscription: boolean;
   updateMediaConnections: (mediaConnections: any[]) => void;
   endCallInitiateJoinReq: any;
   endJoinReqResp: any;
@@ -1761,8 +1762,8 @@ export default class Meeting extends StatelessWebexPlugin {
       LOCUSINFO.EVENTS.CONTROLS_MEETING_TRANSCRIBE_UPDATED,
       ({caption, transcribing}) => {
         // @ts-ignore - config coming from registerPlugin
-        if (transcribing && this.transcription && this.config.receiveTranscription) {
-          this.receiveTranscription();
+        if (transcribing && !this.transcription && this.config.receiveTranscription) {
+          this.startTranscription();
         } else if (!transcribing && this.transcription) {
           Trigger.trigger(
             this,
@@ -2261,8 +2262,17 @@ export default class Meeting extends StatelessWebexPlugin {
         });
       }
     });
-    this.locusInfo.on(LOCUSINFO.EVENTS.SELF_ADMITTED_GUEST, (payload) => {
+    this.locusInfo.on(LOCUSINFO.EVENTS.SELF_ADMITTED_GUEST, async (payload) => {
       this.stopKeepAlive();
+      // @ts-ignore
+      if (!this.transcription && (this.config.receiveTranscription || this.receiveTranscription)) {
+        if (this.isTranscriptionSupported()) {
+          await this.startTranscription();
+          LoggerProxy.logger.info(
+            'Meeting:index#setUpLocusInfoSelfListener --> enabled to receive transcription for guest user!'
+          );
+        }
+      }
 
       if (payload) {
         Trigger.trigger(
@@ -3725,9 +3735,9 @@ export default class Meeting extends StatelessWebexPlugin {
    * @private
    * @returns {Promise<void>} a promise to open the WebSocket connection
    */
-  private async receiveTranscription() {
+  private async startTranscription() {
     LoggerProxy.logger.info(
-      `Meeting:index#receiveTranscription -->
+      `Meeting:index#startTranscription -->
       Attempting to generate a web socket url.`
     );
 
@@ -3744,7 +3754,7 @@ export default class Meeting extends StatelessWebexPlugin {
       });
 
       LoggerProxy.logger.info(
-        `Meeting:index#receiveTranscription -->
+        `Meeting:index#startTranscription -->
         Generated web socket url succesfully.`
       );
 
@@ -3756,7 +3766,7 @@ export default class Meeting extends StatelessWebexPlugin {
       );
 
       LoggerProxy.logger.info(
-        `Meeting:index#receiveTranscription -->
+        `Meeting:index#startTranscription -->
         opened LLM web socket connection successfully.`
       );
 
@@ -3777,7 +3787,7 @@ export default class Meeting extends StatelessWebexPlugin {
       // @ts-ignore - fix type
       this.transcription.connect(this.webex.credentials.supertoken.access_token);
     } catch (error) {
-      LoggerProxy.logger.error(`Meeting:index#receiveTranscription --> ${error}`);
+      LoggerProxy.logger.error(`Meeting:index#startTranscription --> ${error}`);
       Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.RECEIVE_TRANSCRIPTION_FAILURE, {
         correlation_id: this.correlationId,
         reason: error.message,
@@ -3950,6 +3960,7 @@ export default class Meeting extends StatelessWebexPlugin {
       .then((join) => {
         joinSuccess(join);
         this.deferJoin = undefined;
+        this.receiveTranscription = !!options.receiveTranscription;
         Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.JOIN_SUCCESS, {
           correlation_id: this.correlationId,
         });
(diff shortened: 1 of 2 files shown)


arun3528 (approved):
Lgtm

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -2261,8 +2262,17 @@ export default class Meeting extends StatelessWebexPlugin {
         });
       }
     });
-    this.locusInfo.on(LOCUSINFO.EVENTS.SELF_ADMITTED_GUEST, (payload) => {
+    this.locusInfo.on(LOCUSINFO.EVENTS.SELF_ADMITTED_GUEST, async (payload) => {
       this.stopKeepAlive();
+      // @ts-ignore
+      if (!this.transcription && (this.config.receiveTranscription || this.receiveTranscription)) {
```
rarajes2, line 2268: Question: Why do we have two different checks `this.config.receiveTranscription` and `this.receiveTranscription`?

rarajes2 (approved):
LGTM

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: `this.config.receiveTranscription` checks if we have a `receiveTranscription` property when we initialise SDK with config object.

`this.receiveTranscription` is when user selects `startsReceivingTranscription()` in samples app before joining a meeting means if we are passing `receiveTranscription: true` in join meeting options.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1761,8 +1762,8 @@ export default class Meeting extends StatelessWebexPlugin {
       LOCUSINFO.EVENTS.CONTROLS_MEETING_TRANSCRIBE_UPDATED,
       ({caption, transcribing}) => {
         // @ts-ignore - config coming from registerPlugin
-        if (transcribing && this.transcription && this.config.receiveTranscription) {
-          this.receiveTranscription();
+        if (transcribing && !this.transcription && this.config.receiveTranscription) {
```
chrisadubois, line 1765: shouldn't a test be changed for this logic flip? i.e., `this.transcription -> !this.transcription` ?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5698,7 +5709,7 @@ export default class Meeting extends StatelessWebexPlugin {
     if (layoutType) {
       if (!LAYOUT_TYPES.includes(layoutType)) {
         return this.rejectWithErrorLog(
-          'Meeting:index#changeVideoLayout --> cannot change video layout, invalid layoutType recieved.'
+          'Meeting:index#changeVideoLayout --> cannot change video layout, invalid layoutType received.'
```
chrisadubois, line 5712: nice typo fix

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1761,8 +1762,8 @@ export default class Meeting extends StatelessWebexPlugin {
       LOCUSINFO.EVENTS.CONTROLS_MEETING_TRANSCRIBE_UPDATED,
       ({caption, transcribing}) => {
         // @ts-ignore - config coming from registerPlugin
-        if (transcribing && this.transcription && this.config.receiveTranscription) {
-          this.receiveTranscription();
+        if (transcribing && !this.transcription && this.config.receiveTranscription) {
+          this.startTranscription();
```
chrisadubois, line 1766: question: doesn't startTranscription call the service to actually set up a transcription? Doesn't the transcription already exist at this point? Is the API fully idempotent such that this is not an issue ?

chrisadubois (commented):
Have you tested this with the web client or samples app? can you please include your test scenarios from the web client? 

if you want to test with the web client app you can simply run SDK_DIR=<pathToSdk> yarn start and then start a meeting with transcription and join as guest to make sure this is functioning properly.

Or if it's working with the samples app (not sure if samples app has the transcriptions functionality in it) that would be fine too.

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: I checked but there's no test written for this if condition. For the new piece of logic which I have added as part of this PR, I have added a test case for it.

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: `startTranscription()` api basically creates a new `transcription` object and assigns it on meeting object and also subscribes to incoming transcriptions.

This method is called while [joining the meeting](https://github.com/meta-dipanshu-sharma/webex-js-sdk/blob/guest-transcription-issue-fix/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3975) also if guest wants to receive transcription but it fails because of empty `datachannelUrl` as I mentioned in the PR description. Then, once we have `datachannelUrl` when user is admitted into the meeting, we are calling `startTranscription` method again which gets completed successfully and user starts receiving transcriptions.

meta-dipanshu-sharma:
> Have you tested this with the web client or samples app? can you please include your test scenarios from the web client?
> 
> if you want to test with the web client app you can simply run SDK_DIR= yarn start and then start a meeting with transcription and join as guest to make sure this is functioning properly.
> 
> Or if it's working with the samples app (not sure if samples app has the transcriptions functionality in it) that would be fine too.

@chrisadubois I have tested it on SDK's samples app. We have transcription functionality available there. I have mentioned both the testing scenarios in the PR description at the top under `The following scenarios where tested` title which were failing previously but working now after this PR

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: But if you changed the logic here... there should be a test for this if condition (unless it's not feasible to do so for some reason?) In the tests below - I don't see anything for checking with the !this.transcription, it's all solely based off of the config.

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: ah I see. That makes sense, so it doesn't make an API call, it just starts receiving the open data channel, this makes sense.

chrisadubois (approved):
Great that this has been tested on samples app. It looks like this should work with the web client (or at least be transparent to the web client such that if a failure occurs it's entirely contained within the SDK)

I'd caution that I think the startTranscription should be probably wrapped in a try catch and I would also think there should be tests for the reversal of the if check.

Is there any plan for merging this into the beta branch? I see it's only on next branch which web client isn't even using today.

packages/@webex/plugin-meetings/src/meeting/index.ts
meta-dipanshu-sharma: Added tests for locus event `CONTROLS_MEETING_TRANSCRIBE_UPDATED` as well for both if and else condition.

meta-dipanshu-sharma:
> Great that this has been tested on samples app. It looks like this should work with the web client (or at least be transparent to the web client such that if a failure occurs it's entirely contained within the SDK)
> 
Yes, correct.
> I'd caution that I think the startTranscription should be probably wrapped in a try catch and I would also think there should be tests for the reversal of the if check.
> 
`startTranscription` method have it's own [try](https://github.com/meta-dipanshu-sharma/webex-js-sdk/blob/guest-transcription-issue-fix/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3744) [catch](https://github.com/meta-dipanshu-sharma/webex-js-sdk/blob/guest-transcription-issue-fix/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3789) block which will catch the error and will contain it inside SDK only as we're not throwing error further from catch block. Do you still want line from where we call `startTranscription` method wrapped in try catch block?
> Is there any plan for merging this into the beta branch? I see it's only on next branch which web client isn't even using today.

I'm not 100% sure what's the plan but I think we're going to merge next into beta at some point of time in future.
cc: @chrisadubois

PR #3182, opened 2023-11-02
fix: do a full Locus sync if a delta sync fails

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-472336

When we get out of order delta events we start a timer and we do a Locus delta sync when that timer expires. We are seeing a large number of cases where that Locus delta sync request fails - it's not clear why, but when this happens SDK is stuck in a state where it will keep re-trying the delta sync every 10-15s until it succeeds or until we get the missing delta events from Locus, which seems to never happen.

If delta sync fails, do a full sync - this should get us out of the state where we keep retrying the delta sync.

Also added some metrics to help us investigate further what is causing the problem of failing delta syncs in the first place.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/index.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/index.ts
@@ -129,8 +129,18 @@ export default class LocusInfo extends EventsScope {
 
           isDelta = false;
 
-          return meeting.meetingRequest.getLocusDTO({url: meeting.locusUrl});
+          return meeting.meetingRequest.getLocusDTO({url: meeting.locusUrl}).catch((err) => {
+            LoggerProxy.logger.info(
+              'Locus-info:index#doLocusSync --> fallback full sync failed, destroying the meeting'
+            );
+            this.webex.meetings.destroy(meeting, MEETING_REMOVED_REASON.LOCUS_DTO_SYNC_FAILED);
+            throw err;
+          });
         }
+        LoggerProxy.logger.info(
+          'Locus-info:index#doLocusSync --> fallback full sync failed, destroying the meeting'
+        );
+        this.webex.meetings.destroy(meeting, MEETING_REMOVED_REASON.LOCUS_DTO_SYNC_FAILED);
         throw e;
       })
       .then((res) => {
@@ -145,8 +155,6 @@ export default class LocusInfo extends EventsScope {
         } else {
           meeting.locusInfo.onFullLocus(res.body);
         }
-      })
-      .finally(() => {
         // Notify parser to resume processing delta events.
         // Any deltas in the queue that have now been superseded by this sync will simply be ignored
         this.locusParser.resume();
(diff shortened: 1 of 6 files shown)


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -110,6 +112,27 @@ export default class LocusInfo extends EventsScope {
     // return value ignored on purpose
     meeting.meetingRequest
       .getLocusDTO({url})
+      .catch((e) => {
```
chrisadubois, line 115: question: this is good - this will handle the 409 but isn't it possible that this would still get called in an infinite loop via

```
  applyLocusDeltaData(action: string, locus: any, meeting: any) {
    const {DESYNC, USE_CURRENT, USE_INCOMING, WAIT} = LocusDeltaParser.loci;

    switch (action) {
      case USE_INCOMING:
        meeting.locusInfo.onDeltaLocus(locus);
        break;
      case USE_CURRENT:
      case WAIT:
        // do nothing
        break;
      case DESYNC:
        this.doLocusSync(meeting);
        break;
      default:
        LoggerProxy.logger.info(
          `Locus-info:index#applyLocusDeltaData --> Unknown locus delta action: ${action}`
        );
    }
  }
  ```
  
  Shouldn't we *also* have a check for the meeting to be over by handling
  
  `2409113=Invalid client sequence data on state synchronization (check if meeting has ended).`
  
  i.e., if the sync request returns with 2409113 code we could trigger a destroy sequence of the meeting itself?

chrisadubois (approved):
Approved but a question that might need to be a separate PR?

packages/@webex/plugin-meetings/src/locus-info/index.ts
marcin-bazyl: my assumption was that the full locus sync will not fail, but you're right. I'll try to reproduce this scenario now that I understand better how to trigger the 409 error and confirm what happens

marcin-bazyl, line 115: anyway, this fix doesn't do anything, because inside getLocusDTO() we are returning the error instead of throwing, so when request to Locus fails, getLocusDTO resolves with the error instead of rejecting. I'll add a fix for it.

marcin-bazyl: added handling for the case if full locus also fails

marcin-bazyl: fixed

marcin-bazyl:
maybe we could do something more to proactively avoid the locus delta sync in the first place, but we'll see if and how to do it based on the metrics I'm adding in this PR

PR #3186, opened 2023-11-03
fix(diagnostics): make sure authorization token is not logged

chrisadubois:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-472217

logging auth token in call diagnostic calls

-send an object through so the `walkAndFilter` function in logger can actually redact the authorization header
-add a JSON stringifier (that handles circular references as well) to the logger
-adds tests for the JSON stringifier for objects passed through to the logger
-removes known locations where we pre-stringify the payload or response before sending it to the logger

Screenshot of new log
Screenshot 2023-11-06 at 12 32 29 PM

### Change Type

test the call diagnostics metric that comes through using devtools breakpoints to see the string - then check it in the console/buffer

### I certified that

---

](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-472217)


chrisadubois:
this works because of this function in plugin-logger

```
function walkAndFilter(object, visited = []) {
  if (visited.includes(object)) {
    // Prevent circular references
    return object;
  }

  visited.push(object);

  if (isArray(object)) {
    return object.map((o) => walkAndFilter(o, visited));
  }
  if (!isObject(object)) {
    if (isString(object)) {
      if (patterns.containsEmails.test(object)) {
        return object.replace(patterns.containsEmails, '[REDACTED]');
      }
    }

    return object;
  }

  for (const [key, value] of Object.entries(object)) {
    if (authTokenKeyPattern.test(key)) {
      Reflect.deleteProperty(object, key);
    } else {
      object[key] = walkAndFilter(value, visited);
    }
  }

  return object;
}
```

we should not be JSON.stringify objects to send to the logger as the logger itself already handles that and has a sanitizer for emails and auth tokens.

jjkusowski (approved):

Coread (changes requested):
The reason the change was made to stringify, was because in the logs, if you don't, you just get lines like
```js
,2023-11-04T13:42:23.352Z,wx-js-sdk,NewMetrics: @postPreLoginMetric. Request successful:,response:,[object Object]
```

There are also two other instances of this not covered by this PR in packages/@webex/internal-plugin-metrics/src/new-metrics.ts

arun3528 (approved):

arun3528:
@chrisadubois  it wont have any impact to developers as there is no api changes as such , correct me if I am wrong

Coread (approved):

packages/@webex/plugin-logger/test/unit/spec/logger.js
```diff
@@ -185,6 +185,52 @@ describe('plugin-logger', () => {
...
+          authorization: 'Bearer',
+          trackingid: '123',
+        },
+        test: 'object',
+        nested: {
+          test2: 'object2',
+        }
+      }
+
+      obj.selfReference = obj;
```
marcin-bazyl, line 224: maybe it would be worth also adding a test case for the circular reference being inside some array? and also for it to reference some other node in the 'obj', not just the `obj` itself

marcin-bazyl (approved):

PR #3194, opened 2023-11-13
feat: added getPermissionTokenTimeLeft function

cscheffauer:

the issue of not knowing how much seconds are left till we reach permissionToken timeout

- Added `getPermissionTokenTimeLeftInSec` public method to return

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Unit tested the function, to cover return values of:
  - `undefined`
  - positive value
  - negative value

### I certified that


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3021,6 +3022,7 @@ export default class Meeting extends StatelessWebexPlugin {
         webexMeetingInfo?.hostId ||
         this.owner;
       this.permissionToken = webexMeetingInfo?.permissionToken;
+      this.permissionTokenTtl = webexMeetingInfo?.permissionTokenTtl;
```
martroos, line 3025: suggestion: For whatever reason, the `permissionTokenTtl` field is actually supplied as a string from the API:

Screenshot 2023-11-13 at 15 15 23

Not sure if this is a mistake on the service side (feels like it is), but we may want to ensure it's a number so that we can rely on the type from this point onwards. I realise the calculation will happen to work because js is overly permissive, but it would be nice to be able to rely on it.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -5671,6 +5672,7 @@ describe('plugin-meetings', () => {
               meetingNumber: '12345',
               permissionToken:
                 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwZXJtaXNzaW9uIjp7InVzZXJQb2xpY2llcyI6eyJhIjp0cnVlfX0sImlhdCI6MTY4OTE2NDEwMn0.9uL_U7QUdYyMerrgHC_gCKOax2j_bz04u8Ikbv9KiXU',
+              permissionTokenTtl: 12345678,
```
martroos, line 5675: suggestion: This should be a string

docs/samples/webex.min.js
martroos, line 1: I don't think these should be comitted - at least I haven't seen other PRs with these.

docs/samples/webex.min.js.map
martroos, line 1: I don't think these should be comitted - at least I haven't seen other PRs with these.

martroos (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7179,4 +7181,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * (from the time the function has been fired)
+   *
+   * @returns {number} time left in seconds
+   */
+  public getPermissionTokenTimeLeftInSec(): number | undefined {
+    if (!this.permissionTokenTtl) {
+      return undefined;
+    }
+
+    // substract current time from the expiration time
```
martroos, line 7196: nitpick: Typo!

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -5657,12 +5666,25 @@ describe('plugin-meetings', () => {
...
+          const permissionTokenPayloadData = {
+            exp: '12345678',
+            permission: {
+              userPolicies: {
+                a: true
+              }
+            }
+          };
+
+          const jwtDecodeStub = sinon.stub(jwt, 'decode').returns(permissionTokenPayloadData);
```
marcin-bazyl, line 5686: hmm not sure about the changes in this test. Why do we need to stub jwt.decode now? the permissionToken in FAKE_MEETING_INFO seems to be already populated as a JWT with the data that test expects, can't we just let the decode actually run and decode it?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3022,8 +3022,8 @@ export default class Meeting extends StatelessWebexPlugin {
         webexMeetingInfo?.hostId ||
         this.owner;
       this.permissionToken = webexMeetingInfo?.permissionToken;
-      this.permissionTokenTtl = webexMeetingInfo?.permissionTokenTtl;
-      this.setSelfUserPolicies(this.permissionToken);
+      this.setPermissionTokenPayload(webexMeetingInfo?.permissionToken);
```
martroos, line 3025: nitpick: Why not just use `this.permissionToken` here?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -5563,42 +5563,51 @@ describe('plugin-meetings', () => {
         });
       });
 
+      describe('#setPermissionTokenPayload', () => {
+        it('sets correctly', () => {
+          assert.notOk(meeting.permissionTokenPayload);
+
+          const permissionTokenPayloadData = {permission: {userPolicies: {a: true}}, exp: '1234'};
+          
+          const jwtDecodeStub = sinon.stub(jwt, 'decode').returns(permissionTokenPayloadData);
```
martroos, line 5572: suggestion: This probably doesn't need stubbing - surely we could just set the permission token on the meeting to an actual token value, and let the decode run? I still think it might be worth spying on decode (to ensure it is used), but I think there is value in it actually running.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7189,12 +7198,18 @@ export default class Meeting extends StatelessWebexPlugin {
...
   public getPermissionTokenTimeLeftInSec(): number | undefined {
-    if (!this.permissionTokenTtl) {
+    if (!this.permissionTokenPayload) {
       return undefined;
     }
 
-    // substract current time from the expiration time
-    // (permissionTokenTtl is a epoch timestamp, not a time to live duration)
-    return (this.permissionTokenTtl - Date.now()) / 1000;
+    const permissionTokenExpValue = Number(this.permissionTokenPayload.exp);
```
martroos, line 7205: nitpick: This should be a number already, but it doesn't hurt to ensure it anyway.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8002,25 +8027,33 @@ describe('plugin-meetings', () => {
...
+    })
+
+    it('should return undefined if exp is undefined', () => { 
+      assert.equal(meeting.getPermissionTokenTimeLeftInSec(), undefined)
+    });
 
+    it('should return the expected positive exp', () => { 
       // set permission token as now + 1 sec
-      meeting.permissionTokenTtl = now + 1000;
+      meeting.permissionTokenPayload = {exp: (now + 1000).toString()};
```
martroos, line 8050: nitpick: This doesn't need to be made a string as it is actually a number in the permissionToken, doesn't matter too much though as this tests the conversion!

martroos (approved):

marcin-bazyl (approved):

PR #3210, opened 2023-11-21
feat: add API for renewal of the permission token

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-475304

There is no API to refresh permission token

Added API to refresh permission token.

NOTE: currently this doesn't work correctly for cases when we start an instant space meeting - the new meeting info we get after refresh is not the same as we got from /spaceInstant API call - this is being discussed with WbxAppAPI team

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1398,10 +1398,15 @@ export default class Meeting extends StatelessWebexPlugin {
   /**
    * Refreshes the meeting info permission token (it's required for joining meetings)
    *
+   * @param {string} [reason] used for metrics and logging purposes (optional)
    * @returns {Promise}
    */
-  public async refreshPermissionToken(): Promise<void> {
+  public async refreshPermissionToken(reason?: string): Promise<void> {
     if (!this.meetingInfo?.permissionToken) {
+      LoggerProxy.logger.info(
+        `Meeting:index#refreshPermissionToken --> cannot refresh the permission token, because we don't have it (reason=${reason})`
+      );
+
       return;
     }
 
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1420,6 +1395,92 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      // @ts-ignore - config coming from registerPlugin
+      this.config.experimental.enableAdhocMeetings &&
+      // @ts-ignore
+      this.webex.meetings.preferredWebexSite;
+
+    const destination = isStartingSpaceInstantV2Meeting
+      ? this.meetingInfo.meetingJoinUrl
+      : this.destination;
+    const destinationType = isStartingSpaceInstantV2Meeting ? _MEETING_LINK_ : this.destinationType;
+
```
cscheffauer, line 1419: thought: shall we add additional metrics here about that the `refresh` of the permission token has happened?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1420,6 +1395,92 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    await this.fetchMeetingInfoInternal({
```
cscheffauer, line 1420: question: the part of catching the `captcha` and `password` errors is not in the internal fetchMeetingInfo function. We want to throw these errors when we refreshPermissionToken no? (for example the password changes while a user waits a long time in interstitial). It seems the captcha and password errors are still getting thrown in the call stack below, but then i am wondering why we had these handlers in line 1457 originally?

cscheffauer (changes requested):
some questions before approving it

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: you mean the checks that we do in fetchMeetingInfo() before we call `this.attrs.meetingInfoProvider.fetchMeetingInfo()`? These are not possible to happen if we already have the permissionToken  (checked in line 1404). If we have the token, it means we must have passed correct password and captcha or they were not required. If password changes or for whatever reason wbxappapi decides we need to show a captcha when refreshing the token, the errors will be thrown by the call to `this.attrs.meetingInfoProvider.fetchMeetingInfo()` and are caught inside fetchMeetingInfoInternal() (line 1339)

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yeah, good point, I forgot about the metrics, I thought there was a separate jira for that. I'll add them now.

marcin-bazyl: done

cscheffauer (approved):

PR #3211, opened 2023-11-22
fix: downscopeing issues

maxinteger:
# COMPLETES SPARK-474816

Scopes defined by the app and the scopes from the idbroker can mismatch and cause "invalid scope" error during downscoping.

It can happen when a new release of the app define more scopes, but the user did not logout and in again to get the new scopes from idbroker. Unfortunately super token refresh does not update the scopes.

Exclude invalid scopes from the downscope request (and log a warning), except when the downscope called with only invalid scopes, in this case we let handle the error by the downscope function and fallback to the super token.

This should greatly reduce the the number of 400 error during downscope.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -258,8 +258,28 @@ const Credentials = WebexPlugin.extend({
    * @returns {Promise<Token>}
    */
   downscope(scope) {
+    const superTokenScopeArr = this.supertoken.scope.split(' ');
+    const scopeArr = scope.split(' ');
+
+    const scopeDiff = difference(scopeArr, superTokenScopeArr);
+    const newScope = without(scopeArr, ...scopeDiff).join(' ');
+
```
lismith2-cisco, line 266: I think it would be better to do this earlier, so that the scopes are stored correctly, and error cases don't keep happening. Suggest:
- in credentials refresh: check each token if it is a subset of the supertoken. If not, just revoke and remove.
- in credentials getUserToken, change the default from the config scope minus spark:kms to the supertoken scope minus spark:kms
- keep this function as it was, but with the improved log message if defaulting to supertoken (which we would no longer expect to see, so maybe make it a warning?) (we could add a metric here maybe?)
- in credentials refresh: just after refreshing the supertoken, compare its scope to the config scope, and emit a warning (and ideally a metric) if the scope is different (this could be a different PR) (this is where we would intend to force logout)

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -258,8 +258,28 @@ const Credentials = WebexPlugin.extend({
    * @returns {Promise<Token>}
    */
   downscope(scope) {
+    const superTokenScopeArr = this.supertoken.scope.split(' ');
```
lismith2-cisco, line 261: These token scope calculations look like they belong in '.scope' with the other scope utils

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -528,9 +548,7 @@ const Credentials = WebexPlugin.extend({
         this.scheduleRefresh(this.supertoken.expires);
       })
       .catch((error) => {
-        const {InvalidRequestError} = grantErrors;
-
-        if (error instanceof InvalidRequestError) {
+        if (error instanceof OAuthError) {
```
lismith2-cisco, line 551: I am kind of worried about this tbh because we have no idea if this is happening or why. Incidentally, Token refresh function catches processGrantError

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -528,9 +548,7 @@ const Credentials = WebexPlugin.extend({
...
           // Error: The refresh token provided is expired, revoked, malformed, or invalid. Hence emit an event to the client, an opportunity to logout.
```
lismith2-cisco, line 552: If you are increasing the scope of when we will be logged out, this comment is no longer fully accurate

lismith2-cisco (changes requested):

packages/@webex/webex-core/src/lib/credentials/credentials.js
maxinteger: I moved the error handling closer to the super token refresh, so we can be sure that failed and not a user token fetch.

Regarding the error check, we can not do anything except logout the user when the super token refresh fails

packages/@webex/webex-core/src/lib/credentials/credentials.js
maxinteger: @lismith2-cisco 
Is it necessary? In the `refresh` method after the super token refresh, we revoke all user tokens and fetch a new one. 

I could filter out invalid scopes here, but the downscope method will do it anyway.

packages/@webex/webex-core/src/lib/credentials/scope.js
```diff
@@ -32,3 +34,17 @@ export function filterScope(toFilter, scope) {
     .sort()
     .join(' ');
 }
+
+/**
+ * Returns the scopes not included in the other given scope
```
lismith2-cisco, line 39: It's not obvious from the function description, parameter decription or variable names, which scope is which, please make it as clear as possible :)

packages/@webex/webex-core/test/unit/spec/credentials/scope.js
```diff
@@ -0,0 +1,46 @@
...
+        assert.deepEqual(diffScopes(undefined, ''), '');
+        assert.deepEqual(diffScopes('', undefined), '');
+        assert.deepEqual(diffScopes('', ''), '');
+        assert.deepEqual(diffScopes('a', 'a'), '');
+        assert.deepEqual(diffScopes('a b c', 'a b c'), '');
+        assert.deepEqual(diffScopes(undefined, 'a b c'), '');
+        assert.deepEqual(diffScopes('a b c', 'a b c d'), '');
+      });
+
+      it('should return different when the first scope is not a subset of the second one', () => {
```
lismith2-cisco, line 37: suggestion 'should return a string containing all items in the first scope that are not in the second scope'

packages/@webex/webex-core/test/unit/spec/credentials/scope.js
```diff
@@ -0,0 +1,46 @@
...
+        assert.equal(filterScope('a', undefined), '');
+        assert.equal(filterScope('a', ''), '');
+        assert.equal(filterScope('a', 'a'), '');
+        assert.equal(filterScope('a', 'a b c'), 'b c');
+        assert.equal(filterScope('c', 'a c b'), 'a b');
+      });
+    });
+
+    describe('diffScopes', () => {
+      it('should return empty string when the first scope is subset of the second one', () => {
```
lismith2-cisco, line 26: suggestion: 'should return an empty string, if all items in the first scope are contained in the second scope'
I know your description is accurate, but it requires more mental effort to interpret I think

packages/@webex/webex-core/src/lib/credentials/scope.js
lismith2-cisco: Perhaps just change description to 'Returns a string containing all items in scopeA that are not in scopeB, or an empty string if there are none.'

packages/@webex/webex-core/test/unit/spec/credentials/scope.js
```diff
@@ -0,0 +1,46 @@
+import {assert} from '@webex/test-helper-chai';
+import {sortScope, filterScope, diffScopes} from '@webex/webex-core/src/lib/credentials/scope';
+
+describe('webex-core', () => {
+  describe('scope utils', () => {
```
lismith2-cisco, line 5: thank you for adding tests

lismith2-cisco (changes requested):

packages/@webex/webex-core/src/lib/credentials/credentials.js
lismith2-cisco: as far as I can see, if we have an invalid scope, we will fall back to supertoken, with the stored scope of the token object being the invalid scope still

packages/@webex/webex-core/src/lib/credentials/scope.js
maxinteger: I would like to keep the function general, so I picked the later suggestion and updated the docs.

packages/@webex/webex-core/src/lib/credentials/credentials.js
maxinteger: the only case when it is matter when the user update the app to the version which contains these fixes.

lismith2-cisco: Yes, but that is all existing users, and it will happen every time they refresh from now on

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -259,8 +260,12 @@ const Credentials = WebexPlugin.extend({
    */
   downscope(scope) {
     return this.supertoken.downscope(scope).catch((reason) => {
-      this.logger.trace(`credentials: failed to downscope supertoken to ${scope}`, reason);
+      const failReason = reason?.body ?? reason;
+      this.logger.warn(`credentials: failed to downscope supertoken to "${scope}"`, failReason);
       this.logger.trace(`credentials: falling back to supertoken for ${scope}`);
+      this.webex.internal.metrics.submitClientMetrics(METRICS.JS_SDK_CREDENTIALS_DOWNSCOPE_FAILED, {
+        fields: {failReason},
```
lismith2-cisco, line 267: Does the fail reason include the supertoken scope and the attemped scope? If not, could we add this to the metric?

packages/@webex/webex-core/src/lib/credentials/credentials.js
maxinteger: sorry, I thought you will see the fix.

I agree and it was fixed in the last commit.

lismith2-cisco (changes requested):

lismith2-cisco (approved):

jor-row (approved):

PR #3212, opened 2023-11-22
fix: ignore VMN reachability when deciding if we should use TURN-TLS or not

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-477129

We call isAnyClusterReachable() to decide if TURN-TLS should be attempted or not. That function checks if we reached any media cluster during reachability checks and that includes also video mesh nodes (VMNs). VMNs will most likely be reachable even when customer has a firewall blocking all udp and tcp to public nodes and when we join a call we might be placed on a public media node, so we should not take VMNs into account.

- renamed isAnyClusterReachable() to isAnyPublicClusterReachable() and changed the implementation so that it only checks the public nodes.
- we were sending the whole contents of stored reachability results to Locus when joining meetings, but now these results also include a `isVideoMesh` property that we don't need to send, so changed existing code that was reading local storage directly to instead use a new method `getReachabilityResults()` - that method reads results from local storage and strips isVideoMesh from them.
- also added some types to the existing reachability code to make it easier to understand. It's still not ideal, because when we read local storage we just get an "any". This could be improved by adding type guards, but I've decided it's not worth it for now. 
- any additions to metrics mentioned in the jira will be done in a separate PR.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, also tested manually with sample app

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -301,7 +301,7 @@ export default class Reachability {
     });
 
     return Promise.all(clusters)
-      .then(this.parseIceResultsToReachabilityResults)
+      .then(this.parseIceResultsToInternalReachabilityResults)
       .then((reachabilityLatencyResults) => {
         this.logUnreachableClusters();
 
@@ -472,7 +472,7 @@ export default class Reachability {
    * @protected
    * @memberof Reachability
    */
-  protected parseIceResultsToReachabilityResults(
+  protected parseIceResultsToInternalReachabilityResults(
     iceResults: Array<ICECandidateResult>
   ): InternalReachabilityResults {
     const reachabilityMap = {};
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -19,8 +19,8 @@ const VIDEO_MESH_TIMEOUT = 1000;
 // result for a specific transport protocol (like udp or tcp)
 export type TransportResult = {
   reachable: 'true' | 'false';
-  latencyInMilliseconds: string;
-  clientMediaIPs: string[];
+  latencyInMilliseconds?: string;
+  clientMediaIPs?: string[];
   untested?: 'true';
 };
 
@@ -139,6 +139,7 @@ export default class Reachability {
   }
 
   /**
+   * Reachability results as an object in the format that backend expects
    *
    * @returns {any} reachability results that need to be sent to the backend
    */
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -112,7 +112,7 @@ export default class Reachability {
    * @public
    * @memberof Reachability
    */
-  async isAnyClusterReachable() {
+  async isAnyPublicClusterReachable() {
     let reachable = false;
     // @ts-ignore
     const reachabilityData = await this.webex.boundedStorage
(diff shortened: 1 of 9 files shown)


packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -106,13 +138,52 @@ export default class Reachability {
...
+      results = JSON.parse(resultsJson);
+
+      // remove any internal SDK data from the results, leave only the stuff that backend needs
+      Object.values(results).forEach((result) => {
+        Object.keys(result).forEach((prop) => {
+          if (!reachabilityResultsProps.includes(prop)) {
+            delete result[prop];
+          }
+        });
+      });
```
maxinteger, line 168: please do not use `delete` operator because it can force browser to de-optimise code, also it is mutate in place.

you can use `mapValues` and `pick` from lodash, like this:
```suggestion
  results = mapValues(results, (result) => pick(result, reachabilityResultsProps))
```

maxinteger (changes requested):
looks good, I have only one note

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -16,7 +16,39 @@ import ReachabilityRequest from './request';
 const DEFAULT_TIMEOUT = 3000;
 const VIDEO_MESH_TIMEOUT = 1000;
 
-export type ICECandidateResult = {clusterId: string; elapsed?: string | null; publicIPs?: string[]};
+// result for a specific transport protocol (like udp or tcp)
+export type TransportResult = {
+  reachable: 'true' | 'false';
```
torgeadelin, line 21: any reason we're using string literals and not `boolean`?

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -121,10 +192,12 @@ export default class Reachability {
 
     if (reachabilityData) {
       try {
-        const reachabilityResults = JSON.parse(reachabilityData);
+        const reachabilityResults: InternalReachabilityResults = JSON.parse(reachabilityData);
```
torgeadelin, line 195: Why not call getReachabilityResults instead of manually grabbing them and parsing again?

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -392,10 +472,12 @@ export default class Reachability {
    * @protected
    * @memberof Reachability
    */
-  protected parseIceResultsToReachabilityResults(iceResults: Array<ICECandidateResult>) {
+  protected parseIceResultsToReachabilityResults(
```
torgeadelin, line 475:
```suggestion
  protected parseIceResultsToInternalReachabilityResults(
```

torgeadelin (commented):
just some question/suggestion

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: This is existing code, I only added the type for it, I guess it's like that because that's what the server expects. I don't want to risk changing it now. I mean I could try, but then I need to join a test meeting and confirm in Orpheus logs that it reads it correctly and takes it into account correctly when deciding call placement, so it will take time.

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: because getReachabilityResults() returns the results without the internal stuff like `isVideoMesh` field that we need to read here

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: done, thanks for the suggestion, it's much better now

packages/@webex/plugin-meetings/src/reachability/index.ts
torgeadelin: nah, it's ok i don't mind, i thought there is a reason for it

torgeadelin (approved):
LGTM

maxinteger (approved):

PR #3214, opened 2023-11-23
fix: refresh permission token on meetingjoin

cscheffauer:

refresh permission token when meetingJoin happens

- combine `getPermissionTokenTimeLeft` and `refreshPermissionToken` to refresh the token on join if the token is about to expire in the next 10 seconds
- added unit tests

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- checkAndRefreshPermissionToken method unit tests
- additional join unit tests (catching errors, etc.)

### I certified that

### DIFF
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -1250,5 +1250,6 @@ export const IP_VERSION = {
 
 export type IP_VERSION = (typeof IP_VERSION)[keyof typeof IP_VERSION];
 
-// constant for if the permissionToken is about to expire in the next 10 seconds, refresh it
-export const MEETING_PERMISSION_TOKEN_REFRESH_THRESHOLD_IN_SEC = 10;
+// constant for if the permissionToken is about to expire in the next 30 seconds, refresh it
+export const MEETING_PERMISSION_TOKEN_REFRESH_THRESHOLD_IN_SEC = 30;
+export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -1249,3 +1249,6 @@ export const IP_VERSION = {
 } as const;
 
 export type IP_VERSION = (typeof IP_VERSION)[keyof typeof IP_VERSION];
+
+// constant for if the permissionToken is about to expire in the next 10 seconds, refresh it
+export const MEETING_PERMISSION_TOKEN_REFRESH_THRESHOLD_IN_SEC = 10;
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4562,7 +4562,7 @@ export default class Meeting extends StatelessWebexPlugin {
           error instanceof PermissionError
         ) {
           // if refresh permission token requires captcha, password or permission, we are throwing the errors
-          // and bubbble it up to Cantina
+          // and bubble it up to Cantina
           throw error;
         }
       })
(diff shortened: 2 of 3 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4548,6 +4549,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      LoggerProxy.logger.error('Meeting:index#join --> Failed to refresh permission token:', error);
+
+      if (
+        error instanceof CaptchaError ||
+        error instanceof PasswordError ||
+        error instanceof PermissionError
+      ) {
+        // if refresh permission token requires captcha, password or permission, we are throwing the errors
+        // and bubble it up to Cantina
+        return Promise.reject(error);
```
marcin-bazyl, line 4568: there is a lot of code in the catch() block starting at line 4587 that normally gets executed when something fails, but if `checkAndRefreshPermissionToken()` fails with one of captcha/password/permisison errors, then we fail and we don't execute that code. I think we need that to be executed.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4548,6 +4549,26 @@ export default class Meeting extends StatelessWebexPlugin {
 
     this.isMultistream = !!options.enableMultistream;
 
+    try {
+      // refresh the permission token if its about to expire in 10sec
+      await this.checkAndRefreshPermissionToken(
+        MEETING_PERMISSION_TOKEN_REFRESH_THRESHOLD_IN_SEC,
+        'ttl-join'
```
shnaaz, line 4556: Shouldn't this be in constant fine too?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8308,4 +8371,39 @@ describe('plugin-meetings', () => {
       assert.equal(meeting.getPermissionTokenTimeLeftInSec(), -1);
     });
   });
+
+  describe('#checkAndRefreshPermissionToken', () => {
```
marcin-bazyl, line 8375: might be worth adding a case for when there is no permission token

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1249,3 +1249,6 @@ export const IP_VERSION = {
 } as const;
 
 export type IP_VERSION = (typeof IP_VERSION)[keyof typeof IP_VERSION];
+
+// constant for if the permissionToken is about to expire in the next 10 seconds, refresh it
+export const MEETING_PERMISSION_TOKEN_REFRESH_THRESHOLD_IN_SEC = 10;
```
marcin-bazyl, line 1254: I know UCF has it as 10s but maybe we should start with a bit bigger value, like 30s?

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7401,4 +7422,22 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   *
+   * @param {number} threshold - time in seconds
+   * @param {string} reason - reason for refreshing the permission token
+   * @returns {Promise<void>}
+   */
+  public checkAndRefreshPermissionToken(threshold: number, reason: string): Promise<void> {
+    const permissionTokenTimeLeft = this.getPermissionTokenTimeLeftInSec();
+
+    if (permissionTokenTimeLeft <= threshold) {
+      return this.refreshPermissionToken(reason);
```
shnaaz, line 7438: Question: looked at refreshPermissionToken, it can return without throwing error, so how do we deal that scenario?

cscheffauer: we just continue the join flow - refreshing the permission token is a "safety" mechanism - but if for example the permission token is fresh and not about to expire, the function will just return and the join flow just continues as normal.

marcin-bazyl (approved):

shnaaz (approved):

PR #3233, opened 2023-12-06
fix: added logging for getCurrentConnectionType() error cases

marcin-bazyl:
More logs so that we can understand better why we get "unknown" connection type sometimes.

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/properties.ts
+++ b/packages/@webex/plugin-meetings/src/media/properties.ts
@@ -259,9 +259,24 @@ export default class MediaProperties {
 
         return true;
       }
+      LoggerProxy.logger.warn(
+        `Media:properties#getCurrentConnectionType --> missing localCandidate.protocol, candidateType=${localCandidate.candidateType}`
+      );
 
       return false;
     });
+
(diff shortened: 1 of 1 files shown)


packages/@webex/plugin-meetings/src/media/properties.ts
```diff
@@ -259,10 +259,25 @@ export default class MediaProperties {
...
         return true;
       }
+      LoggerProxy.logger.warn(
+        `Media:properties#getCurrentConnectionType --> missing localCandidate.protocol, candidateType=${localCandidate.candidateType}`
+      );
 
       return false;
     });
 
+    if (foundConnectionType === 'unknown') {
```
Coread, line 269: question: should we not have tests for this?

maxinteger (approved):

packages/@webex/plugin-meetings/src/media/properties.ts
marcin-bazyl: I've tested it manually e2e - confirmed the log is displayed correctly, I don't care about potential regressions here, because we're planning to remove this whole function soon.

jor-row (approved):

PR #3235, opened 2023-12-06
fix(metrics): catalog-delay

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to extend the duration in which we wait for a catalog entry for the `metrics` service from `5` to `15`.

## Notes

After investigation, it appears that a majority of the initialization metrics payloads are not being sent to the metrics service. It appears that this is due to machines that are failing to receive the `metrics` service urls within the first 5 seconds of the applications initialization.

To accommodate for this, we will start by extending this value from `5` seconds to `15` seconds to allow for slower machines and/or slower connections to establish the `metrics` catalog entry instead of timing out and failing to send the metric.

This specifically helps alleviate `missing.webexapp.launched` errors, but will assist in margin-fixing for all early-sent metrics payloads.

## Links

* [SPARK-481044](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-481044)


packages/@webex/webex-core/test/unit/spec/services/interceptors/service.js
```diff
@@ -5,13 +5,14 @@ import chai from 'chai';
...
 import {ServiceInterceptor} from '@webex/webex-core';
+import CONFIG from '../../../../../src/config';
 
 const {assert} = chai;
 
 chai.use(chaiAsPromised);
 sinon.assert.expose(chai.assert, {prefix: ''});
 
-describe('webex-core', () => {
+describe.only('webex-core', () => {
```
arun3528, line 15: remove only

chrisadubois (approved):
Please add a fallback `waitForServiceTimeout` in case the value is set to negatives or something else. i.e., don't let the waitForServiceTimeout be set below 5?

this is a thought, not a blocker.

Please also remove the .only

arun3528 (approved):

PR #3239, opened 2023-12-08
fix(breakouts): user gets removed from meeting after end of session

chburket:
# COMPLETES
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-468666

Sometimes when host ends breakout sessions, a user that was in a bo session will get removed from the meeting instead of just moving back to the main session.  This happens because the SDK is trying to compare a main-session locus DTO to a breakout-session DTO (different locus urls) and decides that the deltas are out of order. The SDK then waits to see if the "missing" DTO comes in, but after a timeout will then try to do a full sync. The problem is that it's trying to do a full sync on the BO session instead of the main session, even though the user is not in the BO session any longer.

This results in a locus "Join the meeting before retrieving the roster" error and we force the user out of the meeting.

I adjusted the locus-info parser to recognize that the urls are different when comparing deltas and result in a LOCUS_URL_CHANGED instead of WAIT in that case.  I also added a sync trigger when the urls change from one delta to the next.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added unit tests and manually tested.

### I certified that

---


packages/@webex/plugin-meetings/src/locus-info/parser.ts
```diff
@@ -716,10 +723,16 @@ export default class Parser {
...
       this.pause();
     } else if (lociComparison === USE_INCOMING) {
+      const locusUrlChanged = newLoci?.url !== this.workingCopy?.url;
+
       // update working copy for future comparisons.
       // Note: The working copy of parser gets updated in .onFullLocus()
       // and here when USE_INCOMING locus.
       this.workingCopy = newLoci;
+
+      if (this.workingCopy?.url && locusUrlChanged) {
```
marcin-bazyl, line 733: I think this will trigger a sync every time on the first delta event that we get, because `this.workingCopy` will be `null` when it's read in line 726 and that's not what we want, right?

packages/@webex/plugin-meetings/src/locus-info/parser.ts
```diff
@@ -288,7 +289,13 @@ export default class Parser {
         break;
 
       case LT:
-        if (extract(Parser.compareSequence(incoming.baseSequence, incoming.sequence)) === EQ) {
+        if (incoming.url !== current.url) {
```
marcin-bazyl, line 292: this will work, but conceptually it feels like this check should be done before the `switch()` statement as it really doesn't make any sense to be doing any logic based on the comparison if the delta event is for different url, so I think it would be better to move this up before line 285

packages/@webex/plugin-meetings/test/unit/spec/locus-info/lib/SeqCmp.json
```diff
@@ -517,6 +517,22 @@
       "new": "seq4g",
       "result": "DESYNC",
       "description": "New seq is greater than current, but both current and new base have some unique entries."
+    },
+    "updt25": {
+      "current": "seq4",
+      "newbase": "seq5c",
```
marcin-bazyl, line 523: minor thing, but shouldn't `new` be greater than `newbase`?

packages/@webex/plugin-meetings/test/unit/spec/locus-info/parser.js
```diff
@@ -133,15 +140,21 @@ describe('locus-info/parser', () => {
...
+    const CURRENT_LOCI = {sequence: {entries: [100]}, url: 'CURRENT LOCI', baseSequence: {
+      "entries": [100], "rangeStart": 100, "rangeEnd": 100}};
+    const NEW_LOCI = {sequence: {entries: [200]}, url: 'NEW LOCI', baseSequence: {
+      "entries": [200], "rangeStart": 200, "rangeEnd": 200}};
+
     let sandbox = null;
     let parser;
 
     beforeEach(() => {
+      console.log('>>> beforeEach')
```
marcin-bazyl, line 152: console.log

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/test/unit/spec/locus-info/lib/SeqCmp.json
chburket: It *is* later.  Don't look at me, I didn't name the sequences. :-)

packages/@webex/plugin-meetings/src/locus-info/parser.ts
marcin-bazyl: but it's set to `newLoci` just 2 lines above, so it will always be truthy, won't it?

packages/@webex/plugin-meetings/src/locus-info/parser.ts
```diff
@@ -716,10 +723,16 @@ export default class Parser {
...
     } else if (lociComparison === USE_INCOMING) {
+      const locusUrlChanged = this.workingCopy?.url && newLoci?.url !== this.workingCopy.url;
+
       // update working copy for future comparisons.
       // Note: The working copy of parser gets updated in .onFullLocus()
       // and here when USE_INCOMING locus.
       this.workingCopy = newLoci;
+
+      if (locusUrlChanged) {
+        this.triggerSync('locus url changed, likely due to breakout session move');
```
marcin-bazyl, line 734: looking at this a bit more I'm not sure if this is correct... we call `triggerSync()` here, but we will still call `this.onDeltaAction()` in line 768, so we will start processing the incoming delta and applying it to our Locus DTO copy **before** the sync is completed, so it will mess things up. 

This makes me think that maybe we should also take a different approach and not use USE_INCOMING but have a new value added to `LocusDeltaParser.loci`, something like `LocusDeltaParser.loci.LOCUS_URL_CHANGED` and return that in `compareDelta()` so that it's more clear what's happening and then in `processDeltaEvent()` we would handle the new value and just update the `this.workingCopy` and in `applyLocusDeltaData()` for the new value just do the same we do for `DESYNC` there, because by that point the workingCopy.url would be already updated to the right one. What do you think?

chburket: I like the suggestion.  Just pushed the rework.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -1884,6 +1884,27 @@ describe('plugin-meetings', () => {
...
+        const {LOCUS_URL_CHANGED} = LocusDeltaParser.loci;
+        const fakeDeltaLocus = {id: 'fake delta locus'};
+        const meeting = {
+          meetingRequest: {
+            getLocusDTO: sandbox.stub().resolves({body: fakeDeltaLocus}),
+          },
+          locusInfo: {
+            handleLocusDelta: sandbox.stub(),
+          },
+          locusUrl: 'oldLocusUrl',
```
marcin-bazyl, line 1897: nitpick: the value "oldLocusUrl" is slightly confusing, because for the LOCUS_URL_CHANGED case we actually want to do the sync with a new locus url not the old one, but the new url is set somewhere else - in processDeltaEvent() so it's not part of this unit test. So the code is fine, but maybe a value "current locus url" or something like that would be better?

marcin-bazyl (approved):

InteractiveTimmy (approved):
LGTM!

PR #3271, opened 2023-12-19
fix: added support for Roap optimized flow for offer/answer/ok

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484740

Existing Roap offer/answer/ok flow is slow and prone to errors, because it consists of 2 http requests and 1 websocket message:
1. http request to Locus with OFFER
2. websocket message from Mercury with ANSWER
3. http request to Locus with OK

Using the "optimized" flow, where we indicate in Roap headers ("includeAnswerInHttpResponse" and "noOkInTransaction") that we want to receive the ANSWER in the http response and we don't want to send any OK message. Server then confirms that they don't expect an OK from us by adding the "noOkInTransaction" header to their ANSWER. This way the whole offer/answer/ok exchange can be done in 1 http request.

related internal-media-core PR: https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/204

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5138,7 +5140,74 @@ export default class Meeting extends StatelessWebexPlugin {
     }
   };
 
+  /**
+   * Handles an incoming Roap message
+   * @internal
+   * @param {RoapMessage} roapMessage roap message
+   * @returns {undefined}
+   */
+  public roapMessageReceived = (roapMessage: RoapMessage) => {
```
marcin-bazyl, line 5149: this function is just code moved from meetings util

torgeadelin (approved):
looks good

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -209,7 +212,41 @@ export default class Roap extends StatelessWebexPlugin {
             meeting.updateMediaConnections(mediaConnections);
           }
 
-          return locus;
+          let roapAnswer;
+
+          if (mediaConnections?.[0]?.remoteSdp) {
```
chrisadubois, line 217: Should we do something when mediaConnections[0]?.remoteSDP is undefined?

I think this might be handled elsewhere, so a comment may be nice to have here for the cases when this would be undefined.

chrisadubois: I think it would indicate a malformed response from the server in this case at this stage, right?

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: Is there any reason why we can't now move this into the roap folder and remove roap references (except for the one roap object) from meeting/index?

chrisadubois (approved):
This is nice, I like the new abstraction for it. One place would be nice to improve for the future would be for the SDK (in meeting/index) to not have any calls to ROAP at all, such as `roapMessageReceived` but instead perhaps handle this outside and completely remove roap references.

Other than that, just one small comment/question re: mediaConnections[0]

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: I'm not convinced if that belongs better in the roap folder. The code in SDK's roap folder now is only dealing with preparing the http requests for Locus when sending roap messages and handling the responses, it doesn't even reference mediaProperties or webrtcMediaConnection anywhere. Meeting object is the one that knows about the media connection it has and the roap messages need to be forwarded to it.
There is also other code related to roap in meeting/index - see inside setupMediaConnectionListeners(). Maybe we could look at moving all of this into roap folder... not sure...

packages/@webex/plugin-meetings/src/roap/index.ts
marcin-bazyl: I think mediaConnections[0]?.remoteSDP is always there, I've added the code to handle it missing just in case, I don't think we should ever get a 2xx response from Locus and that not be there.

PR #3275, opened 2023-12-20
fix(calling): exporting interfaces, types from index.ts file

Kesari3008:
# COMPLETES #[SPARK-479560](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-479560) 

Exporting interfaces, types from Calling SDK so that in web client imports can be done directly from @webex/calling instead of accessing @webex/calling/dist/types and having different paths for each type.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/calling/src/index.ts
+++ b/packages/calling/src/index.ts
@@ -1,4 +1,26 @@
 import {NoiseReductionEffect, createMicrophoneStream} from '@webex/media-helpers';
+import {ERROR_LAYER} from './Errors/types';
+import {ICallingClient} from './CallingClient/types';
+import {ICallHistory, JanusResponseEvent} from './CallHistory/types';
+import {
+  CallForwardSetting,
+  CallSettingResponse,
+  ICallSettings,
+  VoicemailSetting,
+} from './CallSettings/types';
+import {Contact, ContactResponse, GroupType, IContacts} from './Contacts/types';
+import {IVoicemail, SummaryInfo, VoicemailResponseEvent} from './Voicemail/types';
+import {ILine, LINE_EVENTS} from './CallingClient/line/types';
+import {
+  CALLING_CLIENT_EVENT_KEYS,
+  CALL_EVENT_KEYS,
+  CallerIdDisplay,
+  Disposition,
+  LINE_EVENT_KEYS,
+} from './Events/types';
+import {CallDetails, CallDirection, CallType, DisplayInformation, SORT} from './common/types';
+import {CallError} from './Errors';
+import {ICall, TransferType} from './CallingClient/calling/types';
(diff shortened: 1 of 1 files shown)


packages/calling/src/index.ts
```diff
@@ -1,4 +1,26 @@
 import {NoiseReductionEffect, createMicrophoneStream} from '@webex/media-helpers';
+import {ERROR_LAYER, ERROR_TYPE} from './Errors/types';
```
sreenara, line 2: Are imports necessary? Why not directly export?

sreenara (changes requested):
Please see the comment. Also please add the tests in the description. In this case, it would be the change in imports on the web client.

packages/calling/src/index.ts
Kesari3008: In that case, from their location itself I will have to directly export

Shreyas281299: Wont that make the import statement really long? For example:
Import {ILine, LineError} from "@webex/calling/CallingClient/line/types".

Kesari3008: updated changes

Kesari3008: @ShreyasSharma28 Even if we export directly, we should be able to import using @webex/calling

Shreyas281299: Understood.

Shreyas281299 (approved):
The changes look good. Please add the tested scenarios.

sreenara (changes requested):
[https://github.com/webex/webex-js-sdk/blob/00466bbc4fcff3a667a51a1da98a5abe613f6573/packages/calling/src/Events/types.ts#L222](This line) can be removed.

Does it also make sense to export Call History types like UserSession and COMMON_EVENT_KEYS which can be used by developers like Salesforce integrations?

Have we checked if we've covered all types needed across all modules in @webex/calling?

Kesari3008:
> [https://github.com/webex/webex-js-sdk/blob/00466bbc4fcff3a667a51a1da98a5abe613f6573/packages/calling/src/Events/types.ts#L222](This line) can be removed.
> 
> Does it also make sense to export Call History types like UserSession and COMMON_EVENT_KEYS which can be used by developers like Salesforce integrations?
> 
> Have we checked if we've covered all types needed across all modules in @webex/calling?

@sreenara I have added UserSession and COMMON_EVENT_KEYS as well. I have checked voicemail, contact, callingclient and call setting to see if more exports might be required. So it looks it's covered

sreenara (approved):

PR #3282, opened 2023-12-22
feat(fraud): add localIP to join request

chrisadubois:
# COMPLETES #[< INSERT LINK TO ISSUE >](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-485996)

Fraud

add localIp field to the join request to locus for fraud monitoring

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- unit tests

### I certified that

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/request.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/request.ts
@@ -4,6 +4,7 @@ import {debounce} from 'lodash';
 import {StatelessWebexPlugin} from '@webex/webex-core';
 // @ts-ignore
 import {deviceType} from '@webex/common';
+import {anonymizeIPAddress} from '@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util';
 
 import LoggerProxy from '../common/logs/logger-proxy';
 import {
@@ -162,6 +163,8 @@ export default class MeetingRequest extends StatelessWebexPlugin {
         url: deviceUrl,
         // @ts-ignore - config comes from registerPlugin
         deviceType: this.config.meetings.deviceType,
+        // TODO
+        // machineId: this.webex.internal.device.getMachineId(),
       },
       usingResource: resourceId || null,
       moveMediaToResource: (resourceId && moveToResource) || false,
@@ -198,6 +201,9 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       body.device.countryCode = this.webex.meetings.geoHintInfo.countryCode;
       // @ts-ignore
       body.device.regionCode = this.webex.meetings.geoHintInfo.regionCode;
+      body.device.localIp =
+        // @ts-ignore
+        anonymizeIPAddress(this.webex.meetings.geoHintInfo?.clientAddress) || undefined;
     }
 
     if (moderator !== undefined) {
--- a/packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
+++ b/packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
@@ -11,9 +11,10 @@ import {IP_VERSION} from '@webex/plugin-meetings/src/constants';
 describe('plugin-meetings', () => {
   let meetingsRequest;
   let locusDeltaRequestSpy;
+  let webex;
 
   beforeEach(() => {
-    const webex = new MockWebex({
+    webex = new MockWebex({
       children: {
         meetings: Meetings,
       },
@@ -22,6 +23,7 @@ describe('plugin-meetings', () => {
     webex.meetings.geoHintInfo = {
       countryCode: 'US',
       regionCode: 'WEST-COAST',
+      clientAddress: '127.0.0.1',
     };
 
     webex.internal = {
@@ -195,7 +197,42 @@ describe('plugin-meetings', () => {
         assert.equal(requestParams.body.device.countryCode, 'US');
         assert.equal(requestParams.body.permissionToken, 'permission-token');
         assert.equal(requestParams.body.device.regionCode, 'WEST-COAST');
+        assert.include(requestParams.body.device.localIp, '127.0.0');
       });
+
+      describe('clientAddress geoHintInfo undefined', () => {
+        beforeEach(() => {
+          webex.meetings.geoHintInfo = {};
+        });
+    
+        // reset
+        afterEach(() => {
+          webex.meetings.geoHintInfo = {
+            countryCode: 'US',
+            regionCode: 'WEST-COAST',
+            clientAddress: '127.0.0.1',
+          };
+        });
+
+        it('doesnt send the clientAddress if not available as localIp', async () => {
+          const locusUrl = 'locusURL';
+          const deviceUrl = 'deviceUrl';
+          const correlationId = 'random-uuid';
+          const roapMessage = 'roap-message';
+          const permissionToken = 'permission-token';
+  
+          await meetingsRequest.joinMeeting({
+            locusUrl,
+            deviceUrl,
+            correlationId,
+            roapMessage,
+            permissionToken,
+          });
+          const requestParams = meetingsRequest.request.getCall(0).args[0];
+  
+          assert.equal(requestParams.body.device.localIp, undefined);
+        });
+      })
 
       it('sends /call with meetingNumber if inviteeAddress does not exist', async () => {
         const deviceUrl = 'deviceUrl';


packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -162,6 +163,8 @@ export default class MeetingRequest extends StatelessWebexPlugin {
         url: deviceUrl,
         // @ts-ignore - config comes from registerPlugin
         deviceType: this.config.meetings.deviceType,
+        // TODO
+        // machineId: this.webex.internal.device.getMachineId(),
```
torgeadelin, line 167: todo?

torgeadelin (approved):

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
```diff
@@ -195,7 +197,42 @@ describe('plugin-meetings', () => {
         assert.equal(requestParams.body.device.countryCode, 'US');
         assert.equal(requestParams.body.permissionToken, 'permission-token');
         assert.equal(requestParams.body.device.regionCode, 'WEST-COAST');
-      });
+        assert.include(requestParams.body.device.localIp, '127.0.0');
```
martroos, line 200: question: Why is this an `include` instead of `equal`? The only way we know that the IP address is being anonymized is with this check, this presumably should check for `'127.0.0.0'`? Really I'd like to spy on the anonymize function to ensure it is being used correctly, but this simple check would be better than nothing.

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
```diff
@@ -195,7 +197,42 @@ describe('plugin-meetings', () => {
...
+        beforeEach(() => {
+          webex.meetings.geoHintInfo = {};
+        });
+    
+        // reset
+        afterEach(() => {
+          webex.meetings.geoHintInfo = {
+            countryCode: 'US',
+            regionCode: 'WEST-COAST',
+            clientAddress: '127.0.0.1',
```
martroos, line 213: suggestion: To avoid issues in future when adding more items to the geo hint (if there are others) I suggest changing uses of it to use a constant:

```js
const geoHintInfoDefaults = {countryCode: 'US', regionCode: 'WEST-COAST', clientAddress: '127.0.0.1'};

...

// Wherever it needs setting
webex.meetings.geoHintInfo = {...geoHintInfoDefaults};
```

Makes this a little more DRY and easier to maintain (read: harder to miss) later.

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -198,6 +201,9 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       body.device.countryCode = this.webex.meetings.geoHintInfo.countryCode;
       // @ts-ignore
       body.device.regionCode = this.webex.meetings.geoHintInfo.regionCode;
+      body.device.localIp =
+        // @ts-ignore
+        anonymizeIPAddress(this.webex.meetings.geoHintInfo?.clientAddress) || undefined;
```
martroos, line 206: nitpick: Why the optional chaining operator? If `getHintInfo` was undefined at this point, it shouldn't be in this block due to the guard statement, and even if that wasn't there, the previous lines would have error-ed before this line executed.

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -4,6 +4,7 @@ import {debounce} from 'lodash';
 import {StatelessWebexPlugin} from '@webex/webex-core';
 // @ts-ignore
 import {deviceType} from '@webex/common';
+import {anonymizeIPAddress} from '@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util';
```
martroos, line 7: issue: This fails when building for me (and at a glance, for CI also)

martroos (changes requested):
This doesn't build, not sure how this could have been tested outside of unit tests without building it, but it can't be merged in this state.

packages/@webex/plugin-meetings/src/meeting/request.ts
chrisadubois: yeah - this one will need a discussion because it will require fingerprinting, this is to mark the API it should follow.

packages/@webex/plugin-meetings/src/meeting/request.ts
chrisadubois: yes. this was opened as a draft. Once I discussed it with folks (catalin and colin) I changed it from draft to real PR.

The build issue does need to be resolved and CI did catch it which is great.

packages/@webex/plugin-meetings/src/meeting/request.ts
chrisadubois: yeah I thought about that too, im happy to remove this.

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
chrisadubois: Well if it's getting anonymized, the last 4 digits would always change. I think what would have to change here is that I will have to stub the anonymize function and have it always return the same result, and then use an equal. That's fine I can do that and it's definitely a "better" test

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
chrisadubois: yeah thats cool. easy change. Will do.

chrisadubois:
@martroos changes have been applied (builds)

screenshots of local tests
Screenshot 2024-01-08 at 10 31 22 AM
Screenshot 2024-01-08 at 10 31 16 AM

packages/@webex/plugin-meetings/src/meeting/request.ts
chrisadubois: removed

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
chrisadubois: done

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
chrisadubois: I have changed this, please have a look and let me know if you're not satisfied with it. I didn't feel like it was the best idea to retest the anonymize function, felt the test should be done in the call-diagnostic metrics util source itself, not here as well, so i felt like includes was good enough. In any case, i've changed it now using anonymize as a stub.

packages/@webex/plugin-meetings/test/unit/spec/meeting/request.js
```diff
@@ -53,6 +53,11 @@ describe('plugin-meetings', () => {
 
     meetingsRequest.request = request;
     locusDeltaRequestSpy = sinon.spy(meetingsRequest, 'locusDeltaRequest');
+    sinon.stub(Utils, 'anonymizeIPAddress').returns('0.0.0.0');
```
martroos, line 56: issue: This isn't actually doing anything - if it were I would expect tests to be failing due to this always returning `'0.0.0.0'`.

Why is this being `stub`-ed anyway, why not just spy on it, and check that the spy is being used?

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -162,6 +163,8 @@ export default class MeetingRequest extends StatelessWebexPlugin {
         url: deviceUrl,
         // @ts-ignore - config comes from registerPlugin
         deviceType: this.config.meetings.deviceType,
+        // TODO
```
martroos, line 166: nitpick: I'd prefer there to be an explanation of why there is a todo here if it really does need to stay - I'm not convinced it needs to be in the source right now, but it is not obvious as to why it is without looking at this PR.

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -198,6 +201,10 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       body.device.countryCode = this.webex.meetings.geoHintInfo.countryCode;
       // @ts-ignore
       body.device.regionCode = this.webex.meetings.geoHintInfo.regionCode;
+      body.device.localIp =
```
martroos, line 204: question: I can see this being sent to locus, but where does this appear after this for us to use?

chrisadubois: Sorry not sure what you mean - why would we need this to appear after this for us to use? we need only send it to locus so that they can capture fraudulent activities.

packages/@webex/plugin-meetings/src/meeting/request.ts
chrisadubois: Sure - i can link to the jira or I can just remove the TODO, that is fine.

packages/@webex/plugin-meetings/src/meeting/request.ts
martroos: Ah I see, it's for locus use - I thought it might have been for CA / metrics - thanks for clearing that up.

martroos (approved):

PR #3292, opened 2024-01-05
fix(plugin-meetings): don't attempt reconnect if media connection was never established

jor-row:
# COMPLETES [SPARK-482574](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-482574)

On firefox when attempting to connect media for the first time we may get ICE "Failed" state which triggers a reconnection, but reconnection will always fail if we never successfully established a media connection

check to see if we have connected previously before attempting reconnection

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5453,19 +5467,20 @@ export default class Meeting extends StatelessWebexPlugin {
           this.setNetworkStatus(NETWORK_STATUS.CONNECTED);
           this.reconnectionManager.iceReconnected();
           this.statsAnalyzer.startAnalyzer(this.mediaProperties.webrtcMediaConnection);
+          this.hasMediaConnectionConnectedAtLeastOnce = true;
```
marcin-bazyl, line 5470: when you rebase to latest beta, you'll see that I had to do something similar for setting the right icePhase when we send 'client.ice.end' when we get `ConnectionState.Connected`. I've managed to avoid adding a new flag by using the `Meeting.networkStatus`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5453,19 +5467,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
         case ConnectionState.Disconnected:
           this.setNetworkStatus(NETWORK_STATUS.DISCONNECTED);
-          this.reconnectionManager.waitForIceReconnect().catch(() => {
+          this.reconnectionManager.waitForIceReconnect().catch(async () => {
             LoggerProxy.logger.info(
               'Meeting:index#setupMediaConnectionListeners --> state DISCONNECTED, automatic reconnection timed out.'
             );
 
-            connectionFailed();
+            await connectionFailed();
```
marcin-bazyl, line 5479: what's the point of adding these awaits? We're inside an event handler and the code that emits the events will not wait for the event handler's returned promise - the event handler's return value is ignored, so having these awaits makes the person reading the code think that something will wait for these, but that's not true, so I think it's better not to have them

packages/@webex/plugin-meetings/src/meeting/index.ts
jor-row: Nice I hadn't thought of using that. I've made this change

packages/@webex/plugin-meetings/src/meeting/index.ts
jor-row: Removed

maxinteger (approved):

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -5799,6 +5791,162 @@ describe('plugin-meetings', () => {
...
+              },
+            );
+          };
+
+          it('handles "Failed" state correctly when hasMediaConnectionConnectedAtLeastOnce = false', async () => {
+            meeting.waitForMediaConnectionConnected = sinon.stub().resolves();
+
+            mockFailedEvent();
+
+            checkBehavioralMetricSent();
```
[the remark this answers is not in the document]
jor-row: Done

PR #3307, opened 2024-01-12
fix: added method for getting the meeting with active webrtc connection

marcin-bazyl:
Added a helper method so that everyday debugging is a bit easier....

when using js console and trying to read something from the active meeting we normally have to do this:
`webex.meetings.meetingCollection.meetings[<some_uid>]`
and we have to know which uid to use if there are multiple meetings in the SDK

now we can do this instead:
`webex.meetings.getActiveWebrtcMeeting()`

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/collection.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/collection.ts
@@ -60,4 +60,17 @@ export default class MeetingCollection extends Collection {
 
     return null;
   }
+
+  /**
+   * Gets the meeting that has a webrtc media connection
+   * NOTE: this function assumes there is no more than 1 such meeting
+   *
+   * @returns {Meeting} first meeting found, else undefined
+   * @public
+   * @memberof MeetingCollection
+   */
+  public getActiveWebrtcMeeting() {
+    // @ts-ignore
+    return find(this.meetings, (meeting) => meeting.mediaProperties.webrtcMediaConnection);
+  }
 }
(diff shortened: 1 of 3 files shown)


maxinteger (approved):

packages/@webex/plugin-meetings/src/meetings/collection.ts
```diff
@@ -60,4 +60,17 @@ export default class MeetingCollection extends Collection {
...
+   * Gets the meeting that has a webrtc media connection
+   * NOTE: this function assumes there is no more than 1 such meeting
+   *
+   * @returns {Meeting} first meeting found, else undefined
+   * @public
+   * @memberof MeetingCollection
+   */
+  public getActiveWebrtcMeeting() {
+    // @ts-ignore
+    return find(this.meetings, (meeting) => meeting.mediaProperties.webrtcMediaConnection);
```
maxinteger, line 74: question: why do not you use `filter` it would give you all the connections

jor-row (approved):

packages/@webex/plugin-meetings/src/meetings/collection.ts
marcin-bazyl: right now I think we only support joining 1 meeting at a time (for sure in the web app, not sure about SDK, probably as well, at least the SDK sample app doesn't support more). If I used a filter, it would return an array and every time I would have to add `[0]` at the end in my console command. The API at the meeting level is private (and MeetingCollection is private) so we can easily change it in the future if we ever support joining more meetings in parallel

PR #3314, opened 2024-01-18
fix(plugin-meetings): fix reconnection logs

jor-row:
# COMPLETES [SPARK-484339](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484339)

logging during reconnection was misleading, implying reconnection of media was completed before ice connection was successful or dtls handshake done

waiting until these steps are completed before sending success metrics and logs

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
```diff
@@ -485,14 +467,14 @@ export default class ReconnectionManager {
       const media = await this.reconnectMedia();
 
       LoggerProxy.logger.log(
-        'ReconnectionManager:index#executeReconnection --> Media reestablished'
+        'ReconnectionManager:index#executeReconnection --> webRTC media connection renewed and local sdp offer sent'
       );
       this.status = RECONNECTION.STATE.COMPLETE;
```
jor-row, line 472: This status change isn't strictly accurate either. I'm wondering if we should add a getter/setter for the status in this class and change it to complete in meeting/index.ts where we are sending ```client.media.recovered```

jor-row: I've made this change as I think it makes sense that the status of the reconnection manager is only set to complete once we have actually reestablished media (got answer and done dtls handshake) and not just all the code in reconnectionManager has ran

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4275,6 +4276,8 @@ export default class Meeting extends StatelessWebexPlugin {
 
     return this.reconnectionManager
       .reconnect(options)
+      .then(() => this.waitForRemoteSDPAnswer())
```
marcin-bazyl, line 4279: nitpick: waitForRemoteSDPAnswer() relies on `deferSDPAnswer` being created when we get LOCAL_SDP_OFFER_GENERATED event. In theory it is possible that this event is not sent and then we would be using the old value of `deferSDPAnswer` that was resolved during initial join, so waitForRemoteSDPAnswer() would resolve even though no offer or answer is sent/received. I guess in such theoretical case the call to `meeting.mediaProperties.webrtcMediaConnection.reconnect()` in reconnectMedia() should have also failed, so maybe we can just ignore this.

packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
```diff
@@ -337,73 +350,55 @@ export default class ReconnectionManager {
...
+              fatal: true,
+              name: 'media-engine',
+              shownToUser: false,
+            },
+          ],
+        },
+        options: {
+          meetingId: this.meeting.id,
+        },
+      });
```
jor-row, line 391: Thinking this should be moved to meeting/index.ts in the reconnect function

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4285,6 +4288,18 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        this.webex.internal.newMetrics.submitClientEvent({
+          name: 'client.media.recovered',
+          payload: {
+            recoveredBy: 'new',
+          },
+          options: {
+            meetingId: this.id,
+          },
+        });
+        this.reconnectionManager.setStatus(RECONNECTION.STATE.COMPLETE);
```
marcin-bazyl, line 4302: It doesn't feel right that we have to do this from here, it means that basically the ReconnectionManager is not really fully managing the reconnection: the private details of it (status) are leaked outside and the logic of what we have to do as part of reconnection (wait for answer, wait for "connected" state) is partially here in Meeting and not all inside reconnection manager. Also the CA event for starting reconnection (client.media.reconnecting) is sent from ReconnectionManager, while the final event (client.media.recovered) is sent from Meeting. It feels like to fix this we would need a slightly bigger refactor, so maybe this is fine for now and we'll do it as part of the call-end CA events work in Q3

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1036,7 +1036,9 @@ export const RECONNECTION = {
     DEFAULT_TRY_COUNT: 0,
     DEFAULT_STATUS: '',
   },
-};
+} as const;
+
+export type RECONNECTION_STATUS = Enum<typeof RECONNECTION.STATE>;
```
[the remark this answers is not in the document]
jor-row: I've changed it to RECONNECTION_STATE

packages/@webex/plugin-meetings/src/meeting/index.ts
jor-row: We've discussed this offline but for anyone reading this PR we have agreed that although it is not ideal to have reconnection logic split between classes like this it would require a greater effort to refactor to allow the reconnection to be completely contained within ReconnectionManager. 

I've created https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-492161 under the Q3 call-end epic so that we don't forget that this is something we would like to do at some point, but we will prioritise that work when we start work on call-end stuff.

maxinteger (approved):

PR #3324, opened 2024-01-24
fix(plugin-meetings): add metric for addMedia() retries 

marcin-bazyl:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-490019

We still see Locus error "User is not authorized to operate on Participant" in some cases when we retry. We have code that does rejoin, but it's not happening, because of the meeting state at that point. This metric will help us understand better what meeting state we're in.

Added metric for retry, kept it generic with a specific reason field so that it can be expanded in the future if we do other retries (like for example to try other media nodes)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5939,6 +5939,13 @@ export default class Meeting extends StatelessWebexPlugin {
 
     await this.cleanUpBeforeRetryWithTurnServer();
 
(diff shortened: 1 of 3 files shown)


jor-row (approved):
LGTM 👍🏻 I like the generic metric with 'reason' field

chburket (approved):

maxinteger (approved):

PR #3333, opened 2024-01-26
fix(plugin-meetings): consider undefined return type

shnaaz:

< DESCRIBE YOUR CHANGES >
getPermissionTokenExpiryInfo returns undefined as well, this PR makes sure there is protection in code for that.
We epect typescript to complain but in sdk repo a function returning undefined is considered as having return type of 'any' hence typing getPermissionTokenExpiryInfo didn't help. Will raise a jira so that we can identify the cause of that, could be some misconfiguration

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1525,7 +1525,7 @@ export default class Meeting extends StatelessWebexPlugin {
       : this.destination;
     const destinationType = isStartingSpaceInstantV2Meeting ? _MEETING_LINK_ : this.destinationType;
 
-    const {timeLeft, expiryTime, currentTime} = this.getPermissionTokenExpiryInfo();
+    const {timeLeft, expiryTime, currentTime} = this.getPermissionTokenExpiryInfo() || {};
 
     LoggerProxy.logger.info(
       `Meeting:index#refreshPermissionToken --> refreshing permission token, destinationType=${destinationType}, timeLeft=${timeLeft}, permissionTokenExpiry=${expiryTime}, currentTimestamp=${currentTime},reason=${reason}`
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1525,7 +1525,9 @@ export default class Meeting extends StatelessWebexPlugin {
       : this.destination;
     const destinationType = isStartingSpaceInstantV2Meeting ? _MEETING_LINK_ : this.destinationType;
 
-    const {timeLeft, expiryTime, currentTime} = this.getPermissionTokenExpiryInfo();
+    const timeLeft = this.getPermissionTokenExpiryInfo()?.timeLeft;
+    const expiryTime = this.getPermissionTokenExpiryInfo()?.expiryTime;
```
Coread, line 1529: suggestion: call the function once and extract after

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -9795,7 +9840,7 @@ describe('plugin-meetings', () => {
```
nataliadelmar, line 9840: we should add a text to see if nothing breaks if we do not get a numberable string as the exp. For example, 
- if we get 'woops' as the exp. Number('woops') will return NaN. How should the function behave then?
- if we get '{something: 'hello'} as the exp. Number(Object) will return NaN too. How should the function behave then?

nataliadelmar (approved):

torgeadelin (approved):

robstax (approved):

torgeadelin:
i think we need a 5th one ::))))))

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7962,11 +7966,7 @@ export default class Meeting extends StatelessWebexPlugin {
    *
    * @returns {object} containing timeLeft, expiryTime, currentTime
    */
-  public getPermissionTokenExpiryInfo(): {
-    timeLeft?: number;
-    expiryTime?: number;
-    currentTime: number;
-  } {
+  public getPermissionTokenExpiryInfo() {
```
marcin-bazyl, line 7969: wouldn't it be simpler to just change code in line 7971 to `return {};`?

shnaaz: We decided against that, if there is no payload then that means there is nothing and returning empty object is not ideal

jor-row (approved):
was feeling left out

PR #3334, opened 2024-01-26
fix(plugin-meetings): don't destroy meeting reference on cleanup of reconnection manager

jor-row:
# COMPLETES [SPARK-461372](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-461372)

If we leave a meeting we clean up the reconnection manager and set reconnectionManager.meeting to null. But, if we rejoin that meeting in the same session the meeting object is the same but a new ReconnectionManager instance isn't created as this only happens when a new Meeting instance is created. This means that reconnection attempts on a second join of a meeting always fail.

don't set meeting object reference to null on clean up

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -5639,6 +5642,16 @@ describe('plugin-meetings', () => {
             });
           });
 
+          it('should reconnect successfully if reconnectionManager.cleanUp is called before reconnection attempt', async () => {
```
jor-row, line 5645: I tried to write a test that checked that validate() was called twice, but each time validate is called it's a different instance of ReconnectionManager and the functions in ReconnectionManager are no longer stubbed once it is overwritten with a new instance so it wasn't as simple as just checking ```assert.calledTwice(meeting.reconnectionManager.validate)```.

marcin-bazyl (approved):

maxinteger (approved):

torgeadelin (approved):

PR #3335, opened 2024-01-26
fix(): adding retry timer for locus service unavailable

shivani1211:
# COMPLETES # https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-474918

If the client receives a 503 error message and if the Retry-After header is present, then client will retry with the same request based on the estimated time, if there is no Retry-After header, then client will immediately retry the exact same request one more time before it notifies the user of the error.

Added an interceptor to catch locus 503 errors and added a retry logic using setTimer to retry same request after Retry-After time (if given) or 2000ms (default)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/index.ts
+++ b/packages/@webex/plugin-meetings/src/index.ts
@@ -3,9 +3,13 @@ import {registerPlugin} from '@webex/webex-core';
 
 import Meetings from './meetings';
 import config from './config';
+import {LocusRetryStatusInterceptor} from './interceptors';
 
 registerPlugin('meetings', Meetings, {
   config,
+  interceptors: {
+    LocusRetryStatusInterceptor: LocusRetryStatusInterceptor.create,
+  },
 });
 
 export {
--- a/packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
+++ b/packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
@@ -24,7 +24,7 @@ export default class LocusRetryStatusInterceptor extends Interceptor {
    * @returns {Promise<WebexHttpError>}
    */
   onResponseError(options, reason) {
-    if (reason.statusCode === 503 && options.uri.includes('locus')) {
+    if ((reason.statusCode === 503 || reason.statusCode === 429) && options.uri.includes('locus')) {
       const hasRetriedLocusRequest = rateLimitExpiryTime.get(this);
       const retryAfterTime = options.headers['retry-after'] || 2000;
 
--- a/packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
+++ b/packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
@@ -1,0 +1,67 @@
+/*!
+ * Copyright (c) 2015-2020 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {Interceptor} from '@webex/http-core';
+
+const rateLimitExpiryTime = new WeakMap();
+/**
+ * @class
+ */
+export default class LocusRetryStatusInterceptor extends Interceptor {
+  /**
+   * @returns {LocusRetryStatusInterceptor}
+   */
+  static create() {
+    // @ts-ignore
+    return new LocusRetryStatusInterceptor({webex: this});
+  }
+
+  /**
+   * Handle response errors
+   * @param {Object} options
+   * @param {WebexHttpError} reason
+   * @returns {Promise<WebexHttpError>}
+   */
+  onResponseError(options, reason) {
+    if (reason.statusCode === 503 && options.uri.includes('locus')) {
+      const hasRetriedLocusRequest = rateLimitExpiryTime.get(this);
+      const retryAfterTime = options.headers['retry-after'] || 2000;
+
+      if (hasRetriedLocusRequest) {
+        rateLimitExpiryTime.set(this, false);
+
+        return Promise.reject(options);
+      }
+      rateLimitExpiryTime.set(this, true);
+
+      return this.handleRetryRequestLocusServiceError(options, retryAfterTime);
+    }
+
+    return Promise.reject(reason);
+  }
+
+  /**
+   * Handle retries for locus service unavailable errors
+   * @param {Object} options associated with the request
+   * @param {number} retryAfterTime retry after time in milliseconds
+   * @returns {Promise}
+   */
+  handleRetryRequestLocusServiceError(options, retryAfterTime) {
+    return new Promise((resolve, reject) => {
+      const timeout = setTimeout(() => {
+        clearTimeout(timeout);
+
+        // @ts-ignore
+        this.webex
+          .request({
+            method: options.method,
+            uri: options.uri,
+            body: options.body,
+          })
+          .then(resolve)
+          .catch(reject);
+      }, retryAfterTime);
+    });
+  }
+}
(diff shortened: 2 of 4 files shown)


packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright (c) 2015-2020 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {Interceptor} from '@webex/http-core';
+
+const rateLimitExpiryTime = new WeakMap();
+/**
+ * @class
+ */
+export default class LocusRetryStatusInterceptor extends Interceptor {
```
chrisadubois, line 11: I dont understand why this is a Locus specific retry status interceptor. There's nothing in here specific to Locus. I know it seems like it's just a naming thing but this retry interceptor should be usable for other services that need to be retried based on unavailability.

Also is there any reason why this retry interceptor doesn't handle the x-retry-after-header for 429s as well as 503 ? why can't this be a generic http retry interceptor that could be extended?

packages/@webex/plugin-meetings/src/index.ts
```diff
@@ -3,9 +3,13 @@ import {registerPlugin} from '@webex/webex-core';
 
 import Meetings from './meetings';
 import config from './config';
+import {LocusRetryStatusInterceptor} from './interceptors';
 
 registerPlugin('meetings', Meetings, {
   config,
+  interceptors: {
+    LocusRetryStatusInterceptor: LocusRetryStatusInterceptor.create,
```
chrisadubois, line 11: question: how does this actually apply the interception to the locus service URL?

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Handle response errors
+   * @param {Object} options
+   * @param {WebexHttpError} reason
+   * @returns {Promise<WebexHttpError>}
+   */
+  onResponseError(options, reason) {
+    if (reason.statusCode >= 500 && reason.statusCode < 600 && options.uri.includes('locus')) {
```
chrisadubois, line 27: I see this here, but why wouldn't this be parameterized?

Also shouldn't this be specific to 503? not 5xx ?

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,67 @@
...
+      const hasRetriedLocusRequest = rateLimitExpiryTime.get(this);
+      const retryAfterTime = options.headers['retry-after'] || 2000;
```
chrisadubois, line 29: Why doesn't this do exponential backoff with a parameterized exponential backoff start and end?

why 2000?

packages/@webex/plugin-meetings/src/index.ts
InteractiveTimmy: When the meeting plugin is mounted, this interceptor should be mounted as well.

This then gets applied to all requests, see the `on{phase}` methods written within the child-interceptor's class.

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
InteractiveTimmy: This is incorrect. This is specifically a single retry for locus. This does not perform any backoff logic. Other services try a different regional cluster when a failure of this scope occurs (it marks the host node as failed, see the services plugin catalog [ `packages/@webex/webex-core/src/lib/services/service-catalog.js` ]. ) This is a server-side failover workflow specific for locus.

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
InteractiveTimmy: This could be the case, the JIRA appears to read that it needs to be specifically 503.

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
InteractiveTimmy: The requirement is **retry once only**, based on the `retry-after` header, or **IMMEDIATELY** after the 503 is received. We MUST place a short delay (1-3 seconds) otherwise the service does not have the time to correct its routing.

InteractiveTimmy (changes requested):

packages/@webex/plugin-meetings/test/unit/spec/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,130 @@
...
+                clock.tick(retryAfterTime);
+
+                return promise.then(() => {
+                    expect(interceptor.webex.request.calledOnce).to.be.true;
+                    });
+            });
+        });
+    });
+    });
+});
```
InteractiveTimmy, line 130: needs new line at EOF to prevent our static analysis from throwing.

InteractiveTimmy (changes requested):
1-2 small changes requested, and comments.

InteractiveTimmy (approved):
LGTM! Thanks for making the change requests!

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
chrisadubois: IMO this should just do a normal exponential backoff if there is no retryAfterTime, (that could be a follow up PR - ill leave it to shivani to decide)

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Handle response errors
+   * @param {Object} options
+   * @param {WebexHttpError} reason
+   * @returns {Promise<WebexHttpError>}
+   */
+  onResponseError(options, reason) {
+    if (reason.statusCode === 503 && options.uri.includes('locus')) {
```
chrisadubois, line 27: should be 5xx || 429, right?

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
```diff
@@ -0,0 +1,67 @@
...
+      const retryAfterTime = options.headers['retry-after'] || 2000;
+
+      if (hasRetriedLocusRequest) {
+        rateLimitExpiryTime.set(this, false);
+
+        return Promise.reject(options);
+      }
+      rateLimitExpiryTime.set(this, true);
+
+      return this.handleRetryRequestLocusServiceError(options, retryAfterTime);
```
chrisadubois, line 38: see this is where it could be the standard exponential backoff flow. We have code that does exponential backoffs in the web client.

chrisadubois (approved):
Approved because this is progress but i see some fairly obvious followups.

Code overall looks good.

packages/@webex/plugin-meetings/src/interceptors/locusRetry.ts
shivani1211: the JIRA mentions the request needs to be retried only once again - which makes sense to me because service unavailable should be getting resolved w load balancing on the backend - which should be immediate in-csae of failover imo, but I can create another followup JIRA for this

PR #3337, opened 2024-01-29
feat(plugin-meetings): TCP reachability checks

marcin-bazyl:
# COMPLETES [SPARK-468017](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-468017)

Adds support for TCP reachability checks.

Had to refactor the whole Reachability class - I've split it into:
- Reachability - existing class, manages the overall reachability process for all media clusters
- a new class ClusterReachability - manages the reachability checks for individual media cluster (does UDP and TCP checks)

Also changed some types. We used to store the reachability in the format that backend expects (it's a bit weird as everything is a string in it, there are no booleans or numbers). I've changed it so that we store it in whatever format we think is best and only convert to the backend format when getReachabilityResults() is called.

TCP reachability works by Linus pretending to be a TURN server. We do it this way, because we are able to specify the protocol in TURN URLs. The same cannot be done for STUN server URLs, because browsers send STUN requests only over UDP.

If anyone wants to test it, TCP reachability works in integration and also seems to work in prod for hologrambts.* and canary.* nodes

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests and sample app with and without VPN

### I certified that

---


packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,325 @@
...
+        }
+        if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
+          if (this.result.udp.latencyInMilliseconds === undefined) {
+            const elapsed = this.getElapsedTime();
+
+            LoggerProxy.logger.log(
+              // @ts-ignore
+              `Reachability:index#onIceCandidate --> Successfully reached ${this.name} over UDP: ${elapsed}ms`
+            );
+            this.result.udp.latencyInMilliseconds = elapsed.toString();
```
torgeadelin, line 229: needs to be a string?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,325 @@
...
+      const CANDIDATE_TYPES = {
+        SERVER_REFLEXIVE: 'srflx',
+        RELAY: 'relay',
+      };
+
+      if (e.candidate) {
+        if (e.candidate.type !== 'host') {
+          console.log('marcin: e.candidate', e.candidate);
+        }
+        if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
```
torgeadelin, line 221: these 2 ifs are the same, maybe we can abstract it into a function, should be easier to test and reduces duplication

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
torgeadelin: also does reflexive map to UDP, and relay to TCP?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,325 @@
...
+        reachable: this.numTcpUrls > 0 ? 'false' : undefined,
+        untested: this.numTcpUrls === 0 ? 'true' : undefined,
+      },
+      xtls: {
+        untested: 'true',
+      },
+    };
+
+    try {
+      const offer = await this.pc.createOffer({offerToReceiveAudio: true});
```
torgeadelin, line 289: why only audio?

torgeadelin: or is this "everything but ignore sending audio"

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      this.startTimestamp = performance.now();
+
+      this.pc.setLocalDescription(offer); // not awaiting on purpose
```
torgeadelin, line 293: what is the purpose

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -265,407 +231,72 @@ export default class Reachability {
...
-   * @param {boolean} isVideoMesh
-   * @returns {Promise}
-   */
-  private iceGatheringState(peerConnection: RTCPeerConnection, isVideoMesh: boolean) {
-    const ELAPSED = 'elapsed';
-
-    const timeout = isVideoMesh ? VIDEO_MESH_TIMEOUT : DEFAULT_TIMEOUT;
-
-    return new Promise<ICECandidateResult>((resolve) => {
-      const peerConnectionProxy = new window.Proxy(peerConnection, {
```
torgeadelin, line 458: this was so unnecessarily complicated, thanks for somplifying it

torgeadelin (commented):
much easier to understand now, looks great

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
torgeadelin: also does reflexive = UDP, and relay = TCP?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: the call to `setLocalDescription()` with the offer triggers browser to start gathering ICE candidates and contacting STUN and TURN servers

marcin-bazyl: we don't await it, because we don't care about it completing, we're just waiting for the ICE candidates that we get through the listeners

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: in order to have ICE candidates we need to have at least 1 m-section in the SDP, so we need at least 1 audio or 1 video transport or a data channel, it doesn't matter what it is, we won't be sending or receiving anything anyway. It could be done in a different way by calling `addTransceiver()` or `createDataChannel()` but the existing code was doing `offerToReceiveAudio` so I kept it.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: > also does reflexive = UDP, and relay = TCP?

yes, it's because of how we do it - we use Linus as STUN server to check UDP reachability, so they generate reflexive candidates. Webrtc only supports talking to STUN servers over UDP. For TCP we use Linus pretending to be a TURN server (that generates RELAY candidates) - see `buildPeerConnectionConfig()` where we force the TURN connection to be done only over TCP

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: yes, it's weird but that's what the backend expects, there are comments about it in the type definition - see `TransportResult`, so I just keep the data always in that format. We could have our own type for all the internal logic and storing of the results and only convert to the type expected by the backend just before sending it. I don't mind it that much which way we do it. Do you think it would be better to change it?

torgeadelin: No, i missed the comment in the type -- if that's what the backend wants it's ok

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -59,10 +55,6 @@ tokenElm.addEventListener('change', (event) => {
   localStorage.setItem('date', new Date().getTime() + (12 * 60 * 60 * 1000));
 });
 
-if (isSafari || isiOS) {
-  document.getElementById('sendShareToggle').disabled = true;
```
marcin-bazyl, line 63: this is just something I've found - it prevents the sample app from working in Safari at all, because "sendShareToggle" doesn't exist anymore

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: in the end I've decided to change it and have 2 separate types: 
`ReachabilityResults` - that's our internal type
`ReachabilityResultsForBackend` - that's the weird type that backend wants from us
gives us freedom to store the data however we want it

packages/@webex/plugin-meetings/src/config.ts
```diff
@@ -88,6 +88,7 @@ export default {
       enableMediaNegotiatedEvent: false,
       enableUnifiedMeetings: true,
       enableAdhocMeetings: true,
+      enableTcpReachability: false,
```
marcin-bazyl, line 91: I had to add this, because Linus support for TCP reachability is only available in integration for now. We can remove this toggle once backend supports it in prod

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: moved the common code from the 2 if blocks into `storeLatencyResult()`

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+    try {
+      const config = this.buildPeerConnectionConfig(clusterInfo);
+
+      const peerConnection = new RTCPeerConnection(config);
+
+      return peerConnection;
+    } catch (peerConnectionError) {
+      LoggerProxy.logger.log(
+        `Reachability:index#createPeerConnection --> Error creating peerConnection: ${peerConnectionError}`
+      );
```
jor-row, line 125: nit pick: given then this is an error, would a logger.warn or logger.info be better here?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+    return this.result;
+  }
+
+  /**
+   * Starts the process of gathering ICE candidates
+   *
+   * @returns {Promise} promise that's resolved once reachability checks for this cluster are completed or timeout is reached
+   */
+  private gatherIceCandidates() {
+    const timeout = this.isVideoMesh ? VIDEO_MESH_TIMEOUT : DEFAULT_TIMEOUT;
```
jor-row, line 306: question: I'm curious what the reason for the difference in timeouts is?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+    return ['udp', 'tcp'].every(
+      (protocol) =>
+        this.result[protocol].result === 'reachable' ||
+        this.result[protocol].result === 'not tested'
+    );
+  }
+
+  /**
+   * Stores the latency in the result
+   *
```
jor-row, line 211: suggestion: adding to the jsdoc that this is what marks the result as 'reachable' might be useful as well

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+
+import {ICE_GATHERING_STATE, CONNECTION_STATE} from '../constants';
+
+const DEFAULT_TIMEOUT = 3000;
+const VIDEO_MESH_TIMEOUT = 1000;
+
+// result for a specific transport protocol (like udp or tcp)
+export type TransportResult = {
+  result: 'reachable' | 'unreachable' | 'not tested';
+  latencyInMilliseconds?: number;
```
jor-row, line 14: suggestion: although it's clear from reading the code, maybe a comment here that it's the latency for reaching the first candidate and not the latency for completing all the reachability checks for that protocol type

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,279 @@
...
+      fakePeerConnection.onicecandidate({candidate: {type: 'relay', address: 'someTurnRelayIp2'}});
+
+      await clock.tickAsync(10);
+      fakePeerConnection.onicecandidate({candidate: {type: 'relay', address: 'someTurnRelayIp3'}});
+
+      await clock.tickAsync(3000);// move the clock so that reachability times out
+
+      await promise;
+
+      // latency should be from only the first candidates, but the clientMediaIps should be from all candidates
```
jor-row, line 243: issue: comment isn't accurate as we aren't storing IPs for tcp results

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,279 @@
...
+      await promise;
+
+      assert.deepEqual(clusterReachability.getResult(), {
+        udp: {result: 'reachable', latencyInMilliseconds: 10, clientMediaIPs: ['somePublicIp1', 'somePublicIp2']},
+        tcp: {result: 'reachable', latencyInMilliseconds: 40},
+        xtls: {result: 'not tested'}
+      });
+    });
+  });
+});
```
jor-row, line 279: nit pick: file is missing empty line at the end

jor-row (changes requested):
Hoping the manual test before approving, just leaving these comments in the mean time

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+import {ClusterNode} from './request';
+
+import {ICE_GATHERING_STATE, CONNECTION_STATE} from '../constants';
+
+const DEFAULT_TIMEOUT = 3000;
+const VIDEO_MESH_TIMEOUT = 1000;
+
+// result for a specific transport protocol (like udp or tcp)
+export type TransportResult = {
+  result: 'reachable' | 'unreachable' | 'not tested';
```
chrisadubois, line 13: Might be good for 'not tested' to be changed to `untested` - no spaces? nitpick obviously but sometimes spaces cause problems

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+  latencyInMilliseconds?: number;
+  clientMediaIPs?: string[];
+};
+
+// reachability result for a specific media cluster
+export type ClusterReachabilityResult = {
```
chrisadubois, line 19: <3

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
chrisadubois, line 13: This might also be better as a const enum

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+
+/**
+ * A class that handles reachability checks for a single cluster.
+ */
+export class ClusterReachability {
+  private numUdpUrls: number;
+  private numTcpUrls: number;
+  private result: ClusterReachabilityResult;
+  private pc?: RTCPeerConnection;
+  private defer: Defer; // this defer is resolved once reachability checks for this cluster are completed
```
chrisadubois, line 33: nice

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+export type ClusterReachabilityResult = {
+  udp: TransportResult;
+  tcp: TransportResult;
+  xtls: TransportResult;
+};
+
+/**
+ * A class that handles reachability checks for a single cluster.
+ */
+export class ClusterReachability {
```
chrisadubois, line 28: Should this implement an interface?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+   * @returns {RTCConfiguration} peerConnectionConfig
+   */
+  private buildPeerConnectionConfig(cluster: ClusterNode): RTCConfiguration {
+    const udpIceServers = cluster.udp.map((url) => ({
+      username: '',
+      credential: '',
+      urls: [url],
+    }));
+
+    // STUN servers are contacted only using UDP, so in order to test TCP reachability
```
chrisadubois, line 85: nice comment

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+    // we pretend that Linus is a TURN server, because we can explicitly say "transport=tcp" in TURN urls.
+    // We then check for relay candidates to know if TURN-TCP worked (see registerIceCandidateListener()).
+    const tcpIceServers = cluster.tcp.map((urlString: string) => {
+      // urlString looks like this: "stun:external-media91.public.wjfkm-a-10.prod.infra.webex.com:5004"
+      // and we need it to be like this: "turn:external-media91.public.wjfkm-a-10.prod.infra.webex.com:5004?transport=tcp"
+      const url = new URL(urlString);
```
chrisadubois, line 91: Can this url munging be moved to a utility function?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
chrisadubois: I think the comment here explaining it would be useful

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: Good question :-) It's because we used to have a limitation that reachability for VMN nodes would always fail if you are on the same network as the video mesh nodes (so when you're on the corporate network or using VPN), so we wanted to reduce the timeout for VMN just to minimize the overall time it takes to do reachability. That limitation is gone now (after a Linus change was done last year), so we could consider removing that code and have same timeout for everybody.
Although thinking about it a bit more, before we do that we would need to better understand if customer's VMN nodes are usually reachable from public internet or not, because if they're only reachable from the corporate LAN, then they are would fail when people use our app from home without VPN and we would want to keep the short timeout for them. Also since VMN should be on the same corporate network, it makes sense to have a shorter timeout for them.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -0,0 +1,324 @@
...
+   * @returns {Promise} promise that's resolved once reachability checks for this cluster are completed or timeout is reached
+   */
+  private gatherIceCandidates() {
+    const timeout = this.isVideoMesh ? VIDEO_MESH_TIMEOUT : DEFAULT_TIMEOUT;
+
+    this.registerIceGatheringStateChangeListener();
+    this.registerIceCandidateListener();
+
+    // Set maximum timeout
+    setTimeout(() => {
```
torgeadelin, line 312: do we need to clear this at any point?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
marcin-bazyl: done

torgeadelin (approved):
looks good, only one question but non blocking

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
marcin-bazyl: done (also in 1 other place)

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done, changed to warn as user might still be able to join meetings

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: the first thing that the timer code does is it checks the RTCPeerConnection state if it's closed, so it won't do anything.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: yeah, but I prefer this union type approach for things that have only 2 or 3 values, it's more light-weight and this type is not exported outside of SDK, only used internally, so I think it's fine

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: why? we don't have any plans to have any other instances implementing the same interface in any other way

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

docs/samples/browser-plugin-meetings/index.html
```diff
@@ -78,12 +78,14 @@ <h2 class="collapsible">Meeting Management</h2>
                 <legend>Environment</legend>
                 <input type="checkbox" id="integration-env">
                 <label for="integration-env">Enable Integration Environment</label><br>
+                <input type="checkbox" id="enable-fedramp">
+                <label for="enable-fedramp">Enable Fedramp</label>
               </div>
 
               <div>
-                <legend>FedRAMP</legend>
```
arun3528, line 84: why removing the fedramp check ?/

arun3528 (commented):
looks good confirming on the fedramp toggle we need for the sample page

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
jor-row: Interesting, thanks for the explanation!

docs/samples/browser-plugin-meetings/index.html
marcin-bazyl: I didn't remove it, it's still there, I just moved it up to the environment section:
image

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -110,7 +102,7 @@ function generateWebexConfig({credentials}) {
         enableMediaNegotiatedEvent: false,
         enableUnifiedMeetings: true,
         enableAdhocMeetings: true,
-        enableTurnDiscovery: turnDiscoveryCheckbox.checked,
```
marcin-bazyl, line 113: `enableTurnDiscovery` config flag was removed in PR#3261 but then it was incorrectly re-introduced in PR#3262 (probably bad merge conflict), so I'm removing it again, because it's no longer needed, the feature was GA-ed over a year ago

docs/samples/browser-plugin-meetings/index.html
marcin-bazyl: actually, now after merging in latest beta, the UI looks a bit different, so I've kept Fedramp in the same position as it is and removed turn discovery checkbox as turn discovery config option has already been removed a few weeks ago, so now the sample app looks like this:
image

packages/@webex/plugin-meetings/src/config.ts
sreenara: @marcin-bazyl could we add a TODO for this? How are we tracking the removal of this config?

sreenara (commented):
Just 1 comment regarding the changes

packages/@webex/plugin-meetings/src/config.ts
marcin-bazyl: I've raised https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-496486 to track this

PR #3339, opened 2024-01-29
fix(plugin-meetings): generate types 

mkesavan13:
# COMPLETES Adhoc PR

The `dist/types` isn't getting generated for both `plugin-meetings` and `internal-plugin-metrics`

This PR addresses that issue by

- Adding the right build command to both package.json scripts
- Building webex-core prior to other packages in the CI scripts

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Build command was executed for the whole package to see if types were generated
- Separate package scripts were also executed to see if types get generated

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -3642,6 +3642,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   closeRemoteStreams() {
+    // TODO: Discuss with Marcin and see how to add remoteShareAudioStream back
     const {remoteAudioStream, remoteVideoStream, remoteShareStream} = this.mediaProperties;
 
     /**
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -3642,7 +3642,6 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   closeRemoteStreams() {
-    // TODO: Discuss with Marcin and see how to add remoteShareAudioStream back
     const {remoteAudioStream, remoteVideoStream, remoteShareStream} = this.mediaProperties;
 
     /**
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -67,7 +67,10 @@ jobs:
       - name: Build Tools
         run: yarn workspaces foreach --from '@webex/*-tools' --topological-dev --parallel --verbose run build:src
 
-      - name: Build
+      - name: Build Webex Core
+        run: yarn workspace @webex/webex-core run build:src
+
+      - name: Build Other Packages
         run: yarn workspaces foreach --parallel --verbose run build:src
 
       - name: Cache Distributables
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -156,7 +156,10 @@ jobs:
       - name: Build Tools
         run: yarn workspaces foreach --from '@webex/*-tools' --topological-dev --parallel --verbose run build:src
 
-      - name: Build
+      - name: Build Webex Core
+        run: yarn workspace @webex/webex-core run build:src
+
+      - name: Build Other Packages
         run: yarn workspaces foreach --parallel --verbose run build:src
 
       - name: Cache Distributables
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "prebuild": "yarn run clean",
     "build:publish": "yarn build:package && yarn workspaces foreach --parallel --verbose run build:src",
     "build:tools": "yarn workspaces foreach --from '@webex/*-tools' --topological-dev --parallel --verbose run build:src",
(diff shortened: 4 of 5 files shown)


Shreyas281299 (approved):
Looks good

packages/@webex/internal-plugin-metrics/package.json
```diff
@@ -45,8 +45,8 @@
     "@webex/webex-core": "workspace:^"
   },
   "scripts": {
-    "build": "yarn build:src && yarn run -T tsc --declaration true --declarationDir ./dist/types",
-    "build:src": "webex-legacy-tools build -dest \"./dist\" -src \"./src\" -js -ts -maps",
+    "build": "yarn run -T tsc --declaration true --declarationDir ./dist/types",
+    "build:src": "webex-legacy-tools build -dest \"./dist\" -src \"./src\" -js -ts -maps && yarn build",
```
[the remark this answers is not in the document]
arun3528: For the packages which can be transpiled using tsc we should , the one which we cannot could use the webex-legacy-tool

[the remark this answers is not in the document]
mkesavan13: I think `build:src` is the default command for all the packages as per new tooling and in `plugin-meetings` and `internal-plugin-metrics` alone, we have to build the types as they are the only packages that are migrated to typescript at the moment. 

Having too many commands in the same `build:src` would be confusing I believe. That's why two different commands. We need to clean up the scripts. We shall take that later.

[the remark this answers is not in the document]
marcin-bazyl: > as they are the only packages that are migrated to typescript at the moment

media-helpers is another package that uses typescript

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -217,10 +217,10 @@ Media.createMediaConnection = (
     {
       // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now SPARK-460530
       localTracks: {
-        audio: audioStream?.outputTrack,
-        video: videoStream?.outputTrack,
-        screenShareVideo: shareVideoStream?.outputTrack,
-        screenShareAudio: shareAudioStream?.outputTrack, // TODO: add type for screenShareAudio in internal-media-core SPARK-446923
+        audio: audioStream?.outputStream.getTracks()[0],
```
marcin-bazyl, line 220: we should instead change webrtc-core and have the `outputTrack` as public instead of protected

sreenara (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3605,8 +3605,8 @@ export default class Meeting extends StatelessWebexPlugin {
    * @memberof Meeting
    */
   closeRemoteStreams() {
-    const {remoteAudioStream, remoteVideoStream, remoteShareStream, shareAudioStream} =
-      this.mediaProperties;
```
marcin-bazyl, line 3609: Why are we doing this PR on next and not beta branch? Beta is the only source of truth for plugin meetings, the agreement was that for plugin-meetings we will take all of the source from beta. There should be nothing on the next branch apart from maybe some package.json changes

packages/@webex/plugin-meetings/src/media/index.ts
mkesavan13: Had a discussion with @brycetham for the same. He suggests we shall do this instead of changing the `outputTrack` to `public`

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: The problem here is that `shareAudioStream` isn't a remote stream and it is of type `LocalSystemAudioStream`

This is causing issue when we make call to `stopStream` which only accepts `RemoteStream.` Wasn't going to make this a permanent change. Already had a chat with Bryce about this yesterday and he wanted me to discuss this with you. That's why added a Todo

sreenara: @mkesavan13 how do we reconcile this with the code that's in `beta`?

marcin-bazyl: This is a real bug, it exists on beta too, so it should be fixed on beta. It was probably introduced when the WCME 3.0 changes were done and has gone unnoticed, because closeRemoteStreams() is only called on meeting leave and if it accidentally stops local share audio stream, it's not causing any real problem in the web app, because we always have share stopped anyway when we leave the meeting

packages/@webex/internal-plugin-metrics/package.json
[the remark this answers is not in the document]
mkesavan13: Thanks @marcin-bazyl. The `media-helpers` already generate types but not in the types folder. Changed it now to generate in a separate types folder.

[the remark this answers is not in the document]
arun3528: @InteractiveTimmy  for the new packages should we still use the legacy tool or just tsc would work ?

PR #3344, opened 2024-01-30
fix: typescript incorrect type issues

mkesavan13:
# COMPLETES Adhoc PR

A few type issues that were noticed during types generation in the next branch and is an extension to #3339 

This PR addresses that issue by

- Fixing some type issues
- Removing the shareAudioStream in the stopStream method as there's no separate remote share audio stream

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Build command was executed for the whole package to see if types were generated
- Separate package scripts were also executed to see if types get generated

### I certified that

---


sreenara (approved):

packages/@webex/plugin-meetings/src/locus-info/parser.ts
```diff
@@ -690,7 +690,7 @@ export default class Parser {
    */
   private stopSyncTimer() {
     if (this.syncTimer !== null) {
-      clearTimeout(this.syncTimer);
+      clearTimeout(this.syncTimer as number);
```
marcin-bazyl, line 693: this doesn't seem right, wouldn't it be better to fix the type declaration for syncTimer? Like this:
`syncTimer?: ReturnType<typeof setTimeout>`
(and also change all the places that use null with syncTimer to undefined)

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -217,10 +217,10 @@ Media.createMediaConnection = (
     {
       // TODO: RoapMediaConnection is not ready to use stream classes yet, so we pass the raw MediaStreamTrack for now SPARK-460530
       localTracks: {
-        audio: audioStream?.outputTrack,
-        video: videoStream?.outputTrack,
-        screenShareVideo: shareVideoStream?.outputTrack,
-        screenShareAudio: shareAudioStream?.outputTrack, // TODO: add type for screenShareAudio in internal-media-core SPARK-446923
+        audio: audioStream?.outputStream.getTracks()[0],
```
marcin-bazyl, line 220: like I said in the other PR, it would be better fix webrtc-core and have `outputTrack` as public instead of protected

mkesavan13: Quoting 1:1 message from @brycetham :

> But regarding accessing the outputTrack in transcoded meetings, we deliberately changed webrtc-core from track-based to stream-based because it's better to use streams, since you just need to set srcObject = stream and any modifications made to the stream (like adding an effect) should work automatically. So there shouldn't be any need for accessing outputTrack directly.

and another,

> Ok, for now, if you really want to use the outputTrack, you can do stream.outputStream.getTracks()[0], with the understanding that it's a temporary workaround until transcoded meetings switches to stream classes. And actually, I thought we did do this workaround when we worked on the WCME 3.0 changes, so I'm a bit surprised to see this isn't the case in the beta code. But maybe I'm misremembering.

sreenara: @marcin-bazyl let's fix this as part of another PR after @brycetham fixes the webrtc-core for the above change.
We'd like to get this PR merged so we can run the E2E for the web client with `next`

packages/@webex/plugin-meetings/src/locus-info/parser.ts
mkesavan13: This is done @marcin-bazyl

packages/@webex/plugin-meetings/src/media/index.ts
brycetham: @marcin-bazyl @sreenara I'd really like to avoid exposing `outputTrack`, as it's not intended to be exposed as part of the change from track-based classes to stream-based classes in webrtc-core, and it risks having the track modified in ways we don't expect. Ideally, transcoded meetings should be using the stream-based classes as intended, so passing the track here is a temporary workaround, and I'd rather not change webrtc-core for something like this.

marcin-bazyl: ok, fair enough, it is true that we want to move transcoded API of internal-media-core to use the stream based classes, so let's go ahead with this as a temporary workaroud

PR #3351, opened 2024-01-31
Fix(tools) : migrate to localized config

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to update the configuration objects to be locally linkable instead of module aliased to allow for editor SDKs (specifically for VSCode) to operate as intended.

Before these changes, contributors were unable to see any TS, ESLint, or similar errors in their source changes, and were required to debug via CLI only. They were greeted with the following messages:

image

and many many others that can still be witnessed when loading other packages that consume modern and legacy configs at this time.

This work will also result on follow-up work to further improve these workflows.


config/typescript/modern.json
```diff
@@ -0,0 +1,38 @@
+{
+	"compilerOptions": {
```
arun3528, line 2: @InteractiveTimmy  how does output directory be set, in older tsconfig I see 

```
    "outDir": "./dist/module",
    "declarationDir": "./dist/types",
    ```

InteractiveTimmy: these are handled within `./package.json`'s scripts, or can be handled within `./tsconfig.json` for any particular package. By placing this within the `scripts` of the `./package.json`, we can easily set up a workflow for types and modules without needing two tsconfig files.

arun3528 (approved):

PR #3353, opened 2024-02-01
fix(plugin-meetings): Fix timestamp

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484765

Issue is that if client's clock is wrong then we calculate the timeLeft for permission token to expire incorrectly hence refreshing the token when not required.
To work around this we now get the current time when we receive the token, we also extract iat (issued time) from permission token payload and workout the shift in time. We then use this to work out the adjusted current time before calculating the timeLeft for permission token to expire. This fix stops us from  refreshing the token when it' not required.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -566,6 +566,7 @@ export default class Meeting extends StatelessWebexPlugin {
   meetingState: any;
   permissionToken: string;
   permissionTokenPayload: any;
+  tokenReceivedTime: number;
   resourceId: any;
   resourceUrl: string;
   selfId: string;
@@ -3540,6 +3541,7 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   public setPermissionTokenPayload(permissionToken: string) {
     this.permissionTokenPayload = jwt.decode(permissionToken);
+    this.tokenReceivedTime = new Date().getTime();
   }
 
   /**
@@ -7979,16 +7981,20 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     const permissionTokenExpValue = Number(this.permissionTokenPayload.exp);
+    const permissionTokenIssuedTime = Number(this.permissionTokenPayload.iat);
+
+    const shiftInTime = this.tokenReceivedTime - permissionTokenIssuedTime;
 
     // using new Date instead of Date.now() to allow for accurate unit testing
     // https://github.com/sinonjs/fake-timers/issues/321
-    const now = new Date().getTime();
+    const currentTime = new Date().getTime();
 
-    // substract current time from the permissionTokenExp
-    // (permissionTokenExp is a epoch timestamp, not a time to live duration)
-    const timeLeft = (permissionTokenExpValue - now) / 1000;
+    // This is done incase your machine time is wrong
+    const adjustedCurrentTime = currentTime - shiftInTime;
 
-    return {timeLeft, expiryTime: permissionTokenExpValue, currentTime: now};
+    const timeLeft = (permissionTokenExpValue - adjustedCurrentTime) / 1000;
+
+    return {timeLeft, expiryTime: permissionTokenExpValue, currentTime};
   }
 
   /**
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -566,7 +566,7 @@ export default class Meeting extends StatelessWebexPlugin {
   meetingState: any;
   permissionToken: string;
   permissionTokenPayload: any;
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7979,16 +7981,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    // adjusted time is calculated in case your machine time is wrong
+    const adjustedCurrentTime = currentTime - shiftInTime;
 
-    // substract current time from the permissionTokenExp
-    // (permissionTokenExp is a epoch timestamp, not a time to live duration)
-    const timeLeft = (permissionTokenExpValue - now) / 1000;
+    const timeLeft = (permissionTokenExpValue - adjustedCurrentTime) / 1000;
 
-    return {timeLeft, expiryTime: permissionTokenExpValue, currentTime: now};
+    return {timeLeft, expiryTime: permissionTokenExpValue, currentTime};
```
maxinteger, line 7997: issue: if we adjust the `timeLeft` then we should do the same for `expiryTime`, so `expiryTime` should be `permissionTokenExpValue - shiftInTime`

maxinteger (changes requested):

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: No, expiryTime is being received in permissionToken payload, we are not adjusting timeLeft we are adjusting the current local time

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3540,6 +3541,7 @@ export default class Meeting extends StatelessWebexPlugin {
    */
   public setPermissionTokenPayload(permissionToken: string) {
     this.permissionTokenPayload = jwt.decode(permissionToken);
+    this.tokenReceivedTime = new Date().getTime();
```
marcin-bazyl, line 3544: it would be good to group the `permissionTokenPayload` and `tokenReceivedTime` under some common property in the Meeting object like "permissionTokenInfo" or something like that, otherwise we end up with a lot of scattered properties in the Meeting object, especially that the name "tokenReceivedTime" doesn't say which token it is about

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7979,16 +7981,20 @@ export default class Meeting extends StatelessWebexPlugin {
     }
 
     const permissionTokenExpValue = Number(this.permissionTokenPayload.exp);
+    const permissionTokenIssuedTime = Number(this.permissionTokenPayload.iat);
```
marcin-bazyl, line 7984: maybe it would be worth renaming the variables that store values read from the permission token (iat and exp) so that it's clear that these are server time values not local client time ones

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: `tokenReceivedTime` can be renamed to `permissionTokenReceivedLocalTime`, if I remove `permissionTokenPalyload` from the meeting object and move it into another property say 'permissionTokenInfo' as suggested then that will have lot more code/unittest changes. Or we can have `permissionTokenPalyload` in 2 places, directly on meetingObject and also in the new property

marcin-bazyl: > then that will have lot more code/unittest changes

but it's just a simple renaming everywhere from `.permissionTokenPayload` to `.permissionTokenInfo.payload`. Anyway, just renaming it to `permissionTokenReceivedLocalTime` is I guess fine too

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -9914,18 +9929,58 @@ describe('plugin-meetings', () => {
...
+      meeting.permissionTokenReceivedLocalTime = now;
       assert.deepEqual(meeting.getPermissionTokenExpiryInfo(), {timeLeft: -1, expiryTime: Number(expiryTime), currentTime: now});
     });
+
+    describe('#getPermissionTokenExpiryInfo with wrong current Time', () => {
+      let now;
+      let clock;
+      beforeEach(() => {
+        // current time is 3 hours off
+        now = Date.now() + 10800000;
```
marcin-bazyl, line 9950: might be worth also adding a test where shiftTime is negative

marcin-bazyl (approved):

maxinteger (approved):

PR #3355, opened 2024-02-02
fix(metrics): fix CA events for missing codec

robstax:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-495685

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -511,6 +514,18 @@ export default class CallDiagnosticMetrics extends StatelessWebexPlugin {
       }
     }
 
+    if (isSdpOfferCreationError(rawError)) {
+      // error code is 30005, but that's not specific enough. we also need to check error.cause.type
+      const causeType = rawError.cause?.type || 'GENERAL';
+
+      return this.getErrorPayloadForClientErrorCode({
+        serviceErrorCode: undefined,
+        clientErrorCode: SDP_OFFER_CREATION_ERROR_MAP[causeType],
```
marcin-bazyl, line 523: SDP_OFFER_CREATION_ERROR_MAP has only entries for GENERAL and SDP_MUNGE_MISSING_CODECS, so when we get any other `cause.type` from WCME we will end up with `undefined` here and we probably would want 2050 in those cases, right?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -388,6 +395,18 @@ export const CLIENT_ERROR_CODE_TO_ERROR_PAYLOAD: Record<number, Partial<ClientEv
     category: 'media',
     fatal: true,
   },
+  2050: {
```
marcin-bazyl, line 398: nitpick: should we be using some constants for these?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
robstax: i wasn't sure about that... do we want the other ones to map to `2050` or each have their own specific code for each specific cause?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
robstax: for the code itself? most of the codes don't in this map, but i am happy to if we'd like. it seemed split so i wasn't sure what pattern to follow.

.vscode/settings.json
```diff
@@ -17,8 +17,8 @@
     "**/dist": true
   },
   "editor.codeActionsOnSave": {
-    "source.fixAll": true,
-    "source.organizeImports": false
+    "source.fixAll": "explicit",
```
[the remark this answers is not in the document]
robstax: yes, vscode kept doing it, but i'll take it out of this commit

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
marcin-bazyl: the code is used in more than 1 place so kind of makes sense to have a const, but on the other hand these numbers are pretty unique, so I don't mind that much - that's why I said it's a "nitpick",

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.ts
marcin-bazyl: if we get any error that's a `Errors.SdpOfferCreationError` no matter what `cause.type` it has, it would be better to map it to 2050, because that has an ErrorDescription of "SdpOfferCreationError" which is better than "unknown"

robstax: makes sense to me. will change.

marcin-bazyl (approved):

chrisadubois (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/config.ts
```diff
@@ -120,6 +125,8 @@ export const ERROR_DESCRIPTIONS = {
   DTLS_HANDSHAKE_FAILED: 'DTLSHandshakeFailed',
   ICE_FAILED_WITHOUT_TURN_TLS: 'ICEFailedWithoutTURN_TLS',
   ICE_FAILED_WITH_TURN_TLS: 'ICEFailedWithTURN_TLS',
+  SDP_OFFER_CREATION_ERROR: 'SdpOfferCreationError',
```
marcin-bazyl, line 128: might be worth adding a comment that this string has to match the name from SdpOfferCreationError class from internal-media-core

arun3528 (approved):

PR #3358, opened 2024-02-02
fix(tools/package): migrate to localized configs

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to further refine the newly created localized packages in `./config` to match all remaining modern tooling. These changes also include deployment optimizations, such as not deploying any unnecessary or duplicated code into the `{package}/dist` folder.


config/typescript/modern.json
```diff
@@ -19,7 +19,8 @@
 			"DOM",
 			"webWorker"
 		],
-		"moduleResolution": "node",
+		"module": "node16",
```
arun3528, line 22: nice

arun3528 (approved):
thanks @InteractiveTimmy

config/typescript/modern.json
InteractiveTimmy: This was **required** for the usage of jest modules importing:

```js
// example.test.js
const Example = require('@{scope}/{name}'); // works now and does not throw an error in tests
const Example = require('../../dist/module/index'); // no longer required to do this
```

config/typescript/modern.json
```diff
@@ -19,7 +19,8 @@
...
+		"moduleResolution": "node16",
```
sreenara, line 23: Would it make more sense to use nodenext instead of node16 as we need to move to a higher node version in the future? Or will making this nodenext start causing things to fail?

packages/tools/cli/src/models/index.ts
```diff
@@ -1,11 +1,7 @@
-import Commands from './commands';
+export { default as Commands } from './commands';
```
sreenara, line 1: This is an interesting export. I've never seen this before. Is there a guide for this?

sreenara: Why doesn't `export Commands from './commands';` work?

packages/tools/cli/src/index.ts
```diff
@@ -1,10 +1,5 @@
-import {
-  Commands,
-} from './models';
-
-import {
-  Executor,
-} from './utils';
+export { Commands } from './models';
```
sreenara, line 1: It seems to be working here/

packages/tools/cli/test/module/executor/executor.test.js
```diff
@@ -13,10 +13,6 @@ describe('Executor', () => {
       const exampleCommand = 'hello world';
       let execSpy;
 
-      afterEach(() => {
```
sreenara, line 16: Is this not needed? Do we remember why it was added before?

packages/tools/cli/package.json
```diff
@@ -17,15 +17,15 @@
...
+    "build:module": "tsc --declaration false --outDir ./dist/module",
     "build:src": "yarn build:module && yarn build:types",
-    "build:types": "tsc --declaration --declarationMap --declarationDir ./dist/types --emitDeclarationOnly --module CommonJS",
+    "build:types": "tsc --declaration --declarationMap --declarationDir ./dist/types --emitDeclarationOnly",
     "clean": "yarn clean:dist",
     "clean:dist": "rimraf ./dist",
     "test": "yarn test:style && yarn test:syntax && yarn test:integration && yarn test:coverage",
     "test:coverage": "yarn test:integration --coverage --reporters=\"jest-silent-reporter\"",
-    "test:integration": "jest ./test/module/**/*.test.js",
+    "test:integration": "jest",
```
sreenara, line 28: Question: How does the jest command now find the required test files?

packages/tools/package/test/module/package/package.test.js
```diff
@@ -101,15 +100,17 @@ describe('Package', () => {
 
       it('should attempt to ammend the package definition file to the package path', () => pack.apply()
         .then(() => {
-          expect(spies.path.join).toHaveBeenCalledOnceWith(
+          expect(spies.path.join).toHaveBeenCalledTimes(1);
```
sreenara, line 103: We have extended jest in packages/webex/calling to add the toBeCalledOnceWith() to combine both the commands. Do you think we could add this to the config you've created for jest in this PR and consume it here and other places?

packages/tools/package/jest.config.js
```diff
@@ -0,0 +1,3 @@
+const { modern } = require('../../../config/jest');
```
sreenara, line 1: Question: So this jest.config.js now pulls the configuration from the config folder and applies that globally to jest commands being run on all packages inside webex, calling etc, unless there is another jest.config.js within a package. Is my understanding right?

sreenara (commented):
Mainly questions around the changes @InteractiveTimmy 
Just one suggestion for jest (toBeCalledOnceWith). Changes look  mostly good to me.

mkesavan13 (approved):
Looks good to me

config/typescript/modern.json
InteractiveTimmy: Since this is TSC options, this affects the distributable. If we move this to node-next, then we potentially lose support for consumers using node16 (as TSC will not build with node16 in mind for the distributable artifact).

We ultimately want to move to node-next (which is ultimately LTS) and require all consumers to be on node LTS (which is now node 20), but this is not the case from the Webex Web Client's side.

packages/tools/cli/src/models/index.ts
InteractiveTimmy: This is a common pattern used to optimize for **tree-shaking**. To improve compatibility with various bundlers and tooling, it is typically the most efficient to never re-declare a only-exported value at its modular de-mark point.

i.e:

```ts
// ./src/models/example/index.ts

import Example from './class.ts'; // this is okay here, as this is the lowest leaf in this scope that we attend to be within the public API

export type { ExampleType } from './types.ts'; // this is a standard type-only re-export. ESLint-TS-Plugin will throw if not done this way

export default Example;

```

```ts
// ./src/models/index.ts

export { default as Example } from './example'; // this is a single line export with no locally declared namespaces, thus tree-shaking is even less likely to miss-interpret this as an IIFE and not tree-shake

export type { ExampleType } from './example'; // standard type re-export, see previous comment above.
```

packages/tools/cli/src/index.ts
InteractiveTimmy: The previous comment should cover this, let me know if this is a different question, thanks!

packages/tools/cli/package.json
InteractiveTimmy: jest, by default, will look for any file that matches a specific test patter (`./**/*.test.*`). I removed the glob pattern from this command so that commands like:

```bash
yarn test:integration ./path/to/specific/test.js
```

would work as expected, and a single file can be tested when debugging. When including the glob pattern, it will always include the globbed tests along with the provided file.

packages/tools/package/test/module/package/package.test.js
InteractiveTimmy: I'm a bit hesitant to utilize a custom BDD clause, as it we'd move away from a standard configuration to a more complex one.

**BUT**, we could definitely add this to the `./config/jest/additions/**` folder.

We should discuss this route and how to make it **expandable** and **readable** in a meeting.

sreenara (approved):

packages/tools/package/test/module/package/package.test.js
sreenara: Sure @InteractiveTimmy , let's discuss this in a meeting next week

config/typescript/modern.json
sreenara: Sounds good. Let's do along with the move to node 20.

packages/tools/package/jest.config.js
InteractiveTimmy: This config file is specifically applied to the `@webex/package-tools` package and its scope. It does not run tests against any packages like the `@webex/legacy-tools` does.

* `@webex/legacy-tools` - Provides tooling to legacy packages ( `./packages/@webex/**` ) for building, testing, etc
* `@webex/package-tools` - Provides CI/CD/automation (and soon downstream tooling) tooling to the root of the package, github actions, and soon to be downstream consumers for organizing and managing package versions and deployments.

PR #3362, opened 2024-02-06
Fix/metrics/browser data refetch

stevem63:
# COMPLETES #[< INSERT LINK TO ISSUE >](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-479762)

We believe there were issues with getting browser data at the moment the file was exported. This PR lets us recapture browser data if it is missing from the object at the time a metric is sent (it will be stored for future use and will only have to happen once in this case).

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/common/src/browser-detection.js
```diff
@@ -44,8 +44,22 @@ const createDetectionObject = (results) => {
...
+const checkBrowserDetection = (agent) => {
+  if (browserDetectionData) {
+    return browserDetectionData;
+  }
+
+  if (!agent || !window.navigator?.userAgent) {
+    return mockDetectionObject;
+  }
+
+  const browserData = createDetectionObject(bowser.getParser(agent || window.navigator.userAgent));
```
InteractiveTimmy, line 58: We should perform a quick check here specifically for the `browser name` to validate it exists on the `browserData` object. If not, we should return the mock detection data again. This should cover the linked task as intended (just for browser name)

packages/@webex/common/src/browser-detection.js
```diff
@@ -27,11 +26,12 @@ const mockDetectionObject = {
...
-  const getOSName = () => results?.getOSName() ?? '';
-  const getOSVersion = () => results?.getOSVersion() ?? '';
+  const getOSName = () => results?.getOSName() ?? browserDetection.unableToGetOSName;
+  const getOSVersion = () => results?.getOSVersion() ?? browserDetection.unableToGetOSVersion;
 
-  const getBrowserName = () => results?.getBrowserName() ?? '';
-  const getBrowserVersion = () => results?.getBrowserVersion() ?? '';
+  const getBrowserName = () => results?.getBrowserName() ?? browserDetection.unableToGetBrowserName;
+  const getBrowserVersion = () =>
+    results?.getBrowserVersion() ?? browserDetection.unableToGetOSVersion;
```
InteractiveTimmy, line 34: Should this be `defaults` and not `error messages`? i.e. maybe the `getOSName()` method should return `unknown` when it is not available?

@chrisadubois , thoughts?

InteractiveTimmy: I'm also curious if the `results` object here goes stale (collected by the garbage collector) and begins returning `null`. This could be a potential root cause for this issue, but just a thought.

InteractiveTimmy (changes requested):
One question (non-blocking), and one change request: one condition should be added (and tested), see comments.

A question for @chrisadubois  as well

packages/@webex/common/src/browser-detection.js
chrisadubois: agreed on unknown vs an error message

packages/@webex/common/src/browser-detection.js
stevem63: Added blow

sreenara:
@stevem63 do we still want to merge this PR? It's been inactive for the last few months.

sreenara:
Closing this due to inactivity.

PR #3364, opened 2024-02-07
fix: use exact versions in workspaces

Shreyas281299:
# COMPLETES #Adhoc

Changes the workspace to use absolute versions rather tha range.

## Reference
https://yarnpkg.com/features/workspaces


sreenara (approved):

sreenara (changes requested):
Please update the top-level package.json too, as well as webex and calling

mkesavan13 (changes requested):
The build is failing. Please check if the yarn works fine locally and push the updated `yarn.lock`

mkesavan13 (approved):

PR #3371, opened 2024-02-09
fix(plugin-meetings): syncMeetings should not filter non-locus meetings

nataliadelmar:
# COMPLETES #[SPARK-484517](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484517)

Meetings SyncMeetings() allows the meeting Collection to only keep the meetings that are valid. "Valid" used to mean that they were valid Locus meetings. But meetings are not necesarily created by Locus, which is why an additional condition was added to the filter.

Only filter meetings if there are a locus meeting (if they have a locusUrl) AND if that locusUrl is still recognized as active by Locus.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -559,43 +559,52 @@ describe('plugin-meetings', () => {
...
-            let parse;
+          describe('destroy non active locus meetings', () => {
             let destroySpy;
 
+            const meetingCollectionMeetings =  {
+              stillValidLocusMeeting: {
+                locusUrl: 'still-valid-locus-url',
+                sendCallAnalyzerMetrics: sinon.stub(),
+              },
+              notLongerValidLocusMeeting: {
```
marcin-bazyl, line 570: `notLonger` -> `noLonger`

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -559,43 +559,52 @@ describe('plugin-meetings', () => {
...
                 Promise.resolve({
-                  loci: [],
+                  loci: [
+                    {
+                      something: 'here',
+                    },
+                  ],
                 })
               );
+              webex.meetings.sortLocusArrayToUpdate = sinon.stub().returns(
```
marcin-bazyl, line 596: I think it would be nicer if we didn't mock this method. After all, all that it's supposed to do is sort the input. If we have to mock it, then maybe at least let's just make it return the same input it is called with instead of hardcoding the output (then you would need to change the stub for getActiveMeetings() to return a meeting with the url "still-valid-locus-url")

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1370,10 +1370,10 @@ export default class Meetings extends WebexPlugin {
 
         if (Object.keys(meetingsCollection).length > 0) {
           // Some time the mercury event is missed after mercury reconnect
-          // if sync returns no locus then clear all the meetings
+          // if sync returns invalid locus events, then clear them and keep the rest
```
marcin-bazyl, line 1373: It would be better to use the term "meetings" instead of "events". Also, the code is removing the meetings from meetingCollection not from the list that sync returns, so I think the comment would be better like this: "remove any Locus meetings that are not returned by Locus in the sync"

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -559,43 +559,46 @@ describe('plugin-meetings', () => {
               });
             });
           });
-          describe('destory non active meeting', () => {
-            let initialSetup;
-            let parse;
+          describe.only('destroy non active locus meetings', () => {
```
marcin-bazyl, line 562: .only

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1369,11 +1369,11 @@ export default class Meetings extends WebexPlugin {
         const meetingsCollection = this.meetingCollection.getAll();
 
         if (Object.keys(meetingsCollection).length > 0) {
-          // Some time the mercury event is missed after mercury reconnect
-          // if sync returns no locus then clear all the meetings
+          // Sometimes the mercury meetings are lost after mercury reconnect
```
marcin-bazyl, line 1372: in this line we actually need to keep it as "events", there is no such thing as "mercury meetings", but we receive "mercury events" over the websocket

marcin-bazyl (approved):

shnaaz (approved):
I have not tested it but we should create cantina bump pr and tests if tap tests pass once pipeline is back to normal

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1369,11 +1369,11 @@ export default class Meetings extends WebexPlugin {
         const meetingsCollection = this.meetingCollection.getAll();
 
         if (Object.keys(meetingsCollection).length > 0) {
-          // Some time the mercury event is missed after mercury reconnect
-          // if sync returns no locus then clear all the meetings
+          // Sometimes the mercury events are lost after mercury reconnect
+          // Remove any Locus meetings that are not returned by Locus in the sync
```
martroos, line 1373: nitpick: Strictly speaking, you aren't removing _any_ meetings not returned by Locus, but those that previously had a locus url and are no longer active. Not sure of a succinct way to phrase that though for the comment though!

martroos (approved):

PR #3375, opened 2024-02-12
fix(call): handle connect after setup and fix send digit

sreenara:
# COMPLETES # BEMS01715628

1. The Call State Machine doesn't handle E_RECV_CONNECT while in the S_SEND_SETUP state. 
2. handleRoapEstablished only sets the mediaNegotiationDone flag if the current state of the Call State Machine is in S_RECV_PROGRESS. 
3. sendDigit waits until the call is in the CONNECTED state, but it doesn't need to do this. DTMF can be sent on a media stream despite the state of the call. 

1. Updated the S_SEND_SETUP state to transition to S_CALL_CONNECT when E_RECV_CONNECT is received
2. Add the S_SEND_SETUP state as a source state in the handleRoapEstablished
3. Remove the condition for connected call and instead rely on mediaConnection.insertDTMF() to throw an error if media is not established.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Created a UT to replicate what's observed in the field and added checks at various intervals to ensure that the state changes are now correct.

Here's the failure of the older code with the new UT I'd written.
1. The call doesn't get set to CONNECTED, which is wrong
2. The call state machine is still at CALL_CONNECT state instead of CALL_ESTABLISHED, which is again incorrect

```
● State Machine handler tests › outgoing call with connect after setup, media established before connect 

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      916 |     call.sendCallStateMachineEvt(dummyEvent as CallEvent);
      917 |     expect(call['callStateMachine'].state.value).toBe('S_RECV_CALL_CONNECT');
    > 918 |     expect(call.isConnected()).toBe(true);
          |                                ^
      919 |
      920 |     call.sendCallStateMachineEvt({type: 'E_SEND_CALL_DISCONNECT'});
      921 |     expect(call['callStateMachine'].state.value).toBe('S_SEND_CALL_DISCONNECT');

      at Object.<anonymous> (src/CallingClient/calling/call.test.ts:918:32)
      at TestScheduler.scheduleTests (node_modules/@jest/core/build/TestScheduler.js:333:13)
      at runJest (node_modules/@jest/core/build/runJest.js:404:19)
      at _run10000 (node_modules/@jest/core/build/cli/index.js:320:7)
      at runCLI (node_modules/@jest/core/build/cli/index.js:173:3)
```

Apart from the above, tested these scenarios on the kitchen sink page.
1. Incoming and outgoing calls
2. DTMF before and after media connection
3. Hold and resume functions

### I certified that

---

Logs:
[PR-3375-incoming-call-hold-resume.har.txt](https://github.com/webex/webex-js-sdk/files/14275696/PR-3375-incoming-call-hold-resume.har.txt)
[PR-3375-incoming-call-hold-resume.log](https://github.com/webex/webex-js-sdk/files/14275699/PR-3375-incoming-call-hold-resume.log)
[PR-3375-call-tests.log](https://github.com/webex/webex-js-sdk/files/14275700/PR-3375-call-tests.log)
[PR-3375-call-tests.har.txt](https://github.com/webex/webex-js-sdk/files/14275701/PR-3375-call-tests.har.txt)


packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -165,6 +165,47 @@ describe('Call Tests', () => {
...
+      insertDTMF: jest.fn(() => {
+        throw new Error(errorMessage);
+      }),
+    };
+
+    const callManager = getCallManager(webex, defaultServiceIndicator);
+
+    const call = callManager.createCall(dest, CallDirection.OUTBOUND, deviceId, mockLineId);
+
+    const realMediaConnection = call['mediaConnection'];
```
rarajes2, line 183: `call['mediaConnection']` --> call.mediaConnection

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -165,6 +165,47 @@ describe('Call Tests', () => {
...
+
+    // Spy on the log.warn method
+    const logWarnSpy = jest.spyOn(log, 'warn');
+
+    // Call the sendDigit method
+    call.sendDigit(tone);
+
+    // Expect the log.warn method to be called with the error message
+    expect(logWarnSpy).toHaveBeenLastCalledWith(
+      'Unable to send digit on call: Failed to send digit',
```
rarajes2, line 195: Should we use `` `Unable to send digit on call: ${errorMessage}` `` ?

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -854,6 +888,73 @@ describe('State Machine handler tests', () => {
     expect(call['callStateMachine'].state.value).toBe('S_SEND_CALL_DISCONNECT');
   });
 
+  it('outgoing call with connect after setup, media established before connect ', async () => {
```
rarajes2, line 891: Message is not very clear, can we make it more readable?

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -854,6 +888,73 @@ describe('State Machine handler tests', () => {
...
+
+    const postMediaSpy = jest.spyOn(call as any, 'postMedia');
+
+    webex.request.mockReturnValue(statusPayload);
+
+    call.sendCallStateMachineEvt(dummyEvent as CallEvent);
+    expect(call['callStateMachine'].state.value).toBe('S_SEND_CALL_SETUP');
+    // dummyEvent.type = 'E_RECV_CALL_PROGRESS';
+    // call.sendCallStateMachineEvt(dummyEvent as CallEvent);
+    // expect(call['callStateMachine'].state.value).toBe('S_RECV_CALL_PROGRESS');
```
rarajes2, line 909: We can remove commented code

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -854,6 +888,73 @@ describe('State Machine handler tests', () => {
...
+
+    dummyEvent.type = 'E_SEND_ROAP_OFFER';
```
rarajes2, line 911: Creating a separate object would be cleaner, if any change is needed later it may mess up the code

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -854,6 +888,73 @@ describe('State Machine handler tests', () => {
...
+    call.sendMediaStateMachineEvt(dummyEvent as RoapEvent);
+
+    /**
+     * Since the event doesn't have any data above, we should request media sdk for an offer here.
+     * The below event is expected to be called again my mediaSdk.
+     */
+    dummyEvent.data = {
```
rarajes2, line 918: Same here

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: `const call = callManager.createCall(dest, CallDirection.OUTBOUND, deviceId, mockLineId);` returns an object with interface ICall, which doesn't expose mediaConnection as a member. mediaConnection is a private member of the Call object. Therefore, call.mediaConnection does not work. This is the usual method we've used for all our tests before as well.

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: Yes this makes sense, I've made the change

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: I see it. I've made this heading clearer now.
`outgoing call where we receive connect directly after setup. Media established before connect. test call and media state changes`

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: removed

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: I agree. I've done it for this test which I've added now. Other tests can be taken care of in the future.

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: fixed

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: It makes sense if the type is not exposed but if it is a private member it shouldn't be accessible like this as well unless we treat it as a private member just by not exposing the type.

rarajes2 (approved):

karthikraji2020 (approved):
LGTM

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -854,6 +888,90 @@ describe('State Machine handler tests', () => {
...
+    dummyOfferEvent.data = {
+      seq: 1,
+      messageType: 'OFFER',
+      sdp: 'sdp',
+    };
+    call.sendMediaStateMachineEvt(dummyOfferEvent as RoapEvent);
+    expect(mediaConnection.initiateOffer).toHaveBeenCalledTimes(1);
+    expect(postMediaSpy).toHaveBeenLastCalledWith(dummyOfferEvent.data as RoapMessage);
+
+    // dummyAEvent.type = 'E_RECV_ROAP_ANSWER';
```
karthikraji2020, line 955: could you please remove this commented code as well ?

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: What's a better way of doing this then? Btw I tried doing `call.mediaConnection` and despite typescript throwing an error, the test still passes.

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: Removed the commented code. Thanks

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: > Btw I tried doing call.mediaConnection and despite typescript throwing an error, the test still passes.

This should be because we are not exposing the type but it's in reality not a private member. 

If we want to keep them as "private" and add some checks before setting the value, maybe we can create public getter/setter methods on them.

PR #3376, opened 2024-02-12
fix: avoid duplicate request of meeting info by passing it from cantina

maxinteger:
# COMPLETES SPARK-489305

Avoid duplicate request of meeting info

Adding extra parameter to `meeting.create` method to pass down pre-fetched meeting info, also adding utility methods to process the new parameter.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1399,7 +1399,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @param {string} [meetingLookupUrl] Lookup url, defined when the meeting info fetched
    * @returns {void}
    */
-  private setMeetingInfo(info, meetingLookupUrl = undefined) {
+  private setMeetingInfo(info, meetingLookupUrl) {
     this.meetingInfo = info ? {...info, meetingLookupUrl} : null;
     this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NONE;
 
@@ -1432,16 +1432,18 @@ export default class Meeting extends StatelessWebexPlugin {
    *
    * @param {Object} meetingInfo - Complete meeting info
    * @param {FetchMeetingInfoParams} fetchParams - Fetch parameters for validation
+   * @param {String} meetingLookupUrl - Lookup url, defined when the meeting info fetched
    * @returns {Promise<void>}
    */
   public async injectMeetingInfo(
     meetingInfo: any,
-    fetchParams: FetchMeetingInfoParams
+    fetchParams: FetchMeetingInfoParams,
+    meetingLookupUrl: string
   ): Promise<void> {
     await this.prepForFetchMeetingInfo(fetchParams, 'injectMeetingInfo');
 
     this.parseMeetingInfo(meetingInfo, this.destination);
-    this.setMeetingInfo(meetingInfo);
+    this.setMeetingInfo(meetingInfo, meetingLookupUrl);
   }
 
   /**
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1396,10 +1396,13 @@ export default class Meeting extends StatelessWebexPlugin {
   /**
    * Set meeting info and trigger `MEETING_INFO_AVAILABLE` event
    * @param {any} info
-   * @param {string} meetingLookupUrl lookup url
+   * @param {string} [meetingLookupUrl] Lookup url, defined when the meeting info fetched
    * @returns {void}
    */
-  private setMeetingInfo(info, meetingLookupUrl) {
+  private setMeetingInfo(info, meetingLookupUrl = undefined) {
+    this.meetingInfo = info ? {...info, meetingLookupUrl} : null;
+    this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NONE;
+
     this.requiredCaptcha = null;
     if (
       this.passwordStatus === PASSWORD_STATUS.REQUIRED ||
@@ -1409,10 +1412,6 @@ export default class Meeting extends StatelessWebexPlugin {
     } else {
       this.passwordStatus = PASSWORD_STATUS.NOT_REQUIRED;
     }
-
-    this.parseMeetingInfo(info, this.destination);
-    this.meetingInfo = info ? {...info, meetingLookupUrl} : null;
-    this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NONE;
 
     Trigger.trigger(
       this,
@@ -1426,12 +1425,34 @@ export default class Meeting extends StatelessWebexPlugin {
     this.updateMeetingActions();
   }
 
-  public async injectMeetingInfo(info: any, fetchParams: FetchMeetingInfoParams): Promise<void> {
+  /**
+   * Add pre-fetched meeting info
+   *
+   * The passed meeting info should be be complete, e.g.: fetched after password or captcha provided
+   *
+   * @param {Object} meetingInfo - Complete meeting info
+   * @param {FetchMeetingInfoParams} fetchParams - Fetch parameters for validation
+   * @returns {Promise<void>}
+   */
+  public async injectMeetingInfo(
+    meetingInfo: any,
+    fetchParams: FetchMeetingInfoParams
+  ): Promise<void> {
     await this.prepForFetchMeetingInfo(fetchParams, 'injectMeetingInfo');
 
-    this.setMeetingInfo(info, undefined);
+    this.parseMeetingInfo(meetingInfo, this.destination);
+    this.setMeetingInfo(meetingInfo);
   }
 
+  /**
+   * Validate fetch parameters and clear the fetchMeetingInfoTimeout timeout
+   *
+   * @param {FetchMeetingInfoParams} fetchParams - fetch parameters for validation
+   * @param {String} caller - Name of the caller for logging
+   *
+   * @returns {Promise<void>}
+   * @private
+   */
   private prepForFetchMeetingInfo(
     {password = null, captchaCode = null, extraParams = {}}: FetchMeetingInfoParams,
     caller: string
@@ -1491,6 +1512,7 @@ export default class Meeting extends StatelessWebexPlugin {
         {meetingId: this.id, sendCAevents}
       );
 
+      this.parseMeetingInfo(info?.body, this.destination, info?.errors);
       this.setMeetingInfo(info?.body, info?.url);
 
       return Promise.resolve();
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1386,6 +1393,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        file: 'meetings',
+        function: 'fetchMeetingInfo',
+      },
+      EVENT_TRIGGERS.MEETING_INFO_AVAILABLE
+    );
+
+    this.updateMeetingActions();
+  }
+
+  public async injectMeetingInfo(info: any, fetchParams: FetchMeetingInfoParams): Promise<void> {
```
martroos, line 1429: nitpick: The linter complains about this lacking a JSDoc comment.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1386,6 +1393,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    await this.prepForFetchMeetingInfo(fetchParams, 'injectMeetingInfo');
+
+    this.setMeetingInfo(info, undefined);
+  }
+
+  private prepForFetchMeetingInfo(
```
martroos, line 1435: nitpick: The linter complains about this lacking a JSDoc comment.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
martroos, line 1: nitpick: Would be much easier to review this if it didn't contain all the auto-formatter changes (or at least if they were done in a separate commit).

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4621,6 +4786,105 @@ describe('plugin-meetings', () => {
...
+          conversationUrl: 'some_convo_url',
+          locusUrl: 'some_locus_url',
+          sipUrl: 'some_sip_url', // or sipMeetingUri
+          meetingNumber: '123456', // this.config.experimental.enableUnifiedMeetings
+          hostId: 'some_host_id', // this.owner;
+        };
+        const FAKE_MEETING_INFO_LOOKUP_URL = 'meetingLookupUrl';
+
+        it('calls meetingInfoProvider with all the right parameters and parses the result', async () => {
+          meeting.requiredCaptcha = FAKE_SDK_CAPTCHA_INFO;
```
martroos, line 4811: suggestion: This is a little confusing, since for the inject case, a captcha is never needed - it is initialised as `null` and is only set to something non-`null` when doing an actual fetch. I realise the reason for this is because you also provide a captcha code, but again I don't think this should be provided generally for inject info. I think it may be simpler to just not set / provide anything captcha (or password) related for this first test.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4621,6 +4786,105 @@ describe('plugin-meetings', () => {
...
+            meeting,
+            {file: 'meetings', function: 'fetchMeetingInfo'},
+            'meeting:meetingInfoAvailable'
+          );
+          assert.calledWith(meeting.updateMeetingActions);
+        });
+
+        it('fails if captchaCode is provided when captcha not needed', async () => {
+          meeting.attrs.meetingInfoProvider = {
+            injectMeetingInfo: sinon.stub().resolves(),
```
martroos, line 4845: question: Why is this necessary? `injectMeetingInfo` isn't a property of `meetingInfoProvider` (so of course it doesn't get called). Should this actually be `fetchMeetingInfo`?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4621,6 +4786,105 @@ describe('plugin-meetings', () => {
...
+            Error,
+            'injectMeetingInfo() called with captchaCode when captcha was not required'
+          );
+
+          assert.notCalled(meeting.attrs.meetingInfoProvider.injectMeetingInfo);
+        });
+
+        it('fails if password is provided when not required', async () => {
+          meeting.attrs.meetingInfoProvider = {
+            injectMeetingInfo: sinon.stub().resolves(),
```
martroos, line 4867: question: Same question as before regarding if this should actually be `fetchMeetingInfo`

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4621,6 +4786,105 @@ describe('plugin-meetings', () => {
         });
       });
 
+      describe('#injectMeetingInfo', () => {
```
martroos, line 4789: suggestion: Theoretically there could be a delayed fetch pending and `injectMeetingInfo` could be called during this time, so it might be worth checking that `fetchMeetingInfoTimeoutId` gets cleared when this is the case.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4621,6 +4786,105 @@ describe('plugin-meetings', () => {
...
+          assert.calledWith(
+            TriggerProxy.trigger,
+            meeting,
+            {file: 'meetings', function: 'fetchMeetingInfo'},
+            'meeting:meetingInfoAvailable'
+          );
+          assert.calledWith(meeting.updateMeetingActions);
+        });
+
+        it('fails if captchaCode is provided when captcha not needed', async () => {
```
martroos, line 4843: suggestion: Would be nice to check that the `meetingInfoAvailable` event isn't triggered in this (and the password) case.

martroos (changes requested):
Requesting changes mostly to have questions answered before approving. I'm a little concerned that this will only work with complete meeting info (not partial info as you might get from a fetch when a password / captcha is needed but not provided), but documentation should suffice for now (unless there is a reliable way of checking complete info is provided).

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1386,6 +1393,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      EVENT_TRIGGERS.MEETING_INFO_AVAILABLE
+    );
+
+    this.updateMeetingActions();
+  }
+
+  public async injectMeetingInfo(info: any, fetchParams: FetchMeetingInfoParams): Promise<void> {
+    await this.prepForFetchMeetingInfo(fetchParams, 'injectMeetingInfo');
+
+    this.setMeetingInfo(info, undefined);
```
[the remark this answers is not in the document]
maxinteger: I updated the JSDoc and set the default value for the `meetingLookupUrl` parameter to `undefined`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3214,7 +3234,7 @@ export default class Meeting extends StatelessWebexPlugin {
       | any,
     destination: object | string | null = null
   ) {
-    const webexMeetingInfo = meetingInfo?.body;
+    const webexMeetingInfo = meetingInfo?.body || meetingInfo;
```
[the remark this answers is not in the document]
maxinteger: We needed the `meetingInfo.body` format because later it uses `meetingInfo.error`

I added an error parameter and moved `parseMeetingInfo` out from `setMeetingInfo` to make it more clear

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
maxinteger: Sorry about that. I was lazy to undo my commit. Next time I will separate them

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1386,6 +1393,74 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  private setMeetingInfo(info, meetingLookupUrl) {
+    this.requiredCaptcha = null;
+    if (
+      this.passwordStatus === PASSWORD_STATUS.REQUIRED ||
+      this.passwordStatus === PASSWORD_STATUS.VERIFIED
+    ) {
+      this.passwordStatus = PASSWORD_STATUS.VERIFIED;
+    } else {
+      this.passwordStatus = PASSWORD_STATUS.NOT_REQUIRED;
+    }
```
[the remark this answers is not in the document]
maxinteger: I moved this and the `parseMeetingInfo` independently and I forget to restore the original order.

I moved out `parseMeetingInfo` from the `setMeetingInfo` and no the order is the same as in the original code.

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
martroos: My point wasn't that we shouldn't be providing `undefined` as an argument, but that we should allow for the lookup URL to be provided when calling `injectMeetingInfo`, so that externally we can state where the information came from when injecting it. For example, in guest join cases, it will come from a preJoin API. Knowing this helps us debug issues with meeting info. So my suggestion was effectively that we should provide an extra argument / add to options to provide this here and in the create method (where this is used). i.e.

```js
public async injectMeetingInfo(info: any, fetchParams: FetchMeetingInfoParams, meetingLookupUrl: string) {
```

Or something like that.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
martroos: The checks for the event triggers appear to be commented out - did they not work or was this just work in progress?

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
maxinteger: I see you point now. 

I refactored it, so meetingLookUp url can passed from cantina if it is available

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
maxinteger: Good catch

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1432,16 +1432,18 @@ export default class Meeting extends StatelessWebexPlugin {
...
    * @param {Object} meetingInfo - Complete meeting info
    * @param {FetchMeetingInfoParams} fetchParams - Fetch parameters for validation
+   * @param {String} meetingLookupUrl - Lookup url, defined when the meeting info fetched
    * @returns {Promise<void>}
    */
   public async injectMeetingInfo(
     meetingInfo: any,
-    fetchParams: FetchMeetingInfoParams
+    fetchParams: FetchMeetingInfoParams,
+    meetingLookupUrl: string
```
martroos, line 1441: suggestion: This additional parameter needs unit test(s) adding.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1432,16 +1432,18 @@ export default class Meeting extends StatelessWebexPlugin {
    *
    * @param {Object} meetingInfo - Complete meeting info
    * @param {FetchMeetingInfoParams} fetchParams - Fetch parameters for validation
+   * @param {String} meetingLookupUrl - Lookup url, defined when the meeting info fetched
```
martroos, line 1435: nitpick: This still may be undefined

martroos (approved):

PR #3377, opened 2024-02-12
Spark 486000

peter7cole:
# COMPLETES [SPARK-486000](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-486000)

Adding `installationId` and testing to the device section of the join request

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

### Screenshots
With installationId from webex web client
Screenshot 2024-02-29 at 1 04 32 PM

Without installationId from webex web client
Screenshot 2024-02-29 at 1 04 50 PM


packages/@webex/internal-plugin-device/src/device.js
```diff
@@ -255,6 +255,29 @@ const Device = WebexPlugin.extend({
...
    */
   derived: {
+    /**
+     * This property assigns a unique machineId per session to assist with
+     * fraud prevention.
+     *
+     * @type {string}
+     */
+    machineId: {
+      deps: ['url'],
```
chrisadubois, line 265: I dont think this is right. This was discussed and the intention was to grab to the machineID from another value we already use like amplitude user ID or ci UUID

Have discussed this with the team and the design approach was set as follows and was added to the JIRA ticket on 11,Jan,2024

1. Define a new constant, `machineId` in constants/Storage.ts for the machineId, and store it as a functional type (the reason being detecting fraud is for functional, but GDPR forces us to allow consumers to turn off tracking of any sort), but it's certainly better than performance storage. The name is knowingly not that great, but sets us up for it's purpose and can thus be changed later on (implementation wise) without us having to change the variable name
2. Define a new key at app startup time (before login or anything) that gets stored in the `machineId` storage location. For now, the best one that we could use would be the same value as the current machineId (from localStorage) OR `amplitude-id` (if it exists) or the `teams-anonymous-id` and persist this value in localStorage
3. Then, we would need to pass this id into the webex SDK config for it to be used later on by other SDK plugins if defined
4. Then, in packages/@webex/plugin-meetings/src/meeting/request.ts in the SDK, we will need to add a new field to the body.device when we do a locus join. That field would be called `machineId`. machineId should only be defined if webex.internal.device.getMachineId() is defined.
5. getMachineId() should (for now) always return the value of webex.internal.device.config.machineId. You will have to set webex.internal.device.config.machineId dynamically after the app has started up. If it is undefined, it should remain undefined, and thus not be appended to the join request body. In the future, it is possible that getMachineId() might return a device fingerprint, but this is unlikely given privacy laws.
6. manual test / write unit tests
7. PR review, deploy SDK, bump SDK in cantina, with cantina changes, and deploy cantina.

chrisadubois (changes requested):

chrisadubois (approved):

PR #3384, opened 2024-02-14
fix(meetings): add keepOnlyLocusMeetings prop to syncMeetings

nataliadelmar:
# COMPLETES #[SPARK-484517](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484517)

When mercury disconnect and reconnect, meetings' syncMeetings method is called to sync our meetings collection with the server. In this sync, only locus meetings with an active locus url where synced, all other meetings were destroyed. That would cause some issues as sometimes we want to keep meetings that are not locus meetings (have no locus url) too as they are meetings created by the client. However, we do not want this to be the default behaviour of syncMeetings method as that would break other instances of this function.

Add keepOnlyLocusMeetings as a prop to syncMeetings that defaults to true, to keep the previous behaviour as default. Only if this prop is false, like when mercury reconnection calls it, then meetings that are not from locus will not be destroyed in the sync.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


arun3528 (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1374,7 +1375,12 @@ export default class Meetings extends WebexPlugin {
           // (they had a locusUrl previously but are no longer active) in the sync
           for (const meeting of Object.values(meetingsCollection)) {
             // @ts-ignore
-            if (meeting.locusUrl && !activeLocusUrl.includes(meeting.locusUrl)) {
+            const isNotActiveLocusMeeting = !activeLocusUrl.includes(meeting.locusUrl);
+            if (
+              (keepOnlyLocusMeetings && isNotActiveLocusMeeting) ||
+              // @ts-ignore
+              (!keepOnlyLocusMeetings && meeting.locusUrl && isNotActiveLocusMeeting)
+            ) {
```
martroos, line 1383: suggestion: This can be simplified:

```ts
if ((keepOnlyLocusMeetings || meeting.locusUrl) && !activeLocusUrl.includes(meeting.locusUrl)) {
```

Which works on the assumption that the active URL array never contains `undefined` as an entry (which it shouldn't!). This is also marginally more efficient, since if we are keeping non-locus meetings, it doesn't even bother checking if it is an active URL or not (the meeting should be kept since it lacks a locusUrl).

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -593,11 +593,24 @@ describe('plugin-meetings', () => {
               );
               MeetingUtil.cleanUp = sinon.stub().returns(Promise.resolve());
             });
-            it('destroy only non active locus meetings and keep active locus meetings and any other non-locus meeting', async () => {
+
+            it('destroy any meeting that has no active locus url if keepOnlyLocusMeetings', async () => {
```
martroos, line 597: nitpick: For completeness, a version of this test where `true` is explicitly passed in would be nice, but not essential!

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1346,11 +1346,12 @@ export default class Meetings extends WebexPlugin {
 
   /**
    * syncs all the meeting from server
+   * @param {boolean} keepOnlyLocusMeetings - whether the sync should keep only locus meetings or any other meeting in meetingCollection
    * @returns {undefined}
    * @public
    * @memberof Meetings
    */
-  public syncMeetings() {
+  public syncMeetings(keepOnlyLocusMeetings = true): undefined {
```
martroos, line 1354: nitpick: The name of the parameter might be clearer as `removeNonLocusMeetings`, but not too worried if you think `keepOnlyLocusMeetings` is clearer.

martroos (changes requested):
Functionally this change looks good, though (as usual) I have made some suggestions, and the type-ing of the sync function needs to be corrected.

packages/@webex/plugin-meetings/src/meetings/index.ts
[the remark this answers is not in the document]
nataliadelmar: Good idea; done!

nataliadelmar: sdk team is not so happy with the term 'non-locus meetings' which is why I prefer thisname

packages/@webex/plugin-meetings/src/meetings/index.ts
nataliadelmar: I like this better! Thank you!

martroos (approved):

PR #3388, opened 2024-02-16
fix(plugin-meetings): update MQE data related to hop-by-hop loss

brycetham:
# COMPLETES [SPARK-498304](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-498304)

Updating MQE stats required to calculate rxHbhLostPercent and txHbhLostPercent in MQR.

From a receiver stats perspective, the following MQE metrics were updated:

- fecPackets (audio only)
- rtpPackets
- rtpHopByHopLost

From a sender stats perspective, the following MQE metrics were updated:

- remoteLossRate 

I also talked with @mccarthytyler about making it easier to test MQE data, and I think we have a pretty good framework now for testing successive MQEs in the unit tests. The existing test for framerate was incorrect, so I've fixed it, as well as added my own tests for the MQE metrics listed above.

Tests are still missing for most MQE stats but those can be added in the future.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Unit tests to verify MQE data is correct.

### I certified that

---


packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1235,7 +1235,7 @@ export const AVAILABLE_RESOLUTIONS = {
  *  mqa Interval for sending stats metrics
  */
 
-export const MQA_INTEVAL = 60000; // mqa analyzer interval its fixed to 60000
+export const MQA_INTERVAL = 60000; // mqa analyzer interval its fixed to 60000
```
mccarthytyler, line 1238: Thanks for this!

mccarthytyler (approved):
Looks good to me!

packages/@webex/plugin-meetings/src/statsAnalyzer/mqaUtil.ts
```diff
@@ -156,16 +158,15 @@ export const getVideoReceiverMqa = ({videoReceiver, statsResults, lastMqaDataSen
...
 
-  const totalPacketLoss =
-    statsResults[mediaType][sendrecvType].totalPacketsLost - lastPacketsLost || 0;
-
-  // Hope by hop are numbers and not percentage so we compare on what we sent the last min
+  // Hop by hop are numbers and not percentage so we compare on what we sent the last min
   // this is including packet lost
-  videoReceiver.common.mediaHopByHopLost = totalPacketLoss;
-  videoReceiver.common.rtpHopByHopLost = totalPacketLoss;
+  const totalPacketsLost =
```
chrisadubois, line 163: Could we just make all this totalPacketsLost stuff a function and hand it a receiver as well as setting the common.mediaHopByHopLost ? it seems like there's some redundancy between this and the audio stuff... obviously not required, but it seems like an easy thing to do and then have a separate util function and a smaller test for it

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
```diff
@@ -478,14 +504,118 @@ describe('plugin-meetings', () => {
         );
       });
 
-      it('emits the correct frameRate', async () => {
-        await startStatsAnalyzer({expected: {receiveVideo: true}});
+      it('emits the correct transmittedFrameRate/receivedFrameRate', async () => {
```
chrisadubois, line 507: each of these tests are absolutely ginormous - could we break it up into smaller component parts and nest the its? does that make sense?

chrisadubois (approved):
Approving with the following comments - nothing is functionally blocking

1. there *should* be a test for mqaUtil for the new bits added which is covered by extension in stats-analyzer, but if you add that function I mentioned
2. Repetition in code in mqaUtil
3. some of the tests are really hard to grok.... might be easier just if they were split up

packages/@webex/plugin-meetings/src/statsAnalyzer/mqaUtil.ts
brycetham: There's probably a lot more than just this where there's redundancy between `audioReceiver` and `videoReceiver` (and even more between `audioSender` and `videoSender`). I agree that it would be great to have this logic in a common location/function shared across media types... though if we do this, we should probably do it all in one go (i.e. instead of having separate `getAudioReceiverMqa` and `getVideoReceiverMqa` functions, have a single `getReceiverMqa` function that handles both). Honestly, that's probably a pretty big change that's out-of-scope for this PR... I'd suggest doing it as a separate task.

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
brycetham: Yep, will see what I can do to better organize them.

PR #3390, opened 2024-02-21
Feat/Simultaneous Interpretation - send/receive named media group request

mickelr:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-443306
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-443837

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -36,13 +37,20 @@ export class RemoteMediaGroup {
...
       (slot) =>
         new RemoteMedia(slot, this.mediaRequestManager, {
           resolution: this.options.resolution,
         })
     );
+    this.namedMedia = this.unpinnedRemoteMedia.filter(
+      (item) => item !== null && !!item.namedMediaGroup && !!item.namedMediaGroup.value
+    );
+    this.unpinnedRemoteMedia = this.unpinnedRemoteMedia.filter(
+      (item) => item !== null && (!item.namedMediaGroup || !item.namedMediaGroup.value)
```
marcin-bazyl, line 52: I don't think `this.unpinnedRemoteMedia` can ever have any null or undefined values, can it? if it can't then this code can be simplified a bit

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,6 +533,42 @@ export class RemoteMediaManager extends EventsScope {
...
+    };
+    if (mediaType === MediaType.AudioMain) {
+      this.invalidateCurrentRemoteMedia({
+        audio: true,
+        video: false,
+        screenShareAudio: false,
+        screenShareVideo: false,
+        commit: false,
+      });
+      this.slots.audio.reverse();
```
marcin-bazyl, line 564: might be worth adding a comment why we're reversing the array here. I guess we're doing it so that we release the slots in reverse order and maintain the same slots in same order after the call to createAudioMedia() is done as they were before setReceiveNamedMediaGroup() was called, right?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5807,6 +5807,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
     // publish the streams
     if (this.mediaProperties.audioStream) {
+      this.setSendNamedMediaGroup(MediaType.AudioMain);
```
marcin-bazyl, line 5810: what about the case when the target language code is changed? In that case we also need to call setSendNamedMediaGroup(), right? but I don't see this in this PR

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -22,9 +22,10 @@ export class RemoteMediaGroup {
 
   private unpinnedRemoteMedia: RemoteMedia[];
 
-  private mediaRequestId?: MediaRequestId; // id of the "active-speaker" media request id
+  private mediaRequestIds: MediaRequestId[]; // id of the "active-speaker" media request id
 
   private pinnedRemoteMedia: RemoteMedia[];
+  private namedMedia: RemoteMedia[];
```
marcin-bazyl, line 28: hmm... I don't know... it doesn't feel great having unpinned, pinned and named media arrays all together in this class. The relationship between them is not clear. I see you're splitting the `unpinned` into `named` and `unpinned` in the constructor, but what if someone calls setNamedMediaGroup() on a RemoteMedia object that's in unpinned array later? or what if someone calls `RemoteMediaGroup.pin()` on a named one? This can easily get very messy. Currently named ones are only applicable to audio, right? and pinning/unpinning only really makes sense for video, so maybe we need to create a class hierarchy with AudioRemoteMediaGroup that will support named media groups but no pinning and VideoRemoteMediaGroup that has pinning but no named media groups and both classes would extend RemoteMediaGroup which would be an abstract class now. Unless we can have named media groups for video too? then we would need to do it differently, if that's the case, let's have a call to discuss this.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -168,6 +172,10 @@ export const DefaultConfiguration: Configuration = {
       ScreenShareView: RemoteScreenShareWithSmallThumbnailsLayout,
     },
   },
+  receiveNamedMediaGroup: {
```
marcin-bazyl, line 175: would it be better to have it `undefined` here?

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -49,6 +49,10 @@ export interface Configuration {
 
     layouts: {[key: LayoutId]: VideoLayout}; // a map of all available layouts, a layout can be set via setLayout() method
   };
+  receiveNamedMediaGroup?: {
```
marcin-bazyl, line 52: this is RemoteMediaManager class so it only deals with receiving of remote media, not sending, so the "receive" bit in the name is redundant. Maybe instead we should have "audioNamedMediaGroup", because it only applies to audio?

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -533,9 +577,17 @@ export class RemoteMediaManager extends EventsScope {
...
       this.slots.audio.push(slot);
     }
+    // create slot for interpretation language audio
+    if (this.config.receiveNamedMediaGroup.value) {
+      const siSlot = await this.receiveSlotManager.allocateSlot(
+        MediaType.AudioMain,
+        this.config.receiveNamedMediaGroup
+      );
+
+      this.slots.audio.push(siSlot);
```
marcin-bazyl, line 589: I'm wondering if maybe we should have a separate `this.slots.siAudio` instead of appending it to the end of `this.slots.audio` array. It doesn't make much difference for now, but feels like it would be cleaner.

packages/@webex/plugin-meetings/src/meeting/index.ts
mickelr: Cantina will call setSendNamedMediaGroup() when changed the target language in UI. I will submit another cantina PR for this later.

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -406,15 +409,17 @@ export class MediaRequestManager {
     return newId;
   }
 
-  public cancelRequest(requestId: MediaRequestId, commit = true) {
-    const mediaRequest = this.clientRequests[requestId];
+  public cancelRequest(requestIds: MediaRequestId[], commit = true) {
```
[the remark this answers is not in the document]
mickelr: Sure.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Right, homer team suggests that we keep same slots for main audio request to get better performance. However I don't think we can always make sure that, e.g. if in future there are more than one named media group request in one media request

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: It's ok to me to remove "receive".

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
mickelr: According to the definition of the namedMediaGroup, I think we cannot suppose it's only for audio.  I use the **named** array just for simplifing the logic when creating meida request. I can remove the array for your concern.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Do you mean that you want to add a new parameter in the constructor of the RemoteMeidaGroup class?

packages/@webex/plugin-meetings/src/multistream/receiveSlotManager.ts
```diff
@@ -50,14 +51,21 @@ export class ReceiveSlotManager {
    * Creates a new receive slot or returns one from the existing pool of free slots
    *
    * @param {MediaType} mediaType
+   * @param isNamedMeida {boolean}
    * @returns {Promise<ReceiveSlot>}
    */
-  async allocateSlot(mediaType: MediaType): Promise<ReceiveSlot> {
+  async allocateSlot(
+    mediaType: MediaType,
+    namedMediaGroup: NamedMediaGroup = undefined
```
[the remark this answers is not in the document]
mickelr: Agree with you that we don't have to introduce named media group into receiveSlotManager

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: > Would it be more meanful in the Configuration definition with "receive" prefix, unless the config is only for recieving?
yes, the config is only for receiving, because the whole RemoteMediaManager class is only for remote media, so only for receiving

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: no, just having a separate siAudio here in remoteMediaManager, so `this.slots.siAudio = siSlot;`

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: BTW, why "value" is a number if it's supposed to describe the media type? this is very confusing, why isn't it named `mediaType` and isn't an enum?

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -213,7 +214,30 @@ export class RemoteMediaGroup {
   private sendActiveSpeakerMediaRequest(commit: boolean) {
     this.cancelActiveSpeakerMediaRequest(false);
 
-    this.mediaRequestId = this.mediaRequestManager.addRequest(
+    const namedMedia = this.unpinnedRemoteMedia.filter((item) => item?.namedMediaGroup?.value);
```
marcin-bazyl, line 217: what about `this.pinnedRemoteMedia`? in theory the app can call pin/unpin on any remote media including the named ones. The pinned ones are not part of the active speaker request, but each of them sends their own receive-selected request see RemoteMedia.sendMediaRequest(). Do we need to add namedMediaGroups to the policyInfo there?

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Sorry, should be the "type" member, which use the Homer side definition : ( 0 – RESERVED, 1 – Audio Language, 2 – Inter-meeting media exchange, 3 – Media origin, 4 – 255 – RESERVED)

mickelr: So I will remove the "receive" prefix

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
mickelr: According to the JMP design: https://confluence-eng-gpk2.cisco.com/conf/display/WMT/JMP+-+Json+Multistream+Protocol , the namedMediaGroups is only defined in "active-speaker" policy.

For "active-speaker" policy,:
- priority (required): the loudest active speakers are mapped to the highest priority requests.  Minimum value is 0, maximum value is 255.  A given active speaker priority MUST NOT be present in multiple different MediaRequests.  Different MediaRequests using active speaker policy MUST have different priorities.
- crossPriorityDuplication (required): when true, the active speakers mapped to a higher priority request are also mapped to lower priority requests.  This parameter MUST have the same value for all requests with "active-speaker" policy.
- crossPolicyDuplication (required): when true, the active speakers mapped to an active-speaker request MAY also be mapped to non-active-speaker requests.
- preferLiveVideo (required): when true, active speakers with live video (e.g. are not audio only or video muted) will be preferred over active speakers without live video.
- namedMediaGroups (optional) = for audio only.  Indicates which named media groups the receiver wants.

marcin-bazyl: hmm.... ok, but we need to decide what should happen if the app calls pin() on a remote media object that has a named media group. Maybe we should just not allow that to happen and fail the call to pin()?

mickelr: Right, will throw an error if app calls pin() for named media group

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,6 +529,42 @@ export class RemoteMediaManager extends EventsScope {
...
+      value === this.config.namedMediaGroup?.value
+    ) {
+      return;
+    }
+
+    this.config.namedMediaGroup = {
+      type,
+      value,
+    };
+    if (mediaType === MediaType.AudioMain) {
```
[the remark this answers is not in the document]
mickelr: Agree.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -533,9 +575,16 @@ export class RemoteMediaManager extends EventsScope {
...
       const slot = await this.receiveSlotManager.allocateSlot(MediaType.AudioMain);
-
       this.slots.audio.push(slot);
     }
+    // create slot for interpretation language audio
+    if (this.config.namedMediaGroup?.value) {
+      const siSlot = await this.receiveSlotManager.allocateSlot(MediaType.AudioMain);
+      if (siSlot) {
+        siSlot.setNamedMediaGroup(this.config.namedMediaGroup);
+        this.slots.audio.push(siSlot);
```
marcin-bazyl, line 585: I think I mentioned it in the past, but maybe it got lost among the comments: I feel it would be cleaner to store this audio si slot separately instead of appending it to the main audio, so like this:
`this.slots.siAudio = siSlot;`

and actually this makes me realize that then we would have to create a separate RemoteMediaGroup just for the si audio, but it means that we wouldn't need most of the changes you did in RemoteMediaGroup and ReceiveSlot. Basically. we would just add namedMediaGroup as a parameter to RemoteMediaGroup constructor and would construct that simultaneous interpretation remote media group here. We would also need to emit a new type of event to indicate that a new media group for SI is created. What do you think, would that be a better approach?

There is actually a problem with the current implementation that when the app calls setReceiveNamedMediaGroup(), the existing RemoteMediaGroup for main audio gets destroyed and app doesn't know about it and it only gets an event `Event.AudioCreated` out of nowhere with the new group that includes SI audio. If we do the new approach I'm proposing above, we wouldn't need to destroy the main audio group and would only need to emit some new event like "si audio updated" or "si audio destroyed" and "si audio created".

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -6012,6 +6013,34 @@ describe('plugin-meetings', () => {
       });
     });
 
+    describe('#setSendNamedMediaGroup', () => {
+      beforeEach(() => {
+        meeting.sendSlotManager.setNamedMediaGroups = sinon.stub().returns(undefined);
+      });
+      it('should throw error if not audio type', () => {
+        // try {
```
marcin-bazyl, line 6021: commented out code - if not needed, remove it, if needed, uncomment it :-)

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,6 +529,44 @@ export class RemoteMediaManager extends EventsScope {
...
 
+  /**
+   * Sets which named media group need receiving
+   *
+   *
+   *
+   */
+  public async setReceiveNamedMediaGroup(mediaType: MediaType, languageId: number) {
+    if (mediaType !== MediaType.AudioMain) {
+      throw new Error(`can not set receive named media group which media type is ${mediaType}`);
```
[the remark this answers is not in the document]
mickelr: fixed

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -121,6 +122,12 @@ export class RemoteMediaGroup {
    *
    */
   public pin(remoteMedia: RemoteMedia, csi?: CSI): void {
+    if (remoteMedia.namedMediaGroup?.value) {
```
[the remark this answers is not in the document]
mickelr: Done

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,6 +529,44 @@ export class RemoteMediaManager extends EventsScope {
...
+    }
+    const type = mediaType === MediaType.AudioMain ? 1 : 0;
```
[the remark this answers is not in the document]
mickelr: Code changed

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mickelr: Removed

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7915,6 +7919,34 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * @param {number} languageCode of the stream
+   * @returns {void}
+   */
+  public setSendNamedMediaGroup(mediaType: MediaType, languageCode = 0): void {
+    if (mediaType !== MediaType.AudioMain) {
+      throw new Error(`can not set send named media group which media type is ${mediaType}`);
+    }
+
+    const value = languageCode || this.simultaneousInterpretation.getTargetLanguageCode();
+    const type = mediaType === MediaType.AudioMain ? 1 : 0;
```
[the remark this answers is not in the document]
mickelr: Code changed

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: According to the design, client need send a complete request (include main and named media if have) out when changing media request. That's why we have to cancel previous request and create a new one when add/change named media group request. We can create a separate RemoteMediaGroup anyway and always keep the main audio RemoteMediaGroup. But will it make logic simpler to destory old one and create a all new  RemoteMediaGroup, instead of maintaining two RemoteMediaGroup  or an array of RemoteMediaGroups in remoteMediaManager?
And it's fine to add a si audio slot IMO.

PR #3394, opened 2024-02-22
fix(plugin-meetings): only log no packets if slot has live state

jor-row:
# COMPLETES [SPARK-498007](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-498007)

We were logging no packets being received even in cases where we did not expect to receive any (e.g. receive slot has "avatar" state because the user isn't sending any video)

Checking the slots state before logging this info

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -938,7 +939,7 @@ export class StatsAnalyzer extends EventsScope {
       this.statsResults[mediaType][sendrecvType].totalPacketsReceived = result.packetsReceived;
 
       if (currentPacketsReceived === 0) {
-        if (receiveSlot) {
+        if (receiveSlot && sourceState === 'live') {
```
marcin-bazyl, line 942: while you're changing the code here, can you also fix the log from the next line? it shows the `currentPacketsReceived` value at the end, but that value is always going to be zero, it's more useful to show `result.packetsReceived`, because then it will tell us if there were any packets ever received on that slot or not (we will see either zero or some integer).

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
```diff
@@ -761,6 +763,47 @@ describe('plugin-meetings', () => {
...
+          csi: 2,
+          id: "4",
+        };
+
+        await startStatsAnalyzer();
+
+        // don't increase the packets when time progresses.
+        await progressTime();
+
+        assert.calledWith(loggerSpy, 'StatsAnalyzer:index#processInboundRTPResult --> No packets received for receive slot id: "4" and csi: 2');
```
marcin-bazyl, line 779: shouldn't there be 1 more parameter passed to loggerSpy? (the `currentPacketsReceived` that I made a comment about above)

jor-row: Yes, I've added it for the tests I've added now

marcin-bazyl (approved):

maxinteger (approved):

PR #3395, opened 2024-02-22
fix(calling): added new janus event in Call History module

Kesari3008:

For Call on Webex(Locus call) option, Janus triggers a different event(janus.user_sessions) which is not being listened for in Calling SDK hence no event is triggered for this type of call from the SDK. Salesforce need an event for Locus call as well from Calling SDK.

Added listening for janus.user_sessions event in CallHistory module and emitting callHistory:user_reecent_sessions for locus call as well.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Testcase: Triggered Call On Webex option and verified in the logs event being received.

### I certified that

---

### DIFF
--- a/packages/calling/src/CallHistory/CallHistory.ts
+++ b/packages/calling/src/CallHistory/CallHistory.ts
@@ -139,6 +139,14 @@ export class CallHistory extends Eventing<CallHistoryEventTypes> implements ICal
         }
       }
     );
+    this.sdkConnector.registerListener<CallSessionEvent>(
+      MOBIUS_EVENT_KEYS.LOCUS_CALL_SESSION_EVENT,
+      async (event?: CallSessionEvent) => {
+        if (event && event.data.userSessions.userSessions) {
+          this.emit(COMMON_EVENT_KEYS.CALL_HISTORY_USER_SESSION_INFO, event as CallSessionEvent);
+        }
+      }
+    );
   }
 }
 /**
--- a/packages/calling/src/Events/types.ts
+++ b/packages/calling/src/Events/types.ts
@@ -155,6 +155,7 @@ export type Item = {
 export enum MOBIUS_EVENT_KEYS {
   SERVER_EVENT_INCLUSIVE = 'event:mobius',
   CALL_SESSION_EVENT_INCLUSIVE = 'event:janus.user_recent_sessions',
+  LOCUS_CALL_SESSION_EVENT = 'event:janus.user_sessions',
 }
 
 export type CallSessionData = {


packages/calling/src/CallHistory/callHistoryFixtures.ts
```diff
@@ -337,3 +337,16 @@ export const MOCK_SESSION_EVENT: CallSessionEvent = {
   timestamp: 12345,
   trackingId: 'tracking-id',
 };
+
+export const MOCK_LOCUS_SESSION_EVENT: CallSessionEvent = {
+  id: 'id',
+  data: {
+    userSessions: {
+      userSessions: [WEBEX_CALL_SESSION],
```
sreenara, line 345: could we use a different object that matches the legacy session type?

packages/calling/src/CallHistory/CallHistory.ts
```diff
@@ -139,6 +139,14 @@ export class CallHistory extends Eventing<CallHistoryEventTypes> implements ICal
         }
       }
     );
+    this.sdkConnector.registerListener<CallSessionEvent>(
+      MOBIUS_EVENT_KEYS.LOCUS_CALL_SESSION_EVENT,
+      async (event?: CallSessionEvent) => {
```
sreenara, line 144: Seems like this is just a duplicate of the code on lines 136-139? Could we optimize?

sreenara (changes requested):
1. Can we reword the title a bit? I believe this is the legacy event that we are adding right?
2. Could you attach a console log with the event emitted for legacy event?
3. Could you add a screenshot from the kitchen sink page with the new event showing up in the history table?

packages/calling/src/Events/types.ts
```diff
@@ -155,14 +155,17 @@ export type Item = {
 export enum MOBIUS_EVENT_KEYS {
   SERVER_EVENT_INCLUSIVE = 'event:mobius',
   CALL_SESSION_EVENT_INCLUSIVE = 'event:janus.user_recent_sessions',
+  LOCUS_CALL_SESSION_EVENT = 'event:janus.user_sessions',
```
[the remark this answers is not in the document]
Kesari3008: done

packages/calling/src/CallHistory/callHistoryFixtures.ts
Kesari3008: Yeah I am yet to fetch the object from the logs, will update

packages/calling/src/CallHistory/CallHistory.ts
```diff
@@ -139,6 +139,14 @@ export class CallHistory extends Eventing<CallHistoryEventTypes> implements ICal
...
+        if (event && event.data.userSessions.userSessions) {
```
Shreyas281299, line 145: I think this will given an error in some edge case, if data.userSessions is undefined

Shreyas281299, line 145:
```suggestion
        if (event && event.data.userSessions?.userSessions) {
```

Shreyas281299 (approved):
Just 1 small change. Otherwise looks good to me.

packages/calling/src/CallHistory/CallHistory.ts
Shreyas281299:
```suggestion
        if (event && event.data.userSessions && event.data.userSessions.userSessions) {
```

Shreyas281299:
```suggestion
        if (event && event.data.userSessions && event.data.userSessions.userSessions) {
```

packages/calling/src/CallHistory/CallHistory.ts
Kesari3008: done

packages/calling/src/CallHistory/CallHistory.ts
Kesari3008: I believe data.userSessions will not be undefined because it will always return an array which can be empty but the object always has some value. Hence we don't need to add a check for it.

packages/calling/src/CallHistory/callHistoryFixtures.ts
```diff
@@ -325,6 +325,46 @@ const WEBEX_CALL_SESSION = {
   correlationIds: ['58ea6cd9-852b-4a77-957f-e704c8b0e63e'],
 };
 
+const SPARK_CALL_SESSION = {
+  id: '2ce4b9e0-7ee7-12a1-4a94-df524443b520',
+  durationSecs: 0,
+  self: {
+    id: '0fea4a63-4e27-46ee-99c3-2472cb12bf68',
+    name: 'Priya Kesari',
```
sreenara, line 333: Please change these names and put in test names. Applies in other parts of the fixture as well

sreenara (approved):
One minor change in the fixtures. Else it LGTM

packages/calling/src/CallHistory/callHistoryFixtures.ts
Kesari3008: done

PR #3396, opened 2024-02-22
fix: when reconnecting, JS-SDK calls Locus /media API with empty mediaId twice

marcin-bazyl:
# COMPLETES #[SPARK-500905](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-500905)

When we reconnect, we send some roap messages and 2 of them (TURN_DISCOVERY_REQUEST and OFFER) are being sent with empty media Id. Only the first one of them should be sent with empty media id. Empty media id indicates to Locus that this is a new connection and that they should create a new confluence in the server

As we are now hardcoded to always send TURN_DISCOVERY_REQUEST as the first Roap message when reconnecting, we can now also hardcode the code for sending offer to never use empty media id.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/roap/index.ts
+++ b/packages/@webex/plugin-meetings/src/roap/index.ts
@@ -181,7 +181,7 @@ export default class Roap extends StatelessWebexPlugin {
    * @memberof Roap
    */
   sendRoapMediaRequest(options: any) {
-    const {meeting, seq, sdp, reconnect, tieBreaker} = options;
+    const {meeting, seq, sdp, tieBreaker} = options;
     const roapMessage = {
       messageType: ROAP.ROAP_TYPES.OFFER,
       sdps: [sdp],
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -191,64 +191,62 @@ export default class Roap extends StatelessWebexPlugin {
...
 
-    // When reconnecting, it's important that the first roap message being sent out has empty media id.
-    // Normally this is the roap offer, but when TURN discovery is enabled,
-    // then this is the TURN discovery request message
-    return this.turnDiscovery.isSkipped(meeting).then((isTurnDiscoverySkipped) => {
-      const sendEmptyMediaId = reconnect && isTurnDiscoverySkipped;
+    // The only time we want to send an empty media id is when we are reconnecting, because this way we tell Locus
+    // that it needs to create a new confluence, but when reconnecting we always send TURN_DISCOVERY_REQUEST first,
+    // so we don't need to ever send an empty media id here
+    const sendEmptyMediaId = false;
```
torgeadelin, line 197: this will always be false no? what's the point adding it here?

torgeadelin: as far as i can tell this method is the generic sendRoapMediaRequest regardless if it's a reconnection or connection

marcin-bazyl: yeah I could remove it, I left it so that it's kind of more explicit and easier to understand and also if potentially we change the logic again in the future about when to do TURN discovery when reconnecting then we might need to send empty media Id here

jor-row (approved):

torgeadelin (approved):

PR #3397, opened 2024-02-23
feat(tools/packages): update command

InteractiveTimmy:
# NOTE

When this pull request is merged, this will publish the `@webex/cli-tools` and `@webex/package-tools` packages to **NPM** for consumption within down-stream clients.

# Description

The scope of the changes in this pull request are to add the `update` command to the `@webex/package-tools` package. This new command allows for consumers of the JS SDK to update their dependencies based on the latest available from `NPMJS`.

# Example

## Before

```jsonc
/* ./package.json */
{
  /* ... */
  "dependencies": {
    /* ... */
    "querystring-es3": "^0.2.1"
  },
  "devDependencies": {
    /* ... */
    "@babel/cli": "^7.17.10",
    "@babel/core": "^7.17.10"
  }
}
```

## Command

```bash
# Input
# package.json script `webex-package-tools` calls `webex-package-tools`
yarn webex-package-tools update --packages @babel/cli querystring-es3 --tag next

# Output
querystring-es3: ^0.2.1 => 1.0.0-0
@babel/cli: ^7.17.10 => 8.0.0-alpha.6
```

## After

```jsonc
/* ./package.json */
{
  /* ... */
  "dependencies": {
    /* ... */
    "querystring-es3": "1.0.0-0" /* updates to eplicit exact */
  },
  "devDependencies": {
    /* ... */
    "@babel/cli": "8.0.0-alpha.6", /* updates all dependency groups */
    "@babel/core": "^7.17.10" /* does not impact non-provided packages */
  }
}
```

# Links

* [Task](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-497350)

### DIFF
--- a/packages/tools/cli/package.json
+++ b/packages/tools/cli/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@webex/cli-tools",
-  "private": true,
   "packageManager": "yarn@3.4.1",
   "engines": {
     "node": ">=16.0.0",
--- a/packages/tools/package/README.md
+++ b/packages/tools/package/README.md
@@ -1,4 +1,4 @@
-# @webex/cli-tools
+# @webex/package-tools
 
 [![icense: mit](https://img.shields.io/badge/License-MIT-blueviolet?style=flat-square)](https://github.com/webex/webex-js-sdk/blob/master/LICENSE)
 ![state: beta](https://img.shields.io/badge/State\-Beta-blue?style=flat-square)
--- a/packages/tools/package/index.js
+++ b/packages/tools/package/index.js
@@ -8,6 +8,7 @@ const main = () => {
   commands.mount(PackageTools.list);
   commands.mount(PackageTools.scripts);
   commands.mount(PackageTools.sync);
+  commands.mount(PackageTools.update);
 
   commands.process();
 };
--- a/packages/tools/package/package.json
+++ b/packages/tools/package/package.json
@@ -1,9 +1,8 @@
 {
   "name": "@webex/package-tools",
-  "private": true,
   "packageManager": "yarn@3.4.1",
   "engines": {
-    "node": ">=18.0.0",
+    "node": ">=16.0.0",
     "npm": ">=8.0.0",
     "yarn": ">=3.0.0"
   },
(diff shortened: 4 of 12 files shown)


packages/tools/cli/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@webex/cli-tools",
-  "private": true,
```
sreenara, line 3: Is there a reason why we are making this also public?

sreenara (commented):
@InteractiveTimmy could you please add some examples of the outputs of the update command to the PR description?

packages/tools/cli/package.json
sreenara: Nevermind, I see `CommandsCommand` being used.

sreenara:
Another general comment @InteractiveTimmy , if the pipeline automatically publishes the current cli-tools and package-tools, it'll do a sync, make them versions 0.0.0-next.0 and publish those versions to NPM. Perhaps we should avoid that and manually publish with version 1.0.0 or something similar?

InteractiveTimmy:
> @InteractiveTimmy could you please add some examples of the outputs of the update command to the PR description?

I have updated the description in this pull request.

> Another general comment @InteractiveTimmy , if the pipeline automatically publishes the current cli-tools and package-tools, it'll do a sync, make them versions 0.0.0-next.0 and publish those versions to NPM. Perhaps we should avoid that and manually publish with version 1.0.0 or something similar?

I do not believe this should be published at version 1.0.0, as this is still an EFT of these packages for consumers. Whenever your team gets a request from a customer that would like this feature, or if you plan on providing this featureset in the core Webex JS SDK utils for external developers, we can then bump to 1.0.0.

sreenara (approved):
LGTM!

sreenara:
@InteractiveTimmy one note is that this only updates the root level package.json and doesn't recursively run across all packages within the mono repo.

InteractiveTimmy:
> @InteractiveTimmy one note is that this only updates the root level package.json and doesn't recursively run across all packages within the mono repo.

This will update any package that runs the command as a sub-module. So it can be ran from the context of any package or sub-package in a monorepo, if installed at that context.

PR #3401, opened 2024-02-26
fix: don't do TCP reachability on VMN, add API to toggle TCP reachability

marcin-bazyl:
# COMPLETES #[SPARK-501381](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-501381)

Linus on VMN does not support TCP reachability, so we should not attempt it, because it will fail.

- stopped doing TCP reachabillity on video mesh nodes
- added private API to toggle the TCP reachability config - this is needed so that we can roll this out behind a feature toggle (same approach was used in the past for similar things)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -699,6 +699,24 @@ export default class Meetings extends WebexPlugin {
...
   }
 
+  /**
+   * API to toggle TCP reachability, needs to be called before webex.meetings.register()
+   * @param {Boolean} newValue
+   * @private
+   * @memberof Meetings
+   * @returns {undefined}
+   */
+  private _toggleTcpReachability(newValue: boolean) {
```
marcin-bazyl, line 709: yes, this function is a bit weird, but I just copied the approach from similar other existing ones

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -699,6 +699,24 @@ export default class Meetings extends WebexPlugin {
...
+    if (typeof newValue !== 'boolean') {
+      return;
+    }
+    // @ts-ignore
+    if (this.config?.experimental?.enableTcpReachability !== newValue) {
+      // @ts-ignore
+      this.config.experimental.enableTcpReachability = newValue;
```
maxinteger, line 716: based on the `if` both `config` and `experimental` can be undefined. it would be better to use [Lodash set](https://lodash.com/docs/4.17.15#set) or similar

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -341,18 +341,21 @@ export default class Reachability {
...
     }
 
-    // @ts-ignore
-    const includeTcpReachability = this.webex.config.meetings.experimental.enableTcpReachability;
-
     LoggerProxy.logger.log(
       `Reachability:index#performReachabilityChecks --> doing UDP${
-        includeTcpReachability ? ' and TCP' : ''
+        // @ts-ignore
+        this.webex.config.meetings.experimental.enableTcpReachability ? ' and TCP' : ''
```
maxinteger, line 347: in the previous file you set `config.experimental.enableTcpReachability` but here you read `config.meetings.experimental.enableTcpReachability` how is this values copied?

Also, is the path always OK? Do not you need Optional chaining operator?

maxinteger (changes requested):

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: good point, I tested it with the web app and it was working when I called _toggleTcpReachability() so it must be copied somewhere, but I'll check

I'll also add the optional chaining

marcin-bazyl: I've checked and webex.config.meetings is the same as this.config in Meetings class:
image

packages/@webex/plugin-meetings/src/meetings/index.ts
marcin-bazyl: actually I've removed the optional chaining. There are lots of other existing places in SDK where config properties are accessed without optional chaining and the meetings plugin has a default config defined with all these props.

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: as mentioned in the other comment, optional chaining is not needed

PR #3209, opened 2023-11-20
feat(meetings): refactor public join method

torgeadelin:
# COMPLETES https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-475778

## This pull request addresses

- refactors join to no longer wait for llm connection and receive transcribing
- there is a relatively big side effect caused by this refactor
- - before we did not allow users to join the meeting if either of these 2 failed
- - with the new refactor, now the users will always be able to join the meeting regardless if llm or transcribing succeeds. When this happens, the user will experience a smaller set of features in the meeting. 
- - future improvements will include to retry this operations in case they fail, but for now we agreed to not do it

## by making the following changes

- refactor public join method
- add unit tests

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

- unit test, manual test

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3405, opened 2024-02-27
fix(plugin-meetings): listen to transcript for guests

mkesavan13:
# COMPLETES Adhoc PR

- Right now, the SDK does not listen to events until we do `startTranscription` and Guest users do not need to startTranscription to receive transcription. This event listening doesn't happen at the Voicea level too. 
- The SDK doesn't wait for the LLM connection to be established
- The `Invalid right-hand for instanceOf Blob` error

- Listening to events in Voicea from meetings by default
- Listening to LLM events from voicea by default
- Making Voicea toggle call only if the current user's role is 'host'
- Wait for LLM connection to be established and then resolve Meeting Join
- Making a toggle in the Sample app to enable transcription on login
- In http-core, for the invalid error, the Blob is mocked instead of importing it from Buffer since we imported it only for jest tests

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Enabling Transcription after joining the meeting as a host
- Enabling Transcription after joining the meeting as a guest
- Did a regression testing for the flow where the Sample app waits for Meeting Join to complete and enables transcription. Works fine

### I certified that

- [] All existing and new tests passed
- [] I have updated the documentation accordingly

---

### DIFF
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -563,6 +563,13 @@ function doPostMediaSetup(meeting) {
 
 async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevice: false}) {
   const meeting = webex.meetings.getAllMeetings()[selectedMeetingId];
+
+  meeting.on('meeting:llm:connected', () => {
+    if (enableTranscript.checked) {
+      toggleTranscription(true);
+    }
+  });
+
   let resourceId = null;
 
   if (!meeting) {
@@ -635,10 +642,6 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
       await meeting.joinWithMedia({joinOptions, mediaOptions});
 
       doPostMediaSetup(meeting);
-    }
-
-    if(webex.meetings.config.enableAutomaticLLM && enableTranscript.checked) {
-      toggleTranscription(true);
     }
 
     enableMeetingDependentButtons(true);
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -53,6 +53,9 @@ const voiceaTranscriptionFormattedDisplay = document.querySelector('#voicea-tran
 const toggleUnifiedMeetings = document.getElementById('toggle-unified-meeting');
 const currentMeetingInfoStatus = document.getElementById('current-meeting-info-status');
 
+const enableLLM = document.getElementById('meetings-enable-llm');
+const enableTranscript = document.getElementById('meetings-enable-transcription');
+
 // Store and Grab `access-token` from localstorage
 if (localStorage.getItem('date') > new Date().getTime()) {
   tokenElm.value = localStorage.getItem('access-token');
@@ -114,7 +117,7 @@ function generateWebexConfig({credentials}) {
         enableAdhocMeetings: true,
         enableTcpReachability: tcpReachabilityConfigElm.checked,
       },
-      enableAutomaticLLM: true,
+      enableAutomaticLLM: enableLLM.checked,
     },
     credentials,
     // Any other sdk config we need
@@ -634,6 +637,10 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
       doPostMediaSetup(meeting);
     }
 
+    if(webex.meetings.config.enableAutomaticLLM && enableTranscript.checked) {
+      toggleTranscription(true);
+    }
+
     enableMeetingDependentButtons(true);
 
     // For meeting controls button onclick handlers
@@ -1096,7 +1103,7 @@ function setTranscriptEvents() {
     });
 
     meeting.on('meeting:receiveTranscription:stopped', () => {
-      generalToggleTranscription.innerText = "Stop Transcription";
+      generalToggleTranscription.innerText = "Start Transcription";
       generalTranscriptionContent.innerHTML = 'Transcription Content: Webex Assistant must be enabled, check the console!';
     });
   }
--- a/docs/samples/browser-plugin-meetings/app.js
+++ b/docs/samples/browser-plugin-meetings/app.js
@@ -564,7 +564,7 @@ function doPostMediaSetup(meeting) {
 async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevice: false}) {
   const meeting = webex.meetings.getAllMeetings()[selectedMeetingId];
 
-  meeting.on('meeting:llm:connected', () => {
+  meeting.on('meeting:transcription:connected', () => {
     if (enableTranscript.checked) {
       toggleTranscription(true);
     }
--- a/docs/samples/browser-plugin-meetings/index.html
+++ b/docs/samples/browser-plugin-meetings/index.html
@@ -88,6 +88,9 @@
               <legend>Meetings</legend>
               <input type="checkbox" id="enable-tcp-reachability">
               <label for="enable-tcp-reachability">Enable TCP reachability (experimental)</label>
+              <br/>
+              <input id="meetings-enable-llm" checked type="checkbox">
+              <label for="meetings-enable-llm-label">Enable LLM</label>
             </div>
           </div>
 
@@ -294,9 +297,9 @@
           <fieldset>
             <legend>Voicea Transcription</legend>
             <div class="u-mv">
-              You will see the formatted transcription below if you have <b>did one of the below</b>,
+              You will see the formatted transcription below if you have did the following,
               <ul>
-                <li>Set the <span class="inline-code">receiveTranscription</span> config to <span class="inline-code">true</span> before joining the meeting</li>
+                <li>Set the <span class="inline-code">enableAutomaticLLM</span> config to <span class="inline-code">true</span> before doing webex.init</li>
                 <li>Invoked <span class="inline-code">meeting.startTranscription()</span> after joining the meeting</li>
               </ul>
               <div id="voicea-transcription-formatted-display">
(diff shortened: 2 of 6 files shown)


torgeadelin:
We don't want to wait for llm for the user to join, we recently changed that. [Colin](spark-mention://33e6d0d8-6fab-470d-ba74-e482d90ccc73)
this change was introduced here: https://github.com/webex/webex-js-sdk/pull/3209

torgeadelin:
sorry, i closed by mistake

torgeadelin (changes requested):
We don't want to wait for llm for the user to join, we recently changed that. Colin
this change was introduced here: https://github.com/webex/webex-js-sdk/pull/3209

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -634,6 +637,10 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
       doPostMediaSetup(meeting);
     }
 
+    if(webex.meetings.config.enableAutomaticLLM && enableTranscript.checked) {
```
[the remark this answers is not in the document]
mkesavan13: Addressed this. The change for this is now part of the new event listener. Please check

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5011,10 +5017,10 @@ export default class Meeting extends StatelessWebexPlugin {
 
         return Promise.reject(error);
       })
-      .then((join) => {
+      .then(async (join) => {
```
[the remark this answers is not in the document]
mkesavan13: This is addressed. Kindly check

packages/@webex/http-core/src/request/request.shim.js
```diff
@@ -12,7 +12,6 @@
 import {defaults, isArray, pick} from 'lodash';
 import qs from 'qs';
 
-import {Blob} from 'buffer';
```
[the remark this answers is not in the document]
mkesavan13: Blob is available in both JSDOM (Browser) as well as the Node.JS environment by default. Just that the Jest test environment doesn't have it. 

So, in the real world, we would have the type `Blob` available.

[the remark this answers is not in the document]
mkesavan13: We imported `Blob` as part of the fix for http-core unit tests to make it available in the Jest test environment. However, that was causing troubles in real world as we are expecting the Blob from Browser and not the 'buffer' 

The older fix was wrong.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -560,6 +563,13 @@ function doPostMediaSetup(meeting) {
 
 async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevice: false}) {
   const meeting = webex.meetings.getAllMeetings()[selectedMeetingId];
+
+  meeting.on('meeting:llm:connected', () => {
```
arun3528, line 567: i am not sure if we need to name it llm as developer does not know anything about it 

meeting:socket:connected

thoughts ?

packages/@webex/http-core/src/request/request.shim.js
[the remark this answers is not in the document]
arun3528: we need to move to fetch soon :-|

arun3528 (commented):
change the event name rest looks good , are you pushing to beta as well ?

docs/samples/browser-plugin-meetings/app.js
sreenara: @arun3528 does LLM replace the existing mercury connection that gets created when registering a device?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5014,15 +5020,31 @@ export default class Meeting extends StatelessWebexPlugin {
...
-              reason: error?.message,
-              stack: error.stack,
+              Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.LLM_CONNECTION_AFTER_JOIN_FAILURE, {
+                correlation_id: this.correlationId,
+                reason: error?.message,
+                stack: error.stack,
+              });
+            })
+            .then(() => {
+              LoggerProxy.logger.info('Meeting:index#join --> Update LLM Connection Success');
```
sreenara, line 5037: same comment here

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -375,6 +375,7 @@ export const EVENT_TRIGGERS = {
   MEETING_LOCUS_URL_UPDATE: 'meeting:locus:locusUrl:update',
   MEETING_STREAM_PUBLISH_STATE_CHANGED: 'meeting:streamPublishStateChanged',
 
+  MEETING_LLM_CONNECTED: 'meeting:llm:connected',
```
sreenara, line 378: agree with Arun here. we can change the name.

sreenara (changes requested):
Minor comment on the name of the event.
Minor change in the logs

docs/samples/browser-plugin-meetings/app.js
mkesavan13: @sreenara - No. They are two different socket connections. Both exist in the browser. So, should we name it like `meeting:transcription:connected` ?

sreenara: @mkesavan13 so with this we'll have `meeting:transcription:connected` and `meeting:transcription:started` that a developer has to listen for?

mkesavan13: Connected is to know that the socket is connected. Transcription started will be received when they toggle the transcription

sreenara: Sounds good @mkesavan13 . Let's go ahead with this.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5014,15 +5020,31 @@ export default class Meeting extends StatelessWebexPlugin {
       .then((join) => {
         // @ts-ignore - config coming from registerPlugin
         if (this.config.enableAutomaticLLM) {
-          this.updateLLMConnection().catch((error) => {
-            LoggerProxy.logger.error('Meeting:index#join --> Update LLM Connection Failed', error);
+          this.updateLLMConnection()
+            .catch((error) => {
+              LoggerProxy.logger.error(
+                'Meeting:index#join --> Update LLM Connection Failed',
```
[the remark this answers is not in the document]
mkesavan13: Updated this @sreenara

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Updated this @sreenara

packages/@webex/plugin-meetings/src/constants.ts
mkesavan13: Addressed this @sreenara

mkesavan13:
> We don't want to wait for LLM for the user to join, we recently changed that. Colin this change was introduced here: #3209

@torgeadelin - Following the conversation in the code review space, have removed the waiting for the LLM connection and triggered an event instead. Kindly check

sreenara (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7673,6 +7697,9 @@ export default class Meeting extends StatelessWebexPlugin {
       if (roles.includes(SELF_ROLES.COHOST)) {
         return 'cohost';
       }
+      if (roles.includes(SELF_ROLES.PRESENTER)) {
```
rarajes2, line 7700: Looks like the sequence of roles matters here.

mkesavan13: I think so and I think the Presenter takes precedence than attendee. Attendees would not have meeting control at all.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -560,6 +563,13 @@ function doPostMediaSetup(meeting) {
 
 async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevice: false}) {
   const meeting = webex.meetings.getAllMeetings()[selectedMeetingId];
+
+  meeting.on('meeting:transcription:connected', () => {
+    if (enableTranscript.checked) {
+      toggleTranscription(true);
```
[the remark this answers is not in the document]
mkesavan13: We cannot do this. The `toggleTranscription` method is also being called when clicked on the 'Start Transcription' button in the UI.

So, when joining the meeting, if the check box isn't checked, we would not call the method at all. If we do this change, we will still call the function with 'enable' as false which seems unnecessary.

docs/samples/browser-plugin-meetings/index.html
```diff
@@ -294,9 +297,9 @@ <h2 class="collapsible">
           <fieldset>
             <legend>Voicea Transcription</legend>
             <div class="u-mv">
-              You will see the formatted transcription below if you have <b>did one of the below</b>,
+              You will see the formatted transcription below if you have did the following,
               <ul>
-                <li>Set the <span class="inline-code">receiveTranscription</span> config to <span class="inline-code">true</span> before joining the meeting</li>
+                <li>Set the <span class="inline-code">enableAutomaticLLM</span> config to <span class="inline-code">true</span> before doing webex.init</li>
```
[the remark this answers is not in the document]
mkesavan13: Changed this

rarajes2 (approved):

PR #3408, opened 2024-02-28
fix(plugin-meeting-samples): fixes the VBG and screen re-share issue

rarajes2:
# COMPLETES #< [SPARK-500552](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-500552) >

**Bugs**

- **Remote Screenshare** - When a remote participant stops sharing, the sample app takes out the Remote Share Video. However, when the remote share is on again, we are not able to see it in the sample app
- **VBG Video** - When a VBG image/video is enabled and the user toggles the video, only the Camera source gets shut and the effect tends to stay in the Local Stream

Fixes the bugs in the samples page mentioned above

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1691,6 +1696,15 @@ function toggleSendVideo() {
     localMedia.cameraStream.setMuted(newMuteValue);
 
     console.log(`MeetingControls#toggleSendVideo() :: Successfully ${newMuteValue ? 'muted': 'unmuted'} video!`);
+
+    if (newMuteValue && vbgEffect?.isEnabled) {
+      handleVbg();
+      wasVBGEnabledBeforeToggle = true;
+    } else if(wasVBGEnabledBeforeToggle) {
+      handleVbg();
```
mkesavan13, line 1704: handleVbg also handles the VBG button text. So, when the user has enabled VBG, then they mute and if the button says "Enable VBG" again, won't it be misleading?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -651,9 +651,11 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
       viewBreakouts();
     });
 
-        meeting.on('meeting:stoppedSharingRemote', () => {
-          meetingStreamsRemoteShare.srcObject = null;
-        });
+    meeting.on('meeting:stoppedSharingRemote', () => {
+      const remoteShareTemp = meetingStreamsRemoteShare.srcObject;
+      meetingStreamsRemoteShare.srcObject = null;
+      meetingStreamsRemoteShare.srcObject = remoteShareTemp;
```
mkesavan13, line 657: Could you please explain how this fixes the issue?

rarajes2: Earlier we were simply assigning the null to the video element which was removing the remote media reference added during `addMedia()` call. Here we first remove the media to stop showing the remote stream being received and again attach the remote media stream but since the remote is not sending the frames anymore it does not show the screenshare but keeps the event attached. Hence, when we again start sharing the screen, the media flows correctly to the video element.

docs/samples/browser-plugin-meetings/app.js
rarajes2: That's why introduced an additional check `wasVBGEnabledBeforeToggle`. It keeps the flow intact.

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: it would be good to add this explanation as a comment in the code

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: I don't understand the point of this fix. What are we trying to fix here? When I enable an effect and then mute the video and unmute it I expect the effect to be applied and that is happening currently without this PR and also with this PR it still works that way.
When user mutes the video, we call setMuted(true) and that's all we're supposed to be doing. It's down to the webrtc-core lib and the effects lib to make sure that resources are not wasted when a track with an effect is muted. There should be no need for any fix at the app level. If there is any real problem with the fact that the effect is still there, then this should be fixed in webrtc-core or the effects lib.

Also, with this new code we get very weird behavior with these steps:
1. enable the effect
2. mute the video
I see now the effect button saying "Enable VBG", which is very misleading to the user, because if they don't press it and just unmute they see the effect is already enabled anyway. 

Also another scenario is broken:
1. enable effect
4. mute video
5. press "enable vbg"
6. unmute video
now you can see the video without the effect despite the fact that you've pressed "enable vbg" in step 3.

marcin-bazyl (changes requested):
Why is this raised against the "next" branch? we should aim for having minimum diff between beta and next. As far as I understand the plan for beta/next branches, any fix like this should be done against beta branch. @sreenara , @arun3528 , @InteractiveTimmy am I misunderstanding something?

docs/samples/browser-plugin-meetings/app.js
mkesavan13: Yes. Please add a short comment explaining it and also does this work in both "Stop Share" and "Unpublish" cases?

rarajes2:
> Why is this raised against the "next" branch? we should aim for having minimum diff between beta and next. As far as I understand the plan for beta/next branches, any fix like this should be done against beta branch. @sreenara , @arun3528 , @InteractiveTimmy am I misunderstanding something?


Since the changes were only added to the samples page and not to the SDK code, raised the PR only for next branch. Will cherry pick the change and raise the PR against beta as well.

docs/samples/browser-plugin-meetings/app.js
rarajes2: > does this work in both "Stop Share" and "Unpublish" cases?

Yes

rarajes2: Added comment.

docs/samples/browser-plugin-meetings/app.js
rarajes2: I have removed this change as the issue needs to be fixed in either the effect package or webrtc-core. Have created a separate JIRA - https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-502882

mkesavan13 (approved):

karthikraji2020 (approved):
LGTM

marcin-bazyl (approved):

PR #3409, opened 2024-02-28
test: test-user set preferred webexsite

cscheffauer:

- Setting the preferred WebexSite of a created Test User

- Adding a new `setPreferredSite` function in test-users package, to allow for setting the preferred webexsite after a test user has been created
- Didn't add integration tests, since they are not testable / there is an issue currently that test-users package doesn't allow running the integration tests (saying "Found 0 tests" when running).
image

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tested the function locally

### I certified that

---

### DIFF
--- a/packages/@webex/test-users/src/index.js
+++ b/packages/@webex/test-users/src/index.js
@@ -321,6 +321,52 @@ export function removeTestUser(options = {}) {
   });
 }
 
+/**
+ * Sets the preferredWebexSite for the provided user
+ * @param {Object} options
+ * @param {string} options.userId user id to set site
+ * @param {string} options.preferredSite new preferred webexsite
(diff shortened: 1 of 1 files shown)


arun3528 (approved):
Can we get some documenation on this , if all the test users need a site patched , does it work based on client id and secret , if we need to use it would we need to get it authorized some where ?

cscheffauer:
> Can we get some documenation on this , if all the test users need a site patched , does it work based on client id and secret , if we need to use it would we need to get it authorized some where ?

The setPreferredWebexSite is only necessary if in the test org, where the test users are created, multiple sites are configured. This method helps to define which of the test site should be used.

shnaaz (approved):

PR #3414, opened 2024-02-29
fix(plugin-meetings): sample app mute button fix in participant list

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-500553 

Device Management - The Device Management section seems to be disabled all the time. Needs to be enabled and tested
Participant list section (Mute button) - The mute button in the participants list doesn't change the behaviour based on the chosen participant. For example, the button shows "Mute" for a chosen participant whose Audio state is already muted.

The mute button in the participant list gets updated based on the audio mute status of the participant selected.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

https://github.com/webex/webex-js-sdk/assets/65543166/86d1ff86-3f32-460e-9426-62da6d72242c

### I certified that

---


Kesari3008:
Ignore console logs as I will be removing those

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -622,6 +622,11 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
       viewBreakouts();
     });
 
+    meeting.on('meeting:self:mutedByOthers', () => {
+      console.log('pkesari_received event to be muted by others');
```
rarajes2, line 626: Please update the log comments everywhere added newly.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -1496,7 +1502,8 @@ function toggleSendAudio() {
     const newMuteValue = !localMedia.microphoneStream.muted;
 
     localMedia.microphoneStream.setMuted(newMuteValue);
-    handleAudioButton(newMuteValue);
+    console.log('pkesari_mute state in stream and new mute value: ', localMedia.microphoneStream.muted, newMuteValue);
+    handleAudioButton(localMedia.microphoneStream.muted);
```
rarajes2, line 1506: Question: Why do we need to change the argument from `newMuteValue` to `localMedia.microphoneStream.muted` here?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2855,20 +2868,17 @@ function removeMember(removeButton) {
 
 function muteMember(muteButton) {
   const meeting = getCurrentMeeting();
-  const unmute = muteButton.getAttribute('data-unmute');
-  const mute = unmute !== 'true';
-
   const participantID = getRadioValue('participant-select');
+  const selectedMember = meeting.members.membersCollection.get(participantID);
+  const mute = selectedMember.isAudioMuted ? 'false' : 'true';
```
rarajes2, line 2873: `const mute` -> `const isMuted`

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2811,6 +2818,12 @@ const participantsList = document.querySelector('#participant-list');
 const participantTable = document.querySelector('#participant-table');
 const participantButtons = document.querySelector('#participant-btn');
 
+participantTable.addEventListener('click', (event) => {
+  const selectedParticipant = meeting.members.membersCollection.get(event.target.id);
```
rarajes2, line 2822: - We should add `event.preventDefault()` at the beginning and `event.stopPropagation()` at the end.
- We should add a check `if (!selectedParticipant) { return; }`

rarajes2 (commented):
Please remove unnecessary logs.

docs/samples/browser-plugin-meetings/app.js
mkesavan13: - Table click doesn't have an action by default that needs prevention (@rarajes2)
- We need that check as well as a check to do anything only if the radio button is clicked. This will keep calling the membersCollection.get when clicked anywhere in the table any number of times

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2811,6 +2818,12 @@ const participantsList = document.querySelector('#participant-list');
...
+  const muteButton =  document.querySelector('#participant-btn .btn-group button:first-child');
+  muteButton.innerText = selectedParticipant.isAudioMuted ? 'Unmute': 'Mute';
```
mkesavan13, line 2824: I thought if the mute state is already muted, we will just disable the button because Host can only Mute a participant but cannot unmute

mkesavan13 (changes requested):
In addition to the comments, do we need all those new logs or were they pushed by mistake after debugging?

docs/samples/browser-plugin-meetings/app.js
Kesari3008: These console logs will be removed. I have kept them for testing purpose

docs/samples/browser-plugin-meetings/app.js
Kesari3008: When a user is muted from the mute button in the participant list, the button under the streams section doesn't reflect the correct state hence this change was required to keep that button text updated properly

docs/samples/browser-plugin-meetings/app.js
Kesari3008: We cannot unmute but there is a request send to unmute which is visible on the other participant's event list hence I didn't want to break that functionality. We can show one message to indicate on the host that unmute request has been sent to make it more clear

docs/samples/browser-plugin-meetings/app.js
Kesari3008: Variable name in the .mute function is expected as mute hence kept it as it is

docs/samples/browser-plugin-meetings/app.js
rarajes2: Correct @mkesavan13, We don't need preventDefault() but `stopPropagation()` would stop the further bubbling.

docs/samples/browser-plugin-meetings/app.js
rarajes2: Okay, but `mute` function with the variable name `mute` as an argument may confuse

docs/samples/browser-plugin-meetings/app.js
Kesari3008: Console logs are removed now

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2823,6 +2827,17 @@ const participantsList = document.querySelector('#participant-list');
 const participantTable = document.querySelector('#participant-table');
 const participantButtons = document.querySelector('#participant-btn');
 
+participantTable.addEventListener('click', (event) => {
+  if (event.target.type === 'radio') {
+    const selectedParticipant = meeting.members.membersCollection.get(event.target.id);
+    if (!selectedParticipant) {
+      return;
+    }
+    const muteButton =  document.querySelector('#participant-btn .btn-group button:first-child');
```
mkesavan13, line 2836: Can't we add ID to the button and do getElementByID? This will break if anyone adds another button before mute

docs/samples/browser-plugin-meetings/app.js
mkesavan13: In that case, can we make this button "Request to unmute" instead of just "Unmute"?

docs/samples/browser-plugin-meetings/app.js
mkesavan13: Again, we know the table won't have another click event listener. So stopPropogation doesn't seem mandatory to me.

mkesavan13 (changes requested):
Just a couple more minor comments and we will be good to go.

docs/samples/browser-plugin-meetings/app.js
Shreyas281299: Changes the argument to muteStatus.

docs/samples/browser-plugin-meetings/app.js
Shreyas281299: Changed test to -> 'Request to unmute'

docs/samples/browser-plugin-meetings/app.js
Shreyas281299: Changed to use getEelementById.

InteractiveTimmy (commented):
The `beta` branch is currently frozen (until `2024`-`03`-`04`) as a part of the WWC `next`/`beta` convergence, should we try to make these changes against `next`?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -625,15 +625,19 @@ async function joinMeeting({withMedia, withDevice} = {withMedia: false, withDevi
...
+       * attach the remote media stream (which gets added during addMedia() call) but since the remote
+       * is not sending the frames anymore it does not show the screenshare but keeps the event attached.
        * Hence, when we again start sharing the screen, the media flows correctly to the video element.
        */
       const remoteShareTemp = meetingStreamsRemoteShare.srcObject;
       meetingStreamsRemoteShare.srcObject = null;
       meetingStreamsRemoteShare.srcObject = remoteShareTemp;
     });
 
+    meeting.on('meeting:self:mutedByOthers', () => {
```
marcin-bazyl, line 637: what about "meeting:self:unmutedByOthers"?

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2823,6 +2827,17 @@ const participantsList = document.querySelector('#participant-list');
...
 const participantButtons = document.querySelector('#participant-btn');
 
+participantTable.addEventListener('click', (event) => {
+  if (event.target.type === 'radio') {
+    const selectedParticipant = meeting.members.membersCollection.get(event.target.id);
+    if (!selectedParticipant) {
+      return;
+    }
+    const muteButton = document.getElementById('mute-participant-btn')
+    muteButton.innerText = selectedParticipant.isAudioMuted ? 'Request to unmute': 'Mute';
```
marcin-bazyl, line 2837: FYI, depending on meeting configuration, it is sometimes possible to actually unmute someone not just request to unmute, so the text "request to unmute" is not entirely correct, but I guess that can be improved further in another PR, not a blocker for this one

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2867,20 +2882,14 @@ function removeMember(removeButton) {
 
 function muteMember(muteButton) {
   const meeting = getCurrentMeeting();
-  const unmute = muteButton.getAttribute('data-unmute');
-  const mute = unmute !== 'true';
-
   const participantID = getRadioValue('participant-select');
+  const selectedMember = meeting.members.membersCollection.get(participantID);
+  const muteStatus = selectedMember.isAudioMuted ? 'false' : 'true';
```
marcin-bazyl, line 2887: why are we using string values 'false' and 'true' and not booleans? the method Meeting.mute() takes a boolean, doesn't it? see here: https://github.com/webex/webex-js-sdk/blob/d3b5012e341426ddb6aea689b540c045c32929f8/packages/%40webex/plugin-meetings/src/meeting/index.ts#L3224
image

it would also be good to rename that variable to `newMuteStatus` or something like that because it's not the current mute status

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2867,20 +2882,14 @@ function removeMember(removeButton) {
...
-    meeting.mute(participantID, mute).then((res) => {
-      console.log(res, `participant is ${mute ? 'mute' : 'unmute'}`);
-      if (mute) {
-        muteButton.setAttribute('data-unmute', 'true');
-      }
-      else {
-        muteButton.removeAttribute('data-unmute');
-      }
+    meeting.mute(participantID, muteStatus).then((res) => {
+      console.log(res, `participant is ${muteStatus ? 'mute' : 'unmute'}`);
```
marcin-bazyl, line 2891: log should say "muted" or "unmuted"

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3528,8 +3537,9 @@ function viewParticipants() {
     const btnDiv = document.createElement('div');
 
     btnDiv.classList.add('btn-group');
-
-    btnDiv.appendChild(createButton('Mute', muteMember));
+    const muteBtn = createButton('Mute', muteMember)
```
marcin-bazyl, line 3540: minor thing, but it would be nicer to just modify createButton to take an extra parameter like "options" or "props" that would then set these props on the new button so we could just call it like this: 
`btnDiv.appendChild(createButton('Mute', muteMember, {id: 'mute-participant-btn'}));`

docs/samples/browser-plugin-meetings/index.html
```diff
@@ -477,6 +477,9 @@ <h2 class="collapsible">
         <div>
           <fieldset id="meetingControlsDevices">
             <legend>Media Devices</legend>
+
+            <p class="note">NOTE: You must join the meeting to update the devices.</p>
```
marcin-bazyl, line 481: there is no technical limitation preventing this from being fixed, it's just how the sample app is currently implemented, so it would be good to raise a jira to fix this properly and allow device selection outside of the meeting

marcin-bazyl (changes requested):
looks good, marking it as "requires changes" only because of the type issue (string vs boolean) when calling Meeting.mute()

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: now every call to `handleAudioButton()` passes `localMedia.microphoneStream.muted` as the argument, so I think it would be better to change `handleAudioButton()` to not have any arguments and just hardcode `localMedia.microphoneStream.muted` inside it

docs/samples/browser-plugin-meetings/app.js
Kesari3008: We don't have any action to be done in Samples app associated with this event hence not listening for it here.

docs/samples/browser-plugin-meetings/app.js
Kesari3008: Makes sense....Updated it now

docs/samples/browser-plugin-meetings/app.js
Kesari3008: It was an existing log so didn't touch it. Updated it now

docs/samples/browser-plugin-meetings/app.js
Kesari3008: There are scenarios where Unmute is possible to be done so keeping it Request To unmute wouldn't be always correct. Updating it to Toggle Audio now

docs/samples/browser-plugin-meetings/app.js
Kesari3008: mute function inside plugin meetings expects the parameter name as mute only else it starts to fail hence kept it like that.

docs/samples/browser-plugin-meetings/index.html
Kesari3008: Sure, we will raise a separate Jira for this

docs/samples/browser-plugin-meetings/app.js
Kesari3008: By mistake, I added inverted commas and since muting the participants was working missed the type mismatch. Thanks for highlighting it. Corrected it now. Are you suggesting changing the variable name in the mute method inside the meeting object or in the app.js

docs/samples/browser-plugin-meetings/index.html
Kesari3008: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503859

mkesavan13 (approved):
Looks good to me.

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: > Are you suggesting changing the variable name in the mute method inside the meeting object or in the app.js

just in the app.js

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: don't we need to update the UI, like labels on the button?

docs/samples/browser-plugin-meetings/app.js
marcin-bazyl: but "toggle audio" doesn't feel correct. In most cases when someone muted themselves and you press "toggle audio" their audio won't be toggled, only the request to unmute will be sent, right?

Kesari3008: yes that's correct

docs/samples/browser-plugin-meetings/app.js
Kesari3008: My understanding here is that Unmuting by others is not possible so no update would be required.

marcin-bazyl: it is possible for certain meetings, you can configure it: 
image

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -2823,6 +2827,21 @@ const participantsList = document.querySelector('#participant-list');
...
 
+participantTable.addEventListener('click', (event) => {
+  if (event.target.type === 'radio') {
+    const selectedParticipant = meeting.members.membersCollection.get(event.target.id);
+    if (!selectedParticipant) {
+      return;
+    }
+    const muteButton = document.getElementById('mute-participant-btn')
+    if (selectedParticipant.isAudioMuted) {
+      muteButton.innerText = meeting.selfId === selectedParticipant.id ? 'Unmute' : 'Request to unmute';
```
marcin-bazyl, line 2838: there is also an option to unmute someone if "moderated unmute mode" is used, but it's not a blocker for this PR, can be added later

docs/samples/browser-plugin-meetings/app.js
Kesari3008: taken care

docs/samples/browser-plugin-meetings/app.js
Kesari3008: taken care now

docs/samples/browser-plugin-meetings/app.js
Kesari3008: addressed..added the event now

marcin-bazyl (approved):

PR #3424, opened 2024-03-01
fix(meetings): types building

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to update the build process to build types to the same location as they are being build on `next`.


sreenara:
This PR is no longer relevant

PR #3439, opened 2024-03-14
feat(metrics): add pre login batching

torgeadelin:
# COMPLETES [#< INSERT LINK TO ISSUE >](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-483903)

- add pre login batching for clientmetrics prelogin

### Change Type
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- unit test, manual test

### I certified that

---


packages/@webex/internal-plugin-metrics/src/batcher.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable require-jsdoc */
```
arun3528, line 1: why now its throwing error and not before ?

arun3528 (approved):

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
```diff
@@ -0,0 +1,97 @@
...
+/* eslint-disable valid-jsdoc */
+
+import {uniqueId} from 'lodash';
+import Batcher from './batcher';
+import {prepareDiagnosticMetricItem} from './call-diagnostic/call-diagnostic-metrics.util';
+import {generateCommonErrorMetadata} from './utils';
+
+const PRE_LOGIN_METRICS_IDENTIFIER = 'Pre Login Metrics -->';
+
+const PreLoginMetricsBatcher = Batcher.extend({
```
InteractiveTimmy, line 11: Why is this batcher created when `call-diagnostics-metrics-batcher` exists? Is there a reason why we did not reuse the existing batcher, or expand on the functionality of the existing batcher?

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable class-methods-use-this */
+/* eslint-disable valid-jsdoc */
```
InteractiveTimmy, line 2: Why are these here? It appears that all of your JSDoc is invalid currently, or, the ESLint rule needs to be updated.

`./.eslintrc.js`, within the `.ts` overrides:

Screenshot 2024-03-14 at 10 21 04 AM

Note that this updates some of the error messages (they wont be correct due to our extremely damaged eslint/typescript configuration), but when these changes exist, it works as intended.

Screenshot 2024-03-14 at 10 20 28 AM

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
```diff
@@ -0,0 +1,97 @@
...
+
+import {uniqueId} from 'lodash';
+import Batcher from './batcher';
+import {prepareDiagnosticMetricItem} from './call-diagnostic/call-diagnostic-metrics.util';
+import {generateCommonErrorMetadata} from './utils';
+
+const PRE_LOGIN_METRICS_IDENTIFIER = 'Pre Login Metrics -->';
+
+const PreLoginMetricsBatcher = Batcher.extend({
+  namespace: 'Metrics',
```
InteractiveTimmy, line 12: Is there a reason or purpose to assigning this `namespace` attribute?

This can be potentially dangerous (in the same way that assigning two elements the same ID attribute within an HTML document can be dangerous). [See Documentation for AmpState](https://github.com/AmpersandJS/ampersand-state).

packages/@webex/internal-plugin-metrics/test/unit/spec/prelogin-metrics-batcher.ts
```diff
@@ -0,0 +1,250 @@
...
+ */
+
+import {assert} from '@webex/test-helper-chai';
+import {config, Utils} from '@webex/internal-plugin-metrics';
+import {CallDiagnosticUtils} from '@webex/internal-plugin-metrics';
+import MockWebex from '@webex/test-helper-mock-webex';
+import sinon from 'sinon';
+import FakeTimers from '@sinonjs/fake-timers';
+import {NewMetrics} from '@webex/internal-plugin-metrics';
+import {uniqueId} from 'lodash';
```
InteractiveTimmy, line 12: Thanks for importing these all via their aliases and not using relative paths! good work!

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.ts
```diff
@@ -5,7 +5,7 @@ import {WebexHttpError} from '@webex/webex-core';
 import CallDiagnosticMetrics from '../../../../src/call-diagnostic/call-diagnostic-metrics';
 import CallDiagnosticLatencies from '../../../../src/call-diagnostic/call-diagnostic-metrics-latencies';
 import * as Utils from '../../../../src/call-diagnostic/call-diagnostic-metrics.util';
```
InteractiveTimmy, line 7: See the positive comment in the other test file. This approach (absolute/relative imports) breaks existing scripts in both `beta` and `next`, such as `yarn distsrc` and similar (review `./CONTRIBUTING.md`) on `beta`, and all runtime tests on `next`. Please consider switching these to use aliased imports if you could

(note that this may be out of scope for this pull request).

InteractiveTimmy (commented):
No blocking comments, as this appears to work as intended. There are a few architectural concerns / tooling concerns, but these are non-blocking.

Let me know if you have some time to discuss any of the items listed, thanks!

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
torgeadelin: they're hitting different APIs which is why i had it separate.
This will be refactored once Metrics MVP comes into play.. metrics a supports multiple types of metrics, and we need to prepare the events in different ways. In this case It's the prepare item and request match the call-diagnostic-metrics-batcher, but it won't always do. Nonetheless could still be merged into 1 batcher, but I didn't want to refactor too much code now, when I know this will be refactored in the near future.

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
torgeadelin: I just copied this over from the other batcher file.

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
torgeadelin: No reason, just following the pattern.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.ts
torgeadelin: I've refactored this.

torgeadelin: Seems a bit silly that I have to export those classes from the plugin in order to test them. Not sure if there's a way around that.. but I don't mind too much.

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
torgeadelin: I've fixed all of these, in prelogin and call diagnostic batcher

packages/@webex/internal-plugin-metrics/src/batcher.js
torgeadelin: I have fixed this too.

packages/@webex/internal-plugin-metrics/src/prelogin-metrics-batcher.ts
torgeadelin: Fixed the batcher.js too

shnaaz (approved):

PR #3440, opened 2024-03-14
feat(ca): add download time jmt

torgeadelin:
# COMPLETES https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506401

- add download jmt

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

-unit test, manual test

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -270,6 +270,14 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
     return this.precomputedLatencies.get('internal.client.pageJMT') || undefined;
   }
 
+  /**
+   * Download Time JMT
+   * @returns - latency
+   */
+  public getDownloadTimeJMT() {
+    return this.precomputedLatencies.get('internal.download.time') || undefined;
```
InteractiveTimmy, line 278: Is there a reason why this is a `Map` class object? I've noticed this throughout the codebase in areas where a regular object/array would have worked.

Assuming this is potentially an artifact from when this was attempted to be a singleton/instance hybrid plugin?

Since these keys are required to exist as strings, have he considered moving them to `constants`?

InteractiveTimmy, line 278: There is no place, within this branch, that `'internal.download.time'` is being explicitly assigned to `precomputedLatencies`, so this should always return `undefined`. Is this being handled via spread/map elsewhere? If so, we should potentially introduce some form of map/document/interface for maintainability of this segment of code.

**I would recommend**: moving away from `Map` instances unless you're specifically need to **key** an `Object` with `Object` instances.

Also, is there redundancy with `/* ... */ || undefined`? Just want to confirm, as I may be misunderstanding this area.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -241,6 +241,9 @@ export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
   const cdl = webex.internal.newMetrics.callDiagnosticLatencies;
 
   switch (eventName) {
+    case 'client.webexapp.launched':
```
InteractiveTimmy, line 244: Case options should be constants in this context. Would these be available via `@webex/event-dictionary-ts` within the `json` scope or should we simply create a "white-list" of constants we use?

Note that we could also generate another artifact from event-dictionary that could include just "event names" for consistency.

InteractiveTimmy (commented):
**NO BLOCKING COMMENTS**

It appears there may be some redundant code, as well as some smaller areas of improvement. Nothing is blocking, as it appears to work as expected.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
torgeadelin: I don't remember exactly why I decided to use Map, but I don't think was a matter of one better than other. I agree both can work.

Not sure it had to with the singleton issue.

the latencyTimestamps are typed, so no need for constants.
the precomputedLatencies are not typed, but I can create a type if needed.

torgeadelin: That's because we're setting that in Cantina and not in SDK. Some latencies are precomputed in Cantina and we just save them all here in one place.

|| undefined was added I think because in case latencies/timestamps equal to 0, would rather return undefined. CA doesn't want 0 values.

torgeadelin: Any reason why Map is problematic?
Also, the timestamps benefit from the insertion order of the Map. The precomputedLatencies, I guess that can be changed to {} if you really think it's worth it.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
torgeadelin: these are typed, so no need for constants

shnaaz (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
arun3528: still feel this need to be constant for 
single place to visualize event accoss the board , easy to maintain test and actual src constants

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
InteractiveTimmy: `Map` instances are not particularly problematic (hence the non-blocking). They don't support the level of type-enforcement you can get from typical interfaces when applied to objects:

```ts
const myData: MyData = {}; // fully typed, inspectable, exportable
const map: Map<string, any> = new Map(); // similar interface and generic params as `Record<string, any>`
```

In addition to this, the benefits of a `Map` instance, when compared to a traditional `Object` are not used here. We do not key the `Map` instance using any mutable data, so it is used as if it were an `Object` with a `get()` and `set()` method, losing the ability to inspect/cast and shape the data being set as a `Map`.

Let me know if there is a benefit to using a `Map` instance here if you could, as it may change my perception of this area of code.

InteractiveTimmy: In this case, constants are definitely needed, an engineer needs the exact string `internal.download.time` to receive the **internal download time**. If this was keyed in TS as:

```ts
interface PrecomputedLatencies {
  'internal.download.time': number;
}
```

or even Mapped as:

```ts
export type PrecomputedLatancy = 'internal.download.time';

/* ... somewhere else in code ... */

precomputedLatencies: Map<PrecomputedLatancy, number>;
```

This would allow for at least the keys to be inferred (partially). This also applies to a later comment.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
InteractiveTimmy: This makes sense, we don't want to "thicken" the bundle size by importing anything material from `@webex/event-dictionary-ts`. But, it would be helpful to have a parody of this locally:

```ts
/* ...constants.ts... */

const EVENT_NAMES: Record<string, MetricEventNames> = {
  EVENT_NAME_A: 'some-metrics-event-name',
};
```

These can then be used as a whitelist for what is currently consumed by this plugin without having to search the plugin file-by-file.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
torgeadelin: added types for the precomputed map

torgeadelin: added types for the precomputed map

arun3528 (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
marcin-bazyl: Typescript already guarantees that the event name used here is correct, if you do a typo in it, it will flag an error. I see no added value in making a constant or enum that you then have to click and jump to another file to see the value and when you search for it your first result is not in the file where it's actually used and you have to do another search for the constant to find the code. 
We normally use constants to avoid magic values like 1 or 10 in the code that can have many meanings and a constant guarantees that you know which "1" you want to use in case you ever want to change it. Here, these values like 'client.webexapp.launched' are unique and it's not like you will see the same string 'client.webexapp.launched' somewhere else that would mean a different thing.

PR #3442, opened 2024-03-15
Feat/SI: send and receive named media groups reuest

mickelr:

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7917,6 +7926,33 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      throw new Error(`cannot set send named media group which media type is ${mediaType}`);
+    }
+
+    const value = languageCode || this.simultaneousInterpretation.getTargetLanguageCode();
+    let groups = [];
+
+    if (value) {
+      groups = [
+        {
+          type: 1,
```
marcin-bazyl, line 7946: is there some enum or const we can use here instead of "1"?

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -210,28 +212,47 @@ export class RemoteMediaGroup {
     }
   }
 
-  private sendActiveSpeakerMediaRequest(commit: boolean) {
+  public sendActiveSpeakerMediaRequest(commit: boolean) {
```
marcin-bazyl, line 215: This method should remain private, see my other comment about creating setNamedMediaGroup() instead.

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,33 +543,96 @@ export class RemoteMediaManager extends EventsScope {
...
+      throw new Error(`cannot set receive named media group which media type is ${mediaType}`);
+    }
+
+    const value = languageId;
+    if (value === this.config.namedMediaGroup?.value) {
+      return;
+    }
+
+    this.config.namedMediaGroup = {
+      type: 1,
```
marcin-bazyl, line 563: same as previous comment

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,33 +543,96 @@ export class RemoteMediaManager extends EventsScope {
...
     );
   }
 
+  /**
+   * Creates the audio slots for named media
+   */
+  private async createInterpretationAudioMedia(commitRequest: boolean) {
+    // release existing si slot
+    if (this.slots.audio.si) {
+      this.receiveSlotManager.releaseSlot(this.slots.audio.si);
```
marcin-bazyl, line 608: createAudioMedia() only allocates the slots, doesn't release them, but createInterpretationAudioMedia() also releases the existing slot, so this is inconsistent. We should either have a separate create/destroy functions for si or rename this one to "recreate" or something like that? I think I prefer the 1st option.

Actually, after you do the changes from my other comments, you won't need to release the slot at all, right? because si audio will be created once at the start and the same one will be re-used all the time. Unless you want to destroy it completely when user no longer wants to receive any si audio, then we would probably also need to emit a new "media:remoteInterpretationAudio:destroyed" event

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,33 +543,96 @@ export class RemoteMediaManager extends EventsScope {
...
+    }
+    this.media.audio.si?.stop(false);
+    this.media.audio.si = undefined;
+
+    // create slot for interpretation language audio
+    if (this.config.namedMediaGroup?.type === 1 && this.config.namedMediaGroup?.value) {
```
marcin-bazyl, line 614: another hardcoded "1"

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,33 +543,96 @@ export class RemoteMediaManager extends EventsScope {
...
+    if (value === this.config.namedMediaGroup?.value) {
+      return;
+    }
+
+    this.config.namedMediaGroup = {
+      type: 1,
+      value,
+    };
+
+    await this.createInterpretationAudioMedia(false);
```
marcin-bazyl, line 567: I thought we discussed that we don't even need to recreate the RemoteMediaGroup for si audio. If you just add a `setNamedMediaGroup()` method to RemoteMediaGroup and inside it just update the stored named media group and call `this.sendActiveSpeakerMediaRequest()` it will work and client will receive new language on the same stream - see how it's done for example in `setPreferLiveVideo()`.

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
```diff
@@ -210,28 +212,47 @@ export class RemoteMediaGroup {
...
 
-    this.mediaRequestId = this.mediaRequestManager.addRequest(
-      {
-        policyInfo: {
-          policy: 'active-speaker',
-          priority: this.priority,
-          crossPriorityDuplication: false,
-          crossPolicyDuplication: false,
-          preferLiveVideo: !!this.options?.preferLiveVideo,
+    if (this.options.namedMediaGroup?.value) {
```
marcin-bazyl, line 218: why do we need a whole big if and else block for namedMediaGroup? it's just 1 extra parameter, the rest of the request remains the same right? so we could instead just add 1 line:
`namedMediaGroups: this.options.namedMediaGroup?.value ? [this.options?.namedMediaGroup] : undefined,`
right?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -55,6 +57,32 @@ export default class SendSlotManager {
     return slot;
   }
 
+  /**
+   * Allow users to specify 'namedMediaGroups' to indicate which named media group its audio should be sent to.
+   * @param {MediaType} mediaType MediaType of the sendSlot to which the audio stream needs to be send to the media group
+   * @param {[]}namedMediaGroups - Allow users to specify 'namedMediaGroups'.
+   * @returns {void}
+   */
+  public setNamedMediaGroups(mediaType: MediaType, namedMediaGroups: NamedMediaGroup[]) {
```
marcin-bazyl, line 66: how can the client app say that they don't want to be sending any si audio anymore with this API?

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
```diff
@@ -525,33 +543,96 @@ export class RemoteMediaManager extends EventsScope {
     this.mediaRequestManagers.video.commit();
   }
 
+  /**
+   * Sets which named media group need receiving
+   * @param {MediaType} mediaType of the stream
+   * @param {number} languageCode of the stream
+   * @returns {void}
+   */
+  public async setReceiveNamedMediaGroup(mediaType: MediaType, languageId: number) {
```
marcin-bazyl, line 552: how can the client app say with this API that they don't want to receive any si audio anymore?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
mickelr: Call the function with the value member of the namedMediaGroup as zero

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Call the function with the languageId as zero

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
mickelr: Previously I used if else block for considering the case that there may be multi groups , but in current implementation, yes, we can simplify it

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Code changed

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Rmoved destroy part

packages/@webex/plugin-meetings/src/meeting/index.ts
mickelr: Done

packages/@webex/plugin-meetings/src/multistream/remoteMediaGroup.ts
mickelr: Done

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Done

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -121,14 +121,31 @@ const SimultaneousInterpretation = WebexPlugin.extend({
    * Update self's interpretation information (self is interpreter)
    * @param {Object} interpretation
    * @param {String} selfParticipantId
-   * @returns {void}
+   * @returns {bool} is target language changed
    */
   updateSelfInterpretation({interpretation, selfParticipantId}) {
```
marcin-bazyl, line 126: updateSelfInterpretation

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -121,14 +121,31 @@ const SimultaneousInterpretation = WebexPlugin.extend({
...
+    this.set('selfIsInterpreter', true);
+    const preTargetLanguage = this.targetLanguage;
     const {originalLanguage, sourceLanguage, order, isActive, targetLanguage, receiveLanguage} =
       interpretation || {};
     this.set({originalLanguage, sourceLanguage, order, isActive, targetLanguage, receiveLanguage});
     this.set('selfParticipantId', selfParticipantId);
+
+    return !!(targetLanguage && preTargetLanguage !== targetLanguage);
```
marcin-bazyl, line 134: in some other PR comment you mentioned that if user wants to disable the translation they should set the target language to 0, but here if targetLanguage is zero, we return false as if "nothing has changed", but disabling interpretation is a change, isn't it? or am I misunderstanding how this works?

also what's the relationship of isActive with targetLanguage being zero, can you have isActive=true and targetLanguage=0?

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -121,14 +121,31 @@ const SimultaneousInterpretation = WebexPlugin.extend({
...
   },
+
+  /**
+   * Get the language code of the interpreter target language
+   * @returns {number}
+   */
+  getTargetLanguageCode() {
+    if (this.selfIsInterpreter) {
+      return this.siLanguages.get(this.targetLanguage)?.languageCode;
```
marcin-bazyl, line 143: should getTargetLanguageCode() take into account also the value of `this.isActive`?

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
```diff
@@ -144,6 +150,79 @@ describe('RemoteMediaGroup', () => {
...
+
+      assert.calledOnce(fakeMediaRequestManager.addRequest);
+
+      assert.calledWith(
+        fakeMediaRequestManager.addRequest,
+        sinon.match({
+          policyInfo: sinon.match({
+            policy: 'active-speaker',
+            priority: 255,
+            nameMediaGroups: sinon.match.any,
```
marcin-bazyl, line 174: we should check the corrrect values here, not "any"

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -299,6 +300,184 @@ describe('RemoteMediaManager', () => {
...
+      assert.callCount(fakeReceiveSlotManager.allocateSlot, 4);
+
+
+      resetHistory();
+
+
+      remoteMediaManager.setReceiveNamedMediaGroup(MediaType.AudioMain, 28);
+
+      // check that setNamedMediaGroup has been called
+      assert.calledOnce(createdInterpretationAudioGroup.setNamedMediaGroup);
```
marcin-bazyl, line 391: we should also check that the right value is passed in

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -299,6 +300,184 @@ describe('RemoteMediaManager', () => {
...
+          numOfScreenShareStreams: 0,
+        },
+        video: {
+          preferLiveVideo: false,
+          initialLayoutId: 'empty',
+          layouts: {
+            empty: {},
+          },
+        },
+        namedMediaGroup: {type: 1, value: 20},
```
marcin-bazyl, line 318: I don't see a check that the RemoteMediaGroup is created with the right namedMediaGroup value (20)

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
```diff
@@ -299,6 +300,184 @@ describe('RemoteMediaManager', () => {
...
+        fakeMediaRequestManagers,
+        config
+      );
+
+      await remoteMediaManager.start();
+
+      resetHistory();
+
+      try {
+        remoteMediaManager.setReceiveNamedMediaGroup(MediaType.VideoMain, 0);
```
marcin-bazyl, line 474: this test is still passing even if I remove the code that throws from setReceiveNamedMediaGroup(), please use assert.throws() or something similar to test this properly

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
marcin-bazyl: ok, so this needs to be documented in the jsdoc

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
marcin-bazyl: this needs to be documented in the jsdoc

packages/@webex/plugin-meetings/src/interpretation/index.ts
mickelr: In actual implementation, an interpreter will never allowed be changed to normal attendee, which means he will never have chance to speak in the floor. For example, if host assign one attendee as an interpreter who take charge of "English" to "French", the interpreter's target language will always be English, or be French in verse when he click the toggle button, and never be zero.
Host can set up several interpreters for one channel, e.g. "Engish to French", and hand over one by one. The working person is active.

packages/@webex/plugin-meetings/src/interpretation/index.ts
mickelr: I don't think getTargetLanguageCode() should take into account also the value of this.isActive, because targeLanguage is always there no matter is active or not. But you reminded me that we need take into account the value when call setSendNamedMediaGroups()!

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
mickelr: Done

packages/@webex/plugin-meetings/src/multistream/remoteMediaManager.ts
mickelr: Done

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
mickelr: Fixed

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
mickelr: Fixed

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaManager.ts
mickelr: Fixed

packages/@webex/plugin-meetings/test/unit/spec/multistream/remoteMediaGroup.ts
mickelr: Fixed

marcin-bazyl (approved):

JudyZhuHz (approved):
GLTM

PR #3449, opened 2024-03-19
fix: Roap optimizations phase 2

marcin-bazyl:
# COMPLETES #[SPARK-484742](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-484742)

Currently TURN discovery is done only if UDP reachability to public clusters fails and is done as a separate http request to /media Locus API after the join request. If we fail to establish a media connection without TURN discovery, we also retry again with it, but sometimes that can still fail, because in some error conditions Locus requires us to re-join first.

Always do TURN discovery and do it together with the join request to Locus, so there is almost no delay added. I've still kept the code that does TURN discovery as a separate step and also the code to re-try with TURN discovery if initial connection attempt fails without it, but that code should never really be executed now, but probably worth keeping it for now until we confirm that this new approach is better for JMF.

Summary of the code changes:
- added turnServerInfo argument to Meeting.addMedia() and Meeting.establishMediaConnection() - this allows me to call addMedia() with turn server information from joinWithMedia()
- taken out the code that starts TURN discovery as a separate /media Locus API call into a new method Meeting.doTurnDiscovery()
- added reachability to the options that get passed into MeetingUtil.joinMeeting() and Meeting.meetingRequest.joinMeeting() (roapMessage was already there, but we need to also send reachability)
- added types for TurnServerInfo and TurnDiscoveryResult, there are some more types that need to be added, I'll add them when making the proper PR
- did some refactoring in turnDiscovery.ts to extract code that's common between TURN discovery being done as part of join or part of a separate /media call into new functions: generateTurnDiscoveryRequestMessage() and handleTurnDiscoveryHttpResponse()

These changes can be tested with the sample app. Web app currently doesn't call Meeting.joinWithMedia() - the work for that is in progress, draft PR here: https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/5578

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4316,34 +4316,71 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      let turnServerInfo;
+      let turnDiscoverySkippedReason;
 
-        return Promise.reject(error);
-      });
+      // @ts-ignore
+      joinOptions.reachability = await this.webex.meetings.reachability.getReachabilityResults();
+      const turnDiscoveryRequest = await this.roap.generateTurnDiscoveryRequestMessage(this, true);
+
+      turnDiscoverySkippedReason = turnDiscoveryRequest.skippedReason;
```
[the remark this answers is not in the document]
marcin-bazyl: good catch, too bad typescript didn't catch it, but it's because of the wrapping of TurnDiscovery class done in roap/index.ts, but I've fixed that now too so that typescipt catches this now too

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4316,34 +4316,71 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        media: mediaResponse,
+      };
+    } catch (error) {
+      LoggerProxy.logger.error('Meeting:index#joinWithMedia --> ', error);
+
+      let leaveError;
+
+      if (joined) {
+        try {
+          await this.leave({resourceId: joinOptions?.resourceId, reason: 'joinWithMedia failure'});
```
jor-row, line 4361: some other reasons that are passed through to ```this.leave(``` are constants. Do we want this to also be a constant somewhere for consistency?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6130,6 +6167,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
     // @ts-ignore - config coming from registerPlugin
     if (!this.turnServerUsed) {
+      // todo: at some point we can remove this block as now this.turnServerUsed should always be true
```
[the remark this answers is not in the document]
marcin-bazyl: yeah, I've only put the todo comment here for now for this draft PR so people looking at the PR realize this and can comment on it :-)
created jira: SPARK-509313

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: if you follow the existing code closely you'll see that there are constants from different enums being passed in as reason - for example sometimes they are from CALL_REMOVED_REASON, sometimes they are from MEETING_REMOVED_REASON, so it just shows that using constants doesn't guarantee that there will be no mess. The `reason` parameter needs to be typed properly first and then all the calls to leave() need to be fixed and this would be big enough change to be a separate PR on its own, so I don't want to do this as part of this work.

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
jor-row: Nice fix

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4316,34 +4316,71 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+      if (joinOptions.roapMessage) {
+        ({turnServerInfo, turnDiscoverySkippedReason} =
+          await this.roap.handleTurnDiscoveryHttpResponse(this, joinResponse));
+
+        this.turnDiscoverySkippedReason = turnDiscoverySkippedReason;
+        this.turnServerUsed = !!turnServerInfo;
+      }
+
+      const mediaResponse = await this.addMedia(mediaOptions, turnServerInfo);
```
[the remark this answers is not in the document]
jor-row: Ok nice that makes sense, I wouldn't have thought to do that

jor-row (approved):
Changes and tests look good to me, I tested this time using the draft Cantina PR and everything is working as expected

packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -18,6 +20,17 @@ const TURN_DISCOVERY_TIMEOUT = 10; // in seconds
 // and do the SDP offer with seq=1
 const TURN_DISCOVERY_SEQ = 0;
 
+const TURN_DISCOVERY_SKIP_REASON = {
```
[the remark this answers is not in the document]
marcin-bazyl: I've changed these 2:
TURN_DISCOVERY_SKIP_REASON -> TurnDiscoverySkipReason
missing_http_response -> missingHttpResponse

packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -18,6 +20,17 @@ const TURN_DISCOVERY_TIMEOUT = 10; // in seconds
...
+  missing_http_response: 'missing http response', // when we asked for the TURN discover response to be in the http response, but it wasn't there
```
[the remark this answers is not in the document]
marcin-bazyl: fixed

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -260,8 +262,15 @@ export default class MeetingRequest extends StatelessWebexPlugin {
       };
     }
 
-    if (roapMessage) {
-      body.localMedias = roapMessage.localMedias;
+    if (roapMessage || reachability) {
+      body.localMedias = [
+        {
+          localSdp: JSON.stringify({
```
torgeadelin, line 268: why do we include reachability in this now and before we didn't?

packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -10,6 +10,8 @@ import RoapRequest from './request';
 import Meeting from '../meeting';
 import MeetingUtil from '../meeting/util';
 
+type Enum<T extends Record<string, unknown>> = T[keyof T];
```
torgeadelin, line 13: does this just return whatever the value type of the record is? and unknown by default?

packages/@webex/plugin-meetings/test/unit/spec/roap/index.ts
```diff
@@ -32,11 +40,58 @@ describe('Roap', () => {
 
           assert.calledOnceWithExactly(doTurnDiscoveryStub, meeting, isReconnecting, isForced);
           assert.deepEqual(result, RESULT);
+        });
+      });
+    });
+
+    describe('generateTurnDiscoveryRequestMessage', () => {
```
torgeadelin, line 47: missing test case: promise already in progress

packages/@webex/plugin-meetings/test/unit/spec/roap/turnDiscovery.ts
```diff
@@ -676,4 +676,275 @@ describe('TurnDiscovery', () => {
...
+            correlation_id: testMeeting.correlationId,
+            locus_id: FAKE_LOCUS_ID,
+          })
+        );
+
+        checkNextTurnDiscovery();
+      });
+
+      describe('abort', () => {
+        it('allows starting a new TURN discovery', async () => {
```
torgeadelin, line 922: non blocking: missing test case when no in progress defer

torgeadelin (approved):
looking good, some small non-blocking comments

packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
marcin-bazyl: it creates a type that has the possible values of the enum - see https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/How+to+use+enums+and+alternatives

Actually I've just noticed that SDK already has this `Enum` type defined in plugin-meetings/src/constants, so I'll just import it from there

packages/@webex/plugin-meetings/src/meeting/request.ts
marcin-bazyl: because previously we never included any roap message with the join request, now we include the TURN_DISCOVERY_REQUEST Roap message and that requires reachability to be included in the request.

packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -158,18 +179,189 @@ export default class TurnDiscovery {
...
+   */
+  public abort() {
+    if (this.defer) {
+      this.defer.reject(new Error('TURN discovery aborted'));
+      this.defer = undefined;
+    }
+  }
+
+  /**
+   *
```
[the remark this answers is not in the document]
marcin-bazyl: done

PR #3482, opened 2024-03-26
feat(metrics): allow metric fields to be passed to meetings.register()

chburket:
# COMPLETES #[SPARK-510169](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-510169)

Allow metric fields to be passed to meetings.register() to be included in the success and failure metrics.  This is to allow cantina to pass the correlation id when crosslaunched to assist in debugging meetings registration failures.

meetings.register() now accepts an object of metric fields to be included when the success/failure metric is submitting.

### Change Type

- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added tests and tested manually with Cantina.

### I certified that

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -721,11 +721,12 @@ export default class Meetings extends WebexPlugin {
    * Explicitly sets up the meetings plugin by registering
    * the device, connecting to mercury, and listening for locus events.
    *
+   * @param {object} metricFields - optional fields to use when sending metrics
    * @returns {Promise}
    * @public
    * @memberof Meetings
    */
-  public register() {
+  public register(metricFields: object = {}) {
     // @ts-ignore
     if (!this.webex.canAuthorize) {
       LoggerProxy.logger.error(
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -774,7 +775,10 @@ export default class Meetings extends WebexPlugin {
           EVENT_TRIGGERS.MEETINGS_REGISTERED
         );
         this.registered = true;
-        Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MEETINGS_REGISTRATION_SUCCESS);
+        Metrics.sendBehavioralMetric(
+          BEHAVIORAL_METRICS.MEETINGS_REGISTRATION_SUCCESS,
+          metricFields
```
marcin-bazyl, line 780: maybe better to have some sanity check on `metricFields` in case some app accidentally passes some huge object here? although not sure if it's worth it, because if they want to be malicious they can always call webex.internal.metrics.submitClientMetrics directly and send whatever they want

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
arun3528: it terms of functionality i don't see a issue , if someone wants to bombad the service they can by checking our network tab

only thing I am concerned is the documentation should show up in metricsField and it might confuse the developers. do you want to change the name metricFields to _metricFields or something which does make them to pass anything

arun3528 (approved):

chburket:
decided not to do this

PR #3515, opened 2024-04-03
feat(internal-plugin-metrics): added helpers

shnaaz:

Added helper measureLatency and improved saveLatency

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -92,12 +92,33 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
...
    * @param key - key
-   * @param value -value
+   * @param value - value
+   * @param overwrite - overwrite existing value or add it
    * @throws
    * @returns
    */
-  public saveLatency(key: PreComputedLatencies, value: number) {
-    this.precomputedLatencies.set(key, value);
+  public saveLatency(key: PreComputedLatencies, value: number, overwrite = true) {
```
shnaaz, line 100: We may have to expand PreComputedLatencies type or name it differently?

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,48 @@ describe('internal-plugin-metrics', () => {
...
+      afterEach(() => {
+        clock.restore();
+        sinon.restore();
+      });
+      
+      it('checks measureLatency with overwrite false', async () => {
+        const key = 'internal.client.pageJMT';
+        const overwrite = false;
+
+        clock.tick(100);
```
torgeadelin, line 150: i don't think is needed

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,48 @@ describe('internal-plugin-metrics', () => {
...
+        await promise;
+        assert.calledOnce(callbackStub);
+        sinon.assert.calledWith(saveLatencySpy, key, 50, overwrite)       
+      });
+
+      it('checks measureLatency with overwrite true', async () => {
+        const key = 'internal.client.pageJMT';
+        const overwrite = true;
+
+        clock.tick(100);
```
torgeadelin, line 163: not needed

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,48 @@ describe('internal-plugin-metrics', () => {
       assert.deepEqual(cdl.getMeetingInfoReqResp(), 10);
     });
 
+    describe('measureLatency', () => {
+      let clock;
+      let callbackStub;
+      let saveLatencySpy;
+      beforeEach(() => {
```
torgeadelin, line 135: newline

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
torgeadelin: I think it's fine for now, it's still valid -- it stores precomputed latencies

torgeadelin (commented):
looks good, just some small comments

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
       assert.deepEqual(cdl.getMeetingInfoReqResp(), 10);
     });
 
+    describe.only('measureLatency', () => {
```
torgeadelin, line 131: remove only

torgeadelin (commented):
one more

torgeadelin (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
martroos, line 100: nitpick: For future use we may want an `options` object instead of only having overwrite, but not sure what other functionality we'll want so maybe not an issue right now.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -92,12 +92,33 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
...
+   * Measure latency for a request
+   * @param key - key
+   * @param callback - callback for which you would like to measure latency
+   * @param overwrite - overwite existing value or add to it
+   * @returns
+   */
+  public measureLatency(
+    callback: () => Promise<any>,
+    key: PreComputedLatencies,
+    overwrite = false
```
martroos, line 115: nitpick: For future use we may want an `options` object instead of only having overwrite, but not sure what other functionality we'll want so maybe not an issue right now.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+        sinon.restore();
+      });
+      
+      it('checks measureLatency with overwrite false', async () => {
+        const key = 'internal.client.pageJMT';
+        const overwrite = false;
+        callbackStub = sinon.stub().resolves();
+
+        const promise = cdl.measureLatency(callbackStub, 'internal.client.pageJMT', overwrite);
+        clock.tick(50);
```
martroos, line 153: suggestion: Whilst this does happen to work, it is entirely reliant on the fact that nothing else runs between creating the promise, and incrementing the clock. If an `await` is added between these, the test will fail since the clock won't have incremented soon enough.

To avoid this, my suggestion is to have a custom stub implementation which increments the time when called. So you would end up defining the callback as something like:

```js
const callback = sinon.stub().callsFake(() => {
  clock.tick(50);
  return Promise.resolve(resolution);
});
```

This way, the clock is incremented as a part of being called, as though it took some time to run. The `clock.tick` here is no longer needed of course.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+
+        await promise;
```
martroos, line 155: issue: This needs to check the promise resolution is the expected resolution from the callback.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
       assert.deepEqual(cdl.getMeetingInfoReqResp(), 10);
     });
 
+    describe('measureLatency', () => {
+      let clock;
+      let callbackStub;
```
martroos, line 133: nitpick: This probably doesn't need to be a top level `let` since it is explicitly defined in each test.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+      it('checks measureLatency with overwrite false', async () => {
+        const key = 'internal.client.pageJMT';
+        const overwrite = false;
+        callbackStub = sinon.stub().resolves();
+
+        const promise = cdl.measureLatency(callbackStub, 'internal.client.pageJMT', overwrite);
+        clock.tick(50);
+
+        await promise;
+        assert.calledOnce(callbackStub);
```
martroos, line 156: suggestion: Check the callback is called with exact arguments:

```js
assert.calledOnceWithExactly(callback);
```

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+        sinon.assert.calledWith(saveLatencySpy, key, 50, overwrite)
```
martroos, line 157: suggestion: Check this is called once also:

```js
assert.calledOnceWithExactly(saveLatencySpy, key, 50, overwrite)     
```

Note the explicit use of `sinon` isn't needed, this should already be registered with chai.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+
+      it('checks measureLatency when callBack rejects', async () => {
+        const key = 'internal.client.pageJMT';
+        const overwrite = true;
+        callbackStub = sinon.stub().rejects(new Error('some error'));
+
+        const promise = cdl.measureLatency(callbackStub, 'internal.client.pageJMT', overwrite);
+        clock.tick(50);
+
+        await await assert.isRejected(promise);
```
martroos, line 181: issue: This should check the rejection value is the same as the rejection value used by the callback.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -111,6 +128,62 @@ describe('internal-plugin-metrics', () => {
...
+        const promise = cdl.measureLatency(callbackStub, 'internal.client.pageJMT', overwrite);
+        clock.tick(50);
+
+        await promise;
+        assert.calledOnce(callbackStub);
+        sinon.assert.calledWith(saveLatencySpy, key, 50, overwrite)       
+      });
+
+      it('checks measureLatency with overwrite true', async () => {
+        const key = 'internal.client.pageJMT';
```
martroos, line 161: nitpick: I'd like to use at least one other key here, to prove that it isn't just hard coded!

martroos (changes requested):
Mostly just test changes - the comments on the tests apply to multiple `it`s, but I haven't repeated comments everywhere.

martroos (approved):
Would like a couple of tweaks to test, but not worth blocking the PR since the tests do cover the changes.

PR #3549, opened 2024-04-19
feat(metrics): add downloadIntelligenceModelsReqResp to latency metrics

maxinteger:
# COMPLETES [SPARK-503933](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503933)

Implement new downloadIntelligenceModelsReqResp latency for CA metrics

Updated the types and added new utility function called `accumulateLatency`

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -121,6 +122,21 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
...
+   * @returns
+   */
+  public accumulateLatency(callback: () => Promise<unknown>, key: PreComputedLatencies) {
+    const start = performance.now();
+
+    return callback().finally(() => {
+      const currentLatency = this.precomputedLatencies.get(key) ?? 0;
+      this.saveLatency(key, currentLatency + (performance.now() - start), true);
+    });
+  }
```
torgeadelin, line 138: we already support this, see saveLatency overwrite param

torgeadelin (changes requested):
one issue

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
maxinteger: This a special case of the overwrite when we want to add the current value to the next one before overwrite it. 

It is doable with `measureLatency` as well, but we need to access to the current value and I really do not want to use `precomputedLatencies` inside cantina. Actually `precomputedLatencies` should be private.

maxinteger: never mind, I just checked again and the saveLatency does what you described.

The `overwrite` name is a bit misleading, because it not just simple overwrite, but also accumulate values

torgeadelin: measureLatency exposes the overwrite property too, so you don't have to access precomputedLatencies directly in cantina

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -93,31 +94,31 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
    * Store precomputed latency value
    * @param key - key
    * @param value - value
-   * @param overwrite - overwrite existing value or add it
+   * @param accumulate - when it true, it overwrite existing value with sum of the current value and the new measurement otherwise just store the new measurement
```
torgeadelin, line 97: its

torgeadelin: overwrites

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -93,31 +94,31 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
...
    * @throws
    * @returns
    */
-  public saveLatency(key: PreComputedLatencies, value: number, overwrite = true) {
-    const existingValue = overwrite ? 0 : this.precomputedLatencies.get(key) || 0;
+  public saveLatency(key: PreComputedLatencies, value: number, accumulate = true) {
```
torgeadelin, line 101: I've just realised, we've changed the meaning here too.
Should I accumulate? 
  - accumulate = true it overrides.
  - accumulate = false, it accumulates.
  
Also the default for this method should be to overwrite.

torgeadelin: So maybe change the default value and swap the tertiary statement outcomes on line 102.

torgeadelin (changes requested):
one small issue

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
maxinteger: good point, I fixed all of them

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -145,64 +150,66 @@ describe('internal-plugin-metrics', () => {
...
         assert.deepEqual(resolvedValue, 'test123');
         assert.calledOnceWithExactly(callbackStub);
-        assert.calledOnceWithExactly(saveLatencySpy, key, 20, overwrite)       
+        assert.calledOnceWithExactly(saveLatencySpy, key, 20, accumulate);
       });
 
       it('checks measureLatency when callBack rejects', async () => {
         const key = 'internal.client.pageJMT';
-        const overwrite = true;
+        const accumulate = true;
```
torgeadelin, line 200: in order to keep the test the same as before, we should say accumulate = false, as before it was overwrite = true

torgeadelin (commented):
some more comments, but non really blocking.
also,  there's a lot of linting changes, not sure if we should add those

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -51,29 +51,29 @@ describe('internal-plugin-metrics', () => {
...
     });
 
-    it('should save latency correctly when overwrite is false', () => {
+    it('should save latency correctly when accumulate is true', () => {
       assert.deepEqual(cdl.precomputedLatencies.size, 0);
-      cdl.saveLatency('internal.client.pageJMT', 10, false);
+      cdl.saveLatency('internal.client.pageJMT', 10, true);
       assert.deepEqual(cdl.precomputedLatencies.size, 1);
       assert.deepEqual(cdl.precomputedLatencies.get('internal.client.pageJMT'), 10);
     });
```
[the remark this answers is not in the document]
maxinteger: It looks good to me. It tests the case when there is no current value. Here we can end up with `NaN` when the missing value not handled correctly.

maxinteger:
@torgeadelin I think the linting is coming from a recently merged PR, there is not too many and all related to metrics plugin so I think it is OK

packages/@webex/internal-plugin-metrics/test/unit/spec/prelogin-metrics-batcher.ts
```diff
@@ -149,7 +150,8 @@ describe('internal-plugin-metrics', () => {
       });
 
       it('fails if preLoinId is not set', async () => {
-        webex.internal.newMetrics.callDiagnosticMetrics.preLoginMetricsBatcher.preLoginId = undefined;
+        webex.internal.newMetrics.callDiagnosticMetrics.preLoginMetricsBatcher.preLoginId =
+          undefined;
```
shnaaz, line 154: Why this change?

packages/@webex/internal-plugin-metrics/test/unit/spec/prelogin-metrics-batcher.ts
```diff
@@ -177,9 +179,10 @@ describe('internal-plugin-metrics', () => {
 
     describe('prepareItem', () => {
       it('calls prepareDiagnosticMetricItem correctly', async () => {
-         // avoid setting .sent timestamp
-        webex.internal.newMetrics.callDiagnosticMetrics.preLoginMetricsBatcher.prepareRequest = (q) =>
-          Promise.resolve(q);
+        // avoid setting .sent timestamp
+        webex.internal.newMetrics.callDiagnosticMetrics.preLoginMetricsBatcher.prepareRequest = (
+          q
+        ) => Promise.resolve(q);
```
shnaaz, line 185: Again why has this changed

shnaaz (commented):
Are the changes in indentation intentional? Seems like some are but some are not?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
shnaaz: This PR was created for adding downloadIntelligenceModelsReqResp, I strongly feel that changing any util should have been a separate PR so that it could be focused and tested. I will still recommend to create a different PR for this.

shnaaz (changes requested):
Please check latest comment

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -93,31 +95,31 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
...
+   * @param key - key
+   * @param accumulate - when it is true, it overwrites existing value with sum of the current value and the new measurement otherwise just store the new measurement
    * @returns
    */
   public measureLatency(
-    callback: () => Promise<any>,
+    callback: () => Promise<unknown>,
     key: PreComputedLatencies,
-    overwrite = false
+    accumulate = false
```
shnaaz, line 117: Are you planning to change definition in Cantina too, I think we should

maxinteger: Yes, I already have the PR for that
https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/5653

maxinteger:
@shnaaz I just updated my branch to the latest beta and run formatting, probably these lines where not correctly formatted.

I can roll it back if you want to.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
shnaaz: We don't have accumulateLatency anymore right?

shnaaz (approved):
Just roll back couple of linter changes we discussed, otherwise looks good

PR #3555, opened 2024-04-22
feat(metrics): Add client.media.render.start event

johnsoter13:
# COMPLETES #<[https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-518038](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-518038)>

- There was a missing event that caused 100% receiver side failures for CA analysis of share flow

< DESCRIBE YOUR CHANGES >
- Added missing receiver event for share client.media.render.start, without it, CA marks all receiver flows as failures

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7702,6 +7702,18 @@ export default class Meeting extends StatelessWebexPlugin {
         ) {
           layoutInfo.content = {width: contentWidth, height: contentHeight};
         }
+
+        // @ts-ignore
+        this.webex.internal.newMetrics.submitClientEvent({
```
marcin-bazyl, line 7707: This place here is only applicable to transcoded meetings. For multistream meetings, the layout is changed via RemoteMediaManager.setLayout() method. Also it doesn't feel right to be sending "client.media.render.start" at the point when we switch layout. Probably at this point we should be sending "client.share.layout.displayed" and "client.media.render.start" should I guess be sent when first frame is decoded and displayed. This will be tricky for SDK to know, because it doesn't have a reference to the video element that renders the stream and it only polls getStats() API every 5s so it can't immediately detect when frames are decoded - it may detect it up to 5s later. So maybe as a compromise we could send it when the share ReceiveSlot's sourceState changes to "live", but it still poses some challenges: 
- in theory there can be many share receive slots, which one should send it? the first one or all of them?
- ReceiveSlots are not used for transcoded meetings/calls

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8630,34 +8652,6 @@ describe('plugin-meetings', () => {
 
           checkParseMeetingInfo(expectedInfoToParse);
         });
-
-        it('should parse meeting info, set values, and return null when permissionToken is not present', () => {
```
marcin-bazyl, line 8634: why is this unrelated test being removed as part of this PR? is this some merge gone wrong?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6040,6 +6040,20 @@ export default class Meeting extends StatelessWebexPlugin {
         EVENT_TRIGGERS.MEETING_MEDIA_REMOTE_STARTED,
         data
       );
+      if (data.type === 'share') {
```
marcin-bazyl, line 6043: nitpick: not sure if it matters, but ideally we should be sending "client.media.render.start" after "client.media.rx.start", right? so this block should be 11 lines down.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6040,6 +6040,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        // @ts-ignore
+        this.webex.internal.newMetrics.submitClientEvent({
+          name: 'client.media.render.start',
```
marcin-bazyl, line 6046: BTW, there is also a "client.media.render.stop" event in the event dictionary. Is it OK that we're not sending it at all?

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: CA organizes the events as they come in, but I will change it to match the desired flow of events for readability

packages/@webex/plugin-meetings/src/meeting/index.ts
johnsoter13: Yes it is ok, Desktop barely sends this event. But you're right, there's no reason not to send in this case.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
johnsoter13: I didn't remove this? I'll see what happened here

johnsoter13: Ok I must've command + z'd this test out of the file. Good catch!

marcin-bazyl (approved):

arun3528 (approved):

PR #3562, opened 2024-04-24
fix(plugin-meetings): fixed unittests after effect lib update

shnaaz:
Web media effect library update (https://github.com/webex/webex-js-sdk/pull/3557/files) caused couple of unittests to fail
This PR fixes those tests

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
+++ b/packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
@@ -413,7 +413,13 @@ describe('plugin-meetings', () => {
           assert.instanceOf(result, NoiseReductionEffect);
           assert.containsAllKeys(result, ['audioContext', 'isEnabled', 'isReady', 'options']);
           assert.equal(result.options.authToken, 'fake_token');
-          assert.deepEqual(result.options.audioContext, {});
+          assert.deepEqual(result.options, {
+            audioContext: {},
+            authToken: 'fake_token',
+            mode: 'WORKLET',
+            env: 'prod',
+            avoidSimd: false,
+          });
           assert.exists(result.enable);
           assert.exists(result.disable);
           assert.exists(result.dispose);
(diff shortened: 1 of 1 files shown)


packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -412,10 +412,8 @@ describe('plugin-meetings', () => {
           assert.exists(result);
           assert.instanceOf(result, NoiseReductionEffect);
           assert.containsAllKeys(result, ['audioContext', 'isEnabled', 'isReady', 'options']);
-          assert.deepEqual(result.options, {
-            authToken: 'fake_token',
-            audioContext: {},
-          });
+          assert.equal(result.options.authToken, 'fake_token');
+          assert.deepEqual(result.options.audioContext, {});
```
martroos, line 416: suggestion: Since the next test will fail if the media effects library changes what is in the options (so that test will need fixing), we may as well do a full check of the options here also, adding in the missing items.

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -426,6 +424,7 @@ describe('plugin-meetings', () => {
             audioContext: {},
             mode: 'WORKLET',
             env: 'prod',
+            avoidSimd: false,
```
martroos, line 427: suggestion: These are all the current default values - in order to prove that the custom options are being used, they should be non-default values.

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
marcin-bazyl: another thing we could do is to just modify the next test to just check the individual options fields that we pass. It seems weird that we're verifying all the options when they actually belong to the effects lib and are controlled by that lib and they can keep changing them.

sreenara (approved):

PR #3585, opened 2024-05-02
fix: wrong type used inside LocalCameraStreamEvents

marcin-bazyl:
small fix for types


packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -32,12 +32,10 @@ export type ServerMuteReason =
   | 'clientRequestFailed' // client called setUserMuted() but server request failed
   | 'localUnmuteRequired'; // server forced the client to be unmuted
 
-// these events are in addition to WCME events. This will be properly typed once webrtc-core event types inheritance is fixed
```
marcin-bazyl, line 35: I've removed these comments, because they are out-of-date, the types inheritance is already fixed in webrtc-core

marcin-bazyl:
closing as the same fix is already done in https://github.com/webex/webex-js-sdk/pull/3584

PR #3569, opened 2024-04-25
fix(plugin-meetings): add callingServiceType to MQE

SomeBody16:
# COMPLETES [SPARK-510208](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-510208)

## This pull request addresses

Missing `callingServiceType` in MQE which is always `LOCUS` for `webex-js-sdk`

## by making the following changes

Adds `callingServiceType` to MQE

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [ ] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3593, opened 2024-05-10
fix(internal-plugin-metrics): add callingServiceType to MQE

SomeBody16:
# COMPLETES [SPARK-510208](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-510208)

Missing `callingServiceType` in MQE which is always `LOCUS` for `webex-js-sdk`

It's a follow to PR #3569 (previously to beta branch)

Adds `callingServiceType` to MQE

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -1326,8 +1326,3 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
-
-// constant for statsAnalyzer report.callingServiceType
-export enum CALLING_SERVICE_TYPE {
-  TYPE_LOCUS = 'LOCUS',
-}
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -1326,3 +1326,6 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
+
+// constant for statsAnalyzer report.callingServiceType
+export const CALLING_SERVICE_TYPE = 'LOCUS' as const;
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -1328,4 +1328,6 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
 
 // constant for statsAnalyzer report.callingServiceType
-export const CALLING_SERVICE_TYPE = 'LOCUS' as const;
+export enum CALLING_SERVICE_TYPE {
+  TYPE_LOCUS = 'LOCUS',
+}
--- a/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
+++ b/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
@@ -1,5 +1,3 @@
-import {CALLING_SERVICE_TYPE} from '../constants';
-
 export const emptyMqaInterval = {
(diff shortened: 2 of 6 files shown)


packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1316,3 +1316,6 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
+
+// constant for statsAnalyzer report.callingServiceType
+export const CALLING_SERVICE_TYPE = 'LOCUS' as const;
```
arun3528, line 1321: not sure if the calling service is a right name , isn't locus a signaling server ?

or calling service type indicate what service is sending metrics ?

arun3528 (approved):

packages/@webex/plugin-meetings/src/constants.ts
SomeBody16: It's specification for MQE requested [here](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-510208)
calls are LOCUS-based and that's where name is from. I can change it to `SIGNALING_SERVICE_TYPE` if you want but later we would still need to do `callingServiceType: SIGNALING_SERVICE_TYPE`

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1322,3 +1322,6 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
...
```
brycetham, line 1327: Should we change this to an enum and list out all the available options in [CallingServiceType.raml](https://sqbu-github.cisco.com/WebExSquared/event-dictionary/blob/05fb283bf3d11cd3a684512bbec3967e5489ac29/types/CallingServiceType.raml)?

I know we will only be setting this as `LOCUS` for multistream meetings but I think there are a few benefits to having all the options listed out:

1. It feels more complete.
2. It provides more context in terms of what these values mean (i.e. seeing the other options makes it easier to understand what `LOCUS` means here).
3. It makes it easier to change use different values in the future. For example, Webex Calling will be using this soon, and they would probably want to use the `WEBEXCALLING` option.

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
```diff
@@ -1,3 +1,5 @@
+import {_UNKNOWN_, CALLING_SERVICE_TYPE} from '../constants';
```
brycetham, line 1: Maybe I'm missing something but why does `_UNKNOWN_` need to be imported here?

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1322,3 +1322,8 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
+
+// constant for statsAnalyzer report.callingServiceType
+export enum CALLING_SERVICE_TYPE {
+  TYPE_LOCUS = 'LOCUS',
```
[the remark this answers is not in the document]
brycetham: Oh that's surprising. I tried running some sample code in TypeScript Playground, it seemed to work for me. What's the error message here?

I noticed a lot of other variables in this file use `const` instead of `enum`. Would doing that fix this issue? (As an aside, there seems to be a lot of debate in the TypeScript community about using enums vs consts, I don't have a strong opinion but seems like people in general prefer avoiding enums?)

And I guess another thing... I feel like we should include 'CUCM', 'WEBEXCALLING', and 'BROADWORKS' here, even if we're not using them, for the reason I described in my previous comment: it provides more context in terms of what these options mean.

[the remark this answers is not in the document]
antsukanova: I would add about enums, that if you're not using CALLING_SERVICE_TYPE as a type, I agree that it's not so useful to keep it as enum. I prefer keeping enum as a way to utilize duplication for types and values, but mostly for types, so it then should be inside type file, not a constants

PR #3599, opened 2024-05-14
fix(plugin-meetings): remove logs and circular dependency with voicea

mkesavan13:
# PARTIALLY COMPLETES [SPARK-515520](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-515520) and FULLY COMPLETES [SPARK-523031](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-523031)

- PII Information is being logged from Voicea
- There are Circular Dependency between Voicea and Meetings because of using TriggerProxy in Voicea

- Removed Circular Dependency by getting rid of TriggerProxy usage in Voicea
- In plugin-meetings, again got rid of TriggerProxy and started logging in place

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

- Tested all the Voicea-based events for Closed Captions
- Ensured sample app is working fine for the newest changes and that the user-facing events are emitted appropriately
- Checked logs through Sample applications for any of the Voicea events not logging any Payload/PII information

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-voicea/src/voicea.ts
+++ b/packages/@webex/internal-plugin-voicea/src/voicea.ts
@@ -1,5 +1,4 @@
 import uuid from 'uuid';
-import {TriggerProxy as Trigger} from '@webex/plugin-meetings';
 import {WebexPlugin, config} from '@webex/webex-core';
 
 import {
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -52,6 +52,7 @@ import {
 import {StatsAnalyzer, EVENTS as StatsAnalyzerEvents} from '../statsAnalyzer';
 import NetworkQualityMonitor from '../networkQualityMonitor';
 import LoggerProxy from '../common/logs/logger-proxy';
+import EventsUtil from '../common/events/util';
 import Trigger from '../common/events/trigger-proxy';
```
Kesari3008, line 56: Won't we also remove this import here or is still getting used in this file? If it is still being used, we are not planning to remove it from other places?

Kesari3008 (commented):
One minor comment
Please update the test scenarios tested in the description

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: No, @Kesari3008. We won't remove it from the file because it's being used by other methods. 

The reason why we got rid of `TriggerProxy` inside the `plugin-meetings` for Voicea alone is that, TriggerProxy logs the payload and for Voicea while doing so, it lets out PII information. To avoid that, we log separately and emit event from inside iteslf. While logging, we don't log the payload for any of the Voicea events.

Kesari3008 (approved):
Would be good if we can add logs or screenshot of testing

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -110,90 +109,55 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
...
-          },
-          EVENT_TRIGGERS.NEW_CAPTION,
-          {
-            isFinal: false,
-            transcriptId: voiceaPayload.transcript_id,
-            transcripts: voiceaPayload.transcripts,
-          }
-        );
+        // @ts-ignore
+        this.trigger(EVENT_TRIGGERS.NEW_CAPTION, {
```
Shreyas281299, line 113: Cant we add trigger in the VoiceaChannel class to avoid using @ts-ignore in the file .

Shreyas281299 (approved):
Changes look good. Please see the comment.

packages/@webex/internal-plugin-voicea/src/voicea.ts
mkesavan13: No. I don't think it can be done. We have to use AmpState if we have to as it extends from AmpState

sreenara (commented):
Question. Which object has the `trigger()` function?

mkesavan13:
> Question. Which object has the `trigger()` function?

@sreenara - The `trigger()` and `emit()` comes from AmpState plugin

marcin-bazyl (approved):

PR #3603, opened 2024-05-16
Catalog failure handling

Coread:

If the service catalog fails to initialize for whatever reason, e.g. network failure, the consuming application doesn't know

Added a new attribute on the services called initFailed which is set if the catalog fails to init in the on webex ready handler. Consuming applications can either check webex.internal.services.initFailed directly or can listen for the change:initFailed event to spot when this occurs.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/lib/services/services.js
+++ b/packages/@webex/webex-core/src/lib/services/services.js
@@ -52,7 +52,7 @@ const Services = WebexPlugin.extend({
 
   props: {
     validateDomains: ['boolean', false, true],
-    initFailed: 'boolean',
+    initFailed: ['boolean', false, true],
   },
 
   _catalogs: new WeakMap(),
@@ -1011,7 +1011,6 @@ const Services = WebexPlugin.extend({
     // to update the service catalogs
     this.listenToOnce(this.webex, 'ready', () => {
       const {supertoken} = this.webex.credentials;
-
       // Validate if the supertoken exists.
       if (supertoken && supertoken.access_token) {
         this.initServiceCatalogs()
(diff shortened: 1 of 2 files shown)


packages/@webex/webex-core/test/integration/spec/services/services.js
```diff
@@ -795,8 +795,8 @@ describe('webex-core', () => {
     describe('#fetchClientRegionInfo()', () => {
       it('returns client region info', () =>
         services.fetchClientRegionInfo().then((r) => {
-          assert.isDefined(r.countryCode);
-          assert.isDefined(r.timezone);
+          assert.isDefined(r.regionCode);
```
Coread, line 798: This test was failing due to an unrelated change in the response from the service

packages/@webex/webex-core/test/unit/spec/services/services.js
```diff
@@ -72,6 +139,12 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#initFailed', () => {
```
marcin-bazyl, line 142: we're missing a test that `initFailed===false` when initialization is successful

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -52,6 +52,7 @@ const Services = WebexPlugin.extend({
 
   props: {
     validateDomains: ['boolean', false, true],
+    initFailed: ['boolean', false, true],
```
marcin-bazyl, line 55: if I'm reading this correctly, the `false, true` bit means that it's not required and has a default `true` - is that right? so we say that init failed before init is attempted? That doesn't feel right....

packages/@webex/webex-core/test/unit/spec/services/services.js
Coread: added the test

packages/@webex/webex-core/src/lib/services/services.js
Coread: yeah it wasn't right. I changed `initFailed: 'boolean'` to add the default sequence after writing the check boolean test as it was returning undefined.. and got the default the wrong way round. Tests now cover this.

marcin-bazyl (approved):

PR #3606, opened 2024-05-17
fix: add reachability checks for TLS

k-wasniowski:
# COMPLETES [WEBEX-384657](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-384657)

Currently, we are doing only reachability checks for UDP and TCP.

This PR introduces a reachability check also for TurnTLS. It will allow checking before the call if client is capable of reaching to our turn servers. TurnTLS will be marked as experimental.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested with the JS SDK Sample app.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
@@ -204,7 +204,7 @@ export class ClusterReachability {
    * @returns {boolean} true if we have all results, false otherwise
    */
   private haveWeGotAllResults(): boolean {
-    return ['udp', 'tcp'].every(
+    return ['udp', 'tcp', 'xtls'].every(
       (protocol) =>
         this.result[protocol].result === 'reachable' || this.result[protocol].result === 'untested'
     );
--- a/packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
@@ -2,7 +2,7 @@ import {Defer} from '@webex/common';
 
 import LoggerProxy from '../common/logs/logger-proxy';
 import {ClusterNode} from './request';
-import {convertStunUrlToTurn} from './util';
+import {convertStunUrlToTurn, convertStunUrlToTurnTls} from './util';
 
 import {ICE_GATHERING_STATE, CONNECTION_STATE} from '../constants';
 
@@ -21,6 +21,7 @@ export type ClusterReachabilityResult = {
   udp: TransportResult;
   tcp: TransportResult;
   xtls: TransportResult;
+  tls: TransportResult;
 };
 
 /**
@@ -29,6 +30,7 @@ export type ClusterReachabilityResult = {
 export class ClusterReachability {
   private numUdpUrls: number;
   private numTcpUrls: number;
+  private numTlsUrls: number;
   private result: ClusterReachabilityResult;
   private pc?: RTCPeerConnection;
   private defer: Defer; // this defer is resolved once reachability checks for this cluster are completed
@@ -46,6 +48,7 @@ export class ClusterReachability {
     this.isVideoMesh = clusterInfo.isVideoMesh;
     this.numUdpUrls = clusterInfo.udp.length;
     this.numTcpUrls = clusterInfo.tcp.length;
+    this.numTlsUrls = clusterInfo.xtls.length;
 
     this.pc = this.createPeerConnection(clusterInfo);
 
@@ -55,6 +58,9 @@ export class ClusterReachability {
         result: 'untested',
       },
       tcp: {
+        result: 'untested',
+      },
+      tls: {
         result: 'untested',
       },
       xtls: {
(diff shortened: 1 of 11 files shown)


packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -22,3 +22,27 @@ export function convertStunUrlToTurn(stunUrl: string, protocol: 'udp' | 'tcp') {
 
   return url.toString();
 }
+
+/**
+ * Converts a stun url to a turn url
```
marcin-bazyl, line 27: the comment should say "to turns url"

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -22,3 +22,27 @@ export function convertStunUrlToTurn(stunUrl: string, protocol: 'udp' | 'tcp') {
...
+ *
+ * @param {string} stunUrl url of a stun server
+ * @param {'tcp'|'udp'} protocol what protocol to use for the turn server
+ * @returns {string} url of a turn server
+ */
+export function convertStunUrlToTurnTls(stunUrl: string, protocol: 'udp' | 'tcp') {
```
marcin-bazyl, line 33: I guess it doesn't make sense to have the protocol argument here, because you can't have TLS over UDP when talking to a TURN server, right?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -239,7 +254,8 @@ export class ClusterReachability {
         }
 
         if (e.candidate.type === CANDIDATE_TYPES.RELAY) {
-          this.storeLatencyResult('tcp', this.getElapsedTime());
+          const protocol = e.candidate.port === TURN_TLS_PORT ? 'tls' : 'tcp';
```
marcin-bazyl, line 257: this code here relies on the assumption that we will always get TLS URLs with port 443 from Orpheus, I guess it's fine, but maybe it would be better for us to make sure that it is always 443 and set the port to 443 inside `convertStunUrlToTurnTls()`?

packages/@webex/plugin-meetings/src/reachability/util.ts
k-wasniowski: Updated, thanks for catching it.

packages/@webex/plugin-meetings/src/reachability/util.ts
k-wasniowski: That seems right, removed.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
k-wasniowski: Done

k-wasniowski:
There are many changes made by auto-formatter on save. Should I remove those changes, or is it fine to leave it?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -50,8 +49,26 @@ describe('ClusterReachability', () => {
...
+        },
+        {
+          username: 'webexturnreachuser',
+          credential: 'webexturnreachpwd',
+          urls: ['turn:tcp2.webex.com:5004?transport=tcp'],
+        },
+        {
+          username: 'webexturnreachuser',
+          credential: 'webexturnreachpwd',
+          urls: ['turns:xtls1.webex.com?transport=tcp'],
```
marcin-bazyl, line 65: `convertStunUrlToTurnTls()` always sets the port to 443, but this url doesn't have the port set to 443. It should have it, right?

k-wasniowski: The tricky thing is that, setting port for URL is not working if protocol is not http/https. I decided to leave this port setting to 443, but in general it won't change the current value. (Not a big issue, we are always getting xTLS with port attached).

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -22,3 +22,25 @@ export function convertStunUrlToTurn(stunUrl: string, protocol: 'udp' | 'tcp') {
...
+  // stunUrl looks like this: "stun:external-media1.public.wjfkm-a-15.prod.infra.webex.com:443"
+  // and we need it to be like this: "turns:external-media1.public.wjfkm-a-15.prod.infra.webex.com:443?transport=tcp"
+  const url = new URL(stunUrl);
+
+  if (url.protocol !== 'stun:') {
+    throw new Error(`Not a STUN URL: ${stunUrl}`);
+  }
+
+  url.protocol = 'turns:';
+  url.port = '443'; // Make sure that the port is 443
```
k-wasniowski, line 42: Just general comment, URL class is not changing port properly here because for some reason it's not working properly if protocol is not http/https.
Wondering if we should leave it anyway, just for informational purposes, or delete this line.

marcin-bazyl: ok, maybe remove it then so we don't have the confusion in unit tests. So looks like the only way to guarantee this to work would be to not use the URL class and just do string manipulations, probably not worth it as the server always sends us the port anyway

packages/@webex/plugin-meetings/test/unit/spec/reachability/index.ts
```diff
@@ -288,15 +315,18 @@ describe('gatherReachability', () => {
     assert.calledWith(clusterReachabilityCtorStub, 'cluster 2', {
       udp: ['udp2.1', 'udp2.2'],
       tcp: [],
-      xtls: ['xtls2.1', 'xtls2.2'],
+      xtls: [],
```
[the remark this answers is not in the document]
k-wasniowski: Done

packages/@webex/plugin-meetings/src/reachability/util.ts
k-wasniowski: Removed it for now

marcin-bazyl (approved):

PR #3609, opened 2024-05-22
fix(calling): wdm deregister method

mkesavan13:
# COMPLETES [SPARK-509669](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-509669)

2 device registrations occur in the Calling SDK.

- `calling.register` which calls the `internal.device.register` from `webex/calling.js`
- `line.register` which does the Mobius line registration

While we have the above two methods for deregistering, we do not have a deregister method only for `line` and not for `calling` which does the WDM device deletion operation

- This PR introduces that method in `webex/calling.js` - The Calling class such that, users can do this deregister upon successful deregistering of line
- Or, they can call `calling.deregister()` method alone which will deregister all the active lines, disconnect mercury and delete WDM device as well

## Recording
In the below recordings, you can see that the WDM device and Mercury registration doesn't happen before this change and they happen in the after-change screen recording,

### Before:

https://github.com/webex/webex-js-sdk/assets/8044050/e7d7bb7a-c03f-4eac-8256-637a4f123419

#### calling.deregister cleans up active lines automatically and deletes the WDM registration

https://github.com/webex/webex-js-sdk/assets/8044050/b0af4d86-cac5-4a56-b9ae-6d6e84b8437b 

#### calling.deregister just deletes the WDM registration

https://github.com/webex/webex-js-sdk/assets/8044050/2eb4af9f-e965-42b8-a7c6-5a0889e25da1

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tested if the device registers and deregisters
- Tried registering and deregistering several times to see if the "User has excessive device registrations" error pops up and it doesn't now

### I certified that

---

### DIFF
--- a/packages/webex/src/calling.js
+++ b/packages/webex/src/calling.js
@@ -85,7 +85,7 @@ class Calling extends EventEmitter {
     for (const lineId of lineIds) {
       const line = lines[lineId];
       if (line.getStatus() === 'active') {
-        const thisLinePromise = new Promise((resolve, reject) => {
+        const thisLinePromise = new Promise((resolve) => {
           line.deregister();
           line.on('unregistered', () => {
             this.log.info(
@@ -94,9 +94,6 @@ class Calling extends EventEmitter {
             );
             resolve();
           });
-          setTimeout(() => {
-            reject();
-          }, 30000);
         });
         promises.push(
           thisLinePromise.catch(() => {
--- a/packages/webex/src/calling.js
+++ b/packages/webex/src/calling.js
@@ -77,14 +77,43 @@ class Calling extends EventEmitter {
       return Promise.resolve();
     }
 
-    return (
+    const promises = [];
+
+    const lineIds = Object.keys(this.callingClient?.getLines());
+
+    const lines = lineIds.length ? this.callingClient.getLines() : {};
+    for (const lineId of lineIds) {
+      const line = lines[lineId];
+      if (line.getStatus() === 'active') {
+        const thisLinePromise = new Promise((resolve, reject) => {
+          line.deregister();
+          line.on('unregistered', () => {
+            this.log.info(
+              `line.deregister() successful for the line with ID: ${lineId}`,
+              logContext
+            );
+            resolve();
+          });
+          setTimeout(() => {
+            reject();
+          }, 30000);
+        });
+        promises.push(
+          thisLinePromise.catch(() => {
+            this.log.warn(`Timeout during line.deregister() for ${line}`, logContext);
+          })
+        );
+      }
+    }
+
+    promises.push(
       // @ts-ignore
       this.webex.internal.mercury
         .disconnect()
-        // @ts-ignore
         .then(() => {
           this.log.info('Authentication: webex.internal.mercury.disconnect successful', logContext);
 
+          // @ts-ignore
           return this.webex.internal.device.unregister();
         })
         .then(() => {
@@ -98,6 +127,8 @@ class Calling extends EventEmitter {
           );
         })
     );
+
+    return Promise.all(promises);
   }
 
   async initializeClients() {
--- a/packages/webex/src/calling.js
+++ b/packages/webex/src/calling.js
@@ -77,33 +77,17 @@ class Calling extends EventEmitter {
       return Promise.resolve();
     }
 
-    const promises = [];
-
     const lineIds = Object.keys(this.callingClient?.getLines());
 
     const lines = lineIds.length ? this.callingClient.getLines() : {};
     for (const lineId of lineIds) {
       const line = lines[lineId];
       if (line.getStatus() === 'active') {
-        const thisLinePromise = new Promise((resolve) => {
-          line.deregister();
-          line.on('unregistered', () => {
-            this.log.info(
-              `line.deregister() successful for the line with ID: ${lineId}`,
-              logContext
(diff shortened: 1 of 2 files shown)


packages/webex/src/calling.js
```diff
@@ -68,6 +70,29 @@ class Calling extends EventEmitter {
...
+  async deregister() {
+    if (!this.registered) {
+      this.log.info('Authentication: deregister already done', logContext);
+
+      return Promise.resolve();
+    }
+
+    return (
+      // @ts-ignore
+      this.webex.internal.mercury
```
Kesari3008, line 82: Do we need to disconnect mercury as well with WDM deregistration?

packages/webex/src/calling.js
```diff
@@ -68,6 +70,29 @@ class Calling extends EventEmitter {
       });
   }
 
+  async deregister() {
+    if (!this.registered) {
+      this.log.info('Authentication: deregister already done', logContext);
```
Kesari3008, line 75: These logs should reflect which deregister we are doing. We should update the test to say webex.internal.device.unregister succesful

packages/webex/src/calling.js
mkesavan13: We connect mercury while registering. So, it made sense to get rid of the mercury connection as well every time we deregister.

packages/webex/src/calling.js
mkesavan13: Done. Updated the logs appropriately. Please check.

packages/webex/src/calling.js
```diff
@@ -68,6 +70,67 @@ class Calling extends EventEmitter {
...
+
+      return Promise.resolve();
+    }
+
+    const promises = [];
+
+    const lineIds = Object.keys(this.callingClient?.getLines());
+
+    const lines = lineIds.length ? this.callingClient.getLines() : {};
+    for (const lineId of lineIds) {
```
Kesari3008, line 85: Why can't we run the loop directly on the lines array? Why do we need to maintain another array of lineIds?

Kesari3008 (changes requested):

packages/webex/src/calling.js
```diff
@@ -68,6 +70,67 @@ class Calling extends EventEmitter {
...
+        const thisLinePromise = new Promise((resolve, reject) => {
+          line.deregister();
+          line.on('unregistered', () => {
+            this.log.info(
+              `line.deregister() successful for the line with ID: ${lineId}`,
+              logContext
+            );
+            resolve();
+          });
+          setTimeout(() => {
```
sreenara, line 97: We won't need a timeout and reject here, once a deregister is called, assume that the user doesn't care about the result anymore.

sreenara (changes requested):
Just one change needed regarding timeout

packages/webex/src/calling.js
mkesavan13: - lines isn't an array. It is an object that contains `{ lineId: lineObject }`
- When we do a for..in loop, it would go through prototype keys as well. Linter fails
- That is why we had to get lineIds in an array and then use it.

packages/webex/src/calling.js
```diff
@@ -68,6 +70,67 @@ class Calling extends EventEmitter {
...
+
+    const promises = [];
+
+    const lineIds = Object.keys(this.callingClient?.getLines());
+
+    const lines = lineIds.length ? this.callingClient.getLines() : {};
+    for (const lineId of lineIds) {
+      const line = lines[lineId];
+      if (line.getStatus() === 'active') {
+        const thisLinePromise = new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
mkesavan13: Because the line deregister can be confirmed as success only after receiving the success event. Therefore, instead of asking the user to wait for deregister of all the lines, we wait for event and resolve it. 

When we do this the `await Calling.deregister()` will wait until all lines are deregistered, mercury is disconnected and WDM device is deleted

packages/webex/src/calling.js
mkesavan13: Ah. I agree. Makes sense. Let me remove it.

packages/webex/src/calling.js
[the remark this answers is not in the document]
mkesavan13: It just occurs to me, even here, we can assume that the user doesn't care about results anymore. Right?

@sreenara

sreenara (approved):

packages/webex/src/calling.js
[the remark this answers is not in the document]
sreenara: Yes. We don't have to care about preserving status of response and handling it anywhere because this is the end of the workflow. @mkesavan13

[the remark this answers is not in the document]
mkesavan13: Removed the promise mechanism as well

packages/webex/src/calling.js
Kesari3008: getLines() API returns Record<string, ILine> so similar to how we are doing Object.keys() to fetch the keys which is lineId, we can do Object.values() which will return line array and then we can run for loop in the line array returned.

Kesari3008 (approved):
Please take care of the last comment and check why integration test is failing

PR #3610, opened 2024-05-22
refactor(plugin-meetings): conditional stream addition for MQA

edvujic:
# COMPLETES [WEBEX-378949](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-378949)

This pull request introduces enhancements to MQA, particularly focusing on the logic that determines which streams are sent to the MQA based on their `isRequested` status.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that


mkesavan13 (changes requested):
Changes mostly look good. Just a few comments on the test

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
```diff
@@ -1617,7 +1630,7 @@ describe('plugin-meetings', () => {
               requestedFrames: 0,
               rtpPackets: 0,
               ssci: 0,
-              transmittedBitrate: 0.13333333333333333,
+              transmittedBitrate:  0.13333333333333333,
```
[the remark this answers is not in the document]
edvujic: The issue was that when bytesSent equaled 0, the system did not acknowledge the stream as having been requested. To resolve this, I modified the bytesSent value to 1. Consequently, I also adjusted the transmittedBitrate to reflect this change, ensuring the reported bitrate corresponds to the bytesSent value of 1.

mkesavan13 (approved):
Looks good to me. Thank you. 

Please update the branch before merging

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -997,6 +1010,10 @@ export class StatsAnalyzer extends EventsScope {
       this.statsResults[mediaType][sendrecvType].headerBytesSent = result.headerBytesSent;
       this.statsResults[mediaType][sendrecvType].retransmittedBytesSent =
         result.retransmittedBytesSent;
+      this.statsResults[mediaType][sendrecvType].isRequested =
+        result.isRequested ||
+        (result.lastRequestedUpdateTimestamp &&
+          performance.timeOrigin + performance.now() - result.lastRequestedUpdateTimestamp < 60000);
```
antsukanova, line 1016: There is a var for timeout `MQA_INTERVAL` Suggest using it here and other code parts with timestamps

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
```diff
@@ -1718,6 +1732,41 @@ describe('plugin-meetings', () => {
...
+        });
+
+        it('should send a stream if it was requested in the last 60 seconds', async () => {
+          fakeStats.audio.receivers[0].report[0].lastRequestedUpdateTimestamp = performance.timeOrigin + performance.now() - 30 * 1000;
+          fakeStats.audio.receivers[0].report[0].isRequested = false;
+          await progressTime(MQA_INTERVAL);
+          assert.strictEqual(mqeData.audioReceive[0].streams.length, 1);
+        })
+        it('should not send a stream if it was not requested in the last 60 seconds', async () => {
+          fakeStats.audio.receivers[0].report[0].lastRequestedUpdateTimestamp = performance.timeOrigin + performance.now() + 30 * 1000;
```
antsukanova, line 1762: To achieve consistent results in tests, especially with time values, it's better to use stubs to consistently return expected results.

like

```
const stub = sinon.stub(performance, 'now');
stub.returns(60000);
```

edvujic: Please correct me if I am wrong, but is this necessary? 
We already have: 
```javascript
const progressTime = async (time = initialConfig.analyzerInterval) => {
await clock.tickAsync(time);
await testUtils.flushPromises();
};
```

Given the `progressTime` function we have, it looks like we're already using [sinon's fake timers](https://github.com/webex/webex-js-sdk/blob/2ea5778de5128968082a07a61b320d0dd3103417/packages/%40webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js#L306) to simulate the passage of time in our tests. The `clock.tickAsync(time)` function advances the clock by a specified amount of time and resolves any promises that are due to be resolved after this simulated time has passed.

SomeBody16 (approved):
lgtm

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
antsukanova, line 1016: I suppose here is also need to be `MQA_INTERVAL`

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
antsukanova: When writing tests, mocking time helps ensure that you get the same expected result each time you run the test. The goal is to stub unreliable functions and write assertions based on the expected behavior of your code when a specific value is returned by performing.now(). This approach makes your tests more predictable and reliable.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
edvujic: Done

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
edvujic: Done

antsukanova (approved):
LGTM, thank you for adding tests cases

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -997,6 +1010,11 @@ export class StatsAnalyzer extends EventsScope {
       this.statsResults[mediaType][sendrecvType].headerBytesSent = result.headerBytesSent;
       this.statsResults[mediaType][sendrecvType].retransmittedBytesSent =
         result.retransmittedBytesSent;
+      this.statsResults[mediaType][sendrecvType].isRequested =
+        result.isRequested ||
+        (result.lastRequestedUpdateTimestamp &&
+          performance.timeOrigin + performance.now() - result.lastRequestedUpdateTimestamp <
+            MQA_INTERVAL);
```
brycetham, line 1017: This doesn't feel like the right place to put this calculation. The problem with putting this calculation here is that `processOutboundRTPResult` is called every 5 seconds, so this would not tell us whether the stream was requested within the last interval, but rather whether the stream was requested within the last 60 seconds, which could include the previous interval.

Technically, at the time the MQE is sent, this may end up being correct because the last 60 seconds at that point would be the within the last interval. But I think it would be better to just store the `isRequested` and `lastRequestedUpdateTimestamp` values here, and only do this calculation every 60 seconds when we send the MQE out.

Anna's PR does something similar to this, so you may use it as a reference: https://github.com/webex/webex-js-sdk/pull/3651

antsukanova: I agree with @brycetham. It's a great point. With that logic, it should be calculated every 60 seconds, not 5.

brycetham (commented):
Looks great! Having the helper util function for `isStreamRequested` is a good idea. Once the unit tests are fixed (the nested `it` block issue that we discussed offline) I can approve.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -1003,6 +1017,11 @@ export class StatsAnalyzer extends EventsScope {
       this.statsResults[mediaType][sendrecvType].totalRtxBytesSent = result.retransmittedBytesSent;
       this.statsResults[mediaType][sendrecvType].totalBytesSent = result.bytesSent;
       this.statsResults[mediaType][sendrecvType].headerBytesSent = result.headerBytesSent;
+      this.statsResults[mediaType][sendrecvType].retransmittedBytesSent =
+        result.retransmittedBytesSent;
```
brycetham, line 1021: I think you may have accidentally added these lines in here from another task but that should be ok :)

brycetham (approved):

PR #3622, opened 2024-05-29
fix(internal-plugin-metrics): webex-386242 Correct report JMF Reports

k-wasniowski:

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
@@ -383,11 +383,18 @@ export const generateClientErrorCodeForIceFailure = ({
     errorCode = MISSING_ROAP_ANSWER_CLIENT_CODE;
   }
 
-  if (signalingState === 'stable' && iceConnectionState === 'connected') {
+  if (
+    signalingState === 'stable' &&
+    (iceConnectionState === 'connected' || iceConnectionState === 'disconnected')
+  ) {
     errorCode = DTLS_HANDSHAKE_FAILED_CLIENT_CODE;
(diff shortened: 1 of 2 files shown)


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -383,11 +383,18 @@ export const generateClientErrorCodeForIceFailure = ({
     errorCode = MISSING_ROAP_ANSWER_CLIENT_CODE;
   }
 
-  if (signalingState === 'stable' && iceConnectionState === 'connected') {
+  if (
+    signalingState === 'stable' &&
+    (iceConnectionState === 'connected' || iceConnectionState === 'disconnected')
```
marcin-bazyl, line 388: Ah, great, so you've managed to do this fix without having to store any state. 
I think there is also a chance that the state may reach "failed" if our timeout is long enough, but probably we cannot assume that "failed" means that it was "connected" at some point, because I think both transitions are possible: connecting->failed and connecting->connected->disconnected->failed. 
Anyway this PR improves things and we can worry about "failed" state later if we find that it happens in the field often enough.

marcin-bazyl (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
k-wasniowski: Good.
Lets treat it as a short term solution, and we will be working on the long term

PR #3623, opened 2024-05-29
fix(calling): mute/unmute fix with bnr

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-511579 

Issue 1: After disabling BNR on the same call, if we try to enable BNR in a muted state, then stream goes missing completely, and after unmuting, the stream is still missing.

Steps to reproduce: Enable BNR and disable BNR while the call is unmuted. Now mute the call and enable BNR. Unmute the call now and the stream should go missing without the fix.

Issue 2: In subsequent calls after disabling BNR, enabling BNR in an unmuted state messes up the mute-unmute behaviour again and the stream remains present always.

Steps to reproduce: Enable BNR and disable BNR while the call is unmuted and disconnect the call. Now establish second call and enable BNR in unmuted state. Now try to mute the call and the stream will always be present and can be heard on the other side. Muting the stream stops working without the fix.

Instead of changing localAudioTrack.enabled property using setUserMuted() function on localAudioStream fixes the issue.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Issue 1 and Issue 2 were tested with the fix.

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -233,6 +233,7 @@ describe('Call Tests', () => {
         getAudioTracks: jest.fn().mockReturnValue([mockTrack]),
       },
       on: jest.fn(),
+      setUserMuted: jest.fn(),
     };
 
     const localAudioStream = mockStream as unknown as MediaSDK.LocalMicrophoneStream;
@@ -244,9 +245,9 @@ describe('Call Tests', () => {
     expect(Object.keys(callManager.getActiveCalls()).length).toBe(1);
     call.mute(localAudioStream);
     expect(call.isMuted()).toEqual(true);
-    expect(mockTrack.enabled).toEqual(false);
+    expect(mockStream.setUserMuted).toBeCalledOnceWith(true);
     call.mute(localAudioStream);
-    expect(mockTrack.enabled).toEqual(true);
+    expect(mockStream.setUserMuted).toBeCalledWith(false);
     expect(call.isMuted()).toEqual(false);
     call.end();
     await waitForMsecs(50); // Need to add a small delay for Promise and callback to finish.
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -229,9 +229,6 @@ describe('Call Tests', () => {
     const callManager = getCallManager(webex, defaultServiceIndicator);
 
     const mockStream = {
-      outputStream: {
-        getAudioTracks: jest.fn().mockReturnValue([mockTrack]),
-      },
       on: jest.fn(),
       setUserMuted: jest.fn(),
     };
--- a/packages/calling/src/CallingClient/calling/call.ts
+++ b/packages/calling/src/CallingClient/calling/call.ts
@@ -2717,8 +2717,15 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    * @param localAudioTrack -.
    */
   public mute = (localAudioStream: LocalMicrophoneStream): void => {
-    localAudioStream.setUserMuted(!this.muted);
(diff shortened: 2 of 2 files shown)


packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2717,14 +2717,8 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
   public mute = (localAudioStream: LocalMicrophoneStream): void => {
-    const localAudioTrack = localAudioStream.outputStream.getAudioTracks()[0];
-    if (this.muted) {
-      localAudioTrack.enabled = true;
-      this.muted = false;
-    } else {
-      localAudioTrack.enabled = false;
-      this.muted = true;
-    }
+    localAudioStream.setUserMuted(!this.muted);
```
Shreyas281299, line 2720: Should we add a check here to see if localAudioStream is passed or not? I remember in one of the contact center scenarios they were calling the answer method without localAudioStream, since they were using a js file it didn't show as an error. So a similar scenario might come here also.

Shreyas281299 (changes requested):

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -233,6 +233,7 @@ describe('Call Tests', () => {
         getAudioTracks: jest.fn().mockReturnValue([mockTrack]),
       },
       on: jest.fn(),
+      setUserMuted: jest.fn(),
```
Shreyas281299, line 236: We can remove the outputStream object from mockStream as we have removed that logic in the src file.

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: @Shreyas281299 localAudioStream is a mandatory parameter for dial, answer and mute functions. So expectation here is they should always be present and specifically for mute, if there is no localAudiostream present then call won't exist and hence mute function is not applicable for such cases. Do you still feel we should add the check for stream here?

packages/calling/src/CallingClient/calling/call.test.ts
Kesari3008: removed

Shreyas281299 (approved):
Looks good. LGTM!

sreenara (changes requested):
This change reminds me of the change that was recently done on the main web client to handle the system mute option on certain devices. Should we handle that event too on the Calling SDK?

Kesari3008:
> https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-511579

Yes, that makes sense. We should handle the system muted event too in Calling SDK and it will require thorough testing where we need to test the mute status getting updated and rfeflecting correctly irrespective whether user mute or system mute was triggered. Both the mute functions should be aligned along with the local audio stream getting updated and functioning correctly. And with BNR as well, we need to do testing as updateLocalTracks impacts the stream too which causes issue with mute/unmute behavior. Hence I will create a separate story where we can take care of the mute/unmute flow entirely.

Kesari3008:
> > https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-511579
> 
> Yes, that makes sense. We should handle the system muted event too in Calling SDK and it will require thorough testing where we need to test the mute status getting updated and rfeflecting correctly irrespective whether user mute or system mute was triggered. Both the mute functions should be aligned along with the local audio stream getting updated and functioning correctly. And with BNR as well, we need to do testing as updateLocalTracks impacts the stream too which causes issue with mute/unmute behavior. Hence I will create a separate story where we can take care of the mute/unmute flow entirely.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-530539

sreenara (approved):

PR #3634, opened 2024-06-07
feat(fedramp):  only add domains if commercial

robstax:
# COMPLETES https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-531900

in fedramp, it's important that clients do not send auth token to resources outside of the "fedramp" boundary. therefore, we should only send auth headers in the following cases
1. the resource is in the u2c catalog
2. the resource is in the list of allowed domains

i changed the allowedDomains to be empty in FedRAMP by default. if the user wants more domains in fedramp, they can pass them through config or by `setAllowedDomains()`. in commercial, we keep the existing list of allowedDomains.
the reason webex.com cannot be a default allowedDomain in fedramp is because there are commercial sites, like cisco.webex.com and we don't want fedramp users sending their auth tokens to arbitrary commercial *.webex sites

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. manually tested with config value `fedramp: false`
2. manually tested with config value `fedramp: true`
3. ran web client automated playwright E2E tests with config value `fedramp: false`
4. ran web client automated playwright E2E tests with config value `fedramp: true`
5. in commercial make request to cisco.webex.com (request added auth headers since within `*.webex` allowed domains
6. in commercial make request to cisco.webex.com (SDK does not add auth header) since it's not within catalog/not allowed domain

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1,5 +1,5 @@
 /* eslint no-shadow: ["error", { "allow": ["eventType"] }] */
-
+import {union} from 'lodash';
 import '@webex/internal-plugin-mercury';
 import '@webex/internal-plugin-conversation';
 import '@webex/internal-plugin-metrics';
@@ -1001,7 +1001,15 @@ export default class Meetings extends WebexPlugin {
   fetchUserPreferredWebexSite() {
     return this.request.getMeetingPreferences().then((res) => {
       if (res) {
-        this.preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+        const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+        this.preferredWebexSite = preferredWebexSite;
+        // @ts-ignore
+        this.webex.internal.services._getCatalog().setAllowedDomains(
+          // @ts-ignore
+          union(this.webex.internal.services._getCatalog().getAllowedDomains(), [
+            preferredWebexSite,
+          ])
+        );
       }
 
       // fall back to getting the preferred site from the user information
@@ -1014,6 +1022,13 @@ export default class Meetings extends WebexPlugin {
               user?.userPreferences?.userPreferencesItems?.preferredWebExSite;
             if (preferredWebexSite) {
               this.preferredWebexSite = preferredWebexSite;
+              // @ts-ignore
+              this.webex.internal.services._getCatalog().setAllowedDomains(
+                // @ts-ignore
+                union(this.webex.internal.services._getCatalog().getAllowedDomains(), [
+                  preferredWebexSite,
+                ])
+              );
             } else {
               throw new Error('site not found');
             }
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1004,12 +1004,7 @@ export default class Meetings extends WebexPlugin {
         const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
         this.preferredWebexSite = preferredWebexSite;
         // @ts-ignore
-        this.webex.internal.services._getCatalog().setAllowedDomains(
-          // @ts-ignore
-          union(this.webex.internal.services._getCatalog().getAllowedDomains(), [
-            preferredWebexSite,
-          ])
-        );
+        this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
       }
 
       // fall back to getting the preferred site from the user information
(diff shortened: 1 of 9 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -996,7 +996,15 @@ export default class Meetings extends WebexPlugin {
   fetchUserPreferredWebexSite() {
     return this.request.getMeetingPreferences().then((res) => {
       if (res) {
-        this.preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+        const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+        this.preferredWebexSite = preferredWebexSite;
+        // @ts-ignore
+        this.webex.internal.services._getCatalog().setAllowedDomains(
+          // @ts-ignore
+          union(this.webex.internal.services._getCatalog().getAllowedDomains(), [
```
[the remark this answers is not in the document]
robstax: this is a tiny optimization (probably overkill) so that the list of allowedDomains is unique. trying to prevent pushing the same domain again fro some reason. honestly it's probably overkill, because we don't expect this to get very big. i also considered converting the array to a Set() but ampersand's support of Set seems a little dodgy

packages/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -361,6 +362,15 @@ const ServiceCatalog = AmpState.extend({
     this.allowedDomains = [...allowedDomains];
   },
 
+  /**
+   *
+   * @param {Array<string>} newAllowedDomains - new alllowed doamins to add to existing set of allowed domains
```
sreenara, line 367: typo in the description

sreenara (commented):
Can we expand on what tests were done in the appropriate section of the PR description?

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -941,6 +944,11 @@ const Services = WebexPlugin.extend({
         catalog.updateServiceUrls('override', formattedOverrideServices);
       }
 
+      // if not fedramp, append on the commercialAllowedDomains
+      if (!fedramp) {
+        services.allowedDomains = [...services.allowedDomains, ...COMMERCIAL_ALLOWED_DOMAINS];
```
Kesari3008, line 949: I see in the service-catalog file we have used the union to merge 2 arrays to ensure that allowedDomains list is unique. Would there be no possibility to have duplicates here between services.AllowedDomains and COMMERCIAL_ALLOWED_DOMAINS?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1001,7 +1001,10 @@ export default class Meetings extends WebexPlugin {
   fetchUserPreferredWebexSite() {
     return this.request.getMeetingPreferences().then((res) => {
       if (res) {
-        this.preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+        const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
```
Kesari3008, line 1004: Why did we change this? I believe there is no difference between the direct and indirect assignments in terms of performance.

mkesavan13 (approved):
Couple of minor non-blocking comments on test files. Otherwise, looks good to me

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -996,7 +996,15 @@ export default class Meetings extends WebexPlugin {
...
+        this.preferredWebexSite = preferredWebexSite;
+        // @ts-ignore
+        this.webex.internal.services._getCatalog().setAllowedDomains(
```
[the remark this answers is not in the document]
robstax: done

packages/@webex/webex-core/src/lib/services/service-catalog.js
robstax: will fix

packages/@webex/webex-core/src/lib/services/services.js
robstax: valid point. will change

packages/@webex/plugin-meetings/src/meetings/index.ts
robstax: main reason i did it was because on line 1007 i do
```
this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]
```
and i didn't want to have to call `MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);` again

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -1917,32 +1925,36 @@ describe('plugin-meetings', () => {
...
-              getMeetingPreferences: sinon.stub().returns(Promise.resolve({})),
-            },
             user: {
               get: sinon.stub().returns(Promise.resolve(user)),
             },
           });
+
+          Object.assign(webex.internal.services, {
+            getMeetingPreferences: sinon.stub().returns(Promise.resolve({})),
+          });
```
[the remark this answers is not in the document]
robstax: woops. i can remove this

packages/@webex/webex-core/test/unit/spec/services/service-catalog.js
```diff
@@ -189,6 +177,30 @@ describe('webex-core', () => {
...
+
+      beforeEach(() => {
+        domains.push('example-a', 'example-b', 'example-c');
+
+        catalog.setAllowedDomains(domains);
+      });
+
+      afterEach(() => {
+        domains.length = 0;
+      });
```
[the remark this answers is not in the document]
robstax: i don't think it really hurts anything having it and makes it easier to add test cases in the future if we want

robstax:
i addressed all comments. thanks for reviews!

PR #3638, opened 2024-06-10
fix: 2 reconnections started in parallel

marcin-bazyl:
# COMPLETES #[SPARK-531442](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-531442)

When a reconnection is started, we first do reachability checks and if during that time we get a FAILED connection state notification, we end up starting another reconnection. We should only do 1 reconnection at a time.

Setting ReconnectionManager.status to IN_PROGRESS sooner and also refactored the code a bit so that the status is private and only ReconnectionManager changes it. Also moved sending of some metrics and emitting events from Meeting into ReconnectionManager so that now ReconnectionManager really is the one managing the reconnection.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -66,6 +66,7 @@ import {createMuteState} from './muteState';
 import LocusInfo from '../locus-info';
 import Metrics from '../metrics';
 import ReconnectionManager from '../reconnection-manager';
+import ReconnectionNotStartedError from '../common/errors/reconnection-not-started';
 import MeetingRequest from './request';
 import Members from '../members/index';
 import MeetingUtil from './util';
@@ -76,8 +77,6 @@ import MediaUtil from '../media/util';
 import {Reactions, SkinTones} from '../reactions/reactions';
 import PasswordError from '../common/errors/password-error';
 import CaptchaError from '../common/errors/captcha-error';
-import ReconnectionError from '../common/errors/reconnection';
-import ReconnectInProgress from '../common/errors/reconnection-in-progress';
 import {
   _CONVERSATION_URL_,
   _INCOMING_,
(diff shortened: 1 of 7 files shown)


packages/@webex/plugin-meetings/src/reconnection-manager/index.ts
```diff
@@ -265,149 +256,183 @@ export default class ReconnectionManager {
...
 
     try {
-      await this.webex.meetings.startReachability();
-    } catch (err) {
-      LoggerProxy.logger.info(
-        'ReconnectionManager:index#reconnect --> Reachability failed, continuing with reconnection attempt, err: ',
-        err
-      );
-    }
+      this.status = RECONNECTION.STATE.IN_PROGRESS;
```
marcin-bazyl, line 325: this is the actual fix - that we're setting "in progress" state earlier now

packages/@webex/plugin-meetings/test/unit/spec/reconnection-manager/index.js
```diff
@@ -2,37 +2,38 @@ import 'jsdom-global/register';
...
 import LoggerConfig from '../../../../src/common/logs/logger-config';
-
 const {assert} = chai;
 
 chai.use(chaiAsPromised);
 sinon.assert.expose(chai.assert, {prefix: ''});
 
 describe('plugin-meetings', () => {
   describe('ReconnectionManager.reconnect', () => {
-    const sandbox = sinon.createSandbox();
```
marcin-bazyl, line 17: I've removed the sandbox, because it's not needed. I think this is something that was required in very old versions of sinon, now sinon has a default sandbox always there.

packages/@webex/plugin-meetings/test/unit/spec/reconnection-manager/index.js
```diff
@@ -316,29 +416,41 @@ describe('plugin-meetings', () => {
 
     describe('waitForIceReconnect()', () => {
```
marcin-bazyl, line 417: the changes inside this `describe` are not strictly related to my stuff, but I've noticed that these tests were causing the process to not exit correctly at the end of the tests, because the timer was still running for another 10s after the tests finished, so I've fixed it by using fake timers

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4664,90 +4663,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
-            file: 'meeting/index',
-            function: 'reconnect',
-          },
-          EVENT_TRIGGERS.MEETING_RECONNECTION_FAILURE,
-          {
-            error: new ReconnectionError('Reconnection failure event', error),
-          }
-        );
+        if (error instanceof ReconnectionNotStartedError) {
+          LoggerProxy.logger.log('Meeting:index#reconnect --> Meeting reconnect not started');
```
jor-row, line 4676: Should this log say "or reconnection is not enabled in config" too since we aren't logging the whole error string?

jor-row: Actually, ```canStartReconnection``` adds a log if it's not in then config so that should suffice.

packages/@webex/plugin-meetings/test/unit/spec/reconnection-manager/index.js
jor-row: Thanks for fixing that

jor-row (approved):
Some small non-blocking comments. Have tested locally and the fix/refactor is working as expected

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yes, correct, that's why I left this one as more generic message

Shreyas281299 (approved):
Looks good.

PR #3640, opened 2024-06-11
chore(codeowners): update codeowners file

sreenara:
# COMPLETES # NA

the issue of the Web SDK team not being added into reviews automatically as codeowners for packages where the Web Client and WCME teams are making changes.
The older CODEOWNERS file was fine because the teams were using different branches (`beta` and `master`). However, now that all teams are using the `next` branch, there is a need to be careful about breaking changes which can cause issues in applications consuming the SDK (Webex Web Client, Instant Connect, Widgets, 3rd party developer applications).

Update @webex/web-sdk as owner for `@webex/plugin-meetings`, `@webex/internal-plugin-voicea` and all their dependent packages as these are the main packages that the SDK team needs notifications on.

### Change Type

- [ ] Documentation update
- [x] Tooling change
- [ ] Internal code refactor

NA

### I certified that

---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,19 +5,28 @@
 
 # Commonly Owned Packages
 
-packages/@webex/webex-core                                @webex/web-client @webex/web-sdk
 packages/@webex/common                                    @webex/web-client @webex/web-sdk
-packages/@webex/helper-html                               @webex/web-client @webex/web-sdk
-packages/@webex/helper-image                              @webex/web-client @webex/web-sdk
+packages/@webex/helper-html                               @webex/web-client
+packages/@webex/helper-image                              @webex/web-client
 packages/@webex/http-core                                 @webex/web-client @webex/web-sdk
-packages/@webex/internal-plugin-*                         @webex/web-client @webex/web-sdk
-packages/@webex/plugin-authorization-browser-first-party  @webex/web-client @webex/web-sdk
+packages/@webex/internal-plugin-conversation              @webex/web-client @webex/web-sdk
+packages/@webex/internal-plugin-device                    @webex/web-client @webex/web-sdk
+packages/@webex/internal-plugin-llm                       @webex/web-client @webex/web-sdk
+packages/@webex/internal-plugin-mercury                   @webex/web-client @webex/web-sdk
+packages/@webex/internal-plugin-metrics                   @webex/web-client @webex/web-sdk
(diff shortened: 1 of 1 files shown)


mkesavan13 (changes requested):
1. Since we are the calling owners, we can specify ourselves as Code Owners there
2. Considering that, we should also ensure if the packages that Calling depend on are also explicitly specified

Looks good otherwise

sreenara:
> 1. Since we are the calling owners, we can specify ourselves as Code Owners there
> 2. Considering that, we should also ensure if the packages that Calling depend on are also explicitly specified
> 
> Looks good otherwise

We'd fall under the default rule of `*` since we are the Web SDK team right? The codeowners file says "Web SDK Team Ownership *". Calling comes under this, so it should be ok? If we create a separate rule for calling, then in the future we would have to create rules for all other plugins we are creating.

Calling only depends on `@webex/media-helpers` and that's added in the codeowners file.

mkesavan13:
> > 1. Since we are the calling owners, we can specify ourselves as Code Owners there
> > 2. Considering that, we should also ensure if the packages that Calling depend on are also explicitly specified
> > 
> > Looks good otherwise
> 
> We'd fall under the default rule of `*` since we are the Web SDK team right? The codeowners file says "Web SDK Team Ownership *". Calling comes under this, so it should be ok? If we create a separate rule for calling, then in the future we would have to create rules for all other plugins we are creating.
> 
> Calling only depends on `@webex/media-helpers` and that's added in the codeowners file.

If we do not mention it specifically, it does ask for a review from us but, it doesn't ask for approval from our team and Calling needs approval from us for sure, right?

mkesavan13 (approved):

sreenara:
> > > 1. Since we are the calling owners, we can specify ourselves as Code Owners there
> > > 2. Considering that, we should also ensure if the packages that Calling depend on are also explicitly specified
> > > 
> > > Looks good otherwise
> > 
> > 
> > We'd fall under the default rule of `*` since we are the Web SDK team right? The codeowners file says "Web SDK Team Ownership *". Calling comes under this, so it should be ok? If we create a separate rule for calling, then in the future we would have to create rules for all other plugins we are creating.
> > Calling only depends on `@webex/media-helpers` and that's added in the codeowners file.
> 
> If we do not mention it specifically, it does ask for a review from us but, it doesn't ask for approval from our team and Calling needs approval from us for sure, right?

Please see #3641 which requires a review and approval from the Web SDK team for calling change.

Kesari3008 (approved):

PR #3642, opened 2024-06-11
fix(plugin-meetings): unpublish streams on moveTo

mkesavan13:
# COMPLETES [SPARK-533903](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-533903)

Instant connected reported that, the `moveTo` API doesn't unpublish the localStreams. This is because the method `closeRemoteStreams` that was there before `cleanupLocalStreams` handled it appropriately and `cleanupLocalStreams` doesn't

- Added the cleanup logic in `moveTo` method that closes the existing media connection, cleans up remote and local streams and also stops statsAnalyzer from sending any data
- Removed reconnect logic as that is not required anymore
- Instead we do `addMedia` with a forced retry to do turn discovery (`isMoveToInProgress` flag ensures of that in the `establishMediaConnection` method)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Join a Meeting with Audio and Video > Pair to a device > `moveTo` - This ensured the Video and Audio gets cut off
- Join a Meeting with Audio and Video > Share Screen > Pair to a device > `moveTo` - This ensured the screenshare is properly cut off on `moveTo`

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5436,6 +5436,7 @@ export default class Meeting extends StatelessWebexPlugin {
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
         // once the DX answers we establish connection back the media server with only receiveShare enabled
         this.closePeerConnections();
+        this.unsetPeerConnections();
 
         await this.unpublishStreams([
           this.mediaProperties.audioStream,
@@ -5444,7 +5445,6 @@ export default class Meeting extends StatelessWebexPlugin {
           this.mediaProperties.shareVideoStream,
         ]);
 
-        this.unsetPeerConnections();
         await this.addMedia({
           audioEnabled: false,
           videoEnabled: false,
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5440,7 +5440,9 @@ export default class Meeting extends StatelessWebexPlugin {
 
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
         // once the DX answers we establish connection back the media server with only receiveShare enabled
-        await this.reconnectionManager.reconnectMedia().then(() => {
+        this.closePeerConnections();
+        this.unsetPeerConnections();
+        await this.createMediaConnection({}).then(() => {
           Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
         });
       } catch (error) {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5489,6 +5489,28 @@ export default class Meeting extends StatelessWebexPlugin {
           },
         };
 
+        const streamsToUnpublish = [];
+
+        if (this.mediaProperties?.audioStream) {
+          streamsToUnpublish.push(this.mediaProperties?.audioStream);
+        }
+
+        if (this.mediaProperties?.videoStream) {
+          streamsToUnpublish.push(this.mediaProperties?.videoStream);
+        }
+
+        if (this.mediaProperties?.shareAudioStream) {
+          streamsToUnpublish.push(this.mediaProperties?.shareAudioStream);
+        }
+
+        if (this.mediaProperties?.shareVideoStream) {
+          streamsToUnpublish.push(this.mediaProperties?.shareVideoStream);
+        }
+
+        if (streamsToUnpublish.length) {
+          this.unpublishStreams(streamsToUnpublish);
+        }
+
         this.cleanupLocalStreams();
 
         this.mediaProperties.setMediaDirection(mediaSettings.mediaDirection);
@@ -5496,10 +5518,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
         // once the DX answers we establish connection back the media server with only receiveShare enabled
-        // @ts-ignore - reconnectMedia does not accept any argument
-        await this.reconnectionManager.reconnectMedia(mediaSettings).then(() => {
-          Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
-        });
+        Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
       } catch (error) {
         LoggerProxy.logger.error('Meeting:index#moveTo --> Failed to moveTo resourceId', error);
         Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_FAILURE, {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5439,13 +5439,6 @@ export default class Meeting extends StatelessWebexPlugin {
         this.closePeerConnections();
         this.unsetPeerConnections();
 
-        await this.unpublishStreams([
-          this.mediaProperties.audioStream,
-          this.mediaProperties.videoStream,
-          this.mediaProperties.shareAudioStream,
-          this.mediaProperties.shareVideoStream,
-        ]);
-
         await this.addMedia({
           audioEnabled: false,
           videoEnabled: false,
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -631,7 +631,7 @@ export default class Meeting extends StatelessWebexPlugin {
   turnDiscoverySkippedReason: TurnDiscoverySkipReason;
   turnServerUsed: boolean;
   areVoiceaEventsSetup = false;
-  isMoveTo = false;
+  isMoveToInProgress = false;
 
   voiceaListenerCallbacks: object = {
     [VOICEAEVENTS.VOICEA_ANNOUNCEMENT]: (payload: Transcription['languageOptions']) => {
@@ -5434,10 +5434,9 @@ export default class Meeting extends StatelessWebexPlugin {
 
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
         // once the DX answers we close the old connection and create new media server connection with only share enabled
-        const stopStatsAnalyzer = this.statsAnalyzer
-          ? this.statsAnalyzer.stopAnalyzer()
-          : Promise.resolve();
-        await stopStatsAnalyzer;
+        if (this.statsAnalyzer) {
+          await this.statsAnalyzer.stopAnalyzer();
+        }
         await this.closeRemoteStreams();
         await this.closePeerConnections();
         this.cleanupLocalStreams();
@@ -5449,10 +5448,9 @@ export default class Meeting extends StatelessWebexPlugin {
           audioEnabled: false,
           videoEnabled: false,
           shareVideoEnabled: true,
-        }).then(() => {
-          Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
-          this.isMoveTo = false;
         });
+        Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
+        this.isMoveToInProgress = false;
       } catch (error) {
         LoggerProxy.logger.error('Meeting:index#moveTo --> Failed to moveTo resourceId', error);
         Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_FAILURE, {
@@ -5461,7 +5459,7 @@ export default class Meeting extends StatelessWebexPlugin {
           reason: error.message,
           stack: error.stack,
         });
-        this.isMoveTo = false;
+        this.isMoveToInProgress = false;
       }
     });
 
@@ -5470,7 +5468,7 @@ export default class Meeting extends StatelessWebexPlugin {
       resourceId
     );
 
-    this.isMoveTo = true;
+    this.isMoveToInProgress = true;
 
     return MeetingUtil.joinMeetingOptions(this, {resourceId, moveToResource: true})
       .then(() => {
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5489,17 +5489,36 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        if (this.mediaProperties?.shareAudioStream) {
+          streamsToUnpublish.push(this.mediaProperties?.shareAudioStream);
+        }
+
+        if (this.mediaProperties?.shareVideoStream) {
+          streamsToUnpublish.push(this.mediaProperties?.shareVideoStream);
+        }
+
+        if (streamsToUnpublish.length) {
+          this.unpublishStreams(streamsToUnpublish);
```
marcin-bazyl, line 5511: unpublishStreams() has a check for each member of the passed in array to be truthy and skips the ones that aren't, so instead of the block of lines 5494-5508 you could just call it like this:
```
this.unpublishStreams([
  this.mediaProperties.audioStream,
  this.mediaProperties.videoStream,
  this.mediaProperties.shareAudioStream,
  this.mediaProperties.shareVideoStream
]
```
if any of the streams are undefined it will skip those and only unpublish the ones that are defined. Also `this.mediaProperties` is guaranteed to exist, so no need for the optional chaining operator.

mkesavan13: Took care of this comment.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -6243,24 +6244,19 @@ describe('plugin-meetings', () => {
           );
 
           // beacuse we are calling callback so we need to wait
+          assert.calledWith(meeting.unpublishStreams,[
```
marcin-bazyl, line 6247: just to double check - are these streams inside mediaProperties actually set to something other than undefined or null in this test?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5489,17 +5489,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
         this.cleanupLocalStreams();
 
         this.mediaProperties.setMediaDirection(mediaSettings.mediaDirection);
         this.mediaProperties.unsetRemoteMedia();
 
-        // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
-        // once the DX answers we establish connection back the media server with only receiveShare enabled
-        // @ts-ignore - reconnectMedia does not accept any argument
-        await this.reconnectionManager.reconnectMedia(mediaSettings).then(() => {
```
marcin-bazyl, line 5500: The call to reconnectMedia() was creating a new confluence in the backend (because the reconnection triggers a call to Locus /media API with a new SDP offer and an empty `mediaId`). It would be good to confirm with the backend teams (Locus, Linus) that we don't need to do that for this moveTo flow

marcin-bazyl:
The PR description says "COMPLETES AdHoc" - we're missing a jira link here

mkesavan13:
> The PR description says "COMPLETES AdHoc" - we're missing a jira link here

We do not have a Jira for this. I took it as an Adhoc activity to fix this @marcin-bazyl

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Oh. Looks like they were null. I have pushed a new commit to have dummy strings for them on `beforeEach`. Kindly check.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: update for anyone reading this: after various attempts and discussions, the conclusion is that we need to stop the old media connection and cleanup everything related to that including the local streams and then call addMedia() again and make sure that TURN discovery is done if it was done during the initial meeting join.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6506,7 +6526,12 @@ export default class Meeting extends StatelessWebexPlugin {
     turnServerInfo?: TurnServerInfo
   ): Promise<void> {
     const LOG_HEADER = 'Meeting:index#addMedia():establishMediaConnection -->';
-    const isRetry = this.retriedWithTurnServer;
+    const isRetry = this.isMoveTo || this.retriedWithTurnServer;
```
marcin-bazyl, line 6529: maybe `isReconnecting` would be a better name now for this variable? Because in the moveTo flow we are not retrying, we are just doing a new connection. Even `isReconnecting` is not a perfect name, but I can't think of a better one...

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -631,6 +631,7 @@ export default class Meeting extends StatelessWebexPlugin {
   turnDiscoverySkippedReason: TurnDiscoverySkipReason;
   turnServerUsed: boolean;
   areVoiceaEventsSetup = false;
+  isMoveTo = false;
```
marcin-bazyl, line 634: maybe `isMoveToInProgress` would be a better name?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5428,16 +5429,31 @@ export default class Meeting extends StatelessWebexPlugin {
...
 
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
-        // once the DX answers we establish connection back the media server with only receiveShare enabled
-        // @ts-ignore - reconnectMedia does not accept any argument
-        await this.reconnectionManager.reconnectMedia(mediaSettings).then(() => {
+        // once the DX answers we close the old connection and create new media server connection with only share enabled
+        const stopStatsAnalyzer = this.statsAnalyzer
+          ? this.statsAnalyzer.stopAnalyzer()
+          : Promise.resolve();
+        await stopStatsAnalyzer
```
marcin-bazyl, line 5440: this whole block of code from line 5437 to 5448 is now a mix of `await` and `then`. There is no reason for that, let's try to clean it up and just use await for all the calls. We also don't need to do the stuff from lines 5437-5439, we can just simply do this:
```
if (this.statsAnalyzer) {
  await this.statsAnalyzer.stopAnalyzer();
}
```

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5428,16 +5429,31 @@ export default class Meeting extends StatelessWebexPlugin {
...
+            this.unsetRemoteStreams();
+            this.unsetPeerConnections();
+            this.reconnectionManager.cleanUp();
+          });
+
+        await this.addMedia({
+          audioEnabled: false,
+          videoEnabled: false,
+          shareVideoEnabled: true,
+        }).then(() => {
```
marcin-bazyl, line 5454: `.then` here is not needed as we're using await already, right?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5455,6 +5472,8 @@ export default class Meeting extends StatelessWebexPlugin {
       resourceId
     );
 
+    this.isMoveTo = true;
```
marcin-bazyl, line 5475: is it possible for `MeetingUtil.joinMeetingOptions()` to succeed but for us to never the the `LOCUSINFO.EVENTS.SELF_OBSERVING` event? In such case the `isMoveTo` flag would remain `true` forever. Not sure if we need to handle such edge case..... we would need to add some timeout if we wanted to handle it...

marcin-bazyl (commented):
Overall looks good, it should work correctly now. I haven't tested it myself. 

The unit tests are failing, I think they need some updates after recent changes, so I haven't reviewed them.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: We still have to cleanup the streams right?

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: This event is emitted from inside locusInfo but I am doubtful if there will be a scenario when the event won't be emitted. (Unless locus sends selfObserving flag wrongly)

However, what would be the timeout we can set for this? I thought of a minute but what if that is too less of a time to timeout and addMedia isn't even called by that time

marcin-bazyl: I don't know how long the timeout should be, maybe Locus team has something to say about this. This is not blocking for this PR, but maybe try to find out if we need to handle this edge case

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -6232,36 +6243,38 @@ describe('plugin-meetings', () => {
...
           );
+          
 
-          // beacuse we are calling callback so we need to wait
-
-          assert.called(meeting.cleanupLocalStreams);
-
-          // give queued Promise callbacks a chance to run
+          // Verify that the event handler behaves as expected
+          expect(meeting.statsAnalyzer.stopAnalyzer.calledOnce).to.be.true;
```
marcin-bazyl, line 6261: vast majority of existing tests in plugin-meetings do
`assert.calledOnce(meeting.statsAnalyzer.stopAnalyzer)`
instead of `expect(...calledOnce).to.be.true`
any reason why you're doing it this way?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -6232,36 +6243,38 @@ describe('plugin-meetings', () => {
...
+          expect(meeting.unsetRemoteStreams.calledOnce).to.be.true;
+          expect(meeting.unsetPeerConnections.calledOnce).to.be.true;
+          expect(meeting.reconnectionManager.cleanUp.calledOnce).to.be.true;
+          expect(meeting.mediaProperties.setMediaDirection.calledOnce).to.be.true;
+          expect(meeting.addMedia.calledOnceWithExactly({
+            audioEnabled: false,
+            videoEnabled: false,
+            shareVideoEnabled: true
+          })).to.be.true;
+          await Promise.resolve();
```
marcin-bazyl, line 6276: I think we have a util for this: `await testUtils.flushPromises();` (we could use it also in line 6263 and 6265)

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Sure. I'll have a TODO to pursue this

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -6232,36 +6243,38 @@ describe('plugin-meetings', () => {
...
+          assert.equal(meeting.isMoveToInProgress, false);
         });
 
         it('should throw an error if moveTo call fails', async () => {
```
marcin-bazyl, line 6280: we could add a check that isMoveToInProgress is false at the end of this test

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: yes

mkesavan13: So, I believe the code now is correct. Shall I resolve this conversation?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: It is just a way of writing tests. I am more used to writing `expects` than `asserts`

I could change everything to assert if required. I saw there are few places in some tests that use `expect` and therefore decided to go with expect for everything

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Added a TODO for this

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5428,17 +5429,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
-
         this.mediaProperties.setMediaDirection(mediaSettings.mediaDirection);
         this.mediaProperties.unsetRemoteMedia();
 
         // when a move to is intiated by the client , Locus delets the existing media node from the server as soon the DX answers the meeting
-        // once the DX answers we establish connection back the media server with only receiveShare enabled
-        // @ts-ignore - reconnectMedia does not accept any argument
-        await this.reconnectionManager.reconnectMedia(mediaSettings).then(() => {
-          Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MOVE_TO_SUCCESS);
+        // once the DX answers we close the old connection and create new media server connection with only share enabled
```
sreenara, line 5436: Is this code only applicable for DX? Shouldn't it be any device that is paired? If that's the case, can we change "DX" to "paired device"?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1390,11 +1391,11 @@ export default class Meeting extends StatelessWebexPlugin {
     this.remoteMediaManager = null;
 
     this.localAudioStreamMuteStateHandler = () => {
-      this.audio.handleLocalStreamMuteStateChange(this);
+      this.audio?.handleLocalStreamMuteStateChange(this);
```
sreenara, line 1394: Just a question. Why have the audio and video objects become optional?

sreenara (approved):
LGTM. Couple of non-blocking comments. The description could also be a little clearer, perhaps a sequence of what happens today?

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: It is applicable for all devices that can be paired. Updated the comment

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: We close the media connection and do cleanup. There is a possibility that the audio or video stream is already cleaned up by the time the event is emitted. That is why we have a null check here

PR #3643, opened 2024-06-12
fix(plugin-meetings): numberOfCores stats

antsukanova:
# COMPLETES # [WEBEX-379500](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-379500)

Missing MQE stats `cpuInfo.numberOfCores` 

Adds `cpuInfo.numberOfCores` to MQE to expose the number of logical processors

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] unit tests for plugin-meetings
- [x] manual testing via SDK test app

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/package.json
+++ b/packages/@webex/plugin-meetings/package.json
@@ -74,7 +74,7 @@
     "@webex/media-helpers": "workspace:*",
     "@webex/plugin-people": "workspace:*",
     "@webex/plugin-rooms": "workspace:*",
-    "@webex/web-capabilities": "1.3.0",
+    "@webex/web-capabilities": "^1.3.0",
     "@webex/webex-core": "workspace:*",
     "ampersand-collection": "^2.0.2",
     "bowser": "^2.11.0",
--- a/packages/@webex/plugin-meetings/package.json
+++ b/packages/@webex/plugin-meetings/package.json
@@ -74,6 +74,7 @@
     "@webex/media-helpers": "workspace:*",
     "@webex/plugin-people": "workspace:*",
     "@webex/plugin-rooms": "workspace:*",
+    "@webex/web-capabilities": "1.3.0",
     "@webex/webex-core": "workspace:*",
     "ampersand-collection": "^2.0.2",
     "bowser": "^2.11.0",
--- a/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
+++ b/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
@@ -6,7 +6,7 @@ export const emptyMqaInterval = {
     peripherals: [],
     cpuInfo: {
       numberOfCores: 1, // default value from spec if CpuInfo.getNumLogicalCores cannot be determined
-      description: 'web client',
+      description: 'NA',
       architecture: 'unknown',
     },
     processAverageCPU: 0,
--- a/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
+++ b/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
@@ -4,6 +4,9 @@ export const emptyMqaInterval = {
   intervalMetadata: {
     peerReflexiveIP: '0.0.0.0',
     peripherals: [],
+    cpuInfo: {
+      numberOfCores: 0,
+    },
     processAverageCPU: 0,
     processMaximumCPU: 0,
     systemAverageCPU: 0,
--- a/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
+++ b/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
@@ -5,7 +5,9 @@ export const emptyMqaInterval = {
     peerReflexiveIP: '0.0.0.0',
     peripherals: [],
     cpuInfo: {
-      numberOfCores: 0,
+      numberOfCores: 1,
+      description: 'web client',
+      architecture: 'unknown',
     },
     processAverageCPU: 0,
     processMaximumCPU: 0,
--- a/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
+++ b/packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
@@ -5,7 +5,7 @@ export const emptyMqaInterval = {
     peerReflexiveIP: '0.0.0.0',
     peripherals: [],
     cpuInfo: {
-      numberOfCores: 1, // default value from spec if web-caability cannot be determined
+      numberOfCores: 1, // default value from spec if CpuInfo.getNumLogicalCores cannot be determined
       description: 'web client',
       architecture: 'unknown',
     },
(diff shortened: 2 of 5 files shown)


edvujic (approved):
Great work Anna, thank you for handling this!

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
```diff
@@ -4,6 +4,11 @@ export const emptyMqaInterval = {
   intervalMetadata: {
     peerReflexiveIP: '0.0.0.0',
     peripherals: [],
+    cpuInfo: {
+      numberOfCores: 1,
+      description: 'web client',
+      architecture: 'unknown',
```
brycetham, line 10: Not something we need to do now, but I think it might be good to have some type checking for architecture since it can only be specific values, even if we're hardcoding it as 'unknown' for now. But we'll handle that when we import the types from the event dictionary when we do the StatsAnalyzer refactor.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -379,6 +377,7 @@ export class StatsAnalyzer extends EventsScope {
     });
 
     newMqa.intervalMetadata.peerReflexiveIP = this.statsResults.connectionType.local.ipAddress;
+    newMqa.intervalMetadata.cpuInfo.numberOfCores = CpuInfo.getNumLogicalCores();
```
brycetham, line 380: We're unlikely to run into this since all the major browsers support this feature, but what happens when `CpuInfo.getNumLogicalCores()` is undefined? I would assume it would fail validation.

I think there was a conversation in the "Ask MQA" space about, if we are unable to get any information about the CPU, we should just leave `newMqa.intervalMetadata.cpuInfo` undefined. Do you think we should do this here?

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
antsukanova: At the beginning, I want to mention that we have a special TS library for types from 'event-dictionary'. As part of our types improvements, we should add this here later. I'd like to create a task specifically for this update so we don't forget. It's important to handle this separately as it may require some changes. We need to be prepared for potential issues or mismatches that may arise.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
antsukanova: Thank you for pointing this out. I believe it's worth handling it as `undefined` too.

packages/@webex/plugin-meetings/package.json
```diff
@@ -74,6 +74,7 @@
     "@webex/media-helpers": "workspace:*",
     "@webex/plugin-people": "workspace:*",
     "@webex/plugin-rooms": "workspace:*",
+    "@webex/web-capabilities": "1.3.0",
```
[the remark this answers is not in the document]
antsukanova: @brycetham I saw that @webex/internal-media-core is used without `^`. So I thought that we do not use `^` to ensure maximum stability and predictability. Do you know is that so?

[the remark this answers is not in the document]
brycetham: For internal-media-core I think we took `^` away because we were changing it constantly and we were worried that we might make a breaking change that would affect the SDK. That's less the case for web-capabilities, so I think we should keep the `^` here. (To be honest I think we should think about putting `^` back into internal-media-core as well, but eventually internal-media-core is going to go away so we can just handle it then.)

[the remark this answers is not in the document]
antsukanova: @brycetham It seems like a good idea to stick to this approach, too. Thank you for your comment.

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -399,6 +397,11 @@ export class StatsAnalyzer extends EventsScope {
 
     newMqa.intervalMetadata.peerReflexiveIP = this.statsResults.connectionType.local.ipAddress;
 
+    const numLogicalCores = CpuInfo.getNumLogicalCores();
+    if (numLogicalCores) {
+      newMqa.intervalMetadata.cpuInfo.numberOfCores = numLogicalCores;
+    }
```
brycetham, line 403: Up to you: I think this could be simplified to a single line `newMqa.intervalMetadata.cpuInfo.numberOfCores = CpuInfo.getNumLogicalCores() || 1;`

brycetham (approved):
Looks good! Left an optional comment.

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
```diff
@@ -4,6 +4,11 @@ export const emptyMqaInterval = {
   intervalMetadata: {
     peerReflexiveIP: '0.0.0.0',
     peripherals: [],
+    cpuInfo: {
+      numberOfCores: 1, // default value from spec if CpuInfo.getNumLogicalCores cannot be determined
+      description: 'web client',
```
Shreyas281299, line 9: According to [types/CpuInfo.raml](https://sqbu-github.cisco.com/WebExSquared/event-dictionary/blob/main/types/CpuInfo.raml), description is  -> "This is a free-form text string representation of the CPU".  However here we are using "web client".  If we have put architecture as undefined maybe we can put desc as undefined as well.

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
```diff
@@ -2113,6 +2114,37 @@ describe('plugin-meetings', () => {
...
         });
       });
-    });
+
+      describe('CPU Information Reporting', async () => {
+        let getNumLogicalCoresStub;
+
+        beforeEach(async () => {
+          if (getNumLogicalCoresStub) {
+            getNumLogicalCoresStub.restore();
```
Shreyas281299, line 2123: I think we can remove this if block because we are already calling a restore in afterEach.

Shreyas281299 (changes requested):

packages/@webex/plugin-meetings/test/unit/spec/stats-analyzer/index.js
Shreyas281299: If it's causing test failures, we can leave it as it is.

antsukanova: @Shreyas281299 thanks, good idea, works without if and I believe it'd be more clearer, good catch

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
antsukanova: @Shreyas281299 It's a valid point. I checked the spec again and it's an enum, so the value 'unknown' seems logical to me too here. I recall that this value was suggested during a conversation with the MQA team, but I will verify with them and hopefully, they will also agree. Thanks

antsukanova: @Shreyas281299 I discussed this with the MQA team, and I put an `NA` value that works for them. Since we can't omit this value, it will be a more suitable string value.

Shreyas281299: Sounds good.

Shreyas281299 (approved):

PR #3620, opened 2024-05-28
Feat/cart: manual caption attendee support

mickelr:
# COMPLETES #< INSERT LINK TO ISSUE >

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-519720
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-519723
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-519725

## This pull request addresses

CART support for attendee

## by making the following changes

1. Handle new mannual transcript event and locus DTO
2 Use v2 final result payload format for voicea transcript

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

### I certified that

- [ ] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [ ] I discussed changes with code owners prior to submitting this pull request

- [ ] I have not skipped any automated checks
- [ ] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3644, opened 2024-06-12
fix(plugin-meetings): update transcription for new voicea payload

mkesavan13:
# COMPLETES [SPARK-533905](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-533905)

As part of this PR: #3620 breaking changes were made to how captions are emitted in the events. This broke the Transcriptions in the plugin-meetings

- Change everything to new payload structure
- Add timestamp in new payload for Voicea final captions
- Update voicea and SDK tests in `plugin-meetings`
- We also have refactored a util method,
    - Earlier, we were processing NEW_CAPTIONS in two data formats, `isFinal:true` came with "transcript" object while `isFinal:false` came with "transcripts" array
    - Now, the breaking change has aligned such that we always get only "transcripts" array
    - Therefore, this method now performs all the caption processing on "transcripts" array and only does the needed changes based on "isFinal" boolean. This means that we process `isFinal:false` as an interim caption while `isFinal:true` as final caption thereby removing the interim caption once we receive the final caption for the respective `transcriptId`
- Additionally, we also made changes to emitting spoken and caption languages as follows,
   - Earlier, the processing of spoken language and caption language wasn't accurate and this PR fixes it based on the SDK value of the caption language they had chosen earlier or set it from what the new caption event responds if not chosen via the SDK already

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

1. Ensured all the unit tests passed locally
2. Tested out from the Kitchen Sink app by enabling transcription and seeing if the live captions are generated without any errors
3. Set the Spoken Language to Hindi and spoke in Hindi. This translated the Caption from whatever I spoke in Hindi to English
4. Similarly set the Spoken Language back to English and the Caption language to Hindi. This translated the caption to Hindi from whatever I spoke in English. Verified the translation with Google Translate also

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
@@ -1,3 +1,22 @@
+type VoiceaTranscripts = {
+  start_mills?: number;
+  end_mills?: number;
+  text: string;
+  last_packet_timestamp_ms?: number;
+  csis: Array<number>;
+  transcript_language_code: string;
+  translations?: {
+    [key: string]: string;
+  };
+  timestamp?: string;
+};
+
+type VoiceaTranscriptPayload = {
+  isFinal: boolean;
+  transcriptId: string;
+  transcripts: Array<VoiceaTranscripts>;
+};
+
 export const getSpeaker = (members, csis = []) =>
   Object.values(members).find((member: any) => {
     const memberCSIs = member.participant.status.csis ?? [];
@@ -31,7 +50,13 @@ export const getSpeakerFromProxyOrStore = ({csisKey, meetingMembers, transcriptD
   return {speaker, needsCaching};
 };
 
-export const processNewCaptions = ({data, meeting}) => {
+export const processNewCaptions = ({
+  data,
+  meeting,
+}: {
+  data: VoiceaTranscriptPayload;
+  meeting: any;
+}) => {
   const {transcriptId} = data;
   const transcriptData = meeting.transcription;
 
--- a/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
@@ -47,7 +47,6 @@ export const processNewCaptions = ({data, meeting}) => {
 
     const newCaption = `${transcriptsPerCsis.get(csisMember)?.text ?? ''} ${text}`.trim();
 
-    // eslint-disable-next-line camelcase
     transcriptsPerCsis.set(csisMember, {
       ...transcript,
       text: newCaption,
--- a/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
@@ -1,21 +1,4 @@
-type VoiceaTranscripts = {
-  start_mills?: number;
-  end_mills?: number;
-  text: string;
-  last_packet_timestamp_ms?: number;
-  csis: Array<number>;
-  transcript_language_code: string;
-  translations?: {
-    [key: string]: string;
-  };
-  timestamp?: string;
-};
-
-type VoiceaTranscriptPayload = {
-  isFinal: boolean;
-  transcriptId: string;
-  transcripts: Array<VoiceaTranscripts>;
-};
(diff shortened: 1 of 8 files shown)


packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
```diff
@@ -81,7 +48,11 @@ export const processNewCaptions = ({data, meeting}) => {
     const newCaption = `${transcriptsPerCsis.get(csisMember)?.text ?? ''} ${text}`.trim();
 
     // eslint-disable-next-line camelcase
```
sreenara, line 50: Why is this disable needed?

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
```diff
@@ -91,31 +62,51 @@ export const processNewCaptions = ({data, meeting}) => {
...
-    const interimId = `${transcriptId}_${speakerId}`;
+
     const captionData = {
       id: interimId,
       isFinal: data.isFinal,
       translations: value.translations,
       text: value.text,
-      currentCaptionLanguage: value.currentSpokenLanguage,
+      currentCaptionLanguage:
+        meeting.transcription?.languageOptions?.currentCaptionLanguage ||
```
sreenara, line 78: Can we explain these changes in the description a little more? I don't think these changes are covered there.
This is also applicable to the (isFinal) block that's been removed from earlier and added below. Can we explain that as well?

Perhaps it would also help to paste a transcription object for interim and final in the description for reference of the reviewer.

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
```diff
@@ -35,39 +35,6 @@ export const processNewCaptions = ({data, meeting}) => {
   const {transcriptId} = data;
```
sreenara, line 35: Do we have any type definition of this `data`?

sreenara (dismissed):
Please see the comments.
Also, please elaborate more on the observation in the tests when we changed the spoken language, e.g what was spoken and what was the result?

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
mkesavan13: Updated it. Not sure why it was there earlier

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
mkesavan13: Done. Please check

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
mkesavan13: We have it in the tests as sample data.

sreenara: @mkesavan13 why aren't we defining a type for this accordingly to be used in src as well? It would be easier to understand the code.

sreenara (commented):
Not a blocking comment but it would be useful for the future.

sreenara (approved):

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
```diff
@@ -91,31 +61,51 @@ export const processNewCaptions = ({data, meeting}) => {
...
+        const interimTranscriptIndex = transcriptData.captions.findIndex(
+          (transcript) => transcript.id === innerInterimId
+        );
+
+        if (interimTranscriptIndex !== -1) {
+          transcriptData.captions.splice(interimTranscriptIndex, 1);
+        }
+      });
+      delete transcriptData.interimCaptions[transcriptId];
+      captionData.id = transcriptId;
```
Shreyas281299, line 107: Would it be better to put the logic to find and remove the interimTranscript data into a separate function and call it here?

Shreyas281299 (approved):
Looks good. Just a minor non breaking comment.

packages/@webex/plugin-meetings/src/meeting/voicea-meeting.ts
mkesavan13: Decided not to move forward after 1:1 Webex discussion that having a separate method just for splicing alone would also complicate it or would render no difference.

PR #3651, opened 2024-06-17
fix(plugin-meetings): isActiveSpeaker MQE fix

antsukanova:
# COMPLETES [WEBEX-379504](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-379504)

Currently, `isActiveSpeaker` is always `true` in the MQE, which is not correct

Fix for MQE `isActiveSpeaker`

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] automated tests for plugin-meetings
- [x] SDK test meeting correctly [sends MQE](https://gist.github.com/antsukanova/64a55a9f667248498124ba9bafd83fda) in network tab

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -225,7 +225,7 @@ export const getBuildType = (
  */
 export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
   const origin: Partial<Event['origin']> = {
-    buildType: exports.getBuildType(
+    buildType: getBuildType(
```
antsukanova, line 228: temporary for your convenience

packages/@webex/media-helpers/package.json
```diff
@@ -22,7 +22,7 @@
     "deploy:npm": "yarn npm publish"
   },
   "dependencies": {
-    "@webex/internal-media-core": "2.5.0",
+    "@webex/internal-media-core": "link:../../../../webrtc-media-core",
```
antsukanova, line 25: also temporary for testing

packages/@webex/plugin-meetings/src/mediaQualityMetrics/config.ts
```diff
@@ -165,7 +165,7 @@ export const emptyVideoReceiveStream = {
     ssci: 0, // Not avaliable
   },
   h264CodecProfile: 'BP',
-  isActiveSpeaker: true,
+  isActiveSpeaker: false,
```
antsukanova, line 168: So far, from my understanding, we need only set `isActiveSpeaker` to `true` when the person on a call is talking, and we see this active video from the user, right? Or did I miss something?

@edvujic @SomeBody16

szotrabh: Or if the person was active/ speaking for the last 60 sec?

antsukanova: @szotrabh Yes, so my point of view is that the user initializes with `false`. When a person is an active speaker, we need to be able to send out information that this user was the active speaker within the last 60 seconds. There is logic involved in calculating the time for that.

SomeBody16 (approved):

edvujic (approved):
LGTM! Thanks for doing this :)

packages/@webex/plugin-meetings/src/statsAnalyzer/mqaUtil.ts
```diff
@@ -336,6 +336,18 @@ export const getVideoReceiverStreamMqa = ({
...
   videoReceiverStream.requestedKeyFrames =
     statsResults[mediaType][sendrecvType].totalPliCount - lastPliCount || 0;
+
+  if (statsResults[mediaType][sendrecvType].isActiveSpeaker !== undefined) {
+    if (statsResults[mediaType][sendrecvType].lastActiveSpeakerTimestamp !== undefined) {
+      const currentTimestamp =
+        statsResults[mediaType][sendrecvType].lastActiveSpeakerTimestamp + MQA_INTERVAL;
+      // check if last true active speaker was set at least once during the last MQA interval
+      videoReceiverStream.isActiveSpeaker =
+        currentTimestamp > performance.now() + performance.timeOrigin;
```
brycetham, line 346: This calculation is confusing to me, probably because `currentTimestamp` sounds like it should be the time that it is right now.

So I would personally write it this way:

```
const currentTimestamp = performance.now() + performance.timeOrigin;
videoReceiverStream.isActiveSpeaker = statsResults[mediaType][sendrecvType].lastActiveSpeakerTimestamp > currentTimestamp - MQA_INTERVAL;
```

Let's also write some unit tests for this to make sure we cover all cases, since this is a bit tricky. In particular:
- A user was constantly an active speaker for the last 5 minutes and did not change their active speaker state within the last interval (`isActiveSpeaker` MQE should be true)
- A user was never an active speaker for the last 5 minutes and did not change their active speaker state within the last interval (`isActiveSpeaker` MQE should be false)
- A user changed their active speaker state within the last interval (`isActiveSpeaker` MQE should be true)

antsukanova: @brycetham Thank you for the ideas. I agree with you on the naming. Your suggestion also seems more natural and logical, so I will make the change. Regarding tests, I plan to add them after your review, so your test cases will also be useful to consider. This logic definitely should be covered with tests. Thanks for the review.

PR #3653, opened 2024-06-18
fix(calling): switching off stream listeners to avoid having multiple listeners

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-511584 

While updating the local track, a DOM exception error was encountered complaining for the peer connection to be closed. Hence enabling BNR fails after the first call. This happens because the old listener is used for the OutputTrackChange event on the localAudioStream, which uses the previous media connection, which is closed while updating the local track.

index.js:6603 Uncaught (in promise) DOMException: The peer connection is closed.
    at https://localhost:8000/samples/calling.min.js:2:3712596
    at Array.forEach (<anonymous>)
    at r.value (https://localhost:8000/samples/calling.min.js:2:3712231)
    at r.value (https://localhost:8000/samples/calling.min.js:2:3713170)
    at r.value (https://localhost:8000/samples/calling.min.js:2:3847575)
    at Je.<anonymous> (https://localhost:8000/samples/calling.min.js:2:3976135)
    at Ne.emit (https://localhost:8000/samples/calling.min.js:2:3059270)
    at Xe.emit (https://localhost:8000/samples/calling.min.js:2:3061589)
    at r.changeOutputTrack (https://localhost:8000/samples/calling.min.js:2:3064449)
    at NoiseReductionEffect.t (https://localhost:8000/samples/calling.min.js:2:3065066)

Steps to reproduce:

1. Initiate a call, enable BNR then disable BNR followed by call disconnect. Everything works fine till this point.
2. Initiated 2nd call and enabled BNR, updating local track fails and it complains of peer connection being closed.

Switching off the localAudioStream listener for the OutputTrackChange event fixes this issue.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Initiate a call, enable BNR then disable BNR followed by call disconnect. Everything works fine till this point.
2. Initiated 2nd call and enabled BNR, enabling BNR works smoothly and no error is observed.

[DOM_Exception_Fix.log](https://github.com/user-attachments/files/15895528/DOM_Exception_Fix.log)

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/calling/call.ts
+++ b/packages/calling/src/CallingClient/calling/call.ts
@@ -2467,6 +2467,10 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     );
   };
 
+  private updateTrack = (audioTrack: MediaStreamTrack) => {
+    this.mediaConnection.updateLocalTracks({audio: audioTrack});
+  };
+
   private registerEffectListener = (addedEffect: TrackEffect) => {
     if (this.localAudioStream) {
       const effect = this.localAudioStream.getEffectByKind(NOISE_REDUCTION_EFFECT);
@@ -2488,13 +2492,12 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
       }
 
       this.localAudioStream.off(LocalStreamEventNames.EffectAdded, this.registerEffectListener);
+      this.localAudioStream.off(LocalStreamEventNames.OutputTrackChange, this.updateTrack);
     }
   }
 
   private registerListeners(localAudioStream: LocalMicrophoneStream) {
-    localAudioStream.on(LocalStreamEventNames.OutputTrackChange, (audioTrack: MediaStreamTrack) => {
-      this.mediaConnection.updateLocalTracks({audio: audioTrack});
-    });
+    localAudioStream.on(LocalStreamEventNames.OutputTrackChange, this.updateTrack);
 
     localAudioStream.on(LocalStreamEventNames.EffectAdded, this.registerEffectListener);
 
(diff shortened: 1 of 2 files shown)


Shreyas281299 (approved):
Looks good! LGTM

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2488,13 +2492,12 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
       this.localAudioStream.off(LocalStreamEventNames.EffectAdded, this.registerEffectListener);
+      this.localAudioStream.off(LocalStreamEventNames.OutputTrackChange, this.updateTrack);
     }
   }
 
   private registerListeners(localAudioStream: LocalMicrophoneStream) {
-    localAudioStream.on(LocalStreamEventNames.OutputTrackChange, (audioTrack: MediaStreamTrack) => {
-      this.mediaConnection.updateLocalTracks({audio: audioTrack});
-    });
+    localAudioStream.on(LocalStreamEventNames.OutputTrackChange, this.updateTrack);
```
sreenara, line 2500: We haven't added an expect for this statement. Is there a reason for that?

sreenara (changes requested):
In the PR description where you are talking about the problem, can you please mention the sequence of events that leads to this issue?
Also, please mention the tests performed in the PR description.

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: I have added it

Kesari3008:
> In the PR description where you are talking about the problem, can you please mention the sequence of events that leads to this issue? Also, please mention the tests performed in the PR description.

PR description updated

packages/calling/src/CallingClient/calling/call.ts
sreenara: Could you point to the line where this change has been made? I am not able to find it.

sreenara (changes requested):
The description of the problem still needs a change. The details that are provided in the test should be present in the problem description.
The test scenarios should contain only the observations with the fix.

sreenara (approved):
LGTM. Thanks for helping test this.

PR #3657, opened 2024-06-20
fix(webex): remove plugin-presence from src

mkesavan13:
# COMPLETES AdHoc

As part of this PR: #3645, we removed `@webex/plugin-presence` from the package thereby avoiding conflict while doing `yarn install`

This however is still not removed from the source and is causing issues when trying to import the `webex` package

- Removing @webex/plugin-presence from the src/webex.js main file also 

**When trying to bundle the webex package using parcel bundler:**

Screenshot 2024-06-21 at 2 24 17 AM

---

**When trying to use the SDK on a node-js app:**

Screenshot 2024-06-21 at 2 22 20 AM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/legacy/tools/src/utils/server.ts
+++ b/packages/legacy/tools/src/utils/server.ts
@@ -5,8 +5,6 @@
 
 // import * as path from 'path';
 import { spawn } from 'child_process';
-import * as fs from 'fs';
-import * as path from 'path';
 
 const debug = require('debug')('monorepo:test:server');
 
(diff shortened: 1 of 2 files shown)


packages/legacy/tools/src/utils/server.ts
```diff
@@ -37,7 +37,12 @@ export async function startServer(): Promise<void> {
 
   return new Promise((resolve) => {
     // TODO:  move the logic for spawn the server to test-helper-server
-    const serverPath = '../../@webex/test-helper-server';
+    let serverPath;
```
sreenara, line 40: What does this change do?

sreenara (changes requested):
Looks like a miss when I removed plugin-presence from package.json.
I have a question regarding a change in the server.ts
Looks like the coverage has failed. Please change the test accordingly.

packages/legacy/tools/src/utils/server.ts
mkesavan13: Basically we were setting a relative path for server that caused the original error. This was an attempt to fix that error which threw another error. 

That error requires more time to investigate and fix. Therefore, created a Jira here for the same: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-535983

I have reverted the changes here.

Shreyas281299 (approved):

PR #3658, opened 2024-06-24
fix: circular dependency b/w feature and mercury

mkesavan13:
# COMPLETES [SPARK-515520](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-515520)

The final circular dependency that is left out. This PR cuts off the circular dependency between `internal-plugin-mercury` and `internal-plugin-feature` 

- Remove the `internal-plugin-mercury` dependency from `internal-plugin-feature`
- On successful connect of `webex.mercury.connect` event, the mercury object listens to the `event_featureToggle`
- Whenever this event is called, the handler checks if it has a legitimate envelope and identifies the feature and sends it to feature plugin by calling `webex.internal.feature.updateFeature`
- This `webex.internal.feature.updateFeature` is the method that is modified from earlier `handleFeatureUpdate` call back listener

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-feature/src/feature.js
+++ b/packages/@webex/internal-plugin-feature/src/feature.js
@@ -47,7 +47,7 @@ const Feature = WebexCore.WebexPlugin.extend({
    * @param {Object} envelope
    * @returns {undefined}
    */
-  handleFeatureUpdate(envelope) {
+  updateFeature(envelope) {
     if (envelope && envelope.data) {
       const feature = envelope.data.featureToggle;
       const keyType = feature.type.toLowerCase();
@@ -56,21 +56,6 @@ const Feature = WebexCore.WebexPlugin.extend({
         this.webex.internal.device.features[keyType].add([feature], {merge: true});
       }
     }
-  },
-
-  /**
-   * Register to listen for incoming feature events
-   * @instance
-   * @returns {undefined}
-   */
-  listen() {
-    if (this.mercury) {
-      this.listenTo(this.mercury, 'event:featureToggle_update', this.handleFeatureUpdate);
-      this.isListeningToMercury = true;
-
-      return;
-    }
-    this.isListeningToMercury = false;
   },
 
   /**
@@ -142,13 +127,6 @@ const Feature = WebexCore.WebexPlugin.extend({
       'change:internal.device.features.user',
       this.trigger.bind(this, 'change:user')
     );
-  },
-
-  setMercury(mercury) {
-    this.mercury = mercury;
-    if (typeof this.isListeningToMercury !== 'undefined' && !this.isListeningToMercury) {
-      this.listen();
-    }
   },
 });
 
--- a/packages/@webex/internal-plugin-feature/src/feature.js
+++ b/packages/@webex/internal-plugin-feature/src/feature.js
@@ -43,6 +43,11 @@ const Feature = WebexCore.WebexPlugin.extend({
 
   /**
    * Handles a feature toggle update from the server.
+   * When one of these legacy feature gets updated, this event handler be triggered
+   *     * group-message-notifications
+   *     * mention-notifications
+   *     * thread-notifications,
+   * This handler is being listened from mercury connection success method for the event "event:featureToggle_update"
    * @param {Object} envelope
    * @returns {undefined}
    */
--- a/packages/@webex/internal-plugin-feature/src/feature.js
+++ b/packages/@webex/internal-plugin-feature/src/feature.js
@@ -8,6 +8,7 @@ import * as WebexCore from '@webex/webex-core';
 
 const Feature = WebexCore.WebexPlugin.extend({
   namespace: 'Feature',
+  isListeningToMercury: undefined,
 
   /**
    * Returns the value of the requested feature toggle.
@@ -63,11 +64,13 @@ const Feature = WebexCore.WebexPlugin.extend({
    * @returns {undefined}
    */
   listen() {
-    this.listenTo(
-      this.webex.internal.mercury,
-      'event:featureToggle_update',
-      this.handleFeatureUpdate
-    );
+    if (this.mercury) {
+      this.listenTo(this.mercury, 'event:featureToggle_update', this.handleFeatureUpdate);
+      this.isListeningToMercury = true;
+
+      return;
+    }
+    this.isListeningToMercury = false;
   },
 
   /**
@@ -139,6 +142,13 @@ const Feature = WebexCore.WebexPlugin.extend({
       'change:internal.device.features.user',
       this.trigger.bind(this, 'change:user')
     );
+  },
+
+  setMercury(mercury) {
+    this.mercury = mercury;
+    if (typeof this.isListeningToMercury !== 'undefined' && !this.isListeningToMercury) {
+      this.listen();
+    }
   },
 });
 
--- a/packages/@webex/internal-plugin-feature/src/feature.js
+++ b/packages/@webex/internal-plugin-feature/src/feature.js
@@ -8,7 +8,6 @@ import * as WebexCore from '@webex/webex-core';
 
 const Feature = WebexCore.WebexPlugin.extend({
   namespace: 'Feature',
-  isListeningToMercury: undefined,
 
   /**
    * Returns the value of the requested feature toggle.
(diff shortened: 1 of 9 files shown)


packages/@webex/internal-plugin-feature/src/feature.js
```diff
@@ -63,11 +64,13 @@ const Feature = WebexCore.WebexPlugin.extend({
    * @returns {undefined}
    */
   listen() {
-    this.listenTo(
-      this.webex.internal.mercury,
-      'event:featureToggle_update',
-      this.handleFeatureUpdate
-    );
+    if (this.mercury) {
+      this.listenTo(this.mercury, 'event:featureToggle_update', this.handleFeatureUpdate);
```
marcin-bazyl, line 68: this doesn't feel like really fully fixing the circular dependency, as the feature plugin will still be listening on the events from mercury and needs to know what these events are. I think, if we wanted to do this fully we should instead of adding a `setMercury()` method, just add a method that mercury calls when a feature is updated -  it would be basically almost the same as handleFeatureUpdate() method, just the arguments should be changed from `envelope` to `feature` and maybe it could be renamed to `featureUpdated()` or `setFeature()` or `updateFeature()` or something like that

marcin-bazyl: hmm.... not sure... but probably should be ok. We don't need that call to `listen()` anymore, because the features plugin will always listen - it will always get called from Mercury with a new feature toggle. The only possible issue is what happens if a feature toggle is updated before web app calls `webex.internal.feature.listen()` - I think with the current code as it is the web app would just miss that event anyway, so with the new code Mercury might call features plugin handleFeatureUpdate() before web app registers the listeners for the feature toggle events, but I think it's just the same is it currently is, and hasn't caused any problems, so probably should be fine.... Can you do some testing with the web app with this change and the SDK changes and change some toggle that web app listens for (for example "web-tcp-reachability") and check that it still works correctly (and configures sdk to do tcp reachability)?

packages/@webex/test-helper-mock-webex/src/index.js
```diff
@@ -282,6 +282,7 @@ function makeWebex(options) {
     feature: {
       setFeature: sinon.stub().returns(Promise.resolve(false)),
       getFeature: sinon.stub().returns(Promise.resolve(false)),
+      setMercury: sinon.stub(),
```
Shreyas281299, line 285: I think we can remove this as we have changed the approach.

packages/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -148,6 +150,9 @@ describe('plugin-mercury', () => {
           assert.isTrue(mercury.connected, 'Mercury is connected');
           assert.isFalse(mercury.connecting, 'Mercury is not connecting');
           assert.calledWith(socketOpenStub, sinon.match(/ws:\/\/example.com/), sinon.match.any);
+          mercury._emit('event:featureToggle_update', updatedFeature)
+          assert.calledOnce(webex.internal.feature.updateFeature);
+          assert.calledWith(webex.internal.feature.updateFeature, updatedFeature);
```
Shreyas281299, line 155: Can we not user assert.calledOncewith() here?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -312,6 +312,7 @@ const Mercury = WebexPlugin.extend({
         this.connected = true;
         this.hasEverConnected = true;
         this._emit('online');
+        this.on('event:featureToggle_update', this.webex.internal.feature.updateFeature);
```
Shreyas281299, line 315: Should we have `this.off()` to clear this listener?

Shreyas281299 (changes requested):
Just a couple of comments.

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Maybe in `_onClose` we can clear this listener.
[here](https://github.com/webex/webex-js-sdk/blob/34340d71ce38227676d37ba081d76e57bc3ddff7/packages/%40webex/internal-plugin-mercury/src/mercury.js#L399)

packages/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
mkesavan13: There was no `calledOnceWith`

However found a `calledOnceWithExactly` and changed

packages/@webex/test-helper-mock-webex/src/index.js
mkesavan13: Oops. Thanks for noticing. Removed

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Okay. If the event handler will be cleaned up automatically then it's alright.  I just wanted to make sure we dont have unused event handlers.

packages/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
```diff
@@ -138,7 +138,9 @@ describe('plugin-mercury', () => {
       });
 
       it('connects to Mercury using default url', () => {
+        webex.internal.feature.updateFeature = sinon.stub();
```
Shreyas281299, line 141: Can we `webex.internal.feature.updateFeature.restore()` the stub at the end of the test? However, we are calling the sandbox. restore but while fixing the test I've noticed it doesn't get cleared sometimes.

Shreyas281299 (commented):
Please see the comment before merging.

Shreyas281299 (approved):

packages/@webex/internal-plugin-feature/src/feature.js
mkesavan13: I got info from the mercury and the feature Ask spaces and triggered this event handler by updating one of the following features:
* group-message-notifications
* mention-notifications
* thread-notifications

Looks like that event is triggered only for these three legacy toggles. I tested these changes with the WWC changes as well here: https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/5975

The event is triggered alright. Kindly re-review

@marcin-bazyl

packages/@webex/internal-plugin-mercury/src/mercury.js
marcin-bazyl: hmm I'm not sure about this.... the event is registered on the `this` instance and `this` is the whole Mercury plugin, so it won't be destroyed when the connection is closed, while this code is executed on every new connection, so I think what will happen is that every time we reconnect mercury connection, we will register the new event handler for 'event:featureToggle_update' and never unregister the old one

packages/@webex/internal-plugin-feature/src/feature.js
```diff
@@ -43,10 +43,15 @@ const Feature = WebexCore.WebexPlugin.extend({
...
+   * When one of these legacy feature gets updated, this event handler be triggered
+   *     * group-message-notifications
+   *     * mention-notifications
+   *     * thread-notifications,
+   * This handler is being listened from mercury connection success method for the event "event:featureToggle_update"
    * @param {Object} envelope
    * @returns {undefined}
    */
-  handleFeatureUpdate(envelope) {
+  updateFeature(envelope) {
```
marcin-bazyl, line 54: this is an API of internal-plugin-feature and that plugin shouldn't need to know anything about how the event data coming from mercury is formatted, so I think it would be better if the argument was not `envelope`, because that's a mercury thing, but instead just have the `feature` as the argument

mkesavan13: Done @marcin-bazyl 

Please check

packages/@webex/internal-plugin-mercury/src/mercury.js
mkesavan13: Thanks, @marcin-bazyl 

It makes sense now. 

@Shreyas281299 - Sorry. I didn't realize I've setup the event listener in connect which is prone to calling more than once.

@marcin-bazyl - I've updated the code to turn on the event handler at the time of `initialize` which is like a Class's constructor in AmpState

Kindly check. Thank you.

packages/@webex/internal-plugin-mercury/test/unit/spec/mercury.js
mkesavan13: I've added `sinon.restore` at the end of the test(s)

Looks like `webex.internal.feature.updateFeature.restore()` isn't a valid method. Thanks, @Shreyas281299

marcin-bazyl (approved):

PR #3665, opened 2024-06-26
fix: missed xtls reachability accounting in isAnyPublicClusterReachable

k-wasniowski:
# COMPLETES #[SPARK-535587](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-535587)

This PR adds missing taking into account if we should do turn discovery based on reachability results. Now it's not particularly used, since we do turn discovery always.

Will calculate need of turn discovery with xTLS Servers (TurnTLS)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested on JS-SDK Plugin Meeting App

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -284,7 +284,9 @@ export default class Reachability {
         reachable = Object.values(reachabilityResults).some(
           (result) =>
             !result.isVideoMesh &&
-            (result.udp?.result === 'reachable' || result.tcp?.result === 'reachable')
+            (result.udp?.result === 'reachable' ||
+              result.tcp?.result === 'reachable' ||
+              result.xtls?.result === 'reachable')
         );
       } catch (e) {
         LoggerProxy.logger.error(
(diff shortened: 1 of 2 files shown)


marcin-bazyl (approved):
Code change looks good, but the PR description is not accurate. The isAnyPublicClusterReachable() method was used for deciding if TURN discovery needs to be done or not, Now web app doesn't use it at all, because we call joinWithMedia() which always does TURN discovery, but potentially other apps that call join() and addMedia() may be impacted by this

Shreyas281299 (commented):
@marcin-bazyl  @k-wasniowski  I don't have much context about turndiscovery so can you please elaborate on how this will impact the joining of the meeting? Is there any scenario that you are expecting this to break?

marcin-bazyl (changes requested):
Actually, thinking about it a bit more I've realised that we don't want this change. `isAnyPublicClusterReachable()` is currently checking only UDP and TCP reachability results and this PR adds TLS checks to the logic. When `isAnyPublicClusterReachable()` returns `true`, we don't do TURN discovery when `addMedia()` is called. So, in a case where UDP and TCP reachability fail but TLS reachability succeeds, before this PR `isAnyPublicClusterReachable()` would return `false` and we would do TURN discovery and potentially use TURN-TLS connection for the meeting, but after this PR `isAnyPublicClusterReachable()` would return `true` and we wouldn't do TURN discovery, potentially preventing the user from having a working media connection. Instead of this PR, maybe we should just rename `isAnyPublicClusterReachable()` to `isAnyPublicClusterReachableWithoutTurnTLS()` or `isAnyPublicClusterReachableOverUdpTcp()` or something like that?

PR #3668, opened 2024-06-27
fix(plugin-meetings): trigger transcript connected after llm online

mkesavan13:
# COMPLETES AdHoc

So, while joining a meeting using the Web SDK, we check if the `enableAutomaticLLM` is set to true and connect the LLM  websocket and once we do this, we notify the customer through an event `meeting:transcription:connected`

We do this once the `updateLLMConnection` method is resolved. However, this is inaccurate sometimes leading to early emission of the event

To avoid such inaccuracy, this change will here-on listen to the 'online' event emitted by the LLM plugin and then trigger the `meeting:transcription:connected` event

### Screenrecording

Here's [link to the sharepoint upload](https://cisco-my.sharepoint.com/:v:/p/kmadavan/EUEyyMAsfDJNhsFgdQwS-nEB1Dheba4JmIeONRGf7MHnRw?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0NvcHkifX0&e=IVG4eC).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Wrote unit tests for the changes
- Updated the code and tested it by enabling Transcription from the sample app to see if the event is triggered
- Also tried to speak a few words to ensure we receive the transcription payload

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4954,8 +4954,10 @@ export default class Meeting extends StatelessWebexPlugin {
   }
 
   /**
-   * This callback is called when LLM comes online
-   * This method in turn will notify the meetings developer that LLM is connected
+   * This is a callback for the LLM event that is triggered when it comes online
+   * This method in turn will trigger an event to the developers that the LLM is connected
+   * @private
+   * @memberof Meeting
    * @returns {null}
    */
   private handleLLMOnline = (): void => {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4954,6 +4954,25 @@ export default class Meeting extends StatelessWebexPlugin {
   }
 
   /**
+   * This callback is called when LLM comes online
+   * This method in turn will notify the meetings developer that LLM is connected
+   * @returns {null}
+   */
+  private handleLLMOnline = (): void => {
+    // @ts-ignore
+    this.webex.internal.llm.off('online', this.handleLLMOnline);
+    Trigger.trigger(
+      this,
+      {
+        file: 'meeting/index',
+        function: 'handleLLMOnline',
+      },
+      EVENT_TRIGGERS.MEETING_TRANSCRIPTION_CONNECTED,
+      undefined
+    );
+  };
+
+  /**
    * Specify joining via audio (option: pstn), video, screenshare
    * @param {JoinOptions} options A configurable options object for joining a meeting
    * @returns {Promise} the join response
@@ -5172,6 +5191,8 @@ export default class Meeting extends StatelessWebexPlugin {
       .then((join) => {
         // @ts-ignore - config coming from registerPlugin
         if (this.config.enableAutomaticLLM) {
+          // @ts-ignore
+          this.webex.internal.llm.on('online', this.handleLLMOnline);
           this.updateLLMConnection()
             .catch((error) => {
               LoggerProxy.logger.error(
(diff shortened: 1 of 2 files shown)


Kesari3008 (commented):
Can we update some sort of screenshot or logs for the test performed via sample app?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4953,6 +4953,25 @@ export default class Meeting extends StatelessWebexPlugin {
     );
   }
 
+  /**
+   * This callback is called when LLM comes online
+   * This method in turn will notify the meetings developer that LLM is connected
+   * @returns {null}
+   */
+  private handleLLMOnline = (): void => {
```
sreenara, line 4961: This function description doesn't conform with the remaining function descriptions in this class. Can we align it accordingly?

sreenara (changes requested):
> However, this is inaccurate sometimes leading to early emission of the event

Have we found out when the resolution of this promise is inaccurate? 
Also, was there a design change on the LLM side that caused this issue?
Third, is there an `offline` event analogous to the `online` event which we need to take care of? What if the LLM goes offline and we need to report this to the user and wait for online again? I see that we are switching off the listener in the function that's added.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: I have updated the description with more info. Please check.

mkesavan13:
@sreenara 

**Have we found out when the resolution of this promise is inaccurate?**
- It is only matter of time(speed) in which code gets executed. At the moment, as soon as the promise resolves, we trigger the SDK event to Instant Connect and Instant Connect starts the transcription
- Now, although the LLM is connected at that time, the parameter in the SDK would still say it is not connected and when they start transcription, we check if LLM is connected through this boolean and it says not connected at the times of lag.
- On the other hand, with this change, if we listen to the online event of llm and then trigger our event for Instant connect, we can be sure that LLM is connected already

**Also, was there a design change on the LLM side that caused this issue?**
- No, I believe. Just our issue of how we handled it earlier.

**Third, is there an offline event analogous to the online event which we need to take care of? What if the LLM goes offline and we need to report this to the user and wait for online again? I see that we are switching off the listener in the function that's added.**
- I think if the LLM goes offline and comes back online by itself, transcription will just continue to work.
- However, there is an offline event that we can make use of to notify user. Should we do that too?

mkesavan13:
@Kesari3008 - Updated the PR description with recording share point link. Please check

packages/@webex/plugin-meetings/src/meeting/index.ts
sreenara:
```suggestion
  private function handleLLMOnline(): void {
```

sreenara (approved):
Please see the suggestion for the code in terms of convention.

mkesavan13:
Oh. We cannot do that. So, we are just passing on this function to the `llm.on` as callback. However, if we do not have it as an arrow function, TriggerProxy would think we refer 'this' as `LLM` and not `Meetings`

For arrow functions, `this` always is where it is declared. Whereas for regular functions, it takes `this` as the object from where it is being called.

We cannot also bind it. If you see the test file, we test assert.calledWith(llm.on, 'online', meeting.handleLLMOnline)

If we bind it, llm.on would receive a bound function and not the actual meeting.handleLLMOnline

And if we do assert.calledWith(llm.on, 'online', meeting.handleLLMOnline.bind(meeting)) in the test, this would be considered as a separate bound function and not the same as it is bound in the meeting.

You can check `processRelayEvent` method as an example of this type of declarations in the same file

@sreenara

Kesari3008 (approved):

PR #3670, opened 2024-06-28
fix(plugin-meetings): update multistreamEnabled

antsukanova:
# COMPLETES # [WEBEX-389398](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-389398)

Currently, `multistreamEnabled` has the wrong value for MQE. It should be `true` for **multistream** and `false` for **transcoded** meetings

Fix value for MQE `multistreamEnabled`, add tests

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] automated tests for plugin-meetings
- [x] SDK test meeting correctly [sends MQE](https://gist.github.com/antsukanova/367800e16040da8360dbd8abf658eb55) in automated diagnostics

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -6317,12 +6317,13 @@ export default class Meeting extends StatelessWebexPlugin {
     if (this.config.stats.enableStatsAnalyzer) {
       // @ts-ignore - config coming from registerPlugin
       this.networkQualityMonitor = new NetworkQualityMonitor(this.config.stats);
-      this.statsAnalyzer = new StatsAnalyzer(
+      this.statsAnalyzer = new StatsAnalyzer({
         // @ts-ignore - config coming from registerPlugin
-        this.config.stats,
-        (ssrc: number) => this.receiveSlotManager.findReceiveSlotBySsrc(ssrc),
-        this.networkQualityMonitor
-      );
+        config: this.config.stats,
+        receiveSlotCallback: (ssrc: number) => this.receiveSlotManager.findReceiveSlotBySsrc(ssrc),
+        networkQualityMonitor: this.networkQualityMonitor,
+        isMultistream: this.isMultistream,
+      });
       this.setupStatsAnalyzerEventHandlers();
       this.networkQualityMonitor.on(
         EVENT_TRIGGERS.NETWORK_QUALITY,
--- a/packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
+++ b/packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
@@ -92,22 +92,31 @@ export class StatsAnalyzer extends EventsScope {
   successfulCandidatePair: any;
   localIpAddress: string; // Returns the local IP address for diagnostics. this is the local IP of the interface used for the current media connection a host can have many local Ip Addresses
   receiveSlotCallback: ReceiveSlotCallback;
+  isMultistream: boolean;
 
   /**
    * Creates a new instance of StatsAnalyzer
    * @constructor
    * @public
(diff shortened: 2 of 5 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6317,12 +6317,13 @@ export default class Meeting extends StatelessWebexPlugin {
     if (this.config.stats.enableStatsAnalyzer) {
       // @ts-ignore - config coming from registerPlugin
       this.networkQualityMonitor = new NetworkQualityMonitor(this.config.stats);
-      this.statsAnalyzer = new StatsAnalyzer(
```
antsukanova, line 6320: for better DX changed arguments to object

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -198,6 +208,21 @@ export class StatsAnalyzer extends EventsScope {
     const videoReceiver = cloneDeep(emptyVideoReceive);
     const videoShareReceiver = cloneDeep(emptyVideoReceive);
 
+    if (this.isMultistream) {
```
antsukanova, line 211: @brycetham  This approach makes changing the `multistreamEnabled` clearer than passing it to all utility functions. We also have some calculations for MQE data not only with the utility function. This results in the need to pass and change all instances of "audioSender, audioShareSender, audioReceiver, audioShareReceiver, videoSender, videoShareSender, videoReceiver, videoShareReceiver" with the value of `this.isMultistream.` I don't feel good about its way of leaking logic.

brycetham: Hmm in my opinion, I still think it makes sense to pass it to the MQA util functions. Even though it does require tweaking the params for all of them (that's more of a consequence of these functions written poorly which we'll fix in the refactor), I like it better for a few reasons:

1. All the logic related to populating the MQE is in the util functions, so it's good to have them all in the same place.
2. In the refactor, we'll want to pass `isMultistream` into the MQE builders anyways (though we'd do this in the constructor which is much cleaner). So having it in the util functions would more closely align with what we'd end up doing in the refactor.

antsukanova: @brycetham updated with the discussed version, it looks good to me, please tell me how you feel about that

packages/@webex/plugin-meetings/src/statsAnalyzer/index.ts
```diff
@@ -91,22 +91,31 @@ export class StatsAnalyzer extends EventsScope {
...
+    receiveSlotCallback = () => undefined,
+    networkQualityMonitor = {},
+    statsResults = defaultStats,
+    isMultistream = false,
+  }: {
+    config: any;
+    receiveSlotCallback: ReceiveSlotCallback;
+    networkQualityMonitor: any;
+    statsResults?: any;
+    isMultistream?: boolean;
```
brycetham, line 117: Noticed this is optional here. Should we worry about the case in which `isMultistream` is undefined? Then `multistreamEnabled` in the MQE would be undefined as well I think, and I'm not sure if that would fail validation.

antsukanova: I use `false` as a fallback here, so far it will always work with either `false` or `true`, you can also notice that `StatsAnalyzer` `isMultistream` works only with `boolean` values.
Screenshot 2024-06-28 at 16 51 31

brycetham: Ah awesome, I didn't notice that. Thanks!

brycetham (approved):

PR #3671, opened 2024-06-30
fix(internal-plugin-metrics): remove exports.getBuildType error

mkesavan13:
# COMPLETES [SPARK-535057](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-535057)

Removal of the non-blocking `exports.getBuildType` error that appears in the logs which was annoying

As per [this ticket](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-470747), the `exports` was added to the `getBuildType` call so that it can be stubbed and tested in the unit tests. To address this, I've changed the getBuildType code to show a log that this method was called and then test `webex.logger.log` instead of `getBuildType`

The aim was to know if this method was called with the right attributes and that is accomplished

**Logs before and after**

| Before | After |
| --- | --- |
| [webex.github.io-1719717587011.log](https://github.com/user-attachments/files/16043036/webex.github.io-1719717587011.log) | [pr-3671.d3m3l2kee0btzx.amplifyapp.com-1719717796375.log](https://github.com/user-attachments/files/16043035/pr-3671.d3m3l2kee0btzx.amplifyapp.com-1719717796375.log) |

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

I opened the sample app locally and tried joining a meeting and by this time, I already saw several logs for getBuildType being called and no errors that appeared earlier

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
@@ -198,16 +198,9 @@ export const isBrowserMediaErrorName = (errorName: any) => {
  * @returns
  */
 export const getBuildType = (
-  webex,
   webClientDomain,
   markAsTestEvent = false
 ): Event['origin']['buildType'] => {
-  webex.logger.log(
-    `CallDiagnosticMetricsUtil: getBuildType is called`,
-    webClientDomain,
-    markAsTestEvent
-  );
-
   // used temporary to test pre join in production without creating noise data, SPARK-468456
   if (markAsTestEvent) {
     return 'test';
@@ -233,7 +226,6 @@ export const getBuildType = (
 export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
   const origin: Partial<Event['origin']> = {
     buildType: getBuildType(
-      webex,
       item.eventPayload?.event?.eventData?.webClientDomain,
       item.eventPayload?.event?.eventData?.markAsTestEvent
     ),
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
@@ -198,9 +198,16 @@ export const isBrowserMediaErrorName = (errorName: any) => {
  * @returns
  */
 export const getBuildType = (
+  webex,
   webClientDomain,
   markAsTestEvent = false
 ): Event['origin']['buildType'] => {
+  webex.logger.log(
+    `CallDiagnosticMetricsUtil: getBuildType is called`,
+    webClientDomain,
+    markAsTestEvent
+  );
+
   // used temporary to test pre join in production without creating noise data, SPARK-468456
   if (markAsTestEvent) {
     return 'test';
@@ -225,7 +232,8 @@ export const getBuildType = (
  */
 export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
   const origin: Partial<Event['origin']> = {
-    buildType: exports.getBuildType(
+    buildType: getBuildType(
+      webex,
       item.eventPayload?.event?.eventData?.webClientDomain,
       item.eventPayload?.event?.eventData?.markAsTestEvent
     ),
@@ -314,7 +322,6 @@ export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
 
   item.eventPayload.origin = Object.assign(origin, item.eventPayload.origin);
 
-  // @ts-ignore
   webex.logger.log(
     `CallDiagnosticLatencies,prepareDiagnosticMetricItem: ${JSON.stringify({
       latencies: Object.fromEntries(cdl.latencyTimestamps),
--- a/packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
+++ b/packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
@@ -234,8 +234,14 @@ describe('internal-plugin-metrics', () => {
   });
 
   describe('getBuildType', () => {
+    let webex;
     beforeEach(() => {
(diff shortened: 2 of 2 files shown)


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -198,9 +198,16 @@ export const isBrowserMediaErrorName = (errorName: any) => {
  * @returns
  */
 export const getBuildType = (
+  webex,
   webClientDomain,
   markAsTestEvent = false
 ): Event['origin']['buildType'] => {
+  webex.logger.log(
```
marcin-bazyl, line 205: `getBuildType()` is called lots of times, we don't want to be logging every time it's called as this will flood the log with unnecessary log lines.
Wouldn't it be better to just move the `getBuildType()` function into a separate file like util.ts or something so than there is no problem with spying or mocking it in unit tests.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -225,7 +232,8 @@ export const getBuildType = (
  */
 export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
   const origin: Partial<Event['origin']> = {
-    buildType: exports.getBuildType(
```
rarajes2, line 228: **Question**: Why did this work? We have syntax like `module.exports` in node.js but this syntax seems bit weird and I also don't see any custom object named `exports` being used.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
mkesavan13: @marcin-bazyl - If you notice, it is already in a util file. If we move it to another util file, it might not look good on the code. That is why I chose to do it this way. 

Is it okay to have it in another utils file?

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
mkesavan13: It works alright in Modules (NPM Package) because even though there's no explicit mention of `module.exports`, the `export const` syntax means that it is added as part of the exports object.

It was failing in CDN all this while. That is what we are trying to fix

@rarajes2

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
mkesavan13: Ah. That makes sense too. I'll try to do that @marcin-bazyl

mkesavan13: Done, @marcin-bazyl

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -406,20 +412,28 @@ describe('internal-plugin-metrics', () => {
...
-      });
+      };
+
+      // just submit any event
+      prepareDiagnosticMetricItem(webex, item);
+      assert.deepEqual(item.eventPayload.origin.buildType, 'test');
 
-      assert.calledOnceWithExactly(getBuildTypeSpy, webClientDomain, markAsTestEvent);
+      delete item.eventPayload.origin.buildType;
+      item.eventPayload.event.eventData.markAsTestEvent = false;
```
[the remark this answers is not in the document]
mkesavan13: Since it is listed in `getBuildType` only, I also didn't test all the cases here. 

We're just ensuring getBuildType is called with appropriate arguments and that's why wanted to alter the argument and see if we get the other outcome.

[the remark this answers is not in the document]
marcin-bazyl: yeah, but as we are unable to really verify that `prepareDiagnosticMetricItem()` calls `getBuildType()`, then in theory we shouldn't be relying on the fact that getBuildType tests pass to imply that `prepareDiagnosticMetricItem()` also works. In theory `prepareDiagnosticMetricItem()` might be executing any other code that happens to return the right result for the 2 values of `markAsTestEvent` but not for the rest of webClientDomain or NODE_ENV values. Like I said, it's just theoretical, in practice it's very unlikely that `prepareDiagnosticMetricItem()` would do something else instead of calling `getBuildType()` and still work correctly just for `markAsTestEvent`, that's why I've approved the PR.

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -234,8 +234,14 @@ describe('internal-plugin-metrics', () => {
   });
 
   describe('getBuildType', () => {
+    let webex;
     beforeEach(() => {
       process.env.NODE_ENV = 'production';
+      webex = {
```
rarajes2, line 240: We can remove this

rarajes2 (approved):

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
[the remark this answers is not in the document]
mkesavan13: Makes sense. I hope we do not land in such a situation. 

Thanks @marcin-bazyl

packages/@webex/internal-plugin-metrics/test/unit/spec/call-diagnostic/call-diagnostic-metrics.util.ts
mkesavan13: Done, @rarajes2

PR #3691, opened 2024-07-08
fix(webex-core): fix parsing of u2c hostmap

Coread:
# COMPLETES # [SPARK-540203](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-540203)

There was a fundamental misunderstanding in the previous parsing algorithm. It assumed that the `id` of each service in the host catalog was the same as the service name as referenced by the serviceLinks. This is not the case. The result was that for certain services, no alternative hosts would be identified. The previous version also created a mock version of the primary with a higher priority. I have also removed this.

The code does the following:
* Iterates over the serviceLinks. Each key of the service links is the service name.
* Finds the corresponding URL key in the host catalog 
* Extract the ID from the first host under this entry
* Generate the first entries in the hosts for this service. homeCluster is set to true for hosts entered here, as indicated by the serviceLink pointing to this entry.
* Go over the rest of the host catalog and find other entries with the same ID and add hosts

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/lib/services/services.js
+++ b/packages/@webex/webex-core/src/lib/services/services.js
@@ -702,16 +702,20 @@ const Services = WebexPlugin.extend({
 
       const matchingCatalogEntry = serviceHostmap.hostCatalog[host];
 
-      if (!matchingCatalogEntry) {
-        return;
-      }
-
       const formattedHost = {
         name: serviceName,
         defaultUrl: serviceUrl,
         defaultHost: host,
         hosts: [],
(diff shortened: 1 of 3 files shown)


robstax (approved):
tested locally. appears to work as described.

sreenara (approved):

sreenara (approved):
Tested this. Looks good. Thanks

PR #3696, opened 2024-07-09
fix: refactor how reachability timeouts work

marcin-bazyl:
# COMPLETES #[SPARK-540198](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-540198)

Currently amplitude metrics show that there are about 3k successful meeting joins from users that failed all reachability checks. Web app wants to show a UI warning to the user when reachability fails, so we need to make sure that reachability results are reliable and when reachability says that all checks failed, it is indeed the case that all clusters are not reachable.

The suspicion is that false negative results are caused by the very short (3s) timeout used for reachability checks.

Reorganizing how reachability times out - instead of each ClusterReachability instance having its own timeout, the timeout is managed by the Reachability class. We keep existing behavior of gatherReachability() resolving at most after 3s, but we will now keep collecting the results in the background after that and eventually send metrics with the stats - this will allow us to make a decision on how long the timeouts need to be.

more details here:
https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/Change+of+reachability+timeouts

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -211,22 +263,43 @@ export class ClusterReachability {
...
       result.latencyInMilliseconds = latency;
       result.result = 'reachable';
+      if (publicIp) {
+        result.clientMediaIPs = [publicIp];
+      }
+
+      this.emit(
+        {
+          file: 'clusterReachability',
+          function: 'registerIceCandidateListener',
```
torgeadelin, line 293: i think this should be `saveResult` no?

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,196 @@ export default class Reachability {
...
+    });
+  }
+
+  /**
+   * Sends a metric with all the statistics about how long reachability took
+   *
+   * @returns {void}
+   */
+  private async sendMetric() {
+    const getStatistics = (results, protocol, isVideoMesh) => {
```
torgeadelin, line 465: nitpick: might be better to extract as separate func, that way testing is easier

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,196 @@ export default class Reachability {
...
+
+        this.resolveReachabilityPromise();
+      }
+    }, VIDEO_MESH_TIMEOUT * 1000);
+
+    this.publicCloudTimer = setTimeout(() => {
+      this.publicCloudTimer = undefined;
+
+      LoggerProxy.logger.log(
+        `Reachability:index#gatherReachability --> Reachability checks timed out (${DEFAULT_TIMEOUT}s)`
```
torgeadelin, line 532: should be `index#startTimers`

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,196 @@ export default class Reachability {
...
+   * @returns {void}
+   */
+  private startTimers() {
+    this.vmnTimer = setTimeout(() => {
+      this.vmnTimer = undefined;
+      // if we are only missing VMN results, then we don't want to wait for them any longer
+      // as they are likely to fail if users are not on corporate network
+      if (this.areAllPublicClusterResultsReady()) {
+        LoggerProxy.logger.log(
+          'Reachability:index#gatherReachability --> Reachability checks timed out (VMN timeout)'
```
torgeadelin, line 521: should be `index#startTimers`

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,196 @@ export default class Reachability {
...
+          file: 'reachability',
+          function: 'overallTimer timeout',
+        },
+        'reachability:done',
+        {}
+      );
+      this.sendMetric();
+
+      LoggerProxy.logger.log(
+        `Reachability:index#gatherReachability --> Reachability checks fully timed out (${OVERALL_TIMEOUT}s)`
```
torgeadelin, line 553: should be `index#startTimers`

torgeadelin (commented):
some small comments, but haven't had a chance to test, will continue tomorrow

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: fixed

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: Ok, I've done that, but both `getStatistics()` and `sendMetric()` are private so they shouldn't be called from the tests anyway. I've worked around this by declaring them as `protected` so in the test I created a class that extends `Reachability` class and added public methods that call `getStatistics()` and `sendMetric()` to allow us to test them. I've added more unit tests for these methods now, I agree it makes it easier to test all the edge cases, but I feel that it makes the unit tests too closely coupled to the implementation, which then makes the tests not very useful if you try to refactor something inside `Reachability` class, so it's a trade off....

torgeadelin: I missed that it was a private function. 
Yeah, I see your point and I agree, but the effort for still doing it is very appreciated!

torgeadelin (approved):
looking good, tested common join flows and all working good

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -23,10 +21,32 @@ export type ClusterReachabilityResult = {
   xtls: TransportResult;
 };
 
+// data for the Events.resultReady event
+export type ResultEventData = {
+  protocol: 'udp' | 'tcp' | 'xtls';
+  result: 'reachable' | 'unreachable' | 'untested';
```
rarajes2, line 27: Q. From where are we triggering the partial reachability event?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -211,22 +263,43 @@ export class ClusterReachability {
...
     if (result.latencyInMilliseconds === undefined) {
       LoggerProxy.logger.log(
         // @ts-ignore
-        `Reachability:index#storeLatencyResult --> Successfully reached ${this.name} over ${protocol}: ${latency}ms`
+        `Reachability:index#saveResult --> Successfully reached ${this.name} over ${protocol}: ${latency}ms`
       );
       result.latencyInMilliseconds = latency;
       result.result = 'reachable';
+      if (publicIp) {
+        result.clientMediaIPs = [publicIp];
```
rarajes2, line 287: Do we need to append the `publicIp` to `clientMediaIPs` here or is this intentional?

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -3,19 +3,26 @@
  */
 
 /* eslint-disable class-methods-use-this */
-import {mapValues} from 'lodash';
+import {isEqual, mapValues, mean} from 'lodash';
```
rarajes2, line 6: Should we import these methods individually? Not sure if tree shaking works this way

e.g.
```
import isEqual from 'lodash/isEqual';
```

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,207 @@ export default class Reachability {
...
+    this.resultsCount.public.xtls = 0;
+  }
+
   /**
    * Performs reachability checks for all clusters
    * @param {ClusterList} clusterList
-   * @returns {Promise<ReachabilityResults>} reachability check results
+   * @returns {Promise<void>} promise that's resolved as soon as the checks are started
    */
-  private async performReachabilityChecks(clusterList: ClusterList): Promise<ReachabilityResults> {
```
rarajes2, line 403: Why did we remove the return type?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: I'm not sure what are you asking about, the `ResultEventData ` is used only with "resultReady" event and that event will actually always have result set to "reachable", because it's only emitted when we get an ICE candidate. I could change the type here to just say  `result: 'reachable'.
Or if you're asking about the partial reachability event that was mentioned on the confluence page, then that its emitted from [here](https://github.com/webex/webex-js-sdk/blob/7f573abc7ef45707c945ea9d3c4c9b59fc7b7f5f/packages/%40webex/plugin-meetings/src/reachability/index.ts#L685) but I've renamed the event and called it 'reachability:firstResultAvailable' instead of "reachability:partiallyDone". I feel it's a better name, once the PR is approved I'll update the confluence.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
marcin-bazyl: it's intentional, because this is the first time we're saving the result (see the check in line 279) so there is no array to append to

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: I haven't checked if it works correctly with tree shaking but there are 108 other existing occurrences of SDK importing methods from lodash this way, so I just kept the same convention and if we think that this is not the right way, we should probably raise a jira to investigate and fix it across the whole SDK separately to this PR.

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: Previously gatherReachability() was calling performReachabilityChecks() and that was returning all the results that were then stored in localStorage by gatherReachability() in one go. Now we update results in local storage every time a new result comes in through the resultReady event, so gatherReachability() doesn't need to get the results from performReachabilityChecks() anymore and performReachabilityChecks() doesn't wait for all results to be ready before it resolves and only gatherReachability() waits for that. Also performReachabilityChecks() is a private function so we can change what it returns at any time.

packages/@webex/plugin-meetings/test/unit/spec/reachability/index.ts
```diff
@@ -1033,3 +1762,309 @@ describe('getReachabilityMetrics', () => {
...
+
+  public testSendMetric() {
+    return this.sendMetric();
+  }
+
+  public setFakeClusterReachability(fakeClusterReachability) {
+    this.clusterReachability = fakeClusterReachability;
+  }
+}
+describe('getStatistics', () => {
```
sreenara, line 1787: nitpick, please add a new line here

sreenara (approved):
Just a couple of questions and non-blocking comments. Nice work on the tests!

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -23,10 +21,32 @@ export type ClusterReachabilityResult = {
...
+  protocol: 'udp' | 'tcp' | 'xtls';
+  clientMediaIPs: string[];
+};
+
+export const Events = {
+  resultReady: 'resultReady', // emitted when a cluster is reached successfully using specific protocol
+  clientMediaIpsUpdated: 'clientMediaIpsUpdated', // emitted when more public IPs are found after resultReady was already sent for a given protocol
+} as const;
+
+export type Events = Enum<typeof Events>;
```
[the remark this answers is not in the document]
marcin-bazyl: yes, it's explained here: https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/How+to+use+enums+and+alternatives

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -60,24 +67,39 @@ export type ReachabilityResults = Record<
   }
 >;
 
+// timeouts in seconds
+const DEFAULT_TIMEOUT = 3;
+const VIDEO_MESH_TIMEOUT = 1;
+const OVERALL_TIMEOUT = 15;
```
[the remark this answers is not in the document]
marcin-bazyl: yes, I'm starting with 15s as I want to gather the metrics to see the distribution of latencies and what are the longest ones we get. For now the app won't do anything on the overall timeout anyway, only after gathering those metrics we'll decide on the actual best value for that overall timeout. I've updated the wiki to say that.

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -395,16 +409,207 @@ export default class Reachability {
...
+    }, OVERALL_TIMEOUT * 1000);
+  }
+
+  /**
+   * Stores given reachability results in local storage
+   *
+   * @param {ReachabilityResults} results
+   * @returns {Promise<void>}
+   */
+  private async storeResults(results: ReachabilityResults) {
```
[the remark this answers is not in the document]
marcin-bazyl: SDK has always been storing it in local storage, I haven't changed it. I guess the idea was to have results ready to use in case we reload the page and don't do reachability before joining a meeting, but right now reachability is always done as part of `webex.meetings.register()` call that you have to make before being able to join anything, so there is not much benefit of using local storage. We might want to re-evaluate this, but it's out of scope for this PR.

packages/@webex/plugin-meetings/test/unit/spec/reachability/index.ts
marcin-bazyl: done

PR #3704, opened 2024-07-12
feat: add share video encoder implementation to share video mute event

brycetham:
# COMPLETES [SPARK-542195](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-542195)

Being able to determine whether a hardware or a software encoder is being used when share video gets muted.

Including the share video encoder implementation in the SDK event that is emitted whenever the share video mute state changes.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -7897,6 +7897,7 @@ export default class Meeting extends StatelessWebexPlugin {
     Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.MEETING_SHARE_VIDEO_MUTE_STATE_CHANGE, {
       correlationId: this.correlationId,
       muted,
+      encoderImplementation: this.statsAnalyzer.shareVideoEncoderImplementation,
     });
   };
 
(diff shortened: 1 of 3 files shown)


edvujic (approved):

marcin-bazyl (approved):
Maybe it would have been better to have the `encoderImplementation` added as a more generic field in StatsAnalyzer that would be there for all outbound video not just specific to share, but approving anyway since only the share one is used right now

k-wasniowski (approved):

brycetham:
> Maybe it would have been better to have the `encoderImplementation` added as a more generic field in StatsAnalyzer that would be there for all outbound video not just specific to share, but approving anyway since only the share one is used right now

Yeah, I think this would be a good enhancement for the future. I wasn't sure if it was possible for the encoder implementations to be different for different simulcast layers. https://datatracker.ietf.org/doc/html/rfc8853 kinda implies that it does ("the same video source encoded with different video-encoder types"), so I think it requires a bit more thought for main video. Decided to leave it out until we have a use case for it.

Kesari3008 (approved):
Would this change not require adjustment in test files as well

brycetham:
> Would this change not require adjustment in test files as well

Ah yeah, thank. I'll add them. It's a small change so should be simple to test.

PR #3706, opened 2024-07-15
fix(samples): fixes the access token in the url issue with OAuth

rarajes2:
# COMPLETES #<[SPARK-499535](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-499535)>

When using the OAuth in the samples page, after the successful authorization the access token stays in the url which should be removed as it poses a security issue.

**Root cause**
The issue was coming because we are not able to maintain the state for the webex object when it gets initialized. After successful OAuth it gets redirected back to the page but since there is no webex object causing that event listener to get removed. After getting back to the page if we again call the OAuth it gets the token from the url and cleans the url as well.

The access token gets removed from the url on successful authorization.

**Screen recording**

https://github.com/user-attachments/assets/4566a4fe-4f76-4412-bea5-888844d0e04e

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tested with the cisco and a gmail user
- Also tested in incognito mode

### I certified that

---


sreenara (approved):
Code change looks good. However, can we add these details to the description?
1. What was the root cause and how does the fix solve the problem?
2. In the tests section, please list the manual tests done
3. Have we tried to do this login with a gmail user instead of Cisco SSO?

rarajes2:
> Approved. Just curious as to why amplify failed

Some glitch. Worked after re-run.

PR #3725, opened 2024-07-19
fix(webex-core): fix findServiceUrlFromUrl

Coread:

After the latest hostmap parsing refactor, the ordering of the services changed. This exposed a bug in findServiceUrlFromUrl, which would find the first host that matched the URL instead of checking the defaultUrl matched.

I changed the algorithm to do the following:
1. check the defaultUrl of the service, if the URL we are checking starts with this defaultUrl then the service matches
2. for each of the hosts for the service, replace the host within the defaultUrl with this host. Then check to see if the URL we are checking starts with this alternate URL

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/lib/services/service-catalog.js
+++ b/packages/@webex/webex-core/src/lib/services/service-catalog.js
@@ -246,10 +246,14 @@ const ServiceCatalog = AmpState.extend({
     ];
 
     return serviceUrls.find((serviceUrl) => {
+      // Check to see if the URL we are checking starts with the default URL
       if (url.startsWith(serviceUrl.defaultUrl)) {
         return true;
       }
 
+      // If not, we check to see if the alternate URLs match
+      // These are made by swapping the host of the default URL
+      // with that of an alternate host
       for (const host of serviceUrl.hosts) {
         const alternateUrl = new URL(serviceUrl.defaultUrl);
         alternateUrl.host = host.host;
--- a/packages/@webex/webex-core/src/lib/services/service-catalog.js
+++ b/packages/@webex/webex-core/src/lib/services/service-catalog.js
@@ -237,7 +237,6 @@ const ServiceCatalog = AmpState.extend({
    * @returns {serviceUrl} - ServiceUrl assocated with provided url
    */
   findServiceUrlFromUrl(url) {
-    const incomingUrlObj = Url.parse(url);
     const serviceUrls = [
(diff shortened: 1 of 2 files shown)


packages/@webex/webex-core/src/lib/services/service-catalog.js
```diff
@@ -247,12 +246,21 @@ const ServiceCatalog = AmpState.extend({
     ];
 
     return serviceUrls.find((serviceUrl) => {
-      if (incomingUrlObj.hostname === Url.parse(serviceUrl.defaultUrl).hostname) {
+      // Check to see if the URL we are checking starts with the default URL
+      if (url.startsWith(serviceUrl.defaultUrl)) {
```
marcin-bazyl, line 250: question: is it ever at all possible for defaultUrl to be empty? startsWith() would always return true for it

packages/@webex/webex-core/test/unit/spec/services/service-catalog.js
```diff
@@ -201,6 +201,56 @@ describe('webex-core', () => {
...
+        hosts: [{host: 'example1.com'}, {host: 'example2.com'}],
+      };
+
+      it.each([
+        'discovery',
+        'preauth',
+        'signin',
+        'postauth',
+        'override'
+      ])('matches a default url correctly', (serviceGroup) => {
```
marcin-bazyl, line 217: `serviceGroup`  doesn't seem to be used at all, so what's the point of these test cases? same for the next test

Coread: got spot. The intention was to check that it would use services from each service group, not just preauth. I've fixed this now.

packages/@webex/webex-core/src/lib/services/service-catalog.js
Coread: while in theory it might be possible for the catalog to have a blank serviceLink (this would be an error in the catalog), the parsing in _formatReceivedHostmap in services.js would ignore that entry as we do the following during parsing:
```js
      try {
        host = new URL(serviceUrl).host;
      } catch (e) {
        return;
      }
```
the defaultUrl is the serviceUrl here so we exclude invalid serviceUrls early on.

marcin-bazyl (approved):

PR #3726, opened 2024-07-22
fix(plugin-meetings): prevent fetchMeetingInfo service call for one2one meeting. 

rsarika:
1-1 calls cause fetchmeeting request to be sent to wbxappapi that always fails

# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-274559

create meeting API

Added new type to create meeting api to state its a one2one meeting so by using that type to skip fetchmeeting api

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

MANUAL Testing done by passing ONE_ON_ONE_CALL in WWC and added below network logs
[withFix.txt](https://github.com/user-attachments/files/16487469/withFix.txt)
[withoutFix.txt](https://github.com/user-attachments/files/16487470/withoutFix.txt)

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -14,7 +14,7 @@ export const ANSWER = 'answer';
 export const CALL = 'call';
 export const CORRELATION_ID = 'correlationId';
 export const MEETINGNUMBER = 'meetingNumber';
-export const CONVERSATION_URL = 'conversationUrl';
+export const CONVERSATION_URL_KEY = 'conversationUrl';
 export const CALENDAR = 'calendar';
 export const CALENDAR_EVENTS_API = 'calendarEvents';
 export const CMR_MEETINGS = 'cmrmeetings';
@@ -64,7 +64,7 @@ export const READY = 'ready';
 
 export const SEND_DTMF_ENDPOINT = 'sendDtmf';
 export const SENDRECV = 'sendrecv';
-export const SIP_URI = 'sipUri';
+export const SIP_URI_KEY = 'sipUri';
 export const SHARE = 'share';
 
 export const TYPE = 'type';
@@ -81,7 +81,6 @@ export const _ACTIVE_ = 'ACTIVE';
 export const _CALL_ = 'CALL';
 export const _CREATED_ = 'CREATED';
 export const _CONFLICT_ = 'CONFLICT';
-export const _CONVERSATION_URL_ = 'CONVERSATION_URL';
 
 export const _ERROR_ = 'ERROR';
 
@@ -97,22 +96,17 @@ export const _ID_ = 'id';
 export const _JOIN_ = 'JOIN';
 export const _JOINED_ = 'JOINED';
 
-export const _LOCUS_ID_ = 'LOCUS_ID';
 export const _LEFT_ = 'LEFT';
 export const _MOVED_ = 'MOVED';
 export const _ON_HOLD_LOBBY_ = 'ON_HOLD_LOBBY';
-export const _MEETING_LINK_ = 'MEETING_LINK';
-export const _MEETING_UUID_ = 'MEETING_UUID';
 export const _MEETING_ = 'MEETING';
 export const _MEETING_CENTER_ = 'MEETING_CENTER';
-export const _MEETING_ID_ = 'MEETING_ID';
 
 export const _NOT_IN_MEETING_ = 'NOT_IN_MEETING';
 export const _NONE_ = 'NONE';
 
 export const _OBSERVE_ = 'OBSERVE';
 
-export const _PERSONAL_ROOM_ = 'PERSONAL_ROOM';
 export const _PEOPLE_ = hydraTypes.PEOPLE;
 
 export const _REQUESTED_ = 'REQUESTED';
@@ -122,7 +116,6 @@ export const _REMOVE_ = 'REMOVE';
 export const _ROOM_ = hydraTypes.ROOM;
 
 export const _SIP_BRIDGE_ = 'SIP_BRIDGE';
-export const _SIP_URI_ = 'SIP_URI';
 export const _SEND_RECEIVE_ = 'SENDRECV';
 export const _SEND_ONLY_ = 'SENDONLY';
 export const _INACTIVE_ = 'INACTIVE';
@@ -135,8 +128,6 @@ export const _UNKNOWN_ = 'UNKNOWN';
 export const _WEBEX_MEETING_ = 'WEBEX_MEETING';
 export const _WAIT_ = 'WAIT';
 export const _MOVE_MEDIA_ = 'MOVE_MEDIA';
-
-export const _ONE_2_ONE_MEEETING_ = 'ONE_2_ONE_MEETING';
 
 // *********** PARTICIPANT DELTAS ***********
 export const PARTICIPANT_DELTAS = {
@@ -1333,3 +1324,24 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
+
+export enum MeetingType {
+  CONVERSATION_URL = 'CONVERSATION_URL',
+  MEETING_LINK = 'MEETING_LINK',
+  SIP_URI = 'SIP_URI',
+  PERSONAL_ROOM = 'PERSONAL_ROOM',
+  ONE_ON_ONE_CALL = 'ONE_ON_ONE_CALL',
+  LOCUS_ID = 'LOCUS_ID',
+  MEETING_ID = 'MEETING_ID',
+  MEETING_UUID = 'MEETING_UUID',
+}
+
+// TODO: remove this after the meeting type is fully migrated to MeetingType
+export const _CONVERSATION_URL_ = 'CONVERSATION_URL';
+export const _MEETING_LINK_ = 'MEETING_LINK';
+export const _SIP_URI_ = 'SIP_URI';
+export const _PERSONAL_ROOM_ = 'PERSONAL_ROOM';
+export const _ONE_ON_ONE_CALL_ = 'ONE_ON_ONE_CALL';
+export const _LOCUS_ID_ = 'LOCUS_ID';
+export const _MEETING_ID_ = 'MEETING_ID';
+export const _MEETING_UUID_ = 'MEETING_UUID';
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -12,9 +12,6 @@ export const ALIAS = 'alias';
 export const ANSWER = 'answer';
 
 export const CALL = 'call';
-export const CORRELATION_ID = 'correlationId';
-export const MEETINGNUMBER = 'meetingNumber';
-export const CONVERSATION_URL_KEY = 'conversationUrl';
 export const CALENDAR = 'calendar';
 export const CALENDAR_EVENTS_API = 'calendarEvents';
 export const CMR_MEETINGS = 'cmrmeetings';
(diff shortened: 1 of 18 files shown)


packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -1358,6 +1361,28 @@ describe('plugin-meetings', () => {
             );
           });
 
+          it('creates the meeting avoiding meeting info fetch by passing one2OneMeeting true', async () => {
+
+            const meeting = await webex.meetings.createMeeting('test destination', 'test type', false, {}, undefined, false, undefined, undefined, true);
+
+            const expectedMeetingData = {
+              permissionToken:
+                'eyJhbGciOiJIUzI1NiJ9.eyJleHAiOiIxMjM0NTYiLCJwZXJtaXNzaW9uIjp7InVzZXJQb2xpY2llcyI6eyJhIjp0cnVlfX19.wkTk0Hp8sUlq2wi2nP4-Ym4Xb7aEUHzyXA1kzk6f0V0',
```
mkesavan13, line 1370: Let us please refrain from adding any kind of tokens to UTs as it is a publicly maintained code. Please change this to `FAKE_TOKEN` and then update tests accordingly

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1290,19 +1295,17 @@ export default class Meetings extends WebexPlugin {
...
-        waitingTime > 0
-      ) {
-        meeting.fetchMeetingInfoTimeoutId = setTimeout(
-          () => meeting.fetchMeetingInfo(meetingInfoOptions),
-          waitingTime
-        );
-        meeting.parseMeetingInfo(undefined, destination);
-      } else {
-        await meeting.fetchMeetingInfo(meetingInfoOptions);
+      } else if (!isOne2OneMeeting) {
```
mkesavan13, line 1298: Just thinking about it, would it make more sense to use the existing `type` parameter in `meetings.create` and then not do anything related to `meetingInfo` if the type is the newly defined one?

Also, `isOne2OneMeeting` doesn't feel like a standard one.

mkesavan13 (changes requested):
If we have tested this with WWC, please mention the manual tests done and also add a screen recording of changes reflecting in WWC.

packages/@webex/plugin-meetings/src/meetings/index.ts
Kesari3008, line 1298: Do we not have some parameter within the meeting object that could be used to identify if it is 1:1 meeting instead of introducing new parameter?

Kesari3008 (commented):
Please capture Network logs with and without change and upload it to the PR. Also update Test Performed section in the description with the manual testing done

packages/@webex/plugin-meetings/src/meetings/index.ts
rsarika: yes. removed isOne2OneMeeting and utilised type

rsarika: Currently no. It will be passed by web client based on the video call button clcik

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -136,6 +136,8 @@ export const _WEBEX_MEETING_ = 'WEBEX_MEETING';
 export const _WAIT_ = 'WAIT';
 export const _MOVE_MEDIA_ = 'MOVE_MEDIA';
 
+export const _ONE_2_ONE_MEEETING_ = 'ONE_2_ONE_MEETING';
```
mkesavan13, line 139: I am not sure what the name should be but this still doesn't feel like something standard. We need a better name that can be documented.

Maybe, "WEBEX_DIRECT_MEETING" or so.

CC - @marcin-bazyl - Your expertise on naming is required here 😁

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -1782,6 +1783,30 @@ describe('plugin-meetings', () => {
...
+                meeting,
+                Meeting,
+                'createMeeting should eventually resolve to a Meeting Object'
+              );
+              assert.notCalled(webex.meetings.meetingInfo.fetchMeetingInfo);
+              assert.calledThrice(TriggerProxy.trigger);
+            } catch (err) {
+              assert.instanceOf(err, NoMeetingInfoError);
+            }
+          };
```
mkesavan13, line 1804: I don't think we need a separate function block as we would use this method in only one test case. Let's add the logic directly to the callback of the test case in line 1807

mkesavan13 (commented):
Minor comments. Rest looks good to me

Kesari3008:
> 1. Please capture Network logs with and without change and upload them to the PR.
> 2. Update the Test Performed section in the description with the manual testing done.
> 3. Add the validated label for checks to run

@rsarika Please address these comments

mkesavan13 (approved):
Looks good to me. However, please ensure to have @Kesari3008's review comments addressed and also to have an approval from WWC

packages/@webex/plugin-meetings/src/constants.ts
marcin-bazyl: Maybe the name should be "ONE_2_ONE_CALL" rather than "MEETING", because I think that's how the server teams call these things - they call them "1-1 Locus calls", they are not meetings.
What I really don't like about the already existing code is that the `type` is declared as a `string` and there is no documentation or anything to point the user of the API what values are allowed/supported for `type`. The constants currently supported are for example `_CONVERSATION_URL_` or `_LOCUS_ID_`, but these are just scattered around the codebase not even grouped together in the constants file, How is anyone supposed to know what values to pass as `type`? As you're touching this area, can we improve this and define a proper Enum for the possible values for `type` supported by the `create()` API?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1090,7 +1092,7 @@ export default class Meetings extends WebexPlugin {
    */
   public create(
     destination: string,
-    type: string = null,
+    type: string | MeetingType = null,
```
marcin-bazyl, line 1095: why is the type still allowed to be string too? and why do we convert it to string later on and still use old set of consts? Changing everywhere to use the new type enum shouldn't be that difficult, it's only mainly used in the fetching meeting info code (as destinationType) and only in a couple of other place that are easy to fix (like MeetingsUtil.getMeetingAddedType). It feels like maybe we should even rename the new enum from `MeetingType` to `DestinationType`

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1331,3 +1324,24 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
 
 // constant for named media group type
 export const NAMED_MEDIA_GROUP_TYPE_AUDIO = 1;
+
+export enum MeetingType {
```
marcin-bazyl, line 1328: see this page about enums:
https://confluence-eng-gpk2.cisco.com/conf/x/uluyEw
and consider following the proposed approach there, you can see examples doing that in SDK: NETWORK_STATUS or IP_VERSION that follow that convention

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1099,12 +1098,15 @@ export default class Meetings extends WebexPlugin {
     meetingInfo = undefined,
     meetingLookupUrl = undefined
   ) {
+    LoggerProxy.logger.warn('ravi chandra sekhar');
```
marcin-bazyl, line 1101: please remove the temporary logs

packages/@webex/plugin-meetings/src/constants.ts
rsarika: changed the enum implementation and renamed it to DESTINATION_TYPE

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -64,7 +64,7 @@ export const READY = 'ready';
 
 export const SEND_DTMF_ENDPOINT = 'sendDtmf';
 export const SENDRECV = 'sendrecv';
-export const SIP_URI = 'sipUri';
+export const SIP_URI_KEY = 'sipUri';
```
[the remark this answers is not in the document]
rsarika: moved all of the keys to MEETING_KEY Enum

mkesavan13 (changes requested):
Just a few comments and a change request. Rest looks good to me

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1331,3 +1319,16 @@ export const MEETING_PERMISSION_TOKEN_REFRESH_REASON = 'ttl-join';
...
+  MEETING_LINK: 'MEETING_LINK',
+  SIP_URI: 'SIP_URI',
+  PERSONAL_ROOM: 'PERSONAL_ROOM',
+  ONE_ON_ONE_CALL: 'ONE_ON_ONE_CALL',
+  LOCUS_ID: 'LOCUS_ID',
+  MEETING_ID: 'MEETING_ID',
+  MEETING_UUID: 'MEETING_UUID',
+} as const;
+
+export type DESTINATION_TYPE = Enum<typeof DESTINATION_TYPE>;
```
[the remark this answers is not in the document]
rsarika: i changed this according Marcin's review

[marcin-bazyl](https://github.com/marcin-bazyl) [5 days ago](https://github.com/webex/webex-js-sdk/pull/3726#discussion_r1698331048)
see this page about enums:
https://confluence-eng-gpk2.cisco.com/conf/x/uluyEw
and consider following the proposed approach there, you can see examples doing that in SDK: NETWORK_STATUS or IP_VERSION that follow that convention

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -12,9 +12,6 @@ export const ALIAS = 'alias';
 export const ANSWER = 'answer';
 
 export const CALL = 'call';
-export const CORRELATION_ID = 'correlationId';
-export const MEETINGNUMBER = 'meetingNumber';
-export const CONVERSATION_URL = 'conversationUrl';
```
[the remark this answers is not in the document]
rsarika: It still works if they use 'correlationId' but for CORRELATION_ID const i remove as per below Marcin's comments

Marcin Wojtczak: yeah it will be a breaking change but it will be easy for any client to update and use the correct value from the new enum and if they just use hardcoded strings and JS (not TS) then it will still work fine if we keep the same values in the enum as the old consts. Also I think majority of other SDK clients are still not on the 3.x version right? and 3.x introduces lots of breaking API changes so now is a good time to do any other API breaking changes (it was even better time when we were still on beta branch, but now is still better than later)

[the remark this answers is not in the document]
mkesavan13: @rsarika - 

> then it will still work fine if we keep the same values in the enum as the old consts

This is what I am concerned about. In your change, the enum doesn't keep the old values. The old values go away.

packages/@webex/plugin-meetings/src/constants.ts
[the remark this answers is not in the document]
mkesavan13: Alright. Just was curious to know.

packages/@webex/plugin-meetings/src/constants.ts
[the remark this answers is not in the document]
mkesavan13: My bad. I think I was confused between the two constants `MEETING_KEY` and `DESTINATION_TYPE`

You were right to remove the old constants. We'll discuss this in person on what confused me. Sorry about this @rsarika

mkesavan13 (approved):
Let's remove all the unused constants.

I do not think there's any breaking change. It was just grouping together. Sorry for the confusions. Approving

marcin-bazyl (approved):

PR #3732, opened 2024-07-23
fix(CallingClient): add c line at session level for answer sdp

mkesavan13:
# COMPLETES [SPARK-522437](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-522437)

Currently, the SDP from the chrome extension has SDP connection (c) information at the media level. Some SBCs do not like it. 

Copying the c-line from first m-line and adding it at the session-level as indicated in the Jira

#### Network logs

[📎 new_sdp.zip](https://github.com/user-attachments/files/16356520/new_sdp.zip)

This zip contains har files for both outgoing call and incoming call where we can see the media requests having modified SDPs

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tested the SDP from an outgoing call
- Put the call on-hold and then resumed and checked both the cases for SDP updates
- Checked for an incoming call to have the right SDP sent through for media requests. Looks good

### I certified that

---


packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2373,6 +2373,24 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     });
   }
 
+  /* istanbul ignore next */
+  /**
+   * Copy SDP's c-line to session level from media level.
+   * SPARK-522437
+   */
+  private addSessionConnection(sdp: string): string {
```
Kesari3008, line 2381: Can we name it as updateSDP ?

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2416,10 +2434,13 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
             break;
           }
 
-          case RoapScenario.ANSWER:
+          case RoapScenario.ANSWER: {
```
Kesari3008, line 2437: Update the corresponding test file to cover unit test for this change

Kesari3008 (changes requested):
Please upload the logs highlighting the SDP change.

packages/calling/src/CallingClient/calling/call.ts
Kesari3008, line 2437: What about the ROAP offer and other cases here? We wouldn't require updating SDP in those cases

mkesavan13: We had a discussion about this in the Scrum meeting this morning and Sreekanth ruled out that it is a string manipulation and we need not write UTs for this.

I saw for the `RoapScenario.OFFER` case also. The test file doesn't check for the manipulation

mkesavan13: I didn't think so. We would be sending out SDPs only on answers right? 

Do we send out SDP for any offers?

packages/calling/src/CallingClient/calling/call.ts
mkesavan13: `updateSDP` - looks very generic. That is why I named it so.

Would something like `addSessionConnectionToSDP` be better?

packages/calling/src/CallingClient/calling/call.ts
sreenara: Yes @mkesavan13 we send SDP for offer, offer request and answer

sreenara: @mkesavan13 I was under the impression that the manipulation of the SDP was happening within an existing function which is already being tested. Since the changeset is actually a new function, let's write a UT for this.

sreenara (commented):
@mkesavan13 replied to a couple of comments

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2416,10 +2434,13 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+            const sdp = this.addSessionConnection(event.roapMessage.sdp);
```
marcin-bazyl, line 2438: I don't think calling SDK should be doing any SDP manipulation. The idea was that all media related things, including SDP handling is managed by internal-mediac-core, so I think these changes should be done in there under a config option that calling would enable.
Also, internal-media-core is already using ts-sdp library for proper parsing of the SDP so it's easier to do any SDP manipulation.

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -712,6 +712,33 @@ describe('Call Tests', () => {
       {file: 'call', method: 'updateMedia'}
     );
   });
+
+  describe('#addSessionConnection', () => {
+    let call;
+
+    beforeEach(async () => {
```
sreenara, line 719: any reason why this is async?

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -712,6 +712,33 @@ describe('Call Tests', () => {
...
+      call = callManager.createCall(dest, CallDirection.INBOUND, deviceId, mockLineId);
+    });
+
+    it('should add a session connection', () => {});
```
sreenara, line 723: this is an empty test?

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2416,10 +2439,11 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
             break;
           }
 
-          case RoapScenario.ANSWER:
+          case RoapScenario.ANSWER: {
```
sreenara, line 2442: These 2 changes in lines 2442 and 2446 are not required right? None of the other cases have this?

sreenara (changes requested):
Have we also tested an incoming call to ensure the ANSWER is correctly populated with the information?

packages/calling/src/CallingClient/calling/call.test.ts
mkesavan13: It is the Github Copilot that must have auto-filled it. Sorry, I overlooked it. Removed it now

packages/calling/src/CallingClient/calling/call.test.ts
mkesavan13: Sorry. I started it there but somehow missed to remove it. Removed it now.

mkesavan13:
> Have we also tested an incoming call to ensure the ANSWER is correctly populated with the information?

Yes. I have updated the description accordingly now

sreenara (approved):

Kesari3008 (approved):

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: This code will be moved anyway so not a blocker

PR #3736, opened 2024-07-23
fix: fail sooner if ice connection goes to failed state

marcin-bazyl:
# COMPLETES #[SPARK-545103](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-545103)

When trying to connect media, we have quite long timeouts and also we do some retries, so it can take very long for the user to finally see the message about connection failure (even more than a minute).

When browser's ice connection state goes to "failed" state, it cannot recover, so there is no point in waiting anymore, we can fail immediately.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/media/MediaConnectionAwaiter.ts
```diff
@@ -83,6 +92,17 @@ export default class MediaConnectionAwaiter {
       `Media:MediaConnectionAwaiter#connectionStateChange --> connection state: ${this.webrtcMediaConnection.getConnectionState()}`
     );
 
+    if (this.isFailed()) {
```
chrisadubois, line 95: perhaps we should add some call analyzer metrics on this.

chrisadubois (approved):

packages/@webex/plugin-meetings/src/media/MediaConnectionAwaiter.ts
marcin-bazyl: we already send CA event "client.ice.end" with the appropriate error when we fail to establish the connection - see Meeting.waitForMediaConnectionConnected() - we will do it no matter if we get "failed" state from the browser or just "disconnected" or "connecting" and we eventually time out. I don't think sending any extra info to CA about the fact that we got "failed" from browser vs us just timing out waiting for "connected" is necessary, because it doesn't tell us any more information, it's just an implementation detail of the browser, for example Firefox very quickly  goes to "failed", while chrome usually stays in "disconnected" for a lot longer and we time out before it gets to "failed"

k-wasniowski (approved):

PR #3746, opened 2024-07-25
fix(metrics): fixed-calling_sdk_version_metrics

adhmenon:
# COMPLETES #< [SPARK-466095](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-466095) >

Currently, in calling SDK, any metric had calling sdk version as "beta" string. This has been changed and now includes the proper versioning, which we get from the package.json. Since we are getting this from webpack, it's not possible to test this in the sample app. However the code has unit tests.

< DESCRIBE YOUR CHANGES >

We now get from the package.json in the webpack. This is then stored in the `process.env` variable and accessed in the metrics class.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/calling/src/Metrics/index.ts
```diff
@@ -60,7 +60,7 @@ class MetricManager implements IMetricManager {
           fields: {
             device_url: this.deviceInfo?.device?.clientDeviceUri,
             mobius_url: this.deviceInfo?.device?.uri,
-            calling_sdk_version: VERSION,
+            calling_sdk_version: process.env.CALLING_SDK_VERSION || VERSION,
```
adhmenon, line 63: Added a fallback here as well, so that even if version fails, we can have some value.

webpack.config.js
```diff
@@ -170,6 +171,7 @@ module.exports = (env = {NODE_ENV: process.env.NODE_ENV || 'production'}) => ({
               WHISTLER_API_SERVICE_URL: JSON.stringify(
                 'https://whistler-prod.allnint.ciscospark.com/api/v1'
               ),
+              CALLING_SDK_VERSION: JSON.stringify(callingPackage.version),
```
adhmenon, line 174: Added this in webpack, as I felt it that we can access it through the `process.env` variable across the codebase and hence easier to access the appropriate version number.
Open to any suggestions though.

packages/calling/src/Metrics/index.test.ts
```diff
@@ -25,6 +26,10 @@ describe('CALLING: Metric tests', () => {
 
   beforeEach(() => {
     mockSubmitClientMetric.mockClear();
+    process.env = {
+      ...originalEnv,
+      CALLING_SDK_VERSION: MOCK_VERSION_NUMBER,
```
Kesari3008, line 31: Do we have a testcase where constant VERSION gets assigned instead of the version picked from process.env?

Kesari3008 (changes requested):
Please upload network logs or screenshot capturing the payload for the metric request where we can see the calling sdk version populated.
In my testing, I still see the version as "beta"

adhmenon:
> Please upload network logs or screenshot capturing the payload for the metric request where we can see the calling sdk version populated. And also provide steps or details of testing in the tests performed section of the description. In my testing, I still see the version as "beta" Please add the validated label to PR. Required for all SDK PRs

Yes, in the sample app it will still be 'beta'. I don't think npm publish gets run there.
I will see how I can try and get some output reflecting the changes. It is a little tricky... but it should be possible

packages/calling/src/Metrics/index.test.ts
```diff
@@ -534,4 +540,37 @@ describe('CALLING: Metric tests', () => {
       );
     });
   });
+
+  describe('Calling_Sdk_Version fallback test', () => {
```
adhmenon, line 544: Added test for fallback. Currently just testing with one particular metric.

packages/calling/src/Metrics/index.ts
mkesavan13: cc - @sreenara

webpack.config.js
mkesavan13: Just curious, why do we need to `JSON.stringify` a string?

mkesavan13 (approved):
I have one comment on this review that needs addressing. Also, please address @Kesari3008's comments. Rest looks good to me.

Great PR 🎉

webpack.config.js
adhmenon: Thanks for the catch Kesava. Yes, I had added it under the assumption that we _might_ not be getting a string. There was also some older logic which I had used before and I forgot to remove this stringify. You are correct, it's not required. Will update the PR with the same changes and the VERSIOn string changes.

packages/calling/src/Metrics/index.test.ts
adhmenon: Added the test case for this.

packages/calling/src/Metrics/index.test.ts
Kesari3008: Yes, just one is fine. Make the env object return null or empty object, that's when actual fallback will happen

Kesari3008 (commented):
Few UTs seem to be failing because of beta and unknown difference in the VERSION contant, please check and fix it

Kesari3008 (approved):

PR #3750, opened 2024-07-31
fix: don't retry if we're missing h264 codec

marcin-bazyl:
# COMPLETES #[SPARK-547931](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-547931)

When we're missing H264 codec, we will fail to add media, but joinWithMedia() will still do a 2nd attempt - this is unnecessary. Also, for any other case where we fail to create the offer, there is no point retrying.

Don't do retries when we fail to create the SDP offer (that's what happens when the codec is missing).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual test with webapp

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4647,7 +4647,16 @@ export default class Meeting extends StatelessWebexPlugin {
         }
       );
 
-      if (!isRetry) {
+      // if this was the first attempt, let's do a retry
+      let shouldRetry = !isRetry;
+
+      if (CallDiagnosticUtils.isSdpOfferCreationError(error)) {
+        // errors related to offer creation (for example missing H264 codec) will happen again no matter how many times we try,
+        // so there is no point doing a retry
+        shouldRetry = false;
+      }
+
+      if (shouldRetry) {
         LoggerProxy.logger.warn('Meeting:index#joinWithMedia --> retrying call to joinWithMedia');
         this.joinWithMediaRetryInfo.isRetry = true;
         this.joinWithMediaRetryInfo.prevJoinResponse = joinResponse;
--- a/packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
+++ b/packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
@@ -902,6 +902,38 @@ describe('plugin-meetings', () => {
             }
           );
         });
+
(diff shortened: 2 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4647,7 +4647,16 @@ export default class Meeting extends StatelessWebexPlugin {
         }
       );
 
-      if (!isRetry) {
+      // if this was the first attempt, let's do a retry
+      let shouldRetry = !isRetry;
+
+      if (CallDiagnosticUtils.isSdpOfferCreationError(error)) {
```
mkesavan13, line 4653: Question: Would the SDP Offer Creation fail only if the codec is missing? Should we have this check specific to the Codec error? or is it that if the SDP creation failed once for any reason, it will always fail?

mkesavan13 (commented):
Also, how did we manually test this with Web App? I mean, in which scenario would the Codec not be present?

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: @kesava codec could not be present if firefox does not install the h264 plugin extension

chrisadubois (approved):

marcin-bazyl:
> Also, how did we manually test this with Web App? I mean, in which scenario would the Codec not be present?

Easiest is to test it by running Chromium instead of Chrome. Chromium by default doesn't have h264 codec. Also a fresh install of Firefox doesn't have it.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: @mkesavan13 in theory there could be other reasons why we fail to create the SDP offer, but in practise the missing h264 is the only one we see happening in prod. Anyway, whatever the reason, if it does fail, then on retry it will fail the same way, because we don't change anything that could affect the SDP offer creation when we do a retry, so there is no point retrying.

mkesavan13 (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Thanks @marcin-bazyl & @chrisadubois

PR #3767, opened 2024-08-07
fix: joinWithMedia() retries fail to send SDP offer

marcin-bazyl:
# COMPLETES #[SPARK-549702](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-549702)

We see from the metrics that a vast majority of automatic retries done by Meeting.joinWithMedia() fail, because we fail to send the SDP offer. It's because we send it with seq=0 and server expects seq=2.

Changed the logic so that we do 2 main things:
1. call Locus /media API with empty mediaId - this triggers Locus to create a new confluence in the backend so then the seq counter starts from zero again
2. we do a new TURN discovery on retry

implementation details:
- when doing a retry we need a LocusMediaRequest instance, but it was being deleted in closePeerConnections() called from cleanUpOnAddMediaFailure(), so I've moved the creation of LocusMediaRequest instance to Meeting.join() so it's created once and not on every addMedia() call and moved the deletion from closePeerConnections() to MeetingUtil.cleanUp(), so it's deleted only when we are dropped or leave the meeting
- added LocusMediaRequest.isConfluenceCreated() so that we know when we need to set isReconnecting flag when calling doTurnDiscovery() - that sets empty mediaId in the /media Locus API call and triggers backend to create a new confluence, which is exactly what we want when doing a retry
- added forceTurnDiscovery argument to Meeting.addMedia() - it's needed, otherwise TurnDiscovery class skips doing TURN discovery on retry triggered from joinWithMedia() if UDP reachability was successful. I was thinking about removing that logic of skipping TURN discovery completely, because we don't use it in the web app, but it will impact other clients of SDK, so I've left it for now.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

tested manually with the web app by blocking/unblocking connections when retry was being done

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4577,21 +4578,34 @@ export default class Meeting extends StatelessWebexPlugin {
...
         );
 
         joinResponse = await this.join(joinOptions);
+      } else {
+        // This is a retry, when join succeeded but addMedia failed, so we'll just call addMedia() again,
+        // but we need to ensure that it also does a new TURN discovery
+        forceTurnDiscovery = true;
+
+        // joinOptions.roapMessage is from the previous join attempt, we need to reset it
+        joinOptions.roapMessage = undefined;
```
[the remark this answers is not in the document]
marcin-bazyl: yes, that's correct, I can maybe expand that comment to explain it better or actually I think I could just move that block (lines 4613-4623) inside the previous if (so insert it just after calling this.join() in line 4601). I will also have to have the call to set `joined = true` copied there too.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6726,24 +6742,57 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    forceTurnDiscovery = false
   ): Promise<void> {
     this.retriedWithTurnServer = false;
     this.hasMediaConnectionConnectedAtLeastOnce = false;
     const LOG_HEADER = 'Meeting:index#addMedia -->';
     LoggerProxy.logger.info(
-      `${LOG_HEADER} called with: ${JSON.stringify(options)}, ${JSON.stringify(turnServerInfo)}`
+      `${LOG_HEADER} called with: ${JSON.stringify(options)}, ${JSON.stringify(
+        turnServerInfo
+      )}, ${forceTurnDiscovery}`
```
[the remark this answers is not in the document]

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
marcin-bazyl: I've moved the code

jor-row (approved):
LGTM

chrisadubois (approved):

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -865,8 +866,7 @@ describe('plugin-meetings', () => {
 
         it('should not call leave() if addMedia fails the first time and succeeds the second time and should only call join() once', async () => {
           const addMediaError = new Error('fake addMedia error');
-          const leaveError = new Error('leave error');
-          const leaveStub = sinon.stub(meeting, 'leave').rejects(leaveError);
+          const leaveStub = sinon.stub(meeting, 'leave');
```
mkesavan13, line 868: Why have we changed this?

packages/@webex/plugin-meetings/test/unit/spec/meeting/locusMediaRequest.ts
```diff
@@ -414,6 +414,8 @@ describe('LocusMediaRequest.send()', () => {
 
     describe('confluence creation', () => {
       it('resolves without sending the request if LocalMute is requested before Roap Offer is sent (confluence state is "not created")', async () => {
+        assert.equal(locusMediaRequest.isConfluenceCreated(), false);
+
         const result = await sendLocalMute({audioMuted: false, videoMuted: true});
```
mkesavan13, line 419: In the `locusMediaRequest.send` method, can't we start using the new API instead of a string check? That way we keep all the checks for isConfluenceCreated pointed to one API. Avoids confusions.

mkesavan13 (approved):
Few minor comments. Looks good otherwise

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: I just noticed that it was not needed. Looks like when someone created this test they copied from another one where Meeting.leave() needed to reject, but in this test the SDK is not expected to call Meeting.leave() at all - see the check in line 886, so making Meeting.leave() fail is unnecessary logic that could confuse someone reading this test.

packages/@webex/plugin-meetings/test/unit/spec/meeting/locusMediaRequest.ts
marcin-bazyl: the check in `LocusMediaRequest.send()` is checking `this.confluenceState === 'not created'`. The `confluenceState` can have 3 different values `'not created' | 'creation in progress' | 'created'` so we can't use `!this.isConfluenceCreated()` there as this would be true for `'creation in progress'` value and in that case we want `LocusMediaRequest.send()` to queue the request and not resolve immediately.

PR #3781, opened 2024-08-16
fix(plugin-meetings): Enable join-button and join-with-media button by adding title to the Space Instant API request body.

RAM1232:
# COMPLETES #[SPARK-408977](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-408977)

The issue where the join button was disabled in 1:1 Space meeting because the title of the request body of Space Instant API request were not paticular for the 1:1 Space meeting.This issue was reproducible while using person's Conversation URL in 1:1 Space meeting.This pull request has resolved this issue by adding the title to the Space Instant API request body and able to disable the join button.

By making changes in packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts the issue of join-button disabled was resolved.

https://github.com/user-attachments/assets/d9acee39-1a2c-4455-aee6-de18f5fdfd1b

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

I had tested the PR request in the Kitchen Sink App(Meeting) for the below process
1.Webex Initialization & Registration (with JWT Token)
2.Manage Meeting Session(Using person's Conversational URL)
3.Join without media and with media.
4.Media Controls(BNR ,VBG)
5.Streams (local/remote)
6.Screen share
7.Participants section
8.Advance Meetings contr

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
+++ b/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
@@ -208,7 +208,7 @@ export default class MeetingInfoV2 {
       .request({uri: conversationUrl, qs: {includeParticipants: true}, disableTransform: true})
       .then(({body: conversation}) => {
         const body = {
-          title: conversation.displayName ? conversation.displayName : conversation.computedTitle,
+          title: conversation.displayName ? conversation.displayName : 'AdHoc Meeting',
           spaceUrl: conversation.url,
           keyUrl: conversation.encryptionKeyUrl,
           kroUrl: conversation.kmsResourceObjectUrl,
--- a/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
+++ b/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
@@ -208,7 +208,7 @@ export default class MeetingInfoV2 {
       .request({uri: conversationUrl, qs: {includeParticipants: true}, disableTransform: true})
       .then(({body: conversation}) => {
         const body = {
-          title: conversation.displayName ? conversation.displayName : 'AdHoc Meeting',
+          title: conversation.displayName ? conversation.displayName : '',
           spaceUrl: conversation.url,
           keyUrl: conversation.encryptionKeyUrl,
           kroUrl: conversation.kmsResourceObjectUrl,
--- a/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
+++ b/packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
@@ -208,7 +208,7 @@ export default class MeetingInfoV2 {
       .request({uri: conversationUrl, qs: {includeParticipants: true}, disableTransform: true})
       .then(({body: conversation}) => {
         const body = {
-          title: conversation.displayName,
+          title: conversation.displayName ? conversation.displayName : conversation.computedTitle,
           spaceUrl: conversation.url,
           keyUrl: conversation.encryptionKeyUrl,
           kroUrl: conversation.kmsResourceObjectUrl,


packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
```diff
@@ -208,7 +208,7 @@ export default class MeetingInfoV2 {
       .request({uri: conversationUrl, qs: {includeParticipants: true}, disableTransform: true})
       .then(({body: conversation}) => {
         const body = {
-          title: conversation.displayName,
+          title: conversation.displayName ? conversation.displayName : 'AdHoc Meeting',
```
rsarika, line 211: Add/modify a test case to test this

rsarika: Can we pass the name from first participant from participant list(as per WWC) if  conversation.displayName is not available?

mkesavan13: The title should anyway come from **App.js**, right?

How does WWC do it today?

mkesavan13 (changes requested):

mkesavan13:
Also, please correct the PR title. Type "fix" already means it is a bug fix. The Component should not be "bug"

It should be "plugin-meetings"

Next, the commit message should reflect the change and not the end goal. Here the fix is adding title to fix the failure of `/spaceinstant` API.

Let us re-write it based on these

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
RAM1232: @mkesavan13 I had keep a dummy tittle as a string.Genrally it is the conversation title for the 1:1 meeting there is no conversation title so i had keep the condition.I had a thought in my mind like to keep an empty string if the conversation title was not defined.

RAM1232: As @rsarika suggestion can we try like that or we can keep .

RAM1232: @rsarika yes i think so in the test cases they had mentioned only the group space titles not the 1:1 meeting titles. we need to look through this

RAM1232:
> Next, the commit message should reflect the change and not the end goal. Here the fix is adding title to fix the failure of `/spaceinstant` API.

I have to change the commit title or PR title

packages/@webex/plugin-meetings/src/meeting-info/meeting-info-v2.ts
mkesavan13: If this is a bug in SDK that we are fixing, how does Webex Web Client handle it?

They should have hit this issue as well, right? Did we check how does WWC handle this one?

RAM1232: @mkesavan13, At present no,I had seen the webex web client but i could not see the Conversation title or Space Instant API s request body to create an AdHoc meeting like that.

RAM1232: @mkesavan13 In the app.js i did not encounter error and it has created meeting object but due to conversation title missing in the request body the create meeting method could not create the meeting info.Ideally if we give the title for the create AdHoc meeting it will give the response meeting info that was used in the method return.

RAM1232:
After further investigation, we found that this feature is neither supported nor utilized anywhere. The bug fix is not feasible as it cannot support or enable the feature on the Webex Web Client 1:1 Space. Therefore, we have decided not to proceed with fixing this issue.

PR #3783, opened 2024-08-18
fix(plugin-meetings): transcription:started event for every participant

mkesavan13:
# COMPLETES [SPARK-542912](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-542912)

When a meeting host has joined and started transcription already and a non-host participant joins the meeting, the `meeting:receiveTranscription:started` event is not triggered. We had the condition because:
1. A non-host participant cannot `toggleTranscribing` and now, a participant can `turnOnCaptions` even if they're not host. 
2. A non-host participant would receive captions even without having to request Voicea to `turnOnCaptions`

Now, we are removing the host condition because:
1. `turnOnCaptions` is allowed by locus for non-host participants also
2. The original issue of `meeting:receiveTranscription:started` event not triggered is because Voicea doesn't think it has to trigger that event unless we `turnOnCaptions`

Additionally, in this PR we also restrict `setSpokenLanguage` to be called only by a host participant at the SDK level itself instead of an earlier API reject approach. This is just to cut-down an unwanted API call. We already updated this restriction in the documentation [here](https://developer.webex.com/docs/sdks/webex-meetings-sdk-web-advanced-meeting-controls#transcription).

Supporting Screenshot:
Screenshot 2024-08-18 at 6 08 51 AM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manually tested the following,
1. `meeting.startTranscription()` call triggers `meeting:receiveTranscription:started` event for both host and non-host participant (including a guest issuer participant)
2. `meeting.setSpokenLanguage()` rejects if the current user is not a `host`
3. `meeting.setCaptionLanguage()` sends the translated language

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -645,6 +645,10 @@ export default class Meeting extends StatelessWebexPlugin {
         })}event#${EVENT_TRIGGERS.MEETING_STARTED_RECEIVING_TRANSCRIPTION}`
       );
 
+      if (this.getCurUserType() !== 'host') {
+        delete payload.spokenLanguages;
+      }
+
       // @ts-ignore
       this.trigger(EVENT_TRIGGERS.MEETING_STARTED_RECEIVING_TRANSCRIPTION, payload);
     },
@@ -4831,6 +4835,14 @@ export default class Meeting extends StatelessWebexPlugin {
         reject(new Error('Webex Assistant is not enabled/supported'));
       }
 
+      if (this.getCurUserType() !== 'host') {
+        LoggerProxy.logger.error(
+          'Meeting:index#setSpokenLanguage --> Only host can set spoken language'
+        );
+
+        reject(new Error('Only host can set spoken language'));
+      }
+
       try {
         const voiceaListenerLanguageUpdate = (payload) => {
           // @ts-ignore
@@ -4884,10 +4896,8 @@ export default class Meeting extends StatelessWebexPlugin {
           this.setUpVoiceaListeners();
         }
 
-        if (this.getCurUserType() === 'host') {
-          // @ts-ignore
-          await this.webex.internal.voicea.turnOnCaptions(options?.spokenLanguage);
-        }
+        // @ts-ignore
+        await this.webex.internal.voicea.turnOnCaptions(options?.spokenLanguage);
       } catch (error) {
         LoggerProxy.logger.error(`Meeting:index#startTranscription --> ${error}`);
         Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.RECEIVE_TRANSCRIPTION_FAILURE, {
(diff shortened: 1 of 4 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4884,10 +4896,8 @@ export default class Meeting extends StatelessWebexPlugin {
           this.setUpVoiceaListeners();
         }
 
-        if (this.getCurUserType() === 'host') {
-          // @ts-ignore
-          await this.webex.internal.voicea.turnOnCaptions(options?.spokenLanguage);
-        }
+        // @ts-ignore
+        await this.webex.internal.voicea.turnOnCaptions(options?.spokenLanguage);
```
Shreyas281299, line 4900: Is it possible to add some ut for this.

Shreyas281299 (approved):
Looks good. Please see the comment.

packages/@webex/plugin-meetings/src/meeting/index.ts
mkesavan13: Thanks, @Shreyas281299 - Addressed this comment [in this line of code](https://github.com/webex/webex-js-sdk/pull/3783/files#diff-8dae4d0be73ef303fd51e1243f57b86f234ed724645f1ccdea3b26d0e2e51f57R1019).

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1010,14 +1010,13 @@ describe('plugin-meetings', () => {
...
           meeting.areVoiceaEventsSetup = false;
-          meeting.roles = ['COHOST'];
 
           await meeting.startTranscription();
 
           assert.equal(webex.internal.voicea.on.callCount, 4);
           assert.equal(meeting.areVoiceaEventsSetup, true);
           assert.equal(webex.internal.voicea.listenToEvents.callCount, 1);
-          assert.notCalled(webex.internal.voicea.turnOnCaptions);
+          assert.calledOnce(webex.internal.voicea.turnOnCaptions);
```
rarajes2, line 1019: Can we update the test description as well?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1010,14 +1010,13 @@ describe('plugin-meetings', () => {
             state: 'JOINED',
           };
           meeting.areVoiceaEventsSetup = false;
-          meeting.roles = ['COHOST'];
```
rarajes2, line 1013: Should we keep this so that it's clear that it's not a host and even then caption can be enabled?

rarajes2 (approved):
Minor comments. Please address them.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: I don't think that's required because we do not have any conditions at all and we check that the turnOnCaptions is called

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Thanks, @rarajes2. Good catch. Updated the test description

PR #3786, opened 2024-08-19
fix: incorrect icePhase when joinWithMedia() fails on 1st attempt

marcin-bazyl:
# COMPLETES #[SPARK-554031](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-554031)

When joinWithMedia() fails to establish media connection, but succeeds on the retry, Call Analyzer still marks the join with mediaOutcome=failed

Introduced a callback for determining the value of icePhase sent on failure. The callback has different logic when addMedia() is called from joinWithMedia(), because joinWithMedia() knows that it will do a retry as opposed to the public addMedia() method, which also has its own retry logic, but it's different - it only does a retry if TURN server is not used on 1st attempt.

There was also another property called `retriedWithTurnServer` that was purely used by addMedia() internally so I've created a `addMediaData` object inside `Meeting` to store all "addMedia" related things in it.

I've also split the public addMedia() method into a protected addMediaInternal() and public addMedia(). This allows us to easily add more parameters to addMediaInternal() for internal SDK use without impacting the public API of addMedia().

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests and tested manually with web app - simulated failure on 1st attempt and success on retry, confirmed CA marked it as success

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -1459,7 +1459,10 @@ export default class Meeting extends StatelessWebexPlugin {
     this.turnServerUsed = false;
 
     /**
-     * Contains information used during the addMedia() operation
+     * Contains information used during the addMedia() operation:
+     * retriedWithTurnServer - whether retry was done using TURN Discovery
+     * icePhaseCallback - callback for determining the value for icePhase when sending failure event to CA
+     *
      * @instance
      * @type {Object}
      * @private
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -155,7 +155,7 @@ import PermissionError from '../common/errors/permission';
 import {LocusMediaRequest} from './locusMediaRequest';
 import {ConnectionStateHandler, ConnectionStateEvent} from './connectionStateHandler';
 
-// default callback so we don't call an undefined function, but it practice it should never be used
+// default callback so we don't call an undefined function, but in practice it should never be used
 const DEFAULT_ICE_PHASE_CALLBACK = () => 'JOIN_MEETING_FINAL';
 
 const logRequest = (request: any, {logText = ''}) => {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -155,6 +155,9 @@ import PermissionError from '../common/errors/permission';
 import {LocusMediaRequest} from './locusMediaRequest';
 import {ConnectionStateHandler, ConnectionStateEvent} from './connectionStateHandler';
 
+// default callback so we don't call an undefined function, but it practice it should never be used
+const DEFAULT_ICE_PHASE_CALLBACK = () => 'JOIN_MEETING_FINAL';
+
 const logRequest = (request: any, {logText = ''}) => {
   LoggerProxy.logger.info(`${logText} - sending request`);
 
@@ -679,7 +682,11 @@ export default class Meeting extends StatelessWebexPlugin {
     },
   };
 
-  private retriedWithTurnServer: boolean;
+  private addMediaData: {
+    retriedWithTurnServer: boolean;
+    icePhaseCallback: () => string;
+  };
+
   private sendSlotManager: SendSlotManager = new SendSlotManager(LoggerProxy);
   private deferSDPAnswer?: Defer; // used for waiting for a response
   private sdpResponseTimer?: ReturnType<typeof setTimeout>;
@@ -1452,13 +1459,16 @@ export default class Meeting extends StatelessWebexPlugin {
     this.turnServerUsed = false;
 
     /**
-     * Whether retry was done using TURN Discovery.
+     * Contains information used during the addMedia() operation
      * @instance
-     * @type {boolean}
+     * @type {Object}
      * @private
      * @memberof Meeting
      */
-    this.retriedWithTurnServer = false;
+    this.addMediaData = {
+      retriedWithTurnServer: false,
+      icePhaseCallback: DEFAULT_ICE_PHASE_CALLBACK,
+    };
 
     /**
      * Whether or not the media connection has ever successfully connected.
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -155,6 +155,9 @@ import PermissionError from '../common/errors/permission';
 import {LocusMediaRequest} from './locusMediaRequest';
 import {ConnectionStateHandler, ConnectionStateEvent} from './connectionStateHandler';
 
+// default callback so we don't call an undefined function, but it practice it should never be used
```
mkesavan13, line 158:
```suggestion
// default callback so we don't call an undefined function, but in practice it should never be used
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -2996,16 +3038,6 @@ describe('plugin-meetings', () => {
...
-        it('should pass the forceTurnDiscovery parameter to doTurnDiscovery()', async () => {
-          meeting.meetingState = 'ACTIVE';
-
-          // call addMedia() with forceTurnDiscovery=true
-          await meeting.addMedia({}, undefined, true);
-
-          // Check that the value was passed on to doTurnDiscovery()
-          assert.calledOnceWithExactly(meeting.roap.doTurnDiscovery, meeting, false, true);
-        });
-
```
mkesavan13, line 3008: Do we test this elsewhere? Why is this test being removed?

mkesavan13 (commented):
Changes mostly look good to me. Left a few nitpick comments. Just curious. None of these changes ensure that the statsAnalyzer sends the right data. While I understand this PR fixes the root cause, do you think it is still good to have a statsAnalyzer test to ensure it always sends the right joinWithMedia outcome?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1452,13 +1459,16 @@ export default class Meeting extends StatelessWebexPlugin {
...
-     * @type {boolean}
+     * @type {Object}
      * @private
      * @memberof Meeting
      */
-    this.retriedWithTurnServer = false;
+    this.addMediaData = {
+      retriedWithTurnServer: false,
+      icePhaseCallback: DEFAULT_ICE_PHASE_CALLBACK,
+    };
```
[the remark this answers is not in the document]
jor-row: I guess this does exist in the jsdoc of addMediaInternal so not essential that it is moved here too

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4627,7 +4637,14 @@ export default class Meeting extends StatelessWebexPlugin {
         joined = true;
       }
 
-      const mediaResponse = await this.addMedia(mediaOptions, turnServerInfo, forceTurnDiscovery);
+      const mediaResponse = await this.addMediaInternal(
+        () => {
+          return this.joinWithMediaRetryInfo.isRetry ? 'JOIN_MEETING_FINAL' : 'JOIN_MEETING_RETRY';
```
jor-row, line 4642: nitpick: this can just be ```isRetry ? ... ``` because ```this.joinWithMediaRetryInfo``` is destructured on line  4580

jor-row (approved):
Changes look good to me, no blocking issues. 

Do we have an idea of how many/what percentage of joins per day are wrongly categorised by CA as a failure due to this bug? Do we have an estimate of how much the mediaOutcome failure rate should drop once this is in production?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: It's removed, because addMedia() doesn't have the `forceTurnDiscovery` argument anymore. This argument exists now only in addMediaInternal() which is now a non-public function, so I didn't want to call it directly from the tests. I thought the logic of forcing TURN discovery through addMediaInternal() was already tested in one of joinWithMedia() tests, but turns it those tests were stubbing addMediaInternal(), so I've now modified one of them so that it checks that TURN discovery is forced by addMediaInternal() when it's called from joinWithMedia().

packages/@webex/plugin-meetings/src/meeting/index.ts

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
marcin-bazyl: I've added the comment

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: hmm not too sure about that one.... the code in the callback is executed later and we need to use the current up-to-date value of `Meeting.joinWithMediaRetryInfo.isRetry` at the point when that callback is called. If we use the `isRetry` closure then I think it will contain the old value, before we've set it to true in line 4700

marcin-bazyl:
> Changes look good to me, no blocking issues.
> 
> Do we have an idea of how many/what percentage of joins per day are wrongly categorised by CA as a failure due to this bug? Do we have an estimate of how much the mediaOutcome failure rate should drop once this is in production?

I think it's about a thousand calls in the last 7 days:
https://app.eu.amplitude.com/analytics/cisco-cross-project/chart/e-krfq778o?source=copy+url

marcin-bazyl:
> Changes mostly look good to me. Left a few nitpick comments. Just curious. None of these changes ensure that the statsAnalyzer sends the right data. While I understand this PR fixes the root cause, do you think it is still good to have a statsAnalyzer test to ensure it always sends the right joinWithMedia outcome?

I think you've confused stats analyzer with Call Analyzer - the event that we're fixing is sent from waitForMediaConnectionConnected(). I've added now a test that also verifies that the right CA events are sent with the right icePhase in them

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
mkesavan13: Thanks for the explanation, @marcin-bazyl

mkesavan13 (approved):
Looks good to me

PR #3789, opened 2024-08-21
feat(webex-core): hostmap interceptor

Coread:

The hostmap from u2c maps the logical name of the service (these are actually FQDNs) to the hosts. Until recently, the first host listed for each logical name was the same as the logical name. Going forward this will not be the case. Services will continue to use logical names i.e. conversations/locus and the client needs to replace these with the correct host.

Existing methods in webex-core/services do not handle this (e.g. getServiceFromUrl)

I've created a new method replaceHostFromHostmap which looks up the host in the hostmap and replaces it in a supplied URI. I've used this new method in an interceptor, so it will automatically happen for each request.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/index.js
+++ b/packages/@webex/webex-core/src/index.js
@@ -25,6 +25,7 @@ export {
   Services,
   ServiceHost,
   ServiceUrl,
+  HostMapInterceptor,
 } from './lib/services';
 
 export {
--- a/packages/@webex/webex-core/src/lib/services/index.js
+++ b/packages/@webex/webex-core/src/lib/services/index.js
@@ -18,6 +18,7 @@ registerInternalPlugin('services', Services, {
 export {constants};
 export {default as ServiceInterceptor} from './interceptors/service';
 export {default as ServerErrorInterceptor} from './interceptors/server-error';
+export {default as HostMapInterceptor} from './interceptors/hostmap';
 export {default as Services} from './services';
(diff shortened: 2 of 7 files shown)


chrisadubois (approved):

sreenara (changes requested):
@Coread can you please fill in the manual tests you've done in the PR description test section? It would be good to see the real-world implication of this interceptor.

I see that we're replacing the hostname with the first instance that we find in the hostmap. Just for my understanding, what happens when the request fails to the first hostname and we need to use the second entry that's found in the hostmap? Do we plan on adding any logic for this at a later time?

Coread:
> @Coread can you please fill in the manual tests you've done in the PR description test section? It would be good to see the real-world implication of this interceptor.
> 
> I see that we're replacing the hostname with the first instance that we find in the hostmap. Just for my understanding, what happens when the request fails to the first hostname and we need to use the second entry that's found in the hostmap? Do we plan on adding any logic for this at a later time?

This is logic that already exists in the desktop apps and has done for many years, the original intention of which was to provide high availability. The U2C hostmap will always contain at least one entry in the hosts section. All the entries will have the same host, but with different ids. See the below screenshot. i.e. the high availability feature of the catalog was never completed, so there is no point in using it like that, no point in checking other hosts.

Screenshot 2024-08-29 185036

sreenara (approved):

PR #3794, opened 2024-08-26
Modified the outbound flow for guest calling

adhmenon:
# COMPLETES #< [SPARK-52966](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-529662)>

* Guest calling support for Calling SDK.

* Currently in the callind SDK, there was no support for guest calling outbound flows.
* Code has been added so that it checks the outbound flows for a guest user (ie in that case destination address will be empty), and based on that removes the callee object from the post request load.
* Unit tests have also been added.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Make Call Flow (Posting request to mobius)
2. End Call Flow (Delete request to mobius)
3. Media Request 
All the above was done to properly test the entire call flow.

* Open the sample app via the amplify link given - [Sample App](https://pr-3794.d3m3l2kee0btzx.amplifyapp.com/)
* Inside the sample app, click on 'Calling' in the samples
* Select 'guest user' in credentials - 
Screenshot 2024-08-26 at 4 45 00 PM
* Click on 'Generate guest token' and you should see a guest token generated.
Screenshot 2024-08-26 at 4 47 13 PM
* Go this this site [JWT Site](https://jwt.io/) and copy the jwe token (from the encoded area). Paste it into the sample app's JWE token for destination.
* In the sample, app press on 'Initialize calling'
* Once that is done, press on 'register calling'
* Now open up the network logs.
* Once it has been registered, click on the following button  (it should auto enable without any address)- 
Screenshot 2024-08-26 at 4 49 40 PM
* You should now see in the network logs a post request for mobius gets sent with the following payload (similar type0 - 
Screenshot 2024-08-26 at 4 53 25 PM
* Now, proceed to end the call. In the network logs, you should see a DELETE request that gets sent to the mobius URI.

### I certified that

---

### DIFF
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -570,10 +570,15 @@ function createCall(e) {
   console.log(destination.value);
   makeCallBtn.disabled = true;
   outboundEndElm.disabled = false
-  call = line.makeCall({
-    type: 'uri',
-    address: destination.value,
-  });
+  if (serviceIndicator !== 'guestcalling') {
+    line.makeCall({
+      type: 'uri',
+      address: destination.value,
+    });
+  }
+  else {
+    line.makeCall();
+  }
 
   call.on('caller_id', (CallerIdEmitter) => {
     callDetailsElm.innerText = `Name: ${CallerIdEmitter.callerId.name}, Number: ${CallerIdEmitter.callerId.num}, Avatar: ${CallerIdEmitter.callerId.avatarSrc} , UserId: ${CallerIdEmitter.callerId.id}`;
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -574,13 +574,13 @@ function createCall(e) {
   makeCallBtn.disabled = true;
   outboundEndElm.disabled = false
   if (serviceIndicator.value !== 'guestcalling') {
-    line.makeCall({
+    call = line.makeCall({
       type: 'uri',
       address: destination.value,
     });
   }
   else {
-    line.makeCall();
+    call = line.makeCall();
   }
 
   call.on('caller_id', (CallerIdEmitter) => {
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -188,7 +188,7 @@ async function handleServiceSelect(e) {
   const value = e.target.value;
   tokenElm.value = '';
 
-  if (value === 'guestCalling') {
+  if (value === 'guestcalling') {
     guestContainerElm.classList.remove('hidden');
   } else {
     guestContainerElm.classList.add('hidden');
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -248,7 +248,7 @@ async function initCalling(e) {
     level: 'info'
   }
 
-  const {region, country} = credentialsFormElm.elements;
+  const {region, country, guestName} = credentialsFormElm.elements;
 
   const serviceData = {indicator: 'calling', domain: ''};
 
@@ -258,6 +258,10 @@ async function initCalling(e) {
 
   if (serviceDomain.value) {
     serviceData.domain = serviceDomain.value;
+  }
+
+  if(guestName){
+    serviceData.guestName = guestName.value
   }
 
   const callingClientConfig = {
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -260,7 +260,7 @@ async function initCalling(e) {
     serviceData.domain = serviceDomain.value;
   }
 
-  if (guestName) {
+  if (guestName && serviceData.indicator === 'guestcalling') {
     serviceData.guestName = guestName.value
   }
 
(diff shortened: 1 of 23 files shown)


docs/samples/calling/app.js
```diff
@@ -188,7 +188,7 @@ async function handleServiceSelect(e) {
   const value = e.target.value;
   tokenElm.value = '';
 
-  if (value === 'guestCalling') {
+  if (value === 'guestcalling') {
```
adhmenon, line 191: There are changes from Shreyas's PR here as well. Will highlight my changes only.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -191,12 +191,12 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
     activeUrl: string,
     webex: WebexSDK,
-    destination: CallDetails,
     direction: CallDirection,
     deviceId: string,
     lineId: string,
     deleteCb: DeleteRecordCallBack,
-    indicator: ServiceIndicator
+    indicator: ServiceIndicator,
+    destination?: CallDetails
```
adhmenon, line 199: Made this an optional so that guest calling does not send the object.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2081,20 +2092,15 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
-        callee: {
-          type: this.destination.type,
-          address: this.destination.address,
-        },
-        localMedia: {
-          roap: roapMessage,
-          mediaId: uuid(),
-        },
-      },
+      body: this.destination
```
adhmenon, line 2095: Only if the destination is empty, we remove `callee` object payload (as per requirement).

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -69,6 +69,8 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
 
   private callManager: ICallManager;
 
+  private serviceData: CallingClientConfig['serviceData'];
```
adhmenon, line 72: Had to add this private variable over here so that we can access the `GUEST_CALLING` enum value. This is so that we can check for it in the `makeCall()` flow.

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -263,6 +268,16 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
 
       return call;
     }
+    if (this.serviceData?.indicator === ServiceIndicator.GUEST_CALLING) {
```
adhmenon, line 271: Added the 'if' case over here, in case the destination is not present it moves here and then sets up guest calling flow.
Nit: Will need to change the log to include 'guest calling call' probably. Will make the change.

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -282,5 +335,25 @@ describe('Line Tests', () => {
         done(error);
       }
     });
+
+    it('attempt to create call with guest calling service indicator', () => {
```
adhmenon, line 339: Just added some tests to test the latest changes.

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -103,6 +106,7 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
     const serviceData = serviceDataConfig?.indicator
       ? serviceDataConfig
       : {indicator: ServiceIndicator.CALLING, domain: ''};
+    this.serviceData = serviceData;
```
Kesari3008, line 109: One way to avoid creating a new private variable is to move the serviceData declaration under the constructor outside and make it a global variable with a default value of Calling. And within the constructor it will get updated with the value received in the parameter

Kesari3008 (changes requested):
Please upload the logs once testing is finished

packages/calling/src/CallingClient/calling/callManager.ts
```diff
@@ -172,10 +172,10 @@ export class CallManager extends Eventing<CallEventTypes> implements ICallManage
            */
           const lineId = this.getLineId(mobiusEvent.data.deviceId);
           newCall = this.createCall(
-            {} as CallDetails,
```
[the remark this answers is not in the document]
adhmenon: Yup, this is the reason.

adhmenon:
Have temporarily commented out some tests so that we get the PR deployed on Amplify. Will revert these once we have successfully tested it.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2073,6 +2073,17 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    * @param roapMessage
    */
   private post = async (roapMessage: RoapMessage): Promise<MobiusCallResponse> => {
```
Kesari3008, line 2075: We should have some sort of UT even to test different payload being sent out of this post request

Kesari3008 (approved):
Approving with 1 minor comment. Please add the UT request before merge

Shreyas281299 (approved):

PR #3808, opened 2024-09-04
fix(plugin-meetings): add to peer connection bare turn tcp server

k-wasniowski:
# COMPLETES #[SPARK-559031](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-559031)

Sometimes TURN-TCP reachability completes succesfully, but then media fails to connect over TCP. Adding TURN-TCP to peer connection configuration, to verify if it will help in this scenario.

Modifying discovered TURN-TLS to work as bare TURN-TCP server.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested with sample app(Chrome,Edge,Firefox) and blocked UDP to force connections over TCP.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/index.ts
+++ b/packages/@webex/plugin-meetings/src/media/index.ts
@@ -163,6 +163,19 @@ Media.createMediaConnection = (
   // we might not have any TURN server if TURN discovery failed or wasn't done or
   // we might get an empty TURN url if we land on a video mesh node
   if (turnServerInfo?.url) {
+    if (!isBrowser('firefox')) {
+      let bareTurnServer = turnServerInfo.url;
+      bareTurnServer = bareTurnServer.replace('turns', 'turn');
+      bareTurnServer = bareTurnServer.replace('443', '5004');
+
+      iceServers.push({
+        urls: bareTurnServer,
+        username: turnServerInfo.username || '',
+        credential: turnServerInfo.password || '',
+      });
+    }
+
+    // TURN-TLS server
     iceServers.push({
       urls: turnServerInfo.url,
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -163,6 +163,19 @@ Media.createMediaConnection = (
   // we might not have any TURN server if TURN discovery failed or wasn't done or
   // we might get an empty TURN url if we land on a video mesh node
   if (turnServerInfo?.url) {
+    if (!isBrowser('firefox')) {
```
k-wasniowski, line 166: Limiting it to non-firefox, as we might be not handling it properly with Full-ICE. Once we test it properly with Firefox, we should enable it also for Firefox.

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -163,6 +163,19 @@ Media.createMediaConnection = (
...
+      let bareTurnServer = turnServerInfo.url;
+      bareTurnServer = bareTurnServer.replace('turns', 'turn');
```
marcin-bazyl, line 168: maybe better to have a more specific search term like "turns:" just in case if we end up with "turns" string being part of the hostname

marcin-bazyl (approved):
looks good, just 1 minor comment

packages/@webex/plugin-meetings/src/media/index.ts
k-wasniowski: Done

szotrabh (approved):

PR #3818, opened 2024-09-09
Operational buisness metrics

covidal-cisco:

## This pull request addresses https://jira-eng-sjc12.cisco.com/jira/browse/CX-15991

Existing behavioral metrics enable to build and send metrics for user usage of feature. We also need to support two other types of metrics to enable to build and send metrics which target technical details (i.e. the kind of questions "is the websocket session failing", etc.). This is why this PR add supports for operational and business metrics (the difference between the two metrics enable to either tap into metrics which can be stored a long time (and slow to analyze) or metrics stored few days max, but quick, it's just different usages).

From technical changes POV, this PR factorize existing a bunch of code used for behavioral metrics into a "GenericMetric" class which is re-used also for operational and business metrics.

## by making the following changes we enable consumer to send operational and business metrics to webex backends.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/internal-plugin-metrics/src/utils.ts
```diff
@@ -15,3 +21,66 @@ export const generateCommonErrorMetadata = (error) => {
 
   return error;
 };
+
+/**
+ * Returns the deviceId from our registration with WDM.
+ * @returns {string} deviceId or empty string
+ */
+export function getDeviceId(): string {
+  const {url} = this.device;
```
torgeadelin, line 30: this will break, the previous implementation relies on this.webex.internal.device which can only be accessed inside a plugin

covidal-cisco: Yeah - as for the `this.version` below. I need to fix that.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+import {EventContext, TaggedEvent, EventPayload} from './metrics.types';
+
+const {getOSVersion, getBrowserName, getBrowserVersion} = BrowserDetection();
+
+/**
+ * @description top-level abstract class to handle Metrics and common routines.
+ * @export
+ * @class GenericMetrics
+ */
+export default abstract class GenericMetrics extends StatelessWebexPlugin {
```
torgeadelin, line 15: note: i feel like this could get confusing in the future.. we have this generic metrics class that is used for behavioural, operational and business but not call diagnostic.

Not sure what the best approach could be here..

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -102,7 +102,7 @@ export interface ClientEvent {
   options?: SubmitClientEventOptions;
 }
 
-export interface BehavioralEventContext {
+export interface EventContext {
```
torgeadelin, line 105: suggestion: I don't think I like this generalisation `EventContext`.. the problem with this is that the consumer can easily assume this context applies for all metrics, but that's not true... CA metrics don't have this AFAIK.

I'm all in for reusability and common types/structures, but we should be mindful how this affect the understanding of how the plugin works.

Also, if there's a new requirement in the future to support some sort of other metrics which are sent via a diff service than metrics-a, it would get confusing.

torgeadelin (commented):
Code looks ok, I'm just worried about over generalising.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
covidal-cisco: It centralizes the logic of the vast majority of the metric backends (3/4) which doesn't have common concepts, so it feels right to call it generic in this respect. The fact it's not used by one backend doesn't make it less generic or confusing IMO: there is always specific code for specific things, it doesn't make generic things less relevant (as soon as they are mostly used, of course).

That said I'm 100% open to other suggestions, but so far it seems the cleanest approach to me.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
covidal-cisco: Fair point, especially since `EVentContext` isn't super precise name either: it doesn't really tells what's kind of context it holds. What about `DeviceContext`? It better defines what's inside and I don't think it would make things more confusing with future requirements.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
chrisadubois: I don't really have a problem with GenericMetrics. It won't be directly used by anything (consumers will just call the behavioral/business/operational) functions as far as i can tell, so it seems fine to me.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+    payload,
+  }: {
+    type: [string];
+    name: string;
+    payload: EventPayload;
+  }): TaggedEvent {
+    let allTags: EventPayload = payload;
+    allTags = merge(allTags, this.getBrowserDetails());
+
+    const event: TaggedEvent = {
```
chrisadubois, line 133: I don't think you need to type this inline, the return type will handle that for you

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -190,7 +201,7 @@ export type SubmitBehavioralEvent = (args: {
   agent: MetricEventAgent;
   target: string;
   verb: MetricEventVerb;
-  payload?: BehavioralEventPayload;
+  payload?: EventPayload;
```
chrisadubois, line 204: nice

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -112,23 +112,33 @@ export interface BehavioralEventContext {
   };
 }
 
-export interface BehavioralEvent {
-  context: BehavioralEventContext;
+export type EventPayload = Record<string, string | number | boolean>;
```
chrisadubois, line 115: i know this seems a bit odd, and it's kind of a thought - but I think we could imagine sending some different types of payloads here, perhaps we should create a facade for the event Payload type which would go in the Record, and use that. That way if we need to change anything, we won't have to update the string | number | boolean in numerous locations, but instead, just update the internal type and it will propagate

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -200,9 +211,8 @@ export type SubmitClientEvent = (args: {
 }) => Promise<any>;
 
 export type SubmitOperationalEvent = (args: {
-  name: OperationalEvent['name'];
-  payload?: RecursivePartial<OperationalEvent['payload']>;
-  options?: any;
+  name: OperationalEvent['metricName'];
```
chrisadubois, line 214: could we add a constants file for all these accessors, such as `metricName`, 'payload', etc? for ease of use and lower chance of typos

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
```diff
@@ -63,6 +67,10 @@ class Metrics extends WebexPlugin {
       this.callDiagnosticMetrics = new CallDiagnosticMetrics({}, {parent: this.webex});
       // @ts-ignore
       this.behavioralMetrics = new BehavioralMetrics({}, {parent: this.webex});
+      // @ts-ignore
```
chrisadubois, line 70: IMO - the actual `creation` of these objects *should be* configurable, i.e., we dont NEED to create a businessMetrics object if we are never going to use it. The consumer should be able to decide if these are needed.

chrisadubois (changes requested):
I didn't review the tests, overall very good, very few blocking request changes, most others is commentary which I am open to discussing ! in fact, all of them I am open to discussing, but right now, marking as request changes . Feel free to ping me and we can have a discussion on them.

Sorry for being a stickler on it - I've just seen so many problems with metrics governance, and want to try and get this right. I really appreciate you working on it and making these improvements and expansion to the module.

packages/@webex/internal-plugin-metrics/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@webex/internal-plugin-metrics",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
covidal-cisco: It's a mistake (for local testing with wxcc), and it slip in the PR. Will remove, thanks for spotting this.

packages/@webex/internal-plugin-metrics/src/business/business-metrics.ts
```diff
@@ -0,0 +1,31 @@
...
+export default class BusinessMetrics extends GenericMetrics {
+  /**
+   * Submit a buisness metric to our metrics endpoint.
+   * @param {string} name of the metric
+   * @param {EventPayload} user payload of the metric
+   * @returns {Promise<any>}
+   */
+  public submitBusinessEvent({name, payload}: {name: string; payload: EventPayload}) {
+    const event = {
+      type: ['business'],
```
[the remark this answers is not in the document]
covidal-cisco: You mean at the GenericMetrics level, then each subclass specify it in the ctor, i.e. `this.type = 'business'` ?

Sure I can do that. I'm just a little worried to add cognitive complexity (i.e. the contract that a subclass of GenericMetrics has to define the metric type in the ctor) because we would still need to use it (i.e. `this.type`) here (as each metrics has a specific shape, so the event is build in each metric implementation anyway).

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
covidal-cisco: Yeah I should probably have clarified this: GenericMetric is really just an internal implementation thing and should never be exposed to the consumer, absolutely.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return this.clientMetricsBatcher.request(event);
+  }
+
+  /**
+   * Returns the deviceId from our registration with WDM.
+   * @returns {string} deviceId or empty string
+   */
+  protected getDeviceId(): string {
+    const {url} = this.device;
```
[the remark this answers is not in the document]
covidal-cisco: Oh yeah that's a good point. It's very easy to cache that in each metric instance (so it would be computed 3 times max). Will do.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Returns the context object to be submitted with all metrics.
+   * @returns {DeviceContext}
+   */
+  protected getContext(): DeviceContext {
+    return {
+      app: {
+        version: this.version,
```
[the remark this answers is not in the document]
covidal-cisco: @lialang-cisco any input here? I guess this is something we got wrong for behavioral metrics in existing code then?

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+   * @param {string} metric name
+   * @param {EventPayload} user payload
+   * @returns {EventPayload}
+   */
+  protected createTaggedEventObject({
+    type,
+    name,
+    payload,
+  }: {
+    type: [string];
```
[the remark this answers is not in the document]
covidal-cisco: Sure, will do. Thanks

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+    name: string;
```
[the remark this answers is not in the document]
covidal-cisco: This is a tricky one (we discussed this internally with ex-Webex, new-Wxcc guys). Conclusion of our discussion were that it's true for behavioral metrics, as those should be clearly defined at PM level (i.e. for a new feature, what's the usage, etc.)

But it feels that should be an allow-list provided by the consumer, as it doesn't feel totally right to have to modify the SDK to enable names from various consumers. (i.e. what is a non-Cisco consumer wants to use it, etc., generally speaking it make a central put-it-all place which wouldn't necessary scale with lot of engineer -- we hit this kind of problems on Webex desktop). But this is probably a good mid-term discussion to have outside of this PR.

Regarding operational and business metrics, however, it's a different story: those metrics can be added by engineers in any places when needed, and removed when not needed anymore, based on a bug report of something else. In this respect it doesn't really make sense to modify the SDK every time for this (well, again, if there was an allow-list thing provided by the consumer that would be okay, but we don't have this ATM)

packages/@webex/internal-plugin-metrics/src/operational/config.ts
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable import/prefer-default-export */
+
+export const OPERATIONAL_LOG_IDENTIFIER = 'operational-events -> ';
```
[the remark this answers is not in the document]
covidal-cisco: TBH I followed the existing pattern (as it is currently for behavioral metrics) but I don't really see the point either. I'd prefer to not have those config files at all (and not necessary have those metrics implementation in separate folder, I don't see the point ATM)

Let me know if you're okay with that, I'll do the change. Thanks

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
covidal-cisco: So you're thinking about extra config flags when we instantiate the plugin/SDK?

covidal-cisco: Or more a lazy-apporach: we create the metric instance the first time the user try to send an event -- which makes more sense I guess.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
covidal-cisco: Sure - so you think about something in that spirit?

```
type Payload = string | number | boolean [ | something-else-later ]
export type EventPayload = Record<string, Payload>
```

covidal-cisco: (I'm not 100% sure if extending the type here would make sense as we depends of the backend eventually, but I get your point from a coding engineering perspective)

packages/@webex/internal-plugin-metrics/src/behavioral/behavioral-metrics.ts
```diff
@@ -166,14 +33,14 @@ export default class BehavioralMetrics extends StatelessWebexPlugin {
...
+    payload?: EventPayload;
   }) {
-    this.logger.log(
-      BEHAVIORAL_LOG_IDENTIFIER,
-      `BehavioralMetrics: @submitBehavioralEvent. Submit Behavioral event: ${product}.${agent}.${target}.${verb}`
-    );
-    const behavioralEvent = this.createEventObject({product, agent, target, verb, payload});
-
-    return this.clientMetricsBatcher.request(behavioralEvent);
+    const name = `${product}.${agent}.${target}.${verb}`;
```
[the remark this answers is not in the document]
covidal-cisco: Makes sense, I mostly re-used this existing code (from behavioral) so I don't know why we wouldn't do that already, I'll try to dig a bit and find out.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
covidal-cisco: Actually I was wondering about those: I saw this pattern in previous code I change so I didn't want to change it, but what's wrong with `string`? (I get the point of using this syntax for complex types/subject to changes though). 

Otherwise, sure, I can move those under a constant... Or isn't the following a better pattern? (It's an actual question - I don't have much experience in this code base/TS):

```
type MetricName = string;

type OperationalEvent {
....
 metricName: MetricName
```

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
```diff
@@ -0,0 +1,143 @@
...
+    payload: EventPayload;
+  }): TaggedEvent {
+    let allTags: EventPayload = payload;
+    allTags = merge(allTags, this.getBrowserDetails());
```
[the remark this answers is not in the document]
covidal-cisco: So what happens ATM (in the current code) is that the browser detection code would return an empty name/version and undefined values like the weight/height. We could consider adding a check to figure out if we're in a nodejs context. I'm not quite sure yet what we would want to do, i.e. to not add those fields at all in the metrics, or add just the name "node", etc. 

Perhaps something to think about as a follow-up? (Considering it's how it works currently for behavioral metrics)

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -102,7 +102,7 @@ export interface ClientEvent {
   options?: SubmitClientEventOptions;
 }
 
-export interface BehavioralEventContext {
+export interface DeviceContext {
   app: {version: string};
```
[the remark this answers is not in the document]
covidal-cisco: Problem then is that it create a tight coupling from the SDK to the consumer, i.e. a third-party consumer wouldn't be able to set the version it wants if it's not part of the SDK enum (or I misunderstood what you meant here)

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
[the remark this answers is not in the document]
chrisadubois: Yeah that tracks. I would *definitely* accept a consumer provided allow-list as a middle ground.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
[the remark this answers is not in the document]
chrisadubois: I would request @sreenara or @arun3528 on this one, I don't have any concerns for node env, but they may.

packages/@webex/internal-plugin-metrics/src/generic-metrics.ts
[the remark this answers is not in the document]
chrisadubois: well but deviceIds *could* get updated mid flow, so it needs a way to bust the cache.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
chrisadubois: yes that's what I was imagining.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
chrisadubois: no because then they are not linked together, i.e., OperationalEvent['metricName'] type may change at some point in time from string to enum, or something along these lines

packages/@webex/internal-plugin-metrics/src/new-metrics.ts
chrisadubois: lazy approach works fine for me as well, however, there's 2 usecases

1. pre-warming (use config to get the object already created, but just the ones the consumer wants)
2. lazy created similar to a JIT-C, which is great

However, the performance impact of something like this is so infinitesmal it's hard to argue about it

tldr: i'm cool with a lazy approach

chrisadubois (approved):
i no longer have any blocking requests, everything else is an improvement. Assuming this is a non-breaking change for web client, it seems good to me. It'd be nice to have a good consumption document added into this package to explain how it all works.

I have not tested this change locally with the web client. Without that, it may trigger the need to revert (for example if we can't upgrade the SDK because the metrics plugin doesn't work in our repo as is, however you have indicated it's a non breaking change)

let me know if you want to talk about any of the remaining comments.

needs another approval from SDK team.

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
covidal-cisco: Sure yeah. I don't see any example of this in metric code yet (looking at other metrics). Should I add a simple

```
const MetricNameKey = "metricName"
```

then replace all occurrences of `OperationalEvent['metricName']` into `OperationalEvent[MetricNameKey]` then? (similarly for behavioral and business of course).

chrisadubois: yes

mkesavan13: Is this addressed? I do not see this change happened yet?

packages/@webex/internal-plugin-metrics/test/unit/spec/business/business.ts
```diff
@@ -0,0 +1,120 @@
+import sinon from 'sinon';
```
mkesavan13, line 1: Can we rename this file to be consistent with what we already have?
`business-metrics.ts` instead of `business.ts`

Same comment for `operational.ts` as well

mkesavan13 (approved):
Left a few nitpicks. Otherwise, looks good to me

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
covidal-cisco: No - after trying it it turns out that putting a symbol (variable) instead of a literal string changes the semantic and break the build. Discussed with @chrisadubois we'll leave as it is.

packages/@webex/internal-plugin-metrics/test/unit/spec/business/business.ts
covidal-cisco: Oh yes, good point. Will do, thanks.

PR #3823, opened 2024-09-10
fix(meetings): sending ice candidate errors with add media success metric

marcin-bazyl:
Submitting info about candidate errors also with `js_sdk_add_media_success` metric event (we already send it with `js_sdk_add_media_failures` metric) so that we can see how often we have issues with DNS resolution of Linus hostname used for TLS


k-wasniowski (approved):

arun3528 (approved):
LGTM just change the title to mention plugin-meetings

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -7022,6 +7022,7 @@ export default class Meeting extends StatelessWebexPlugin {
         await this.mediaProperties.getCurrentConnectionInfo();
       // @ts-ignore
       const reachabilityStats = await this.webex.meetings.reachability.getReachabilityMetrics();
+      const iceCandidateErrors = Object.fromEntries(this.iceCandidateErrors);
```
mkesavan13, line 7025: Is there a possibility that this `iceCandidateErrors` be `undefined` in any case? 
Screenshot 2024-09-12 at 11 58 56 AM

mkesavan13 (approved):
One minor comment and rest looks good to me. Approved

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: no, it's never undefined, it's initialised to an empty Map in constructor and is cleared when new connections are created, but never set to undefined

mkesavan13: Great. Thanks for the clarification

PR #3838, opened 2024-09-16
feat(meetings): do ip version detection in parallel with reachability

marcin-bazyl:
# COMPLETES #[SPARK-562699](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-562699)

Orpheus team wants us to tell them if we are on an ipv4 or ipv6 network (or both) when we call the discovery/v1/clusters API. To know this we would need to gather local ICE candidates before making that call - that can take some time (few hundred milliseconds), so for now we will do this in parallel to making the API call just so that we can measure how long it takes and send the timings with reachability metrics.

- starting ip network detection when we do reachability, but not waiting for it to complete
- sending the information about how long it took in reachability metric

This means that now we'll be sending the `ipver` parameter to Locus when we're joining a meeting based on the results of ip network detection done during last reachability check. Previously we've just been always sending ipver=0 which means "we don't know if we're on ipv4 or ipv6 network")

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -149,6 +149,11 @@ export default class Reachability extends EventsScope {
   public async gatherReachability(): Promise<ReachabilityResults> {
     // Fetch clusters and measure latency
     try {
+      // kick off ip version detection. For now we don't await it, as we're doing it
+      // to gather the timings and send them with our reachability metrics
+      // @ts-ignore
+      this.webex.internal.device.ipNetworkDetector.detect();
+
       const {clusters, joinCookie} = await this.getClusters();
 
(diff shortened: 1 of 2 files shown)


chrisadubois (approved):

mkesavan13 (approved):
Looks good to me. Do we have a separate Jira for sending the ip version later when we have the duration measured through this metric?

marcin-bazyl:
> Looks good to me. Do we have a separate Jira for sending the ip version later when we have the duration measured through this metric?

Not yet, because depending on the timings we see from the metrics, we'll decide if and how we'll do it. The Jira once created will be added to the parent epic of this PR's jira

PR #3854, opened 2024-09-24
fix(meetings): webex.internal.device doesn't know when user is in a meeting

marcin-bazyl:
# COMPLETES #[SPARK-564815](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564815)

Raised from a customer BEMS case:
When users have configured inactivity timeout in control hub for example for 30 mins, when they join a meeting, after 30 min they get logged out by the SDK and web app doesn't update the UI, so they remain in the meeting but cannot do any operations like mute/unmute/start recording (they just fail) and also Mercury connection is dropped, so they won't get any updates about the meeting.

Internal device plugin has the inactivity timer and it relies on meetingStarted/meetingEnded methods to be called when a user joins or leaves (or gets dropped from) a meeting to avoid being logged out while in a meeting. These methods were not being called from anywhere, so I've added calls to them in plugin-meetings.

This solution is not ideal as it won't work correctly if user joins more than one meeting in parallel, but AFAIK none of the SDK clients do that, so for now it's fine. A bigger refactor of internal device plugin would be needed to fix this more properly.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5232,6 +5232,9 @@ export default class Meeting extends StatelessWebexPlugin {
         this.meetingFiniteStateMachine.join();
         this.setupLocusMediaRequest();
 
+        // @ts-ignore
+        this.webex.internal.device.meetingStarted();
+
         LoggerProxy.logger.log('Meeting:index#join --> Success');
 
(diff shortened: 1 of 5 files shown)


jor-row (approved):

sreenara (approved):
LGTM

PR #3879, opened 2024-10-02
fix(calling): added handling of system mute

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-530539 

Handling system mute for calling

Listeners for system mute and user mute were added along with appropriate conditions to reflect the right mute status.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

I tested using the samples page and tried user mute and system mute.
1. User mute works as before, clicking on the mute button triggers mute-unmute.
2. When the system mute is done, the mute button takes effect based on the mute state and user mute cannot be done while the call is system muted.
3. In the user muted state, the system mute will take effect but won't impact the mute button text.

https://github.com/user-attachments/assets/897c397c-2305-4d8a-bfb4-6c40fb72571c

### I certified that

---


Shreyas281299 (commented):
Changes look, but I tried to use my usb headphone which has a mute button on it and when I muted it didnt reflect in the samples app. Apart from muting from setting is there any other scenario that we should test?

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2764,7 +2763,7 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    * @param localAudioTrack -.
    */
   public mute = (localAudioStream: LocalMicrophoneStream): void => {
-    if (localAudioStream) {
+    if (localAudioStream && !localAudioStream.systemMuted) {
       localAudioStream.setUserMuted(!this.muted);
       this.muted = !this.muted;
```
sreenara, line 2768: Would it make sense to mute the call object when a system mute occurs? Or are we leaving this completely to the developer?

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2764,7 +2763,7 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    * @param localAudioTrack -.
    */
   public mute = (localAudioStream: LocalMicrophoneStream): void => {
-    if (localAudioStream) {
+    if (localAudioStream && !localAudioStream.systemMuted) {
```
sreenara, line 2766: we'll need a UT for this.

docs/samples/calling/app.js
```diff
@@ -610,6 +609,16 @@ function createCall(e) {
     document.getElementById('remote-audio').srcObject = new MediaStream([track]);
   });
 
+  localAudioStream.on('system-mute-state-change', (systemMuted) => {
```
sreenara, line 612: Are we not going to make any changes to the call object on receiving this?

sreenara (changes requested):
Please fix the title. there is a typo

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: When I mute the call object on system mute, it messes up the mute state entirely. Tried it and it was messing up the mute state.

docs/samples/calling/app.js
Kesari3008: No changes are required. It works without the call object changes

packages/calling/src/CallingClient/calling/call.ts
sreenara: @Kesari3008 let's discuss this.. wouldn't we have a situation where the call object doesn't show it as mute but the call is actually muted?

packages/calling/src/CallingClient/calling/types.ts
```diff
@@ -290,7 +290,7 @@ export interface ICall extends Eventing<CallEventTypes> {
    *
    * @param localAudioStream - The local audio stream to mute or unmute.
    */
```
sreenara, line 292: Please fix the description.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2763,10 +2762,26 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+          log.warn(`Call is muted by user already ${this.getCorrelationId()}.`, {
+            file: CALL_FILE,
+            method: 'mute',
+          });
+        }
+      } else if (!localAudioStream.systemMuted) {
+        localAudioStream.setUserMuted(!this.muted);
+        this.muted = !this.muted;
+      } else {
+        log.warn(`Call is system muted ${this.getCorrelationId()}.`, {
```
sreenara, line 2780:
```suggestion
        log.info(`Call is muted on the system - ${this.getCorrelationId()}.`, {
```

sreenara (changes requested):
Would it make sense to provide an object where a developer could specify which mute they wanted instead of having a parameter only for system based mute?

packages/calling/src/CallingClient/calling/call.ts
sreenara: This comment hasn't been addressed.

Kesari3008:
> Please fix the title. there is a typo

Fixed the title

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: done

packages/calling/src/CallingClient/calling/types.ts
Kesari3008: done

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2763,10 +2762,26 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    *
    * @param localAudioTrack -.
    */
```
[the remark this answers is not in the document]
Kesari3008: done

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2542,7 +2542,6 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
 
   private registerListeners(localAudioStream: LocalMicrophoneStream) {
     localAudioStream.on(LocalStreamEventNames.OutputTrackChange, this.updateTrack);
-
```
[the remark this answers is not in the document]
Kesari3008: taken care now

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2761,14 +2762,35 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
    */
-  public mute = (localAudioStream: LocalMicrophoneStream): void => {
-    if (localAudioStream) {
+  public mute = (localAudioStream: LocalMicrophoneStream, muteType?: MUTE_TYPE): void => {
+    if (!localAudioStream) {
+      log.warn(`Did not find a local stream while muting the call ${this.getCorrelationId()}.`, {
+        file: CALL_FILE,
+        method: 'mute',
+      });
+    }
```
sreenara, line 2829: Shouldn't we return in this condition?

packages/calling/src/CallingClient/calling/types.ts
```diff
@@ -155,6 +155,11 @@ export enum TransferType {
   CONSULT = 'CONSULT',
 }
 
+export enum MUTE_TYPE {
```
sreenara, line 158: Could we use the new way of doing enums?

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -225,10 +226,10 @@ describe('Call Tests', () => {
     expect(call).toBeTruthy();
     /* After creation , call manager should have 1 record */
     expect(Object.keys(callManager.getActiveCalls()).length).toBe(1);
-    call.mute(localAudioStream);
+    call.mute(localAudioStream, MUTE_TYPE.USER);
```
sreenara: These tests should not require changes and must still pass.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2761,14 +2762,35 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+    if (muteType === MUTE_TYPE.SYSTEM) {
```
sreenara, line 2830: Won't this fail if muteType is undefined?

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -718,6 +719,89 @@ describe('Call Tests', () => {
       {file: 'call', method: 'updateMedia'}
     );
   });
+
+  it('test system mute and user mute different scnearios', async () => {
```
sreenara, line 732: Good test

sreenara (changes requested):
Please see the comments

packages/calling/src/CallingClient/calling/call.test.ts
Kesari3008: That's correct, removed them now

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: Yes we should, missed to add that

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: Why would it be undefined? We are planning to send Usermute and system mute. If nothing is passed, it will be usermute and go to else block. But for system it always has to be there

packages/calling/src/CallingClient/calling/types.ts
Kesari3008: Changing it for just this and having other enums in the old format is alright. For new SDKs we can follow the appaorach but for calling SDK also if we do it then it will be 2 different formats.

packages/calling/src/CallingClient/calling/call.ts
sreenara: muteType is an optional argument. That's the reason for the question.

sreenara (approved):

Shreyas281299 (approved):

PR #3893, opened 2024-10-09
fix: send rtc metrics at the beginning every 5 seconds

k-wasniowski:
… seconds

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -3213,7 +3213,8 @@ export default class Meeting extends StatelessWebexPlugin {
           options: {meetingId: this.id},
         });
       }
-      this.rtcMetrics?.sendNextMetrics();
+      const numberOfReportsToSend = 30000 / 5000; // 30 seconds divided by 5 seconds
+      this.rtcMetrics?.sendNextMetrics(numberOfReportsToSend);
       this.updateLLMConnection();
     });
 
--- a/packages/@webex/plugin-meetings/src/rtcMetrics/index.ts
+++ b/packages/@webex/plugin-meetings/src/rtcMetrics/index.ts
@@ -34,7 +34,7 @@ export default class RtcMetrics {
 
   connectionId: string;
 
-  shouldSendMetricsOnNextStatsReport: boolean;
+  shouldSendMetricsOnNextStatsReport: number;
 
   /**
    * Initialize the interval.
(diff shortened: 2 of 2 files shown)


packages/@webex/plugin-meetings/src/rtcMetrics/index.ts
```diff
@@ -151,7 +152,7 @@ export default class RtcMetrics {
    */
   private resetConnection() {
     this.connectionId = uuid.v4();
-    this.shouldSendMetricsOnNextStatsReport = true;
+    this.shouldSendMetricsOnNextStatsReport = 1;
```
marcin-bazyl, line 155: we should set this to 30000/5000 here too, right? This is called from constructor so this is the default used when we join the meeting straight in, without lobby.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3213,7 +3213,8 @@ export default class Meeting extends StatelessWebexPlugin {
           options: {meetingId: this.id},
         });
       }
-      this.rtcMetrics?.sendNextMetrics();
+      const numberOfReportsToSend = 30000 / 5000; // 30 seconds divided by 5 seconds
```
marcin-bazyl, line 3216: it's really annoying that the constant for calling getStats every 5s is buried deep in WCME and not even exported from it, but at least the 30s timer for uploading is defined in rtcMetrics/index.ts in SDK so we could define a constant and use it here instead of hardcoding 30000. Maybe define a constant for the 5s inside rtcMetrics/index.ts too so that both are kept together? Or don't bother with the math and just set numberOfReportsToSend=5 or 6, after all we just want "a few" reports with data from initial meeting phase, don't care about the exact number of seconds the data is from. If you go with that, then maybe define a constant so the same value is used also when we don't go through lobby (see my other comment)

packages/@webex/plugin-meetings/src/rtcMetrics/index.ts
```diff
@@ -95,7 +96,7 @@ export default class RtcMetrics {
         // this is the first useful set of data (WCME gives it to us after 5s), send it out immediately
         // in case the user is unhappy and closes the browser early
         this.sendMetricsInQueue();
-        this.shouldSendMetricsOnNextStatsReport = false;
+        this.shouldSendMetricsOnNextStatsReport -= 1;
```
marcin-bazyl, line 99: maybe worth changing the condition in line 95 to `this.shouldSendMetricsOnNextStatsReport > 0` just in case if that variable ever became negative somehow?

PR #3903, opened 2024-10-10
fix(voicemail): added /v2.0/ check for BWRKS xsiendipoint

akulakum:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-568088

The BroadWorks API was failing due to an issue where the v2.0 version string was appearing twice in the endpoint URL. This was happening because some customers were configuring the XSI endpoint in the Partner Hub with /v2.0/ instead of the expected /v2.0. As a result, this duplication was causing the API calls to fail.

We have added an additional check inside the getXsiActionEndpoint function to handle this scenario:

If the xsiEndpoint contains /v2.0/, we slice it to remove the trailing / and avoid duplicating the version string in the URL.
This ensures that the URL is correctly formatted and prevents further issues with BroadWorks API calls.
This fix now accounts for both /v2.0 and /v2.0/, providing more flexibility in the endpoint configuration while resolving the API issue.

< DESCRIBE YOUR CHANGES >

Before fix
Screenshot 2024-10-10 at 11 45 24 AM

After fix
Screenshot 2024-10-10 at 11 47 49 AM

Screenshot 2024-10-10 at 12 12 08 PM

Screenshot 2024-10-10 at 4 52 27 PM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

We conducted thorough testing to ensure the changes work correctly in both scenarios where the XSI endpoint URL includes /v2.0 and /v2.0/. The following tests were performed:

Manual Tests:

Scenario 1: XSI Endpoint with /v2.0:
Verified that the BroadWorks API successfully processes requests when the XSI endpoint URL is configured with /v2.0.
Checked that the API responds correctly and the issue with the double versioning in the URL does not occur.

Scenario 2: XSI Endpoint with /v2.0/:
Verified that the API works when the XSI endpoint URL is configured with /v2.0/ (extra slash).
Ensured that the additional logic we implemented slices the trailing slash and properly processes the request.

### I certified that

---

### DIFF
--- a/packages/calling/src/common/Utils.ts
+++ b/packages/calling/src/common/Utils.ts
@@ -119,6 +119,7 @@ import {
   SCIM_USER_FILTER,
   WEBEX_API_PROD,
   WEBEX_API_BTS,
+  BW_XSI_ENDPOINT_VERSION_WITH_SLASH,
 } from './constants';
 import {Model, WebexSDK} from '../SDKConnector/types';
 import SDKConnector from '../SDKConnector';
@@ -1110,8 +1111,13 @@ export async function getXsiActionEndpoint(
 
         let xsiEndpoint = response[DEVICES][0][SETTINGS][BW_XSI_URL];
 
-        if (response[DEVICES][0][SETTINGS][BW_XSI_URL].endsWith(BW_XSI_ENDPOINT_VERSION)) {
-          xsiEndpoint = response[DEVICES][0][SETTINGS][BW_XSI_URL].slice(0, -5);
+        const xsiUrl = response[DEVICES][0][SETTINGS][BW_XSI_URL];
+
+        // Check if it ends with specific version and slice accordingly
+        if (xsiUrl.endsWith(BW_XSI_ENDPOINT_VERSION)) {
+          xsiEndpoint = xsiUrl.slice(0, -5); // Remove 'v2.0'
+        } else if (xsiUrl.endsWith(BW_XSI_ENDPOINT_VERSION_WITH_SLASH)) {
+          xsiEndpoint = xsiUrl.slice(0, -6); // Remove 'v2.0/'
         }
 
         return xsiEndpoint;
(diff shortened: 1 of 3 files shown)


packages/calling/src/common/Utils.ts
```diff
@@ -1110,8 +1111,13 @@ export async function getXsiActionEndpoint(
 
         let xsiEndpoint = response[DEVICES][0][SETTINGS][BW_XSI_URL];
 
-        if (response[DEVICES][0][SETTINGS][BW_XSI_URL].endsWith(BW_XSI_ENDPOINT_VERSION)) {
-          xsiEndpoint = response[DEVICES][0][SETTINGS][BW_XSI_URL].slice(0, -5);
+        const xsiUrl = response[DEVICES][0][SETTINGS][BW_XSI_URL];
+
+        // Check if it ends with specific version and slice accordingly
+        if (xsiUrl.endsWith(BW_XSI_ENDPOINT_VERSION)) {
+          xsiEndpoint = xsiUrl.slice(0, -5); // Remove 'v2.0'
```
Shreyas281299, line 1118: Rather than making it specific to version 2, can we use regex to find if it ends with a version, otherwise we would have to re-visit this once the version changes to v3

Shreyas281299 (changes requested):
Please see comments

packages/calling/src/common/Utils.ts
Waseemraj: Screenshot 2024-10-14 at 10 44 32 AM

Hey Shreyas, as per the 
Broadworks solution guide version will be v2
https://help.webex.com/en-us/article/no4dlj3/Webex-for-Cisco-BroadWorks-Solution-Guide#reference-template_9e47637b-bf3e-4cd4-b19e-89fbc4e0e7d1

This will not change, but we will reach out to Broadworks PM for confirmation and make the necessary changes in the next PR.

Shreyas281299 (approved):

mkesavan13 (approved):

sreenara (approved):

PR #3906, opened 2024-10-11
feat(calling): add rtcMetrics in the call

rarajes2:
# COMPLETES #< [SPARK-529191](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-529191) >

Adds RtcMetrics to the Calling package.

Calling SDK is now able to send the RTC Metrics.

**Screenshot**
Calling-RtcMetrics

**Note**: Please check the JIRA for [HAR](https://jira-eng-gpk2.cisco.com/jira/secure/attachment/1324588/calling-rtc-metrics.har) file.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested both Calling and Meeting stats being sent and attached the HAR files on the JIRA.

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
+++ b/packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
@@ -2,6 +2,7 @@
 import uuid from 'uuid';
 import * as CallDiagnosticUtils from '../call-diagnostic/call-diagnostic-metrics.util';
 import RTC_METRICS from './constants';
+import {IdType, IMetricsAttributes} from '../metrics.types';
 
 const parseJsonPayload = (payload: any[]): any | null => {
   try {
@@ -28,7 +29,9 @@ export default class RtcMetrics {
 
   webex: any;
 
-  meetingId: string;
+  meetingId?: string;
+
+  callId?: string;
 
   correlationId: string;
 
@@ -40,16 +43,27 @@ export default class RtcMetrics {
    * Initialize the interval.
    *
    * @param {object} webex - The main `webex` object.
-   * @param {string} meetingId - The meeting id.
+   * @param {IdType} Ids - Meeting or Calling id.
    * @param {string} correlationId - The correlation id.
    */
-  constructor(webex, meetingId, correlationId) {
+  constructor(webex, {meetingId, callId}: IdType, correlationId) {
     // `window` is used to prevent typescript from returning a NodeJS.Timer.
     this.intervalId = window.setInterval(this.sendMetricsInQueue.bind(this), 30 * 1000);
     this.meetingId = meetingId;
+    this.callId = callId;
     this.webex = webex;
     this.correlationId = correlationId;
     this.resetConnection();
+  }
+
+  /**
+   * Updates the call identifier with the provided value.
+   *
+   * @param {string} callId - The new call identifier to set.
+   * @returns {void}
+   */
+  public updateCallId(callId: string) {
+    this.callId = callId;
   }
 
   /**
@@ -160,6 +174,21 @@ export default class RtcMetrics {
    * @returns {void}
    */
   private sendMetrics() {
+    const metricsAttributes: IMetricsAttributes = {
+      type: 'webrtc',
+      version: '1.1.0',
+      userId: this.webex.internal.device.userId,
+      correlationId: this.correlationId,
+      connectionId: this.connectionId,
+      data: this.metricsQueue,
+    };
+
+    if (this.meetingId) {
+      metricsAttributes.meetingId = this.meetingId;
+    } else if (this.callId) {
+      metricsAttributes.callId = this.callId;
+    }
+
     this.webex.request({
       method: 'POST',
       service: 'unifiedTelemetry',
@@ -169,17 +198,7 @@ export default class RtcMetrics {
         appId: RTC_METRICS.APP_ID,
       },
       body: {
-        metrics: [
-          {
-            type: 'webrtc',
-            version: '1.1.0',
-            userId: this.webex.internal.device.userId,
-            meetingId: this.meetingId,
-            correlationId: this.correlationId,
-            connectionId: this.connectionId,
-            data: this.metricsQueue,
-          },
-        ],
+        metrics: [metricsAttributes],
       },
     });
   }
--- a/packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
+++ b/packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
@@ -27,7 +27,7 @@ describe('RtcMetrics', () => {
     clock = sinon.useFakeTimers();
     window.setInterval = setInterval;
     webex = new MockWebex();
-    metrics = new RtcMetrics(webex, 'mock-meeting-id', 'mock-correlation-id');
+    metrics = new RtcMetrics(webex, {meetingId: 'mock-meeting-id'}, 'mock-correlation-id');
     anonymizeIpSpy = sandbox.spy(metrics, 'anonymizeIp');
   });
 
@@ -151,5 +151,46 @@ describe('RtcMetrics', () => {
     // and another stats-report should trigger another upload of the metrics
     metrics.addMetrics({ name: 'stats-report', payload: [STATS_WITH_IP] });
     assert.callCount(webex.request, 3);
+  });
+
+  describe('RtcMetrics - callId', () => {
+    let metrics: RtcMetrics;
+    let webex: MockWebex;
+    let clock;
+    let sandbox;
+  
+    beforeEach(() => {
+      clock = sinon.useFakeTimers();
+      window.setInterval = setInterval;
+      webex = new MockWebex();
+      metrics = new RtcMetrics(webex, {callId: 'mock-call-id'}, 'mock-correlation-id');
+      sandbox = sinon.createSandbox();
+    });
+  
+    afterEach(() => {
+      sandbox.restore();
+    });
+  
+    it('sendMetrics should send a webex request with callId', () => {
+      assert.notCalled(webex.request);
+  
+      metrics.addMetrics(FAKE_METRICS_ITEM);
+      (metrics as any).sendMetrics();
+  
+      assert.callCount(webex.request, 1);
+      assert.calledWithMatch(webex.request, sinon.match.has('headers', {
+        type: 'webrtcMedia',
+        appId: RTC_METRICS.APP_ID,
+      }));
+      assert.calledWithMatch(webex.request, sinon.match.hasNested('body.metrics[0].data[0].payload', FAKE_METRICS_ITEM.payload));
+      assert.calledWithMatch(webex.request, sinon.match.hasNested('body.metrics[0].callId', 'mock-call-id'));
+      assert.calledWithMatch(webex.request, sinon.match.hasNested('body.metrics[0].correlationId', 'mock-correlation-id'));
+    });
+
+    it('should update the callId correctly', () => {
+      const newCallId = 'new-call-id';
+      metrics.updateCallId(newCallId);
+      assert.strictEqual(metrics.callId, newCallId);
+    });
   });
 });
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -6376,7 +6376,7 @@ export default class Meeting extends StatelessWebexPlugin {
   private async createMediaConnection(turnServerInfo, bundlePolicy?: BundlePolicy) {
     this.rtcMetrics = this.isMultistream
       ? // @ts-ignore
-        new RtcMetrics(this.webex, this.id, this.correlationId)
+        new RtcMetrics(this.webex, {meetingId: this.id}, this.correlationId)
       : undefined;
 
     const mc = Media.createMediaConnection(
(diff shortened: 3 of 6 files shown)


packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -160,6 +168,21 @@ export default class RtcMetrics {
...
+      userId: this.webex.internal.device.userId,
+      correlationId: this.correlationId,
+      connectionId: this.connectionId,
+      data: this.metricsQueue,
+    };
+
+    if (this.meetingId) {
+      metricsAttributes['meetingId'] = this.meetingId;
+    } else if (this.callId) {
+      metricsAttributes['callId'] = this.callId;
```
rarajes2, line 183: Adding these conditionally as we don't want to sent the key `meetingId` or `callingId` with value as `undefined`. A key will only be added if the value exists for it.

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -346,7 +346,10 @@ describe('Call Tests', () => {
     expect(mockInternalMediaCoreModule.RoapMediaConnection).toBeCalledOnceWith(
       roapMediaConnectionConfig,
       roapMediaConnectionOptions,
-      expect.any(String)
+      expect.any(String),
+      expect.any(Function),
+      expect.any(Function),
+      expect.any(Function)
```
rarajes2, line 352: These are added as we are using anonymous/arrow functions. Adding the function directly was not possible as we have a bit different logic to create the RtcMetrics object. Please see the screenshot attached below for reference.

rtcmetrics-functions

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1927,6 +1933,33 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+  private forceSendStatsReport = async ({callFrom}: {callFrom?: string}) => {
+    const loggerContext = {
+      file: CALL_FILE,
+      method: this.forceSendStatsReport.name,
+    };
+
+    try {
+      await this.mediaConnection.forceRtcMetricsSend();
+      log.info(`Successfully uploaded available webrtc telemetry statistics`, loggerContext);
+      log.info(`callFrom: ${callFrom}`, loggerContext);
```
rarajes2, line 1951: Simply logging this here as it was throwing linting error due to not being used. This function is created but not being used as of now, this will be utilized later. @Kesari3008 Please correct me if I mentioned wrong.

rarajes2:
Style checks are failing for other packages,

sreenara (changes requested):

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -28,7 +33,9 @@ export default class RtcMetrics {
 
   webex: any;
 
-  meetingId: string;
+  meetingId?: string;
```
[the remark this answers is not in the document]
rarajes2: The only change included here is the `callId` so, I think we won't be making much changes and RTC Stats would be standard format data. If some drastic change is required such as the format of the data being sent or the way data is sent is different then we can make it an abstract/parent class and have the children implement own logic.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1936,6 +1969,16 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
    * @param settings.debugId - String.
    */
   private initMediaConnection(localAudioTrack: MediaStreamTrack, debugId?: string) {
+    let rtcMetrics: RtcMetrics;
+
+    // Adding a check to ensure that the callId is not the default callId as initMediaConnection is called before callId is set
+    const interval = setInterval(() => {
```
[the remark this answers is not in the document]
rarajes2: Yes, it gets sent only after the call establishes. Created a method in the RtcMetric class to update the callId and invoking the same in the `setCallId` in the call class. Tested it and it's working as expected. Thanks

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: @rarajes2 Yes that's correct

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -40,18 +58,23 @@ export default class RtcMetrics {
    * Initialize the interval.
    *
    * @param {object} webex - The main `webex` object.
-   * @param {string} meetingId - The meeting id.
+   * @param {IdType} Ids - Meeting or Calling id.
    * @param {string} correlationId - The correlation id.
    */
-  constructor(webex, meetingId, correlationId) {
+  constructor(webex, {meetingId, callId}: IdType, correlationId) {
```
Kesari3008, line 64: Why do we need to add a new type for this? We can just have 2 optional parameters here: meetingId, callId and based on what is passed while creating object of this class, rtctstats will be sent with that

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -15,6 +15,22 @@ const parseJsonPayload = (payload: any[]): any | null => {
   }
 };
 
+interface IdType {
+  meetingId?: string;
+  callId?: string;
+}
+
+interface IMetricsAttributes {
```
Kesari3008, line 23: Do we have any types or interface file where this can be moved?

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -15,6 +15,22 @@ const parseJsonPayload = (payload: any[]): any | null => {
...
+  type: string;
+  version: string;
+  userId: string;
+  correlationId: string;
+  connectionId: string;
```
Kesari3008, line 28: Would connectionId be applicable for calling cases?

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
```diff
@@ -40,18 +58,23 @@ export default class RtcMetrics {
...
     // `window` is used to prevent typescript from returning a NodeJS.Timer.
     this.intervalId = window.setInterval(this.sendMetricsInQueue.bind(this), 30 * 1000);
     this.meetingId = meetingId;
+    this.callId = callId;
     this.webex = webex;
     this.correlationId = correlationId;
     this.resetConnection();
   }
 
+  public updateCallId(callId: string) {
```
Kesari3008, line 74: Did we need this to update callId for outbound call cases since callId is not generated initially?

Kesari3008 (changes requested):

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
sreenara: @Kesari3008 yes, please see my comment above

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
rarajes2: That can also be done, I have kept them in one object. Both are optional.

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
rarajes2: This value is local to rtcMetrics, not being passed from meeting also so, kept it as it is.

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
rarajes2: Moved to `packages/@webex/internal-plugin-metrics/src/metrics.types.ts`

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
sreenara: @rarajes2 can you add a description for this?

packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
```diff
@@ -27,7 +27,7 @@ describe('RtcMetrics', () => {
     clock = sinon.useFakeTimers();
     window.setInterval = setInterval;
     webex = new MockWebex();
-    metrics = new RtcMetrics(webex, 'mock-meeting-id', 'mock-correlation-id');
+    metrics = new RtcMetrics(webex, {meetingId: 'mock-meeting-id'}, 'mock-correlation-id');
```
sreenara, line 30: Shouldn't we have a test with callId too to cover all conditions?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6376,7 +6376,7 @@ export default class Meeting extends StatelessWebexPlugin {
   private async createMediaConnection(turnServerInfo, bundlePolicy?: BundlePolicy) {
     this.rtcMetrics = this.isMultistream
       ? // @ts-ignore
-        new RtcMetrics(this.webex, this.id, this.correlationId)
+        new RtcMetrics(this.webex, {meetingId: this.id}, this.correlationId)
```
sreenara, line 6379: No changes in tests needed?

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1927,6 +1937,33 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     }
   }
 
+  /**
+   * Media failed, so collect a stats report from webrtc using the wcme connection to grab the rtc stats report
```
sreenara: We don't use wcme. Could we fix this description?

packages/calling/src/CallingClient/calling/call.test.ts
sreenara: @rarajes2 I was able to avoid this error by doing this..
```
  private handleClose() {
    this.rtcMetrics.closeMetrics();
  }

  private handleSendMetric() {
    this.rtcMetrics.sendMetricsInQueue();
  }

  /* istanbul ignore next */
  /**
   * Initialize Media Connection.
   *
   * @param settings -.
   * @param settings.localAudioTrack - MediaStreamTrack.
   * @param settings.debugId - String.
   */
  private initMediaConnection(localAudioTrack: MediaStreamTrack, debugId?: string) {
    const mediaConnection = new RoapMediaConnection(
      {
        skipInactiveTransceivers: true,
        iceServers: [],
        iceCandidatesTimeout: ICE_CANDIDATES_TIMEOUT,
        sdpMunging: {
          convertPort9to0: true,
          addContentSlides: false,
          copyClineToSessionLevel: true,
        },
      },
      {
        localTracks: {audio: localAudioTrack},
        direction: {
          audio: 'sendrecv',
          video: 'inactive',
          screenShareVideo: 'inactive',
        },
      },
      debugId || `WebexCallSDK-${this.correlationId}`,
      (data) => this.rtcMetrics.addMetrics(data),
      this.handleClose,
      this.handleSendMetric,
    );
```

Is it not the same?

Kesari3008 (approved):

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: I had tried this way as well. With add metrics also we should be able to simply pass `this.rtcMetrics.addMetrics` as this is equivalent to `(data) => this.rtcMetrics.addMetrics(data)` but was getting the issue with this keyword/object.

packages/@webex/internal-plugin-metrics/src/rtcMetrics/index.ts
rarajes2: Done

packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
rarajes2: Added one

packages/@webex/plugin-meetings/src/meeting/index.ts
rarajes2: [This change](https://github.com/webex/webex-js-sdk/pull/3906#discussion_r1802833741) covers this

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Updated

packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
```diff
@@ -152,4 +152,39 @@ describe('RtcMetrics', () => {
...
+      webex = new MockWebex();
+      metrics = new RtcMetrics(webex, {callId: 'mock-call-id'}, 'mock-correlation-id');
+      sandbox = sinon.createSandbox();
+    });
+  
+    afterEach(() => {
+      sandbox.restore();
+    });
+  
+    it('sendMetrics should send a webex request with callId', () => {
```
sreenara, line 174: We aren't testing updateCallId()

sreenara (changes requested):
Just one comment about the new UT.

packages/@webex/internal-plugin-metrics/test/unit/spec/rtcMetrics/index.ts
rarajes2: Added

sreenara (approved):

chrisadubois (approved):

PR #3911, opened 2024-10-11
fix(calling): failover immediately after 4 keepalive failures for cc flow

rarajes2:
# COMPLETES #< [SPARK-569580](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-569580) >

Failover to secondary mobius server immediately after 4 keep alive failure for contact center flow.

Here we have introduced a new variable/flag to check we need to failover immediately or not. That flag gets set only when the flow is contact centre and keepalive is failing. It gets reset back when the immediate failover logic runs.

Able to failover to backup mobius server immediately after 4 keepalive failure.

**Note**: Please check the [HAR](https://jira-eng-gpk2.cisco.com/jira/secure/attachment/1324962/failover-after-keepalive-public-repo.har) file attached on the JIRA.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested on the samples page by blocking the primary server domain and after 4 keepalive failure registration switched to secondary server.

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/registration/register.test.ts
+++ b/packages/calling/src/CallingClient/registration/register.test.ts
@@ -732,7 +732,7 @@ describe('Registration Tests', () => {
       expect(reg.getStatus()).toBe(RegistrationStatus.INACTIVE);
       expect(reg.keepaliveTimer).not.toBe(timer);
 
-      webex.request.mockRejectedValueOnce(failurePayload).mockResolvedValue(successPayload);
+      webex.request.mockResolvedValue(successPayload);
 
       jest.advanceTimersByTime(REG_TRY_BACKUP_TIMER_VAL_FOR_CC_IN_SEC * SEC_TO_MSEC_MFACTOR);
       await flushPromises();
--- a/packages/calling/src/CallingClient/registration/register.ts
+++ b/packages/calling/src/CallingClient/registration/register.ts
@@ -76,6 +76,7 @@ export class Registration implements IRegistration {
   private registerRetry = false;
   private reconnectPending = false;
   private isCCFlow = false;
+  private failoverImmediately = false;
 
   /**
    */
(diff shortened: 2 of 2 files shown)


Shreyas281299 (approved):
LGTM

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -732,7 +732,7 @@ describe('Registration Tests', () => {
       expect(reg.getStatus()).toBe(RegistrationStatus.INACTIVE);
       expect(reg.keepaliveTimer).not.toBe(timer);
 
-      webex.request.mockRejectedValueOnce(failurePayload).mockResolvedValue(successPayload);
+      webex.request.mockResolvedValue(successPayload);
```
sreenara, line 735: How does this change ensure we are failing over immediately? I'm not clear about how this statement change covers the new flow.

sreenara (changes requested):
Please add some more details to the description on why we are using a new variable to control this immediate failover. That'll help others to return to this PR in the future and understand the change better.

packages/calling/src/CallingClient/registration/register.test.ts
rarajes2: The failure payload was for the primary server try but now we don't try primary server for the cc flow when it comes from the keepalive failure, here we show the success for the secondary server immediately.

sreenara (commented):
> Tested on the samples page by blocking the primary server domain and after 4 keepalive failure registration switched to secondary server.

@rarajes2 in the HAR file 'failover-after-keepalive-public-repo.har', it looks like the SDK tries to register with the primary twice before failing over to secondary. Why is that happening?

rarajes2:
> > Tested on the samples page by blocking the primary server domain and after 4 keepalive failure registration switched to secondary server.
> 
> @rarajes2 in the HAR file 'failover-after-keepalive-public-repo.har', it looks like the SDK tries to register with the primary twice before failing over to secondary. Why is that happening?

@sreenara Fixed that. It was happening because we have logic to reregister (a function) after keepalive failures which first tries to attempt primary server then starts the failover. Updated the logic in both the PR and also updated JIRA with latest HAR files.

sreenara (approved):

PR #3920, opened 2024-10-15
fix: added type for LocusInfo inside BasicMeetingInformation

marcin-bazyl:
This is to address PR comment from PR#3918

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -143,7 +143,14 @@ export type BasicMeetingInformation = {
   environment: string;
   id: string;
   locusUrl: string;
-  locusInfo: any; // it's only a very small subset of the locus info, see what's populated in the destroy() method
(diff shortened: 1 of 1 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1092,7 +1099,6 @@ export default class Meetings extends WebexPlugin {
       meetingInfo: cloneDeep(meeting.meetingInfo),
       locusInfo: {
         // locusInfo can be quite big, so keep just the minimal info
-        sequence: meeting.locusInfo?.sequence,
```
marcin-bazyl, line 1095: I've removed the sequence, because it's not really needed, I only added it in the first place, because I thought it was a simple string and I thought that maybe it will be useful in the future, more for debugging then actual metrics, but it turns out that sequence is an object that can contain also arrays so I don't want to be bothered with storing it.

chrisadubois (approved):

PR #3942, opened 2024-10-28
feat(plugin-meetings): Introduce SDK changes to call WCME and Locus APIs when current user steps away or returns

antsukanova:
# COMPLETES # [SPARK-559645](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-559645)

Introduce SDK changes to call WCME and Locus APIs when the current user steps away or returns.

Related PRs:

- [PR-3942](https://github.com/webex/webex-js-sdk/pull/3942): Introduce SDK changes to call WCME and Locus APIs when current user steps away or returns
- [PR-3946](https://github.com/webex/webex-js-sdk/pull/3946): Step Away feature unit tests

This PR only contains updates for SDK logic for the sender mostly. 
Since the PR is still under development with unit tests (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574651), it will be presented as a draft.
Also, for now, it will contain local linking `"link:../../../../webrtc-media-core"` for better dx until [WCME changes](https://sqbu-github.cisco.com/CPaaS/web-client-media-engine/pull/418) will be ready and release.

Also, either this PR or a separate one will be updated [logic for test-app](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574653).

How to test: 
1. Create PMR with guest join from SDK and native/web from your user.
2. Click on the `Be right back true / Be right back false` buttons in "Participants, Breakout Sessions, and Events published" section. Check the console with data about Be right back requests (for example `locus-info:updateSelf->event#SELF_MEETING_BRB_CHANGED`)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- tested flow with good/bad requests using a sample app.
- tested flow with native / web clients and a sample app.

### I certified that

---


packages/@webex/media-helpers/package.json
```diff
@@ -22,7 +22,7 @@
     "deploy:npm": "yarn npm publish"
   },
   "dependencies": {
-    "@webex/internal-media-core": "2.11.3",
+    "@webex/internal-media-core": "link:../../../../webrtc-media-core",
```
antsukanova: temporary until WCME changes is not merged

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3330,6 +3330,26 @@ function toggleBreakout() {
   }
 }
 
+function stepAwayTrue() {
+  const meeting = getCurrentMeeting();
+
+  meeting.stepAway(true).then((res) => {
+    console.log(`meeting.stepAway(true) done successfully with result: ${res}`);
+  }).catch((err) => {
+    console.log(`meeting.stepAway(true) done unsuccessfull with err: ${err}`);
```
edvujic: typo: did you want to say `unsuccessfully` here?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3492,6 +3506,28 @@ export default class Meeting extends StatelessWebexPlugin {
     return this.members.admitMembers(memberIds, locusUrls);
   }
 
+  /**
+   * Manages step away updates for the current participant.
+   *
+   * @param {boolean} enabled - Indicates whether the use is stepped away or not.
```
edvujic: typo: did you mean `user` here? also probably we should say something like `indicates whether the user *has* stepped away or no`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -8653,6 +8689,23 @@ export default class Meeting extends StatelessWebexPlugin {
...
   }
 
+  /**
+   * Sets the step away status for the specified media type.
+   *
+   * @param {MediaType} mediaType - The type of media (e.g., VideoMain).
+   * @param {boolean} enabled - Whether the step away status is enabled.
+   * @returns {void}
+   */
+  private setSendStepAway(mediaType: MediaType, enabled: boolean) {
```
edvujic: Do we need to send `MediaType`as a parameter here? Isn't it always going to be `VideoMain`?

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -1333,6 +1333,19 @@ export default class LocusInfo extends EventsScope {
...
 
+      if (parsedSelves.updates.stepAwayChanged) {
+        this.emitScoped(
+          {
+            file: 'locus-info',
+            function: 'updateSelf',
+          },
+          LOCUSINFO.EVENTS.SELF_MEETING_STEP_AWAY_CHANGED,
+          {
+            brb: parsedSelves.current.brb,
```
brycetham, line 1344: Just a comment on terminology. In some places we say "step away" and in other places we say "brb". I'm wondering if we should try to be consistent and refer to everything as "brb" in the SDK.

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -916,4 +916,29 @@ export default class MeetingRequest extends StatelessWebexPlugin {
...
+   * Sends a request to set the step away (be right back) status.
+   *
+   * @param {Object} options - The options for the step away request.
+   * @param {boolean} options.enabled - Whether the step away status is enabled.
+   * @param {string} options.locusUrl - The URL of the locus.
+   * @param {string} options.deviceUrl - The URL of the device.
+   * @param {string} options.selfId - The ID of the participant.
+   * @returns {Promise}
+   */
+  getStepAway({enabled, locusUrl, deviceUrl, selfId}: StepAwayOptions) {
```
brycetham: Is "get" the right word to use here? Should this be `setStepAway` or `sendStepAway` or simply just `stepAway`?

packages/@webex/plugin-meetings/src/locus-info/index.ts
antsukanova: @brycetham, it would be nice to discuss together and align boundaries for this part.

packages/@webex/plugin-meetings/src/meeting/request.ts
antsukanova: @brycetham I feel `sendStepAway` or `stepAway` is better, too. I would go with `sendStepAway` because I see there are a lot of similar naming patterns in the `request.ts` file, and it clears action in this function better.

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
```diff
@@ -159,6 +162,9 @@ SelfUtils.breakoutsChanged = (previous, current) =>
 SelfUtils.interpretationChanged = (previous, current) =>
   !isEqual(previous?.interpretation, current?.interpretation) && !!current?.interpretation;
 
+SelfUtils.brbChanged = (previous, current) =>
+  !isEqual(previous?.brb, current?.brb) && !!current?.brb;
```
SomeBody16:
```suggestion
SelfUtils.brbChanged = (previous, current) =>
  !isEqual(previous?.brb, current?.brb) && current?.brb !== undefined;
```

Because with current code:
```
(previous -> current | result)
undefined -> true    | brbChanged = true
true      -> false   | brbChanged = false // this is not true
false     -> true    | brbChanged = true
```

Catched with unit tests

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3330,6 +3330,26 @@ function toggleBreakout() {
...
 
+function beRightBackTrue() {
+  const meeting = getCurrentMeeting();
+
+  meeting.beRightBack(true).then((res) => {
+    console.log(`meeting.beRightBack(true) done successfully with result: ${res}`);
+  }).catch((err) => {
+    console.log(`meeting.beRightBack(true) done unsuccessfull with err: ${err}`);
+  });
+}
```
SomeBody16: Why not just

```suggestion
function beRightBack(enabled) {
  const meeting = getCurrentMeeting();

  meeting.beRightBack(enabled).then((res) => {
    console.log(`meeting.beRightBack(${enabled}) done successfully with result: ${res}`);
  }).catch((err) => {
    console.log(`meeting.beRightBack(true) done unsuccessfull with err: ${err}`);
  });
}
```

And later
```typescript
<button id="beRightBack" type="button" onclick="beRightBack(true)">Step Away true</button>
```

SomeBody16 (changes requested):
With 2 comments

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
antsukanova: Great catch, will update this part

docs/samples/browser-plugin-meetings/app.js
antsukanova: I completed this part in the simplest way possible and created a task to update it properly in https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574653. Don't feel it's worth changing here if this still should be rewritten.

brycetham: Could we throw in a TODO or something in the code with a link to the JIRA so we have a reference to it?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3290,6 +3290,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      Trigger.trigger(
+        this,
+        {
+          file: 'meeting/index',
+          function: 'setUpLocusInfoSelfListener',
+        },
+        EVENT_TRIGGERS.MEETING_BRB_UPDATE
+      );
+      if (this.mediaProperties.videoStream) {
+        this.setSendBrb(MediaType.VideoMain, payload.brb.enabled);
```
brycetham: If I'm understanding this correctly, we don't call this function (which sends the "away" state to Homer) until after we get the Locus DTO telling us our own BRB state has changed. By this point, the web client will likely have already muted the camera, which means other clients will briefly see our avatar before we send the "away" state. Ideally, we'd like to send the "away" state first, so other clients will see the "away" state straight away. Perhaps, instead of calling this function here, could we send the "away" state as soon as the call to `this.meetingRequest.sendBrb` succeeds...

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3492,6 +3506,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * @throws {Error} - Throws an error if the request fails.
+   */
+  public beRightBack(enabled: boolean) {
+    return this.meetingRequest
+      .sendBrb({
+        enabled,
+        locusUrl: this.locusUrl,
+        deviceUrl: this.deviceUrl,
+        selfId: this.selfId,
+      })
```
brycetham, line 3555: ...so like here (before the catch)? I assume the web client will listen to this promise before muting the camera, so if we send the away state before the promise is returned, it'll be reflected in the other clients before our camera is muted...

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -8653,6 +8689,23 @@ export default class Meeting extends StatelessWebexPlugin {
...
   }
 
+  /**
+   * Sets the be right back status for the specified media type.
+   *
+   * @param {MediaType} mediaType - The type of media (e.g., VideoMain).
+   * @param {boolean} enabled - Whether the brb status is enabled.
+   * @returns {void}
+   */
+  private setSendBrb(mediaType: MediaType, enabled: boolean): void {
```
brycetham: ...and then I guess we won't need this private function anymore, since it'll just be a part of the `beRightBack` API. I think that ends up being clearer code-wise as well so we don't have this logic separated out in a different function.

brycetham (commented):
Overall looks good! I left a rather lengthy comment about the timing of sending the "away" state. Ideally, we'd like to send the "away" state as soon as possible, and with how it is right now, I suspect we'd see a noticeable delay.

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @brycetham good idea, Bryce, thank you. I believe that the client works always assuming that sending a request will be successful and tries to change the UI accordingly asap, and only if it isn't successful will it change it to something else. (like a case if the server is not available or any other similar issue)

docs/samples/browser-plugin-meetings/app.js
antsukanova: @brycetham  @SomeBody16  Added Todo

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3492,6 +3506,36 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  public beRightBack(enabled: boolean) {
+    return this.meetingRequest
+      .sendBrb({
+        enabled,
+        locusUrl: this.locusUrl,
+        deviceUrl: this.deviceUrl,
+        selfId: this.selfId,
+      })
+      .then((payload) => {
+        if (this.isMultistream && this.mediaProperties.webrtcMediaConnection) {
```
marcin-bazyl: just to double check: what about non-multistream cases? do we support this feature for 1-1 calls (that's the only type of calls webex web app currently has that are still transcoded). Do we need to signal anything to Edonus via the SDP or some other way? I guess not, but I just want to be sure, maybe it's worth asking the Edonus team about this.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3297,11 +3297,11 @@ export default class Meeting extends StatelessWebexPlugin {
           file: 'meeting/index',
           function: 'setUpLocusInfoSelfListener',
         },
-        EVENT_TRIGGERS.MEETING_BRB_UPDATE
+        EVENT_TRIGGERS.MEETING_BRB_UPDATE,
```
brycetham: Is this event only going to be used when BRB is updated for self? If so, I'm wondering if we should rename it to `EVENT_TRIGGERS.MEETING_BRB_SELF_UPDATE`/`meeting:self:brbUpdate`.

brycetham (commented):
I agree with Marcin about checking transcoded meetings, though we don't have to do it in this PR (as long as you stick a TODO in there). Left a small question about the event we're emitting but other than that looks good.

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: Created jira for that: [SPARK-578667](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-578667)
As for now we'll focus on finishing this feature for multistream

FYI @brycetham

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -738,6 +738,49 @@ describe('plugin-meetings', () => {
...
+        }
+
+        assertBrb(true)
+        assertBrb(false)
+      })
+
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state changed', () => {
+        const assertBrb = (enabled) => {
+          const selfWithBrbChanged = cloneDeep(self);
+          selfWithBrbChanged.controls.brb = enabled
```
antsukanova: @SomeBody16 I'm not sure if the code style should allow missing parts without a ";". I believe there is something not set up correctly in this logic for JavaScript files, or I’m missing something here. But if this is a linter issue, we need to report that

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -738,6 +738,49 @@ describe('plugin-meetings', () => {
...
+            LOCUSINFO.EVENTS.SELF_MEETING_BRB_CHANGED,
+            { brb: enabled }
+          )
+        }
+
+        assertBrb(true)
+        assertBrb(false)
+      })
+
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state changed', () => {
```
antsukanova: @SomeBody16 why you do not use `locusInfo.self = undefined;` here for test?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8531,6 +8531,33 @@ describe('plugin-meetings', () => {
           });
         });
       });
+
+      describe('#brb', () => {
```
antsukanova: @SomeBody16  What do you think about adding a test case for API existence? I see that the other test cases follow this pattern.

    it('should have #beRightBack', () => {
        ...
        });

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8531,6 +8531,33 @@ describe('plugin-meetings', () => {
...
+        it('send brb request', () => {
+          meeting.meetingRequest.locusDeltaRequest = sinon.stub().resolves();
+          meeting.locusUrl = 'locus url'
+          meeting.deviceUrl = 'device url'
+          meeting.selfId = 'self id'
+
+          const assertBrb = (enabled) => {
```
antsukanova: @SomeBody16  I'm not sure if we should test the exact implementation details for this test case. We simply need to verify its functionality since we don’t know which internal requests API is used to make brb work.

For example, you can look at how the `toggleReactions` public API is tested. It doesn’t focus on the HTTP request details, and I haven’t found any similar approaches, and I believe this should also have something similar. Let’s ensure our tests are consistent with the existing logic for testing the public API.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8531,6 +8531,33 @@ describe('plugin-meetings', () => {
           });
         });
       });
+
+      describe('#brb', () => {
+        it('send brb request', () => {
```
antsukanova: @SomeBody16  I’m not sure why the test is named "brb." If we are testing the "beRightBack" public method, I would expect the test name to be consistent with the API. Additionally, I would suggest changing the naming of the test case because it focuses on one specific aspect of the method, while there are multiple things involved (request, wcme api). As I mentioned in another comment, the current naming feels for me overly detailed regarding the implementation. I would suggest making it more abstract, focusing on the logic instead. Something like "should enable/disable the brb with the correct data and return a promise."
Open to discussion.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8611,6 +8638,29 @@ describe('plugin-meetings', () => {
           );
         });
 
+        it('listens to the brb state changed event', () => {
+          meeting.sendSlotManager.setSourceStateOverride = sinon.stub()
+          meeting.mediaProperties.webrtcMediaConnection = true
```
antsukanova: @SomeBody16  How does this line affect the logic for `setUpLocusInfoSelfListener` tests?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -8611,6 +8638,29 @@ describe('plugin-meetings', () => {
...
+
+          const assertBrb = (enabled) => {
+            meeting.locusInfo.emit(
+              { file: 'locus-info', function: 'updateSelf' },
```
antsukanova: @SomeBody16 I believe this should be "function: 'test', file: 'test'"

docs/samples/browser-plugin-meetings/index.html
```diff
@@ -217,6 +217,14 @@ <h2 class="collapsible">
                 </div>
               </div>
 
+              <div>
+                <fieldset>
+                  <legend>Step Away</legend>
+                  <input type="checkbox" id="brb" onclick="toggleBrb()">
```
brycetham, line 223: I feel like this should be a button and not a checkbox, since it'll be a button in the web client. The button should reflect the brb status of the current user, and would be a good way for us to test whether we are able to accurately show our brb status in the UI, as well as handle any potential race condition that might occur because of it.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -83,6 +84,36 @@ export default class SendSlotManager {
...
+    }
+
+    if (state) {
+      slot.setSourceStateOverride(state);
+    } else {
+      slot.clearSourceStateOverride();
+    }
+
+    this.LoggerProxy.logger.info(
+      `SendSlotsManager->setSourceStateOverride#set source state override to ${state}`
```
brycetham: Nit: I know we only override for main video for now, but might be good to log the media type here too in case it changes in the future.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
brycetham: Should this say "when brb state *not* changed"?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
brycetham: Haven't written an SDK unit test in a while... just wondering if it would be more appropriate to put these tests under the "Public Api Contract" describe block? It would test the multistream vs non-multistream case. There's also "locusMediaRequestStub" that you might be able to reuse there.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
brycetham: Prettier detects and fixes this automatically for me, though I'm using VS Code. I'd be curious to see why it isn't catching stuff like this for some of y'all.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
edvujic: Hey @brycetham, I agree with moving the test to the other part, but since `stepAway` will be only available for multistream meetings, does it make sense to test the multistream vs non-multistream scenario?

docs/samples/browser-plugin-meetings/index.html
antsukanova: It's a good comment, and I believe it's worth changing. We can improve this part in this task and not block the RP with this issue. https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574653

docs/samples/browser-plugin-meetings/app.js
antsukanova: @edvujic there is another fuction now without this part, `toggleBrb`, but thanks, there was a typo

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3330,6 +3330,23 @@ function toggleBreakout() {
...
+async function toggleBrb() {
+  const meeting = getCurrentMeeting();
+
+  if (meeting) {
+    const enabled = document.getElementById('brb').checked;
+    try {
+      const result = await meeting.beRightBack(enabled);
+      console.log(`meeting.beRightBack(${enabled}): SUCCESS | result: ${result}`);
+    } catch (error) {
+      console.error(`meeting.beRightBack({${enabled}): ERROR`, error);
```
antsukanova: I believe that uppercase messages can attract attention and may be useful. However, if everything in the console were written this way, I'm not sure it would be helpful or easy to read. I haven't seen any patterns for logs using this approach, but I like the idea of having concise messages. I have updated this part a little bit, if anything please let me know @SomeBody16 .

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @edvujic thanks Ed, updated this part, appreciate your review

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @edvujic We have updated this logic. Now we only use the `setSourceStateOverride` function with `VideoMain`.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -548,6 +548,41 @@ describe('plugin-meetings', () => {
...
+        describe('#beRightBack', () => {
+          it('should have #beRightBack', () => {
+            assert.exists(meeting.beRightBack);
+          });
+
+          beforeEach(() => {
+            meeting.meetingRequest.locusDeltaRequest = sinon.stub().resolves();
+          });
+
+          it('should enable #beRightBack with the correct data and return a promise', async () => {
```
antsukanova: The test description says here that function should enable brb with the correct data, and then I'd expect that we will check here if we called the request with proper data on something similar, like in admit test cases, for example. What do you think, @edvujic? 
Otherwise, now I do not see how we check data correctness in test here.

antsukanova:
@SomeBody16 @edvujic added small final (hope so) comments, but overall I feel good with changes, thanks for the updates so far!

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3492,6 +3506,35 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  public beRightBack(enabled: boolean) {
+    return this.meetingRequest
+      .sendBrb({
+        enabled,
+        locusUrl: this.locusUrl,
+        deviceUrl: this.deviceUrl,
+        selfId: this.selfId,
+      })
+      .then(() => {
+        if (this.isMultistream && this.mediaProperties.webrtcMediaConnection) {
```
SomeBody16: This check should be at the start of beRightBack function.

antsukanova: @SomeBody16 updated function, please check it again

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3632,6 +3632,83 @@ describe('plugin-meetings', () => {
         });
       });
 
+      describe(`#beRightBack`, () => {
```
edvujic, line 3637: Hey @brycetham, I had a discussion with @antsukanova about the placement of these tests. The reason why these tests are placed here is because we are testing the public API. Previously, these tests were placed under the `[true, false].forEach((isMultistream) => describe('addMedia/updateMedia semi-integration tests ...`. However, these tests are specifically for adding and updating media, as well as publishing streams. This way, the tests are aligned with the other public APIs' tests.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3632,6 +3632,83 @@ describe('plugin-meetings', () => {
...
+        const fakeMultistreamRoapMediaConnection = {
+          createReceiveSlot: sinon.stub().resolves({on: sinon.stub()}),
+          createSendSlot: sinon.stub().returns({
+            setSourceStateOverride: sinon.stub().resolves(),
+            clearSourceStateOverride: sinon.stub().resolves(),
+          }),
+        };
+
+        beforeEach(() => {
+          meeting.meetingRequest.sendBrb = sinon.stub().resolves({body: 'test'});
```
edvujic, line 3646: I had a discussion with @antsukanova about the stubbing process, using `leaveMeeting` as a reference for testing. It appears that `locusMediaRequestStub` is used for actions related to adding/updating media and publishing streams. In contrast, `locusDeltaRequest` is for member updates. While we can verify member updates through `locusDeltaRequest`, using `sendBrb` seems like a better approach, as it aligns with how other tests were conducted for the functions in `request.ts`.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3632,6 +3632,82 @@ describe('plugin-meetings', () => {
...
 
+      describe(`#beRightBack`, () => {
+        const fakeMultistreamRoapMediaConnection = {
+          createSendSlot: sinon.stub().returns({
+            setSourceStateOverride: sinon.stub().resolves(),
+            clearSourceStateOverride: sinon.stub().resolves(),
+          }),
+        };
+
+        beforeEach(() => {
```
antsukanova, line 3645: I think would be good to add also clearing part in  `afterEach`
```
afterEach(() => {
          sinon.restore();
        });
```

antsukanova: Also thinking of doing code block 
```
  it('should have #beRightBack', () => {
            assert.exists(meeting.beRightBack);
          });
```

before _describe_ because it's common for two flows, LMKWYT

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3632,6 +3632,82 @@ describe('plugin-meetings', () => {
...
+          });
+
+          it('should have #beRightBack', () => {
+            assert.exists(meeting.beRightBack);
+          });
+
+          it('should ignore enabling #beRightBack', async () => {
+            const brbResult = meeting.beRightBack(true);
+
+            await brbResult;
```
antsukanova: I believe we do not need to await _**brbResult**_ here because it's not a promise for transcoded flow, but other logic lgtm

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3509,30 +3509,31 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        .catch((error) => {
+          LoggerProxy.logger.error('Meeting:index#beRightBack --> Error ', error);
 
-        return Promise.reject(error);
-      });
+          return Promise.reject(error);
+        });
+    }
+
+    return undefined;
```
brycetham: Do we want to return undefined here or do we want to return an empty promise? Maybe we can omit the return statement entirely? I'm a bit surprised we're allowed to return undefined here without TypeScript complaining, since we don't specify undefined can be returned in the function signature.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -739,46 +739,44 @@ describe('plugin-meetings', () => {
...
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state did not change', () => {
         const assertBrb = (enabled) => {
-          const selfWithBrbChanged = cloneDeep(self);
-          selfWithBrbChanged.controls.brb = enabled
 
-          locusInfo.self = selfWithBrbChanged
+          locusInfo.self = undefined;
           locusInfo.emitScoped = sinon.stub();
-          locusInfo.updateSelf(selfWithBrbChanged, [])
+          locusInfo.updateSelf(undefined, []);
```
brycetham: Is this a valid case? Will we ever actually call `locusInfo.updateSelf` with `undefined` when `locusInfo.self` is undefined?

Is it also possible for `locusInfo.self` to be defined and the `locusInfo.updateSelf` to be called with the `brb` state unchanged? Should we test that also?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3632,6 +3632,82 @@ describe('plugin-meetings', () => {
...
+            fakeMultistreamRoapMediaConnection,
+            MediaType.VideoMain
+          );
+
+          meeting.locusUrl = 'locus url';
+          meeting.deviceUrl = 'device url';
+          meeting.selfId = 'self id';
+        });
+
+        describe('when multistream meeting', () => {
```
brycetham: Nit: `when in a multistream meeting`

packages/@webex/plugin-meetings/src/meeting/index.ts
edvujic: I am more inclined to have an empty promise here, though @antsukanova has a point to have it as undefined as other functions have similar return statements.

SomeBody16: If we made this method `async`, no return statement will be required

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
antsukanova: I agree with the idea to test that brb is not changed. it's a good one/ for `undefined` @edvujic, could you please check if this flow is actually possible, and if not, then it would be good to update it to Bryce's suggestion with brb.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3532,8 +3533,7 @@ export default class Meeting extends StatelessWebexPlugin {
           return Promise.reject(error);
         });
     }
-
-    return undefined;
+    LoggerProxy.logger.error('Meeting:index#beRightBack --> Not a multistream meeting');
```
brycetham: This would trigger if there if `this.mediaProperties.webrtcMediaConnection` is not defined, so I think we need to tweak this error message a bit or maybe have 2 separate error messages.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -761,23 +761,22 @@ describe('plugin-meetings', () => {
...
-            { brb: enabled }
-          )
+            { brb: updatedBrb }
+          );
         };
 
-        assertBrb(true);
-        assertBrb(false);
-      })
+        assertBrbUnchanged(true, true); // BRB state unchanged and remains true
```
brycetham: If both params are going to be the same, we can just use a single one right?

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @brycetham, as discussed, let's then do throwing error way to handle transcoded flow. @edvujic, could you please support and change this part alongside with test cases pls? Appreciate that

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3513,10 +3513,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
     // this logic should be applied only to multistream meetings
+    if (!this.isMultistream) {
+      LoggerProxy.logger.error('Meeting:index#beRightBack --> Not a multistream meeting');
+    }
+    if (!this.mediaProperties.webrtcMediaConnection) {
+      LoggerProxy.logger.error(
+        'Meeting:index#beRightBack --> WebRTC media connection is not defined'
+      );
+
+      return;
```
brycetham: Why do we need to return here? Shouldn't `LoggerProxy.logger.error` throw and therefore we don't need a return?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3513,10 +3513,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+    }
     if (this.isMultistream && this.mediaProperties.webrtcMediaConnection) {
```
brycetham: I don't think we need this if statement anymore.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3513,10 +3513,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      // eslint-disable-next-line consistent-return
```
brycetham: Why is this needed?

packages/@webex/plugin-meetings/src/meeting/index.ts
edvujic: `LoggerProxy.logger.error` doesn't throw an error, it uses `console.error` to print the error but doesn't throw it.

edvujic: Double checked this, it seems like you're right, @brycetham. Thank you for pointing it out.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3513,20 +3513,16 @@ export default class Meeting extends StatelessWebexPlugin {
    * @returns {Promise<void>} resolves when the brb status is updated or does nothing if not in a multistream meeting.
    * @throws {Error} - Throws an error if the request fails.
    */
+  // eslint-disable-next-line consistent-return
```
edvujic: Seems like this is generated because ESLint is not recognising `return` statements within `if` statements.
Screenshot 2024-11-18 at 9 35 51 AM

brycetham: What if you remove the `else`? So something like:
```
public async beRightBack(enabled: boolean): Promise<void> {
    // this logic should be applied only to multistream meetings
    if (!this.isMultistream) {
      LoggerProxy.logger.error('Meeting:index#beRightBack --> Not a multistream meeting');
    } else if (!this.mediaProperties.webrtcMediaConnection) {
      LoggerProxy.logger.error(
        'Meeting:index#beRightBack --> WebRTC media connection is not defined'
      );
    }

    return this.meetingRequest
      .sendBrb({
        enabled,
        locusUrl: this.locusUrl,
        deviceUrl: this.deviceUrl,
        selfId: this.selfId,
      })
      .then(() => {
        this.sendSlotManager.setSourceStateOverride(MediaType.VideoMain, enabled ? 'away' : null);
      })
      .catch((error) => {
        LoggerProxy.logger.error('Meeting:index#beRightBack --> Error ', error);

        return Promise.reject(error);
      });
  }
```

edvujic: @brycetham, this would cause a source state to be always sent. I think that we should not be sending the source state change if the meeting is not multistream or webrtc media is not defined. Or I am wrong?

SomeBody16: If `LoggerProxy.logger.error` is throwing, then third else is not needed because if `!this.isMultistream` or `!this.mediaProperties.webrtcMediaConnection` is true, error will be thrown, and `.sendBrb` won't be called

edvujic: @SomeBody16, seems like that is not the case. Triple checking the `LoggerProxy.logger.error` it seems like it's doing a  `console.error` for the error message. I think we should throw an additional error/have a return statement/or keep the current logic.
Screenshot 2024-11-19 at 10 41 17 AM

Screenshot 2024-11-19 at 10 39 49 AM
Screenshot 2024-11-19 at 10 39 58 AM

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3505,6 +3519,41 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        .then(() => {
+          this.sendSlotManager.setSourceStateOverride(MediaType.VideoMain, enabled ? 'away' : null);
+        })
+        .catch((error) => {
+          LoggerProxy.logger.error('Meeting:index#beRightBack --> Error ', error);
+
+          return Promise.reject(error);
+        });
+    }
+  }
```
SomeBody16:
```suggestion
  public async beRightBack(enabled: boolean): Promise<void> {
    // this logic should be applied only to multistream meetings
    if (!this.isMultistream) {
      LoggerProxy.logger.error('Meeting:index#beRightBack --> Not a multistream meeting');
      return;
    } else if (!this.mediaProperties.webrtcMediaConnection) {
      LoggerProxy.logger.error(
        'Meeting:index#beRightBack --> WebRTC media connection is not defined'
      );
      return;
    }

    return this.meetingRequest
      .sendBrb({
        enabled,
        locusUrl: this.locusUrl,
        deviceUrl: this.deviceUrl,
        selfId: this.selfId,
      })
      .then(() => {
        this.sendSlotManager.setSourceStateOverride(MediaType.VideoMain, enabled ? 'away' : null);
      })
      .catch((error) => {
        LoggerProxy.logger.error('Meeting:index#beRightBack --> Error ', error);
        return Promise.reject(error);
      });
  }
```

Then i think something like this should fix it

antsukanova: @SomeBody16 thanks! I will update this part, align with other public methods

packages/@webex/plugin-meetings/src/locus-info/index.ts
antsukanova: as discussed will keep here only `brb` abbreviation

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3505,6 +3519,49 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      const error = new Error(errorMessage);
+      LoggerProxy.logger.error(error);
+
+      return Promise.reject(error);
+    }
+
+    if (!this.mediaProperties.webrtcMediaConnection) {
+      const errorMessage = 'Meeting:index#beRightBack --> WebRTC media connection is not defined';
+      const error = new Error(errorMessage);
+
```
brycetham, line 3542: Nit: some inconsistent spacing here.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -738,6 +738,75 @@ describe('plugin-meetings', () => {
...
+            LOCUSINFO.EVENTS.SELF_MEETING_BRB_CHANGED,
+            {brb: value}
+          );
+        };
+
+        assertBrbUnchanged(true);
+        assertBrbUnchanged(false);
+      });
+
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state changed to undefined', () => {
```
brycetham, line 790: When does this happen?

brycetham (approved):
Looks good. Left a super minor comment and a small question, but none of them are blocking are I've approved it.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
antsukanova: @brycetham all the time when DTO change happen, but when we do not use in API brb values
Screenshot 2024-11-22 at 10 40 18

SomeBody16 (approved):

PR #3974, opened 2024-11-14
feat(automotive): qr code login flow for automotive app 

xinhyao:
# COMPLETES #[SPARK-569607](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-569607)

Fetch QR code details and start polling the server to check whether the user has completed authorization.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

## Summary by CodeRabbit

- **New Features**
	- Introduced QR code login functionality with methods for obtaining login details and managing polling for authorization status.
	- Added the ability to cancel ongoing polling requests.

- **Bug Fixes**
	- Enhanced error handling for OAuth-related processes.

- **Tests**
	- Increased test coverage for new QR code login methods to ensure proper functionality and error handling, including scenarios for timeouts and specific error responses.


packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +242,128 @@ const Authorization = WebexPlugin.extend({
...
+        resource: '/actions/device/authorize',
+        form: {
+          client_id: this.config.client_id,
+          scope: this.config.scope,
+        },
+        auth: {
+          user: this.config.client_id,
+          pass: this.config.client_secret,
+          sendImmediately: true,
+        },
```
maxinteger, line 284: Since we just requesting the QR code details, I do not think we need `auth` parameter.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +242,128 @@ const Authorization = WebexPlugin.extend({
...
+              form: {
+                grant_type: 'urn:ietf:params:oauth:grant-type:device_code',
+                device_code: deviceCode,
+                client_id: this.config.client_id,
+              },
+              auth: {
+                user: this.config.client_id,
+                pass: this.config.client_secret,
+                sendImmediately: true,
+              },
```
maxinteger: I do not think you need the `auth` param for this request. Could you please check and link the docs if it is needed.

maxinteger (changes requested):
Over all it looks good, but I added few comments. Also, please take a look on the comments from `coderabbitai`.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
xinhyao: I found that if auth is not added, the error "Current state cannot produce an access token" will be reported when making a request. I tried to remove the auth field in other existing methods and found that this error will also be reported. So I added the auth parameter, although this is indeed a bit strange and probably not the right solution
image
image

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
xinhyao: Same as above

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
maxinteger: strange, but OK, we can keep the `auth` parameter

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+      });
+  },
+
+  /**
+   * Polling the server to check whether the user has completed authorization
+   * @instance
+   * @memberof AuthorizationBrowserFirstParty
+   * @param {Object} options
+   * @throws {Error} When the request fails
+   * @returns {Promise}
```
maxinteger: this method should not return with promise because we ignore it `initQRCodeLogin`

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+   */
+  startQRCodePolling(options = {}) {
+    if (!options.device_code) {
+      return Promise.reject(new Error('A deviceCode is required'));
```
maxinteger: issue: instead of reject, you should emit an error and then return.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+
+  /**
+   * Polling the server to check whether the user has completed authorization
+   * @instance
+   * @memberof AuthorizationBrowserFirstParty
+   * @param {Object} options
+   * @throws {Error} When the request fails
+   * @returns {Promise}
+   */
+  startQRCodePolling(options = {}) {
```
maxinteger: recommendation: since it is an internal method, it should start with underscore:

```suggestion
  _startQRCodePolling(options = {}) {
```

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+        },
+      })
+      .then((res) => {
+        const {user_code, verification_uri, verification_uri_complete} = res.body;
+        this.eventEmitter.emit('qrcode-login', {
+          eventType: 'user-code',
+          userData: {
+            user_code,
+            verification_uri,
+            verification_uri_complete,
```
maxinteger: issue: in the JS code we use camelCase instead of snakeCase, please re-map the values from the server response to camelCase

```suggestion
            userCode: user_code,
            verificationUri: verification_uri,
            verificationUriComplete: verification_uri_complete,
```

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+   * @throws {Error} When the request fails
+   * @returns {Promise}
+   */
+  startQRCodePolling(options = {}) {
+    if (!options.device_code) {
+      return Promise.reject(new Error('A deviceCode is required'));
+    }
+
+    if (this.pollingRequest) {
+      return Promise.reject(new Error('There is already a polling request'));
```
maxinteger: issue: same as the error above

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,147 @@ const Authorization = WebexPlugin.extend({
...
+    }
+
+    const {device_code: deviceCode, interval = 2, expires_in: expiresIn = 300} = options;
+
+    let attempts = 0;
+    const maxAttempts = expiresIn / interval;
+
+    return new Promise((resolve, reject) => {
```
maxinteger: again, we do not need this wrapper Promise

maxinteger (changes requested):

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,148 @@ const Authorization = WebexPlugin.extend({
...
+        },
+        auth: {
+          user: this.config.client_id,
+          pass: this.config.client_secret,
+          sendImmediately: true,
+        },
+      })
+      .then((res) => {
+        const {user_code, verification_uri, verification_uri_complete} = res.body;
+        this.eventEmitter.emit('qrcode-login', {
```
maxinteger: recommend: please use camelCase for the event's name. Same for all other cases.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,148 @@ const Authorization = WebexPlugin.extend({
...
+          eventType: 'user-code',
```
maxinteger: recommend: please use camelCase for `eventType`. Same for all other cases.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -240,6 +253,148 @@ const Authorization = WebexPlugin.extend({
       });
   },
 
+  /**
+   * Get an OAuth Login URL for QRCode. Generate QR code based on the returned URL.
+   * @instance
+   * @memberof AuthorizationBrowserFirstParty
+   * @emits #qrcode-login
+   */
+  initQRCodeLogin() {
```
maxinteger, line 262: since `_startQRCodePolling` is private and no point to get the QRCode details and then fail because there is a polling. 

Please add a check here for existing poll, `if (this.pollingRequest) {` and reject if there any.

maxinteger (changes requested):

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
xinhyao: done

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
xinhyao: done

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
xinhyao: done

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -443,6 +443,281 @@ describe('plugin-authorization-browser-first-party', () => {
...
+        
+        webex.authorization._startQRCodePolling(options);
+        clock.tick(4000);
+        await clock.runAllAsync()
+    
+        assert.calledTwice(webex.request);
+        assert.calledOnce(webex.authorization.cancelQRCodePolling);
+        assert.calledThrice(emitSpy);
+        assert.equal(emitSpy.getCall(0).args[1].eventType, 'authorizationSuccess');
+        assert.equal(emitSpy.getCall(1).args[1].eventType, 'pollingCanceled');
```
maxinteger, line 612: issue: after the poll canceled the consumer still get events from the plugin.

Like in this case, you get the `pollingCanceled` event, but then you got a `authorizationPending` event as well. It happens because cancellation stop only the timer, but does not abort the fetch.

Aborting is tricky with webex.request, so better to not emit any event after polling cancelled.

For example:
```js
.then((res) => {
  if (this.pollingRequest === null) return;

  this.eventEmitter.emit('qRCodeLogin', {
    eventType: 'authorizationSuccess',
    data: res.body,
  });
  this.cancelQRCodePolling();
})
```

Same for the catch part.

It is still not perfect because user can re-start the polling and it might happens before the last fetch from the previous poll finished.
So again we either abort the fetch (somehow) or track the cancelled fetch(es) and do not emit event.

maxinteger (changes requested):

maxinteger (approved):

mickelr (approved):

PR #3981, opened 2024-11-18
feat(wxcc): added-relogin-logic

adhmenon:
# COMPLETES #< [SPARK-580609](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-580609)>

* Currently, we are simply firing a re-login request after 15 seconds of logging in, which is incorrect.
* Code has been added to properly implement re-login logic (when network goes and when browser is reloaded).
* There has also been functionality added to take care of the edge case wherein a user is logged in, available state and then proceeds to re-login, we need to reset status to available in that case.

* Added a re-registration every time we reconnect with the websocket
* When we receive a `AgentReLoginFail`, we silently handle the error
* When we get a success, we check the event and if it is `wss-agent-disconnect`, we need to reset the state to `AVAILABLE` again.
* Added an event bus to manage all events (added it as a singleton)

### Vidcast of Switching off internet
https://app.vidcast.io/share/f12acad3-5d34-4b34-9f5c-05501f555553

### Vidcast of Re-Login State Change
https://app.vidcast.io/share/519354fc-24d1-4404-b27f-3225676fe4f3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


docs/samples/contact-center/app.js
```diff
@@ -174,16 +174,6 @@ function doAgentLogin() {
   webex.cc.stationLogin({teamId: teamsDropdown.value, loginOption: agentDeviceType, dialNumber: dialNumber.value}).then((response) => {
     console.log('Agent Logged in successfully', response);
     logoutAgentElm.classList.remove('hidden');
-    // Re-Login Agent after 15 seconds for testing purpose
```
adhmenon, line 177: We no longer require this now that we have the logic working.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -25,7 +25,8 @@ import Services from './services';
 import LoggerProxy from './logger-proxy';
 import {StateChange, Logout} from './services/agent/types';
 import {ConnectionService} from './services/core/WebSocket/connection-service';
-import {getErrorDetails} from './services/core/Utils';
+import {getErrorDetails, getErrorReason} from './services/core/Utils';
+import {EventBus} from './services/core/EventBus';
```
adhmenon, line 29: Have added an `EventBus` in order to have an easy way to manage events. Will explain this in further detail.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -57,15 +59,18 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       this.webSocketManager = new WebSocketManager({webex: this.$webex});
 
-      this.connectionService = new ConnectionService(
-        this.webSocketManager,
-        this.getConnectionConfig()
-      );
+      this.connectionService = new ConnectionService({
+        webSocketManager: this.webSocketManager,
+        onReRegister: this.register.bind(this),
```
adhmenon, line 64: Passing reRegister as a callback to the connection-service, felt this would be the best way so that we do not have any circular dependencies. Open to any other suggestions however.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -118,6 +123,9 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
           const agentConfig = new AgentConfig(agentId, this.$webex, this.httpRequest);
           this.agentConfig = await agentConfig.getAgentProfile();
           this.$webex.logger.log(`file: ${CC_FILE}: agent config is fetched successfully`);
+          // TODO: We need to emit an event when it is successful
+          /* eslint-disable consistent-return */
+          await this.defaultReLogin();
```
adhmenon, line 128: **Note**

This needs to be handled, waiting for Ravi's PR as I think that is required for us to handle this event.
Although, I'm not a 100% sure. But, open to any suggestions on this - might need a discussion.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +253,35 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       allowMultiLogin: this.$config?.allowMultiLogin ?? true,
     };
   }
+
+  /**
+   * Called when we finish registration to silently handle the errors
+   */
+  private async defaultReLogin(): Promise<StationReLoginResponse | void> {
```
adhmenon, line 260: Here, we silently handle the error (usually because agent is not logged in) - this is being done in agentx also, ie they always call relogin after registration.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +253,35 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        return;
+      }
+      throw getErrorDetails(error, 'defaultReLogin');
+    }
+  }
+
+  /**
+   * Handle the custom event and call setAgentState for disconnect event
+   */
+  private async handleAgentWssDisconnect(msg: StateChange) {
```
adhmenon, line 278: For handling the logic to set agent status to `AVAILABLE` again (edge case). Tested and it was working (in vidcast 1)

packages/@webex/plugin-cc/src/services/core/EventBus.ts
```diff
@@ -0,0 +1,25 @@
+import {EventEmitter} from 'events';
```
adhmenon, line 1: Here we will have our class which is an EventBus. This will basically register all the required events, and also the various handlers. When we emit events from the code, it will then be re-routed to wherver it is needed to go.
Note that I have included methods for cleanup and am cleaning it up in the tests as well.
We use a singleton setup here (static instance), and ideally there should be no memory leaks.

Open to any better way to do this (as otherwise it is difficult to route events from aqm-reqs to cc and websocket also was being injected everywhere).

packages/@webex/plugin-cc/src/services/core/Utils.ts
```diff
@@ -10,6 +10,12 @@ const getCommonErrorDetails = (errObj: WebexRequestPayload) => {
   };
 };
 
+export const getErrorReason = (error: any): string => {
```
adhmenon: Decided to write a util method, it can be reused also.

packages/@webex/plugin-cc/src/services/core/WebSocket/WebSocketManager.ts
```diff
@@ -4,8 +4,9 @@ import {SubscribeResponse} from '../../config/types';
 import LoggerProxy from '../../../logger-proxy';
 import workerScript from './keepalive.worker';
 import {KEEPALIVE_WORKER_INTERVAL, CLOSE_SOCKET_TIMEOUT} from '../constants';
+import {EventBus} from '../EventBus';
 
-export class WebSocketManager extends EventTarget {
+export class WebSocketManager {
```
adhmenon, line 9: Now that we use the EventBus, we no longer need to extend the EventTarget, we can just emit via the bus instance.

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
```diff
@@ -32,20 +22,22 @@ export class ConnectionService extends EventTarget {
...
+  private eventBus: EventBus;
+  private onReRegister: () => Promise<void>;
 
-  constructor(webSocketManager: WebSocketManager, subscribeRequest: SubscribeRequest) {
+  constructor(options: ConnectionServiceOptions) {
     super();
-    this.webSocketManager = webSocketManager;
-    this.subscribeRequest = subscribeRequest;
+    this.webSocketManager = options.webSocketManager;
+    this.onReRegister = options.onReRegister;
```
adhmenon, line 31: We register the callback here, it will be called when we reconnect to the socket again.

packages/@webex/plugin-cc/src/services/core/WebSocket/types.ts
```diff
@@ -0,0 +1,17 @@
+import {WebSocketManager} from './WebSocketManager';
+
+export type ConnectionServiceOptions = {
```
adhmenon, line 4: Options can be extended.

packages/@webex/plugin-cc/src/services/core/aqm-reqs.ts
```diff
@@ -1,22 +1,24 @@
...
-  private webSocketManager: WebSocketManager;
   private httpRequest: HttpRequest;
+  private eventBus: EventBus;
 
-  constructor(webSocketManager: WebSocketManager) {
+  constructor() {
     this.httpRequest = HttpRequest.getInstance();
-    this.webSocketManager = webSocketManager;
-    this.webSocketManager.addEventListener('message', this.onMessage);
+    this.eventBus = EventBus.getInstance();
```
adhmenon, line 20: Same logic here, no need to DI the web socket manager, it can just do the web socket related tasks. We can use the bus to manage and route the events.

packages/@webex/plugin-cc/src/services/core/aqm-reqs.ts
```diff
@@ -216,6 +218,22 @@ export default class AqmReqs {
       return;
     }
 
+    if (
```
adhmenon, line 221: The logic to handle the edge case. It does not interfere with any event.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -86,6 +89,7 @@ describe('webex.cc', () => {
 
   afterEach(() => {
     jest.clearAllMocks();
+    eventBusInstance.removeAllListeners();
```
adhmenon, line 92: In order for any remaining event listeners to be removed (otherwise memory used is intensive).

packages/@webex/plugin-cc/test/unit/spec/services/core/aqm-reqs.ts
```diff
@@ -141,21 +141,20 @@ describe('AqmReqs', () => {
...
-              webSocketManagerInstance.dispatchEvent(
-                new CustomEvent('message', {
-                  detail: JSON.stringify({
-                    type: 'RoutingMessage',
-                    data: {
-                      type: 'AgentCtqCancelled',
-                      interactionId: '6920dda3-337a-48b1-b82d-2333392f9905',
-                    },
-                  }),
+              eventBusInstance.emit(
```
adhmenon, line 150: Had to change the way we mock the event, but other than that most test cases are covered)

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -57,15 +59,16 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       this.webSocketManager = new WebSocketManager({webex: this.$webex});
 
-      this.connectionService = new ConnectionService(
-        this.webSocketManager,
-        this.getConnectionConfig()
-      );
+      this.connectionService = new ConnectionService({
```
adhmenon: Sticking to the old way of doing it, but with an options.
We need subscribeRequestConfig so that we can re-initialise the websocket connection rather than once again calling the cc.connectWebSocket method here, we directly register it and connect (same as what agent desktop does).

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -118,6 +121,8 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
           const agentConfig = new AgentConfig(agentId, this.$webex, this.httpRequest);
           this.agentConfig = await agentConfig.getAgentProfile();
           this.$webex.logger.log(`file: ${CC_FILE}: agent config is fetched successfully`);
+          // TODO: Check for config if it supports this auto silent relogin (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574666)
+          await this.silentRelogin();
```
adhmenon: Again, here we will by default call `silentRelogin` as discussed, however once the config PR is ready, we will add a check here (so users can decide if they want to call it or not).

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +257,54 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       allowMultiLogin: this.$config?.allowMultiLogin ?? true,
     };
   }
+
+  /**
+   * Called when we reconnection has been completed
+   */
+  private async handleConnectionLost(msg: ConnectionLostDetails): Promise<void> {
```
adhmenon, line 261: This is an event handler when we lose connection in the web socket. This gets called and it internally will handle things such as overall event emission to the user that the socket has been disconnected or has been reconnected.
Here also we will call `silentRelogin` if the config allows for it.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +257,54 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      );
+      // TODO: Add a check here for config (if it has property to auto silent relogin) (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574666)
+      await this.silentRelogin();
+    }
+  }
+
+  /**
+   * Called when we finish registration to silently handle the errors
+   */
+  private async silentRelogin(): Promise<void> {
```
adhmenon, line 279: Over here, we simply reload the agent and get the required data.
If the agent was in available state and it got disconnected, we will change the state (same logic as in agent desktop).
Again, once all this is done, we need to update the config variable (maybe `isLoggedIn` or something) ... again this will be TODO as it is waiting for the other PR.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +257,54 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+          state: AGENT_STATE_AVAILABLE,
+          auxCodeId,
+          lastStateChangeReason,
+          agentId,
+        };
+        await this.setAgentState(stateChangeData);
+      }
+      // TODO: Update login state of config here (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574666)
+    } catch (error) {
+      if (getErrorReason(error) === 'AGENT_NOT_FOUND') {
```
adhmenon, line 302: Again, silently handle this fialure (mostly will occur on fresh reload if the user is not logged in).

packages/@webex/plugin-cc/src/services/core/WebSocket/WebSocketManager.ts
```diff
@@ -59,6 +61,10 @@ export class WebSocketManager extends EventTarget {
     }
   }
 
+  handleConnectionLost(event: ConnectionLostDetails) {
+    this.isConnectionLost = event.isConnectionLost;
```
adhmenon, line 65: Had forgotten to include this, it is needed here. This gets called when the connection is lost (event will be sent to cc which will call this).
Agent desktop was using a public signal variable, we have used events.

packages/@webex/plugin-cc/src/services/core/WebSocket/WebSocketManager.ts
```diff
@@ -150,7 +156,7 @@ export class WebSocketManager extends EventTarget {
     this.isSocketClosed = true;
     this.keepaliveWorker.postMessage({type: 'terminate'});
     if (this.shouldReconnect) {
-      this.dispatchEvent(new Event('socketClose'));
+      this.emit('socketClose');
```
adhmenon, line 159: As per Arun's advice, stuck to using event emitter and dispatching the events.
Added this all over and updated the event handling as well.

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
```diff
@@ -34,31 +25,33 @@ export class ConnectionService extends EventTarget {
...
-    });
-    this.dispatchEvent(event);
+    const event: ConnectionLostDetails = {
+      isConnectionLost: this.isConnectionLost,
+      isRestoreFailed: this.isRestoreFailed,
+      isSocketReconnected:
+        !this.webSocketManager.isSocketClosed && (socketReconnected || this.isSocketReconnected),
+      isKeepAlive: this.isKeepAlive,
+    };
+    this.emit('connectionLost', event);
```
adhmenon, line 55: Same over here (updated the emitting).

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
```diff
@@ -89,29 +82,24 @@ export class ConnectionService extends EventTarget {
...
     }
     this.isKeepAlive = parsedEvent.keepalive === 'true';
-    const shouldUpdateConnectionData =
-      this.isKeepAlive || (this.isConnectionLost && !this.isRestoreFailed);
-    const shouldDispatchEvent =
-      this.isKeepAlive || (this.isConnectionLost && !this.isRestoreFailed);
-    const shouldDispatchEventWithReconnect = this.isSocketReconnected && this.isKeepAlive;
 
-    if (shouldUpdateConnectionData) {
+    if (
```
adhmenon, line 96: This old logic from agent desktop was working (my logic was not), hence re-added it here.

packages/@webex/plugin-cc/src/services/core/aqm-reqs.ts
```diff
@@ -216,6 +216,12 @@ export default class AqmReqs {
       return;
     }
 
+    if (event.type === 'AgentReloginFailed') {
```
adhmenon, line 219: So that we get no logs for lack of handles.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -172,6 +181,12 @@ describe('webex.cc', () => {
         wrapUpCodes: [],
       };
       const connectWebsocketSpy = jest.spyOn(webex.cc, 'connectWebsocket');
+      const reloadSpy = jest.spyOn(webex.cc.services.agent, 'reload').mockResolvedValue({
```
adhmenon, line 267: Need to mock all this so that the tests pass.

packages/@webex/plugin-cc/test/unit/spec/services/core/WebSocket/connection-service.ts
```diff
@@ -113,15 +112,30 @@ describe('ConnectionService', () => {
     expect(mockWebSocketManager.initWebSocket).toHaveBeenCalledWith({body: mockSubscribeRequest});
   });
 
-  it('should handle ping message without keepalive and not update connection data', () => {
-    const pingMessage = new CustomEvent<string>('message', {
-      detail: JSON.stringify({ someOtherProperty: 'value' }),
+  describe('ConnectionService onPing', () => {
```
adhmenon, line 115: Added more tests for improved coverage.

packages/@webex/plugin-cc/test/unit/spec/services/core/aqm-reqs.ts
```diff
@@ -66,12 +65,22 @@ describe('AqmReqs', () => {
       webex: {} as any,
     }) as jest.Mocked<WebSocketManager>;
 
-    // Mock the addEventListener method
-    webSocketManagerInstance.addEventListener = jest.fn((event, callback) => {
-      if (event === 'message') {
-        webSocketManagerInstance.dispatchEvent = callback;
+    // Mock the on method to handle event listeners
+    const eventListeners: { [key: string]: Function[] } = {};
```
adhmenon, line 69: Had to add this as we need to mock the EventEmitter framework somehow.
Open to suggestions on better ways, if there are.

packages/@webex/plugin-cc/src/cc.ts
rsarika: this can be done now itself. pass the config as part of webexconfig

packages/@webex/plugin-cc/src/services/core/Utils.ts
rsarika: we have this already right? getErrorDetails can enhance or reuse the same

packages/@webex/plugin-cc/src/cc.ts
rsarika: is there a possibility to combine them and move it to services layer completely.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -55,14 +53,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
 
-      this.webSocketManager = new WebSocketManager({webex: this.$webex});
-
-      this.connectionService = new ConnectionService(
-        this.webSocketManager,
-        this.getConnectionConfig()
-      );
-
-      this.services = Services.getInstance(this.webSocketManager);
+      this.services = Services.getInstance(this.$webex, this.getConnectionConfig());
```
rsarika: This is good

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
```diff
@@ -89,29 +83,24 @@ export class ConnectionService extends EventTarget {
     this.connectionProp = prop;
   }
 
-  private onPing = (event: Event): void => {
-    const msg = (event as CustomEvent<string>).detail;
-    const parsedEvent = JSON.parse(msg);
+  private onPing = (event: any): void => {
```
rsarika, line 86: is there a reason we changed this to any?

packages/@webex/plugin-cc/src/services/index.ts
```diff
@@ -1,19 +1,28 @@
...
+    const aqmReq = new AqmReqs(this.webSocketManager);
     this.agent = routingAgent(aqmReq);
+    this.connectionService = new ConnectionService({
+      webSocketManager: this.webSocketManager,
+      subscribeRequest,
+    });
   }
 
-  public static getInstance(webSocketManager: WebSocketManager): Services {
+  public static getInstance(webex: WebexSDK, subscribeRequest: SubscribeRequest): Services {
```
rsarika: use options and deconstruct them.

rsarika (approved):

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
adhmenon: Oh I think it it was a mistake (leftover from some old code), I'll change it back.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -245,4 +253,54 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        );
+        const stateChangeData: StateChange = {
+          state: AGENT_STATE_AVAILABLE,
+          auxCodeId,
+          lastStateChangeReason,
+          agentId,
+        };
+        await this.setAgentState(stateChangeData);
+      }
+      // TODO: Update login state of config here (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-574666)
```
mkesavan13: Looks like this was Ravi's ticket. Should we consider removing this TODO already? His PR is merged

packages/@webex/plugin-cc/src/services/core/Utils.ts
```diff
@@ -12,9 +12,16 @@ const getCommonErrorDetails = (errObj: WebexRequestPayload) => {
 
 export const getErrorDetails = (error: any, methodName: string) => {
   const failure = error.details as Failure;
-  LoggerProxy.logger.error(`${methodName} failed with trackingId: ${failure?.trackingId}`);
+  const reason = failure?.data?.reason ?? `Error while performing ${methodName}`;
+  // We don't want to log silentReLogin errors
+  if (methodName !== 'silentReLogin') {
```
mkesavan13, line 17: I agree it is `silentRelogin` so this is in the background. But am wondering if we should silent the logs as well. We will have no clue of the origin of error if we want to debug an issue with the logs

mkesavan13 (approved):
Just one comment. The other in the review is outdated already.

packages/@webex/plugin-cc/src/services/core/Utils.ts
adhmenon: You're correct, I think it will be good if I add another check which also checks the type - ie AGENT_NOT_FOUND and silentRelogin, this way we know it will be only the case we want silenced.
I have the logger PR Kesava, I will fix it in that :)

adhmenon: You're correct, I think it will be good if I add another check which also checks the type - ie AGENT_NOT_FOUND and silentRelogin, this way we know it will be only the case we want silenced.

PR #3797, opened 2024-08-28
fix(audio-mute): fix sometimes out-of-sync remote audio mute

chburket:
# COMPLETES

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555818

## This pull request addresses

Fixes a BEMS case defect where remote audio mute can get out of sync with local audio mute.

BEMS space: webexteams://im?space=c4706ce0-56e0-11ef-81df-a39e6dd40a42

Steps to recreate bug:

1. start a meeting as host, unmuted
2. uncheck "allow participants to unmute"
3. mute 
4. unmute
5. mute (problem: sdk's representation of server remote audio is still unmuted)
6. unmute (because sdk thinks remote is already unmuted, we do not try to change it)

The server remote audio is still muted (notice the muted icon in the participant list)
With the fix, the server remote audio is unmuted as expected.

## by making the following changes

First, I needed to revert a [fix from months ago](https://github.com/webex/webex-js-sdk/pull/3621). This fix was causing the locus diff update from step 5 to be skipped because the modifiedBy value was the same as the local user.  We did *not* skip the update in step 3 because modifiedBy was undefined.  I met with Rajesh Kadikar about this and he indicated that the modifiedBy should contain the current user in this scenario, and that we should not be skipping locus updates, even it they were modified by the current user.

Second, I needed to find a new way to fix what was reverted. I added console.logs in various places in the flow and found that if you unmute and mute quickly, `_LocalMicrophoneStream.setUserMuted(true)` happens before we receive the locus diff from the unmute.  When the locus diff eventually comes in, `handleServerRemoteMuteUpdate()` unmutes the local stream to match the remote mute state, which should not be done.  So I added an `if` around the local stream update to only execute it when muting, not unmuting.

While testing this PR, I found a few problems around host mute/unmute (unrelated to this change) and opened these Jiras:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-559101
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-559095

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

Manually tested the scenario in the BEMS Jira (SPARK-555818).
[...]

PR #3621, opened 2024-05-29
fix(plugin-meetings): Mute issue with host and "mute all" action

antsukanova:
# COMPLETES # [WEBEX-382445](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-382445)

## This pull request addresses

The issue arises from `mutedByOthersChanged` in `selfUtils`. This function should not trigger a user mute update when the user receives a notification from the server that the server has been updated remotely to a muted state but not from another person. 

## by making the following changes

I have added the fix to `mutedByOthersChanged` that will avoid updating the mute state when the server state doesn't change by others and added testcases for `selfUtils.mutedByOthersChanged`

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

- [x] automated tests for plugin-meetings
- [x] manually tested with client app

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request
- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #3995, opened 2024-11-20
fix(plugin-meetings): moderated unmute when client is remotely muted

marcin-bazyl:
This PR reverts https://github.com/webex/webex-js-sdk/pull/3797 and improves the fix from https://github.com/webex/webex-js-sdk/pull/3621 by having more specific condition to decide when to ignore the Locus update

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
@@ -428,6 +428,11 @@ SelfUtils.mutedByOthersChanged = (oldSelf, changedSelf) => {
     return false;
   }
 
+  // there is no need to trigger user update if no one muted user
+  if (changedSelf.selfIdentity === changedSelf.modifiedBy) {
+    return false;
+  }
+
   return (
     changedSelf.remoteMuted !== null &&
     (oldSelf.remoteMuted !== changedSelf.remoteMuted ||
--- a/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
@@ -419,25 +419,27 @@ SelfUtils.mutedByOthersChanged = (oldSelf, changedSelf) => {
     throw new ParameterError('New self must be defined to determine if self was muted by others.');
   }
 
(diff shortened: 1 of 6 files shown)


packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -359,6 +370,23 @@ export class MuteState {
     }
   }
 
+  public shouldIgnoreRemoteMuteUpdate(remoteMute: boolean, isModifiedBySelf: boolean) {
+    console.log(
+      `marcin: shouldIgnoreRemoteMuteUpdate: flag=${this.state.isRemoteUnmutePendingLocusDtoUpdate} remoteMute=${remoteMute}, isModifiedBySelf=${isModifiedBySelf}`
+    );
+    if (this.state.isRemoteUnmutePendingLocusDtoUpdate && isModifiedBySelf && !remoteMute) {
```
marcin-bazyl, line 377: We only ever send remote unmute requests (not mute), so here I'm only checking for `!remoteMute`, but the method that sends out these requests (`sendRemoteMuteRequestToServer()`) is written in a way that suggests that it could do either mute or unmute, so I think I'll rename `sendRemoteMuteRequestToServer()` to `sendRemoteUnmuteRequestToServer()` and hardcode `remoteMute = false` in it to make it more clear that we only do remote unmute, then this will be more in-line with the hardcoded assumption here in `shouldIgnoreRemoteMuteUpdate()`

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
```diff
@@ -419,20 +419,27 @@ SelfUtils.mutedByOthersChanged = (oldSelf, changedSelf) => {
     throw new ParameterError('New self must be defined to determine if self was muted by others.');
   }
 
+  const isModifiedBySelf = changedSelf.selfIdentity === changedSelf.modifiedBy;
```
marcin-bazyl, line 422: I couldn't do the full check if we should ignore or not here, because selfUtils.ts doesn't have a reference to the meeting or MuteState class, so I've added the field `isModifiedBySelf` that's emitted with the `SELF_REMOTE_MUTE_STATUS_UPDATED` event and I've put all the logic to decide if we ignore the Locus DTO update or not in `MuteState.shouldIgnoreRemoteMuteUpdate()`

PR #3998, opened 2024-11-21
feat(internal-device-plugin): added deleteDevices

shnaaz:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-584363

Added a method deleteDevices which is called when register fails.
This is done to tackle failures where stale device registrations don't get deleted. This is possible when user just closes the browser instead of leaving a meeting.
Added logic where register is tried once more after failure.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method for automatic deletion of devices when the number exceeds a specified limit.
	- Enhanced error handling during device registration to manage device lifecycle more effectively.

- **Bug Fixes**
	- Improved registration process to handle errors and manage device cleanup when registration fails.

- **Tests**
	- Added test cases to validate the new device deletion functionality and error handling in the registration process.


packages/@webex/internal-plugin-device/src/device.js
```diff
@@ -439,7 +440,62 @@ const Device = WebexPlugin.extend({
...
+  /**
+   * Registers and when fails deletes devices
+   */
+  @oneFlight
+  @waitForValue('@')
+  register(deviceRegistrationOptions = {}) {
+    return this._registerInternal(deviceRegistrationOptions).catch((error) => {
+      if (!deviceRegistrationOptions.deleteFlag) {
+        return this.deleteDevices().then(() => {
+          return this._registerInternal({deleteFlag: true});
```
Coread: issue: we don't need the delete flag anymore

Coread: issue: we should be passing in the original options

packages/@webex/internal-plugin-device/src/device.js
```diff
@@ -439,7 +440,62 @@ const Device = WebexPlugin.extend({
...
+  },
 
+  /**
+   * Registers and when fails deletes devices
+   */
+  @oneFlight
+  @waitForValue('@')
+  register(deviceRegistrationOptions = {}) {
+    return this._registerInternal(deviceRegistrationOptions).catch((error) => {
+      if (!deviceRegistrationOptions.deleteFlag) {
```
Coread: issue: instead of checking the delete flag, we should be checking for the error message of excessive device registrations

packages/@webex/internal-plugin-device/src/device.js
```diff
@@ -439,7 +440,62 @@ const Device = WebexPlugin.extend({
...
+          const countToDelete = Math.ceil(totalItems / 3);
+          const urlsToDelete = sortedDevices.slice(0, countToDelete).map((item) => item.url);
+
+          // Send DELETE requests for each URL
+          urlsToDelete.forEach((url) => {
+            this.request({
+              uri: url,
+              method: 'DELETE',
+            }).then(() => {
+              this.clear();
```
marcin-bazyl: we are calling `this.clear()` as many times as many devices we delete. I see that this is even verified in unit tests, but I don't understand why - shouldn't we be calling it only once?

packages/@webex/internal-plugin-device/test/unit/spec/device.js
```diff
@@ -357,6 +357,97 @@ describe('plugin-device', () => {
...
+
+        const requestStub = sinon.stub(device, 'request');
+        requestStub.withArgs(sinon.match({method: 'GET'})).resolves(response);
+        requestStub.withArgs(sinon.match({method: 'DELETE'})).resolves();
+
+        const clearSpy = sinon.spy(device, 'clear');
+        await device.deleteDevices();
+
+        // Calculate expected deletions
+        const expectedDeletions = ['url1'];
```
marcin-bazyl: even though the tests are written in a way that can check that multiple devices are deleted, in all of them the `expectedDeletions` has only 1 entry, so they're not really checking that multiple devices get deleted. It would be nice to increase the number of mocked devices so that we really test that more than 1 is deleted.

Coread (approved):

jor-row (approved):

PR #4005, opened 2024-11-26
feat(plugin-cc): Spark 562169 task accept api implementation

Kesari3008:
# COMPLETES #[SPARK-562169](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-562169) 

Call Accept Flow Implementation in CC SDK

Created a new file contact.ts and moved the contact API code from the agent desktop to CC SDK. This file contains the API invocation a handling of success and failure events as per the agent desktop. This layer is present behind the services layer.
Create a new class called TaksCControl, a public module to access task-specific public methods. Implemented the call accept flow for the browser login option there.
Updated sample app to test the call accept flow.
Added UT for the changes done

Please check the new PR: https://github.com/webex/webex-js-sdk/pull/4016 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Scenario 1: Agent login with Browser option. Initiated an incoming call and answered it from the sample app.
Scenario 2: Agent login with Browser option. Initiated an incoming call and declined it from the sample app.
Scenario 3: Agent login with an Extension option. Initiated an incoming call, answer and decline buttons are disabled as the call needs to be answered from an actual device.

https://app.vidcast.io/share/358a656c-ca26-4d69-8aee-526b33345fe3 

### I certified that

---

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Contact Center module in the Webex SDK, enhancing functionality for managing agent interactions.
	- Added a new link to the "Contact Center" sample in the documentation for improved navigation.
	- Implemented a WebSocket manager to handle connections and events more robustly.
	- Added comprehensive agent management features, including login, logout, and status updates.

- **Bug Fixes**
	- Improved error handling and logging for various operations within the Contact Center functionalities.

- **Documentation**
	- Updated README and various documentation files to reflect new features and usage instructions.

- **Tests**
	- Expanded unit tests for the Contact Center and related services to ensure functionality and error handling.


sreenara (changes requested):
Please update the PR description with details of the new code added such as a new class called Task. Additionally, it would be good if you could talk about what the contact.ts is and what changes if any were made to the agentx-services code while moving the file here.

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -1,5 +1,6 @@
 import {CallingClientConfig} from '@webex/calling';
 import * as Agent from './services/agent/types';
+import * as Contact from './services/Task/types';
```
sreenara: Could we stick to the convention of naming the folder in all small letters?
```suggestion
import * as Contact from './services/task/types';
```

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -217,7 +218,14 @@ export type RequestBody =
   | Agent.Logout
   | Agent.UserStationLogin
   | Agent.StateChange
-  | Agent.BuddyAgents;
+  | Agent.BuddyAgents
+  | Contact.HoldResumePayload
```
sreenara, line 222: Is there a reason why we are importing as Contact? What's the significance of that term?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
+import EventEmitter from 'events';
+import {ICall, LINE_EVENTS, LocalMicrophoneStream, createMicrophoneStream} from '@webex/calling';
+import Services from '..';
+import {getErrorDetails} from '../core/Utils';
+import WebCallingService from '../WebCallingService';
+import {LoginOption} from '../../types';
+import {CC_FILE} from '../../constants';
+import {CC_EVENTS} from '../config/types';
+
+export default class Task extends EventEmitter {
```
sreenara: question. If we extend EventEmitter, what object will the application listen for the event on?
What's the difference between doing this vs const emitter = new EventEmitter(); within this class?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  private localAudioStream: LocalMicrophoneStream;
+  private services: Services; // This will be used later for invoking Contact APIs
```
sreenara: Does the Task need the entire Services object? Doesn't services have unnecessary members which don't pertain to the Task?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(services: Services, webCallingService: WebCallingService) {
+    super();
+    this.services = services;
+    this.webCallingService = webCallingService;
+    this.registerTaskListeners();
+    this.registerCallingEvent();
+  }
+
+  private registerCallingEvent() {
```
sreenara:
```suggestion
  private registerIncomingCallEvent() {
```

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.webCallingService.on(LINE_EVENTS.INCOMING_CALL, (call) => {
+      if (this.taskPayload) {
+        this.emit('task:incoming', this.taskPayload);
```
sreenara: Perhaps we could have an enum for task:incoming and other events too?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+import WebCallingService from '../WebCallingService';
+import {LoginOption} from '../../types';
+import {CC_FILE} from '../../constants';
+import {CC_EVENTS} from '../config/types';
+
+export default class Task extends EventEmitter {
+  private localAudioStream: LocalMicrophoneStream;
+  private services: Services; // This will be used later for invoking Contact APIs
+  private webCallingService: WebCallingService;
+  private taskPayload: any;
```
sreenara: Is there no way to avoid this any?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  private registerTaskListeners() {
+    // TODO: This event reception approach will be changed once state machine is implemented
+    this.services.webSocketManager.on('message', (event) => {
+      const payload = JSON.parse(event);
+      switch (payload.data.type) {
+        case CC_EVENTS.AGENT_CONTACT_RESERVED:
+          this.taskPayload = payload.data;
+          if (this.webCallingService.loginOption !== LoginOption.BROWSER) {
+            this.emit('task:incoming', payload.data);
```
sreenara: Wouldn't this be confusing for a developer? We're emitting the same event with different payloads based on the type of login? Would it make sense to have different event names?

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+          break;
+      }
+    });
+  }
+
+  /**
+   * This is used for incoming task accept by agent.
+   * @param data
+   * @returns Promise<TaskAcceptResponse>
+   * @throws Error
```
sreenara: Please include an example. Applies to all other public functions

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public async accept(taskId: string): Promise<void> {
+    try {
+      if (this.webCallingService.loginOption === LoginOption.BROWSER) {
+        // @ts-ignore
```
sreenara: We shouldn't have this ts-ignore here. This is similar to the issue reported B&S. Perhaps we should use the other approach to create the microphone stream.

packages/@webex/plugin-cc/src/services/Task/types.ts
```diff
@@ -0,0 +1,288 @@
+import {Msg} from '../core/GlobalTypes';
+
+export enum DESTINATION_TYPE {
```
sreenara: Could we use the new way of declaring enums?

packages/@webex/plugin-cc/src/services/Task/types.ts
```diff
@@ -0,0 +1,288 @@
...
+  | 'email'
+  | 'chat'
+  | 'telephony'
+  | 'social'
+  | 'sms'
+  | 'facebook'
+  | 'whatsapp'
+  | string;
+
+export type AgentContact = Msg<{
```
sreenara: We should have TSDoc style descriptions for each of the types. This applies in all types files modified in this PR.

packages/@webex/plugin-cc/src/services/WebCallingService.ts
sreenara, line 1: Is this the file that acts as the wrapper for everything related to the Calling SDK? If so, can we have a new ticket created to investigate how we can ship a contactcenter.min.js without this file and all references to @webex/calling? Do you think that would make it easier for devs to consume the plugin-cc they need?

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: Perhaps we could only pass the members of services that are needed?

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara, line 1: The way this class has been structured makes me feel like this is a bit of a task manager but also a task.
Also shouldn't we have types of Tasks? We won't only be doing calls in the future right?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -57,6 +59,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       });
 
       this.webCallingService = new WebCallingService(this.$webex, this.$config.callingClientConfig);
+      this.task = new Task(this.services, this.webCallingService);
```
sreenara: Why are we initializing a task here.. at the start, there won't be any tasks right?

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: Also, how do we plan to handle multiple tasks at the same time?

docs/samples/contact-center/app.js
```diff
@@ -132,7 +159,7 @@ function register() {
         agentLogin.innerHTML = '<option value="" selected>Choose Agent Login ...</option>'; // Clear previously selected option on agentLogin.
         dialNumber.value = '';
         dialNumber.disabled = true;
-        if(loginVoiceOptions.length > 0) agentLoginButton.disabled = false;
+        if(loginVoiceOptions.length > 0) loginAgentElm.disabled = false;
```
sreenara:
```suggestion
        if (loginVoiceOptions.length > 0) loginAgentElm.disabled = false;
```

docs/samples/contact-center/app.js
```diff
@@ -70,6 +77,26 @@ function toggleDisplay(elementId, status) {
   }
 }
 
+const taskEvents = new CustomEvent('task:incoming', {
+  detail: {
+    task: task,
+  },
+});
+
+function registerListeners(task) {
```
sreenara, line 88: This part is confusing because of the way the Task class has been implemented. I'm listening on webex.cc.task, but the payload is also a task? What would the difference between tasks of different types be.. how would the payload vary..

docs/samples/contact-center/app.js
```diff
@@ -245,6 +276,31 @@ async function fetchBuddyAgents() {
...
 
+incomingCallListener.addEventListener('task:incoming', (event) => {
+  if (task.webCallingService.loginOption === 'BROWSER') {
+    answerElm.disabled = false;
+    declineElm.disabled = false;
+  }
+  taskId = event.detail.task.interactionId;
+  const callerDisplay = event.detail.task.interaction.callAssociatedDetails.ani;
+
+  incomingDetailsElm.innerText = `Call from ${callerDisplay}`;
```
sreenara: In the case of non-desktop mode, prehaps it might be better to say "Call from ..., please answer on the endpoint where the agent's extension is registered".

sreenara (changes requested):
I have a few comments regarding the design. Haven't reviewed the UTs yet. Please see the comments.

packages/@webex/plugin-cc/src/types.ts
Kesari3008: In agent Desktop, these call controls are referred as contact controls and we planned to call them task so replaced agent with contact

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Application listens on the task object. task,on. Even if we create a constant emitter, application will have to listen on task object only right

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: We do not need the entire services object but we need to keep the contact layer behind the services layer and aqmReqs needs to be passed to the contact layer which is present within the services layer

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Addressed

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: It is the same payload, there is no difference in payload based on the login option

packages/@webex/plugin-cc/src/types.ts
sreenara: @Kesari3008 would the developer see these names?

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: ok then it's fine

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: So we only need to pass the services objects which are needed here in task right?

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: What if your task is not a call and instead an email or sms or any other omni-channel type?

adhmenon:
@Kesari3008, are we using the latest wxcc branch - only reason I ask is because I think I see some old code (which has now been changed).

packages/@webex/plugin-cc/src/services/Task/index.ts
sreenara: Discussed this. Code will be changed when state machine is implemented.

packages/@webex/plugin-cc/src/services/Task/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  private registerTaskListeners() {
+    // TODO: This event reception approach will be changed once state machine is implemented
+    this.services.webSocketManager.on('message', (event) => {
+      const payload = JSON.parse(event);
+      switch (payload.data.type) {
+        case CC_EVENTS.AGENT_CONTACT_RESERVED:
+          this.taskPayload = payload.data;
+          if (this.webCallingService.loginOption !== LoginOption.BROWSER) {
+            this.emit('task:incoming', payload.data);
+          } else if (this.call) {
```
adhmenon: Just curious, is it possible to encounter a situation where we do not have the calling service created and we are receicing web socket events (for BROWSER) - I would assume if either was present we would be having both (as per the initialisation we do in cc.ts).
Only reason I'm asking is because we can avoid this double if condition. Just a query, however.

packages/@webex/plugin-cc/src/services/core/WebSocket/WebSocketManager.ts
```diff
@@ -142,7 +142,7 @@ export class WebSocketManager extends EventEmitter {
         this.emit('message', e.data);
         const eventData = JSON.parse(e.data);
 
-        if (eventData.type === 'Welcome') {
+        if (eventData.type === CC_EVENTS.WELCOME) {
```
adhmenon, line 145: Thanks for adding this!

packages/@webex/plugin-cc/src/types.ts
Kesari3008: Addressed

packages/@webex/plugin-cc/src/types.ts
Kesari3008: No

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Passing contact and websocketmanager instead of complete services object

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Addressed

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Addressed

packages/@webex/plugin-cc/src/services/Task/types.ts
Kesari3008: Done

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Done

docs/samples/contact-center/app.js
Kesari3008: done

packages/@webex/plugin-cc/src/services/WebCallingService.ts
Kesari3008: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-587471

docs/samples/contact-center/app.js
Kesari3008: Addressed

packages/@webex/plugin-cc/src/services/Task/index.ts
Kesari3008: Will create a TaskManager as part of the state machine story which will handle multiple tasks and their state transitions

docs/samples/contact-center/app.js
Kesari3008: Updated the name of the file

packages/@webex/plugin-cc/src/services/WebSocket/config.ts
sreenara, line 1: Why were these files deleted?

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
```diff
@@ -1,5 +1,5 @@
 import {EventEmitter} from 'events';
-import {WebSocketManager} from './WebSocketManager';
+import {WebSocketManager} from './WebsocketManager';
```
sreenara: For some reason I'm not seeing a change in the file name... Is that just me? Also, do we want to change the class name also to WebsocketManager?

packages/@webex/plugin-cc/src/services/core/WebSocket/types.ts
```diff
@@ -1,5 +1,5 @@
 import {SubscribeRequest} from '../../../types';
-import {WebSocketManager} from './WebSocketManager';
+import {WebSocketManager} from './WebsocketManager';
```
sreenara: same comment here

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -0,0 +1,101 @@
+import EventEmitter from 'events';
+import {ICall, LINE_EVENTS} from '@webex/calling';
```
sreenara, line 2: We discussed about re-exporting these from WebCallingService right? Could we add a TODO here?

packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * This is used for the incoming task decline by agent.
+   *
+   * @param taskId - Unique Id to identify each task
+   *
+   * @returns Promise<TaskResponse>
+   * @throws Error
+   * ```typescript
```
sreenara, line 74: @example is missing here

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -0,0 +1,101 @@
...
+            this.task = new Task(this.contact, this.webCallingService, payload.data);
+
+            this.taskCollection[payload.data.interactionId] = this.task;
+            if (this.webCallingService.loginOption !== LoginOption.BROWSER) {
+              this.emit(TASK_EVENTS.TASK_INCOMING, this.task);
+            } else if (this.call) {
+              this.emit(TASK_EVENTS.TASK_INCOMING, this.task);
+            }
+            break;
+          case CC_EVENTS.AGENT_CONTACT_ASSIGNED:
```
sreenara, line 63: what does this event mean.. why don't we create a task and emit a task object here?

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -0,0 +1,101 @@
...
+   * @param taskId - string.
+   */
+  public getTask = (taskId: string) => {
+    return this.taskCollection[taskId];
+  };
+
+  /**
+   *
+   */
+  public getActiveTasks = (): Record<TaskId, ITask> => {
```
sreenara:
```suggestion
  public getAllTasks = (): Record<TaskId, ITask> => {
```

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -0,0 +1,101 @@
...
+    return this.taskCollection;
+  };
+
+  /**
+   * webSocketManager - WebSocketManager
```
sreenara, line 89: This description doesn't seem right. Can we fix it.

packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+import routingContact from './contact';
+import {ITask, TaskId, TaskResponse, TaskData} from './types';
+import WebCallingService from '../WebCallingService';
+
+export default class Task implements ITask {
+  private contact: ReturnType<typeof routingContact>;
+  private localAudioStream: LocalMicrophoneStream;
+  private webCallingService: WebCallingService;
+  public data: TaskData;
+
```
sreenara, line 14: Should we be sending in the call object here as well since Task and Call are related to each other? How do we know which task corresponds to which call?
What if we have 2-3 calls on an agent at the same time?

packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      if (this.webCallingService.loginOption === LoginOption.BROWSER) {
+        const constraints = {
+          audio: true,
+        };
+
+        const localStream = await navigator.mediaDevices.getUserMedia(constraints);
+        const audioTrack = localStream.getAudioTracks()[0];
+        this.localAudioStream = new LocalMicrophoneStream(new MediaStream([audioTrack]));
+        this.webCallingService.answerCall(this.localAudioStream, taskId);
```
sreenara, line 53: The answer and decline calls seem to suggest that there will only be one call always.. are we sure this is always the case?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -57,11 +63,26 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
 
+      this.incomingTaskListener();
       LoggerProxy.initialize(this.$webex.logger);
     });
   }
 
+  /**
+   * An Incoming Call listener.
+   */
+  private incomingTaskListener() {
```
sreenara, line 80: We should be switching off all listeners on the logout function being called right? Applies to other listeners in WebCallingService too. Additionally, we should also deregister on the calling SDK?

sreenara (changes requested):
I have a few comments about the Task and TaskManager classes. Please check them.

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
sreenara: The build is failing because of this.

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: Actually, IIRC once an agent answers a call they go into a different state so no other call should arrive.. so maybe my scenario won't happen.. but we should try to link the call and the task somehow.

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: Actually, IIRC once an agent answers a call they go into a different state so no other call should arrive.. so maybe my scenario won't happen.. but we should try to link the call and the task somehow.

packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        return Promise.resolve(); // TODO: Update this with sending the task object received in AgentContactAssigned
+      }
+
+      // TODO: Invoke the accept API from services layer. This is going to be used in Outbound Dialer scenario
+      return this.contact.accept({interactionId: taskId});
```
sreenara, line 59: Should we be changing the state of the agent when the call is accepted?

packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        const constraints = {
+          audio: true,
+        };
+
+        const localStream = await navigator.mediaDevices.getUserMedia(constraints);
+        const audioTrack = localStream.getAudioTracks()[0];
+        this.localAudioStream = new LocalMicrophoneStream(new MediaStream([audioTrack]));
+        this.webCallingService.answerCall(this.localAudioStream, taskId);
+
+        return Promise.resolve(); // TODO: Update this with sending the task object received in AgentContactAssigned
```
sreenara, line 55: When does this TODO get taken care of?

packages/@webex/plugin-cc/src/services/WebSocket/config.ts
adhmenon: This was the older web socket Sreekanth iirc, it was deleted a long time back... not sure why it's coming now. Maybe a github bug.

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: This is supposed to be taken care of in the separate story I created for conditionally adding @webex/calling package and there we can take care of re-exporting. Adding a TODO is also tricky because I will have to add it in multiple places. If that's ok, I will add

packages/@webex/plugin-cc/src/services/core/WebSocket/types.ts
Kesari3008: last set of my changes didn't get pushed, now you should see it

packages/@webex/plugin-cc/src/services/core/WebSocket/connection-service.ts
Kesari3008: Build will pass now, this happened due to missing the last commit push

packages/@webex/plugin-cc/src/services/WebSocket/config.ts
Kesari3008: Yeah this folder was an old one and for some reason, it was coming in my changes so I deleted it but I did pull the latest changes also so this change should go away now

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: done

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: this is updated in the latest push

packages/@webex/plugin-cc/src/services/task/index.ts
Kesari3008: There will be one call at a time

packages/@webex/plugin-cc/src/services/task/index.ts
Kesari3008: There will be one call at a time

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+  }
+
+  private registerTaskListeners() {
+    this.webSocketManager.on('message', (event) => {
+      const payload = JSON.parse(event);
+      if (!event.keepalive) {
+        switch (payload.data.type) {
+          case CC_EVENTS.AGENT_CONTACT_RESERVED:
+            this.task = new Task(this.contact, this.webCallingService, payload.data);
```
adhmenon, line 54: One doubt, are we sure this will properly store the required task details from the payload? While testing I had found some issues, so I wanted to be a 100% sure.

packages/@webex/plugin-cc/src/services/task/types.ts
```diff
@@ -0,0 +1,330 @@
...
+      name: string;
+      qualifier: string;
+      description: string;
+      valueDataType: string;
+      value: string;
+    }
+  >;
+};
+
+export type TaskData = {
```
adhmenon, line 142: Are we not storing the aux id here?

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: This is an event which is receiuved when call is accepted by the agent. Updated the code to emit the task object only

packages/@webex/plugin-cc/src/services/task/index.ts
Kesari3008: added

packages/@webex/plugin-cc/src/services/task/types.ts
Kesari3008: It's present inside the agent profile, why do we need to keep it within the Task object

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: Could you please elaborate on what issue you came across?  Within the payload, data property is what we are passing which has complete event detail which will get saved as one of the property within the task object

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: Deregister is already happening when we are logging out

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: @Kesari3008 have you tested this? Is there an event that arrives when we answer the call from WxCC backend?

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: This comment hasn't been addressed @Kesari3008

packages/@webex/plugin-cc/src/cc.ts
sreenara: @Kesari3008 but deregister doesn't take care of these listeners right? This is a listener on taskManager.

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: @Kesari3008 should we still link a call and task since a call is still a task?

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
sreenara: I don't see this updated in the latest PR #4016

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: Have added the off listener

packages/@webex/plugin-cc/src/services/task/index.ts
Kesari3008: When we have state machine implemented

packages/@webex/plugin-cc/src/services/task/index.ts
Kesari3008: I have tested this, incoming event is not even received for the second call

sreenara:
Replaced with #4016

packages/@webex/plugin-cc/src/services/task/index.ts
sreenara: Resolved through testing. Agent is automatically marked busy on the backend.

PR #4031, opened 2024-12-17
feat(plugin-meetings): add brb logic for plugin meetings when current user steps away or returns

antsukanova:
# COMPLETES # [SPARK-559645](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-559645)

This PR introduces a new meeting API `sendBrb` to allow clients to enable step away feature.

This PR introduces a couple of important updates: 
- adds new API in meeting `beRightBack()`
- adds new unit tests
- adds new test app functionality (see screenshots)

Test app additions to test this feature

Screenshot 2024-12-16 at 19 48 02
Screenshot 2024-12-16 at 19 47 52

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- tested manually with convergedats account using the test app and Webex client
- adds new test cases that covers new logic

### I certified that

---


packages/@webex/plugin-meetings/src/member/types.ts
```diff
@@ -23,6 +23,14 @@ export type ParticipantWithRoles = {
   };
 };
 
+export type ParticipantWithBrb = {
+  controls?: {
```
antsukanova: One thing that I haven't found easily is in what situations `controls` can be `undefined`. I would appreciate it if someone could highlight that. For now, I will leave this part consistent with other member logic and add optional `controls` logic.

docs/samples/calling.min.js
antsukanova, line 1: I am not sure if I need to load new samples for changes in test-app, appreciate it if you comment on that

brycetham: No need to push these, you can remove them from your PR.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -793,6 +793,75 @@ describe('plugin-meetings', () => {
...
+            LOCUSINFO.EVENTS.SELF_MEETING_BRB_CHANGED,
+            {brb: value}
+          );
+        };
+
+        assertBrbUnchanged(true);
+        assertBrbUnchanged(false);
+      });
+
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state changed to undefined', () => {
```
brycetham: This probably isn't the correct test description here. Locus doesn't change the BRB state to undefined, but rather it may be undefined if there is no BRB state change. So I would change this to `should not trigger SELF_MEETING_BRB_CHANGED when brb state is undefined`.

k-wasniowski (approved):

k-wasniowski:
LGTM

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
antsukanova: @brycetham you're right, good idea, your description feels much better, thanks!

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3094,7 +3099,7 @@ participantTable.addEventListener('click', (event) => {
     }
     const muteButton = document.getElementById('mute-participant-btn')
     if (selectedParticipant.isAudioMuted) {
-      muteButton.innerText = meeting.selfId === selectedParticipant.id ? 'Unmute' : 'Request to unmute';
+      muteButton.innerText = isUserSelf(selectedParticipant) ? 'Unmute' : 'Request to unmute';
```
sreenara: Were we observing any issues with this logic, leading to the change?

packages/@webex/plugin-meetings/src/meeting/request.ts
```diff
@@ -909,4 +909,29 @@ export default class MeetingRequest extends StatelessWebexPlugin {
...
+   * Sends a request to set be right back status.
+   *
+   * @param {Object} options - The options for brb request.
+   * @param {boolean} options.enabled - Whether brb status is enabled.
+   * @param {string} options.locusUrl - The URL of the locus.
+   * @param {string} options.deviceUrl - The URL of the device.
+   * @param {string} options.selfId - The ID of the participant.
+   * @returns {Promise}
+   */
+  sendBrb({enabled, locusUrl, deviceUrl, selfId}: BrbOptions) {
```
sreenara: Would `toggleBrb` or `setBrb` be better alternatives for the function name?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -32,8 +32,8 @@ import {
   NETWORK_STATUS,
   ONLINE,
   OFFLINE,
-  ROAP_OFFER_ANSWER_EXCHANGE_TIMEOUT,
-} from '@webex/plugin-meetings/src/constants';
+  ROAP_OFFER_ANSWER_EXCHANGE_TIMEOUT, HTTP_VERBS, PARTICIPANT, CONTROLS,
+} from "@webex/plugin-meetings/src/constants";
```
sreenara: Any reason for this change?
```suggestion
} from '@webex/plugin-meetings/src/constants';
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -4796,6 +4870,7 @@ describe('plugin-meetings', () => {
             });
           });
```
sreenara: This change isn't needed right?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3565,6 +3579,50 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      .sendBrb({
+        enabled,
+        locusUrl: this.locusUrl,
+        deviceUrl: this.deviceUrl,
+        selfId: this.selfId,
+      })
+      .then(() => {
+        this.sendSlotManager.setSourceStateOverride(MediaType.VideoMain, enabled ? 'away' : null);
+      })
+      .catch((error) => {
```
sreenara, line 3646: I'm not able to find a test for this catch block. Could you point me to the test where this is being tested?

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -793,6 +793,75 @@ describe('plugin-meetings', () => {
...
+            LOCUSINFO.EVENTS.SELF_MEETING_BRB_CHANGED,
+            {brb: value}
+          );
+        };
+
+        assertBrbUnchanged(true);
+        assertBrbUnchanged(false);
+      });
+
+      it('should not trigger SELF_MEETING_BRB_CHANGED when brb state is undefined', () => {
```
sreenara, line 846: Can we also add a test when `controls` is undefined because that is also an optional object?

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
sreenara, line 1: I don't see a changeset for selfUtils test file. Could we add them as well?

sreenara (changes requested):
I've got a few comments regarding the UT. Please check.

Additionally, please see this **[vidcast](https://app.vidcast.io/share/2d3ff8b2-bd4d-4038-92bf-5df43314dca3)** for a discrepancy I've observed between the SDK and desktop app behavior with regards to step-away.

Is there a reason we have created this PR from another branch in `upstream` rather than a fork?

antsukanova:
@sreenara, thank you for the review.

Regarding the branch, I have created the upstream version because we divided these changes between the WCME team and did them simultaneously. Once it was finished in one branch (this one brb-feature), I created a PR for the SDK team to review.

docs/samples/browser-plugin-meetings/app.js
antsukanova: @sreenara This change is just adding a new function, `isUserSelf,` that utilises logic for checking `meeting.selfId === member.id` in a couple of places. It does not change the logic for this function.
I need the same check for `brb` and came up with a utility function for this. We can discuss whether the test app needs such a change.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
antsukanova: I'm not sure why linter didn't catch that single quotes were changed (I will double check this), I believe this is an issue with style changes js files. I assume just didn't catch that. Will update. Thanks

packages/@webex/plugin-meetings/src/meeting/request.ts
antsukanova: I like the idea of renaming it to `setBrb`, makes sense thanks for it.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
antsukanova: @sreenara not sure if I got what change you refer to here, could you please specify?

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @sreenara haven't found that case, will add it, thank you for pointing this.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
antsukanova: @sreenara will extend with new test case for `updateSelf` for all event that relies on check that controls is optional, because I haven't found this and it's not hard to do, so I will appreciate your feedback on that idea

packages/@webex/plugin-meetings/src/locus-info/selfUtils.ts
antsukanova: @sreenara Thanks for catching that, I will update this file

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3052,6 +3052,11 @@ function moveFromDevice() {
   });
 }
 
+function isUserSelf(member) {
```
marcin-bazyl: nitpick: For me "User" usually means the current user using the app and in this case we are not checking the current user but some other member, so I think a better name would be `isMemberSelf` for this function.

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3330,6 +3335,25 @@ function toggleBreakout() {
   }
 }
 
+async function toggleBrb() {
+  const meeting = getCurrentMeeting();
+
+  if (meeting) {
+    const brbButton = document.getElementById('brb-btn');
+    const isBrbEnabled = brbButton.innerText === 'Step away';
```
marcin-bazyl: the name `isBrbEnabled` is a bit confusing, because when the button's innerText says "Step away", it means that we are in the meeting so BRB is not enabled, while `isBrbEnabled` will be set to true in that case

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -83,6 +84,36 @@ export default class SendSlotManager {
...
   }
 
+  /**
+   * Sets the source state override for the given media type.
+   * @param {MediaType} mediaType - The type of media (must be MediaType.VideoMain to apply source state changes).
+   * @param {StreamState | null} state - The state to set or null to clear the override value.
+   * @returns {void}
+   */
+  public setSourceStateOverride(mediaType: MediaType, state: StreamState | null) {
+    if (mediaType !== MediaType.VideoMain) {
```
marcin-bazyl, line 94: just out of curiosity: does Homer support sending 'away' status for video slides? If it does, then maybe we should also allow that in the SDK?

When testing this feature in UCF, when I step away while sharing, UCF stops the screen share, so I guess we will do the same logic in the webex web app, but in theory some other app might want to have different logic and keep sharing but with sourceState overridden to "away" maybe?

marcin-bazyl (commented):
LGTM, just some minor comments and questions, I haven't tested it as I had some problems with joining the meeting using sample app, I'll try again tomorrow

docs/samples/browser-plugin-meetings/app.js
antsukanova: @marcin-bazyl thanks for the idea, I like logic behind it, will update

docs/samples/browser-plugin-meetings/app.js
antsukanova: @marcin-bazyl Good point, I have an idea to change it to `enableBrb`. In this way we can directly understand what action will be done: either enable or disable brb. I hope this will also make you feel better with naming here.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: @marcin-bazyl from the JMP spec side and WCME it's possible, so your point is good to verify. Thanks. I will ask Homer team about their expectations regarding  'away' status for video slides. I believe it possible that other sdk clients would like to rely on overridden value for video slides.

packages/@webex/plugin-meetings/src/member/types.ts
marcin-bazyl: I don't know, probably best to ask Locus team in "Ask Locus" space as the participants come to us from Locus

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
marcin-bazyl: if Homer confirms that they support "away" for video slides, then I think it would be best to add a parameter to `Meeting.beRightBack()` so that the app calling that API can decide if they want the status to be applied to just main video, video slides or both

antsukanova:
@sreenara I would appreciate your additional review. The only question regarding the test app logic for now you can omit. I would like to check it by myself and discuss it with the team next week, and then come back to you with a final suggestion.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: @marcin-bazyl Yes, I believe it can be an optional additional param, good idea, thank you

packages/@webex/plugin-meetings/src/member/types.ts
antsukanova: @marcin-bazyl thanks for suggesting, will double-check with them

antsukanova: From Locus team, they pointed it should be always available so will update the logic and remove `?` for `controls`

antsukanova: Updated it to make controls not optional

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
marcin-bazyl: I've approved this PR as we can do this change in another smaller PR

chrisadubois (approved):

antsukanova:
@sreenara updated the logic with the test app that you reviewed https://github.com/webex/webex-js-sdk/pull/4054. As we discussed will proceed with merging this PR and working on documenting it. Appreciate your help

sreenara (approved):

PR #4038, opened 2024-12-20
fix: end meeting for all getCurUserType type issue

rsarika:
# COMPLETES # [CSCwn60262](https://bst.cisco.com/bugsearch/bug/CSCwn60262)

fixing Console error when we call meeting.EndMeetingForAll()

Added below properties and function to BasicMeetingInformation type so that they will be accessible after ending the meeting as well.

roles: string[];
 getCurUserType: () => string | null;
 callStateForMetrics: CallStateForMetrics;

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

 Joined meeting as host in sample app and called endMeetingForAll() => there were no errors observed.

### I certified that

---

## Summary by CodeRabbit

- **New Features**
	- Enhanced meeting data structure to retain user roles and type information after a meeting is deleted.
	- Added new properties for improved meeting state management.

- **Bug Fixes**
	- Improved handling of user scenarios in meeting functionality tests, including guest users and incomplete data.

- **Tests**
	- Refined test cases and added new assertions to ensure accurate validation of meeting functionalities and metrics handling.

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -155,6 +155,9 @@ export type BasicMeetingInformation = {
   };
   meetingInfo: any;
   sessionCorrelationId: string;
+  roles: string[];
+  getCurUserType: () => string | null;
+  callStateForMetrics: CallStateForMetrics;
 };
 
 /**
@@ -1143,6 +1146,9 @@ export default class Meetings extends WebexPlugin {
           sessionId: meeting.locusInfo?.fullState?.sessionId,
         },
       },
+      roles: meeting.roles,
+      callStateForMetrics: meeting.callStateForMetrics,
+      getCurUserType: meeting.getCurUserType,
     });
     this.meetingCollection.delete(meeting.id);
     Trigger.trigger(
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1143,6 +1146,9 @@ export default class Meetings extends WebexPlugin {
           sessionId: meeting.locusInfo?.fullState?.sessionId,
         },
       },
+      roles: meeting.roles,
+      callStateForMetrics: meeting.callStateForMetrics,
+      getCurUserType: meeting.getCurUserType,
```
marcin-bazyl, line 1151: initially I didn't like this approach, because I thought that having a reference to the method of a meeting will prevent that meeting object from being garbage collected, but I think it's OK, you can have references to methods of an object and these methods on their own don't have a reference to the object itself (as long as they don't use any closures), so it should be fine.

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -155,6 +155,9 @@ export type BasicMeetingInformation = {
   };
   meetingInfo: any;
   sessionCorrelationId: string;
+  roles: string[];
```
sreenara, line 158: question. Why have we also added roles and callStateForMetrics?

sreenara (approved):
Just one question.

packages/@webex/plugin-meetings/src/meetings/index.ts
rsarika: roles being read in getCurUserType utility function and callStateForMetrics is being read in next line where previous error occured.

PR #4044, opened 2025-01-06
fix(cc-sdk): fixed-websocket-on-message-event-remove-logout

adhmenon:
# COMPLETES #< [SPARK-592068](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-592068) >

* Currently, in the widgets, whenever we login and then logout and once again login to the widget, we are unable to get any message events.
* This is causing an issue with the agent status and timer as shown in this PR - https://github.com/webex/widgets/pull/352

* Changed the code so that we now listen to websocket messages (again only used in widgets as of now) whenever we login/ relogin and stop listening whenever we do logout.

Vidcast of working widgets - https://app.vidcast.io/share/a1cac1b3-83ec-450d-9e63-4a3066287b93

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested out the code by locally linking to the widgets repo.
* Tested the code in the sdk sample app to ensure no older behaviour is broken.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-cc/src/cc.ts
+++ b/packages/@webex/plugin-cc/src/cc.ts
@@ -194,6 +194,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       await loginResponse;
 
+      this.services.webSocketManager.on('message', this.handleWebSocketMessage);
       this.incomingTaskListener();
       this.taskManager.registerIncomingCallEvent();
 
@@ -294,7 +295,6 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    */
   private setupEventListeners() {
     this.services.connectionService.on('connectionLost', this.handleConnectionLost.bind(this));
-    this.services.webSocketManager.on('message', this.handleWebSocketMessage);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -194,6 +194,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       await loginResponse;
 
+      this.services.webSocketManager.on('message', this.handleWebSocketMessage);
```
adhmenon, line 197: This change will not affect any other functionality as it is only handling `AGENT_STATE_CHANGE` event. We were earlier listening to web socket events during register which would not have worked.
Checked that it works cleanly in the sample app.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -355,6 +355,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       await this.handleDeviceType(deviceType as LoginOption, dn);
       this.agentConfig.isAgentLoggedIn = true;
+      this.services.webSocketManager.on('message', this.handleWebSocketMessage);
```
adhmenon, line 358: Need to add this for relogin as well (in those cases).

mkesavan13 (approved):
Thanks a lot, Adhwaith. Looks good to me. A test for the exact case where you found the issue: "Doing a station logout and logging in back didn't listen to the event" would be a good addition

packages/@webex/plugin-cc/src/cc.ts
Kesari3008, line 197: Where would we be switching off these listeners? Is that already taken care of somewhere?

adhmenon: It's already being switched off on logout.

Kesari3008 (approved):

PR #4078, opened 2025-01-30
fix(cc-sdk): fixed-relogin-issue

adhmenon:
# COMPLETES #< [SPARK-604350](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-604350) >

* On relogin, we were not able to receive a task incoming event.

* Changes were made to once again listen to events on silent relogin

Vidcast - https://app.vidcast.io/share/3bf01944-75f8-45f2-8592-3393519b8c43

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested for regular task incoming event
* Refreshed page and tested for task incoming event
* Switched off network and tested for incoming event

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-cc/src/cc.ts
+++ b/packages/@webex/plugin-cc/src/cc.ts
@@ -345,6 +345,10 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       const reLoginResponse = await this.services.agent.reload();
       const {auxCodeId, agentId, lastStateChangeReason, deviceType, dn} = reLoginResponse.data;
 
+      // To handle re-registration of event listeners on silent relogin
+      this.incomingTaskListener();
+      this.taskManager.registerIncomingCallEvent();
+
       if (lastStateChangeReason === 'agent-wss-disconnect') {
         LoggerProxy.info(
           'event=requestAutoStateChange | Requesting state change to available on socket reconnect',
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -395,6 +399,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
         });
         throw new Error(`Unsupported device type: ${deviceType}`);
     }
+    this.webCallingService.setLoginOption(deviceType);
```
adhmenon, line 402: Need to do this so we can get calls again in the sample app. Based on this we emit the events.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -359,6 +359,10 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
         await this.setAgentState(stateChangeData);
       }
 
+      // To handle re-registration of event listeners on silent relogin
+      this.incomingTaskListener();
+      this.taskManager.registerIncomingCallEvent();
+
       await this.handleDeviceType(deviceType as LoginOption, dn);
       this.agentConfig.isAgentLoggedIn = true;
       this.services.webSocketManager.on('message', this.handleWebSocketMessage);
```
rsarika: The changes looks good. Is it possible to move this line up before setAgentState so the statechange event will be triggered in widgets for re-login

rsarika (changes requested):

packages/@webex/plugin-cc/src/cc.ts
rsarika: I think it will resolve the available state issue in widgets

adhmenon: I don't think it worked Ravi... reason is when we login we don't get the change event always, only if we were available and the internet disconnected - the agent state issue would be something else... we can debug that on the widgets side tho.

rsarika (approved):

Shreyas281299 (approved):

PR #4088, opened 2025-02-07
fix(plugin-meetings): sip 1-1 calls cannot connect

marcin-bazyl:
# COMPLETES #[SPARK-612371](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-612371)

We've recently changed web app to use multistream for all types of meeting and calls, but it turns out that there are still some types of calls the Homer doesn't support: these are SIP 1-1 calls. Homer will support them in a few months, but for now when we try to connect we receive an SDP answer from Linus instead of Homer and multistream code cannot handle that.

Added handling for the case when received SDP answer is not from Homer - we fallback to a transcoded meeting and re-create the media connection, send new TURN discovery request and new SDP offer.

Overview how this works:
 - when we get SDP answer not from homer as we expect, `Meeting.roapMessageReceived()` will throw `MultistreamNotSupportedError`
 - as a result, `Meeting.establishMediaConnection()` will also throw that error and it will be caught in `Meeting.addMediaInternal()` which calls `Meeting.downgradeFromMultistreamToTranscoded()` and then `Meeting.establishMediaConnection()` again - this time forcing TURN discovery to be done too.
 - `Meeting.downgradeFromMultistreamToTranscoded()` does the cleanup of all things that were setup for multistream

Corresponding web app PR: 
https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/7514
(web app needs to just check if multistream is enabled after joinWithMedia() resolves and if not, update its UI to transcoded one).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual e2e test with web app and sip device

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -161,6 +161,7 @@ import {LocusMediaRequest} from './locusMediaRequest';
 import {ConnectionStateHandler, ConnectionStateEvent} from './connectionStateHandler';
 import JoinWebinarError from '../common/errors/join-webinar-error';
 import Member from '../member';
+import MultistreamNotSupportedError from '../common/errors/multistream-not-supported-error';
 
 // default callback so we don't call an undefined function, but in practice it should never be used
 const DEFAULT_ICE_PHASE_CALLBACK = () => 'JOIN_MEETING_FINAL';
@@ -4627,11 +4628,12 @@ export default class Meeting extends StatelessWebexPlugin {
    * Close the peer connections and remove them from the class.
    * Cleanup any media connection related things.
    *
+   * @param {boolean} resetMuteStates whether to also rese the audio/video mute state information
    * @returns {Promise}
    * @public
    * @memberof Meeting
    */
-  public closePeerConnections() {
+  public closePeerConnections(resetMuteStates = true) {
     if (this.mediaProperties.webrtcMediaConnection) {
       if (this.remoteMediaManager) {
         this.remoteMediaManager.stop();
@@ -4648,8 +4650,10 @@ export default class Meeting extends StatelessWebexPlugin {
       this.setNetworkStatus(undefined);
     }
 
-    this.audio = null;
-    this.video = null;
+    if (resetMuteStates) {
+      this.audio = null;
+      this.video = null;
+    }
 
     return Promise.resolve();
   }
@@ -4909,7 +4913,7 @@ export default class Meeting extends StatelessWebexPlugin {
    * @param {Object} options - options to join with media
    * @param {JoinOptions} [options.joinOptions] - see #join()
    * @param {AddMediaOptions} [options.mediaOptions] - see #addMedia()
-   * @returns {Promise} -- {join: see join(), media: see addMedia()}
+   * @returns {Promise} -- {join: see join(), media: see addMedia(), multistreamEnabled: flag to indicate if we managed to join in multistream mode}
    * @public
    * @memberof Meeting
    * @example
@@ -4999,6 +5003,7 @@ export default class Meeting extends StatelessWebexPlugin {
       return {
         join: joinResponse,
         media: mediaResponse,
+        multistreamEnabled: this.isMultistream,
       };
     } catch (error) {
       LoggerProxy.logger.error('Meeting:index#joinWithMedia --> ', error);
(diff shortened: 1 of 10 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5007,7 +5012,17 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      let shouldRetry = !isRetry;
+
+      if (CallDiagnosticUtils.isSdpOfferCreationError(error)) {
+        // errors related to offer creation (for example missing H264 codec) will happen again no matter how many times we try,
+        // so there is no point doing a retry
+        shouldRetry = false;
+      }
+
+      // we only want to call leave if join was successful and this was a retry or we won't be doing any more retries
+      if (joined && (isRetry || !shouldRetry)) {
```
marcin-bazyl, line 5025: These changes here are not strictly required for this fix, but it's something I've found while working on this and initially I thought I would need this fix, but later changed my approach. The issue here was that if joinWithMedia() failed and decided that there is no point in doing a retry, it wasn't calling leave() and it should, so I've fixed it.

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -231,14 +231,16 @@ export default class Roap extends StatelessWebexPlugin {
...
 
-            roapAnswer = {
-              seq: answerSeq,
-              messageType,
-              sdp: sdps[0],
-              errorType,
-              errorCause,
-              headers,
-            };
+            if (messageType === ROAP.ROAP_TYPES.ANSWER) {
```
marcin-bazyl, line 234: Another thing I've found while testing - we might get something else than an ANSWER (for example ERROR) and then the `sdps` array will be empty so we would crash. I've added a unit test for this too.

arun3528 (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6064,6 +6070,11 @@ export default class Meeting extends StatelessWebexPlugin {
   public roapMessageReceived = (roapMessage: RoapMessage) => {
     const mediaServer = MeetingsUtil.getMediaServer(roapMessage.sdp);
 
+    if (this.isMultistream && mediaServer !== 'homer') {
```
chrisadubois, line 6085: can we make a MEDIA_SERVER constant somewhere? I'm not sure if we check for homer in the SDP for example anywhere else?

packages/@webex/plugin-meetings/src/meetings/util.ts
```diff
@@ -90,7 +90,8 @@ MeetingsUtil.getMediaServer = (sdp) => {
       .find((line) => line.startsWith('o='))
       .split(' ')
       .shift()
-      .replace('o=', '');
+      .replace('o=', '')
+      .toLowerCase();
```
chrisadubois, line 94: nice. Hopefully we added a test for the casing

chrisadubois (approved):
Code looks good. I did not review the unit tests. One small comment.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4627,11 +4628,12 @@ export default class Meeting extends StatelessWebexPlugin {
    * Close the peer connections and remove them from the class.
    * Cleanup any media connection related things.
    *
+   * @param {boolean} resetMuteStates whether to also rese the audio/video mute state information
```
jor-row: typo : ```rese``` -> ```reset```

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -8341,7 +8405,7 @@ export default class Meeting extends StatelessWebexPlugin {
     if (layoutType) {
       if (!LAYOUT_TYPES.includes(layoutType)) {
         return this.rejectWithErrorLog(
-          'Meeting:index#changeVideoLayout --> cannot change video layout, invalid layoutType received.'
+          `Meeting:index#changeVideoLayout --> cannot change video layout, invalid layoutType "${layoutType}" received.`
```
jor-row, line 8420: nice

jor-row (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: We don't check anywhere else, this is the only place for homer. There is a MEDIA_SERVER constant in SDK tests that has 3 values, LINUS, HOMER, CALLIOPE - which is not really correct because Linus and Homer are part of Calliope and also there is an issue that getMediaServer() only works correctly for Homer, making it distinguish between Linus and Edonus requires more effort, not even sure if we are able to tell them apart based on the SDP, so this is not something I want to be doing now in this PR.

packages/@webex/plugin-meetings/src/meetings/util.ts
marcin-bazyl: added the test now

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: fixed

PR #4089, opened 2025-02-07
feat(plugin-cc): add task sync on reload

rarajes2:
# COMPLETES #<[SPARK-610352](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610352) >

Adds event to sync the task on page refresh/reload.

e.g.

```js
    webex.cc.on('task:sync', (currentTask) => {
        task = currentTask;
    });
```

Able to get the current task and update the samples page

Vidcast - https://app.vidcast.io/share/36105f68-ca3c-40b9-a108-6731a1be4ba3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manual Testig

- Tested with the `Extension` Agent login - Seems to be working fine
- Tested with `Browser` Agent login - There is one issue with the mobius device registration where when we do the `silentRelogin` we get 403 error on first registration try, then DELETE request goes followed by device registration successful -- causing the call to disconnect and we get the wrap up state in place. This requires investigation in calling SDK. I will update this scenario in the test plan if it's not there already

**Tested the following controls:**

- Answer/Decline
- hold/resume
- end 
- pause/resume recording
- wrapUp

**Code added but needs more testing as it's behaving bit flaky**

- consult
- consultTransfer
- transfer (Seems to be working as expected, could be tested few more times to be sure)

### I certified that

---

### DIFF
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -185,6 +185,8 @@ function updateButtonsPostEndCall() {
   showConsultButton()
   disableTransferControls();
   consultTabBtn.disabled = true;
+  incomingDetailsElm.innerText = '';
+  pauseResumeRecordingElm.innerText = 'Pause Recording';
 }
 
 function showInitiateConsultDialog() {
@@ -589,28 +591,8 @@ function register() {
     });
 
     webex.cc.on('task:sync', (currentTask) => {
-      task = currentTask;
-      
-      const {state, isTerminated} = task.data.interaction;
-
-      if (!isTerminated) {
-        holdResumeElm.disabled = false;
-        holdResumeElm.innerText = 'Hold';
-        pauseResumeRecordingElm.disabled = false;
-        pauseResumeRecordingElm.innerText = 'Pause Recording';
-        endElm.disabled = false;
-        enableConsultControls(); // Enable consult controls
-        enableTransferControls(); // Enable transfer controls
-
-        return;
-      }
-
-      if (state === 'connected') {
-        wrapupCodesDropdownElm.disabled = false;
-        wrapupElm.disabled = false;
-      }
+      handleTaskSync(currentTask);
     });
-  
 
     webex.cc.on('agent:stateChange', (data) => {
       if (data && typeof data === 'object' && data.type === 'AgentStateChangeSuccess') {
@@ -626,6 +608,74 @@ function register() {
       }
     });
     
+}
+
+function handleTaskSync(currentTask) {
+  task = currentTask;
+      
+  if (!task || !task.data) {
+    console.error('task:sync --> No task data found.');
+    
+    return;
+  }
+
+  console.log('Task: ', task);
+
+  const { data, webCallingService } = task;
+  const { interaction, interactionId, mediaResourceId, agentId  } = data;
+  const {
+    state,
+    isTerminated,
+    media,
+    participants,
+    callAssociatedDetails,
+    callProcessingDetails
+  } = interaction;
+
+  if (isTerminated) {
+    // wrapup
+    if (state === 'wrapUp' && !participants[agentId].isWrappedUp) {
+      wrapupCodesDropdownElm.disabled = false;
+      wrapupElm.disabled = false;
+    }
+
+    return;
+  }
+
+  // answer & decline incoming calls
+  const callerDisplay = callAssociatedDetails?.ani;
+  if (webCallingService.loginOption === 'BROWSER') {
+    answerElm.disabled = false;
+    declineElm.disabled = false;
+
+    incomingDetailsElm.innerText = `Call from ${callerDisplay}`;
+  } else {
+    incomingDetailsElm.innerText = `Call from ${callerDisplay}...please answer on the endpoint where the agent's extension is registered`;
+  }
+
+  // end button
+  const hasParticipants = Object.keys(participants).length > 1;
+  endElm.disabled = !hasParticipants;
+
+  // consult, transfer
+  // TODO: Test the consult and transfer features and update the conditions accordingly
+  consultTabBtn.disabled = !hasParticipants; // Enable consult controls
+  transferElm.disabled = !hasParticipants; // Enable transfer controls
+  
+  // TODO: consultTransferBtn
+  // console.log('consultState >>>>>>>>', participants[agentId].consultState);
+
+  // hold/resume call
+  const isHold = media && media[mediaResourceId] && media[mediaResourceId].isHold;
+  holdResumeElm.disabled = isTerminated;
+  holdResumeElm.innerText = isHold ? 'Resume' : 'Hold';
+
+  // pause/resume recording
+  if (callProcessingDetails) {
+    const {pauseResumeEnabled, isPaused} = callProcessingDetails;
+    pauseResumeRecordingElm.disabled = !pauseResumeEnabled;
+    pauseResumeRecordingElm.innerText = isPaused === 'true' ? 'Resume Recording' : 'Pause Recording';
+  }
 }
 
 function populateWrapupCodesDropdown() {
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -588,6 +588,30 @@ function register() {
       incomingCallListener.dispatchEvent(taskEvents);
     });
 
+    webex.cc.on('task:sync', (currentTask) => {
+      task = currentTask;
+      
+      const {state, isTerminated} = task.data.interaction;
+
+      if (!isTerminated) {
+        holdResumeElm.disabled = false;
+        holdResumeElm.innerText = 'Hold';
+        pauseResumeRecordingElm.disabled = false;
+        pauseResumeRecordingElm.innerText = 'Pause Recording';
+        endElm.disabled = false;
+        enableConsultControls(); // Enable consult controls
+        enableTransferControls(); // Enable transfer controls
+
+        return;
+      }
+
+      if (state === 'connected') {
+        wrapupCodesDropdownElm.disabled = false;
+        wrapupElm.disabled = false;
+      }
+    });
+  
+
     webex.cc.on('agent:stateChange', (data) => {
       if (data && typeof data === 'object' && data.type === 'AgentStateChangeSuccess') {
         const DEFAULT_CODE = '0'; // Default code when no aux code is present
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -639,15 +639,13 @@ function register() {
 
 function handleTaskHydrate(currentTask) {
   task = currentTask;
-      
+
   if (!task || !task.data || !task.data.interaction) {
     console.error('task:hydrate --> No task data found.');
     alert('task:hydrate --> No task data found.');
     
     return;
   }
-
-  console.log('Task: ', task);
 
   const { data, webCallingService } = task;
   const { interaction, mediaResourceId, agentId } = data;
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -590,8 +590,8 @@ function register() {
       incomingCallListener.dispatchEvent(taskEvents);
     });
 
-    webex.cc.on('task:sync', (currentTask) => {
-      handleTaskSync(currentTask);
+    webex.cc.on('task:hydrate', (currentTask) => {
+      handleTaskHydrate(currentTask);
     });
 
     webex.cc.on('agent:stateChange', (data) => {
@@ -610,11 +610,12 @@ function register() {
     
 }
 
-function handleTaskSync(currentTask) {
+function handleTaskHydrate(currentTask) {
   task = currentTask;
       
-  if (!task || !task.data) {
-    console.error('task:sync --> No task data found.');
+  if (!task || !task.data || !task.data.interaction) {
+    console.error('task:hydrate --> No task data found.');
+    alert('task:hydrate --> No task data found.');
     
     return;
   }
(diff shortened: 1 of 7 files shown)


packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -347,7 +353,6 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       // To handle re-registration of event listeners on silent relogin
       this.incomingTaskListener();
-      this.taskManager.registerIncomingCallEvent();
```
mkesavan13, line 355: Why are we removing this listener? What is the alternative here?

docs/samples/contact-center/app.js
```diff
@@ -588,6 +588,30 @@ function register() {
...
     });
 
+    webex.cc.on('task:sync', (currentTask) => {
+      task = currentTask;
+      
+      const {state, isTerminated} = task.data.interaction;
+
+      if (!isTerminated) {
+        holdResumeElm.disabled = false;
+        holdResumeElm.innerText = 'Hold';
```
mkesavan13: What if the agent puts the customer on hold and refreshes the page?

docs/samples/contact-center/app.js
```diff
@@ -588,6 +588,30 @@ function register() {
...
+        pauseResumeRecordingElm.disabled = false;
+        pauseResumeRecordingElm.innerText = 'Pause Recording';
```
mkesavan13: Same here. What if the agent already paused recording and refreshed the page?

rarajes2:
@coderabbitai review

mkesavan13 (changes requested):
Let's also test reload scenarios when the agent has already held the call / paused the recording. What if the user is in the middle of consult and transfer and reloads the page? Do we put them back in the consult state and in call with the right party?

docs/samples/contact-center/app.js
Shreyas281299: This might be difficult we dont get any indication in the task object that the call is held or the recording is paused or not. We can investigate this just to be sure, but afaik we dont get this information

packages/@webex/plugin-cc/src/cc.ts
rarajes2: This listener was not required here as this gets called inside the constructor itself so, as soon as we create the task manager, this gets invoked. Confirmed with Adhwaith - He agreed with this change. Tested the functionality, it's not breaking.

docs/samples/contact-center/app.js
mkesavan13: There should be someway. Let's check how the WxCC Agent Desktop does it and come to a conclusion on this

rarajes2: Handled this scenario

docs/samples/contact-center/app.js
rarajes2: Handled this scenario

rarajes2:
> Let's also test reload scenarios when the agent has already held the call / paused the recording. What if the user is in the middle of consult and transfer and reloads the page? Do we put them back in the consult state and in call with the right party?

@mkesavan13  Update the manual test section. Except `consult` and `transfer` all other scenarios are tested and fixed. Will test the remaining scenarios and update the PR.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -79,11 +79,17 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    // @ts-ignore
+    this.trigger(TASK_EVENTS.TASK_SYNC, task);
+  };
+
   /**
    * An Incoming Call listener.
    */
   private incomingTaskListener() {
     this.taskManager.on(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
+    this.taskManager.on(TASK_EVENTS.TASK_SYNC, this.handleTaskSync);
```
Kesari3008: Where are we switching off this listener?

packages/@webex/plugin-cc/src/cc.ts
Kesari3008, line 355: I feel that the constructor might be the place where it is redundant and not here. If we see station login and station logout, we are switching on the incoming call listener during login and switching it off during logout. Now for the case of silent relogin, the usual login is skipped hence, we added a listener being switched on here as well in silent relogin too.

Now, if we decide to remove the listener from here, then we should remove it from the station login method as well otherwise, we need to add this one back and remove it from the TaskManager constructor

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -65,6 +65,11 @@ export default class TaskManager extends EventEmitter {
       const payload = JSON.parse(event);
       if (payload.data) {
         switch (payload.data.type) {
+          case CC_EVENTS.AGENT_CONTACT:
+            this.currentTask = new Task(this.contact, this.webCallingService, payload.data);
```
Kesari3008, line 69: Did we check the possibility of receiving the array of tasks in this event payload data? If that's the case, then creating a single task object might not be the correct approach to handling this event. We might need to have a loop running and updating the task collection with each of the array items (pending tasks).

packages/@webex/plugin-cc/src/services/task/types.ts
```diff
@@ -49,6 +49,7 @@ export const TASK_EVENTS = {
   TASK_END: 'task:end',
   TASK_WRAPUP: 'task:wrapup',
   TASK_REJECT: 'task:rejected',
+  TASK_SYNC: 'task:sync',
```
Kesari3008: Suggestion: Can we come up with some other event name that might be more appropriate to the purpose of the event? task: update probably or  something else.

packages/@webex/plugin-cc/test/unit/spec/services/task/TaskManager.ts
```diff
@@ -306,6 +306,20 @@ describe('TaskManager', () => {
...
+      data: {
+        ...initalPayload.data,
+        type: CC_EVENTS.AGENT_CONTACT,
+      },
+    };
+
+    const taskEmitSpy = jest.spyOn(taskManager, 'emit');
+    webSocketManagerMock.emit('message', JSON.stringify(payload));
+
+    expect(taskEmitSpy).toHaveBeenCalledWith(TASK_EVENTS.TASK_SYNC, taskManager.currentTask);
```
Kesari3008: We should also check the task collection as it gets updated with the task object right

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -567,6 +567,15 @@ describe('webex.cc', () => {
         {module: CC_FILE, method: 'stationReLogin'}
       );
     });
+
+    it('should trigger TASK_SYNC event with the task', () => {
```
Kesari3008: We should also add the check for on and off spy for this listener whenever regsiterIncomingcallEvents is invoked

Kesari3008 (changes requested):

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -65,6 +65,11 @@ export default class TaskManager extends EventEmitter {
...
+            this.taskCollection[payload.data.interactionId] = this.currentTask;
+            this.emit(TASK_EVENTS.TASK_SYNC, this.currentTask);
+            break;
```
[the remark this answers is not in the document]
rarajes2: Not required. It's consistent with all other cases.

docs/samples/contact-center/app.js
```diff
@@ -588,6 +588,30 @@ function register() {
...
+        enableTransferControls(); // Enable transfer controls
+
+        return;
+      }
+
+      if (state === 'connected') {
+        wrapupCodesDropdownElm.disabled = false;
+        wrapupElm.disabled = false;
+      }
+    });
```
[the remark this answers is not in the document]
rarajes2: Added the check. Else block is not required as code is changed.

packages/@webex/plugin-cc/src/cc.ts
rarajes2: Instead of invoking this in two different places which makes the code a bit error prone in my opinion. If keeping it simply in the constructor has same effect, we should keep it in only one place.

I am removing it from the `stationLogin` as well. Please let me know if this doesn't feel right and needs change

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
rarajes2: I didn't encounter anytime getting the list of tasks (inside the data field) in the `AgentContact` event. Had this discussion with Arun, we agreed upon if this case happens, we'll handle that later for now we can go ahead with this.

packages/@webex/plugin-cc/test/unit/spec/services/task/TaskManager.ts
rarajes2: Done

packages/@webex/plugin-cc/test/unit/spec/cc.ts
rarajes2: Done. Added in the logout test

packages/@webex/plugin-cc/src/cc.ts
rarajes2: Added the logic in the `stationLogout`.

packages/@webex/plugin-cc/src/services/task/types.ts
rarajes2: How about `task:hydrate` ? Updating with this name, let me know if you like something else.

rarajes2:
Creating a separate ticket for verifying the consult and transfer state. I have added the code in the samples app to handle the scenario but it behaves bit flaky. For widget work we don't have the consult and transfer feature available as of now so we can go ahead with this change and figure out this change in a separate task.

JIRA - https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-615561

rsarika:
@rarajes2  I remember below is the behaviour in agentdesktop as well(it disconnects). please check and make this as note as limitation for BROWSER case.

Tested with Browser Agent login - There is one issue with the mobius device registration where when we do the silentRelogin we get 403 error on first registration try, then DELETE request goes followed by device registration successful -- causing the call to disconnect and we get the wrap up state in place. This requires investigation in calling SDK. I will update this scenario in the test plan if it's not there already.


and did we test the scenario where customer ends before we relogin?

rsarika (approved):

packages/@webex/plugin-cc/src/services/task/types.ts
rsarika: hydrate is not straight forward i feel. probably task:reload ?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -79,11 +79,17 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
     this.trigger(TASK_EVENTS.TASK_INCOMING, task);
   };
 
+  private handleTaskSync = (task: ITask) => {
```
rarajes2: Rename to `handleTaskHydrate`

packages/@webex/plugin-cc/test/unit/spec/cc.ts
rarajes2: Change to TASK_HYDRATE

packages/@webex/plugin-cc/test/unit/spec/services/task/TaskManager.ts
```diff
@@ -298,14 +297,32 @@ describe('TaskManager', () => {
...
+      callOffSpy.mock.calls[0][1]
+    );
 
     taskManager.unregisterIncomingCallEvent();
     expect(offSpy.mock.calls.length).toBe(2); // 1 for incoming call and 1 for remote media
     expect(offSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.REMOTE_MEDIA, offSpy.mock.calls[0][1]);
     expect(offSpy).toHaveBeenCalledWith(LINE_EVENTS.INCOMING_CALL, offSpy.mock.calls[1][1]);
   });
 
+  it('should emit TASK_SYNC event on AGENT_CONTACT event', () => {
```
rarajes2: TASK_HYDRATE

mkesavan13 (approved):

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: Yeah, this approach is fine. Just ensure that everything is working as expected post these changes

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -567,6 +568,15 @@ describe('webex.cc', () => {
...
```
Kesari3008: Ensure to have the same name in all the places, whatever is decided for event name

packages/@webex/plugin-cc/src/cc.ts
rarajes2: Done

packages/@webex/plugin-cc/test/unit/spec/services/task/TaskManager.ts
rarajes2: Done

packages/@webex/plugin-cc/test/unit/spec/cc.ts
rarajes2: Done

Kesari3008 (approved):
Please check how we need to update the test plan with testcases to cover these changes

PR #4090, opened 2025-02-07
feat(media-helpers): update media helpers APIs

brycetham:
# COMPLETES [SPARK-610118](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610118)/[SPARK-607390](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-607390)

New functionality to `@webex/media-helpers` to aid in the creation of media streams.

The main changes are:

1. A new `createCameraAndMicrophoneStreams` API. This function allows users to create both camera and microphone streams in a single function call, without needing to call `createCameraStream` and `createMicrophoneStream` separately.

2. A new `createDisplayMedia` API. This function allows users to pass in additional options not currently possible with `createDisplayStream` or `createDisplayStreamWithAudio`. Many of these options are only available on Chrome/Edge (see https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia#options for details).

3. The ability to pass a `videoContentHint` to `createDisplayStream` or `createDisplayStreamWithAudio`. This tells the browser whether the stream will contain "motion", "detail", or "text". Browsers will use this to determine how to treat the shared content if insufficient resources are available (i.e. poor network) but actual results will depend on the browser.

Some examples:

1. Creating both camera and microphone streams:

- `[cameraStream, microphoneStream] = await createCameraAndMicrophoneStreams()`

2. Creating a display stream (analogous to calling `createDisplayStream`):

- `[displayStream] = await createDisplayMedia()` or
- `[displayStream] = await createDisplayMedia({video: {}})`

3. Creating both a display stream and a system audio stream (analogous to calling `createDisplayStreamWithAudio`):

- `[displayStream, systemAudioStream] = await createDisplayMedia({video: {}, audio: {}})`

4. Creating a display stream but limit the user to only sharing windows and tabs (Chrome/Edge only):

- `[displayStream] = await createDisplayMedia({video: {monitorTypeSurfaces: 'exclude'}})`

5. Creating both a display stream and a system audio stream but limit the user to only sharing audio from tabs (Chrome/Edge only):

- `[displayStream, systemAudioStream] = await createDisplayMedia({video: {}, audio: {systemAudio: 'exclude'}})`

6. Creating a display stream with content hint set to "motion":

- `displayStream = await createDisplayStream('motion')` or 
- `[displayStream] = await createDisplayMedia({video: {videoContentHint: 'motion'}})`

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

1. Manually tested by calling each individual API in the SDK sample app.
2. Added more robust unit tests to test different combinations of parameters for both existing and new APIs.

### I certified that

---


arun3528 (approved):
Didnt realize we are not exporting createCameraMicrophone steams as well , thanks for adding it

edvujic (approved):

szotrabh (approved):

antsukanova (approved):
lgtm

rsarika (approved):

packages/@webex/media-helpers/src/webrtc-core.ts
```diff
@@ -162,3 +154,43 @@ export const LocalCameraStream = AddEvents<typeof _LocalCameraStream, LocalCamer
...
+}) => wcmeCreateCameraAndMicrophoneStreams(LocalCameraStream, LocalMicrophoneStream, constraints);
+
+export const createDisplayStream = (videoContentHint?: VideoContentHint) =>
+  wcmeCreateDisplayStream(LocalDisplayStream, videoContentHint);
+
+export const createDisplayStreamWithAudio = (videoContentHint?: VideoContentHint) =>
+  wcmeCreateDisplayStreamWithAudio(LocalDisplayStream, LocalSystemAudioStream, videoContentHint);
+
+export const createDisplayMedia = (
+  options: {
```
marcin-bazyl, line 176: shouldn't the type for `options` also be defined and exported by webrtc-core?

packages/@webex/media-helpers/src/index.ts
```diff
@@ -14,8 +14,10 @@ export {
   LocalCameraStream,
   createMicrophoneStream,
   createCameraStream,
+  createCameraAndMicrophoneStreams,
```
marcin-bazyl, line 17: nitpick: feels like our naming between user media and display media is not very consistent. `createCameraAndMicrophoneStreams` is not consistent with `createDisplayStreamWithAudio` - one has "and", the other "with" and one expicitly says what streams are created the other just says "withAudio". Is there a plan to maybe deprecate `createDisplayStreamWithAudio` now that we have the new ` createDisplayMedia()`? If we have now `createDisplayMedia()` then maybe we should have `createUserMedia()` for mic and camera?

packages/@webex/media-helpers/src/webrtc-core.ts
brycetham: The type used in the `createDisplayMedia` API in `webrtc-core` is slightly different from this type in that it includes the class constructors under `options.video` and `options.audio`. So this specific type is not shared between `webrtc-core` and `media-helpers`.

packages/@webex/media-helpers/src/index.ts
brycetham: @bbaldino had originally suggested deprecating `createDisplayStream` and `createDisplayStreamWithAudio` but we ultimately decided to keep them as "simple" versions of the API for devs who are not interested in using any of the advanced options. In terms of why the original APIs were named differently, honestly I think it's mostly because `createDisplayStreamWithAudio` came first, and it felt weird to name the other API `createCameraStreamWithAudio`. But in my mind there is a slight difference that may justify the different name... you can create camera and microphone streams separately (and we have APIs for it) but it is not possible to create a `SystemAudioStream` without a `DisplayStream`.

brycetham: Also I'm not opposed to having a `createUserMedia` API but there aren't any additional options for `getUserMedia` like `getDisplayMedia` does to justify having another API, so we'd basically only be doing it for consistency.

bbaldino: @marcin-bazyl it's a fair point on the naming.  For what it's worth the display stuff is genuinely different: for example you can't get only an audio stream, so there's not really any concept of "just" a system audio stream, so I think that motivated some of those naming differences: you can only get a display stream or a display stream _with_ audio as well.  As opposed to main media where you can get audio, video or audio _and_ video.  But maybe they can be better/more consistent.

marcin-bazyl: ok, @brycetham @bbaldino fair enough, let's keep it different for main media and screenshare then.

marcin-bazyl (approved):

PR #4110, opened 2025-02-21
feat(plugin-meetings): brbState

SomeBody16:
# COMPLETES [SPARK-610065](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610065)

Users can quickly toggle the "step away" feature on and off through the web client. To prevent issues on the server side caused by processing too many rapid requests, it feels good to limit the amount of these requests. This logic could similar to how the mute/unmute microphone functionality works. 

Based on `MuteState` now there is `BrbState` that won't send request to server until first one finishes.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] Samples app tests
- [x] Unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,203 @@
+import {MediaType} from '@webex/internal-media-core';
+import LoggerProxy from '../common/logs/logger-proxy';
+import type Meeting from '.';
+import SendSlotManager from '../multistream/sendSlotManager';
+
+// eslint-disable-next-line import/prefer-default-export
```
antsukanova: I'm not sure if we need to repeat all patterns here from the mute state because mute state works a little bit differently from brb, I'm not sure if we need logic for `remoteBrbRequiresSync`, and also I see some methods that feel redundant here like `isEnabledRemotely`, `isEnabledLocally`

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,203 @@
...
+    `Meeting:brbState#createBrbState: creating BrbState for meeting id ${meeting?.id}`
+  );
+
+  const brbState = new BrbState(meeting, enabled);
+
+  return brbState;
+};
+
+/** The purpose of this class is to manage the local and remote brb state and make sure that the server state always matches
+   the last requested state by the client.
```
edvujic:
```suggestion
/** The purpose of this class is to manage the local and remote brb state and make sure that the server state always matches the last requested state by the client.
```

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,203 @@
...
+
+   This class is exported only for unit tests. It should never be instantiated directly with new BrbState(), instead createBrbState() should be called
```
edvujic: A bit confused here, this is exported because of unit tests, but then I can see an import for both `BrbState` and `createBrbState` in `packages/@webex/plugin-meetings/src/meeting/index.ts`. Is this comment correct?

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    it('can be disabled', async () => {
+      brbState.enable(false, meeting.sendSlotManager);
+      brbState.handleServerBrbUpdate(false);
+      await testUtils.flushPromises();
+
+      assert.isFalse(brbState.isEnabled());
+    });
+
+    it('won\t be able to send local brb state to server if not a multistream meeting', async () => {
```
edvujic: Nit: two suggestions

```suggestion
    it('won\'t be able to send local brb state to server if it is not a multistream meeting', async () => {
``` 
```suggestion
    it('will not be able to send local brb state to server if it is not a multistream meeting', async () => {
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3866,27 +3868,6 @@ describe('plugin-meetings', () => {
...
-            assert.notCalled(meeting.meetingRequest.setBrb);
-          })
-
-          it('should ignore disabling #beRightBack', async () => {
-            meeting.beRightBack(false);
-
-            assert.isRejected((Promise.reject()));
-            assert.notCalled(meeting.meetingRequest.setBrb);
-          })
-        });
```
edvujic, line 3890: Why did we remove the test for transcoded meetings?

edvujic (commented):
Overall looks good, left some comments. I agree with Anna that this should be a `fix` instead of a `feat`.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
brycetham: To add to this, mute is unique in that you can be remotely muted by another participant, but for brb, only you can set your own brb state. So there is not really a need to worry about the "remote" brb state (since such a concept does not really exist, at least not in the same as "remote" mute). We just need to be able to gracefully handle rapidly toggling the local brb state.

edit: Looking at it again, I think the code is mostly fine, the word "remote" is just kind of confusing since it's used in a different way for mute. We can consider renaming some stuff here to better fit the brb context.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  /**
+   * Enables/disables brb
+   *
+   * @param {boolean} enabled
+   * @param {SendSlotManager} sendSlotManager
+   * @returns {Promise}
+   */
+  public enable(enabled: boolean, sendSlotManager: SendSlotManager) {
```
brycetham, line 58: I noticed that for `muteState` the function signature is `enable(meeting: any, enable: boolean)`. I was just wondering if you considered passing in `meeting` instead, or if you felt passing in `sendSlotManager` would be better?

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -0,0 +1,112 @@
...
+      meeting.isMultistream = false;
+      brbState.enable(true, meeting.sendSlotManager);
+      brbState.handleServerBrbUpdate(true);
+      await testUtils.flushPromises();
+
+      assert.isTrue(meeting.meetingRequest.setBrb.notCalled);
+    });
+
+    it('won\t be able to send local brb state to server if webrtc media connection is not defined', async () => {
+      meeting.mediaProperties.webrtcMediaConnection = false;
```
brycetham: Do we really have cases in which the `webrtcMediaConnection` is false instead of undefined?

brycetham (approved):
This looks pretty good! I had a few questions but they are non-blocking. Of course, it would be good to address everyone else's comments also.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
SomeBody16: `sendSlotManager` is private in meeting that's why I'm passing it directly

packages/@webex/plugin-meetings/src/meeting/brbState.ts
SomeBody16: `remote` is more like "did server received and updated our state" and if not, we will repeat this request to make sure it gets updated

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
SomeBody16: This tests are now in brbState file instead as this checks are also there 
`won't be able to send local brb state to server if not a multistream meeting`

edvujic: Ah, got it. Thanks for clarifying.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * Updates the server local and remote brb values so that they match the current client desired state.
+   *
+   * @param {SendSlotManager} sendSlotManager
+   * @returns {Promise}
+   */
+  private applyClientStateToServer(sendSlotManager: SendSlotManager) {
+    if (this.state.syncToServerInProgress) {
```
antsukanova, line 71: What case causes `syncToServerInProgress` be `true` in our client flow?

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  const brbState = new BrbState(meeting, enabled);
+
+  return brbState;
+};
+
+/** The purpose of this class is to manage the local and remote brb state
+ * and make sure that the server state always matches the last requested state by the client.
+ */
+class BrbState {
```
antsukanova: I haven't seen export for this class, but you refer to it as a type and I see an issue in my IDE

Screenshot 2025-02-25 at 15 02 45

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,181 @@
+import {MediaType} from '@webex/internal-media-core';
+import LoggerProxy from '../common/logs/logger-proxy';
+import type Meeting from '.';
+import SendSlotManager from '../multistream/sendSlotManager';
+
+// eslint-disable-next-line import/prefer-default-export
+export const createBrbState = (meeting: Meeting, enabled: boolean) => {
```
antsukanova, line 6: Have you consideretd passing `sendSlotManager` as `createBrbState` param? Feels better for encapsulation in `BrbState`.

edvujic (approved):

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,181 @@
...
+      });
+  }
+
+  /**
+   * This method should be called whenever the server brb state is changed
+   *
+   * @param {Boolean} [brbEnabled] true if user has brb enabled, false otherwise
+   * @returns {undefined}
+   */
+  public handleServerBrbUpdate(brbEnabled?: boolean) {
```
antsukanova: feels right to align with other methods namings `brbEnabled` -> `enabled`

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Returns true if the user has locally or remotely brb enabled.
+   *
+   * @public
+   * @returns {Boolean}
+   */
+  public isEnabled(): boolean {
+    return this.state.client.enabled || this.state.server.enabled;
+  }
```
antsukanova: Where this function will be used?

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    it('can be disabled', async () => {
+      brbState.enable(false, meeting.sendSlotManager);
+      brbState.handleServerBrbUpdate(false);
+      await testUtils.flushPromises();
+
+      assert.isFalse(brbState.isEnabled());
+    });
+
+    it("won't be able to send local brb state to server if it is not a multistream meeting", async () => {
```
antsukanova: nit: let's have one time in test case naming like `does not send instead` of `won't be able to send` and so on

packages/@webex/plugin-meetings/src/meeting/brbState.ts
SomeBody16: I tried, but it causes issues as it can be undefined when creating brbState and initialized later. So instead I'm passing it when enabling so it's always most recent value that is used in meeting

packages/@webex/plugin-meetings/src/meeting/brbState.ts
brycetham: Ah got it, ok sounds like we don't really have a choice here, so that's fine.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
antsukanova: @SomeBody16 thanks for info, then let's keep current logic.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
SomeBody16: It's on line 94 in the same method, starting the process of appluyClientStateToServer will cause `syncToServerInProgress` to be true

antsukanova: @SomeBody16  could you explain what client flow will trigger this flow work?

SomeBody16: When user clicks brb
- enable(true, sendSlotManager)
- applyClientStateToServer
- if syncToServerInProgress, return
- if local state is the same as remote state, return
- **set syncToServerInProgress to true**
- `sendLocalBrbStateToServer`
- locus event will call `handleServerBrbUpdate`
- after `sendLocalBrbStateToServer` we are checking if locus updated brbState
- if it did, set syncToServerInProgress to false
- if not, will call `applyClientStateToServer` to make sure state will be updated and the process will start again

antsukanova: @SomeBody16 thanks! tested this with sdk, loogs good to me, appreciate detailed flow

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @param {SendSlotManager} sendSlotManager
+   * @returns {Promise}
+   */
+  private async sendLocalBrbStateToServer(sendSlotManager: SendSlotManager) {
+    const {enabled} = this.state.client;
+
+    if (!this.meeting.isMultistream) {
+      const errorMessage =
+        'Meeting:brbState#sendLocalBrbStateToServer --> Not a multistream meeting';
```
antsukanova: nit: how we define that here we will have "-->" in logs? 
And in other we are not have them

antsukanova: In `packages/@webex/plugin-meetings/src/meeting/muteState.ts` they all aligned, feels right to align this here as well

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -0,0 +1,114 @@
...
+    it('can be disabled', async () => {
+      brbState.enable(false, meeting.sendSlotManager);
+      brbState.handleServerBrbUpdate(false);
+      await testUtils.flushPromises();
+
+      assert.isFalse(brbState.state.client.enabled);
+      assert.isFalse(brbState.state.server.enabled);
+    });
+
+    it('cannot send local brb state to server if it is not a multistream meeting', async () => {
```
antsukanova: Nit: 
```suggestion
    it('does not send local brb state to server if it is not a multistream meeting', async () => {
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -0,0 +1,114 @@
...
+      meeting.isMultistream = false;
+      brbState.enable(true, meeting.sendSlotManager);
+      brbState.handleServerBrbUpdate(true);
+      await testUtils.flushPromises();
+
+      assert.isTrue(meeting.meetingRequest.setBrb.notCalled);
+    });
+
+    it('cannot send local brb state to server if webrtc media connection is not defined', async () => {
```
antsukanova: nit: a here as well

antsukanova (approved):
Small nits to handle, but overall logic lgtm, thanks for your work on that so far

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
antsukanova: `cannot` feels like function can't do this flow due to some errors, `does not` feels better to me

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3390,6 +3392,7 @@ export default class Meeting extends StatelessWebexPlugin {
     });
 
     this.locusInfo.on(LOCUSINFO.EVENTS.SELF_MEETING_BRB_CHANGED, (payload) => {
+      this.brbState.handleServerBrbUpdate(payload?.brb?.enabled);
```
marcin-bazyl: not sure if this is possible, but can Locus change our BRB state without us being in a meeting (so without Meeting.addMedia() being called and therefore with this.brbState being undefined)?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3615,40 +3618,7 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      })
-      .then(() => {
-        this.sendSlotManager.setSourceStateOverride(MediaType.VideoMain, enabled ? 'away' : null);
-      })
-      .catch((error) => {
-        LoggerProxy.logger.error('Meeting:index#beRightBack --> Error ', error);
-
-        return Promise.reject(error);
-      });
+    return this.brbState.enable(enabled, this.sendSlotManager);
```
marcin-bazyl, line 3656: if someone calls `beRightBack()` before calling addMedia(), `this.brbState` will be undefined. Previously this case was taken care of by the check for `this.mediaProperties.webrtcMediaConnection`.

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: Locus will notify us about self brb change only as a response to us changing brb so it should never be called with brbState being undefined, but just to make sure I've added `?` so it won't break anything

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: I've readded checks to this method for this case

marcin-bazyl (approved):

PR #4114, opened 2025-02-24
fix(sdk): fixed-join-link-parsing

adhmenon:
# COMPLETES #< [SPARK-468090](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-468090)>

* Right now, we have support for using joinLink to create meetings, for eg. `https://cisco.webex.com/cisco/j.php?MTID=m6dfc73927ea656a2f9b5637c32a2dac5`
* However, the SDK level prasing fails for certain cases, due to incorrect checks.

* Code has been added to improve the parsing.
* We simply check for the `joinLink` params in both pathname and search (ie MTID) to validate.
* Unit tests were added

### Vidcast
https://app.vidcast.io/share/9aa6185b-6d9a-41ee-8e4f-4fa1069d28e3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested with different types of join links which had `cisco` and non-cisco domains.
* Tested with bad links which did not have MTID and ensured meeting was not created.
* Ensured it did not break the current meeting link creation.

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -36,6 +36,9 @@ export const HOST = 'host';
 
 export const JOIN = 'join';
 
+export const JOIN_LINK = 'j.php';
+export const JOIN_LINK_MTID = 'MTID';
+
 export const LEAVE = 'leave';
 export const LIVE = 'live';
 export const LOCAL = 'local';
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -36,7 +36,6 @@ export const HOST = 'host';
 
 export const JOIN = 'join';
 
-export const JOIN_LINK = 'j.php';
 export const JOIN_LINK_MTID = 'MTID';
 
 export const LEAVE = 'leave';
--- a/packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
+++ b/packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
@@ -72,7 +72,7 @@ export default class MeetingInfoUtil {
           parsedUrl.pathname.includes(`/${MEET_CISCO}`) ||
           parsedUrl.pathname.includes(`/${MEET_CO}`) ||
           parsedUrl.pathname.includes(`/${JOIN}`) ||
-          parsedUrl.search.includes(JOIN_LINK_MTID));
+          (parsedUrl.search && parsedUrl.search.includes(JOIN_LINK_MTID)));
     }
 
     return hostNameBool && pathNameBool;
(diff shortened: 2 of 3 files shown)


packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -36,6 +36,9 @@ export const HOST = 'host';
 
 export const JOIN = 'join';
 
+export const JOIN_LINK = 'j.php';
```
adhmenon: Had to add this - no other way to check for the join link...

packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
```diff
@@ -70,7 +72,9 @@ export default class MeetingInfoUtil {
           parsedUrl.pathname.includes(`/${MEET_M}`) ||
           parsedUrl.pathname.includes(`/${MEET_CISCO}`) ||
           parsedUrl.pathname.includes(`/${MEET_CO}`) ||
-          parsedUrl.pathname.includes(`/${JOIN}`));
+          parsedUrl.pathname.includes(`/${JOIN}`) ||
+          (parsedUrl.pathname.includes(`/${JOIN_LINK}`) &&
+            parsedUrl.search.includes(JOIN_LINK_MTID)));
```
adhmenon: We will only check if both exists at the same time - hence the `&&`.

packages/@webex/plugin-meetings/src/constants.ts
mkesavan13: Were we able to get to the right team to verify this one?

packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
mkesavan13: Again, we need to confirm if MITD is the only type of query param there

packages/@webex/plugin-meetings/src/constants.ts
rsarika: yes. we need to confirm we always get j.php? any doc reference?

rarajes2:
What was the issue with the `Leave`? Were you able to fix that?

adhmenon:
> What was the issue with the `Leave`? Were you able to fix that?

I am still trying to find the root cause... seems to be something related to the remote media disconnect...

packages/@webex/plugin-meetings/src/constants.ts
adhmenon: https://developer.webex.com/docs/api/v1/meetings/join-a-meeting

adhmenon:
**Note**
Turns out we only need to check for `MTID` in the search parameter of the url. That is enough to determine the validity of the join link it seems.

packages/@webex/plugin-meetings/src/meeting-info/utilv2.ts
adhmenon: For `joinLink` this parameter will be there by default. Hence, we only need to check for this, have updated the code and added a comment also.

rsarika (approved):

mkesavan13 (approved):

PR #4115, opened 2025-02-25
fix(calling): set mediaEngine logger with webex logs for upload logs

rsarika:
# COMPLETES #[SPARK-616868](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-616868)

The uploaded logs from calling sdk to mats does not have Media layer(webrtc-media-core) logs.

called setLogger of Media layer to assign webex logger so that upload logs will have logs of media layer.

[webex-js-sdk_4288e710-ec5a-4895-adc5-80f339eec445.txt](https://github.com/user-attachments/files/19011779/webex-js-sdk_4288e710-ec5a-4895-adc5-80f339eec445.txt)
[webex-js-sdk_ea67d93f-7651-43c5-952c-16e36f86061a.txt](https://github.com/user-attachments/files/19011784/webex-js-sdk_ea67d93f-7651-43c5-952c-16e36f86061a.txt)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

login to calling sdk with token and make a call.

end the call and click upload logs. 

Expected: the logs should be uploaded to mats. and it should contain webrtc-media-core logs eg:  initiateOffer() called

### I certified that

---


sreenara (dismissed):
There needs to be a way to configure the level of the media core logs too. Can we ensure we provide a setting for this?

rsarika:
> There needs to be a way to configure the level of the media core logs too. Can we ensure we provide a setting for this?

@sreenara we are completely replacing the logger(using setLogger). so the level also gets effected in the media core logs. so in the console we see whatever we pass in config while initialising webex/callingclient.

But for uploading logs we upload complete buffer(irrespective of level) as per current logic in webex logger package.

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -120,6 +120,17 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
     this.mediaEngine = Media;
 
+    const adaptedLogger: Media.Logger = {
+      log: (...args) => webex.logger.log(args.join(' ')),
+      error: (...args) => webex.logger.error(args.join(' ')),
+      warn: (...args) => webex.logger.warn(args.join(' ')),
+      info: (...args) => webex.logger.info(args.join(' ')),
+      trace: (...args) => webex.logger.trace(args.join(' ')),
+      debug: (...args) => webex.logger.debug(args.join(' ')),
+    };
```
rarajes2: As discussed, Let's avoid space separated logs

packages/calling/src/common/Utils.ts
```diff
@@ -1554,8 +1554,10 @@ export function modifySdpForIPv4(sdp: string): string {
  * @param webex - Webex object to get service urls.
  * @param data - Data to be uploaded.
  */
-export async function uploadLogs(webex: WebexSDK, data = {}) {
+export async function uploadLogs(data = {}) {
   try {
+    const sdkConnector = SDKConnector;
+    const webex = sdkConnector.getWebex();
```
rarajes2: Do we have to create an extra variable here? Could be simply as 

```
const webex = SDKConnector.getWebex();
```

rsarika: addressed

packages/calling/src/CallingClient/CallingClient.ts
rsarika: media layer uses console log which prints space separated. so using the same.
attached the downloaded log for reference.

bhabalan (approved):
Please address the comments from others before merging

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: @sreenara @mkesavan13  What's your thought on this?

sreenara: @rarajes2 @rsarika I think we should have a meeting for this. I'm not sure what this discussion is about.
@rsarika where is the attached download log? I'm not able to find it.

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -120,6 +120,17 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
     this.mediaEngine = Media;
 
+    const adaptedLogger: Media.Logger = {
+      log: (...args) => webex.logger.log(args.join(' : ')),
+      error: (...args) => webex.logger.error(args.join(' : ')),
+      warn: (...args) => webex.logger.warn(args.join(' : ')),
+      info: (...args) => webex.logger.info(args.join(' : ')),
+      trace: (...args) => webex.logger.trace(args.join(' : ')),
+      debug: (...args) => webex.logger.debug(args.join(' : ')),
+    };
```
rarajes2, line 130: I would suggest using ` ; ` instead of ` : ` as delimiter.

` : ` may be confused with the `key : value` thing

rarajes2 (approved):
Approving with one comment. Please address that before merging.

PR #4124, opened 2025-02-28
fix(meetings): Meeting joined time header

chrisadubois:
# COMPLETES https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555657

add meeting joined time, read from header if exists, and set to system clock if not. Apply to CA metrics thereafter

- CA metrics
- read from header
- add setter method

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

manual,
unit

### I certified that

---


arun3528 (approved):

chrisadubois:
Waiting for response from CA and Locus on Date header

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1634,6 +1634,21 @@ export default class Meeting extends StatelessWebexPlugin {
     return this.#isoLocalClientMeetingJoinTime;
   }
 
+  /**
+   * Setter - sets isoLocalClientMeetingJoinTime
+   * This will be set once on meeting join, and not updated again
+   * @param {string | undefined} isoLocalClientMeetingJoinTime
+   */
+  set isoLocalClientMeetingJoinTime(isoLocalClientMeetingJoinTime: string | undefined) {
+    if (!this.#isoLocalClientMeetingJoinTime) {
```
marcin-bazyl: question (just to double check): in a scenario where user joins, leaves and then joins the same meeting again, are we supposed to still use the fist meeting join time and not the 2nd one?

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -180,7 +180,7 @@ const MeetingUtil = {
       .then((res) => {
         const parsed = MeetingUtil.parseLocusJoin(res);
         meeting.setLocus(parsed);
-
+        meeting.isoLocalClientMeetingJoinTime = res?.headers?.date; // read from header if exist, else fall back to system clock : https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555657
```
marcin-bazyl, line 183: wouldn't it be better to add this to the common code that's always executed when any Locus is parsed? I think we are not guaranteed to have the join time in the http response here, it might come in a delta DTO afterwards. Actually if we always parse it out of Locus DTO then we would end up setting it also in the case when we just load the page and Locus thinks we're already in the meeting, because user was in a meeting and just did a page refresh, so maybe it is better to keep it here as it is and my comment is useless - just making this comment to make sure that all these scenarios were considered.

Also just wondering if there are any other means of ending up in the meeting where we should set the joined time, like maybe for example Meeting.moveFrom()?

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: You are right in calling this out, this is a problem. I made the mistake of not recalling that the meeting object is not always destroyed on leave (in the case of OBTP or meeting ongoing) and was only considering the guest case. I will need to make adjustments for this. Ill let you know once done, I'm still waiting for locus to make their change.

packages/@webex/plugin-meetings/src/meeting/util.ts
chrisadubois: I think this is maybe a misunderstanding in how we are supposed to be "setting meeting joined time" according to the spec, we are *only* supposed to set `meetingJoinedTime` when we receive a locus join response (from `/participant`, or `/call` request). The time is not set in the DTO or anything else, and it's not the system clock. the request is to use the value exposed in the `Date` header field from the join response. We would *not* want the one that comes in a delta DTO afterwards. In order to support this, Locus has added `Date` to `Access-Control-Expose-Headers`.

I *believe* all the "join" functions route through this location. At least every one that I checked did (for example `moveFrom` will also end up here)

marcin-bazyl (approved):

PR #4126, opened 2025-03-03
feat(meeting): spark-623318 return correct info when not reach JBH

JudyZhuHz:
…ng since not reach JBH

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-623318

Return failure reason as JOIN_ERROR_REASN_FOR_END_USER when meet 403003(not reach JBH time) error

< DESCRIBE YOUR CHANGES >
Add new failed reason for 403 error. 
Testing in local env, it's works
Cantina PR: https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/7712

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -203,6 +203,7 @@ export const ICE_AND_DTLS_CONNECTION_TIMEOUT = 20000;
 export const ROAP_OFFER_ANSWER_EXCHANGE_TIMEOUT = 35000;
 export const WEBINAR_ERROR_WEBCAST = [403026];
 export const WEBINAR_ERROR_REGISTRATIONID = [403037, 403137];
+export const JOIN_ERROR_REASON_FOR_END_USER = [403003];
```
marcin-bazyl: the name seems very generic - what does it mean "join error reason for end user"? can we have some more meaningful name here?

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -1333,6 +1334,7 @@ export const MEETING_INFO_FAILURE_REASON = {
   WEBINAR_REGISTRATION: 'WEBINAR_REGISTRATION', // webinar need registration
   NEED_JOIN_WITH_WEBCAST: 'NEED_JOIN_WITH_WEBCAST', // webinar need using webcast join
   WEBINAR_NEED_REGISTRATIONID: 'WEBINAR_NEED_REGISTRATIONID', // webinar need registrationID
+  NOT_REACH_JBH: 'NOT_REACH_JBH', // Meeting is not allow to access since not reach JBH time
```
marcin-bazyl: it would be good to explain what JBH means in the comment here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1790,6 +1791,18 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        // Handle the case where user hasn't reached Join Before Host (JBH) time (error code 403003)
+        if (JOIN_ERROR_REASON_FOR_END_USER.includes(err.wbxAppApiCode)) {
+          this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NOT_REACH_JBH;
+          this.meetingInfoFailureCode = err.wbxAppApiCode;
+
+          if (err.meetingInfo) {
+            this.meetingInfo = err.meetingInfo;
+          }
+
+          throw new JoinWebinarError();
```
marcin-bazyl: why are we throwing a webinar error here? this is not a webinar only specific scenario, it can happen for any meeting, right?

packages/@webex/plugin-meetings/src/constants.ts
JudyZhuHz: Do you have some suggestion?
I think is return a string that needs to show the specific error reason (just display error reason, don't need any other actions), such as JBH case

packages/@webex/plugin-meetings/src/constants.ts
JudyZhuHz: Sure.

packages/@webex/plugin-meetings/src/meeting/index.ts
JudyZhuHz: Yes, but in plugin-meetings/src/meetings/index.ts, only check the CaptchaError,PasswordError,PermissionError,JoinWebinarError. Or add a new error - JoinMeetingMessageError?

packages/@webex/plugin-meetings/src/constants.ts
marcin-bazyl: we can't be displaying the error reason that backend sends us - it won't be localized to user's language

packages/@webex/plugin-meetings/src/meeting/index.ts
JudyZhuHz: Add JoinForbiddenError for this case.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1784,6 +1787,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        if (err.meetingInfo) {
+          this.meetingInfo = err.meetingInfo;
+        }
+
+        // Handle the case where user hasn't reached Join Before Host (JBH) time (error code 403003)
+        if (JOIN_BEFORE_HOST === err.wbxAppApiCode) {
+          this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NOT_REACH_JBH;
+        }
+
+        throw new JoinForbiddenError();
```
marcin-bazyl: maybe it would be good to pass here into JoinForbiddenError the more specific information about the error - the fact that it's JBH so that the app knows how to react to it. I guess you're doing it by just setting meetingInfoFailureReason above, but when an app tries to join a meeting it can also receive this JoinForbiddenError() error and then how would it know that for this error it needs to check meetingInfoFailureReason - the app shouldn't need to know that, it's SDK's implementation detail that we have to get meeting info from app api to be able to join

JudyZhuHz: Thanks for the reminder, updated.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1784,6 +1787,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        if (err.meetingInfo) {
+          this.meetingInfo = err.meetingInfo;
+        }
+
+        // Handle the case where user hasn't reached Join Before Host (JBH) time (error code 403003)
+        if (JOIN_BEFORE_HOST === err.wbxAppApiCode) {
+          this.meetingInfoFailureReason = MEETING_INFO_FAILURE_REASON.NOT_REACH_JBH;
+        }
+
+        throw new JoinForbiddenError(this.meetingInfoFailureReason, err.wbxAppApiCode, err);
```
marcin-bazyl: in JoinForbiddenError constructor, the code is initialized to `ERROR_DICTIONARY.JoinForbiddenError.CODE` (value 18), while here it is set to wbxAppApiCode which is a totally different number/range - it's like a different type. I'm not sure if it's a good idea to mix them, although at least both are numbers and the will never overlap because wbx app api codes are much bigger, so maybe it's OK? Otherwise we would need to add a new field like `wbxAppApiCode` to JoinForbiddenError.

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
JudyZhuHz: yes, it's good, add wbxAppApiCode.

mickelr (approved):

PR #4127, opened 2025-03-03
feat(cc-metrics): add MetricsManager

bhabalan:
# COMPLETES #[SPARK-572789](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-572789)

## This pull request addresses [SPARK-572789](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-572789)

Add behavioral and operational metrics for cc sdk

Added new MetricsManager class

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Emitted a behavioral event and verified it in Amplitude

### I certified that

---


bhabalan:
@coderabbitai review

packages/@webex/plugin-cc/package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "@types/platform": "1.3.4",
     "@webex/calling": "workspace:*",
-    "@webex/internal-plugin-mercury": "workspace:*",
```
arun3528, line 28: why removing mercury here ?

packages/@webex/plugin-cc/src/MetricsManager.ts
```diff
@@ -0,0 +1,47 @@
...
+  private static instance: MetricsManager;
+
+  private constructor({webex}: {webex: WebexSDK}) {
+    this.webex = webex;
+  }
+
+  public trackBehavioralMetric(eventName: eventNames, data?: EventPayload) {
+    if (this.webex?.internal?.newMetrics) {
+      this.webex.internal.newMetrics.submitBehavioralEvent({
+        product: 'wxcc_desktop',
```
arun3528: do you need a wrapper , can you check the web client if they do it based on config for adding default values ?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -201,6 +204,11 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
 
       await loginResponse;
 
+      this.metricsManager.trackBehavioralMetric('agent-login', {
```
arun3528: all constnt to config file

packages/@webex/plugin-cc/src/services/agent/index.ts
```diff
@@ -77,6 +85,14 @@ export default function routingAgent(routing: AqmReqs) {
...
         },
         errId: 'Service.aqm.agent.stationLoginFailed',
+        err(e) {
+          metricsManager.trackBehavioralMetric('agent-login', {
+            ...p.data,
+            isSuccess: false,
+            roles: Array.isArray(p.data.roles) ? p.data.roles.join(',') : '',
+            error: JSON.stringify(e),
+          });
+        },
```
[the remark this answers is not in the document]
bhabalan: Invalid suggestion

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -114,8 +115,9 @@ interface IWebexInternal {
       contactsService: string;
     };
   };
-  metrics: {
-    submitClientMetrics: (name: string, data: unknown) => void;
+  newMetrics: {
+    submitBehavioralEvent: SubmitBehavioralEvent;
+    submitOperationalEvent: SubmitOperationalEvent;
   };
```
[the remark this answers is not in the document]
bhabalan: Out of scope for this PR

packages/@webex/plugin-cc/src/MetricsManager.ts
bhabalan: As discussed, keep this to make events specific to cc so that it can be passed around instead of the webex object

bhabalan:
@coderabbitai re-review and mark comments as resolved if they've been addressed

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+type GenericEvent = {
+  name: string;
+  payload: EventPayload;
+};
+
+export type MetricsType = 'behavioral' | 'operational' | 'business';
+
+export default class MetricsManager {
+  private webex: WebexSDK;
+  private readonly runningEvents: Map<string, number> = new Map<string, number>();
```
sreenara: I recall we used `Record` in the calling package instead of `Map`. Can we check why and see if we can be consistent?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+  private pendingBehavioralEvents: BehavioralEvent[] = [];
+  private pendingOperationalEvents: GenericEvent[] = [];
+  private pendingBusinessEvents: GenericEvent[] = [];
+  private readyToSubmitEvents = false;
+  private submittingEvents = false; // Add a lock for submitting events
+
+  // eslint-disable-next-line no-use-before-define
+  private static instance: MetricsManager;
```
sreenara, line 35: question. Is it necessary to declare an instance within the class itself? Or are we doing this to avoid a global instance?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+      payloadWithCommonMetadata.mem_js_heap_size_limit = memory.jsHeapSizeLimit;
+      payloadWithCommonMetadata.mem_total_js_heap_size = memory.totalJSHeapSize;
+      payloadWithCommonMetadata.mem_used_js_heap_size = memory.usedJSHeapSize;
+    }
+
+    return payloadWithCommonMetadata;
+  }
+
+  private isMetricsDisabled(): boolean {
+    // TODO: Need to return true if in development mode to avoid sending metrics to the server
```
sreenara: Is there a story for this?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
+import {EventPayload} from '@webex/internal-plugin-metrics/src/metrics.types';
+
+import {WebexSDK} from '../types';
+import {BehavioralEventTaxonomy, getEventTaxonomy} from './behavioral-events';
+import LoggerProxy from '../logger-proxy';
+
+type MemoryInfo = {
```
sreenara: Why are we tracking this info? Will this be present in every event emitted?

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export type BehavioralEventTaxonomy = {
+  product: MetricEventProduct;
+  agent: MetricEventAgent;
+  target: string;
+  verb: MetricEventVerb;
+};
+
+export function getEventTaxonomy(name: string): BehavioralEventTaxonomy | undefined {
+  // Adding new metrics? Please add them to the wiki: https://confluence-eng-gpk2.cisco.com/conf/x/e_-_JQ
```
sreenara: We should avoid internal confluence docs in the code. Perhaps we can document this comment elsewhere?

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
```diff
@@ -0,0 +1,78 @@
...
+import {PRODUCT_NAME} from '../constants';
+
+export type BehavioralEventTaxonomy = {
+  product: MetricEventProduct;
+  agent: MetricEventAgent;
+  target: string;
+  verb: MetricEventVerb;
+};
+
+export function getEventTaxonomy(name: string): BehavioralEventTaxonomy | undefined {
```
sreenara: Why is `name` a string when we seem to have clear values for it?

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  switch (name) {
+    case METRICS.EVENT_NAMES.STATION_LOGIN:
+      return {
+        product,
+        agent: 'user',
+        target: 'station',
+        verb: 'login',
+      };
+    case METRICS.EVENT_NAMES.CALL_COMPLETED:
```
sreenara: Are these examples or the final events we are proposing? If it's not final, I'll skip this and review future PRs

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    if (typeof window === 'undefined') {
+      return payload;
+    }
+
+    const payloadWithCommonMetadata = {...payload};
+    payloadWithCommonMetadata.tabHidden = document.hidden;
+    if (window.performance && 'memory' in window.performance) {
+      const memory = window.performance.memory as MemoryInfo;
+      payloadWithCommonMetadata.mem_js_heap_size_limit = memory.jsHeapSizeLimit;
```
sreenara: Won't this become too much information? What's the benefit of having this?

packages/@webex/plugin-cc/src/metrics/constants.ts
```diff
@@ -0,0 +1,15 @@
+const METRICS = () => {
```
sreenara: question. Any reason this is a function that returns an object instead of just having a bunch of contants?

packages/@webex/plugin-cc/src/services/agent/index.ts
```diff
@@ -58,12 +60,24 @@ export default function routingAgent(routing: AqmReqs) {
       url: '/v1/agents/login',
       host: WCC_API_GATEWAY,
       data: p.data,
-      err: /* istanbul ignore next */ (e: any) =>
-        new Err.Details('Service.aqm.agent.stationLogin', {
+      err: /* istanbul ignore next */ (e: any) => {
+        metricsManager.trackEvent(
```
sreenara, line 66: question. This is just an example right? We're not going to send the entire `p.data` to metrics server in the final code?

packages/@webex/plugin-cc/package.json
sreenara: Please reply to this comment

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+  private readyToSubmitEvents = false;
+  private submittingEvents = false; // Add a lock for submitting events
+
+  // eslint-disable-next-line no-use-before-define
+  private static instance: MetricsManager;
+
+  private constructor({webex}: {webex: WebexSDK}) {
+    this.webex = webex;
+
+    if (this.webex.ready) {
```
sreenara: we aren't testing these conditions

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+    });
+  }
+
+  private setReadyToSubmitEvents() {
+    this.readyToSubmitEvents = true;
+    this.submitPendingEvents();
+  }
+
+  private async submitPendingEvents() {
+    if (this.submittingEvents) {
```
sreenara, line 47: here as well. I don't see tests for this and other similar places where we have exit conditions on the submit functions

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  private async submitPendingBehavioralEvents() {
+    if (this.pendingBehavioralEvents.length === 0) {
+      return;
+    }
+    if (this.readyToSubmitEvents) {
+      const eventsToSubmit = [...this.pendingBehavioralEvents];
+      this.pendingBehavioralEvents = [];
+      eventsToSubmit.forEach((event) => {
+        this.webex?.internal?.newMetrics?.submitBehavioralEvent({
```
sreenara: We should add tests where `webex`, `internal` and `newMetrics` are undefined. Applies in all other places

sreenara (changes requested):
Please see the comments. Is there a coverage report in the actions?

packages/@webex/plugin-cc/package.json
bhabalan: Will revert this change as it's not in the scope of this ticket

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: SPARK-637285

sreenara (changes requested):
Please add the behavioral event payload to the PR description for reference.
I took a look at the amplitude chart. Do we need fields like:
browserheight, browserwidth, dataeventtype (which is AgentDesktopMessage), iniframe, tabhidden

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Tracked the same to make see if the process got killed. Similar info tracked in cc desktop. Can remove if not needed

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Yes. So that this can be referred as MetricsManager.instance. Similar pattern followed in other places inside cc
eg: https://github.com/webex/webex-js-sdk/blob/49c76aacf427049b733518e96f6570fdfa283004/packages/%40webex/plugin-cc/src/services/task/TaskManager.ts#L19

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Updated to use Record

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
bhabalan: Removed

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
bhabalan: Using enums now here

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
bhabalan: Example events. Final event names will be published in the confluence for review

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: This is just 3 extra fields. CC Desktop is also tracking these. Gives an insight on available memory and usage to reproduce scenarios with devtools

packages/@webex/plugin-cc/src/metrics/constants.ts
bhabalan: Refactored to be an ENUM now

packages/@webex/plugin-cc/src/services/agent/index.ts
bhabalan: Correct. Infact, will be removing this in favor tracking this in the catch block instead of in AQM

bhabalan:
> Please add the behavioral event payload to the PR description for reference. I took a look at the amplitude chart. Do we need fields like: browserheight, browserwidth, dataeventtype (which is AgentDesktopMessage), iniframe, tabhidden

These come from newMetrics package automatically. I'll try to make these as optional as this shouldn't be required for an SDK
dataeventtype <--- From that particular dummy metric code. This will go away in future.

tabhidden <— I'm tracking it so that we can determine tab is in background(same as CC desktop)

Attaching here the HAR file for metrics payload
[MetricsManager all types of metrics.har.zip](https://github.com/user-attachments/files/19332843/MetricsManager.all.types.of.metrics.har.zip)

Here is the current coverage report. This should improve as I fix the failing mockWebex based tests 
image

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Removed these conditionals as we always submit only when webex.ready or webex.once('ready') happens.

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Done

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Done

bhabalan:
After [adding more tests](https://github.com/webex/webex-js-sdk/pull/4127/commits/31b97e5b00ec53c258d3eb4fe24a06657d4ecc68)
image

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
sreenara: @bhabalan I don't think we need to track this since it's going to be loaded on customer applications

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,327 @@
...
+      const data: EventPayload = {key: 'value'};
+
+      metricsManager.setMetricsDisabled(true);
+      metricsManager.trackEvent(eventName, data, ['behavioral', 'operational', 'business']);
+
+      expect(webex.internal.newMetrics.submitBehavioralEvent).toHaveBeenCalledTimes(0);
+      expect(webex.internal.newMetrics.submitOperationalEvent).toHaveBeenCalledTimes(0);
+      expect(webex.internal.newMetrics.submitBusinessEvent).toHaveBeenCalledTimes(0);
+    });
+    it('should not submit a behavioral event if metricsDisabled is true', () => {
```
sreenara, line 88: Please add a new line between tests

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,327 @@
...
+      const eventName = METRIC_EVENT_NAMES.STATION_LOGIN;
+      const data: EventPayload = {key: 'value'};
+
+      metricsManager.setMetricsDisabled(true);
+      metricsManager.trackBehavioralEvent(eventName, data);
+
+      expect(webex.internal.newMetrics.submitBehavioralEvent).toHaveBeenCalledTimes(0);
+    });
+    it('should not submit a operational event if metricsDisabled is true', () => {
```
sreenara, line 98: same here and all remaining tests. new line please.

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,327 @@
...
+      const data: EventPayload = {key: 'value'};
+
+      metricsManager.trackEvent(eventName, data, ['nonexistent']);
+
+      expect(webex.internal.newMetrics.submitBehavioralEvent).toHaveBeenCalledTimes(0);
+      expect(webex.internal.newMetrics.submitOperationalEvent).toHaveBeenCalledTimes(0);
+      expect(webex.internal.newMetrics.submitBusinessEvent).toHaveBeenCalledTimes(0);
+    });
+
+    it('should not submit a behavioral, operational and business event if metricsDisabled is true', () => {
```
sreenara, line 76: The next 3 tests seem to be a duplicate of this test? What's the difference between the tests?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  public setMetricsDisabled(disabled: boolean) {
+    this.metricsDisabled = disabled;
+    if (disabled) {
+      this.clearPendingEvents();
+    }
+  }
+
+  private clearPendingEvents() {
```
sreenara, line 168: Should we use `this.pendingBehavioralEvents.length = 0` instead?
Please see this discussion: https://stackoverflow.com/questions/1232040/how-do-i-empty-an-array-in-javascript

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+
+  private async submitPendingBusinessEvents() {
+    if (this.pendingBusinessEvents.length === 0) {
+      return;
+    }
+    if (this.readyToSubmitEvents) {
+      const eventsToSubmit = [...this.pendingBusinessEvents];
+      this.pendingBusinessEvents = [];
```
sreenara: same point here about setting the length to 0. I feel that's cleaner. Applies to the other submit functions too

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -493,7 +534,7 @@ describe('webex.cc', () => {
         details: {
           trackingId: '1234',
           data: {
-            reason: 'Error while performing station login',
+            reason: 'Error while performing stationLogin',
```
sreenara, line 537: question. Why have we changed this to one word?

sreenara (changes requested):
Please see the comments.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
bhabalan: This is comparing to the output of `getErrorDetails(error, 'stationLogin', CC_FILE)` ,
See the implementation of getErrorDetails here: https://github.com/webex/webex-js-sdk/blob/48133278312245cb3ae130940188de7a19fdce63/packages/%40webex/plugin-cc/src/services/core/Utils.ts#L13

So this can't be two words. Not sure how this was passing earlier.

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
bhabalan: The implementation of MetricsManager.trackEvent() has a guard condition that does not even invoke trackBehavioral, trackOperational, trackFunctional methods. If we want to test this branch, then this test has to be there. 
The other three tests check the individual methods for this branch. This ensures all the branches are covered.

Refer to the implementation here: https://github.com/bhabalan/webex-js-sdk/blob/a0df91d92989d1a0d13da0a33f9df03f37db6e9b/packages/%40webex/plugin-cc/src/metrics/MetricsManager.ts#L239

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Done

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: done

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
bhabalan: Done

packages/@webex/plugin-cc/test/unit/spec/metrics/MetricsManager.ts
bhabalan: Done

sreenara (approved):
LGTM. Thanks.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -212,13 +217,31 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.STATION_LOGIN,
+        {
+          isSuccess: true,
+          loginType: data.loginOption,
+          agentId: resp.data.agentId,
+          teamId: resp.data.teamId,
+          siteId: resp.data.siteId,
+          roles: resp.data.roles.join(','),
```
rsarika, line 230: roles is an optional value it might break the flow if we dont get roles in the response. Similarly can we check all other fields as well?

packages/@webex/plugin-cc/src/services/agent/index.ts
```diff
@@ -5,14 +5,16 @@ import AqmReqs from '../core/aqm-reqs';
...
+import {METRIC_EVENT_NAMES} from '../../metrics/constants';
 
 /*
  * routingAgent
  * @param reqs
  * @category Routing Service
  */
 
-export default function routingAgent(routing: AqmReqs) {
+export default function routingAgent(routing: AqmReqs, metricsManager: MetricsManager) {
```
rsarika: why do we need to pass it from here? we can directly access MetricsManager.getInstance right?

rsarika (approved):

packages/@webex/plugin-cc/src/services/agent/index.ts
bhabalan: Good point. I have made webex as an optional parameter now

packages/@webex/plugin-cc/src/cc.ts
bhabalan: This is a dummy event in this PR. Will take care of it in the actual cc.ts metrics PR

bhabalan:
@sreenara , I've made webex as an optional instance for MetricsManager.getInstance() for cases when webex isn't available

sreenara (approved):

PR #4135, opened 2025-03-06
fix(plugin-meetings): extra logging of roap messages progress

k-wasniowski:
# COMPLETES #[SPARK-629266](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-629266)

This pull request will add extra logging of the progress in sending SDP Offer to the server. Sometimes we are observing it can take a long time to send, with this change we want to get better understanding where the delays occur.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested manually on JS-SDK Sample app

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
@@ -304,18 +304,12 @@ export class LocusMediaRequest extends WebexPlugin {
       });
 
     if (request.type === 'RoapMessage') {
-      const progressLogger = (direction: string, progressEvent: ProgressEvent) => {
-        LoggerProxy.logger.info(
-          `${request.type}: ${direction} Progress, Timestamp: ${progressEvent.timeStamp}, Progress: ${progressEvent.loaded}/${progressEvent.total}`
-        );
-      };
-
       const setupProgressListener = (direction: string, eventEmitter: EventEmitter) => {
-        const shunt = new EventEmitter();
-        transferEvents('progress', eventEmitter, shunt);
-        shunt.on('progress', (progressEvent: ProgressEvent) =>
-          progressLogger(direction, progressEvent)
-        );
+        eventEmitter.on('progress', (progressEvent: ProgressEvent) => {
+          LoggerProxy.logger.info(
+            `${request.type}: ${direction} Progress, Timestamp: ${progressEvent.timeStamp}, Progress: ${progressEvent.loaded}/${progressEvent.total}`
+          );
+        });
       };
 
       setupProgressListener('Upload', options.upload);
--- a/packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
@@ -1,6 +1,7 @@
 /* eslint-disable valid-jsdoc */
 import {defer} from 'lodash';
-import {Defer} from '@webex/common';
+import {Defer, transferEvents} from '@webex/common';
+import {EventEmitter} from 'events';
 import {WebexPlugin} from '@webex/webex-core';
 import {MEDIA, HTTP_VERBS, ROAP} from '../constants';
 import LoggerProxy from '../common/logs/logger-proxy';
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -294,6 +302,25 @@ export class LocusMediaRequest extends WebexPlugin {
...
+    const progressLogger = (direction: string, progressEvent: ProgressEvent) => {
+      LoggerProxy.logger.info(
+        `${request.type}: ${direction} Progress, Timestamp: ${progressEvent.timeStamp}, Progress: ${progressEvent.loaded}/${progressEvent.total}`
+      );
+    };
+
+    const setupProgressListener = (direction: string, eventEmitter: EventEmitter) => {
+      const shunt = new EventEmitter();
+      transferEvents('progress', eventEmitter, shunt);
+      shunt.on('progress', (progressEvent: ProgressEvent) =>
```
marcin-bazyl: why do we need `shunt`? can't we just listen for `progress` event on the `eventEmitter` directly?

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -250,12 +251,19 @@ export class LocusMediaRequest extends WebexPlugin {
...
-    return this.request({
+    const upload = new EventEmitter();
+    const download = new EventEmitter();
+
+    const options = {
       method: HTTP_VERBS.PUT,
       uri,
       body,
-    })
+      upload,
```
marcin-bazyl, line 261: maybe we should do this extra logging only when `(request.type === 'RoapMessage')`, because LocusMediaRequest is also used for mute operations and these won't be interesting to us and can fill up the log

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
k-wasniowski: I guess we could, decided to keep it consistent with other places that are listening on progress

k-wasniowski: If you think we should skip it here, and listen directly I can remove shunt

marcin-bazyl: if it works without it, then yeah I would rather remove it, just because it can be a bit confusing and others will keep repeating the same pattern when they see it, less code is always better than more :-)

k-wasniowski:
Working on the UT.

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
k-wasniowski: Looks like it works, removed it then

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -294,6 +302,21 @@ export class LocusMediaRequest extends WebexPlugin {
...
+    if (request.type === 'RoapMessage') {
+      const setupProgressListener = (direction: string, eventEmitter: EventEmitter) => {
+        eventEmitter.on('progress', (progressEvent: ProgressEvent) => {
+          LoggerProxy.logger.info(
+            `${request.type}: ${direction} Progress, Timestamp: ${progressEvent.timeStamp}, Progress: ${progressEvent.loaded}/${progressEvent.total}`
+          );
+        });
+      };
+
+      setupProgressListener('Upload', options.upload);
```
marcin-bazyl, line 315: is there any reason why we're setting up these listeners after we called `this.requet()` and not before? Doing it later like this risks that we may miss some events

k-wasniowski: It's not working when we do it before a call to the `this.request()`. Not sure why though.

marcin-bazyl: ok, fine, we'll look into it only if we find cases where we don't get any progress events at all

marcin-bazyl (approved):

szotrabh (approved):

PR #4141, opened 2025-03-07
fix(plugin-meetings): stop reachability when joining a meeting

marcin-bazyl:
# COMPLETES #[SPARK-629490](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-629490)

If reachability is till being done in the background when we create a media connection, gathering of local candidates is very slow on Firefox, this delays meeting join and sometimes even causes Homer to timeout and join to fail.

Stopping reachability when we create a media connection, this is after we've already sent the reachability results to the backend, so continuing reachability doesn't add value. 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -6719,6 +6719,9 @@ export default class Meeting extends StatelessWebexPlugin {
         new RtcMetrics(this.webex, {meetingId: this.id}, this.correlationId)
       : undefined;
 
+    // ongoing reachability checks slow down new media connections especially on Firefox, so we stop them
+    this.getWebexObject().meetings.reachability.stopReachability();
+
     const mc = Media.createMediaConnection(
       this.isMultistream,
       this.getMediaConnectionDebugId(),
--- a/packages/@webex/plugin-meetings/src/reachability/index.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/index.ts
@@ -260,6 +260,32 @@ export default class Reachability extends EventsScope {
   }
 
   /**
+   * Stops all reachability checks that are in progress
+   * @public
+   * @memberof Reachability
+   * @returns {void}
+   */
+  public stopReachability() {
+    // overallTimer is always there only if there is reachability in progress
+    if (this.overallTimer) {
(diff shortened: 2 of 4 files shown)


k-wasniowski (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6719,6 +6719,9 @@ export default class Meeting extends StatelessWebexPlugin {
         new RtcMetrics(this.webex, {meetingId: this.id}, this.correlationId)
       : undefined;
 
+    // ongoing reachability checks slow down new media connections especially on Firefox, so we stop them
+    this.getWebexObject().meetings.reachability.stopReachability();
```
sreenara, line 6723: Quick question. Why do we have `this.getWebexObject()` here but I'm seeing instances of `this.webex` in other places, for example the line above (6719).

adhmenon (commented):
Changes LGTM.
Small doubt, since we are stopping the checks - let's assume some more resilient candidate pairs arrive slowly (edge case) - in that scenario, since we stop the connectivity checks, we don't get these candidates and essentially we will essentially be using a connection which is less optimal. Would this cause any major latency issues? Curious to know...

sreenara (approved):
Just one question about the way we're getting `webex` object.

adhmenon (approved):

k-wasniowski:
> Changes LGTM. Small doubt, since we are stopping the checks - let's assume some more resilient candidate pairs arrive slowly (edge case) - in that scenario, since we stop the connectivity checks, we don't get these candidates and essentially we will essentially be using a connection which is less optimal. Would this cause any major latency issues? Curious to know...

Once we do a join, we would have provide reachability results gathered up to the sending point of the media request. All pairs gathered later was just for information, and was not signaled to the Locus.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: it's really the same thing, I think getWebexObject() was added later, so some newer code calls it, older code uses this.webex. TBH, I'm not sure why the function was added in the first place

PR #4156, opened 2025-03-18
feat(plugin-meetings): reachability- add detection of symmetric NAT

k-wasniowski:
# COMPLETES #[SPARK-637849](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-637849)

This PR introduces possibility of finding out if customer is behind a symmetric NAT during the reachability.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---


brycetham (approved):

szotrabh (approved):

edvujic (approved):

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -290,6 +297,46 @@ export class ClusterReachability extends EventsScope {
...
+   * Determines NAT Type.
+   *
+   * @param {RTCIceCandidate} candidate
+   * @returns {void}
+   */
+  private determineNatType(candidate: RTCIceCandidate) {
+    this.srflxIceCandidates.push(candidate);
+
+    if (this.srflxIceCandidates.length > 1) {
+      const portsFound: {[key: string]: Set<number>} = {};
```
antsukanova: Why not `Record<string, Set<number>>`?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -290,6 +297,46 @@ export class ClusterReachability extends EventsScope {
...
+        if (!portsFound[key]) {
+          portsFound[key] = new Set();
+        }
+        portsFound[key].add(c.port);
+      });
+
+      Object.entries(portsFound).forEach(([, ports]) => {
+        if (ports.size > 1) {
+          // Found candidates with the same address and relatedPort, but different ports
+          this.result.natType = NatType.SymmetricNat;
```
marcin-bazyl: we could also add support for other types of nat like full cone, right? (if we get multiple srflx candidates from different servers all with same port) It could be added in another PR...

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -963,6 +970,15 @@ export default class Reachability extends EventsScope {
         }
       );
 
+      this.clusterReachability[key].on(
+        Events.natTypeUpdated,
+        async (data: NatTypeUpdatedEventData) => {
+          results[key].natType = data.natType;
```
marcin-bazyl: the natType determined from all cluster checks should be always the same (unless some have it as unknown - it can be for clusters that have only 1 stun server), so in the final `results` we don't need to store it per cluster, just one overall natType and show a warning if some clusters detect different natType (other than unknown, but in this PR we don't have any other than symmetric anyway)

k-wasniowski: Correct, only issue is that we are storing results in the:
```
export type ReachabilityResults = Record<
  string,
  ClusterReachabilityResult & {
    isVideoMesh?: boolean;
  }
>;
```
Which is array of clusters. To keep single natType property would require a lot of changes of existing code. Do you think we should do it anyway?

SomeBody16 (approved):

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
k-wasniowski: Done

marcin-bazyl (approved):

PR #4158, opened 2025-03-20
fix(plugin-meetings): added client-environment to /clusters API request body

venky-mediboina:
…b client to send to discovery api

Reading the values from config to send in payload for discovery API to get additional response.

Modified the request payload formation by getting the required values from sdk configuration.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/reachability/request.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/request.ts
@@ -34,12 +34,14 @@ class ReachabilityRequest {
    * @param {string} trigger that's passed to Orpheus
    * @param {IP_VERSION} ipVersion information about current ip network we're on
    * @param {Object} previousReport last reachability result
+   * @param {Object} clientEnvironment information about the client environment
    * @returns {Promise}
    */
   getClusters = (
     trigger: GetClustersTrigger,
     ipVersion?: IP_VERSION,
-    previousReport?: any
+    previousReport?: any,
+    clientEnvironment?: any
   ): Promise<{
     clusters: ClusterList;
     joinCookie: any;
@@ -67,11 +69,7 @@ class ReachabilityRequest {
               'early-call-min-clusters': true,
             },
             'previous-report': previousReport,
-            'client-environment': {
-              components: {
-                [this.webex.config.support.appType]: this.webex.config.support.appVersion,
-              },
-            },
+            'client-environment': clientEnvironment,
             trigger,
           },
           timeout: this.webex.config.meetings.reachabilityGetClusterTimeout,
--- a/packages/@webex/plugin-meetings/src/reachability/request.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/request.ts
@@ -47,6 +47,18 @@ class ReachabilityRequest {
     joinCookie: any;
     discoveryOptions?: Record<string, any>;
   }> => {
+    const appType = this.webex?.config?.support?.appType;
+    const appVersion = this.webex?.config?.support?.appVersion;
+
+    const clientEnv = {
+      'client-environment': {
+        components: {
+          ...(appType && appVersion ? {[appType]: appVersion} : {}),
+          ...clientEnvironment?.components,
+        },
+      },
+    };
+
     // we only measure latency for the initial startup call, not for other triggers
     const callWrapper =
       trigger === 'startup'
@@ -69,7 +81,7 @@ class ReachabilityRequest {
               'early-call-min-clusters': true,
             },
             'previous-report': previousReport,
-            'client-environment': clientEnvironment,
+            ...clientEnv,
             trigger,
           },
           timeout: this.webex.config.meetings.reachabilityGetClusterTimeout,
--- a/packages/@webex/plugin-meetings/src/reachability/request.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/request.ts
@@ -41,7 +41,7 @@ class ReachabilityRequest {
     trigger: GetClustersTrigger,
     ipVersion?: IP_VERSION,
     previousReport?: any,
-    clientEnvironment?: any
+    clientEnvironment?: {components: Record<string, string>}
   ): Promise<{
     clusters: ClusterList;
     joinCookie: any;
--- a/packages/@webex/plugin-meetings/src/reachability/request.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/request.ts
@@ -34,7 +34,6 @@ class ReachabilityRequest {
    * @param {string} trigger that's passed to Orpheus
    * @param {IP_VERSION} ipVersion information about current ip network we're on
    * @param {Object} previousReport last reachability result
-   * @param {Object} clientEnvironment information about the client environment
    * @returns {Promise}
    */
   getClusters = (
--- a/packages/@webex/plugin-meetings/src/reachability/request.ts
+++ b/packages/@webex/plugin-meetings/src/reachability/request.ts
@@ -50,15 +50,6 @@ class ReachabilityRequest {
     const appType = this.webex?.config?.support?.appType;
     const appVersion = this.webex?.config?.support?.appVersion;
 
-    const clientEnv = {
-      'client-environment': {
-        components: {
-          ...(appType && appVersion ? {[appType]: appVersion} : {}),
-          ...clientEnvironment?.components,
-        },
-      },
-    };
-
     // we only measure latency for the initial startup call, not for other triggers
     const callWrapper =
       trigger === 'startup'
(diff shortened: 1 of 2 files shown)


marcin-bazyl:
The changes look fine, just need an update to unit tests, but we need to change the PR title. SDK release notes are generated from PR titles and network checker is not part of the SDK, so it shouldn't be mentioned, change it to something like this:
"fix(plugin-meetings): added client-environment to /clusters API request body"

Also in PR description we should just say that the values are read from config, not that these are network checker app type or version, because they can be anything else, they're controlled by the client app.

packages/@webex/plugin-meetings/src/reachability/request.ts
```diff
@@ -34,12 +34,14 @@ class ReachabilityRequest {
...
    * @param {Object} previousReport last reachability result
+   * @param {Object} clientEnvironment information about the client environment
    * @returns {Promise}
    */
   getClusters = (
     trigger: GetClustersTrigger,
     ipVersion?: IP_VERSION,
-    previousReport?: any
+    previousReport?: any,
+    clientEnvironment?: any
```
marcin-bazyl: I think Orpheus specifies how `clientEnvironment` should look like, right? Can we define a type for this instead of `any`? For example:
`clientEnvironment?: {components: Record<string, string>}`

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
```diff
@@ -57,14 +58,20 @@ describe('plugin-meetings/reachability', () => {
       previousReport = {
         id: 'fake previous report',
       }
+
+      clientEnvironment = {
+        'components': [{
```
marcin-bazyl: nitpick: this is just a unit test, so it doesn't matter that much, but I think Orpheus expects components to be an object not an array, right? so would be better to use in the tests values that resemble what is used in real life

arun3528 (approved):

arun3528 (commented):
lgtm after changes @marcin-bazyl mentioned

packages/@webex/plugin-meetings/src/reachability/request.ts
venky-mediboina: Added type.

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
venky-mediboina: Modified to object.

packages/@webex/plugin-meetings/src/reachability/request.ts
```diff
@@ -34,17 +34,31 @@ class ReachabilityRequest {
...
     joinCookie: any;
     discoveryOptions?: Record<string, any>;
   }> => {
+    const appType = this.webex?.config?.support?.appType;
+    const appVersion = this.webex?.config?.support?.appVersion;
+
+    const clientEnv = {
+      'client-environment': {
+        components: {
+          ...(appType && appVersion ? {[appType]: appVersion} : {}),
```
marcin-bazyl: This line of code is not tested in the tests. The tests only check the clientEnvironment passed in as the argument and don't check the values from config and these are actually the most important part as you have not written any code that ever passes any clientEnvironment to getClusters as an argument anyway.
BTW, I'm not sure why you're doing both - using config and letting more clientEnvironment to be passed in, it's unlikely that this will be ever needed, but I don't mind, it's up to you, but if you have added both mechanisms than you need to test both.

venky-mediboina: Added this to make test cases pass. Removed this block as configured appVersion and appType in test case.

packages/@webex/plugin-meetings/src/reachability/request.ts
```diff
@@ -34,17 +34,22 @@ class ReachabilityRequest {
...
    * @param {Object} previousReport last reachability result
+   * @param {Object} clientEnvironment information about the client environment
    * @returns {Promise}
    */
   getClusters = (
     trigger: GetClustersTrigger,
     ipVersion?: IP_VERSION,
-    previousReport?: any
+    previousReport?: any,
+    clientEnvironment?: {components: Record<string, string>}
```
marcin-bazyl: `clientEnvironment` argument is now not used at all, so should be removed

packages/@webex/plugin-meetings/src/reachability/request.ts
```diff
@@ -67,6 +72,7 @@ class ReachabilityRequest {
               'early-call-min-clusters': true,
             },
             'previous-report': previousReport,
+            'client-environment': {components: {[appType]: appVersion}},
```
marcin-bazyl: `appType` and `appVersion` can be undefined, in that case we shouldn't sent the "client-environment" to Orpheus

packages/@webex/plugin-meetings/src/reachability/request.ts
```diff
@@ -34,6 +34,7 @@ class ReachabilityRequest {
    * @param {string} trigger that's passed to Orpheus
    * @param {IP_VERSION} ipVersion information about current ip network we're on
    * @param {Object} previousReport last reachability result
+   * @param {Object} clientEnvironment information about the client environment
```
marcin-bazyl: this jsdoc is wrong, the function doesn't take this argument anymore

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
```diff
@@ -56,15 +68,17 @@ describe('plugin-meetings/reachability', () => {
...
+
     });
 
     afterEach(() => {
       sinon.restore();
     });
 
     it('sends a POST request with the correct params when trigger is "startup"', async () => {
-      const res = await reachabilityRequest.getClusters('startup', IP_VERSION.only_ipv4, previousReport);
+      const res = await reachabilityRequest.getClusters('startup', IP_VERSION.only_ipv4, previousReport, clientEnvironment);
```
marcin-bazyl: getClusters() doesn't take clientEnvironment as argument, so it shouldn't be passed

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
```diff
@@ -79,6 +93,9 @@ describe('plugin-meetings/reachability', () => {
             'report-version': 1,
             'early-call-min-clusters': true,
           },
+          ...(appType && appVersion && {
```
marcin-bazyl: the checks for `appType && appVersion` are not needed in the tests as the tests fully control these values

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
```diff
@@ -115,4 +135,4 @@ describe('plugin-meetings/reachability', () => {
       assert.notCalled(webex.internal.newMetrics.callDiagnosticLatencies.measureLatency);
     });
   });
-});
+});
```
marcin-bazyl, line 163: having one test that checks that it works fine with webex.config.support.appType and webex.config.support.appVersion being undefined would be nice

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/reachability/request.ts
venky-mediboina: Removed.

packages/@webex/plugin-meetings/src/reachability/request.ts
venky-mediboina: Added conditional check before adding.

packages/@webex/plugin-meetings/src/reachability/request.ts
venky-mediboina: Removed from here too.

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
venky-mediboina: removed from all the TCs.

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
venky-mediboina: Removed from both TCs.

packages/@webex/plugin-meetings/test/unit/spec/reachability/request.js
```diff
@@ -114,5 +129,37 @@ describe('plugin-meetings/reachability', () => {
       assert.deepEqual(res.joinCookie, {anycastEntryPoint: "aws-eu-west-1"})
       assert.notCalled(webex.internal.newMetrics.callDiagnosticLatencies.measureLatency);
     });
+
+    it('sends a POST request with the correct params when appVersion is undefined', async () => {
+      // Mocking appType & appVersion to undefined
+      webex.config.support.appType = undefined;
+      webex.config.support.appVersion = undefined;
+      appType = webex?.config?.support?.appType;
```
marcin-bazyl: what is the purpose of this line and next? both `appType` and `appVersion` are not used in this test

venky-mediboina: Removed.

marcin-bazyl (approved):

PR #4167, opened 2025-03-26
fix(plugin-cc): set autoResumed to false if payload is empty

mkesavan13:
# COMPLETES Ad-hoc

The Resume Recording API was mandating autoResumed parameter while it wasn't required unless enabled

**The Problem (SDK Sample vs Widget)**

| App | Request | Code | Note |
| --- | --- | --- | --- |
| SDK Sample | Screenshot 2025-03-26 at 10 41 37 AM | Screenshot 2025-03-26 at 10 47 22 AM| Note that there's a payload always |
| Widget | Screenshot 2025-03-26 at 10 41 44 AM | Screenshot 2025-03-26 at 10 45 42 AM | Note that the payload is empty always which triggers a GET request instead of POST |

- Adding a payload with `autoResumed` to be false whenever payload is undefined
- The Sample app is also altered to not send any payload if the autoResumed box is not checked

**Screen recording of fix**

https://github.com/user-attachments/assets/1ed9c53a-4752-42fc-9e83-5284ab142881

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tested in the sample app with both autoResumed checked and not checked

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-cc/src/services/task/index.ts
+++ b/packages/@webex/plugin-cc/src/services/task/index.ts
@@ -274,9 +274,7 @@ export default class Task extends EventEmitter implements ITask {
     resumeRecordingPayload: ResumeRecordingPayload
   ): Promise<TaskResponse> {
     try {
-      if (typeof resumeRecordingPayload === 'undefined') {
-        resumeRecordingPayload = {autoResumed: false};
-      }
+      resumeRecordingPayload ??= {autoResumed: false};
 
       const result = await this.contact.resumeRecording({
         interactionId: this.data.interactionId,
--- a/packages/@webex/plugin-cc/src/services/task/index.ts
+++ b/packages/@webex/plugin-cc/src/services/task/index.ts
@@ -274,6 +274,10 @@ export default class Task extends EventEmitter implements ITask {
     resumeRecordingPayload: ResumeRecordingPayload
   ): Promise<TaskResponse> {
     try {
+      if (typeof resumeRecordingPayload === 'undefined') {
+        resumeRecordingPayload = {autoResumed: false};
+      }
+
       const result = await this.contact.resumeRecording({
         interactionId: this.data.interactionId,
         data: resumeRecordingPayload,
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -1046,7 +1046,8 @@ function togglePauseResumeRecording() {
     });
   } else {
     pauseResumeRecordingElm.disabled = true;
-    task.resumeRecording({ autoResumed: autoResumed }).then(() => {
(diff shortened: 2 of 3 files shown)


packages/@webex/plugin-cc/src/services/task/index.ts
```diff
@@ -274,6 +274,10 @@ export default class Task extends EventEmitter implements ITask {
     resumeRecordingPayload: ResumeRecordingPayload
   ): Promise<TaskResponse> {
     try {
+      if (typeof resumeRecordingPayload === 'undefined') {
+        resumeRecordingPayload = {autoResumed: false};
```
Kesari3008: What if we keep it empty object ? Would that not work ?

Kesari3008 (changes requested):
Can we add a screen recording of testing done in sample app that show the network logs open on the side and actions being done on left in sample app?
Also it would be good if we can update the description either with the conversation link from the webex space or add the screenshots where we observed difference in HTTP method and request from widgets failing due to payload params not passed so that it's very clear what was the issue and the fix we have done

packages/@webex/plugin-cc/src/services/task/index.ts
mkesavan13: Empty object did not work. It was a bad request from the backend. That's why we had to send a payload like this

bhabalan: How about using a default value for the param?

mkesavan13: Can do that too. Just wanted the code to be more readable so people know that this check was added there deliberately.

Kesari3008 (approved):

Kesari3008:
We should test it locally with widgets if possible before merging to confirm that it works smoothly there too

packages/@webex/plugin-cc/src/services/task/index.ts
pagour98: Just a suggestion to use Nullish Coalescing Assignment to assign only if resumeRecordingPayload is undefined. By this we can avoid if condition

resumeRecordingPayload ??= { autoResumed: false };

mkesavan13:
> We should test it locally with widgets if possible before merging to confirm that it works smoothly there too

https://github.com/user-attachments/assets/b0879fbf-7106-46a8-b141-91b1da5f1cca

Done @Kesari3008

pagour98 (changes requested):
LGTM. Just a Suggestion to improve the code.

packages/@webex/plugin-cc/src/services/task/index.ts
mkesavan13: Thanks, @pagour98. Done

pagour98 (approved):

PR #4171, opened 2025-03-27
fix(plugin-cc): fixed webex import with contact-center.min.js

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-633914 

Fixed the export from contact-center.js so webex can be imported from the contact center module itself.

1. Updated index.html to use contact-center.min.js
2. Fixed the ESM export syntax in contact-center.js to export webex object.
3. Triggered device.register and mercury.connect within cc.register() to make desktop login mode work.

### Change Type

- [ ] Documentation update
- [x] Tooling change
- [ ] Internal code refactor

Initialized WxCC SDK and subscribed to the websocket connection. Confirmed WDM registration and mercury connection.
Station login using Extension/Browser was successful.
Call controls working for both modes.

https://app.vidcast.io/share/ee45fdcb-96fd-41b8-ae3b-f25110f00bc3 

Attaching logs for testing with  webtrc disabled:
[ConsoleLogswithwebrtcdisabled.log](https://github.com/user-attachments/files/19568445/ConsoleLogswithwebrtcdisabled.log)

[NetworkLogsWithWerbrtcDisabled.log](https://github.com/user-attachments/files/19568457/NetworkLogsWithWerbrtcDisabled.log)

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-cc/src/cc.ts
+++ b/packages/@webex/plugin-cc/src/cc.ts
@@ -114,6 +114,21 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    */
   public async register(): Promise<Profile> {
     try {
+      this.$webex.internal.mercury
+        .connect()
+        .then(() => {
+          LoggerProxy.info('Authentication: webex.internal.mercury.connect successful', {
+            module: CC_FILE,
+            method: this.register.name,
+          });
+        })
+        .catch((error) => {
+          LoggerProxy.warn(`Error occurred during mercury.connect() ${error}`, {
+            module: CC_FILE,
+            method: this.register.name,
+          });
+        });
+
       this.setupEventListeners();
 
       return await this.connectWebsocket();
--- a/packages/@webex/plugin-cc/src/cc.ts
+++ b/packages/@webex/plugin-cc/src/cc.ts
@@ -167,7 +167,10 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
             method: this.connectWebsocket.name,
           });
 
-          if (this.agentConfig.webRtcEnabled) {
+          if (
+            this.agentConfig.webRtcEnabled &&
+            this.agentConfig.loginVoiceOptions.includes(LoginOption.BROWSER)
+          ) {
             this.$webex.internal.mercury
               .connect()
               .then(() => {
@@ -228,7 +231,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
         },
       });
 
-      if (data.loginOption === LoginOption.BROWSER) {
+      if (this.agentConfig.webRtcEnabled && data.loginOption === LoginOption.BROWSER) {
         await this.webCallingService.registerWebCallingLine();
       }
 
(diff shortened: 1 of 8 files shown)


rsarika (approved):

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -114,6 +114,21 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    */
   public async register(): Promise<Profile> {
     try {
+      this.$webex.internal.mercury
```
mkesavan13: If this mercury doesn't get connected, shouldn't we reject this API and not setup event listeners and not connect to the CC web socket?

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -335,6 +339,7 @@ describe('webex.cc', () => {
         webRtcEnabled: false,
         lostConnectionRecoveryTimeout: 0,
       };
+      jest.spyOn(webex.internal.mercury, 'connect').mockResolvedValue(true);
```
mkesavan13, line 301: We need to test the reject case as well

mkesavan13 (changes requested):

packages/@webex/plugin-cc/test/unit/spec/cc.ts
Kesari3008: Do you mean when connect fails ? Testcase for catch block ?

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: If we do that then we will block the extension/dial number login as well which shouldn't be the case. If mercury connection fails then we should not engage in webcallingservice object creation or fail the station login for browser before even attempting to do mobius registration

packages/@webex/plugin-cc/test/unit/spec/cc.ts
Kesari3008: Addressed

packages/@webex/plugin-cc/src/cc.ts
mkesavan13: Right. That makes sense. Just thinking if we should have did this connection request during station login if the device type was Browser. Otherwise we are opening up the connection and keeping it simply or doing WDM registration for nothing while the user is only on Dial Number/Extension mode

packages/webex/src/contact-center.js
```diff
@@ -17,4 +18,4 @@ Webex.init = function init(attrs = {}) {
   return new Webex(attrs);
 };
 
-module.exports = Webex;
+export default Webex;
```
mkesavan13, line 21: Have we tested this in node?

Just curious how `require` and `export` live in the same file.

mkesavan13 (approved):
Looks good to me mostly. Once this gets merged, we should update our S4D docs as well as Beta document to update the import statement.

Let us also ensure to test this once with npm package import as those documents recommend using the npm package

PR #4173, opened 2025-03-28
feat(plugin-cc): metrics tracking for cc.ts

bhabalan:
# COMPLETES #[SPARK-635169](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-635169)

Add metrics to cc.ts

Using MetricsManager to track events described in the confluence

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

- Features that are being tracked were tested

### I certified that

- [] All existing and new tests passed

---


bhabalan:
@coderabbitai review

bhabalan:
[all cc.ts metrics.har.zip](https://github.com/user-attachments/files/19579394/all.cc.ts.metrics.har.zip)
image

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
```diff
@@ -73,6 +73,7 @@ export type MetricEventVerb =
   | 'list'
   | 'load'
   | 'login'
+  | 'relogin'
```
bhabalan: Not needed anymore. Will remove

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -33,7 +35,7 @@ export default class MetricsManager {
 
   // eslint-disable-next-line no-use-before-define
   private static instance: MetricsManager;
-  private metricsDisabled = false;
+  private metricsDisabled = true; // TODO: SPARK-637285
```
mkesavan13: So, merging this PR won't send the metrics is it?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -270,4 +275,46 @@ export default class MetricsManager {
   public static resetInstance() {
     MetricsManager.instance = undefined;
   }
+
+  public static getCommonTrackingFieldForAQMResponse(response: any): Record<string, any> {
+    const fields = {
+      agentId: response?.data?.agentId || EMPTY_STRING,
+      teamId: response?.data?.teamId || EMPTY_STRING, // TODO: handle multiple teams
```
mkesavan13: Is there a Jira to address this? If today there are multiple teams, would this break somehow?

packages/@webex/plugin-cc/src/metrics/constants.ts
```diff
@@ -1,12 +1,16 @@
-type Enum<T extends Record<string, unknown>> = T[keyof T];
-
 export const METRIC_EVENT_NAMES = {
-  STATION_LOGIN: 'Station Login',
-  CALL_COMPLETED: 'Call Completed',
-  TASK_END: 'Task End',
-  CALL_CONSULT_ACTIVATE: 'Call Consult Activate',
-  CALL_TRANSFER_ACTIVATE: 'Call Transfer Activated',
-  CALL_ANSWERED: 'Call Answered',
+  STATION_LOGIN_SUCCESS: 'Station Login Success',
```
mkesavan13, line 4: These are the names that finally land in the metrics dashboards is it? If so, would it be more efficient to not have spaces?

Considering we would share filter based URLs/permalinks around?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -114,10 +115,31 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    */
   public async register(): Promise<Profile> {
     try {
+      this.metricsManager.timeEvent([
```
mkesavan13, line 129: Should we at least have all the UTs in these APIs to check if timeEvent is being called? Just to ensure we do metrics on all the APIs?

mkesavan13 (commented):
Looks good to me mostly. Just a few comments. Also let us have the title reflecting the component like this:
`feat(plugin-cc): metrics tracking for cc.ts`

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: It won't. This is to avoid send metrics during dev/testing

packages/@webex/plugin-cc/src/metrics/constants.ts
bhabalan: These are not the final names.
For behavioral, The names would come from getEventTaxonomy() in packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
For business and operational, these would be uppercased and snakecased

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
sreenara: Did we add this or was it already there? The agent desktop also uses this plugin.

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -270,4 +275,46 @@ export default class MetricsManager {
   public static resetInstance() {
     MetricsManager.instance = undefined;
   }
+
+  public static getCommonTrackingFieldForAQMResponse(response: any): Record<string, any> {
```
sreenara, line 288: Can you add more description to the PR about what these new functions are doing?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -103,33 +105,29 @@ export default class MetricsManager {
...
+          metadata: {
+            appType: PRODUCT_NAME,
+          },
         });
       });
     }
   }
 
-  private addDurationIfTimed(name: string, options?: EventPayload): EventPayload {
+  private addDurationIfTimed(eventName: string, options?: EventPayload): EventPayload {
```
sreenara, line 117: Why did we change this function?

sreenara (changes requested):
As discussed, what do you think about removing the check for sending metrics in dev mode? Considering we have a beta starting next week, isn't everyone in dev mode?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Earlier we had one name for each event type.
Eg: Station Login, So we were sending just one event name for timing. 
Now we have two names per event. i.e: Station Login Success, Station Login Failed
So needed to add ability to support this as the events were being used as keys

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: I'll add more comments to these functions.
Basically these are just utility functions to get common tracking fields for success and failure cases for AQM based requests

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Addressed this in this PR itself

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -270,4 +275,58 @@ export default class MetricsManager {
   public static resetInstance() {
     MetricsManager.instance = undefined;
   }
+
+  public static getCommonTrackingFieldForAQMResponse(response: any): Record<string, any> {
+    let teamId;
+    if (response?.data?.teamId) {
```
sreenara: I'm not seeing any UT that covers all the conditions here. Could you please point me to that?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -270,4 +275,58 @@ export default class MetricsManager {
...
+      agentId: failureResponse?.data?.agentId,
+      trackingId: failureResponse?.trackingId,
+      notifTrackingId: failureResponse?.trackingId,
+      orgId: failureResponse?.orgId,
+      failureType: failureResponse?.type,
+      failureReason: failureResponse?.data?.reason,
+      reasonCode: failureResponse?.data?.reasonCode,
+    };
+
+    Object.keys(fields).forEach((key) => {
```
sreenara: Can you point me to the UT that tests these conditions?

sreenara (changes requested):
I only have comments about the UT. Other things look fine.

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: I've added the tests for these. I have also moved the removal of emptyish fields to the preparePayload method.

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Added now

packages/@webex/plugin-cc/src/cc.ts
bhabalan: Added UTs

packages/@webex/plugin-cc/src/metrics/constants.ts
mkesavan13: In that case why do we have these space separated values? What's the purpose there?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
```diff
@@ -270,4 +284,52 @@ export default class MetricsManager {
...
+    // object and its data property.
+    let teamId: string | undefined;
+    if (response?.data?.teamId) {
+      teamId = response.data.teamId;
+    } else if (response?.teamId) {
+      teamId = response.teamId;
+    } else if (Array.isArray(response?.data?.teams) && response.data.teams.length > 0) {
+      teamId = response.data.teams[0].id;
+    } else if (Array.isArray(response?.teams) && response.teams.length > 0) {
+      teamId = response.teams[0].id;
```
mkesavan13: On what grounds are we choosing zero-th team always? What if the agent chose to login with teams[1] or teams[2] or any random team[n]?

mkesavan13 (commented):
Added a few comments out of this review as well. Please check

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: The metrics-A api accepts only boolean, string or number. Cannot send array. The other option is to concat the teamIds

packages/@webex/plugin-cc/src/metrics/constants.ts
bhabalan: These are Enum values. Just for the sake of readability

packages/@webex/internal-plugin-metrics/src/metrics.types.ts
bhabalan: We added it. Now removed

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Also, I have seen the teams as an array only while we do websocket register. After that we get only one teamId instead of an array after StationLogin. The if..else.. ladder takes care of this. This picking of zero'th element is just a failsafe

mkesavan13: So, while doing station login, the user would have chosen a team before logging in right? Shouldn't we be using that teamId here?

packages/@webex/plugin-cc/src/metrics/constants.ts
mkesavan13: In code, we just use these underscore separated variable names and in metrics dashboards, the space separated values won't show up. So, where are they being read is what am curious about?

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
bhabalan: Handled higher up in the if..else.. ladder

packages/@webex/plugin-cc/src/metrics/constants.ts
bhabalan: Readable by humans. Eventually they will be snake cased on the dashboard.

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
mkesavan13: > Also, I have seen the teams as an array only while we do websocket register. After that we get only one teamId instead of an array after StationLogin. The if..else.. ladder takes care of this. This picking of zero'th element is just a failsafe

If that is the case, am curious if those metrics would somehow mislead us. The metrics we record until station login could possibly have a teamId that is not chosen by the agent after station login. Is my interpretation correct here?

bhabalan: Only one event before station login -> Websocket register. This returns an array of teams.
After StationLogin -> We get direct teamId. No array of teams

mkesavan13: As discussed on Webex, we need to take a call how are we going to do this. Please have a discussion with @sreenara before merging the PR

mkesavan13 (approved):

packages/@webex/plugin-cc/src/metrics/MetricsManager.ts
sreenara: Let's not track the teams in this event. We will have the required information during station login anyway.

sreenara (approved):

Issue #3510, opened 2024-04-02
Webex SDK 3.0.0 not compatible with any version of NodeJS


Tiuipuv:
**Describe the bug**
The following script fails to initialize in NodeJS:
```js
import webex from 'webex';

webex.init({credentials: '<valid credentials were provided here>'});
```

Yielding the following error:
```
ReferenceError: window is not defined
    at Object.<anonymous> ([...]\node_modules\@webex\plugin-meetings\node_modules\@webex\internal-media-core\dist\cjs\index.js:2065:29024)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> ([...]\node_modules\@webex\plugin-meetings\dist\meetings\index.js:35:26)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
```

This has been tested with nodejs `v12`, `v14`, `v16`, `v18`, and `v20`.
package.json:
```json
{
  "name": "test_webex",
  "type": "module",
  [...]
  "dependencies": {
    "webex": "^3.0.0"
  }
}
```
This error is produced for both ESM modules and CommonJS scripts.

**To Reproduce**
1. create a blank npm project using `npm init`
2. install webex sdk with `npm install webex`
3. create `index.js`, with the above code (enter valid bot id)
4. run with `node index.js`
5. See error

**Expected behavior**
Webex SDK successfully imports into script.

**Platform (please complete the following information):**
 - OS: Windows 10 Pro 22H2
 - Browser: N/A (NodeJS)
 - Version: v12.22.12, v14.21.3, v16.20.2, v18.20.0, v20.12.0
 - Device Type: Desktop

valgaze:
+1 was about to file an issue as well

Confirm failure in node 21

<img width="934" alt="image" src="https://github.com/webex/webex-js-sdk/assets/1396559/901e1f5f-a691-4c17-bbd6-ba00909e96d6">

valgaze:
@Tiuipuv 

Two things:

[...]

PR #4187, opened 2025-04-02
feat(webex-node): add a new package webex-node

sreenara:
# COMPLETES #3510 

The issue of being unable to import `webex` in a NodeJS environment. The problems are detailed more in Issue #3510 as well as some workarounds to this. This is a permanent solution.

1. New package `webex-node` created in the workspaces
2. This package does _not_ import `@webex/plugin-meetings`, `@webex/calling`, `@webex/plugin-cc` which all require media and therefore the `window` object.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Created a new package that imports `webex-node` and initializes a space, sends a message on the new space.
Test code is present in this repo: https://github.com/sreenara/test-nodejs-sdk/blob/main/src/index.js

### I certified that

---


packages/webex-node/src/config.js
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright (c) 2015-2020 Cisco Systems, Inc. See LICENSE file.
+ */
+
+// eslint-disable-next-line import/extensions
+import storageConfig from './config-storage.js';
```
sreenara, line 6: This change was required because the ES module import in config.js was missing the file extension. In Node's ES module resolution, you must include the full filename (e.g. ./config-storage.js) when importing a local module.

packages/webex-node/package.json
```diff
@@ -0,0 +1,68 @@
+{
+    "name": "webex-node",
```
mkesavan13, line 2: Should we consider naming this either,
```suggestion
    "name": "webex/node",
```

or

```suggestion
    "name": "@webex/node",
```

Just for uniformity?

mkesavan13 (commented):
This is great. Apart from the comment below, can we also consider having another mono-repo package that imports this package. Like a node sample app?

Or we could also try and revive it: https://github.com/webex/webex-js-sdk/tree/next/packages/%40webex/test-helper-server

packages/webex-node/package.json
sreenara: @mkesavan13 I didn't name it with scoped `@webex` or `webex/` because I wanted this to be exactly the same as the `webex` package. It's just to let developers know that it's `webex` but for Node.js.

Shreyas281299 (approved):
This is really awesome. I think we should make this public ASAP.

But we would need to be very specific when we make this public that this doesnt not support meeting,calling

sreenara:
> This is great. Apart from the comment below, can we also consider having another mono-repo package that imports this package. Like a node sample app?
> 
> Or we could also try and revive it: https://github.com/webex/webex-js-sdk/tree/next/packages/%40webex/test-helper-server

@mkesavan13 I've added a new package `@webex/test-webex-node` that helps test the new `webex-node` package.

sreenara:
> This is really awesome. I think we should make this public ASAP.
> 
> But we would need to be very specific when we make this public that this doesnt not support meeting,calling

Yes @Shreyas281299 . We'll be putting the NodeJS SDK doc back on the developer portal with changes to the doc and ensure we call out which flows aren't supported.

packages/@webex/test-webex-node/README.md
```diff
@@ -0,0 +1,42 @@
...
+  - [Usage](#usage)
+    - [Environment Defaults](#environment-defaults)
+  - [Maintainers](#maintainers)
+  - [Contribute](#contribute)
+  - [License](#license)
+
+## Install
+
+```bash
+npm install --save @webex/test-webex-node
```
mkesavan13, line 18: nitpick: Should we change this to a yarn command as well?

mkesavan13 (approved):

PR #4189, opened 2025-04-03
feat(plugin-cc): add getQueues() method

rarajes2:
# COMPLETES #< [SPARK-567681](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-567681) >

While performing consult or transfer we have option to choose the type of consult such as agent, DN, queue etc.
We needed to fetch the list of queues to perform the consult/transfer operation for queue as a result and for that we have added a method inside `cc.ts` file as `getQueues()`.

Added a method `getQueues(orgId)` inside the `packages/@webex/plugin-cc/src/services/config/index.ts` and exposed to the developer through `cc.ts` file with the same name.

We are able to get the list of queue.

**Vidcast** - https://app.vidcast.io/share/abe60753-715a-4811-b655-e0a40cb7c303

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Able to get the list of queue
- Filtering those queue inside the samples app to perform consult and transfer through queue.
  - **Consult** API is returning 202 status code but the next line is not being executed causing the buttons to remain in the same state --- NEED FURTHER TESTING AND CLARIFICATION
  - In case of error --- We are not able to get the server error message because we are creating our own error message and throwing that for the application to consume. NEED LARGER TEAM DISCUSSION ON THIS.
  - In case of **Transfer** it seems to be working fine.

**Note**: Based on the requirement mentioned on the ticket, the `getQueues()` method is working as expected.

### I certified that

---

### DIFF
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -168,6 +168,13 @@ function initOauth() {
   });
 }
 
+function toggleIfQueueConsultEnabled () {
+  document.querySelectorAll('option[value="queue"]').forEach(item => {
+    if(webex && !webex.cc.agentConfig.allowConsultToQueue) item.style.display = 'none';
+    else item.style.display = 'block';
+  });
+}
+
 const taskEvents = new CustomEvent('task:incoming', {
   detail: {
     task: task,
@@ -236,7 +243,6 @@ async function getQueueListForTelephonyChannel() {
   }
 }
 
-
 async function onConsultTypeSelectionChanged(){
 
   consultDestinationHolderElm.innerHTML = '';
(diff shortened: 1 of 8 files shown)


packages/@webex/plugin-cc/src/services/config/index.ts
```diff
@@ -531,4 +532,44 @@ export default class AgentConfigService {
...
   }
+
+  /**
+   * Fetches the list of queues for the given orgId.
+   * @param {string} orgId
+   * @returns {Promise<ContactServiceQueue[]>}
+   */
+  public async getQueues(
+    orgId: string,
+    page: number,
```
Kesari3008, line 547: I believe even these parameters are optional right

packages/@webex/plugin-cc/src/services/config/index.ts
```diff
@@ -531,4 +532,44 @@ export default class AgentConfigService {
...
+        service: WCC_API_GATEWAY,
+        resource,
+        method: HTTP_METHODS.GET,
+      });
+
+      if (response.statusCode !== 200) {
+        throw new Error(`API call failed with ${response.statusCode}`);
+      }
+
+      LoggerProxy.log('getQueues api success.', {module: CONFIG_FILE_NAME, method: 'getQueues'});
```
Kesari3008: nitpick: API should be in caps. We should keep it consistent through all the logs. I know in other methods same mistake is there but just want to highlight it going forward

docs/samples/contact-center/app.js
```diff
@@ -274,6 +301,20 @@ async function onTransferTypeSelectionChanged() {
 
     const agentNodeList = await fetchBuddyAgentsNodeList();
     agentNodeList.forEach(n => { transferDestinationInput.appendChild(n) });
+  } else if (document.querySelector('#transfer-destination-type').value === 'queue') {
```
Kesari3008, line 324: Here to ensure that we only show the queue when consultToQueue is enabled, this is the parameter you need to look for: agentConfig.allowConsultToQueue. It is present under the agent profile/agentConfig received as response to cc.register(). Please add the check in kictehn sink app

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -1044,4 +1053,48 @@ describe('webex.cc', () => {
...
+        'mockOrgId',
+        0,
+        100,
+        undefined,
+        undefined
+      );
+      expect(result).toEqual(mockQueuesResponse);
+    });
+
+    it('shoule throw an error if orgId is not present', async () => {
```
Kesari3008, line 1084: I believe the third testcase is still valid becasue reject is coming from config/index.ts file and we just need to UT to verify that we return that reject as it is in this method too.

Kesari3008 (approved):
Approving with minor comments

Kesari3008:
Tested it locally and observed below for kitchen sink app:
1. There is no refresh option for the queue if are already in queue selected mode. So if config is changed in between, there is no way to fetch the fresh list unless we go to dropdown, select some other option and select queue again to get the recent list.
2. When there are no queue present, we receive incorrect message saying no buddy agents available. But it should say Queues not found or No Queues available or something like that.

Screenshot 2025-04-04 at 12 37 56 AM

mkesavan13 (approved):

packages/@webex/plugin-cc/src/services/config/index.ts
rarajes2: These are optional from the cc.ts file as we are passing a default value from there. Here they are required (It's internal)

packages/@webex/plugin-cc/src/services/config/index.ts
rarajes2: Done

packages/@webex/plugin-cc/test/unit/spec/cc.ts
rarajes2: Still don't think this test is required. Added anyway

docs/samples/contact-center/app.js
rarajes2: Done

PR #4200, opened 2025-04-09
fix(plugin-meetings): mute state for brb

antsukanova:
# COMPLETES 

Fixes issue flow with brb request, when locus mutes user on the server after brb:true sdk don't know about that. Currently, sdk contains a bug when brb:false then the client should be unmuted but it won't be unmuted on the server because the server didn't do unmute and the client also didn't know that it should do server unmute.

**How to test:** 
in sdk do :
`yarn prebuild:modules`
`yarn @all build:src`

in web client checkout to [webex-web-client step away manual branch](https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/7246)
do `yarn` and  `SDK_DIR=/Users/antsukan/WebstormProjects/webex-js-sdk yarn start` 

The issue with the remote mute should gone and the mute in the participant list should work as expected.

Introduces new `then` statement for `beRightBack` for remote mure updates.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] - manual test cases using sample app and locall linked web client app
- [x] - unit tests 

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/muteState.ts
```diff
@@ -466,4 +466,19 @@ export class MuteState {
...
+  /**
+   * This method sets the `remoteMute` property in the server state to the provided value
+   *
+   * @public
+   * @memberof MuteState
+   * @param {boolean} muteState - The new remote mute state.
+   * @returns {void}
+   */
+  public setServerRemoteMute(muteState: boolean) {
+    this.state.server.remoteMute = muteState;
```
marcin-bazyl: what if instead of this code we just called handleServerRemoteMuteUpdate() with muted=muteState and unmuteAllowed=undefined? This way we would ensure that not only the `this.state.server.remoteMute` is updated but also the web app knows that we are remotely muted - the solution would still be a bit of a hack, but just feels less hacky :-) would it work fine?

antsukanova: @marcin-bazyl Thank you for the suggestion, I have checked it and it works ok, so I will keep this idea, thanks for pointing to handleServerRemoteMuteUpdate, good to keep it there

antsukanova: @marcin-bazyl updated logic, appreciate your review here

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -82,28 +95,44 @@ describe('plugin-meetings', () => {
...
     it('sets source state override when client state does not match server state', async () => {
-      brbState.enable(true, meeting.sendSlotManager);
+      brbState.enable(true, meeting.sendSlotManager, audioMuteStateMock);
       brbState.handleServerBrbUpdate(true);
       await testUtils.flushPromises();
 
       assert.isTrue(meeting.sendSlotManager.setSourceStateOverride.calledOnce);
     });
 
+    it('handles server mute update for participant with brb enabled when client state does not match server state', async () => {
```
edvujic:
```suggestion
    it('processes server mute update when brb is enabled and states mismatch', async () => {
```

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -82,28 +95,44 @@ describe('plugin-meetings', () => {
...
+      brbState.enable(true, meeting.sendSlotManager, audioMuteStateMock);
+      brbState.handleServerBrbUpdate(true);
+      await testUtils.flushPromises();
+
+      assert.isTrue(handleServerRemoteMuteUpdateStub.calledOnce);
+    });
+
+    it('skips handling server mute update for participant with brb enabled when client state does not match server state', async () => {
```
edvujic:
```suggestion
    it('ignores server mute update when brb is disabled and states mismatch', async () => {
```

edvujic (approved):
LGTM

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -82,28 +95,44 @@ describe('plugin-meetings', () => {
...
 
       assert.isTrue(meeting.sendSlotManager.setSourceStateOverride.calledOnce);
     });
 
+    it('handles server mute update for participant with brb enabled when client state does not match server state', async () => {
+      brbState.enable(true, meeting.sendSlotManager, audioMuteStateMock);
+      brbState.handleServerBrbUpdate(true);
+      await testUtils.flushPromises();
+
+      assert.isTrue(handleServerRemoteMuteUpdateStub.calledOnce);
```
marcin-bazyl: we should check also the arguments passed to `handleServerRemoteMuteUpdate()`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3802,7 +3802,7 @@ export default class Meeting extends StatelessWebexPlugin {
       return Promise.reject(error);
     }
 
-    return this.brbState.enable(enabled, this.sendSlotManager);
+    return this.brbState.enable(enabled, this.sendSlotManager, this.audio);
```
brycetham: Did you consider, instead of passing `this.audio` here, to instead handle the call to `handleServerRemoteMuteUpdate` in this file?

I'm asking because it feels a bit strange to me to pass `MuteState` to `BrbState`. To me, they feel like they should be separated, i.e. let's handle enabling the `BrbState` first, and if it succeeds, then handle the `MuteState` (rather than handle the `MuteState` inside of the `BrbState`).

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
antsukanova: @edvujic Thanks, I will update the descriptions for these tests that we agreed with Marcin, please feel free to check them and say if they looks good to you as well

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
antsukanova: @marcin-bazyl changed logic so please could you review it once again, thank you

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: @brycetham I like your idea, thanks for that! it feels clearer and from an architectural point of view, so I have revised the logic in PR based on your idea. I've also tested it with the client, and it works well for our case and with Filip's logic for sync client/server logic.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3887,6 +3887,22 @@ describe('plugin-meetings', () => {
...
+            sinon.assert.calledOnceWithExactly(meeting.audio.handleServerRemoteMuteUpdate, meeting, true);
+          });
+
+          it.only('does not update remote mute state when brb is disabled and audio is available', async () => {
+            meeting.audio = { handleServerRemoteMuteUpdate: sinon.stub() };
+
+            await meeting.beRightBack(false);
+
+            assert.isTrue(meeting.audio.handleServerRemoteMuteUpdate.notCalled);
+          });
```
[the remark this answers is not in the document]
brycetham: @antsukanova this needs to be fixed

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3887,6 +3887,22 @@ describe('plugin-meetings', () => {
               assert.isRejected((Promise.reject()));
             }
           });
+
+          it('updates remote mute state when brb is enabled and audio is available', async () => {
```
brycetham: What does "audio is available" mean? (I know it means `this.audio` exists but under what scenarios would it not exist?)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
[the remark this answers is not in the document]
antsukanova: @brycetham will do, thanks, at least coderabbitai did well his part)

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
edvujic: I agree with @brycetham here, does it mean when the user has a microphone selected? If yes, it could be good to rephrase this.

marcin-bazyl: just FYI, `Meeting.audio` should always exist when there is a media connection, it's created by `Meeting.addMedia()`, so maybe we could rephrase this to "and media connection exists" or just "and audio MuteState instance is set"

antsukanova: @brycetham There are a couple of places in sdk api that make `this.audio` object `null` or `undefined` (see e.g. `clearMeetingData`, `closePeerConnections`). I feel this is a super rare case that this function will be triggered with other api that can give `null` for audio, but it feels good to be saved and have this logic inside the function to prevent this issue appears (let's imagine the user leaves meeting and at the same time auto brb triggered and `this.audio` `null` happens), so it's more for health error handling for this code block. 
Also in other places in this file, there is a check for `this.audio` for `null` so it feels natural to have this check as well here.

@edvujic fyi

antsukanova: @edvujic updated the test names to make their purpose clearer. appreciate your review on that

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -3381,8 +3381,8 @@ async function toggleBrb({unmuteAudio = false, unmuteVideo = false}) {
       isBrb = enableBrb;
 
       if (enableBrb) {
-        localMediaCameraMuted = localMedia.cameraStream.userMuted;
-        localMediaMicMuted = localMedia.microphoneStream.userMuted;
+        localMediaCameraMuted = localMedia.cameraStream?.userMuted;
+        localMediaMicMuted = localMedia.microphoneStream?.userMuted;
```
antsukanova, line 3385: There is a case with `localMedia.cameraStream` that will be `undefined` so I updated this brb code as well to avoid console issues.

edvujic (approved):

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
antsukanova: @marcin-bazyl Thank you for the idea, I have changed names to be focused on main testing logic, I don't feel here it's needed to add test cases with audio `null` so audio part can be omited (because such checks `meeting.audio.handleServerRemoteMuteUpdate.notCalled` won't fit well for audio null)
let me know please if it looks ok to you.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -3887,6 +3887,22 @@ describe('plugin-meetings', () => {
...
+
+            sinon.assert.calledOnceWithExactly(meeting.audio.handleServerRemoteMuteUpdate, meeting, true);
+          });
+
+          it('does not update remote mute state when brb is disabled', async () => {
+            meeting.audio = { handleServerRemoteMuteUpdate: sinon.stub() };
+
+            await meeting.beRightBack(false);
+
+            assert.isTrue(meeting.audio.handleServerRemoteMuteUpdate.notCalled);
```
marcin-bazyl: nitpick: you could just do `assert.notCalled(meeting.audio.handleServerRemoteMuteUpdate);` - feels more readable

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
marcin-bazyl: yeah, ok it's fine to not test for audio null, because Meeting.beRightBack() already checks `this.mediaProperties.webrtcMediaConnection` before calling `this.brbState.enable`, so audio should never really be null if we have a webrtcMediaConnection.

marcin-bazyl (approved):

brycetham (approved):

szotrabh (approved):

SomeBody16 (approved):

PR #4203, opened 2025-04-11
fix: added support for multiple TURN urls, removed TURN-TCP

marcin-bazyl:
# COMPLETES #[SPARK-590974](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-590974)

When backend rolls out native ipv6 support, they will send us 2 TURN URLs (one that has a DNS A record and one that has AAAA record), so we need to support receiving multiple TURN URLs.
As we will have now more TURN URLs, I'm removing the TURN-TCP trick we've added some time ago to minimize the number of URLs. Also we believe that the introduction of TURN-TCP didn't have much impact on JMF, but we will monitor the metrics and if JMF increases as a result of this PR, we'll look at what we can do about it and consider putting TURN-TCP back in.

Adding support for multiple TURN URLs.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

### I certified that

---


packages/@webex/plugin-meetings/src/roap/turnDiscovery.ts
```diff
@@ -18,28 +18,6 @@ const TURN_DISCOVERY_TIMEOUT = 10; // in seconds
 // and do the SDP offer with seq=1
 const TURN_DISCOVERY_SEQ = 0;
 
-const TurnDiscoverySkipReason = {
```
marcin-bazyl, line 21: this stuff is just moved to a separate types.ts file so that it can be imported from media/index.ts without adding a circular dependency

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -160,24 +157,11 @@ Media.createMediaConnection = (
...
-
-      iceServers.push({
-        urls: bareTurnServer,
-        username: turnServerInfo.username || '',
-        credential: turnServerInfo.password || '',
-      });
-    }
-
+  // we might not have any TURN server if TURN discovery failed or wasn't done or we land on a video mesh node
+  if (turnServerInfo?.urls.length > 0) {
```
marcin-bazyl, line 161: previously when we got an empty url (happens on VMN), it was just passed on from turnDiscovery.ts and reached all the way here and checked here with the check `if (turnServerInfo?.url)`. Now I've moved the filtering out of empty urls into handleTurnDiscoveryResponse() in turnDiscovery.ts so here we only need to check the length of urls array

jor-row (approved):

k-wasniowski (approved):

mkesavan13 (commented):
Changes look good to me. However, it does not look like it is backward compatible and your description for the PR says, "When the backend rolls out..." 

Just curious if we would wait for the backend to roll this change out and then merge the PR?

mkesavan13 (approved):

marcin-bazyl:
> Changes look good to me. However, it does not look like it is backward compatible and your description for the PR says, "When the backend rolls out..."
> 
> Just curious if we would wait for the backend to roll this change out and then merge the PR?

discussed 1-1  - changes are backwards compatible in the sense that SDK will work when Homer sends 1 url (like it does currently) or multiple

PR #4226, opened 2025-04-29
fix(plugin-cc): fixed-broken-transfer-issue

adhmenon:
# COMPLETES #< Ad-HOC >

* The `blind transfer` as well as `queue transfer` were both broken in the SDK.
* Attempting any of these was giving an error `task is not defined`
* The reason was because whenever we transferred, we had a recent change in the SDK code which cleared the task from the list on two events - AGENT_VTEAM_TRANSFERRED (for queue) and AGENT_CONTACT_UNASSIGNED (for agent transfer).
* Then when the wrapup state was shown, we got the AgentWrapup event wherein, we were updating the task, which was null, hence the SDK threw an error.

* Added a null check for the wrapup event (workaround)
* Added unit tests to ensure task update only happens when we have a task existing.

Transfer vidcast - https://app.vidcast.io/share/a14a14f2-eaeb-43ad-a7d3-707a3e325d63
Consult Transfer vidcast - https://app.vidcast.io/share/6c3787ea-9156-4df8-91ac-90a870929323

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested blind transfer for both agents and queues
* Tested consult for both agents and queues
* Tested consult-transfer for both agents and queues
* Tested regular call and end call
* Tetsed receiving of consult for both agent and queues.

Reviewer Testing logs between Agent5 and Agent11:
[Agent5ConsoleLogs.log](https://github.com/user-attachments/files/19962056/Agent5ConsoleLogs.log)
[Agent11ConsoleLogs.log](https://github.com/user-attachments/files/19962061/Agent11ConsoleLogs.log)

### I certified that

---

### DIFF
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -463,6 +463,7 @@ async function initiateConsultTransfer() {
     console.log('Consult transfer initiated successfully');
     consultTransferBtn.disabled = true; // Disable the consult transfer button after initiating consult transfer
     consultTransferBtn.style.display = 'none'; // Hide the consult transfer button after initiating consult transfer
+    endConsultBtn.style.display = 'none';
   } catch (error) {
     console.error('Failed to initiate consult transfer', error);
   }
--- a/packages/@webex/plugin-cc/src/services/task/TaskManager.ts
+++ b/packages/@webex/plugin-cc/src/services/task/TaskManager.ts
@@ -224,10 +224,7 @@ export default class TaskManager extends EventEmitter {
             task.emit(TASK_EVENTS.TASK_CONSULT_QUEUE_CANCELLED, task);
             break;
           case CC_EVENTS.AGENT_WRAPUP:
-            // Only update if task exists
-            if (task) {
-              task = this.updateTaskData(task, payload.data);
-            }
+            task = this.updateTaskData(task, payload.data);
             break;
           case CC_EVENTS.AGENT_WRAPPEDUP:
             this.removeTaskFromCollection(task);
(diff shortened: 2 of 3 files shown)


docs/samples/contact-center/app.js
```diff
@@ -463,6 +463,7 @@ async function initiateConsultTransfer() {
     console.log('Consult transfer initiated successfully');
     consultTransferBtn.disabled = true; // Disable the consult transfer button after initiating consult transfer
     consultTransferBtn.style.display = 'none'; // Hide the consult transfer button after initiating consult transfer
+    endConsultBtn.style.display = 'none';
```
adhmenon, line 466: Sample app fix - observed this while testing.

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
```diff
@@ -224,7 +224,10 @@ export default class TaskManager extends EventEmitter {
             task.emit(TASK_EVENTS.TASK_CONSULT_QUEUE_CANCELLED, task);
             break;
           case CC_EVENTS.AGENT_WRAPUP:
-            task = this.updateTaskData(task, payload.data);
+            // Only update if task exists
+            if (task) {
```
Kesari3008: Instead of adding check here, does it make sense to add the check in updateTaskData method itself for task object to be valid so that for any similar cases it would be seamlessly handled and we don't have to worry about it per event basis

adhmenon: Thought about this - in that case we would be returning `undefined` from the method... not sure if that would be the best. Again, just stating my opinion - open to any kind of change.

packages/@webex/plugin-cc/test/unit/spec/services/task/TaskManager.ts
```diff
@@ -1087,5 +1087,22 @@ describe('TaskManager', () => {
     // The task should still exist in the collection based on current implementation
     expect(taskManager.getTask(taskId)).toBeDefined();
   });
+
+  it('should update task data on AGENT_WRAPUP event', () => {
```
Kesari3008, line 1091: Instead of adding this test, we should simply add a check in testcases where Unassiogned event and VTeamTransferred events are received followed by agent wrapup and ensure that cleanup is not attempted twice

adhmenon: Noted.

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
rsarika: I agree but,  i want to refactor this also to updatedata instead of returning new task and assigning it.

i mean task?.updateTaskData(task, payload.data); thats it. it should not return any task instead it will update the task.

This refactor i/we will do it as part of task manager(controls) refactor, so for now this is fine.

rsarika (approved):

packages/@webex/plugin-cc/src/services/task/TaskManager.ts
Kesari3008: ok we can revisit this

Kesari3008 (approved):

PR #4228, opened 2025-04-30
fix(plugin-meetings): fixed-user-policy-undefined-issue

adhmenon:
# COMPLETES #< AD-HOC ISSUE >

* In the meetings plugin, for a customer, they were seeing the following errors when joining the meeting - 
```
wx-js-sdk,Meeting:util#joinMeetingOptions --> Error joining the call, ,TypeError: Cannot read properties of undefined (reading 'supportWhiteboard')
    at Object.canShareWhiteBoard (https://unpkg.com/webex@next/umd/webex.min.js:2:2160210)

```
* The reason was because a recent PR had changed the `canShareWhiteBoard` method in the meeting utils.
* So, in plugin meetings, we use the `permissionToken` which we get from meeting info and decode it to get the user policies object which contains the various values regarding permissions. We have a util method in meetings to extract this.
* In that method, we were not checking for a non-null value of the user policy object and hence whenever it was undefined we got a crash.

* Simply adding both a check and a default value (empty for user policies) results in `false` value for undefined policieis.
* Unit tests also reflect the same fix.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Joined a meeting from the sdk sample app and ensured everything was working smoothly (with `null` permissionToken)

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/util.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/util.ts
@@ -575,6 +575,7 @@ const MeetingUtil = {
 
   canUserRenameOthers: (displayHints) => displayHints.includes(DISPLAY_HINTS.CAN_RENAME_OTHERS),
 
+  // Default value for policies in case we get an undefined value (ie permissionToken is not availableo)
   canShareWhiteBoard: (displayHints, policies = {}) =>
     displayHints.includes(DISPLAY_HINTS.SHARE_WHITEBOARD) &&
     !!policies[SELF_POLICY.SUPPORT_WHITEBOARD],
--- a/packages/@webex/plugin-meetings/src/meeting/util.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/util.ts
@@ -577,7 +577,6 @@ const MeetingUtil = {
 
   canShareWhiteBoard: (displayHints, policies = {}) =>
     displayHints.includes(DISPLAY_HINTS.SHARE_WHITEBOARD) &&
-    policies &&
     !!policies[SELF_POLICY.SUPPORT_WHITEBOARD],
 
   /**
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -575,8 +575,9 @@ const MeetingUtil = {
 
   canUserRenameOthers: (displayHints) => displayHints.includes(DISPLAY_HINTS.CAN_RENAME_OTHERS),
 
-  canShareWhiteBoard: (displayHints, policies) =>
+  canShareWhiteBoard: (displayHints, policies = {}) =>
     displayHints.includes(DISPLAY_HINTS.SHARE_WHITEBOARD) &&
+    policies &&
```
adhmenon, line 580: Unsure about this change - when I was trying unit tests, just adding this policies check was not working.. adding the default value worked overall.

packages/@webex/plugin-meetings/src/meeting/util.ts
```diff
@@ -575,8 +575,9 @@ const MeetingUtil = {
 
   canUserRenameOthers: (displayHints) => displayHints.includes(DISPLAY_HINTS.CAN_RENAME_OTHERS),
 
-  canShareWhiteBoard: (displayHints, policies) =>
+  canShareWhiteBoard: (displayHints, policies = {}) =>
```
adhmenon, line 579: This change alone was working fine - tested with unit tests and it works properly.
We have a similar logic elsewhere and it is working properly.

**Note** - Felt it was better than doing this - 
`
displayHints.includes(DISPLAY_HINTS.SHARE_WHITEBOARD) && policies ? !!policies[SELF_POLICY.SUPPORT_WHITEBOARD] : false
`

arun3528: should be fine , can you just add a comment above stating the permission token not available

arun3528 (approved):

rarajes2 (approved):
Looks good

PR #4231, opened 2025-05-02
feat(plugin-cc): added-unregister-for-cc

adhmenon:
# COMPLETES #< [CAI-6348](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6348) >

* In the webex cc sdk, we did not have a method for uders to unregister wdm, mobius or even the webex cc instance itself.
* Now, we have added a public method which will allow users to unregister the SDK and will in turn dergister mobius, close the web socket connection among others.

* Added an external method to perform unregister
* Added extensive logs to track it.

Vidcast - https://app.vidcast.io/share/3b7ac760-dab3-49a7-a471-59589774a873

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested unregister in the sample app and checked network logs for both BROWSER and EXTENSION mode.
* Tested out getting a call and logging in after unregister.

### I certified that

---


docs/samples/contact-center/app.js
```diff
@@ -850,10 +853,21 @@ function startStateTimer(lastStateChangeTimestamp, lastIdleCodeChangeTimestamp)
   }, 1000);
 }
 
+function updateUnregisterButtonState() {
```
adhmenon, line 856: We should not show the button when the user is logged in - by default this takes care of active tasks or anything else in turn.

docs/samples/contact-center/app.js
```diff
@@ -850,10 +853,21 @@ function startStateTimer(lastStateChangeTimestamp, lastIdleCodeChangeTimestamp)
...
+    !logoutAgentElm.classList.contains('hidden');
+  
+  unregisterBtn.disabled = isLoggedIn;  
+}
 
 function register() {
     webex.cc.register().then((agentProfile) => {
         registerStatus.innerHTML = 'Subscribed';
+        // Update button states upon successful registration
+        registerBtn.disabled = true;
```
adhmenon, line 867: Only show these buttons, inculding upload logs only when we register.

docs/samples/contact-center/app.js
```diff
@@ -938,6 +953,46 @@ function register() {
     
 }
 
+// New function to handle unregistration
+function doUnregister() {
```
adhmenon: Doing some UI cleanup on unregistration - like removing the dropdown values that we see.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
     }
   }
 
+  /**
+   * This is used to unregister the CC SDK and clean up all resources.
+   * @returns Promise<void>
+   * @throws Error
+   */
+  public async unregister(): Promise<void> {
```
adhmenon: Simple public method which does all the required tasks - 
1. Deletes the wdm device (same as meetings flow)
2. Stops websocket connection with backend
3. Stops mercury connection
4. Deletes mobius device, if BROWSER mode is enabled

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      LoggerProxy.log('CC SDK unregistered successfully', {
+        module: CC_FILE,
+        method: 'unregister',
+      });
+
+      this.metricsManager.trackEvent(METRIC_EVENT_NAMES.WEBSOCKET_UNREGISTER_SUCCESS, {}, [
+        'operational',
+      ]);
+    } catch (error) {
+      this.metricsManager.trackEvent(
```
adhmenon, line 220: Added metrics for all cases and extensive logging as well.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -379,12 +441,6 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
         this.webCallingService.deregisterWebCallingLine();
       }
 
-      // TODO: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-626777 Implement the de-register method and close the listener there
```
adhmenon, line 382: We don't need many comments anymore, hence removed them.

packages/@webex/plugin-cc/src/services/WebCallingService.ts
```diff
@@ -130,6 +130,11 @@ export default class WebCallingService extends EventEmitter {
   }
 
   public async deregisterWebCallingLine() {
+    LoggerProxy.log('Deregistering WebCalling line and cleaning up resources', {
```
adhmenon: Added this so we clean up the call whenever we de-register.

packages/@webex/plugin-cc/src/services/core/websocket/WebSocketManager.ts
```diff
@@ -59,7 +59,7 @@ export class WebSocketManager extends EventEmitter {
       this.shouldReconnect = shouldReconnect;
       this.websocket.close();
       this.keepaliveWorker.postMessage({type: 'terminate'});
-      LoggerProxy.error(
+      LoggerProxy.log(
```
adhmenon, line 62: This should be log and error since we only call it when we manually want to close the connection.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -1356,4 +1358,104 @@ describe('webex.cc', () => {
       await expect(webex.cc.uploadLogs('12345')).rejects.toThrow(error);
     });
   });
+
+  describe('unregister', () => {
```
adhmenon, line 1361: Added 2 tests - 
1. Testing if unregister calls all required methods
2. Seeing if websocket connection is removed and errors are handled proeprly.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      ]);
+
+      this.taskManager.off(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
+      this.taskManager.off(TASK_EVENTS.TASK_HYDRATE, this.handleTaskHydrate);
+
+      this.services.webSocketManager.off('message', this.handleWebSocketMessage);
+      this.services.connectionService.off('connectionLost', this.handleConnectionLost);
+
+      if (this.webCallingService) {
+        await this.webCallingService.deregisterWebCallingLine();
```
adhmenon: See this being called in logout also... I think I can remove it from there and just keep it here, but I added it here just to indicate this and discuss that doubt.
Open to any suggestion also.

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: nitpick: can we change the name to deregister

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+   * @throws Error
+   */
+  public async unregister(): Promise<void> {
+    try {
+      this.metricsManager.timeEvent([
+        METRIC_EVENT_NAMES.WEBSOCKET_UNREGISTER_SUCCESS,
+        METRIC_EVENT_NAMES.WEBSOCKET_UNREGISTER_FAILED,
+      ]);
+
+      this.taskManager.off(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
```
Kesari3008, line 180: There is one more listener we need to consider switching off. Please check invocation of this method: unregisterIncomingCallEvent. For webrtc calls, we are switching on this and need to be switched off

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -329,10 +395,6 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
         ['behavioral', 'business', 'operational']
       );
 
-      // TODO: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-626777 Implement the de-register method and close the listener there
```
Kesari3008, line 332: There is a comment in silentRelogin method for the de-register, please remove that too

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        METRIC_EVENT_NAMES.WEBSOCKET_UNREGISTER_FAILED,
+      ]);
+
+      this.taskManager.off(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
+      this.taskManager.off(TASK_EVENTS.TASK_HYDRATE, this.handleTaskHydrate);
+
+      this.services.webSocketManager.off('message', this.handleWebSocketMessage);
+      this.services.connectionService.off('connectionLost', this.handleConnectionLost);
+
+      if (this.webCallingService) {
```
Kesari3008: Yes let's keep it at one place but just something to think of: Webtrc case doesn't have multilogin support and having any cleanup logic done in logout was only causing issue with multilogin

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        await this.$webex.internal.mercury.disconnect();
+        // @ts-ignore
+        await this.$webex.internal.device.unregister();
+        LoggerProxy.log('Mercury disconnected successfully', {
+          module: CC_FILE,
+          method: 'unregister',
+        });
+      }
+
+      LoggerProxy.log('CC SDK unregistered successfully', {
```
Kesari3008: Just saying Deregistered Successfully should ne enough

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -1356,4 +1357,104 @@ describe('webex.cc', () => {
...
+      deregisterWebCallingLineSpy = jest.spyOn(
+        webex.cc.webCallingService,
+        'deregisterWebCallingLine'
+      ).mockResolvedValue();
+    });
+
+    it('should unregister successfully and clean up all resources', async () => {
+      await webex.cc.unregister();
+
+      expect(mockTaskManager.off).toHaveBeenCalledWith(TASK_EVENTS.TASK_INCOMING, expect.any(Function));
```
Kesari3008, line 1399: Along with having expect statements for off listeners being invoked, invoked the respectivae callback and add expect for them being invoked with right set of arguments

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+      // Clear any cached agent configuration
+      this.agentConfig = null;
+
+      if (this.$webex.internal.mercury.connected) {
+        this.$webex.internal.mercury.off('online');
+        this.$webex.internal.mercury.off('offline');
+        await this.$webex.internal.mercury.disconnect();
+        // @ts-ignore
+        await this.$webex.internal.device.unregister();
```
Kesari3008: Wdm registration and deregistration is only done webrtc cases I believe so some checks here to   do conditional cleanups needs to be added because this deregister may start failing for reasons where that failure may not be valid too. And based on those conditions, your testcases should be updated too

Kesari3008 (changes requested):

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Right. I think I can retain it in the `logout`. After all, we do not support multi login for webrtc calls.
When we do change it, we can refactor our logic (we will still need the if check to call the `unregisterIncomingCallEvent`, so it would be an easy refactor)

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      ]);
+
+      this.taskManager.off(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
+      this.taskManager.off(TASK_EVENTS.TASK_HYDRATE, this.handleTaskHydrate);
+
+      this.services.webSocketManager.off('message', this.handleWebSocketMessage);
+      this.services.connectionService.off('connectionLost', this.handleConnectionLost);
+
+      if (this.agentConfig.webRtcEnabled) {
+        this.taskManager.unregisterIncomingCallEvent();
```
Kesari3008: This listener is not webrtc dependent, please move it out of the condition

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,73 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+        if (this.$webex.internal.mercury.connected) {
+          this.$webex.internal.mercury.off('online');
+          this.$webex.internal.mercury.off('offline');
+          await this.$webex.internal.mercury.disconnect();
+          // @ts-ignore
+          await this.$webex.internal.device.unregister();
```
Kesari3008, line 196: We should put login option condition combined with webrtcenabled to do wdm unregister

adhmenon:
Tested with amplify link.
Everything is working good. Tetsed with different modes and deregister works fine.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -1356,4 +1357,143 @@ describe('webex.cc', () => {
...
+        method: 'deregister',
+      });
+
+      const taskIncomingOff = mockTaskManager.off.mock.calls.find(([evt]) => evt === TASK_EVENTS.TASK_INCOMING);
+      expect(taskIncomingOff[1]).toBe(webex.cc['handleIncomingTask']);
+
+      const taskHydrateOff = mockTaskManager.off.mock.calls.find(([evt]) => evt === TASK_EVENTS.TASK_HYDRATE);
+      expect(taskHydrateOff[1]).toBe(webex.cc['handleTaskHydrate']);
+
+      const wsMessageOff = mockWebSocketManager.off.mock.calls.find(([evt]) => evt === 'message');
```
Kesari3008: We are still not invoking the callback. You have to fetch the callback using like mock.calls[0][1]. Indexes have to be correct here and then invoke it. And then have an expect statement for comparing the action that would have happened inside the callback function.

Reference: https://github.com/webex/webex-js-sdk/blob/f027038dcce588cd846cf19994fbf8c4f3a55e56/packages/%40webex/plugin-cc/test/unit/spec/cc.ts#L556

Kesari3008 (approved):

packages/@webex/plugin-cc/src/services/WebCallingService.ts
Shreyas281299: Can the texts come from a const file? Since we are already doing that.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,75 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      });
+
+      this.metricsManager.trackEvent(METRIC_EVENT_NAMES.WEBSOCKET_DEREGISTER_SUCCESS, {}, [
+        'operational',
+      ]);
+    } catch (error) {
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.WEBSOCKET_DEREGISTER_FAIL,
+        {
+          error: error.message || 'Unknown error',
```
Shreyas281299: The text can be moved to a const

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -164,6 +163,75 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        this.agentConfig.webRtcEnabled &&
+        this.agentConfig.loginVoiceOptions.includes(LoginOption.BROWSER)
+      ) {
+        if (this.$webex.internal.mercury.connected) {
+          this.$webex.internal.mercury.off('online');
+          this.$webex.internal.mercury.off('offline');
+          await this.$webex.internal.mercury.disconnect();
+          // @ts-ignore
+          await this.$webex.internal.device.unregister();
+          LoggerProxy.log('Mercury disconnected successfully', {
```
Shreyas281299: Text to come from const file

Shreyas281299 (approved):
Looks good to me just before merging lets move any NEW added text to a const file and use it from there

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Valid point, however I am a little skeptical about moving logs to a constant file - I feel that file should only be for string constants and logs can be left as it is.

PR #4239, opened 2025-05-06
fix(contact-center): add buffer to window object

mkesavan13:
# PARTIALLY COMPLETES [CAI-6287](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6287)

The `Buffer` class was missing in the window object after moving away from `webex.min.js`. The `webex.min.js` had the encryption plugin included which included the `node-jose` package and this package ensured that the `Buffer` class is attached to the global/window scope.

In the `contact-center.min.js`, this was missing because we do not need encryption-plugin, neither the whole `node-jose` library. This PR addresses that.

### The impact

The real impact we had out of this was that the oAuth Login happens but the SDK wasn't initializing. This is because, every time the SDK initializes, it checks the URL for any hashes and ensure the accessToken query param is removed from it if found. While doing this, one of the depndent libraries uses a Buffer object to decode the base64 text and this Buffer class not being available on the Window broke the application and did not let it proceed to the next steps

Import `safe-buffer` polyfill package in `contact-center.js` and then attach it to the `global` object. 

### Screen Recording

This Vidcast explains E2E of what's the root cause and what was the fix and also why it was working in the Meetings Sample app:

https://app.vidcast.io/share/86dd9707-3a47-4779-8db9-61f8fe25c8c3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Did a oAuth Login on CC Samples
* Completed Station Login and tried to update the User state as well

### I certified that

---


adhmenon (approved):

sreenara (commented):
@mkesavan13  Please record a vidcast instead of a screen recording. The screen recording will get auto deleted after a few months.
Also please add the test scenarios.
In the description, it doesn't mention how the Buffer class missing caused OAuth login failure. Can we talk about that as well?

mkesavan13:
> @mkesavan13 Please record a vidcast instead of a screen recording. The screen recording will get auto deleted after a few months. Also please add the test scenarios. In the description, it doesn't mention how the Buffer class missing caused OAuth login failure. Can we talk about that as well?

@sreenara - This is done. Please check

sreenara (approved):

PR #4099, opened 2025-02-14
feat(plugin-encryption): Create new plugin-encryption

bhabalan:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-580068

## This pull request addresses

Creation of plugin-encryption

## by making the following changes

Added new plugin-encryption

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios were tested

Tested that we can download and decrypt files using the plugin

Vidcast: https://app.vidcast.io/share/946359bd-0c9a-4c80-ba71-c788d0183703

### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #4244, opened 2025-05-07
fix(encryption): fix issue with fetching files service url

bhabalan:
# COMPLETES ADHOC

Fixes an issue introduced by #4099 where web client is not able to download encrypted files

Added a null check and removed async keyword from

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested against web client and confirmed that downloads work

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-encryption/src/encryption.js
+++ b/packages/@webex/internal-plugin-encryption/src/encryption.js
@@ -72,7 +72,7 @@ const Encryption = WebexPlugin.extend({
    * @param {Object} options - optional parameters to download a file
    * @returns {promise}
    */
-  async download(fileUrl, scr, options) {
+  download(fileUrl, scr, options) {
     /* istanbul ignore if */
     if (!fileUrl || !scr) {
       return Promise.reject(new Error('`scr` and `fileUrl` are required'));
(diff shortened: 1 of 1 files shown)


arun3528 (approved):
its requires unit test , you can add them as seperate PR its its urgent

packages/@webex/internal-plugin-encryption/src/encryption.js
```diff
@@ -140,12 +140,13 @@ const Encryption = WebexPlugin.extend({
...
       method: 'POST',
       uri: url.format(endpointUrl),
-      body: options
-        ? {
-            ...inputBody,
-            allow: options.params.allow,
-          }
-        : inputBody,
+      body:
+        options?.params?.allow != null
```
chrisadubois, line 144: shouldn't this just be simplified to a truthy or an explicit true value?

i.e., 

`if (options?.params?.allow === true)` OR `if (options?.params?.allow)` OR conversely `if (!options?.params?.allow)`

doin a strict null check seems worrisome because params.allow by default is `undefined` or falsy, and really the only time we want `allow` to be passed through is when it's explicitly set, which would be either `true` or `false` explicitly.

chrisadubois (approved):
Approved because this will unblock the pipeline, but I would request a different check other than `null` if it does infact make sense to do so.

packages/@webex/internal-plugin-encryption/src/encryption.js
sreenara: @bhabalan, curious about the cause. What changed about this `allow` parameter in #4099 ?

bhabalan: @sreenara 
The existing check that was not a part of #4099 relied on whether 'options' is passed in or not. It was not doing null safe access to get the 'allow' field.

 With #4099 , we introduced a field called 'useFileService' which was always being passed from the download() method. Since now the options object will always exist and not necessarily have the 'allow' field, we were trying to access non-existent keys. So an error was being thrown.

PR #4265, opened 2025-05-13
feat(plugin-cc): added-update-profile

adhmenon:
# COMPLETES #<[CAI-6447](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6447)>

* As per EPIC requirements, we need an option for the user to update their login type (ie profile) whilst they are already logged in, rather than _manually_ logging out and logging back in.
* SDK did not _directly_ support such a functionality.

* In the SDK, we expose a public method which takes in `LoginOption` as a data and it includes the new login type.
* This method internally calls logout and login successively and send correct metrics.
* Sample app changes and unit tests changes to incorporate this method and fix few issues.

Vidcast - https://app.vidcast.io/share/5f69a914-fafc-4184-ac3e-761fd77ff173

SS of metrics - 
Screenshot 2025-05-13 at 4 37 21 PM

SS of error with clean logs - 
Screenshot 2025-05-14 at 8 43 49 PM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Manually logged in / logged out and tried to change status.
* Changed the profile and tried to change status
* Changed profile and recieved calls on both EXTENSION and BROWSER
* Tried out consult and transfer flows on updated profile.

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -1132,7 +1132,6 @@ function logoutAgent() {
 
 async function updateAgentDeviceType() {
   const payload = {
-    teamId: teamsDropdown.value,
     loginOption: agentDeviceType,
     dialNumber: dialNumber.value
   };
@@ -1154,7 +1153,6 @@ async function applyupdateAgentDeviceType() {
   const loginOption = updateLoginOptionElm.value;
   const newDial = loginOption === 'BROWSER' ? '' : updateDialNumberElm.value;
   const payload = {
-    teamId: teamsDropdown.value,
     loginOption,
     dialNumber: newDial,
   };
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -35,8 +35,8 @@ const idleCodesDropdown = document.querySelector('#idleCodesDropdown')
 const setAgentStatusButton = document.querySelector('#setAgentStatus');
 const logoutAgentElm = document.querySelector('#logoutAgent');
 const buddyAgentsDropdownElm = document.getElementById('buddyAgentsDropdown');
-const updateProfileElm = document.querySelector('#updateProfile');
-const updateFieldsContainer = document.querySelector('#updateProfileFields');
+const updateAgentDeviceTypeElm = document.querySelector('#updateAgentDeviceType');
+const updateFieldsContainer = document.querySelector('#updateAgentDeviceTypeFields');
 const updateLoginOptionElm = document.querySelector('#updateLoginOption');
 const updateDialNumberElm = document.querySelector('#updateDialNumber');
 const incomingCallListener = document.querySelector('#incomingsection');
@@ -941,7 +941,7 @@ function register() {
       console.log('Agent re-login successful', data);
       loginAgentElm.disabled = true;
       logoutAgentElm.classList.remove('hidden');
-      updateProfileElm.classList.remove('hidden');
+      updateAgentDeviceTypeElm.classList.remove('hidden');
 
       agentLogin.value = data.deviceType;
       agentDeviceType = data.deviceType;
@@ -960,7 +960,7 @@ function register() {
       console.log('Agent station-login success', data);
       loginAgentElm.disabled = true;
       logoutAgentElm.classList.remove('hidden');
-      updateProfileElm.classList.remove('hidden');
+      updateAgentDeviceTypeElm.classList.remove('hidden');
       updateFieldsContainer.classList.add('hidden');
 
       agentLogin.value = data.deviceType;
@@ -1066,7 +1066,7 @@ function doAgentLogin() {
     console.log('Agent Logged in successfully', response);
     loginAgentElm.disabled = true;
     logoutAgentElm.classList.remove('hidden');
-    updateProfileElm.classList.remove('hidden');
+    updateAgentDeviceTypeElm.classList.remove('hidden');
     // Read auxCode and lastStateChangeTimestamp from login response
     const DEFAULT_CODE = '0'; // Default code when no aux code is present
     const auxCodeId = response.data.auxCodeId?.trim() !== '' ? response.data.auxCodeId : DEFAULT_CODE;
@@ -1104,7 +1104,7 @@ function logoutAgent() {
     .then((response) => {
       console.log('Agent logged out successfully', response);
       loginAgentElm.disabled = false;
-      updateProfileElm.classList.add('hidden');
+      updateAgentDeviceTypeElm.classList.add('hidden');
       updateFieldsContainer.classList.add('hidden');
 
      // Clear the timer when the agent logs out.
@@ -1129,14 +1129,13 @@ function logoutAgent() {
   });
 }
 
-async function updateProfile() {
+async function updateAgentDeviceType() {
   const payload = {
-    teamId: teamsDropdown.value,
     loginOption: agentDeviceType,
     dialNumber: dialNumber.value
   };
   try {
-    const response = await webex.cc.updateProfile(payload);
+    const response = await webex.cc.updateAgentDeviceType(payload);
     console.log('Profile updated successfully', response);
   }
   catch (error) {
@@ -1145,20 +1144,19 @@ async function updateProfile() {
   }
 }
 
-function showUpdateProfileUI() {
+function showupdateAgentDeviceTypeUI() {
   updateFieldsContainer.classList.toggle('hidden');
 }
 
-async function applyUpdateProfile() {
+async function applyupdateAgentDeviceType() {
   const loginOption = updateLoginOptionElm.value;
   const newDial = loginOption === 'BROWSER' ? '' : updateDialNumberElm.value;
   const payload = {
-    teamId: teamsDropdown.value,
     loginOption,
     dialNumber: newDial,
   };
   try {
-    const resp = await webex.cc.updateProfile(payload);
+    const resp = await webex.cc.updateAgentDeviceType(payload);
     console.log('Profile updated', resp);
     updateFieldsContainer.classList.add('hidden');
     // Reflect new values in main UI
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -1132,6 +1132,7 @@ function logoutAgent() {
 
 async function updateAgentDeviceType() {
   const payload = {
+    teamId: teamsDropdown.value,
     loginOption: agentDeviceType,
     dialNumber: dialNumber.value
   };
@@ -1153,6 +1154,7 @@ async function applyupdateAgentDeviceType() {
   const loginOption = updateLoginOptionElm.value;
   const newDial = loginOption === 'BROWSER' ? '' : updateDialNumberElm.value;
   const payload = {
+    teamId: teamsDropdown.value,
     loginOption,
     dialNumber: newDial,
   };
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -35,6 +35,10 @@ const idleCodesDropdown = document.querySelector('#idleCodesDropdown')
 const setAgentStatusButton = document.querySelector('#setAgentStatus');
 const logoutAgentElm = document.querySelector('#logoutAgent');
 const buddyAgentsDropdownElm = document.getElementById('buddyAgentsDropdown');
+const updateProfileElm = document.querySelector('#updateProfile');
+const updateFieldsContainer = document.querySelector('#updateProfileFields');
+const updateLoginOptionElm = document.querySelector('#updateLoginOption');
+const updateDialNumberElm = document.querySelector('#updateDialNumber');
 const incomingCallListener = document.querySelector('#incomingsection');
 const incomingDetailsElm = document.querySelector('#incoming-task');
 const answerElm = document.querySelector('#answer');
@@ -873,21 +877,12 @@ function register() {
             teamsDropdown.add(option);
         });
         const loginVoiceOptions = agentProfile.loginVoiceOptions;
-        agentLogin.innerHTML = '<option value="" selected>Choose Agent Login ...</option>'; // Clear previously selected option on agentLogin.
+        populateLoginOptions(
+          loginVoiceOptions.filter((o) => agentProfile.webRtcEnabled || o !== 'BROWSER')
+        );
         dialNumber.value = agentProfile.defaultDn ? agentProfile.defaultDn : '';
         dialNumber.disabled = agentProfile.defaultDn ? false : true;
         if (loginVoiceOptions.length > 0) loginAgentElm.disabled = false;
-        loginVoiceOptions.forEach((voiceOptions)=> {
-          if (!agentProfile.webRtcEnabled && voiceOptions === 'BROWSER') {
-            // Skiping the addition of browser option for webrtc disabled case
-            return;
-          }
-          const option = document.createElement('option');
-          option.text = voiceOptions;
-          option.value = voiceOptions;
-          agentLogin.add(option);
-          option.selected = agentProfile.isAgentLoggedIn && voiceOptions === agentProfile.deviceType;
-        });
 
         if (agentProfile.isAgentLoggedIn) {
           loginAgentElm.disabled = true;
@@ -942,10 +937,48 @@ function register() {
       }
     });
 
-    webex.cc.on('agent:stationLoginSuccess', (data) => {
-      console.log('Agent station login success', data);
+    webex.cc.on('agent:reloginSuccess', (data) => {
+      console.log('Agent re-login successful', data);
+      loginAgentElm.disabled = true;
+      logoutAgentElm.classList.remove('hidden');
+      updateProfileElm.classList.remove('hidden');
+
+      agentLogin.value = data.deviceType;
+      agentDeviceType = data.deviceType;
+
+      if (data.deviceType === 'BROWSER') {
+        dialNumber.disabled = true;
+        dialNumber.value = '';
+      }
+      else {
+        dialNumber.disabled = false;
+        dialNumber.value = data.dn || '';
+      }
     });
-    
+
+    webex.cc.on('agent:stationLoginSuccess', (data) => {
+      console.log('Agent station-login success', data);
+      loginAgentElm.disabled = true;
+      logoutAgentElm.classList.remove('hidden');
+      updateProfileElm.classList.remove('hidden');
+      updateFieldsContainer.classList.add('hidden');
+
+      agentLogin.value = data.deviceType;
+      agentDeviceType = data.deviceType;
+      if (data.deviceType === 'BROWSER') {
+        dialNumber.disabled = true;
+        dialNumber.value = '';
+      }
+      else {
+        dialNumber.disabled = false;
+        dialNumber.value = data.dn || '';
+      }
+
+      const auxId  = data.auxCodeId?.trim() || '0';
+      const idx    = [...idleCodesDropdown.options].findIndex(o => o.value === auxId);
+      idleCodesDropdown.selectedIndex = idx >= 0 ? idx : 0;
+      startStateTimer(data.lastStateChangeTimestamp, data.lastIdleCodeChangeTimestamp);
+    });
 }
 
 // New function to handle unregistration
@@ -1028,12 +1061,12 @@ function doAgentLogin() {
     teamId: teamsDropdown.value,
     loginOption: agentDeviceType,
     dialNumber: dialNumber.value
-  }).then((response) => {
+  })
+  .then((response) => {
     console.log('Agent Logged in successfully', response);
     loginAgentElm.disabled = true;
     logoutAgentElm.classList.remove('hidden');
-    updateUnregisterButtonState();
-    
+    updateProfileElm.classList.remove('hidden');
     // Read auxCode and lastStateChangeTimestamp from login response
     const DEFAULT_CODE = '0'; // Default code when no aux code is present
     const auxCodeId = response.data.auxCodeId?.trim() !== '' ? response.data.auxCodeId : DEFAULT_CODE;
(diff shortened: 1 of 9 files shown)


docs/samples/contact-center/app.js
```diff
@@ -35,6 +35,10 @@ const idleCodesDropdown = document.querySelector('#idleCodesDropdown')
 const setAgentStatusButton = document.querySelector('#setAgentStatus');
 const logoutAgentElm = document.querySelector('#logoutAgent');
 const buddyAgentsDropdownElm = document.getElementById('buddyAgentsDropdown');
+const updateProfileElm = document.querySelector('#updateProfile');
```
adhmenon: Adding new elements for the update profile area in sample app.
Open to suggestions if we want to make it look different (it is shown in the vidcast)

docs/samples/contact-center/app.js
```diff
@@ -873,21 +877,12 @@ function register() {
             teamsDropdown.add(option);
         });
         const loginVoiceOptions = agentProfile.loginVoiceOptions;
-        agentLogin.innerHTML = '<option value="" selected>Choose Agent Login ...</option>'; // Clear previously selected option on agentLogin.
+        populateLoginOptions(
```
adhmenon, line 881: This is to ensure that we change the login options after we login, so that we can see the changes of the update profile on login options by using the older login / logout event flows and not having to make more changes.

docs/samples/contact-center/app.js
```diff
@@ -942,10 +937,48 @@ function register() {
       }
     });
 
+    webex.cc.on('agent:reloginSuccess', (data) => {
```
adhmenon, line 941: We need to listen to this event so that it has the latets and updated login options values. This is incidentally a new event that we are emitting after the event refactor.

docs/samples/contact-center/app.js
```diff
@@ -942,10 +937,48 @@ function register() {
...
+      }
+      else {
+        dialNumber.disabled = false;
+        dialNumber.value = data.dn || '';
+      }
+    });
+
     webex.cc.on('agent:stationLoginSuccess', (data) => {
-      console.log('Agent station login success', data);
+      console.log('Agent station-login success', data);
```
adhmenon, line 961: We need to consume this event also so that we can now change the UI of login after a success - earlier we were relying only on promises but it would not work for update profile, hence I added it here.
Good case of how it can be leveraged in widgets also.

docs/samples/contact-center/app.js
```diff
@@ -1067,9 +1100,12 @@ function setAgentStatus() {
...
 function logoutAgent() {
-  webex.cc.stationLogout({logoutReason: 'logout'}).then((response) => {
-    console.log('Agent logged out successfully', response);
-    loginAgentElm.disabled = false;
+  webex.cc.stationLogout({logoutReason: 'logout'})
+    .then((response) => {
+      console.log('Agent logged out successfully', response);
+      loginAgentElm.disabled = false;
+      updateProfileElm.classList.add('hidden');
+      updateFieldsContainer.classList.add('hidden');
```
adhmenon, line 1109: Only show the button once we have logged in.

docs/samples/contact-center/app.js
```diff
@@ -1093,6 +1129,50 @@ function logoutAgent() {
   });
 }
 
+async function updateProfile() {
```
adhmenon: Standard methods to call the public API `webex.cc.updateProfile`, and modify and show the custom UI built for it.

docs/samples/contact-center/app.js
```diff
@@ -1557,3 +1637,20 @@ function loadEmailWidget(task) {
     ></imi-email-composer>
   `;
 }
+
+function populateLoginOptions(options) {
```
adhmenon, line 1642: As explained above adding this method here so it can be reused - we are calling it from the `register` method also so it does not break anything.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +879,61 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
   public async uploadLogs(): Promise<UploadLogsResponse> {
     return this.webexRequest.uploadLogs();
   }
+
+  public async updateProfile(data: AgentLogin): Promise<StationLoginResponse> {
```
adhmenon: This is the public method.
I have added the `AgentLogin` as the parameter - felt it would be best to take in the entire object as we require it for AQM. The end user can simply update the new login type over here.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +879,61 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    ]);
+
+    LoggerProxy.info('updateProfile | starting profile update', {
+      module: CC_FILE,
+      method: 'updateProfile',
+    });
+
+    try {
+      // Throw an error if the loginOption is of the same type as the current device type
+      if (this.webCallingService && this.webCallingService.loginOption === data.loginOption) {
```
adhmenon: As mentioned, we have a check so that the user does not provide the same option (why waste time logging out and loggin back in that case). We can directly just reject it over here.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +879,61 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+      LoggerProxy.log('updateProfile | profile updated successfully', {
+        module: CC_FILE,
+        method: 'updateProfile',
+      });
+
+      return resp;
+    } catch (error) {
+      const failure = error.details as Failure;
+      this.metricsManager.trackEvent(
```
adhmenon, line 971: As discussed, adding metrics here. Since the response we get from login is essentially an AQM respons, the below will work.
Have posted screenshot of the metric being sent also.

packages/@webex/plugin-cc/test/unit/spec/cc.ts
```diff
@@ -1610,4 +1610,93 @@ describe('webex.cc', () => {
       });
     });
   });
+
+  describe('updateProfile', () => {
```
adhmenon: Tested out 3 major cases
1. Working of `updateProfile` in all cases and seeing the changes
2. Error handling of the same method
3. Testing of duplicate / same login type also.

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: We can build the object within our public method and pass to AQM. All user has to pass is new login mode and extension or dial number detail

Kesari3008: Also updateProfile is little confusing name, maybe we should use something more accurate

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +879,61 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    });
+
+    try {
+      // Throw an error if the loginOption is of the same type as the current device type
+      if (this.webCallingService && this.webCallingService.loginOption === data.loginOption) {
+        throw new Error('Device type is same as current device type');
+      }
+
+      await this.stationLogout({
+        logoutReason: 'User requested logout',
```
Kesari3008: Here we can mention different reason saying user chagning the station login mode or something more appropriate

packages/@webex/plugin-cc/src/metrics/constants.ts
```diff
@@ -50,6 +50,9 @@ export const METRIC_EVENT_NAMES = {
   UPLOAD_LOGS_FAILED: 'Upload Logs Failed',
   WEBSOCKET_DEREGISTER_SUCCESS: 'Websocket Deregister Success',
   WEBSOCKET_DEREGISTER_FAIL: 'Websocket Deregister Failed',
+
+  PROFILE_UPDATE_SUCCESS: 'Profile Update Success',
```
Kesari3008: More than profile update, it would be more accurate to call it StationLoginUpdate or something similar

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +879,61 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+  public async updateProfile(data: AgentLogin): Promise<StationLoginResponse> {
+    this.metricsManager.timeEvent([
+      METRIC_EVENT_NAMES.PROFILE_UPDATE_SUCCESS,
+      METRIC_EVENT_NAMES.PROFILE_UPDATE_FAILED,
+    ]);
+
+    LoggerProxy.info('updateProfile | starting profile update', {
+      module: CC_FILE,
+      method: 'updateProfile',
```
Kesari3008: You can add the method name like this: method: this.updateProfile.name,

Kesari3008 (changes requested):

packages/@webex/plugin-cc/src/metrics/constants.ts
```diff
@@ -50,6 +50,9 @@ export const METRIC_EVENT_NAMES = {
   UPLOAD_LOGS_FAILED: 'Upload Logs Failed',
   WEBSOCKET_DEREGISTER_SUCCESS: 'Websocket Deregister Success',
   WEBSOCKET_DEREGISTER_FAIL: 'Websocket Deregister Failed',
+
+  AGENT_DEVICE_TYPE_UPDATE_SUCCESS: 'Profile Update Success',
```
bhabalan: The value should match the key

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +880,68 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+    LoggerProxy.info('updateAgentDeviceType | starting profile update', {
+      module: CC_FILE,
+      method: this.updateAgentDeviceType.name,
+    });
+
+    try {
+      // ensure we change device type
+      if (this.webCallingService && this.webCallingService.loginOption === data.loginOption) {
+        throw new Error('Device type is same as current device type');
```
bhabalan: Why do we need to throw?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -879,4 +880,68 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
   public async uploadLogs(): Promise<UploadLogsResponse> {
     return this.webexRequest.uploadLogs();
   }
+
+  public async updateAgentDeviceType(data: AgentDeviceUpdate): Promise<StationLoginResponse> {
```
bhabalan: Does the return type have to be Promise<StationLoginResponse> ?

adhmenon: Yes, we need to wait for it as otherwise no other operation can happen.

packages/@webex/plugin-cc/src/cc.ts
adhmenon: I think it's important to throw in this case so users know they cannot pass the same type.

packages/@webex/plugin-cc/src/cc.ts
bhabalan: Sorry, I meant to ask why does the return type have to be `Promise<StationLoginResponse>` ?

packages/@webex/plugin-cc/src/services/agent/types.ts
```diff
@@ -108,7 +108,16 @@ export type StationLoginSuccessResponse = {
   notifsTrackingId: string;
 };
 
-export type Logout = {logoutReason?: 'User requested logout' | 'Inactivity Logout'};
+export type DeviceTypeUpdateSuccess = Omit<StationLoginSuccessResponse, 'type'> & {
```
adhmenon, line 111: New type, we simply use the same structure as `StationLoginSuccessResponse` but we change the type.

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -284,3 +287,4 @@ export type StationLogoutResponse = Agent.LogoutSuccess | Error;
 export type StationReLoginResponse = Agent.ReloginSuccess | Error;
 export type SetStateResponse = Agent.StateChangeSuccess | Error;
 export type BuddyAgentsResponse = Agent.BuddyAgentsSuccess | Error;
+export type UpdateDeviceTypeResponse = Agent.DeviceTypeUpdateSuccess | Error;
```
adhmenon, line 303: Following the same format, this is the type we will use. Have added it to the comments also - we will document this also.

bharath063 (approved):

bhabalan (approved):

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -1,19 +1,22 @@
 import {WebexPlugin} from '@webex/webex-core';
 import EventEmitter from 'events';
+import {v4 as uuidv4} from 'uuid';
```
adhmenon, line 3: Need this for the tracking id - if not now, we would have used it in the future, hence I thought I might as well add it now.

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -246,6 +246,9 @@ export type AgentLogin = {
 
   loginOption: LoginOption;
 };
+
+export type AgentDeviceUpdate = Pick<AgentLogin, 'teamId' | 'loginOption' | 'dialNumber'>;
```
adhmenon, line 250: We need the teamId, because we are not persisting this in the SDK, no clean way to persist also, but the type can remain here as it can be used to differentiate between `AgentLogin` and updating.

cc @Kesari3008

adhmenon:
Final vidcast after updating with stored team id - https://app.vidcast.io/share/34a4345c-0b45-4738-ad4c-1cb43bba8813

packages/@webex/plugin-cc/src/services/config/types.ts
```diff
@@ -615,6 +615,7 @@ export type Profile = {
   tenantTimezone?: string;
   loginVoiceOptions?: LoginOption[];
   deviceType?: LoginOption;
+  currentTeamId?: string;
```
adhmenon, line 618: @Kesari3008 I have added currentTeamId here itself, this gets reset on login or relogin so it can be used for updating.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        ['behavioral', 'business', 'operational']
+      );
+
+      LoggerProxy.error(`updateAgentDeviceType | error updating profile: ${error}`, {
+        module: CC_FILE,
+        method: this.updateAgentDeviceType.name,
+      });
+      throw error;
+    }
+  }
```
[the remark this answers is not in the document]
adhmenon: This will not be an issue - we cannot have login and relogin happening at the same time.

packages/@webex/plugin-cc/src/services/agent/types.ts
Kesari3008: Why can't we use same response type ?

packages/@webex/plugin-cc/src/metrics/behavioral-events.ts
```diff
@@ -271,6 +271,20 @@ const eventTaxonomyMap: Record<string, BehavioralEventTaxonomy> = {
...
+  [METRIC_EVENT_NAMES.AGENT_DEVICE_TYPE_UPDATE_SUCCESS]: {
+    product,
+    agent: 'user',
+    target: 'profile_update',
+    verb: 'complete',
+  },
+  [METRIC_EVENT_NAMES.AGENT_DEVICE_TYPE_UPDATE_FAILED]: {
+    product,
+    agent: 'user',
+    target: 'profile_update',
```
Kesari3008: target also should say device type update

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      module: CC_FILE,
+      method: this.updateAgentDeviceType.name,
+    });
+
+    try {
+      // ensure we change device type
+      if (this.webCallingService?.loginOption === data.loginOption) {
+        const message = 'New Device type is same as current device type';
+        const trackingId = `WX_CC_SDK_${uuidv4()}`;
+        const err: any = new Error(message);
```
Kesari3008: Would it be possible to assign a type this error object here. Using any is not recommended

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+    LoggerProxy.info('updateAgentDeviceType | starting profile update', {
+      module: CC_FILE,
+      method: this.updateAgentDeviceType.name,
+    });
+
+    try {
+      // ensure we change device type
+      if (this.webCallingService?.loginOption === data.loginOption) {
+        const message = 'New Device type is same as current device type';
```
Kesari3008: Also this is something I observed in Agent desktop that it doesn't let you submit the selection if it's the same device type so throwing error for the same, is it the right way to handle it. Or if you wanna throw the error the maybe in the message we should not say it failed but something like we won;t proceed with updating device type as there is no change in it

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        const trackingId = `WX_CC_SDK_${uuidv4()}`;
+        const err: any = new Error(message);
+        err.details = {
+          type: 'Identical Device Change Failure',
+          orgId: this.$webex.credentials.getOrgId(),
+          trackingId,
```
Kesari3008, line 927: Having trackingId one time should be enough

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+          data: {
+            agentId: this.agentConfig.agentId,
+            trackingId,
+            reasonCode: 'R002',
```
Kesari3008, line 930: How did we come up with this reason code?

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +878,109 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.AGENT_DEVICE_TYPE_UPDATE_FAILED,
+        {
+          ...MetricsManager.getCommonTrackingFieldForAQMResponseFailed(failure),
+          loginType: data.loginOption,
+        },
+        ['behavioral', 'business', 'operational']
+      );
+
+      LoggerProxy.error(`updateAgentDeviceType | error updating profile: ${error}`, {
```
Kesari3008: It would be good to start adding trackingId in our logs and even metrics

Kesari3008 (changes requested):

packages/@webex/plugin-cc/src/cc.ts
adhmenon: That would be something to be handled on the widgets, yes.
I can do one thing I want to throw the error just so developers who are experimenting with it can identify - the other details can be taken care of in the app consuming the SDK.
I can modify the message to - "Will not proceed with device update as new Device type is same as current device type"

packages/@webex/plugin-cc/src/cc.ts
adhmenon: As of now it is a dummy value (I noticeed R001 somewhere else), we can modify it later on imo.
The key here is tracking id.

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Metrics will have the tracking id (for station login it gets it from the tracking id of the response). Will add it to the logs.

packages/@webex/plugin-cc/src/services/agent/types.ts
adhmenon: For one, it is easier for end users to demarcate the difference between update and station login. Felt we should keep that as our standard.
Secondly, in the `StationLoginResponse` we have a `type` parameter which has the value `AgentLoginSuccess`, this seemed confusing to me as end users would not understand it easily. Hence I changed the type and we are modifying the `type` parameter to a different value.

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Added the type - we can reuse it throughout sdk when we improve our logging.

packages/@webex/plugin-cc/src/cc.ts
Kesari3008: Please add a comment to mention the same that it's a dummy value and maybe we can come up with some meaningful reason codes and use
 that

packages/@webex/plugin-cc/src/services/agent/types.ts
Kesari3008: I feel if the payload or data is same then creating multiple types is increasing the work for the consumer to keep track of different types. If type does say AgentLoginSuccess then should be ok right because login is successful with new type so should be fine.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +879,113 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      });
+
+      const deviceTypeUpdateResponse: UpdateDeviceTypeResponse = {
+        ...resp,
+        type: 'AgentDeviceTypeUpdateSuccess',
+      };
+
+      return deviceTypeUpdateResponse;
+    } catch (error) {
+      const failure = (error as any).details as Failure;
```
Kesari3008: Update any here also to proper error type

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -873,10 +879,113 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+
+    try {
+      // ensure we change device type
+      if (this.webCallingService?.loginOption === data.loginOption) {
+        const message =
+          'Will not proceed with device update as new Device type is same as current device type';
+        const err = new Error(message) as GenericError;
+        err.details = {
+          type: 'Identical Device Change Failure',
+          orgId: this.$webex.credentials.getOrgId(),
```
Kesari3008, line 926: Single orgid also should be enough

Kesari3008 (approved):
Approving with few comments

PR #4266, opened 2025-05-13
fix(plugin-meetings): subnet reachability enhancement

k-wasniowski:
# COMPLETES #[SPARK-661903](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-661903)

Ensures that isSubnetReachable failure will be marked as false only if any clusters were reachable. Otherwise return null, since such cases will be marked as JMF skip anyway.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested with JS-SDK sample app

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -7159,12 +7159,18 @@ export default class Meeting extends StatelessWebexPlugin {
           },
           options: {
             meetingId: this.id,
+            rawError: error,
           },
         });
       }
-      throw new Error(
+
+      const timedOutError = new Error(
         `Timed out waiting for media connection to be connected, correlationId=${this.correlationId}`
       );
+
+      timedOutError.cause = error;
+
+      throw timedOutError;
     }
   }
 
@@ -7225,6 +7231,9 @@ export default class Meeting extends StatelessWebexPlugin {
           ROAP_OFFER_ANSWER_EXCHANGE_TIMEOUT / 1000
         } seconds`
       );
+
+      const error = new Error('Timed out waiting for REMOTE SDP ANSWER');
+
       // @ts-ignore
       this.webex.internal.newMetrics.submitClientEvent({
         name: 'client.media-engine.remote-sdp-received',
@@ -7237,10 +7246,10 @@ export default class Meeting extends StatelessWebexPlugin {
             }),
           ],
         },
-        options: {meetingId: this.id, rawError: new Error('Timeout waiting for SDP answer')},
+        options: {meetingId: this.id, rawError: error},
       });
 
-      deferSDPAnswer.reject(new Error('Timed out waiting for REMOTE SDP ANSWER'));
+      deferSDPAnswer.reject(error);
     }, ROAP_OFFER_ANSWER_EXCHANGE_TIMEOUT);
 
     LoggerProxy.logger.info(`${LOG_HEADER} waiting for REMOTE SDP ANSWER...`);
@@ -7345,7 +7354,7 @@ export default class Meeting extends StatelessWebexPlugin {
         error
       );
 
-      throw new AddMediaFailed();
+      throw new AddMediaFailed(error);
     }
   }
 
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -7769,14 +7769,10 @@ export default class Meeting extends StatelessWebexPlugin {
 
       const {connectionType, selectedCandidatePairChanges, numTransports} =
         await this.mediaProperties.getCurrentConnectionInfo();
-      // @ts-ignore
-      const reachabilityStats = await this.webex.meetings.reachability.getReachabilityMetrics();
+
       const iceCandidateErrors = Object.fromEntries(this.iceCandidateErrors);
 
-      // @ts-ignore
-      const isSubnetReachable = await this.webex.meetings.reachability.isSubnetReachable(
-        this.mediaServerIp
-      );
+      const reachabilityMetrics = await this.getMediaReachabilityMetricFields();
 
       Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.ADD_MEDIA_SUCCESS, {
         correlation_id: this.correlationId,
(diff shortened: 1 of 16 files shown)


packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -141,17 +141,56 @@ export default class Reachability extends EventsScope {
...
    */
-  public isSubnetReachable(mediaServerIp?: string): boolean | null {
+  public async isSubnetReachable(mediaServerIp?: string): Promise<boolean | null> {
     if (!mediaServerIp) {
       LoggerProxy.logger.error(`Reachability:index#isSubnetReachable --> mediaServerIp is null`);
 
       return null;
     }
 
+    try {
```
marcin-bazyl: wouldn't it be better to just call getReachabilityMetrics() and use the output of that? actually getReachabilityMetrics() is already being called where we now call isSubnetReachable() so we could just have the logic of combining these 2 into one behavioral metric there

k-wasniowski: In general I would prefer to keep it separate. Both functions have it's own narrowed purpose.
Combining them would break the rule to keep a single functionality for each function

marcin-bazyl: I'm not talking about combining them, just using them both inside Meeting.addMediaInternal()

marcin-bazyl: what I mean is that isSubnetReachable() currently as it stands works correctly, returns true or false if subnet is reachable. The logic that we want to send null/undefined metric value for the behavioral addMedia metric when no clusters are reachable at all doesn't strictly belong inside isSubnetReachable(), it belongs more in the code that sends the behavioral metric and that code already calls getReachabilityMetrics() and isSubnetReachable() so it has all the information to decide what metric value to send.

k-wasniowski: Do you mean something like this?
```
      const reachabilityMetrics = await this.webex.meetings.reachability.getReachabilityMetrics();
      const reachedSubnetsCount = ['udp', 'tcp', 'xtls'].reduce((acc, type) => {
        const propertyName = `reachability_public_${type}_success`;
        
        return acc + (reachabilityMetrics[propertyName] || 0);
      }, 0);
      let isSubnetReachable = null;
      if (reachedSubnetsCount > 0) {
        // @ts-ignore
        isSubnetReachable = await this.webex.meetings.reachability.isSubnetReachable(
          this.mediaServerIp
        )
      }
```

marcin-bazyl: yes and this block of code could be put into a new helper method called Meeting.getAddMediaMetricFields() or something like that, because this code is needed in 2 places and getAddMediaMetricFields() could also return the other fields like iceCandidateErrors, selectedCandidatePairChanges, numTransports, connectionType (this one only in success case)

k-wasniowski: Got it. Thank you :)

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -9622,4 +9614,37 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  private async getMediaReachabilityMetricFields(): Promise<MediaReachabilityMetrics> {
+    // @ts-ignore
+    const reachabilityMetrics = await this.webex.meetings.reachability.getReachabilityMetrics();
+    const totalSuccessCases = ['public', 'vmn'].reduce((total, type) => {
+      return (
+        total +
+        ['udp', 'tcp', 'xtls'].reduce((protocolTotal, protocol) => {
+          const propertyName = `reachability_${type}_${protocol}_success`;
+
+          return protocolTotal + (reachabilityMetrics[propertyName] || 0); // Safely access the property
```
marcin-bazyl: just a non-blocking thought: there is a small risk with this approach that if in the future the `ReachabilityMetrics` type changes and this code doesn't get updated, build won't fail but code won't do want we want, so maybe it would be safer to just do a hardcoded addition of the fields we care about instead of dynamically creating them or maybe there is some typescript magic that could be done to add more type safety here.

For example this adds type safety, but it's not ideal that I had to write the protocol names twice:
```
const totalSuccessCases = ['public', 'vmn'].reduce((total, type: 'public' | 'vmn') => {
      return (
        total +
        ['udp', 'tcp', 'xtls'].reduce((protocolTotal, protocol: 'udp' | 'tcp' | 'xtls') => {
          const propertyName: keyof ReachabilityMetrics = `reachability_${type}_${protocol}_success`;

          return protocolTotal + reachabilityMetrics[propertyName];
        }, 0)
      );
    }, 0);
```

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
k-wasniowski: Copilot suggested following implementation:

private async getMediaReachabilityMetricFields(): Promise<MediaReachabilityMetrics> {
  // @ts-ignore
  const reachabilityMetrics: ReachabilityMetrics = await this.webex.meetings.reachability.getReachabilityMetrics();

  // Define the exact keys for success metrics
  const successKeys: Array<keyof ReachabilityMetrics> = [
    'reachability_public_udp_success',
    'reachability_public_tcp_success',
    'reachability_public_xtls_success',
    'reachability_vmn_udp_success',
    'reachability_vmn_tcp_success',
    'reachability_vmn_xtls_success',
  ];

  // Calculate total success cases
  const totalSuccessCases = successKeys.reduce((total, key) => {
    return total + (reachabilityMetrics[key] || 0); // Safely access the property
  }, 0);

  let isSubnetReachable = null;
  if (totalSuccessCases > 0) {
    // @ts-ignore
    isSubnetReachable = await this.webex.meetings.reachability.isSubnetReachable(
      this.mediaServerIp
    );
  }

  return {
    ...reachabilityMetrics,
    isSubnetReachable,
  };
}

To ensure type safety

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -9622,4 +9614,37 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        }, 0)
+      );
+    }, 0);
+
+    let isSubnetReachable = null;
+    if (totalSuccessCases > 0) {
+      // @ts-ignore
+      isSubnetReachable = await this.webex.meetings.reachability.isSubnetReachable(
```
marcin-bazyl: nitpick: await is not needed here

k-wasniowski: Removed

szotrabh (approved):

PR #4267, opened 2025-05-13
Paired device metrics

jjkusowski:
DRAFT:  I will write/modify tests after review and before removing draft status.  I will also remove the commented properties that I have not populated.

# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-608727

Call Analytics team has requested that Paired Device info be sent with events when the web client is paired.

1) The device manager already stores the paired device.  This PR adds a function to return it.
2) While preparing Event items for CA events, use this new function to determine if the web client is paired.  If so, use the information in the device object to add the information to the Event.

Example corellationId: e554d08f-3e3f-48ea-8e24-506367840191

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manually tested in paired and non-paired state with a DX80.

- [X] GAI was not used (or, no additional notation is required)
  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -240,6 +240,24 @@ export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
...
   const upgradeChannel = buildType === 'prod' ? 'gold' : buildType;
 
+  const pairedDevice = await webex.devicemanager.getPairedDevice();
+
+  if (pairedDevice.id) {
+    const devicePayload = {
+      deviceId: pairedDevice.deviceInfo.id,
+      // deviceJoinType:
+      devicePairingType: 'Manual',
+      // deviceType?: string # such as Cloud, Cloud Aware, On-Prem
```
jjkusowski: I'll remove these, but I left in so reviewers could see all the optional properties.  I don't know how to determine the values for these, so I have left them out.

robstax: is it possible to be paired to multiple devices? does this always have the right one?

jjkusowski: I think the sdk can only keep one device in device-manager.js at this._pairedDevice.  That's the one I am grabbing.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -239,6 +239,20 @@ export const prepareDiagnosticMetricItem = (webex: any, item: any) => {
 
   // Set upgradeChannel to 'gold' if buildType is 'prod', otherwise to the buildType value
   const upgradeChannel = buildType === 'prod' ? 'gold' : buildType;
+  if (webex.devicemanager) {
+    const pairedDevice = webex.devicemanager.getPairedDevice();
+    if (pairedDevice) {
+      const devicePayload = {
+        deviceId: pairedDevice.deviceInfo?.id,
+        devicePairingType: 'Manual',
```
marcin-bazyl: what are other possible pairing types? is this for example about manual vs ultrasound pairing? I think we have an upcoming feature to support ultrasound pairing in Q1, so it would be better if we could read this info from somewhere rather hardcoding it, so that we don't have to update this code again when ultrasound pairing is added

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
marcin-bazyl: did you check with Meera if it's OK if those properties are left out?

jjkusowski: I did not, but they are marked as optional in the docs that are linked in the ticket.  It looks similar to what desktop sends.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
jjkusowski: Sorry for delay on responding.  I don't think this is the PR to do this in.  The larger effort of supporting ultrasonic pairing should handle that.

jor-row: I think what marcin is saying is we should have ```pairedMethod``` defined in the pairedDevice object (or in webex.devicemanager if pairing by one device only is supported) and get that info from there for this line

Then when ultrasound pairing is added the ```pairedMethod``` will be updated in the devicemanager package (where most of the work for that feature will be done i imagine?) and this event will just "work"

for now ```pairedMethod``` might be just hardcoded to 'Manual' but that should be done in devicemanager i think

packages/@webex/plugin-device-manager/src/device-manager.js
```diff
@@ -638,6 +638,22 @@ const DeviceManager = WebexPlugin.extend({
...
+   * @returns {device}
+   */
+  getPairedDevice() {
+    if (!this._pairedDevice) {
+      this.logger.info('DeviceManager#getPairedDeviceInfo: Currently no device is paired');
+
+      return undefined;
+    }
+    const pairedDeviceId =
+      this._pairedDevice && (this._pairedDevice.id || this._pairedDevice.identity.id);
```
jor-row: nitpick: we already check if this._pairedDevice is truthy on line 647 so this isn't needed

jor-row: will ```id``` or ```identity.id``` always be defined?

Should we us optional chaining just incase and do another check before calling DeviceCollection.get incase   pairedDeviceId is undefined?

packages/@webex/plugin-device-manager/test/unit/spec/index.js
```diff
@@ -3,9 +3,84 @@
...
+      const result = deviceManager.getPairedDevice();
+      
+      assert.isUndefined(result);
+      assert.isTrue(deviceManager.logger.info.calledOnce);
+      assert.isTrue(deviceManager.logger.info.calledWith(
+        'DeviceManager#getPairedDeviceInfo: Currently no device is paired'
+      ));
+    });
+
+    it('returns the paired device when a device is paired with id property', () => {
```
jor-row, line 73: a test here for if id and identity is undefined in the mockDevice, and also one where id in undefined, identity is defined but without an id inside

packages/@webex/plugin-device-manager/src/device-manager.js
```diff
@@ -638,6 +638,22 @@ const DeviceManager = WebexPlugin.extend({
...
     return Promise.resolve();
   },
+
+  /**
+   * gets the currently paired device info if paired
+   * @returns {device}
+   */
+  getPairedDevice() {
+    if (!this._pairedDevice) {
+      this.logger.info('DeviceManager#getPairedDeviceInfo: Currently no device is paired');
```
jor-row: nitpick: the function is called ```getPairedDevice``` so the "Info" isn't needed

jor-row (changes requested):

jor-row (approved):

arun3528 (approved):
looks good , would wait for other review comments from rob and others if any

PR #4269, opened 2025-05-14
feat(plugin-meetings): allow stopping ice gathering after first relay candidate

k-wasniowski:
# COMPLETES #[SPARK-626519](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-626519)

This PR introduces possibility to stop ICE candidates gathering.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested with JS-SDK

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/media/index.ts
+++ b/packages/@webex/plugin-meetings/src/media/index.ts
@@ -175,9 +175,6 @@ Media.createMediaConnection = (
   if (isMultistream) {
     const config: MultistreamConnectionConfig = {
       iceServers,
-      doFullIce: BrowserInfo.isFirefox(),
-      stopIceGatheringAfterFirstRelayCandidate:
-        BrowserInfo.isFirefox() && stopIceGatheringAfterFirstRelayCandidate,
     };
 
     if (bundlePolicy) {
@@ -186,6 +183,12 @@ Media.createMediaConnection = (
 
     if (disableAudioMainDtx !== undefined) {
       config.disableAudioMainDtx = disableAudioMainDtx;
+    }
+
+    if (BrowserInfo.isFirefox()) {
+      config.doFullIce = true;
+
+      config.stopIceGatheringAfterFirstRelayCandidate = stopIceGatheringAfterFirstRelayCandidate;
     }
 
     return new MultistreamRoapMediaConnection(
--- a/packages/@webex/plugin-meetings/src/media/index.ts
+++ b/packages/@webex/plugin-meetings/src/media/index.ts
@@ -16,6 +16,7 @@ import {
   LocalMicrophoneStream,
 } from '@webex/media-helpers';
 import {RtcMetrics} from '@webex/internal-plugin-metrics';
+import {BrowserInfo} from '@webex/web-capabilities';
 import LoggerProxy from '../common/logs/logger-proxy';
 import {MEDIA_TRACK_CONSTRAINT} from '../constants';
 import Config from '../config';
@@ -143,6 +144,7 @@ Media.createMediaConnection = (
     bundlePolicy?: BundlePolicy;
     iceCandidatesTimeout?: number;
     disableAudioMainDtx?: boolean;
+    stopIceGatheringAfterFirstRelayCandidate?: boolean;
   }
 ) => {
   const {
@@ -155,6 +157,7 @@ Media.createMediaConnection = (
     bundlePolicy,
     iceCandidatesTimeout,
     disableAudioMainDtx,
+    stopIceGatheringAfterFirstRelayCandidate,
   } = options;
 
   const iceServers = [];
@@ -172,6 +175,9 @@ Media.createMediaConnection = (
   if (isMultistream) {
     const config: MultistreamConnectionConfig = {
       iceServers,
+      doFullIce: BrowserInfo.isFirefox(),
+      stopIceGatheringAfterFirstRelayCandidate:
+        BrowserInfo.isFirefox() && stopIceGatheringAfterFirstRelayCandidate,
     };
 
     if (bundlePolicy) {
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -807,6 +807,18 @@ export default class Meetings extends WebexPlugin {
     }
   }
 
+  private _toggleStopIceGatheringAfterFirstRelayCandidate(newValue: boolean) {
+    if (typeof newValue !== 'boolean') {
+      return;
+    }
+
+    // @ts-ignore
+    if (this.config.stopIceGatheringAfterFirstRelayCandidate !== newValue) {
+      // @ts-ignore
+      this.config.stopIceGatheringAfterFirstRelayCandidate = newValue;
+    }
+  }
+
   /**
    * API to toggle usage of audio main DTX, needs to be called before webex.meetings.register()
    *
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -807,18 +807,6 @@ export default class Meetings extends WebexPlugin {
     }
   }
 
-  private _toggleStopIceGatheringAfterFirstRelayCandidate(newValue: boolean) {
-    if (typeof newValue !== 'boolean') {
-      return;
-    }
-
-    // @ts-ignore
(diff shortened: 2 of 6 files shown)


packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -413,6 +413,19 @@ describe('plugin-meetings', () => {
       });
     });
 
+    describe('#_toggleStopIceGatheringAfterFirstRelayCandidate', () => {
+      it('should have _toggleStopIceGatheringAfterFirstRelayCandidate', () => {
+        assert.equal(typeof webex.meetings._toggleStopIceGatheringAfterFirstRelayCandidate, 'function');
+      });
+
+      describe('success', () => {
+        it('should update meetings to disable audio main dtx', () => {
```
marcin-bazyl: test title is incorrect

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -413,6 +413,19 @@ describe('plugin-meetings', () => {
...
+          webex.meetings._toggleStopIceGatheringAfterFirstRelayCandidate(true);
```
marcin-bazyl, line 423: the default in packages/@webex/plugin-meetings/src/config.ts is true, so setting it to true in this test doesn't really test it all that well...... and I'm thinking that maybe better to keep the test as is, but to change the default to false, so that this change doesn't impact other SDK clients and we will gradually roll it out with a toggle in the web app

k-wasniowski: Changed the default to false

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
k-wasniowski: Fixed

marcin-bazyl (approved):

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
```diff
@@ -197,7 +198,68 @@ describe('createMediaConnection', () => {
 
     sendMetricsInQueueCallback();
     assert.calledOnce(rtcMetrics.sendMetricsInQueue);
+  });
+
+  it('multistream non-firefox doesnt care about stopIceGatheringAfterFirstRelayCandidate', () => {
```
edvujic:
```suggestion
  it('multistream non-firefox does not care about stopIceGatheringAfterFirstRelayCandidate', () => {
```

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
```diff
@@ -197,7 +198,68 @@ describe('createMediaConnection', () => {
...
+    assert.calledWith(
+      multistreamRoapMediaConnectionConstructorStub,
+      {
+        iceServers: []
+      },
+      'meeting id'
+    );
+  });
+
+  it('multistream firefox stop gathering after first relay if stopIceGatheringAfterFirstRelayCandidate is true', () => {
```
edvujic:
```suggestion
  it('multistream firefox stops gathering after first relay if stopIceGatheringAfterFirstRelayCandidate is true', () => {
```

edvujic (approved):
Couple of nits, but not blocking.

szotrabh (approved):

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -413,6 +413,19 @@ describe('plugin-meetings', () => {
       });
     });
 
+    describe('#_toggleStopIceGatheringAfterFirstRelayCandidate', () => {
+      it('should have _toggleStopIceGatheringAfterFirstRelayCandidate', () => {
+        assert.equal(typeof webex.meetings._toggleStopIceGatheringAfterFirstRelayCandidate, 'function');
+      });
+
+      describe('success', () => {
+        it('should update meetings to stop ICE candidates gathering after first relay candidate', () => {
```
antsukanova, line 422: I feel here we should test `true/false` as one test case and another small test case about `if (typeof newValue !== 'boolean') {` logic.

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
```diff
@@ -197,7 +198,68 @@ describe('createMediaConnection', () => {
 
     sendMetricsInQueueCallback();
     assert.calledOnce(rtcMetrics.sendMetricsInQueue);
+  });
+
+  it('multistream non-firefox does not care about stopIceGatheringAfterFirstRelayCandidate', () => {
```
antsukanova, line 203: maybe just `it('multistream non-firefox ignores stopIceGatheringAfterFirstRelayCandidate', () => {`?

packages/@webex/plugin-meetings/src/media/index.ts
```diff
@@ -182,6 +185,12 @@ Media.createMediaConnection = (
       config.disableAudioMainDtx = disableAudioMainDtx;
     }
 
+    if (BrowserInfo.isFirefox()) {
+      config.doFullIce = true;
+
+      config.stopIceGatheringAfterFirstRelayCandidate = stopIceGatheringAfterFirstRelayCandidate;
+    }
```
brycetham, line 192: A few questions:

1. What happens if I configure `stopIceGatheringAfterFirstRelayCandidate = true` and the browser isn't Firefox? Your unit tests imply that the option would be ignored, but there's no way an SDK user would know this. If we want to only apply this option to Firefox, it should be clear to the SDK user, e.g. by including "Firefox" in the option name.

2. What happens if `doFullIce = false` but `stopIceGatheringAfterFirstRelayCandidate = true`? I believe WCME would ignore the second option, but again this is not made clear to the SDK user. Perhaps we should throw an error/warning either at the SDK level or the WCME level?

3. Do we really want to condition for Firefox here? Should we let the web client be in full control of deciding when to enable/disable these options?

marcin-bazyl: 1. I think it would be enough to just explain in the config comments that the option currently applies only to FF. Leaving out the Firefox browser name from the config option allows us more flexibility in case we want to use this for other browsers in the future too (unlikely, but you never know...)
2. Such combination currently doesn't make sense, because we don't wait for gathering of candidates when doing ice-lite, but in theory we could do this in the future, so for now I don't mind we could throw an error that such combination is not supported, but I'm fine also with just silently ignoring `stopIceGatheringAfterFirstRelayCandidate` in that case.
3. No, we definitely don't want the app to be controlling which browsers do full ICE, because our backend really requires FF to have full ICE and not sure if it would work with other browsers. Other clients of the JS-SDK won't know which browsers should have full ICE and which shouldn't, so the logic of enabling full ice for FF and only for FF needs to be somewwhere in JS-SDK or layers below.

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -827,6 +827,27 @@ export default class Meetings extends WebexPlugin {
...
+   * API to toggle stopping ICE Candidates Gathering after first relay candidate,
+   * needs to be called before webex.meetings.joinWithMedia()
+   *
+   * @param {Boolean} newValue
+   * @private
+   * @memberof Meetings
+   * @returns {undefined}
+   */
+  private _toggleStopIceGatheringAfterFirstRelayCandidate(newValue: boolean) {
+    if (typeof newValue !== 'boolean') {
```
SomeBody16, line 840: will this method be ever called with non-boolean?
If it can, maybe we could change definition to `(newValue?: boolean)`

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -827,6 +827,27 @@ export default class Meetings extends WebexPlugin {
...
+      return;
+    }
+
+    // @ts-ignore
```
SomeBody16, line 844:
```suggestion
    // @ts-expect-error [explanation]
```

expect-error will also show error if for some reason in future below line will start to be correct and potentially stop silencing important errors

PR #4272, opened 2025-05-15
fix: selfutils - small refactor

marcin-bazyl:
Fixed it so that vscode can now jump to function definitions.

Make sure to tick the "hide whitespace" box when looking at the diff


chrisadubois (approved):
this is a great improvement

k-wasniowski (approved):

adhmenon (approved):

PR #4274, opened 2025-05-15
fix(calling): fixed handling for http response 429 and 400 requests

Kesari3008:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-651487 

We have added triggering fallback retry in case of 429 response only after the number of secons received in Retry-After header
We have also added handling for 400 Bad Request Error cases.
Updated method names in logger as existing approach was not working

Added timeout for invoking scheduleFallbackRetryfor429 after the retry-after interval only. 

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added UT for failover case receiving 429 in different attempts and handling 429 response.
[test-results1.log](https://github.com/user-attachments/files/20553509/test-results1.log)
[test-results2.log](https://github.com/user-attachments/files/20553510/test-results2.log)

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [x] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/common/Utils.ts
```diff
@@ -322,6 +338,30 @@ export async function handleRegistrationErrors(
...
+
+    case ERROR_CODE.TOO_MANY_REQUESTS: {
+      const caller = loggerContext.method || 'handleErrors';
+
+      if (err.headers) {
+        const retryAfter = Number(err.headers['retry-after']);
+        retry429CB(retryAfter, caller);
+      }
+    }
+
```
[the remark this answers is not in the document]
Kesari3008: Fixed

packages/calling/src/common/Utils.ts
```diff
@@ -297,6 +297,7 @@ export async function handleRegistrationErrors(
   err: WebexRequestPayload,
   emitterCb: LineErrorEmitterCallback,
   loggerContext: LogContext,
+  retry429CB?: retry429CallBack,
```
sreenara:
```suggestion
  retry429Cb?: retry429CallBack,
```

packages/calling/src/common/Utils.ts
```diff
@@ -305,6 +306,21 @@ export async function handleRegistrationErrors(
   let finalError = false;
   log.warn(`Status code: -> ${errorCode}`, loggerContext);
   switch (errorCode) {
+    case ERROR_CODE.BAD_REQUEST: {
```
sreenara, line 309: Were we able to find these details?
1. When do we hit the 400 error?
2. Why is it not always reproducible?
3. What parameters are missing or malformed causing 400?
The reason I ask is that it's surprising that this only happens sometimes.

packages/calling/src/common/Utils.ts
```diff
@@ -305,6 +306,21 @@ export async function handleRegistrationErrors(
...
+      finalError = true;
+      log.warn(`400 Bad Request`, loggerContext);
+
+      updateLineErrorContext(
+        loggerContext,
+        ERROR_TYPE.SERVER_ERROR,
+        'Invalid input. Please verify the required parameters,  sign out and then sign back in with the valid data',
```
sreenara:
```suggestion
        'Invalid input. Please verify the required parameters, sign out and then sign back in with the valid data',
```

packages/calling/src/common/Utils.ts
```diff
@@ -322,6 +338,32 @@ export async function handleRegistrationErrors(
       break;
     }
 
+    case ERROR_CODE.DEVICE_NOT_FOUND: {
```
sreenara, line 341: Please add to the description that this PR also takes care of 400, 404 errors and their appropriate handling. The description is only talking about 429

packages/calling/src/common/Utils.ts
```diff
@@ -322,6 +338,32 @@ export async function handleRegistrationErrors(
...
+      finalError = true;
```
sreenara, line 342: What should happen in this case?
1. Device is registered and sending keepalives
2. Mobius responds to one keepalive with 404

Will our logic of 5 keepalives retries still continue?
What should happen with the device if a 404 is received?

packages/calling/src/common/Utils.ts
```diff
@@ -322,6 +338,32 @@ export async function handleRegistrationErrors(
...
+      log.warn(`404 Device Not Found`, loggerContext);
+
+      updateLineErrorContext(
+        loggerContext,
+        ERROR_TYPE.NOT_FOUND,
+        'The client has unregistered. Please wait for the client to register before attempting the call. If error persists, sign out, sign back in and attempt the call.',
```
sreenara: If this is a final error, the client will never register right?

packages/calling/src/common/Utils.ts
```diff
@@ -322,6 +338,32 @@ export async function handleRegistrationErrors(
...
+      );
+      emitterCb(lineError, finalError);
+      break;
+    }
+
+    case ERROR_CODE.TOO_MANY_REQUESTS: {
+      const caller = loggerContext.method || 'handleErrors';
+
+      if (retry429CB && err.headers) {
+        const retryAfter = Number(err.headers['retry-after']);
```
sreenara, line 361: Is there a possibility of 429 arriving without retry-after?

packages/calling/src/common/Utils.test.ts
```diff
@@ -305,7 +319,7 @@ describe('Registration Tests', () => {
       RegistrationStatus.ACTIVE
     );
 
-    handleRegistrationErrors(webexPayload, mockEmitterCb, logObj, mockRestoreCb);
+    handleRegistrationErrors(webexPayload, mockEmitterCb, logObj, mock429RetryCb, mockRestoreCb);
```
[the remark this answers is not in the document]
sreenara: +1. There is no test covering the 429 case.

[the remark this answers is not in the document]
sreenara: This mock429RetryCb should only be sent in for the 429 retry right? Why are we passing it for all tests?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -208,28 +211,30 @@ export class Registration implements IRegistration {
...
-      file: REGISTRATION_FILE,
-      method: FAILBACK_429_RETRY_UTIL,
-    });
-    const interval = this.getRegRetryInterval(this.failback429RetryAttempts);
+  private async handle429Retry(retryAfter: number, caller: string): Promise<void> {
+    if (caller === FAILBACK_UTIL) {
+      if (this.failback429RetryAttempts >= REG_FAILBACK_429_MAX_RETRIES) {
+        return;
+      }
+      this.scheduled429Retry = true;
```
sreenara: shouldn't this only be set after starting the failover timer, i.e. line 230?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -277,13 +282,16 @@ export class Registration implements IRegistration {
     }
 
     let abort;
-
     if (interval > BASE_REG_RETRY_TIMER_VAL_IN_SEC && !this.failoverImmediately) {
       const scheduledTime = Math.floor(Date.now() / 1000);
 
+      if (this.retryAfter != null) {
```
sreenara, line 294: We are initializing `this.retryAfter` as `undefined` at the start. So shouldn't this check also follow the same logic? Does null equal undefined?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -277,13 +282,16 @@ export class Registration implements IRegistration {
...
+        interval = interval < this.retryAfter ? this.retryAfter : interval;
+      }
+
       setTimeout(async () => {
         await this.mutex.runExclusive(async () => {
-          abort = await this.attemptRegistrationWithServers(this.startFailoverTimer.name);
+          abort = await this.attemptRegistrationWithServers(FAILOVER_UTIL);
```
sreenara, line 300: We should have a comment in the description covering why this change was made.

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -676,15 +687,6 @@ export class Registration implements IRegistration {
         if (abort) {
           this.setStatus(RegistrationStatus.INACTIVE);
           break;
-        } else if (caller === this.executeFailback.name) {
```
sreenara, line 679: We should have a comment in the PR description about why this change was done

packages/calling/src/common/Utils.ts
```diff
@@ -425,20 +467,6 @@ export async function handleRegistrationErrors(
       }
       break;
     }
-    case ERROR_CODE.DEVICE_NOT_FOUND: {
```
sreenara, line 428: Why was this code moved from here to above?

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -81,31 +82,44 @@ describe('Registration Tests', () => {
...
+  });
   const successPayload = <WebexRequestPayload>(<unknown>{
     statusCode: 200,
     body: mockPostResponse,
   });
 
   let reg: IRegistration;
   let restartSpy;
-  let failbackRetry429Spy;
+
```
sreenara: can we remove this empty line?

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -204,6 +218,130 @@ describe('Registration Tests', () => {
...
+      expect(webex.request).toBeCalledOnceWith({
+        ...mockResponse,
+        method: 'POST',
+        uri: `${mobiusUris.primary[0]}device`,
+      });
+
+      expect(retry429Spy).not.toBeCalled();
+      expect(reg.retryAfter).toEqual(undefined);
+      expect(failoverSpy).toBeCalledOnceWith(3, 85);
```
sreenara, line 326: Question.. why is this `85` and not `75`? I thought this would be the sum of the 2 retry afters?

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -204,6 +218,130 @@ describe('Registration Tests', () => {
...
+    });
+
+    it('handle 429 received while failing over to backup server', async () => {
```
sreenara: Rather than using `REG_TRY_BACKUP_TIMER_VAL_IN_SEC`, which is 1200, can we use the CC time? All flows must be compatible with CC more than calling due to the usage.

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -204,6 +218,130 @@ describe('Registration Tests', () => {
...
+        .mockRejectedValueOnce(failurePayload)
+        .mockRejectedValueOnce(failurePayload)
+        .mockRejectedValueOnce(failurePayload429One)
+        .mockResolvedValueOnce(successPayload);
+
+      await reg.triggerRegistration();
+
+      jest.advanceTimersByTime(REG_TRY_BACKUP_TIMER_VAL_IN_SEC * SEC_TO_MSEC_MFACTOR);
+      await flushPromises();
+      expect(webex.request).toBeCalledTimes(4);
```
sreenara: With CC flows, this would be a good test to see if the SDK moved to backup within the time of 114 seconds.

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -204,6 +218,130 @@ describe('Registration Tests', () => {
...
+        ...mockResponse,
+        method: 'POST',
+        uri: `${mobiusUris.primary[0]}device`,
+      });
+
+      expect(webex.request).toBeCalledWith({
+        ...mockResponse,
+        method: 'POST',
+        uri: `${mobiusUris.backup[0]}device`,
+      });
```
sreenara: This only covers once each right? How do we get webex.request = 4?

packages/calling/src/CallingClient/line/line.test.ts
```diff
@@ -201,8 +201,9 @@ describe('Line Tests', () => {
         expect.anything(),
         {
           file: REGISTRATION_FILE,
-          method: 'attemptRegistrationWithServers',
+          method: 'triggerRegistration',
         },
+        expect.anything(),
         expect.anything()
```
sreenara, line 207: When did we add expect.anything().. is that safe?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -401,20 +408,22 @@ export class Registration implements IRegistration {
...
           const abort = await this.attemptRegistrationWithServers(FAILBACK_UTIL);
-          if (!abort && !this.isDeviceRegistered()) {
-            const abortNew = await this.restorePreviousRegistration(FAILBACK_UTIL);
+          if (!this.scheduled429Retry) {
+            if (!abort && !this.isDeviceRegistered()) {
+              const abortNew = await this.restorePreviousRegistration(FAILBACK_UTIL);
 
-            if (abortNew) {
-              this.clearFailbackTimer();
+              if (abortNew) {
```
sreenara: This is 5 levels of if-condition.. is there no static code smell raised for this.. Any way to avoid so many levels?

sreenara (changes requested):
Please see the comments

packages/calling/src/common/Utils.ts
Kesari3008: Addressed

packages/calling/src/common/Utils.ts
Kesari3008: When do we hit the 400 error? 
When the payload in the register request has userId missing.

Why is it not always reproducible?
It is reproducible if we don;t run the samples:build or use expiry token probably. WDM registration fails and we can see the device registration failing with 400

What parameters are missing or malformed causing 400?
UserId

packages/calling/src/common/Utils.ts
Kesari3008: Addressed

packages/calling/src/common/Utils.ts
Kesari3008: Updated

packages/calling/src/common/Utils.ts
Kesari3008: Yes this is a final error: 404 Not found

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Failback handling is different so this is to control coninuing the logic in execute failback as we are habdling the retyry here

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: If we use != it works for null and undefined both

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Updated

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Updated

packages/calling/src/common/Utils.ts
Kesari3008: Just keeping all 4xx together

packages/calling/src/CallingClient/registration/register.test.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/registration/register.test.ts
Kesari3008: In the second attempt, we are using interval calculated instead of retryAfter since retry-after is less than the interval value

packages/calling/src/CallingClient/line/line.test.ts
Kesari3008: This is how the existing test was so I just added on top of it. I can try changing it

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: I can reduce the levels of nesting here by updating it to below:

// Best - eliminates nested conditions entirely
if (this.scheduled429Retry || abort || this.isDeviceRegistered()) {
  return;
}

const abortNew = await this.restorePreviousRegistration(FAILBACK_UTIL);

if (abortNew) {
  this.clearFailbackTimer();
  return;
}

if (!this.isDeviceRegistered()) {
  await this.restartRegistration(this.executeFailback.name);
} else {
  this.failbackTimer = undefined;
  this.initiateFailback();
}

Does this look ok ?

packages/calling/src/common/Utils.ts
Kesari3008: For final error, keepalive retries wont continue but failoverImmediately is set to true for CC flow and for non-cc flows usual failover logic would work

packages/calling/src/common/Utils.ts
Kesari3008: Not sure about this

packages/calling/src/common/Utils.ts
sreenara: @Kesari3008 so should we change the wording in the message? Right now it seems like the user can wait for it to register?

packages/calling/src/common/Utils.ts
sreenara: @Kesari3008 can we check with someone in the mobius team?

packages/calling/src/CallingClient/registration/register.ts
sreenara: Should this be `!==` instead of `!=`?

packages/calling/src/CallingClient/registration/register.ts
sreenara: I don't understand. My point here was that shouldn't we set this boolean to true only after we have scheduled the new retry for 429, which happens on line 231? Setting this boolean earlier without actually starting the timer doesn't seem right

packages/calling/src/CallingClient/registration/register.ts
sreenara: Yes this is OK

packages/calling/src/common/Utils.ts
Kesari3008: Message Krishna for this

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: No if we us !== then it's a stricter check for either undefiner or nul based on what we compare against. This way it take care of both

packages/calling/src/CallingClient/registration/register.test.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/registration/register.test.ts
Kesari3008: Current UTs take care of that

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: This is not for failover logic but failback logic. This is present only to control either restore logic is executed from the retryHandler instead of executeFailback

packages/calling/src/common/Utils.ts
Kesari3008: What would you like to change it to

Kesari3008: Addressed

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -204,6 +226,283 @@ describe('Registration Tests', () => {
...
 
+  describe('429 handling tests', () => {
+    const loggerContext = {
+      file: REGISTRATION_FILE,
+      method: FAILOVER_UTIL,
+    };
+    const logSpy = jest.spyOn(log, 'log');
+
+    beforeEach(() => {
+      mobiusUris.backup.pop();
```
sreenara, line 245: Backup url is being popped to control only 1 backup url for 429 tests.

sreenara (approved):

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -276,14 +283,21 @@ export class Registration implements IRegistration {
...
 
+    let abort;
     if (interval > BASE_REG_RETRY_TIMER_VAL_IN_SEC && !this.failoverImmediately) {
       const scheduledTime = Math.floor(Date.now() / 1000);
 
+      if (this.retryAfter != null) {
+        interval = Math.max(interval, this.retryAfter);
+      }
+
       setTimeout(async () => {
```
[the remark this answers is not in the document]
Kesari3008: Won;t the below code take care of that:

     if (this.retryAfter != null && interval < this.retryAfter) {
      this.failoverImmediately = this.retryAfter + timeElapsed > TIMER_THRESHOLD;
    }

    let abort;
    if (interval > BASE_REG_RETRY_TIMER_VAL_IN_SEC && !this.failoverImmediately) {

PR #4289, opened 2025-05-23
fix(plugin-meetings): fix brb state

antsukanova:
# COMPLETES # [SPARK-655626](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-655626)

When a user steps away while using multiple devices during a single call, the SDK's existing logic waits for the 'brb' update state from the Locus side, that won't happen because locus will set brb update only after all devices request brb enabling. This results in unnecessary request loops on the client now. This PR addresses the issue by updating the logic to avoid these request loops and improve efficiency.

updates BrbState implementation for sendLocalBrbStateToServer

Vidcast for fix: https://app.vidcast.io/share/19135395-6eb2-4788-9289-5218cfaf12e3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- unit tests
- web client manual and auto test cases

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/brbState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/brbState.ts
@@ -95,6 +95,9 @@ export class BrbState {
     return this.sendLocalBrbStateToServer(sendSlotManager)
       .then(() => {
         this.state.syncToServerInProgress = false;
+
+        this.handleServerBrbUpdate(this.state.client.enabled);
+
         LoggerProxy.logger.info(
           `Meeting:brbState#applyClientStateToServer: sync with server completed`
         );
--- a/packages/@webex/plugin-meetings/src/meeting/brbState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/brbState.ts
@@ -96,6 +96,9 @@ export class BrbState {
       .then(() => {
         this.state.syncToServerInProgress = false;
 
+        // This is a workaround for the fact that the server does not send the brb state
+        // in the locus update when a user joins with multiple devices.
+        // In the future, this could be improved with a new brb locus update handler
         this.handleServerBrbUpdate(this.state.client.enabled);
 
         LoggerProxy.logger.info(
--- a/packages/@webex/plugin-meetings/src/meeting/brbState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/brbState.ts
@@ -99,6 +99,7 @@ export class BrbState {
         // This is a workaround for the fact that the server does not send the brb state
         // in the locus update when a user joins from multiple devices but not all devices are requested brb.
         // In the future, this could be improved with a new brb locus update handler
+        // https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-655626
         this.handleServerBrbUpdate(this.state.client.enabled);
 
         LoggerProxy.logger.info(
--- a/packages/@webex/plugin-meetings/src/meeting/brbState.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/brbState.ts
@@ -97,7 +97,7 @@ export class BrbState {
         this.state.syncToServerInProgress = false;
 
         // This is a workaround for the fact that the server does not send the brb state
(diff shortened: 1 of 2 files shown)


k-wasniowski (approved):

edvujic (approved):

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -95,6 +95,9 @@ export class BrbState {
     return this.sendLocalBrbStateToServer(sendSlotManager)
       .then(() => {
         this.state.syncToServerInProgress = false;
+
+        this.handleServerBrbUpdate(this.state.client.enabled);
+
```
edvujic: Nice and clean. But, should we add a `TODO` here indicating this is not the optimal solution?

antsukanova: @edvujic thanks you. I would say it is not the super right solution because I feel it breaks some flows for https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610065, but I can't test this now due to https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-655628 issue.

I was considering putting a comment like `// verify this logic correctness flow after SPARK-655628`
But I can keep this on my list as well to check because I'm currently working on that already.

brycetham: We should add a comment here describing that this is a temporary solution, provide some potential long-term solution (they don't have to be finalized, so we can say something like "In the future, we could potentially solve this by..."), and link a JIRA. I think it is important to make it clear that this is just a quick fix and that we are thinking about long-term solutions.

antsukanova: @brycetham @edvujic thanks for your ideas and thoughts, I've added some explanations that should give visibility about this workaround. Please let me know if they are clear for you.

upd: attached task as well

brycetham (approved):

SomeBody16 (approved):

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -95,6 +95,13 @@ export class BrbState {
     return this.sendLocalBrbStateToServer(sendSlotManager)
       .then(() => {
         this.state.syncToServerInProgress = false;
+
+        // This is a workaround for the fact that the server does not send the brb state
+        // in the locus update when a user joins from multiple devices but not all devices are requested brb.
+        // In the future, this could be improved with a new brb locus update handler
+        // https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-655626
+        this.handleServerBrbUpdate(this.state.client.enabled);
```
marcin-bazyl, line 103: as discussed offline, it feels like we should really understand better what is the desired behavior in these cases, for example even if we force our brb status to "on" locally here, if Locus doesn't think it's on, then what will the other meeting participants see as our status? And probably worth discussing this with Colin to confirm what is the desired behaviour for VDI.

Also, it would be good to update some unit test to add the check that this call is being made.

marcin-bazyl (approved):
Approving, so that you're not blocked by this issue, but I feel like we still need better understating of the overall requirements how this feature should work in these edge cases.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
antsukanova: @marcin-bazyl  Thanks I agree with you, I will extend brbState.ts text cases, it feels good to have some extention for them as well. 
I have been already working on VDI flow. I need to check in more detailed why one participant browser+browser devices flow  and one participant browser+native devices flow work a bit differently. Second one currently work as expected and same as native+native (with this fix). And this fix allows unblock some other flows so I will keep it now but as I put comment it can be improved. 
Thank you for your involvement!

PR #4295, opened 2025-05-27
feat(plugin-cc): updated-team-change-for-device-change

adhmenon:
# COMPLETES #< ADHOC >

* In the agent device update method, earlier we had missed the team changing.

* Code has been added to ensure that team can also be changed by providing the team id.

Vidcast - https://app.vidcast.io/share/2e913584-658e-4a4b-8bf3-1f658b48d5d3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Manually tested webrtc voice call.
* Tetsed out extension voice call and device update with team update as well.

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/docs/samples/contact-center/app.js
+++ b/docs/samples/contact-center/app.js
@@ -35,8 +35,8 @@ const idleCodesDropdown = document.querySelector('#idleCodesDropdown')
 const setAgentStatusButton = document.querySelector('#setAgentStatus');
 const logoutAgentElm = document.querySelector('#logoutAgent');
 const buddyAgentsDropdownElm = document.getElementById('buddyAgentsDropdown');
-const updateAgentDeviceTypeElm = document.querySelector('#updateAgentDeviceType');
-const updateFieldsContainer = document.querySelector('#updateAgentDeviceTypeFields');
+const updateAgentProfileElm = document.querySelector('#updateAgentProfile');
+const updateFieldsContainer = document.querySelector('#updateAgentProfileFields');
 const updateLoginOptionElm = document.querySelector('#updateLoginOption');
 const updateDialNumberElm  = document.querySelector('#updateDialNumber');
 const updateTeamDropdownElm = document.querySelector('#updateTeamDropdown');
@@ -74,6 +74,7 @@ const engageElm = document.querySelector('#engageWidget');
 let isBundleLoaded = false; // this is just to check before loading/using engage widgets
 const uploadLogsButton = document.getElementById('upload-logs');
 const uploadLogsResultElm = document.getElementById('upload-logs-result');
+const applyupdateAgentProfileBtn = document.querySelector('#applyupdateAgentProfile');
 
 deregisterBtn.style.backgroundColor = 'red';
 
@@ -956,7 +957,7 @@ function register() {
       console.log('Agent re-login successful', data);
       loginAgentElm.disabled = true;
       logoutAgentElm.classList.remove('hidden');
-      updateAgentDeviceTypeElm.classList.remove('hidden');
+      updateAgentProfileElm.classList.remove('hidden');
 
       agentLogin.value = data.deviceType;
       agentDeviceType = data.deviceType;
@@ -975,7 +976,7 @@ function register() {
       console.log('Agent station-login success', data);
       loginAgentElm.disabled = true;
       logoutAgentElm.classList.remove('hidden');
-      updateAgentDeviceTypeElm.classList.remove('hidden');
+      updateAgentProfileElm.classList.remove('hidden');
       updateFieldsContainer.classList.add('hidden');
 
       agentLogin.value = data.deviceType;
@@ -1081,7 +1082,7 @@ function doAgentLogin() {
     console.log('Agent Logged in successfully', response);
     loginAgentElm.disabled = true;
     logoutAgentElm.classList.remove('hidden');
-    updateAgentDeviceTypeElm.classList.remove('hidden');
+    updateAgentProfileElm.classList.remove('hidden');
     // Read auxCode and lastStateChangeTimestamp from login response
     const DEFAULT_CODE = '0'; // Default code when no aux code is present
     const auxCodeId = response.data.auxCodeId?.trim() !== '' ? response.data.auxCodeId : DEFAULT_CODE;
@@ -1119,7 +1120,7 @@ function logoutAgent() {
     .then((response) => {
       console.log('Agent logged out successfully', response);
       loginAgentElm.disabled = false;
-      updateAgentDeviceTypeElm.classList.add('hidden');
+      updateAgentProfileElm.classList.add('hidden');
       updateFieldsContainer.classList.add('hidden');
 
      // Clear the timer when the agent logs out.
@@ -1144,7 +1145,7 @@ function logoutAgent() {
   });
 }
 
-async function applyupdateAgentDeviceType() {
+async function applyupdateAgentProfile() {
   const loginOption = updateLoginOptionElm.value;
   const newDial = loginOption === 'BROWSER' ? '' : updateDialNumberElm.value;
   const payload = {
@@ -1153,7 +1154,7 @@ async function applyupdateAgentDeviceType() {
     dialNumber: newDial,
   };
   try {
-    const resp = await webex.cc.updateAgentDeviceType(payload);
+    const resp = await webex.cc.updateAgentProfile(payload);
     console.log('Profile updated', resp);
     updateFieldsContainer.classList.add('hidden');
     // Reflect new values in main UI
@@ -1168,7 +1169,7 @@ async function applyupdateAgentDeviceType() {
   }
 }
 
-function showupdateAgentDeviceTypeUI() {
+function showupdateAgentProfileUI() {
   // ensure update dialog reflects current team
   if (updateTeamDropdownElm) {
     updateTeamDropdownElm.value = teamsDropdown.value;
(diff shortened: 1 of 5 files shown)


packages/@webex/plugin-cc/src/types.ts
```diff
@@ -247,7 +247,9 @@ export type AgentLogin = {
   loginOption: LoginOption;
 };
 
-export type AgentDeviceUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber'>;
+export type AgentDeviceUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber'> & {
+  teamId?: string;
```
adhmenon: Made this optional as users don't really need to enter team id, by default we take the current team.

docs/samples/contact-center/app.js
```diff
@@ -1132,6 +1133,7 @@ function logoutAgent() {
 
 async function updateAgentDeviceType() {
   const payload = {
+    teamId: teamsDropdown.value,
```
adhmenon, line 1136: Reason you see 2 payloads here is because we support updating from the `update` section and the main section as well.

docs/samples/contact-center/app.js
```diff
@@ -877,6 +878,19 @@ function register() {
             option.text = team.name;
             teamsDropdown.add(option);
         });
+        if (updateTeamDropdownElm) {
```
adhmenon, line 882: So thaat we can sync between team dropdowns

docs/samples/contact-center/app.js
```diff
@@ -1130,29 +1144,11 @@ function logoutAgent() {
   });
 }
 
-async function updateAgentDeviceType() {
```
adhmenon, line 1133: Dead code, no longer needed

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -919,8 +920,12 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
     });
 
     try {
-      // ensure we change device type
-      if (this.webCallingService?.loginOption === data.loginOption) {
+      const newTeamId = data.teamId ?? this.agentConfig.currentTeamId ?? EMPTY_STRING;
+      // Only block if both loginOption AND teamId remain unchanged
+      if (
```
adhmenon, line 924: As explaineed only block if both team and device type are the same.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -919,8 +920,12 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
     });
 
     try {
-      // ensure we change device type
-      if (this.webCallingService?.loginOption === data.loginOption) {
+      const newTeamId = data.teamId ?? this.agentConfig.currentTeamId ?? EMPTY_STRING;
```
Kesari3008: What happens when teamId is empty string?

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -247,7 +247,9 @@ export type AgentLogin = {
   loginOption: LoginOption;
 };
 
-export type AgentDeviceUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber'>;
+export type AgentDeviceUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber'> & {
```
Kesari3008: teamId would be present in agentLogin type also right so why are we not picking it from there? Also where are we assigning it a default value

Kesari3008 (approved):

packages/@webex/plugin-cc/src/cc.ts
adhmenon: So if `data.teamId` is empty it uses the agent config's team id - which can never be empty for a logged in user.

packages/@webex/plugin-cc/src/types.ts
adhmenon: Yes, but in device update it's optional, there is not default value - in the method, we simply check for the existing config value.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -893,14 +893,15 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    * Updates the agent device type.
    * This method allows the agent to change their device type (e.g., from BROWSER to EXTENSION or anything else).
    * It will also throw an error if the new device type is the same as the current one.
-   * @param data type is AgentDeviceUpdate - The data required to update the agent device type, including the new login option and dial number.
+   * @param data type is AgentDeviceUpdate - The data required to update the agent device type, including the new login option, new team and dial number.
```
mkesavan13: We can remove the "new"
```suggestion
   * @param data type is AgentDeviceUpdate - The data required to update the agent device type, including the new login option, team and dial number.
```

mkesavan13: Should we also consider changing the name of this method? `updateDeviceType` doesn't seem appropriate anymore

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -919,8 +920,12 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
     });
 
     try {
-      // ensure we change device type
-      if (this.webCallingService?.loginOption === data.loginOption) {
+      const newTeamId = data.teamId ?? this.agentConfig.currentTeamId ?? EMPTY_STRING;
+      // Only block if both loginOption AND teamId remain unchanged
```
mkesavan13, line 923: Should we also check if they are even logged in to call this method? Or else, this is going to fail as the station logout would fail

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -919,8 +920,12 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      if (
+        this.webCallingService?.loginOption === data.loginOption &&
```
Kesari3008, line 925: This should be an or condition. What if I only want to update the team and keep the same deviceType. In that case profile update is failing which shouldn't

packages/@webex/plugin-cc/src/types.ts
mkesavan13: I think we should stick to asking for a teamId all the time. Just to avoid confusions. The code to pick existing team can stay there but we need not mark it optional.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -919,8 +920,12 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        newTeamId === this.agentConfig.currentTeamId
+      ) {
         const message =
           'Will not proceed with device update as new Device type is same as current device type';
```
rsarika:
```suggestion
          'Will not proceed with device update as new Device type is same as current device type and teamId is same as current teamId';
```

rsarika (approved):

docs/samples/contact-center/index.html
```diff
@@ -133,6 +133,9 @@ <h2 class="collapsible">
                   <button id="updateAgentDeviceType" class="btn btn-primary my-3 ml-2 hidden" onclick="showupdateAgentDeviceTypeUI()">Update Profile</button>
 
                   <div id="updateAgentDeviceTypeFields" class="hidden" style="margin-top:1rem;">
+                    <select id="updateTeamDropdown" class="form-control w-auto my-2">
+                      <option value="" selected hidden>Choose Team …</option>
+                    </select>
```
Kesari3008, line 138: Save button should be disabled when necessary selections are not present. when I choose extension or dial number and textbox is empty, I am still able to click on save and trigger login request

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Profile update should not fail in this case - OR means the above case you mentioned would fail.
I had tested this out and it was working....

vidcast - https://app.vidcast.io/share/7841bb47-d4cf-45e5-9a27-d422a7f18c53

packages/@webex/plugin-cc/src/cc.ts
adhmenon: Fair - I think `AgentProfileUpdate` would work best in this case.

packages/@webex/plugin-cc/src/cc.ts
```diff
@@ -893,36 +893,40 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
       METRIC_EVENT_NAMES.AGENT_DEVICE_TYPE_UPDATE_FAILED,
     ]);
 
     const trackingId = `WX_CC_SDK_${uuidv4()}`;
 
-    LoggerProxy.info(`[${trackingId}] updateAgentDeviceType | starting profile update`, {
+    LoggerProxy.info(`[${trackingId}] updateAgentProfile | starting profile update`, {
       module: CC_FILE,
-      method: this.updateAgentDeviceType.name,
+      method: this.updateAgentProfile.name,
```
mkesavan13: Didn't we check that in minified versions this name wasn't going through and we had to mention strings here?

cc - @rsarika

packages/@webex/plugin-cc/src/cc.ts
mkesavan13: Have we addressed this? or is it that the loginOption would be empty if already not logged in and that is enough?

mkesavan13 (approved):
Looks good but a few minor comments to be addressed before merging. Approving

packages/@webex/plugin-cc/src/types.ts
```diff
@@ -247,7 +247,7 @@ export type AgentLogin = {
   loginOption: LoginOption;
 };
 
-export type AgentDeviceUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber'>;
+export type AgentProfileUpdate = Pick<AgentLogin, 'loginOption' | 'dialNumber' | 'teamId'>;
```
mkesavan13, line 250: Can't we use the AgentLogin itself now? Do we need a separate type?

adhmenon: Yes, we don't want to confuse end users.
Again, this is something we will document strongly. I've already added the typedoc so that should be good.

packages/@webex/plugin-cc/src/cc.ts
adhmenon: I think it's alright, we will document this.
There is no sure shot way to see if the agent is logged in or not - we have the agent profile flag but that only sets to true when relogin happens.
For now, we will document it strongly, but even if end users try, they will get an error, so it should be fine.

adhmenon:
Verified with amplify link. 
Things are working as expected.

PR #4300, opened 2025-05-28
fix(plugin-meetings): saving subnet ip details in reachability

venky-mediboina:
# COMPLETES #[SPARK-610518](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610518)

Saving the subnet ip details and sending to web-client

Saved the subnet ip details in reachability results and sending to web-client

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested manually whether we are getting the subnets in the response or not in webex-web-client. 

  - [X] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [X] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -49,7 +49,13 @@ export class ClusterReachability extends EventsScope {
   private srflxIceCandidates: RTCIceCandidate[] = [];
   public readonly isVideoMesh: boolean;
   public readonly name;
-  public readonly reachedSubnets: Set<string> = new Set();
+  public readonly reachedSubnets: Set<{
+    protocol: string;
+    serverIp: string;
+    reachable: string;
```
marcin-bazyl: why do we need `reachable` here? the name of this set is `reachedSubnets` so they're all always reachable.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -49,7 +49,13 @@ export class ClusterReachability extends EventsScope {
...
+    latency: number;
```
marcin-bazyl: latency is already stored in reachability results, why do we need it here?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -290,7 +298,21 @@ export class ClusterReachability extends EventsScope {
     }
 
     if (serverIp) {
-      this.reachedSubnets.add(serverIp);
+      // Checking if the serverIp and port combination already exists in the reachedSubnets
+      const isDuplicate = Array.from(this.reachedSubnets).some(
```
marcin-bazyl: this feels wrong... we have `reachedSubnets` defined as a `Set` so that it's fast and efficient to look up entries in it and check for duplicates, converting it to an array to search for entries doesn't feel right.... I guess you've done this because you only want to check the ip, port and protocol and not latency, best I see no reason why we have `latency` in this set in the first place...

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -460,13 +460,33 @@ export default class Reachability extends EventsScope {
...
-        udp: this.mapTransportResultToBackendDataFormat(clusterResult.udp || {result: 'untested'}),
-        tcp: this.mapTransportResultToBackendDataFormat(clusterResult.tcp || {result: 'untested'}),
-        xtls: this.mapTransportResultToBackendDataFormat(
-          clusterResult.xtls || {result: 'untested'}
-        ),
-      }));
+      results = mapValues(allClusterResults, (clusterResult, clusterKey) => {
+        const clusterReachability = this.clusterReachability[clusterKey];
+        const subnets = clusterReachability
+          ? Array.from(clusterReachability.reachedSubnets || [])
```
marcin-bazyl, line 466: this is not ideal, because reachability results are stored in localStorage, but reachedSubnets is not, so in theory we might want to use getReachabilityResults() without doing reachability and relying on the results from localStorage. We should be storing the reached server ips/ports in reachability results themselves, not using the ones from allClusterResults[..].reachedSubnets - Kacper added them there because he only needed it for sending the metric at the end of reachability, but you guys need to send this to Orpheus together with reachabiltiy results, so it needs to be stored with reachability results. 

Also, the returned data from getReachabilityResults() doesn't match now what it was supposed to match - described in [this wiki](https://confluence-eng-gpk2.cisco.com/conf/display/CLO/New+Reachability+Format+and+Metrics) and once you make it match, you need to update the ReachabilityResultsForBackend type definition too.

marcin-bazyl (changes requested):
This is not the right approach, we should be storing the reached server IP and port in reachability results that we store in boundedStorage as `REACHABILITY.localStorageResult` and make sure that getReachabilityResults() returns them in the format as described on Orpheus wiki. We will need to also take into account that server IPs might be IPV6.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: This is subnet level reachability result. There might be chance in the cluster, one subnet fails and other(s) success. This reachedSubnets is list of all subnets in the cluster. Subnet level details include individual subnet IP(serverIp), its result, time for this subnet, etc.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: This Latency is for individual subnet. And I'll try to avoid using array as Rabbit scan suggesting, I'll use Map.

PR #4321, opened 2025-06-05
fix(plugin-meetings): add strict type check to reachability

k-wasniowski:


marcin-bazyl (approved):
LGTM, but I haven't tested it with the web app. Might be worth doing a deploy from this branch to see if it works correctly?

arun3528 (approved):

PR #4323, opened 2025-06-05
fix: JS-SDK not setting member props for paired devices correctly

marcin-bazyl:
# COMPLETES #[SPARK-672183](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-672183)

Some member properties are not set correctly when we are paired with a device, for example, we have code that tries to set `isSelf`, `isHost` and `isContentSharing` on the paired device, but that code depends on `associatedUsers` that is never set, so these props are always undefined - it's been like that for years. Furthermore, when any in the meeting user is paired with some device, existing code sets `associatedUser` on that device to point to our (self) user, which is wrong.

added new fields to Member class to better describe paired relationships. Fixed the code for setting `associatedUsers` - now this is done in a new function `updateRelationsBetweenMembers()` that's called after all members are added/updated.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manually tested with the web app (joining/leaving meetings with paired device)

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -10,7 +10,7 @@ import MemberUtil from './util';
  * @class Member
  */
 export default class Member {
-  associatedUser: any;
+  associatedUsers: Set<string>; // member ids of users associated with this device, empty if this member is not a device
```
chrisadubois: Should this be participant ids or member ids?

this makes sense to me otherwise

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -29,6 +29,7 @@ export default class Member {
   isRecording: any;
   isRemovable: any;
   isSelf: any;
+  isAssociatedWithSelf: boolean; // true for a device that we are paired with
```
chrisadubois: I am not sure about this one

packages/@webex/plugin-meetings/src/member/index.ts
chrisadubois: Should we change the type of string to MemberId or ParticipantId and have that be a representation of a string?

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -207,7 +212,7 @@ export default class Member {
      * @public
      * @memberof Member
      */
-    this.associatedUser = null;
+    this.associatedUsers = new Set();
```
chrisadubois: new Set<>

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -42,6 +43,10 @@ export default class Member {
   supportLiveAnnotation: boolean;
   type: any;
   namespace = MEETINGS;
+  pairedWith: {
```
chrisadubois, line 48: makes sense to me

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -302,6 +319,7 @@ export default class Member {
   private processParticipant(participant: object) {
     this.participant = participant;
     if (participant) {
+      this.processPairedDevice(participant);
```
chrisadubois, line 334: nice

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -406,5 +406,23 @@ const MemberUtil = {
...
   },
+
+  /**
+   * @param {Object} participant - The locus participant object.
+   * @returns {String}
+   */
+  extractPairedWithParticipantUrl: (participant: any): string => {
+    let participantUrl;
+
+    participant?.devices?.forEach((device: any) => {
```
chrisadubois: looks great

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: I could name it `isDevicePairedWithSelf` to make it more clear, but in theory this could be set to true for any type of participant not just a device, I don't know if that's possible in webex to pair with something else than a device or if maybe in the future we will have such feature, but in theory, based on just what Locus sends us, it is possible, so that's why I kept the name more generic - it matches what Locus sends us - they just have a field "associatedWith" that points to another participant and it could be anything not only a device.

Or we could rename it to "isPairedWithSelf"?

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: MemberId is always set to ParticipantId, so they're the same. Overall I think the idea is that SDK abstracts away Locus participants and provides SDK Members instead, right? so I think fields in Member class should use MemberIds.

Sure, I can add type for MemberId, just be aware it kind of gives you a false sense of type safety, because typescript will still handle this as a string, so if you call a function that takes MemberId argument with some other string instead of MemberId, you won't get any error or warning from typescript. But having MemberId will still be useful for a developer reading the function declaration, so I'll add it.

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: I've renamed it to isPairedWithSelf

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: added, although I'm not sure if it's needed because of the type declaration earlier already stating that

chrisadubois (approved):

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -1,18 +1,19 @@
...
 import {IExternalRoles, IMediaStatus, ParticipantWithBrb, ParticipantWithRoles} from './types';
 
 import MemberUtil from './util';
 
+export type MemberId = string;
 /**
  * @class Member
  */
 export default class Member {
-  associatedUser: any;
```
sreenara, line 13: is it possible for any other usage of this field? Could an external developer have made use of it in any way? Just trying to ensure we aren't breaking existing behavior in external applications.

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -406,5 +392,23 @@ const MemberUtil = {
...
+      device?.intents?.forEach((intent: any) => {
```
sreenara: Could we add types for `device` and `intent` since it's a new objects we're introducing? participant can remain `any` for now.

packages/@webex/plugin-meetings/test/unit/spec/members/index.js
```diff
@@ -23,6 +23,7 @@ import {
   ReclaimHostNotAllowedError,
   ReclaimHostNotSupportedError,
 } from '../../../../src/common/errors/reclaim-host-role-errors';
+import { cloneDeep } from 'lodash';
```
sreenara: is there a linting issue here?

sreenara (changes requested):
Minor comments and questions

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: yes, this is a public property, web app is using it and I guess some other clients may have been using it too. Although this property was never entirely correct, because a device can have multiple users paired to it, so in that case I think until now SDK would just set it to the last user that had a Locus DTO update, now it will contain all the users that are paired to the device. So, in theory we could leave `associatedUser` for backward compatibility and set it to the first entry from `associatedUsers` if you want?, but clients really should switch to using `associatedUsers`

packages/@webex/plugin-meetings/src/member/util.ts
marcin-bazyl: they are not new, Locus has been sending them for years and SDK has been parsing them in various other places and they're all typed as `any` in SDK. I've added some types already in this PR, I can try to add a bit more, but I don't want this PR to grow in scope much more.

marcin-bazyl: done - I've added more types

packages/@webex/plugin-meetings/test/unit/spec/members/index.js
marcin-bazyl: fixed

packages/@webex/plugin-meetings/src/member/util.ts
```diff
@@ -406,5 +392,23 @@ const MemberUtil = {
...
+    participant?.devices?.forEach((device) => {
+      device?.intents?.forEach((intent) => {
+        if (intent?.type === _OBSERVE_ && intent?.associatedWith) {
+          participantUrl = intent.associatedWith;
+        }
+      });
+    });
+
+    return participantUrl;
+  },
```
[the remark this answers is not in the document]
marcin-bazyl: fixed return type

packages/@webex/plugin-meetings/src/member/index.ts
sreenara: Yes @marcin-bazyl , let's leave it for backward compatibility with a comment for now. Would like to ensure we don't break it for anyone who's already using it, and we'll phase it out in some other manner.

marcin-bazyl: done

packages/@webex/plugin-meetings/src/member/index.ts
```diff
@@ -468,12 +508,9 @@ export default class Member {
    * @private
    * @memberof Member
    */
-  private processIsSelf(participant: object, selfId: string) {
+  private processIsSelf(participant: Participant, selfId: string) {
     if (MemberUtil.isUser(participant)) {
       this.isSelf = MemberUtil.isSame(participant, selfId);
-    } else if (MemberUtil.isDevice(participant)) {
```
sreenara, line 474: @marcin-bazyl we should return this as well right? Else wouldn't `this.associatedUser` be `null` always?

sreenara (commented):
Just one comment on removal of existing code.

packages/@webex/plugin-meetings/src/member/index.ts
marcin-bazyl: This code here was incorrect, it was always setting `associatedUser` to selfId, which is not correct, because device might be paired with different user. Anyway, I'm setting `associatedUser` now correctly inside `updateRelationsBetweenMembers()`, see commit 6a0401b033a188e0f7c6469159166583e188f111

sreenara (approved):

PR #4347, opened 2025-06-16
fix(calling): metric improvements for registration flows

Kesari3008:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6516 

Updating the fields for registration metrics to cover below details:
1. Server type whether it is primary or backup
2. Error reason
3. Caller of registration flow
Updated lineErrorContext for 429 error case so that metric reflects the error message and error type

Added new property fields within submitRegistrationMetric() method.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

Screenshot 2025-06-17 at 10 39 01 PM

Vidcast: https://app.vidcast.io/share/756884cd-cc36-4781-808d-9b2cf8d54c23?playerMode=vidcast 

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [x] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/common/Utils.ts
```diff
@@ -355,6 +355,13 @@ export async function handleRegistrationErrors(
 
     case ERROR_CODE.TOO_MANY_REQUESTS: {
       log.warn(`429 Too Many Requests`, loggerContext);
+      updateLineErrorContext(
```
sreenara, line 362: Could we talk about this change as well in the description?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -262,6 +262,10 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
               METRIC_EVENT.REGISTRATION_ERROR,
               REG_ACTION.REGISTER,
               METRIC_TYPE.BEHAVIORAL,
+              GET_MOBIUS_SERVERS_UTIL,
```
sreenara, line 288: Are there tests for this and all the branches?

packages/calling/src/Metrics/types.ts
```diff
@@ -47,7 +47,11 @@ export interface IMetricManager {
     name: METRIC_EVENT,
     metricAction: REG_ACTION,
     type: METRIC_TYPE,
-    error: LineError | CallingClientError | undefined
+    caller: string,
+    serverType: string,
```
sreenara: can we create a type with limited set of values?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -262,6 +262,10 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+              '',
+              (err as WebexRequestPayload).headers?.trackingId ?? '',
```
sreenara, line 290: What are the scenarios when we don't have a trackingId in the response?

packages/calling/src/CallingClient/registration/register.ts
sreenara, line 1: I don't see test changes for this file.. am i missing them?

sreenara (changes requested):
Please see the comments

packages/calling/src/Metrics/types.ts
```diff
@@ -47,7 +47,11 @@ export interface IMetricManager {
...
+    trackingId: string,
+    keeapliveCount?: number,
+    error?: LineError | CallingClientError
```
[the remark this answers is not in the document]
Kesari3008: Addressed

packages/calling/src/Metrics/types.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: there were no test changes needed here

packages/calling/src/common/Utils.ts
Kesari3008: done

packages/calling/src/Metrics/types.ts
```diff
@@ -43,6 +43,11 @@ export enum VOICEMAIL_ACTION {
   TRANSCRIPT = 'transcript',
 }
 
+export enum SERVER_TYPE {
```
sreenara:
```suggestion
export SERVER_TYPE = 'primary' | 'backup' | 'unknown'
```

sreenara: I think this might be better, it's stricter than the enum.

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -175,6 +179,16 @@ describe('Registration Tests', () => {
         method: expect.any(String),
       })
     );
+    expect(metricSpy).toBeCalledWith(
+      METRIC_EVENT.REGISTRATION,
+      REG_ACTION.REGISTER,
+      METRIC_TYPE.BEHAVIORAL,
+      REGISTRATION_UTIL,
+      SERVER_TYPE.PRIMARY,
+      '',
```
sreenara: We should return a tracking ID in the webex.request mock response so that it can be filled here to satisfy the condition of the tracking ID being present.

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -645,6 +653,10 @@ export class Registration implements IRegistration {
       return abort;
     }
     for (const url of servers) {
+      const serverType =
```
sreenara, line 656: Are we testing all 3 conditions here? Can you please point me to the tests that have this?

packages/calling/src/CallingClient/registration/register.test.ts
sreenara: Talking about this : `resp.headers?.trackingid`

sreenara (commented):
Please see the comments

packages/calling/src/CallingClient/CallingClient.ts
sreenara: Discussed this offline. trackingId remains same in request and response.

packages/calling/src/CallingClient/constants.ts
```diff
@@ -109,6 +109,7 @@ export const REG_TRY_BACKUP_TIMER_VAL_IN_SEC = 1200;
 export const REG_TRY_BACKUP_TIMER_VAL_FOR_CC_IN_SEC = 114;
 export const REG_FAILBACK_429_MAX_RETRIES = 5;
 export const RETRY_TIMER_UPPER_LIMIT = 60;
+export const REGISTRATION_UTIL = 'triggerRegistration';
```
rarajes2, line 112: Q. Why are we creating constants with different value than the key?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -645,6 +653,10 @@ export class Registration implements IRegistration {
...
+        (this.primaryMobiusUris.includes(url) && SERVER_TYPE.PRIMARY) ||
+        (this.backupMobiusUris?.includes(url) && SERVER_TYPE.BACKUP) ||
+        SERVER_TYPE.UNKNOWN;
```
rarajes2: Should we create a separate util function for this?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -672,17 +684,21 @@ export class Registration implements IRegistration {
           METRIC_EVENT.REGISTRATION,
           REG_ACTION.REGISTER,
           METRIC_TYPE.BEHAVIORAL,
+          caller,
+          serverType,
+          resp.headers?.trackingid ?? '',
+          undefined,
```
rarajes2, line 690: Would it make sense to group these together as one param and call it `payload` ?

rarajes2 (commented):
1. How are we testing if the registration was due to fallback ? For failover we can see if it was secondary server
2. Can we add a vidcast explaining the usecase?

sreenara (approved):

packages/calling/src/CallingClient/constants.ts
Kesari3008: This is for method names reflecting properly as currently we have seen it's not working with this.triggerRegistration.name pattern.

Now why the value is different than key is because for the key name, we are following a common pattern whicch is <flowname>_UTIL and value has the be exact same method name

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: It is only being done in one place so creating util would be overengineering for now

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: I would anyway have to create that object everywhere before passing it to the method right

Kesari3008:
> 1. How are we testing if the registration was due to fallback ? For failover we can see if it was secondary server
> 2. Can we add a vidcast explaining the usecase?

1. Reg source will tell the registration flow caller

packages/calling/src/CallingClient/registration/register.ts
rarajes2: Right, I thought it would be cleaner to pass these args as a single object param instead of passing multiple params. That way if we need to pass more args later, we can update without altering the method signature except the type.

packages/calling/src/CallingClient/registration/register.ts
rarajes2: We don't have to create it as a util function in a separate file. We can keep it in this file itself. It will reduce the function's complexity

rarajes2 (approved):

PR #4371, opened 2025-06-26
fix(brbState): sync issue

SomeBody16:
# COMPLETES [SPARK-669381](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-669381)

When connected to the same meeting from 2 web clients and we:

1. On client A: step away
2. On client B: step away
3. On client A: go back to the meeting
4. On client B: go back to the meeting

client A will see client B as 'away' because client B did not update his source state

If local state matches remote state of away, update source state override to make sure it's relevant

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Double web client described in `This pull request addresses`
- Automated tests 

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -141,10 +147,8 @@ describe('plugin-meetings', () => {
     it('should reject when sendLocalBrbStateToServer fails', async () => {
       const error = new Error('send failed');
       setBrbStub.rejects(error);
-        
-      await expect(
-        brbState.enable(true, meeting.sendSlotManager)
-      ).to.be.rejectedWith(error); 
+
+      await expect(brbState.enable(true, meeting.sendSlotManager)).to.be.rejectedWith(error);
```
SomeBody16, line 151: prettier change

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -23,7 +22,7 @@ describe('plugin-meetings', () => {
         setSourceStateOverride: sinon.stub(),
       },
       meetingRequest: {
-        setBrb: () => {}
+        setBrb: () => {},
```
SomeBody16, line 26: prettier change

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -3,12 +3,11 @@ import {assert, expect} from '@webex/test-helper-chai';
 
 import testUtils from '../../../utils/testUtils';
 import {BrbState, createBrbState} from '@webex/plugin-meetings/src/meeting/brbState';
-import LoggerProxy from '@webex/plugin-meetings/src/common/logs/logger-proxy';
 
 describe('plugin-meetings', () => {
   let meeting: any;
   let brbState: BrbState;
-  let setBrbStub: sinon.SinonStub;  
+  let setBrbStub: sinon.SinonStub;
```
SomeBody16, line 11: prettier change

k-wasniowski (approved):

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -117,6 +119,19 @@ export class BrbState {
       });
   }
 
+  /**
+   * Updates the local state of the brb based on the enabled parameter.
+   *
+   * @param {SendSlotManager} sendSlotManager - the send slot manager instance
+   * @returns {Promise<void>}
+   */
+  private updateLocalState(sendSlotManager: SendSlotManager) {
```
brycetham: A few things:
1. This function name confused me for a bit because it wasn't clear to me what "local state" refers to. I think all we're doing is setting the source state override here, so maybe `updateSourceStateOverride` makes better sense?
2. Should we be checking if we need to update the source state override all (i.e. only call `setSourceStateOverride` if the source state is incorrect)? At the end of the day it probably doesn't matter too much, because we never send duplicate JMP messages anyways, but we do log whenever that happens, so maybe we should try to prevent it?

antsukanova: I have similar thoughts about naming, it's not `updateLocalState` for me because we never have such logic for server-only state and overall applying `sendSlotManager` logic to the server, `updateSourceStateOverride` looks better.

I see also duplicated logic for the current `updateLocalState`, feels good to consider optimization for this flow to not have `setSourceStateOverride` twice here:

Screenshot 2025-06-27 at 17 32 08

SomeBody16: 1. @brycetham @antsukanova I agree and I've changed to `updateSourceStateOverride`
2. @brycetham I'm not sure if we can from this place get the current source state?
We can do
`SendSlotManager.getSlot(MediaType.VideoMain) -> SendSlot`
but SendSlot doesn't have public API to get source state (tranceiver is private there)
3. @antsukanova I'm not sure which logic is duplicated on this screenshot, `updateLocalState` is breaking the flow and second call (inside `sendLocalBrbStateToServer` is at the end of it and called only if everything succeed, so I wouldn't call it duplicated

antsukanova:
@SomeBody16 Double call easily happens when u use step away now. Where u didn't not able to see it?
Screenshot 2025-07-01 at 15 07 51

packages/@webex/plugin-meetings/src/meeting/brbState.ts
antsukanova: 3. @SomeBody16 However, I feel we do not need two calls in the flow. Why do you want to keep both calls? I would prefer to make only one override call for successful brb flow.

brycetham: As discussed offline with point 2, perhaps we can think about storing the overridden source state in `SendSlotManager`, and use that to control whether or not to call the WCME `setSourceStateOverride` API.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -117,6 +119,24 @@ export class BrbState {
       });
   }
 
+  /**
```
antsukanova, line 125: didn't see usage of `sendSlotManager` in `sendLocalBrbStateToServer`

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -105,15 +115,32 @@ export default class SendSlotManager {
...
+   * @returns {StreamState | null} - The current source state override or null if not set.
+   */
+  public getSourceStateOverride(mediaType: MediaType): StreamState | null {
+    if (mediaType !== MediaType.VideoMain) {
+      throw new Error(
+        `sendSlotManager cannot get source state override which media type is ${mediaType}`
+      );
+    }
+
+    return this.sourceStateOverride.get(mediaType) || null;
```
antsukanova: Where is test cases for `getSourceStateOverride`?

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -117,6 +119,24 @@ export class BrbState {
...
+  private updateSourceStateOverride(sendSlotManager: SendSlotManager): void {
+    const isOverrideEnabled =
+      sendSlotManager.getSourceStateOverride(MediaType.VideoMain) === 'away';
+    const isClientEnabled = this.state.client.enabled;
+
+    if (isClientEnabled && !isOverrideEnabled) {
+      sendSlotManager.setSourceStateOverride(MediaType.VideoMain, 'away');
+    } else if (!isClientEnabled && isOverrideEnabled) {
+      sendSlotManager.setSourceStateOverride(MediaType.VideoMain, null);
+    }
```
antsukanova: Could we have something like that? Feels more readable

```
if (isClientEnabled && !isOverrideEnabled) {
  sendSlotManager.setSourceStateOverride(MediaType.VideoMain, 'away');
}
if (!isClientEnabled && isOverrideEnabled) {
  sendSlotManager.setSourceStateOverride(MediaType.VideoMain, null);
}
```

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -117,6 +119,24 @@ export class BrbState {
...
 
+  /**
+   * Updates the local state of the brb based on the enabled parameter.
+   *
+   * @param {SendSlotManager} sendSlotManager - the send slot manager instance
+   * @returns {void}
+   */
+  private updateSourceStateOverride(sendSlotManager: SendSlotManager): void {
+    const isOverrideEnabled =
+      sendSlotManager.getSourceStateOverride(MediaType.VideoMain) === 'away';
```
antsukanova: I'm not sure in `enabled` here gets me correct understanding of this logic, I would rather say here `isSourceStateOverridden`

packages/@webex/plugin-meetings/src/meeting/brbState.ts
```diff
@@ -117,13 +119,31 @@ export class BrbState {
...
+   *
+   * @param {SendSlotManager} sendSlotManager - the send slot manager instance
+   * @returns {void}
+   */
+  private updateSourceStateOverride(sendSlotManager: SendSlotManager): void {
+    const isSourceStateOverridden =
+      sendSlotManager.getSourceStateOverride(MediaType.VideoMain) === 'away';
+    const isClientEnabled = this.state.client.enabled;
+
+    if (isClientEnabled && !isSourceStateOverridden) {
```
brycetham: I wonder if it makes more sense for this check to be in `SendSlotManager`. That is, `updateSourceStateOverride` will always call `sendSlotManager.setSourceStateOverride`, but `SendSlotManager` will only call the WCME API if the override state has changed. That way, if we override in other parts of the code in the future (not just `brbState`), we don't have to write checks like this again, since all the checks will be in `SendSlotManager`.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -7,10 +7,20 @@ import {
   StreamState,
 } from '@webex/internal-media-core';
 
+/**
+ * This class is used to manage the sendSlots for the given media types.
+ */
 export default class SendSlotManager {
   private readonly slots: Map<MediaType, SendSlot> = new Map();
   private readonly LoggerProxy: any;
+  private sourceStateOverride: Map<MediaType, StreamState | null> = new Map();
```
brycetham: A few minor things:

1. Can this be `readonly`? Only the contents of the map I think are being changed; the map itself is never replaced.
2. Can this be renamed to `sourceStateOverrides` (since there is more than one, similar to `slots`).
3. I noticed the value can be set to `null` here, yet...

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -105,15 +115,32 @@ export default class SendSlotManager {
 
     if (state) {
       slot.setSourceStateOverride(state);
+      this.sourceStateOverride.set(mediaType, state);
     } else {
       slot.clearSourceStateOverride();
+      this.sourceStateOverride.delete(mediaType);
```
brycetham: ...here, you delete instead of setting to `null`. Should we try to be consistent here?

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -5,7 +5,7 @@ import 'jsdom-global/register';
 
 // Polyfill for crypto: https://github.com/jsdom/jsdom/issues/1612#issuecomment-663210638
 import {Crypto} from '@peculiar/webcrypto';
-global.crypto = new Crypto();
+// global.crypto = new Crypto();
```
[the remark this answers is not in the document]
brycetham: @SomeBody16 I think we can delete this if it's unneeded, unless this change was unintentional, since I don't see how it's related to your fix.

packages/@webex/plugin-meetings/src/meeting/brbState.ts
SomeBody16: I agree, but the question is, should we log that? If yes, these checks were added because we wanted to remove necessary logs in the browser so they should stay here

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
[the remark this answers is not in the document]
SomeBody16: Committed by accident, with this line not commented my local tests are failing but I've added this to commit by accident

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -103,17 +113,39 @@ export default class SendSlotManager {
...
       `SendSlotsManager->setSourceStateOverride#set source state override for ${mediaType} to ${state}`
     );
   }
 
+  /**
+   * Gets the source state override for the given media type.
+   * @param {MediaType} mediaType - The type of media to get the source state override for.
+   * @returns {StreamState | null} - The current source state override or null if not set.
+   */
+  public getSourceStateOverride(mediaType: MediaType): StreamState | null {
```
brycetham: This doesn't need to be public right? No one else needs to use this other than `SendSlotManager`?

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -104,12 +104,19 @@ describe('plugin-meetings', () => {
       assert.isTrue(meeting.meetingRequest.setBrb.calledOnce);
     });
 
-    it('sets source state override when client state does not match server state', async () => {
+    it('update source state override when client state does not match server state', async () => {
```
brycetham: Nit: change "update" to "updates", so this reads "it updates source state override".

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -104,12 +104,19 @@ describe('plugin-meetings', () => {
...
-      assert.isTrue(meeting.sendSlotManager.setSourceStateOverride.calledOnce);
+      assert.isTrue(meeting.sendSlotManager.setSourceStateOverride.called);
+    });
+
+    it('updates set source state override when client state does match server state', async () => {
+      brbState.enable(false, meeting.sendSlotManager);
+      await testUtils.flushPromises();
+
+      assert.isTrue(meeting.sendSlotManager.setSourceStateOverride.called);
     });
```
brycetham: I'm confused about these tests, but maybe it's because we've moved a lot of the logic to `SendSlotManager`. Should we just test that `sendSlotManager.setSourceStateOverride` is called with the correct state?

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
```diff
@@ -141,10 +148,8 @@ describe('plugin-meetings', () => {
     it('should reject when sendLocalBrbStateToServer fails', async () => {
       const error = new Error('send failed');
       setBrbStub.rejects(error);
-        
-      await expect(
-        brbState.enable(true, meeting.sendSlotManager)
-      ).to.be.rejectedWith(error); 
+
+      await expect(brbState.enable(true, meeting.sendSlotManager)).to.be.rejectedWith(error);
```
brycetham: Maybe also check here that `sendSlotManager.setSourceStateOverride` is called with the correct state.

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -272,4 +272,62 @@ describe('SendSlotsManager', () => {
...
+        } as MultistreamRoapMediaConnection;
+      });
+
+      it('can set source state override', () => {
+        const slot: any = sendSlotsManager.createSlot(mediaConnection, MediaType.VideoMain);
+
+        sendSlotsManager.setSourceStateOverride(MediaType.VideoMain, 'away');
+
+        expect(slot.setSourceStateOverride.calledWith('away')).to.be.true;
+      });
```
brycetham: Do we need this test if we're already testing `VideoMain` below?

SomeBody16: I agree, I'll remove this test

packages/@webex/plugin-meetings/test/unit/spec/meeting/brbState.ts
SomeBody16: It was added in the process of changing this place, it was suppose to ensure that we are updating source state override on every call, but currently these tests are not refering to any part of code, all remove the second test and improve the first one

oleksiijko (approved):

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -103,17 +113,39 @@ export default class SendSlotManager {
       throw new Error(`Slot for ${mediaType} does not exist`);
     }
 
+    const currentStateOverride = this.getSourceStateOverride(mediaType);
+    if (currentStateOverride === state) {
+      return;
+    }
+
```
antsukanova, line 120: Great change!

antsukanova: Where are the tests for this logic? Can find them...

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -272,4 +272,54 @@ describe('SendSlotsManager', () => {
...
+      beforeEach(() => {
+        mediaConnection = {
+          createSendSlot: sinon.stub().returns({
+            setSourceStateOverride: sinon.stub().resolves(),
+            clearSourceStateOverride: sinon.stub().resolves(),
+          }),
+        } as MultistreamRoapMediaConnection;
+      });
+
+      [MediaType.VideoMain].forEach((mediaType) => {
```
edvujic: Why do you have a `forEach` for only one `MediaType`?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -103,17 +113,39 @@ export default class SendSlotManager {
...
 
+  /**
+   * Gets the source state override for the given media type.
+   * @param {MediaType} mediaType - The type of media to get the source state override for.
+   * @returns {StreamState | null} - The current source state override or null if not set.
+   */
+  private getSourceStateOverride(mediaType: MediaType): StreamState | null {
+    if (mediaType !== MediaType.VideoMain) {
+      throw new Error(
+        `sendSlotManager cannot get source state override which media type is ${mediaType}`
```
edvujic: I feel like this can be improved. It's a bit confusing. 

```suggestion
        `Invalid media type '${mediaType}'. Source state overrides are only applicable to MediaType.VideoMain.`
```

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -103,17 +113,39 @@ export default class SendSlotManager {
...
     if (state) {
       slot.setSourceStateOverride(state);
+      this.sourceStateOverrides.set(mediaType, state);
     } else {
       slot.clearSourceStateOverride();
+      this.sourceStateOverrides.delete(mediaType);
     }
 
     this.LoggerProxy.logger.info(
       `SendSlotsManager->setSourceStateOverride#set source state override for ${mediaType} to ${state}`
```
edvujic:
```suggestion
      `SendSlotManager->setSourceStateOverride#set source state override for ${mediaType} to ${state}`
```

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -7,10 +7,20 @@ import {
...
+/**
+ * This class is used to manage the sendSlots for the given media types.
+ */
 export default class SendSlotManager {
   private readonly slots: Map<MediaType, SendSlot> = new Map();
   private readonly LoggerProxy: any;
+  private readonly sourceStateOverrides: Map<MediaType, StreamState> = new Map();
 
+  /**
+   * Constructor for SendSlotsManager
```
edvujic:
```suggestion
   * Constructor for SendSlotManager
```

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
SomeBody16: This one is taken from https://github.com/webex/webex-js-sdk/pull/4371/files#diff-64c869d0e3937b132006eda7d35ae17e4b950be255eb402f6803af594bf6528cR80

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
SomeBody16: To match the lower test, so these media types (MediaType.VideoMain) can be set, and these (MediaType.VideoSlides, MediaType.AudioMain, MediaType.AudioSlides) can't

edvujic: so why not do something like:

```
[
  {mediaType: MediaType.VideoMain, shouldThrow: false, shouldBeCalled: true},
  {mediaType: MediaType.VideoSlides, shouldThrow: true, shouldBeCalled: false},
  {mediaType: MediaType.AudioMain, shouldThrow: true, shouldBeCalled: false},
  {mediaType: MediaType.AudioSlides, shouldThrow: true, shouldBeCalled: false},
].forEach(({mediaType, shouldThrow, shouldBeCalled}) => {
  it(`${shouldThrow ? "can't" : 'can'} set source state override for ${mediaType}`, () => {
    const slot: any = sendSlotsManager.createSlot(mediaConnection, mediaType);
    const set = () => sendSlotsManager.setSourceStateOverride(mediaType, 'away');

    if (shouldThrow) {
      expect(set).to.throw();
    } else {
      expect(set).not.to.throw();
    }

    if (shouldBeCalled) {
      expect(slot.setSourceStateOverride.calledWith('away')).to.be.true;
    } else {
      expect(slot.setSourceStateOverride.called).to.be.false;
    }
  });
});

```

I tested it, and it works fine.

SomeBody16: Yes, but there is `if` for every `expect`, even in test description so it's like two separate tests. This is why I've separated them for more readability even while this is working. So I can easily see which media type is supported and which not from the tests

SomeBody16:
This is ready to merge

antsukanova (approved):

sreenara (approved):

PR #4390, opened 2025-07-11
fix: client sometimes gets stuck in a loop syncing with Locus

marcin-bazyl:
# COMPLETES #[SPARK-686383](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-686383)

## This pull request addresses
When Locus sync fails, we destroy the meeting, but the Locus parser instance is still running and keeps trying to do the sync. Also Locus team confirmed that they might sometimes return 403 when we try to sync (even if earlier we were allowed to get the DTOs) and in that case there is no point trying the fallback to full sync

## by making the following changes
not resuming the parser when meeting is destroyed and not doing the fallback to full sync when we get 403

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

unit tests, manual test with the web app

## The GAI Coding Policy And Copyright Annotation Best Practices ##
  
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [ ] GAI was used to create a draft that was subsequently customized or modified
- [x] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [ ] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

- [x] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [x] I discussed changes with code owners prior to submitting this pull request

- [x] I have not skipped any automated checks
- [x] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #4401, opened 2025-07-18
fix: handle case when we get full DTO when we asked for delta

marcin-bazyl:
After PR https://github.com/webex/webex-js-sdk/pull/4390 the amount of Locus syncs in a loop dropped, but still we are seeing a large number of failures to process the response from the sync, majority of them with the error:
`TypeError: Cannot read properties of undefined (reading &#39;entries&#39;)`

This is caused by the fact that Locus sometimes returns a full DTO (that doesn't contain the baseSequence) instead of the delta one that JS-SDK assumes.

amplitude link for the response processing failures:
https://app.eu.amplitude.com/analytics/cisco-cross-project/chart/new/e-61xtmfxa?source=copy+url

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/index.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/index.ts
@@ -159,17 +159,32 @@ export default class LocusInfo extends EventsScope {
         throw e;
       })
       .then((res) => {
-        if (isDelta) {
-          if (!isEmpty(res.body)) {
-            meeting.locusInfo.handleLocusDelta(res.body, meeting);
-          } else {
+        if (isEmpty(res.body)) {
+          if (isDelta) {
             LoggerProxy.logger.info(
               'Locus-info:index#doLocusSync --> received empty body from syncUrl, so we already have latest Locus DTO'
             );
+          } else {
+            LoggerProxy.logger.info(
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -159,17 +159,32 @@ export default class LocusInfo extends EventsScope {
...
-          if (!isEmpty(res.body)) {
-            meeting.locusInfo.handleLocusDelta(res.body, meeting);
-          } else {
+        if (isEmpty(res.body)) {
+          if (isDelta) {
             LoggerProxy.logger.info(
               'Locus-info:index#doLocusSync --> received empty body from syncUrl, so we already have latest Locus DTO'
             );
+          } else {
+            LoggerProxy.logger.info(
```
marcin-bazyl, line 168: previously the code was not checking for empty body when `isDelta===false` and just calling `meeting.locusInfo.onFullLocus()` but there it would still return early without doing anything, because `isNewFullLocus()` would return false, so now I'm just doing this check for empty body earlier here for all cases and not calling `meeting.locusInfo.onFullLocus()`.

brycetham (approved):

chrisadubois (approved):

PR #4416, opened 2025-07-29
fix: updated u2c logic to filter out urls with negative priority

johnsoter13:

- Current u2c logic doesn't support update to u2c backend that marks some service urls with a negative priority (-1)

- Updating filter function when looking for the priority host that removes urls with negative priority as well as hosts that are marked as failed (current logic only looks if url is marked as failed)

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Manually tested, and tests written

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/lib/services/service-url.js
+++ b/packages/@webex/webex-core/src/lib/services/service-url.js
@@ -60,7 +60,7 @@ const ServiceUrl = AmpState.extend({
       ? this.hosts.filter((host) => host.id === clusterId)
       : this.hosts.filter((host) => host.homeCluster);
 
-    const aliveHosts = filteredHosts.filter((host) => !host.failed && host.priority > 1);
+    const aliveHosts = filteredHosts.filter((host) => !host.failed);
 
     filteredHosts =
       aliveHosts.length === 0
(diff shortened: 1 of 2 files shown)


packages/@webex/webex-core/src/lib/services/service-url.js
```diff
@@ -75,7 +75,9 @@ const ServiceUrl = AmpState.extend({
     return this._generateHostUrl(
       filteredHosts.reduce(
         (previous, current) =>
-          previous.priority > current.priority || !previous.homeCluster ? current : previous,
+          current.priority > 0 && (previous.priority > current.priority || !previous.homeCluster)
```
chrisadubois, line 78: I think we need to do a couple more things

1. homeCluster should not override priority decisions (this is a problem with the current implementation)
2. the array probably needs to be pre-sorted by priority

As it is right now, -1 urls would be removed, but we could still have a logic error based on the following
A. If homeCluster is set to true for an "invalid" FQDN
B. if the order of the array is incorrect

chrisadubois (changes requested):

chrisadubois:
From andrew biggs -> https://wwwin-github.cisco.com/pages/Webex/crr-docs/techdocs/rr-005.html#intolerant-clients

packages/@webex/webex-core/src/lib/services/service-url.js
chrisadubois: I also think we need Colin's input on this re: homeCluster since he added this part

robstax: did we decide we don't have to do these? i thought we could merge this as is?

PR #4432, opened 2025-08-12
fix(breakouts): failed to end breakout sessions

xinhyao:
# COMPLETES #
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-666148

When the host clicks to end breakout sessions and the countdown ends, the session cannot be successfully switched from BO to meeting. As a result, the Web client sometimes still tries to send messages to BO Locus instead of main session Locus.
image
When the host clicks to end BO and after the countdown ends, SDK will receive the `LOCUSEVENT.LOCUS_MERCURY` event. By parsing, the event belongs to `LOCUS_URL_CHANGED` case. According to the existing logic, it will call `doLocusSync `to do the sync with a new locus, but for this bug case, the entry value of the sequence in the new fetched locus is the same as the entry value of the sequence in the `workingCopy`. This will cause the `lociComparison` to be `USE_CURRENT`, for the `USE_CURRENT` case, SDK will do nothing, so the switch from BO to main session is not completed when the BO ends countdown.
Pasted Graphic 21

image

This issue has a relatively high probability of reoccurring (I tested it 10 times and reoccurred 7 times). For those successful cases, I found that when the countdown ends, the SDK will additionally receive the `LOCUSEVENT.LOCUS_MERCURY` event, by parsing, the event belongs to `USE_INCOMING` case. According to the existing logic, it will call `meeting.locusInfo.onDeltaLocus` directly, so locus info can be updated to complete the switch from BO to main session, but this case is not very stable and does not depend on `doLocusSync()`, so my current modification is that when the received action is `LOCUS_URL_CHANGED`, call `meeting.locusInfo.onDeltaLocus` instead of `doLocusSync` directly. In this way, locus info can be updated to complete the switch from BO to main session.
Pasted Graphic 24
Failure log:
LOCUS_URL_CHANGED_ACTION===
Success log:
image
According to the log printed on the console, we can see that for successful cases, it is because the SDK received additional events, which triggered the `meeting.locusInfo.onDeltaLocus`, but this is unstable.

This PR is changed on this PR(https://github.com/webex/webex-js-sdk/pull/3239).

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added unit tests and manually tested.
I tested:

1. Host starts a meeting and 2 others web clients join this meeting, test audio, video, share, chat, content, they work fine.
2. Host enables breakouts sessions and joins BO session 1, client A joins BO session 1, client B joins BO session 2.
3. Test audio, video, share, chat, they work fine.
4. Host ends BO sessions and all clients join the main session at the same time.
5. Test audio, video, share, chat, content, they work fine.

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -220,14 +220,14 @@ export default class LocusInfo extends EventsScope {
 
     switch (action) {
       case USE_INCOMING:
+      case LOCUS_URL_CHANGED:
```
marcin-bazyl: I'm not convinced that this is the right approach. I think it would be better to keep doing the full sync when Locus URL changes. I think the problem is that in [this line](https://github.com/webex/webex-js-sdk/blob/next/packages/%40webex/plugin-meetings/src/locus-info/parser.ts#L735) for the LOCUS_URL_CHANGED case we put the incoming locus that had url changed into the working copy, so then when we do a sync there is a big chance that the full locus in response will be on the same version as our working copy, so I think we need modify that line to either clear the working copy completely (because we want to do a full sync anyway) or at least just reset the sequence in workingCopy so that it's empty and then any sequence value from locus sync response will be always "newer" and will result in USE_INCOMING

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/locus-info/index.ts
xinhyao: Thanks for your suggestions, I agreed with you, for the LOCUS_URL_CHANGED case, I put the incoming locus into the working copy and reset the sequence, then it does result in 
USE_INCOMING case, and I found that the locus in response lose the participants info, and it will result errors after the breakout session ends in this case, so if the participants from locus 
sync response  are empty, I add the participants from workingCopy to the new locus, in this way, it can work correctly.
I also tested clearing the working copy completely, for the current logic, it will use `meeting.locusUrl` to get the full locus, but I receive the error response, so I use the first solution above.
let meetingDestroyed = false;

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -172,6 +173,14 @@ export default class LocusInfo extends EventsScope {
           return;
         }
 
+        // When the locus url changed, if the participants from locus sync response are empty, add the participants from workingCopy to the new locus,
+        // otherwise the participants info will be lost after the breakout session ends.
+        if (isLocusUrlChanged) {
+          if (!res.body.participants?.length) {
+            res.body.participants = this.locusParser.workingCopy.participants;
```
marcin-bazyl: but is there any guarantee that the `workingCopy.participants` is the full list of participants? it will contain only the participants from the last delta we received (the one that had the new locus url), so maybe during testing you got lucky to have all the participants, but in general it's not guaranteed, right?

marcin-bazyl: maybe the reason why you got empty participants in the response is because we did a "delta" locus sync so we used the `this.locusParser.workingCopy.syncUrl` as the `url`. When we detect locus url change, we should always do a full sync and use the `meeting.locusUrl` as the `url` - then I think we would get the full list of participants in the response - can you check this?

xinhyao: > but is there any guarantee that the `workingCopy.participants` is the full list of participants? it will contain only the participants from the last delta we received (the one that had the new locus url), so maybe during testing you got lucky to have all the participants, but in general it's not guaranteed, right?

yes, you are right, `workingCopy.participants` only contain the participants from the last delta, I changed it.

xinhyao: > maybe the reason why you got empty participants in the response is because we did a "delta" locus sync so we used the `this.locusParser.workingCopy.syncUrl` as the `url`. When we detect locus url change, we should always do a full sync and use the `meeting.locusUrl` as the `url` - then I think we would get the full list of participants in the response - can you check this?

Yes, for the LOCUS_URL_CHANGED case, we should do a full locus sync instead of a delta locus sync, I tried to use `meeting.locusUrl` as the url to get locusDTO, but the reponse returned the `Join the meeting before retrieving the roster` error, I suspect that `meeting.locusUrl` at this moment is the BO session locus url,  we should do a full sync on the main session instead of the BO session, so I use the cached main session locus url to do a full sync, and it worked.

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -93,18 +93,25 @@ export default class LocusInfo extends EventsScope {
...
     let url;
     let meetingDestroyed = false;
 
-    if (this.locusParser.workingCopy.syncUrl) {
+    if (this.locusParser.workingCopy?.syncUrl) {
       url = this.locusParser.workingCopy.syncUrl;
       isDelta = true;
     } else {
-      url = meeting.locusUrl;
+      if (isLocusUrlChanged) {
```
marcin-bazyl: shouldn't this check be done at the beginning, before the check for `(this.locusParser.workingCopy?.syncUrl)`?

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -93,18 +93,25 @@ export default class LocusInfo extends EventsScope {
...
+        // If the locus url changed, we should always do a full sync,
+        // use the main session locus url to get the full locus(full participants list in the response)
+        url = this.mainSessionLocusCache.url;
```
marcin-bazyl: what if the locusUrl change is caused by the change in the other direction: from main to breakout? we should know the new locus url, because that's how we detect the url change, by receiving the new one, right? just need to make sure it's stored somewhere

xinhyao: > what if the locusUrl change is caused by the change in the other direction: from main to breakout? we should know the new locus url, because that's how we detect the url change, by receiving the new one, right? just need to make sure it's stored somewhere

I checked our current logic, for the case from main session to breakout session, the last new delta locus(locus url is breakout) did not contain the `baseSequence` property, so it will not execute the `compareDelta` method which means that it will not result in LOCUS_URL_CHANGED case(because it is determined in `compareDelta` method whether the locus URL has changed ). It will result in USE_INCOMING case.

But for the case from breakout to main session, the last new delta locus(locus url is main session) contained the `baseSequence` property, then it will execute `compareDelta` method and return LOCUS_URL_CHANGED comparison, and then we will do a full locus sync.

So I think that in this line, we don’t need to care about the locusUrl changed direction(from main to breakout), because only locusUrl changed from breakout to  main session can result in LOCUS_URL_CHANGED case.
said not contain ase sequence poperty, it will not execute this line
const (extractConparisonState extract) = Parser

marcin-bazyl (approved):

JudyZhuHz (approved):

PR #4464, opened 2025-09-09
feat(plugin-meetings): include suffix value

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-708364
Changed editDisplaName to also edit suffix

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/members/index.ts
+++ b/packages/@webex/plugin-meetings/src/members/index.ts
@@ -1181,11 +1181,17 @@ export default class Members extends StatelessWebexPlugin {
    * @param {string} memberId - id of the participant who is receiving request
    * @param {string} requestingParticipantId - id of the participant who is sending request (optional)
    * @param {string} [alias] - alias name
+   * @param {string} [suffix] - suffix name (optional)
    * @returns {Promise}
    * @public
    * @memberof Members
    */
-  public editDisplayName(memberId: string, requestingParticipantId: string, alias: string) {
+  public editDisplayName(
+    memberId: string,
+    requestingParticipantId: string,
+    alias: string,
+    suffix?: string
+  ) {
     if (!this.locusUrl) {
       return Promise.reject(
         new ParameterError(
@@ -1205,7 +1211,8 @@ export default class Members extends StatelessWebexPlugin {
       memberId,
       requestingParticipantId,
       alias,
-      locusUrl
+      locusUrl,
+      suffix
     );
 
     return this.membersRequest.editDisplayNameMember(options);
--- a/packages/@webex/plugin-meetings/src/members/index.ts
+++ b/packages/@webex/plugin-meetings/src/members/index.ts
@@ -1181,7 +1181,7 @@ export default class Members extends StatelessWebexPlugin {
    * @param {string} memberId - id of the participant who is receiving request
    * @param {string} requestingParticipantId - id of the participant who is sending request (optional)
    * @param {string} [alias] - alias name
-   * @param {string} [suffix] - suffix name (optional)
+   * @param {string} [suffix] - name suffix (optional)
    * @returns {Promise}
    * @public
    * @memberof Members
--- a/packages/@webex/plugin-meetings/src/members/util.ts
+++ b/packages/@webex/plugin-meetings/src/members/util.ts
@@ -316,8 +316,11 @@ const MembersUtil = {
     } = {
       aliasValue: options.alias,
       requestingParticipantId: options.requestingParticipantId,
-      suffixValue: options.suffix || '',
     };
+
+    if (options.suffix !== undefined) {
+      body.suffixValue = options.suffix;
+    }
     const uri = `${options.locusUrl}/${PARTICIPANT}/${options.memberId}/${ALIAS}`;
 
     return {
--- a/packages/@webex/plugin-meetings/src/members/util.ts
+++ b/packages/@webex/plugin-meetings/src/members/util.ts
@@ -190,13 +190,21 @@ const MembersUtil = {
    * @param {String} requestingParticipantId id of the participant who is sending request (optional)
    * @param {String} alias alias name
    * @param {String} locusUrl url
+   * @param {String} suffix optional suffix
    * @returns {Object} consists of {memberID: string, requestingParticipantId: string, alias: string, locusUrl: string}
    */
-  generateEditDisplayNameMemberOptions: (memberId, requestingParticipantId, alias, locusUrl) => ({
+  generateEditDisplayNameMemberOptions: (
     memberId,
     requestingParticipantId,
     alias,
     locusUrl,
+    suffix
+  ) => ({
+    memberId,
+    requestingParticipantId,
+    alias,
+    locusUrl,
+    suffix,
   }),
 
   getMuteMemberRequestParams: (options) => {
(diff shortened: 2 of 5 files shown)


packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -1181,11 +1181,17 @@ export default class Members extends StatelessWebexPlugin {
    * @param {string} memberId - id of the participant who is receiving request
    * @param {string} requestingParticipantId - id of the participant who is sending request (optional)
    * @param {string} [alias] - alias name
+   * @param {string} [suffix] - suffix name (optional)
```
martroos: nitpick: Maybe "name suffix" makes more sense here (it is the suffix for a name, not the name of a suffix)?

packages/@webex/plugin-meetings/test/unit/spec/members/index.js
```diff
@@ -1017,6 +1020,31 @@ describe('plugin-meetings', () => {
       });
 
       it('should make the correct request when called with respective parameters', async () => {
+        const requestingParticipantId = uuid.v4();
+        const memberId = uuid.v4();
+        const alias = 'aliasName';
+        const suffix = 'suffixName';
```
martroos, line 1026: nitpick: It isn't a suffix name, it's just a suffix.

packages/@webex/plugin-meetings/src/members/util.ts
```diff
@@ -301,9 +309,14 @@ const MembersUtil = {
...
   editDisplayNameMemberRequestParams: (options) => {
-    const body = {
+    const body: {
+      aliasValue: string;
+      requestingParticipantId: string;
+      suffixValue?: string;
+    } = {
       aliasValue: options.alias,
       requestingParticipantId: options.requestingParticipantId,
+      suffixValue: options.suffix || '',
```
martroos: issue: There is a difference between setting to an empty string, and not setting it at all. Not setting it at all (i.e. the value is `undefined`) means "don't change the suffix", however in this implementation it means "clear the suffix". This issue can already be seen after using your changes in web client:

1. Start a webinar with native client
2. Join the webinar with web client
3. With native client, create a stage and assign subtitles to important participants
4. Make web client a cohost
5. Edit display names in web client
6. Notice that the subtitles for edited names has now been reset

The end result of this is that any caller of `editDisplayName` now needs to provide the existing suffix in order to preserve it. Instead, what we should do here is simply add it to the body if it is defined:

```ts
const body: {
  aliasValue: string;
  requestingParticipantId: string;
  suffixValue?: string;
} = {
  aliasValue: options.alias,
  requestingParticipantId: options.requestingParticipantId,
};

if (options.suffix !== undefined) {
  body.suffixValue = options.suffix;
}
```

This way when only editing the alias, the suffix is untouched, however it can be set as an empty string still (which is why I suggested the comparison to `undefined`).

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
```diff
@@ -129,7 +129,32 @@ describe('plugin-meetings', () => {
       });
     });
     describe('#generateEditDisplayNameMemberOptions', () => {
-      it('returns the correct options', () => {
+      it('returns the correct options with suffixx when suffix is there', () => {
```
martroos: nitpick: Suffix only has 1 "x" in it! Same comment for the other test in this `describe`.

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
```diff
@@ -306,7 +332,7 @@ describe('plugin-meetings', () => {
         const options = {
           invitee: {
             phoneNumber: '1234567890',
-            isInternalNumber: false
+            isInternalNumber: false,
```
martroos, line 335: question: Did your editor auto-format things?

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
```diff
@@ -452,5 +472,60 @@ describe('plugin-meetings', () => {
...
+            suffixValue: suffix,
+          },
+        });
+      });
+
+      it('returns the correct params when suffix is not available', () => {
+        const locusUrl = 'TestLocusUrl';
+        const memberId = 'test1';
+        const alias = 'alias';
+        const requestingParticipantId = '23131';
```
martroos: suggestion: Would be nice to use different values for each test

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
```diff
@@ -452,5 +472,60 @@ describe('plugin-meetings', () => {
...
+        const suffix = 'suffix';
+        const options = {
+          locusUrl: locusUrl,
+          memberId,
+          alias,
+          requestingParticipantId,
+          suffix,
+        };
+
+        const uri = `${options.locusUrl}/${PARTICIPANT}/${options.memberId}/${ALIAS}`;
```
martroos: nitpick: This is identical to the implementation - I think it would be better if this was formed as much locally as possible, for example:

```js
const uri = `${locusUrl}/participant/${memberId}/alias`;
```

This at least checks the constants used are correct.

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
```diff
@@ -452,5 +472,60 @@ describe('plugin-meetings', () => {
...
+          uri,
+          body: {
+            aliasValue: alias,
+            requestingParticipantId,
+            suffixValue: suffix,
+          },
+        });
+      });
+
+      it('returns the correct params when suffix is not available', () => {
```
martroos, line 504: issue: There should be a test where the suffix is an empty string, as well as this one (where it is undefined).

packages/@webex/plugin-meetings/test/unit/spec/members/request.js
```diff
@@ -426,6 +426,35 @@ describe('plugin-meetings', () => {
...
+
+        const options = {
+          memberId,
+          requestingParticipantId,
+          alias: aliasValue,
+          locusUrl,
+          suffix: suffixValue,
+        };
+
+        await membersRequest.editDisplayNameMember(options);
```
martroos, line 475: nitpick: Would be nice to check the result of `editDisplayNameMember`, though none of the other tests do, so I won't insist on it!

packages/@webex/plugin-meetings/test/unit/spec/members/request.js
```diff
@@ -426,6 +426,35 @@ describe('plugin-meetings', () => {
           body: {
             aliasValue,
             requestingParticipantId,
+            suffixValue: '',
```
martroos, line 455: issue: This shouldn't be there at all when there is no suffix provided. An additional test for when the suffix is an empty string might be nice though.

martroos (changes requested):

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
shnaaz: yes

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
martroos: I actually meant would be nice to use different values for all the inputs (I selected multiple lines!), not just the last one.

packages/@webex/plugin-meetings/test/unit/spec/members/utils.js
martroos: You missed one - the "returns the correct options without suffix when suffix is not there" test

martroos (approved):

lismith2-cisco (approved):

PR #4497, opened 2025-09-24
feat(mercury): handle mercury shutdown

Shreyas281299:
# COMPLETES # https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6913

- The PR adds the support for the shutdown event from the mercury.

- When mercury sends the shutdown event, SDK connects to a new mercury connection using the same mercury url. When the new WS is connected the older one is closed.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Unit Tests
- While on a call triggered a shutdown message on the mercury connection and ensure a new connection was setted up without any interruption
- While in a meeting triggered a shutdown message on the mercury connection and ensure a new connection was setted up without any interruption
- Locally linked SDK with WWC and tested a call and a meeting and triggered a shutdown event.

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


Shreyas281299:
@coderabbit review

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+            pingInterval: this.config.pingInterval,
+            pongTimeout: this.config.pongTimeout,
+            token: token.toString(),
+            trackingId: `${this.webex.sessionId}_${Date.now()}`,
+            logger: this.logger,
+          };
+
+          if (this.webex.config.defaultMercuryOptions) {
+            this.logger.info(`${this.namespace}: setting custom options for switchover`);
+            options = {...options, ...this.webex.config.defaultMercuryOptions};
```
rarajes2: Order of spreading the options seems incorrect. The default options should come first so that the later can override it.

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+  _handleImminentShutdown() {
+    try {
+      if (this._shutdownSwitchoverInProgress) {
+        this.logger.info(`${this.namespace}: [shutdown] switchover already in progress`);
+
+        return;
+      }
+      this._shutdownSwitchoverInProgress = true;
+      this._shutdownSwitchoverId = `${Date.now()}`;
+      this.logger.info(
```
rarajes2, line 127: `this.logger.log`

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+            this.logger.info(`${this.namespace}: setting custom options for switchover`);
+            options = {...options, ...this.webex.config.defaultMercuryOptions};
+          }
+
+          this.logger.info(`${this.namespace}: [shutdown] switchover url: ${webSocketUrl}`);
+
+          return pendingSocket.open(webSocketUrl, options);
+        })
+        .then(() => {
+          this.logger.info(
```
rarajes2: Same here. We were keeping info only at the beginning of the function

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+
+        return;
+      }
+      this._shutdownSwitchoverInProgress = true;
+      this._shutdownSwitchoverId = `${Date.now()}`;
+      this.logger.info(
+        `${this.namespace}: [shutdown] switchover start, id=${this._shutdownSwitchoverId}`
+      );
+
+      const pendingSocket = new Socket();
```
rarajes2: Let's rename it `newSocket`

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+      // Wire listeners prior to open
+      pendingSocket.on('close', (event) => this._onclose(event, pendingSocket));
+      pendingSocket.on('message', (...args) => this._onmessage(...args));
+      pendingSocket.on('pong', (...args) => this._setTimeOffset(...args));
+      pendingSocket.on('sequence-mismatch', (...args) => this._emit('sequence-mismatch', ...args));
+      pendingSocket.on('ping-pong-latency', (...args) => this._emit('ping-pong-latency', ...args));
+
+      Promise.all([this._prepareUrl(), this.webex.credentials.getUserToken()])
+        .then(([webSocketUrl, token]) => {
+          attemptWSUrl = webSocketUrl;
```
rarajes2: `attemptWSUrl` -> `newWSUrl`

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+          this.socket = pendingSocket;
+          this.connected = true; // remain connected throughout
+          this._shutdownSwitchoverInProgress = false;
+
+          // Optional: emit an event for observers
+          try {
+            this._emit('event:mercury_shutdown_switchover_complete', {url: attemptWSUrl});
+          } catch (e) {
+            // ignore observer errors
+          }
```
rarajes2: This does not seem to need a try-catch

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+            );
+          }
+        })
+        .catch((reason) => {
+          this.logger.info(
+            `${this.namespace}: [shutdown] switchover connection attempt failed`,
+            reason
+          );
+          this._shutdownSwitchoverInProgress = false;
+        });
```
rarajes2, line 165: Should we emit an event to notify the subscriber about the failure ?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+    } catch (e) {
+      this.logger.error(`${this.namespace}: [shutdown] error during switchover`, e);
+      this._shutdownSwitchoverInProgress = false;
+    }
```
rarajes2, line 170: Same here

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -211,7 +300,7 @@ const Mercury = WebexPlugin.extend({
     const socket = new Socket();
     let attemptWSUrl;
 
-    socket.on('close', (...args) => this._onclose(...args));
+    socket.on('close', (event) => this._onclose(event, socket));
```
rarajes2: Do we need this change ? Spread operator should be taking care of all the params already

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+
+      const pendingSocket = new Socket();
+      let attemptWSUrl;
+
+      // Wire listeners prior to open
+      pendingSocket.on('close', (event) => this._onclose(event, pendingSocket));
+      pendingSocket.on('message', (...args) => this._onmessage(...args));
+      pendingSocket.on('pong', (...args) => this._setTimeOffset(...args));
+      pendingSocket.on('sequence-mismatch', (...args) => this._emit('sequence-mismatch', ...args));
+      pendingSocket.on('ping-pong-latency', (...args) => this._emit('ping-pong-latency', ...args));
```
rarajes2: This seems like repetitive code. We can create a method which accepts the socket param and adds these listeners and call that method in other paces too

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -70,6 +70,95 @@ const Mercury = WebexPlugin.extend({
...
+          } catch (e) {
+            // ignore observer errors
+          }
+
+          // Do not force-close oldSocket; server will close it with 4001.
+          if (oldSocket) {
+            this.logger.info(
+              `${this.namespace}: [shutdown] old socket retained; awaiting 4001 close`
+            );
+          }
```
rarajes2: Q. Is `4001` a status code? We can improve the message

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -444,64 +533,99 @@ const Mercury = WebexPlugin.extend({
     return handlers;
   },
 
-  _onclose(event) {
+  _onclose(event, sourceSocket) {
     // I don't see any way to avoid the complexity or statement count in here.
     /* eslint complexity: [0] */
 
     try {
+      const isActiveSocket = !sourceSocket || (this.socket && sourceSocket === this.socket);
```
rarajes2: The `sourceSocket` passed is a new socket created (line number 300) which will never be undefined so the first condition will always return false and proceed with the next condition. Please correct if I am wrong

rarajes2: I think we are already updating the `this.socket` with the new one or active one. Shouldn't we be able to call the close on it directly ?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -516,6 +640,20 @@ const Mercury = WebexPlugin.extend({
...
 
+    // Handle shutdown message shape: { type: 'shutdown' }
+    if (envelope && envelope.type === 'shutdown') {
+      this.logger.info(`${this.namespace}: [shutdown] imminent shutdown message received`);
+      // Optional event for observers/tests
+      try {
+        this._emit('event:mercury_shutdown_imminent', envelope);
+      } catch (e) {
+        // ignore observer errors
+      }
```
rarajes2: Same here. Try-catch not needed

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
```diff
@@ -290,16 +290,22 @@ export default class Socket extends EventEmitter {
...
-        this.logger.debug(
-          `socket,${this._domain}: sequence number mismatch indicates lost mercury message. expected: ${this.expectedSequenceNumber}, actual: ${sequenceNumber}`
-        );
-        this.emit('sequence-mismatch', sequenceNumber, this.expectedSequenceNumber);
+      // Gracefully handle frames that omit sequenceNumber (e.g., shutdown frames)
+      if (
+        typeof data.sequenceNumber === 'number' ||
+        (typeof data.sequenceNumber === 'string' && data.sequenceNumber.trim() !== '')
+      ) {
+        const sequenceNumber = parseInt(data.sequenceNumber, 10);
```
rarajes2: Are we getting sequenceNumber in different number system ? We are already checking the type. I think the parsing should have been done before the condition

rarajes2 (changes requested):
1. Are we not adding/updating UTs ?
2. Please add vidcast

packages/@webex/internal-plugin-mercury/src/mercury.js
Kesari3008: What would be the source socket here ? I am unable to understand this, we are assigning negation of source socket but the we are equating source socket with this.socket  which can be either old socket or new based on handleImminentShutdown logic so can you elaobaret on what are we trying to achieve here

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -444,64 +533,99 @@ const Mercury = WebexPlugin.extend({
...
         case 4000:
           // metric: disconnect
           this.logger.info(`${this.namespace}: socket replaced; will not reconnect`);
-          this._emit('offline.replaced', event);
+          if (isActiveSocket) this._emit('offline.replaced', event);
+          // If not active, nothing to do
+          break;
+        case 4001:
+          // replaced during shutdown
+          if (isActiveSocket) {
```
Kesari3008, line 682: I thought that we are not going to close any socket until new socket is established and only when the new one is ready, server will close the old socket. And both these connections are supposed to be different so why are we reconnecting here ?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -444,64 +533,99 @@ const Mercury = WebexPlugin.extend({
...
+          }
           break;
         case 1001:
         case 1005:
         case 1006:
         case 1011:
           this.logger.info(`${this.namespace}: socket disconnected; reconnecting`);
-          this._emit('offline.transient', event);
-          this._reconnect(socketUrl);
+          if (isActiveSocket) {
```
Kesari3008, line 703: Why did we need to add this condition in all the event emissions?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -516,6 +640,20 @@ const Mercury = WebexPlugin.extend({
...
+    // Handle shutdown message shape: { type: 'shutdown' }
+    if (envelope && envelope.type === 'shutdown') {
+      this.logger.info(`${this.namespace}: [shutdown] imminent shutdown message received`);
+      // Optional event for observers/tests
+      try {
+        this._emit('event:mercury_shutdown_imminent', envelope);
+      } catch (e) {
+        // ignore observer errors
+      }
+      this._handleImminentShutdown();
```
Kesari3008, line 754: There is an error handling implemented in this method. If we do enter the catch block for this method, we shouldn't be resolving the promise in all cases below right. Why don't we return whatever is received from _handleImminentShutdown

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -516,6 +640,20 @@ const Mercury = WebexPlugin.extend({
       this.logger.debug(`${this.namespace}: message envelope: `, envelope);
     }
 
+    // Handle shutdown message shape: { type: 'shutdown' }
+    if (envelope && envelope.type === 'shutdown') {
+      this.logger.info(`${this.namespace}: [shutdown] imminent shutdown message received`);
+      // Optional event for observers/tests
+      try {
+        this._emit('event:mercury_shutdown_imminent', envelope);
```
Kesari3008: Why do we need to add this single line under try block ? Does the emission of event fail  and what is this scenario where this can fail ?

Kesari3008 (changes requested):
Did we test calling flows when the shutdown occurs to see if we have any impact there due to connection replacement ?

Kesari3008:
Please update the description with details of testing done for this change and check the boxes required

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: In this file the deafult has been logger.info for the existint code. I feel we can keep this as info only.

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Made a new method for adding the event listeners

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Updated

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Thanks for catching this

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Yes, 4001 is 'replacement of the connection due to shutdown'. Updated the message

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: We can emit a failure, added the event

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: socket is not part of the event data, so we need this

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Ive removed it

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Removed

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: This is the handler for 'message' event on the web socket, this return value is not used. So we are just resolving the promise. 

Also in `_handleImminentShutdown` method we are sending an event to let the user know that there is an error while connecting, so the handling can be left to the application

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: So in this condition all we need to know is that 
- If the closing socket is the current socket (this is a normal closing of mercury connection)
- If the closing socket is an older inactive socket (this is the case of mercury shutdown)

Simplified the condition

Shreyas281299: The negation was kept for backward compatibility, but since this is an internal plugin I can make the changes.

Ive mentioned what we are trying to gather from this condition in the above comment

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: You are correct. When we recieve 4001 is a final code

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: If we wont put this condition any event received while we have two WS, our code wont know for which socket thie event came for and we might end up closing the new socket but infact we should have closed the older socket


AI explanation : 
- Prevents false state changes - Old socket closures don't represent actual disconnection
- Maintains accurate state - Application only knows about the active connection
- Avoids duplicate actions - Prevents reconnecting when already connected
- Enables seamless switchover - Old sockets can close silently

Kesari3008 (commented):
1. Unit test is failing mercury plugin
2. Did we test calling flows with mercury shutdown ?

mkesavan13 (commented):
Did we test this by locally linking the web client once? We need one of WWC member's review as well on this one.

Kesari3008 (approved):

packages/@webex/internal-plugin-mercury/src/mercury.js
rarajes2: I think I got confused with the name `activeSocket`. I thought it means the socket is active but here it means that it is the currently being used socket.

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -444,64 +533,108 @@ const Mercury = WebexPlugin.extend({
...
       const reason = event.reason && event.reason.toLowerCase();
-      const socketUrl = this.socket.url;
 
-      this.socket.removeAllListeners();
-      this.unset('socket');
-      this.connected = false;
-      this._emit('offline', event);
-      this.webex.internal.newMetrics.callDiagnosticMetrics.setMercuryConnectedStatus(false);
+      let socketUrl;
+      if (isActiveSocket && this.socket) {
```
rarajes2, line 638: `isActiveSocket` is already ensuring that the `this.socket` has some value so we don't need to add both the checks.

rarajes2 (approved):

Shreyas281299:
> Did we test this by locally linking the web client once? We need one of WWC member's review as well on this one.

Yes, Ive locally linked and tested a call and a meeting in the WWC

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -96,6 +96,99 @@ const Mercury = WebexPlugin.extend({
...
+      let newWSUrl;
+
+      // Wire listeners prior to open
+      this._attachSocketEventListeners(newSocket);
+
+      Promise.all([this._prepareUrl(), this.webex.credentials.getUserToken()])
+        .then(([webSocketUrl, token]) => {
+          newWSUrl = webSocketUrl;
+
+          let options = {
```
Coread: these options are a complete duplicate of the code in the open method. This could have been factored out to a private method. In fact, most of what is in this .then could have been factored out. It only really differs by the log strings

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -96,6 +96,99 @@ const Mercury = WebexPlugin.extend({
...
+          };
+
+          if (this.webex.config.defaultMercuryOptions) {
+            this.logger.info(`${this.namespace}: setting custom options for switchover`);
+            options = {...this.webex.config.defaultMercuryOptions, ...options};
+          }
+
+          this.logger.info(`${this.namespace}: [shutdown] switchover url: ${webSocketUrl}`);
+
+          return newSocket.open(webSocketUrl, options);
```
Coread: if the socket fails to open, I think it won't ever have the event listeners removed. You should call removeAllListeners on the socket in the catch

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -470,64 +559,108 @@ const Mercury = WebexPlugin.extend({
...
+        socketUrl = this.socket.url;
+      } else if (sourceSocket) {
+        // Old socket closed - get URL from the closed socket
+        socketUrl = sourceSocket.url;
+      }
+
+      if (isActiveSocket) {
+        // Only tear down state if the currently active socket closed
+        if (this.socket) {
+          this.socket.removeAllListeners();
```
Coread, line 649: question: When are the listeners removed from the old socket?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -96,6 +96,99 @@ const Mercury = WebexPlugin.extend({
...
+          this.logger.info(`${this.namespace}: [shutdown] switchover url: ${webSocketUrl}`);
+
+          return newSocket.open(webSocketUrl, options);
+        })
+        .then(() => {
+          this.logger.info(`${this.namespace}: [shutdown] switchover connected, url: ${newWSUrl}`);
+
+          const oldSocket = this.socket;
+          // Atomically switch active socket reference without closing the old one.
+          this.socket = newSocket;
```
Coread: There's a potential race condition here. If the old mercury websocket disconnects before it reaches this line, then the plugin will still emit the offline. This code won't emit the online either so the consuming application would not realise that it has reconnected

packages/@webex/internal-plugin-mercury/src/mercury.js
Coread: This directly connects the websocket where as the first connection uses the backoff strategy. Is this deliberate? Are we not worried that this reconnection attempt is more fragile than the normal connection?

Shreyas281299: Thanks for pointing this out, Ive refactored the handleIminentShutdown method to reuse the backoff strategy

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: @Coread  the idea is that the old mercury connection should not close immediately after the shutdown event there is a time gap between when the event is received and when the actually socket is closed.

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Thanks for catching this, the old socket's eventListeners wernt getting closed. Ive added the logic to remove the eventListeners in the else block of this.

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Update the duplicate code to a new method _prepareAndOpenSocket

packages/@webex/internal-plugin-mercury/src/mercury.js
Shreyas281299: Right, Ive updated the code to only add event listerns when the socket is up- following the existing implementation

Coread (approved):

PR #4513, opened 2025-10-01
feat(contact-center): add outdial ANI entries support for outbound calls

cmullenx:
# COMPLETES # [CAI-7180](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7180) 

- Adds ability to get the list of outdial ani eps
- Adds ability to use selected outdial ani in the outbound call 

- Adds new function to retrieve list of outdial ani eps
- Adds new param to outbound call functionality
- Updates Samples to show functionality
- Adds and updates tests

Vidcast: https://app.vidcast.io/share/d2f13b5b-bf9d-4fa2-b0e1-cdd0ca2c4b73

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Added UTs and tested manually 

  - [ ] Github Copilot
  - [x] Other - Windsurf
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/contact-center/src/cc.ts
```diff
@@ -1489,6 +1493,110 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+   * );
+   *
+   * // Process the results
+   * const aniList = paginatedAni.data || [];
+   * aniList.forEach(ani => {
+   *   console.log(`ANI: ${ani.number} - ${ani.name}`);
+   * });
+   * ```
+   */
+  public async getOutdialAniEntries(
```
arun3528: can we convert this into an object {} , reduces erros as there is no type for this

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1489,6 +1493,110 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        orgId,
+        outdialANI,
+        page,
+        pageSize,
+        search,
+        filter,
+        attributes
+      );
+
+      LoggerProxy.log(`Successfully retrieved outdial ANI entries for ANI ID ${outdialANI}`, {
```
arun3528, line 1585: metric for success and failure would needed

cmullenx: added metrics

packages/@webex/contact-center/src/cc.ts
cmullenx: done

packages/@webex/contact-center/src/metrics/constants.ts
```diff
@@ -62,6 +64,9 @@ type Enum<T extends Record<string, unknown>> = T[keyof T];
  * @property {string} AGENT_DEVICE_TYPE_UPDATE_SUCCESS - Event name for successful agent device type update.
  * @property {string} AGENT_DEVICE_TYPE_UPDATE_FAILED - Event name for failed agent device type update.
  *
+ * @property {string} OUTDIAL_ANI_EP_FETCH_SUCCESS - Event name for successful outdial ANI entries fetch.
```
Kesari3008, line 71: Looks like these 2 lines added twice, please remove one of them

packages/@webex/contact-center/src/services/config/constants.ts
```diff
@@ -67,6 +67,7 @@ export const METHODS = {
   GET_TENANT_DATA: 'getTenantData',
   GET_URL_MAPPING: 'getURLMapping',
   GET_DIAL_PLAN_DATA: 'getDialPlanData',
+  GET_OUTDIAL_ANI_ENTRIES: 'getOutdialAniEntries',
```
Kesari3008: Same constant is added in this file as well: packages/@webex/contact-center/src/constants.ts 
Could we please keep it only in one place and import it from there wherever required

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1489,6 +1499,133 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        search,
+        filter,
+        attributes,
+      });
+
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.OUTDIAL_ANI_EP_FETCH_SUCCESS,
+        {
+          outdialANI,
+          page,
```
Kesari3008: We don't have to submit page, pageSize, search, filter and attributes in the metrics. Please remove these

packages/@webex/contact-center/src/services/config/index.ts
```diff
@@ -686,4 +686,61 @@ export default class AgentConfigService {
...
+      const resource = `organization/${orgId}/v2/outdial-ani/${outdialANI}/entry${
+        queryString ? `?${queryString}` : ''
+      }`;
+      const response = await this.webexReq.request({
+        service: WCC_API_GATEWAY,
+        resource,
+        method: HTTP_METHODS.GET,
+      });
+
+      if (response.statusCode !== 200) {
```
Kesari3008: Could you please check if for cases where we get error response, does the code go to catch block, if it does then we won't need this check. Catch block can handle API error responses as well

packages/@webex/contact-center/src/services/config/index.ts
```diff
@@ -686,4 +686,61 @@ export default class AgentConfigService {
...
   // getQueues removed - use Queue instead
+
+  /**
+   * Fetches outdial ANI (Automatic Number Identification) entries for the given orgId and outdial ANI ID.
+   * @ignore
+   * @param {string} orgId - organization ID for which the outdial ANI entries are to be fetched.
+   * @param {OutdialAniParams} params - parameters object containing outdialANI and optional pagination/filtering options
+   * @returns {Promise<OutdialAniEntriesResponse>} - A promise that resolves to the outdial ANI entries response.
+   * @throws {Error} - Throws an error if the API call fails or if the response status is not 200.
+   * @public
```
Kesari3008: Since this is behind the services object, this is going to be private function from doc description perspective

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1489,6 +1499,133 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+   * outbound calls. The ANI data is associated with an outdial ANI ID and can be filtered
+   * and paginated as needed.
+   *
+   * @param {string} outdialANI - The outdial ANI ID to fetch ANI data for
+   * @param {number} [page] - Optional page number for pagination (0-based)
+   * @param {number} [pageSize] - Optional number of items per page
+   * @param {string} [search] - Optional search term to filter results by name or number
+   * @param {string} [filter] - Optional filter string
+   * @param {string} [attributes] - Optional attributes to include in response
+   * @returns {Promise<any>} Promise resolving to outdial ANI response containing:
```
Kesari3008: Return type here is not any, there is a type created so please mention that here

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -931,6 +931,8 @@ export type DialerPayload = {
   mediaType: 'telephony' | 'chat' | 'social' | 'email';
   /** The outbound type for the task */
   outboundType: 'OUTDIAL' | 'CALLBACK' | 'EXECUTE_FLOW';
+  /** The contact center number that will be used while making a call to the customer.  */
+  origin: string;
```
Kesari3008, line 1068: My understanding her is that this origin is nothing but OutdialANI that gets selected while making outdial call, can we make the description more accurate to reflect that clearly.

Kesari3008 (changes requested):

packages/@webex/contact-center/src/services/config/index.ts
```diff
@@ -9,8 +9,6 @@ import {
   DesktopProfileResponse,
   ListAuxCodesResponse,
   AgentResponse,
-  OrgInfo,
```
adhmenon, line 12: I think we shouldn't be removing these types, I believe we use them whilst getting the agent config.

cmullenx: thanks, added them back

packages/@webex/contact-center/src/services/config/index.ts
cmullenx: Yeah so WebexRequest.request() doesn't throw on HTTP errors so it wouldn't reach the catch block w/out this check. I used the same pattern from the other functions in this file because of this issue.

packages/@webex/contact-center/src/services/task/types.ts
cmullenx: sure, i have updated the documentation. i previously used the property documentation straight from the API docs. I don't feel its necessary to mention that this origin gets selected, because other consumers may not use it in a select so feel this is more clear.

packages/@webex/contact-center/src/metrics/constants.ts
cmullenx: thanks for catching this, removed one of them

packages/@webex/contact-center/src/services/config/constants.ts
cmullenx: sure, removed from packages/@webex/contact-center/src/constants.ts

packages/@webex/contact-center/src/cc.ts
cmullenx: removed

packages/@webex/contact-center/src/services/config/index.ts
cmullenx: updated

packages/@webex/contact-center/src/cc.ts
cmullenx: fixed

packages/@webex/contact-center/src/services/config/index.ts
adhmenon: Can we verify this, I understand it might have been used in other places but iirc we have an interceptor which catches any non 2xx response and throws an error.
In that case we might not need this flow.

packages/@webex/contact-center/src/services/config/constants.ts
Kesari3008: I still see it in both places

packages/@webex/contact-center/src/constants.ts
```diff
@@ -42,6 +42,7 @@ export const METHODS = {
   HANDLE_DEVICE_TYPE: 'handleDeviceType',
   START_OUTDIAL: 'startOutdial',
   GET_QUEUES: 'getQueues',
+  GET_OUTDIAL_ANI_ENTRIES: 'getOutdialAniEntries',
```
Kesari3008, line 45: Still present

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1447,6 +1458,7 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       // Construct the outdial payload.
       const outDialPayload: DialerPayload = {
         destination,
+        origin,
         entryPointId: this.agentConfig.outDialEp,
```
Kesari3008, line 1458: As per the discussion, we were planning to update this too right so that user can pass any outdial EP ID as well ?

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1490,6 +1504,123 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      );
+
+      LoggerProxy.log(`Successfully retrieved outdial ANI entries for ANI ID ${outdialANI}`, {
+        module: CC_FILE,
+        method: CONFIG_METHODS.GET_OUTDIAL_ANI_ENTRIES,
+      });
+
+      return result;
+    } catch (error) {
+      const failure = error.details as Failure;
```
Kesari3008, line 1592: For error formatting I believe we have this function being used : getErrorDetails() and after formatting we throw new formatted error object. Could you please check if same design is followed here

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1490,6 +1504,123 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    if (!orgId) {
+      LoggerProxy.error('Org ID not found.', {
+        module: CC_FILE,
+        method: CONFIG_METHODS.GET_OUTDIAL_ANI_ENTRIES,
+      });
+
+      throw new Error('Org ID not found.');
+    }
+
+    if (!this.agentConfig) {
```
Kesari3008: Why did we need to have this check here ? I do not see agentConfig being consumed for this method ? Also cc object is only available after initialization of the SDK and by then the websocket connection should have been established and agent config should be present. In any case outdialANI seems to received from the user and not being pulled within SDK so we may not need this check.

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1490,6 +1504,123 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+        {
+          outdialANI,
+          resultCount: result?.length || 0,
+        },
+        ['behavioral', 'business', 'operational']
+      );
+
+      LoggerProxy.log(`Successfully retrieved outdial ANI entries for ANI ID ${outdialANI}`, {
+        module: CC_FILE,
+        method: CONFIG_METHODS.GET_OUTDIAL_ANI_ENTRIES,
```
Kesari3008: Can we add trackingId and OutdialANI in this log ?

packages/@webex/contact-center/src/cc.ts
```diff
@@ -1490,6 +1504,123 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+      });
+
+      return result;
+    } catch (error) {
+      const failure = error.details as Failure;
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.OUTDIAL_ANI_EP_FETCH_FAILED,
+        {
+          ...MetricsManager.getCommonTrackingFieldForAQMResponseFailed(failure),
```
Kesari3008, line 1596: We should submit the error object in the metric

Kesari3008 (changes requested):

packages/@webex/contact-center/src/cc.ts
cmullenx: agent desktop is using the outDialEp from agent config and simply passing the selected outdial ani phone number as the origin, so i kept the same pattern.

packages/@webex/contact-center/src/cc.ts
cmullenx: removed

packages/@webex/contact-center/src/cc.ts
cmullenx: updated to throw the formatted error object

packages/@webex/contact-center/src/constants.ts
cmullenx: removed from config/constants and left here

packages/@webex/contact-center/src/cc.ts
cmullenx: updated

packages/@webex/contact-center/src/cc.ts
cmullenx: ive added the tracking id and outdial ani to the error log, but there is not a tracking id on the response object for successes and the outdial ANI is present in the log already.

packages/@webex/contact-center/src/services/config/index.ts
cmullenx: yes, tested again and error is caught, have removed.

adhmenon (approved):
Changes are looking alright to me.

packages/@webex/contact-center/src/services/config/index.ts
```diff
@@ -686,4 +688,56 @@ export default class AgentConfigService {
...
+        resource,
+        method: HTTP_METHODS.GET,
+      });
+
+      LoggerProxy.log('getOutdialAniEntries API success.', {
+        module: CONFIG_FILE_NAME,
+        method: MAIN_METHODS.GET_OUTDIAL_ANI_ENTRIES,
+      });
+
+      return response.body?.data;
```
Kesari3008, line 734: Check if this response object has tracking ID as one of the headers present and if it can be read.

Kesari3008 (approved):

packages/@webex/contact-center/src/services/config/index.ts
cmullenx: The response object only has ```{content-type: 'application/json'}``` as a header

PR #4518, opened 2025-10-06
fix(calling): separate handling for mercury down and internet down cases

Kesari3008:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6518 

Handling of state calls and registration status when either internet goes down or mercury service is down. We have added separate logic for each case to smoothly handle it.

1. Listening for window 'online, 'offline' event listeners and switched the appriach for mercury as well by listenting for online and offline events instead of checking mercury.connected state every 2 seconds.
2. Added new callbacks to be invoked when these events are received.
3. For mercury down case, we will only check for calls status by sending call keepalive to Mobius once mercury comes back up and if the calls are not active anymore, we will do a cleanup.
4. For internet down, we are running an interval to check every 5 seconds if there are any calls still active so that we don't go directly and restore the registration. Restoration will only be performed when there are 0 calls active. And we send call keepalives to check for active calls and if there are stale calls do a cleanup.
5. Added new method to submit metrics for network flap and mercury flap.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

Testcase 1: Register the device, answer a call or dial out. Once the call is connected, bring the internet down. Wait for few seconds, disconnect the call from customer's end. Call will still be active on SDK. Once the network comes back up, we check the call status and keepalive fails since the call doesn't exist on Mobius anymore. SDK will perform a call cleanup and then restore the registration as well.

Testcase 2: Register the device, answer a call or dial out. Once the call is connected, bring the internet down. Wait for few seconds and bring the internet back up. We check the call status and keepalive succeeds hence no restoration will be done. Manually disconnect the call and SDK will then restore the registration.

Testcase 3: Only bring the mercury down and it only does call keepalive checks. Registration is not touched as reg keepalives continue and registration remains stable.

LOGS FOR EACH OF THE ABIVE CASES CAN BE FOUND ATTACHED IN THE JIRA

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +199,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      const controller = new AbortController();
+      const timeoutId = setTimeout(() => controller.abort(), 3000);
+
+      // Using a common connectivity check endpoint that returns 204 with minimal payload.
+      // no-cors mode yields an opaque response but a successful fetch implies reachability.
+      await fetch('https://www.google.com/generate_204', {
+        method: 'GET',
+        cache: 'no-cache',
+        mode: 'no-cors',
+        signal: controller.signal,
```
Shreyas281299, line 253: Screenshot 2025-10-08 at 12 50 51 PM

The response was 204 but I got this NS_Binding_Aborted, I hope this wont cause this logic to fail.

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +199,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+        .catch((err) => {
+          log.warn(`Call Keepalive failed: ${err}`, {
+            file: CALLING_CLIENT_FILE,
+            method: METHODS.NETWORK_ONLINE,
+          });
+
+          callObj.sendCallStateMachineEvt({type: 'E_SEND_CALL_DISCONNECT'});
+        });
+    }
+  }
```
Shreyas281299: We can optimize this method. 
Right now we will take each call and then wait for status response and then move to the next call. 

But since we only have 2 calls max, maybe optimizing this would be an overkill. Il leave it upto you

packages/calling/src/Metrics/types.ts
```diff
@@ -47,6 +48,12 @@ export type SERVER_TYPE = 'PRIMARY' | 'BACKUP' | 'UNKNOWN';
 
 export const UPLOAD_LOGS_ACTION = 'upload_logs';
 
+export enum CONN_ACTION {
```
Shreyas281299: We can just name is as CONNECTION_ACTIONS. 

"Conn" has several meanings, but most commonly refers to the act of directing the steering of a ship or plane (verb), or the station or act of steering (noun).

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +199,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+  // Wondering if we should keep this for timestamp recording purpose
+  private handleNetworkOnline = async () => {
     log.info(METHOD_START_MESSAGE, {
       file: CALLING_CLIENT_FILE,
-      method: METHODS.DETECT_NETWORK_CHANGE,
+      method: METHODS.NETWORK_ONLINE,
     });
-    let retry = false;
+    this.networkUpTimestamp = new Date(Date.now()).toISOString();
+  };
```
Shreyas281299, line 314: Why do we need this?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +199,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      this.metricManager.submitConnectionMetrics(
+        METRIC_EVENT.CONNECTION_ERROR,
+        CONN_ACTION.NETWORK_FLAP,
+        METRIC_TYPE.BEHAVIORAL,
+        this.networkDownTimestamp,
+        this.networkUpTimestamp
+      );
+    } else {
+      if (Object.keys(this.callManager.getActiveCalls()).length) {
+        await this.isCallActive();
```
Shreyas281299: Im unable to understand why we are doing this here? Can we not try to put this at the top of the function?

Shreyas281299 (changes requested):
Please see the comments. Please check the Unit Tests

packages/calling/src/Metrics/types.ts
Kesari3008: Updated

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: This is just for recording the timestamp when network came back online

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: This is the block to be executed when internet is up and stable but mercury service is down and we only to check for stale calls and clean that up. No specific reason to put it here, just how the if condition was done. Since I am checking for networkDown to be true first so this had to be in the else block. If I reverse the condition, then this will go up.

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -96,6 +102,16 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
 
   private lineDict: Record<string, ILine> = {};
 
+  private networkDown = false;
```
rarajes2: nitpick: We can rename it `isNetworkDown`.

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
    */
-  private async detectNetworkChange() {
+  private async checkNetworkReachability(): Promise<boolean> {
+    try {
+      const controller = new AbortController();
+      const timeoutId = setTimeout(() => controller.abort(), 3000);
+
+      // Using a common connectivity check endpoint that returns 204 with minimal payload.
+      // no-cors mode yields an opaque response but a successful fetch implies reachability.
+      await fetch('https://www.google.com/generate_204', {
```
rarajes2, line 249: Can we confirm if this url is ok to use as client in some specific region might block this ?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      });
+
+      return false;
+    }
+  }
+
+  private async isCallActive() {
+    const calls = Object.keys(this.callManager.getActiveCalls());
+    for (const call of calls) {
+      const callObj = this.callManager.getActiveCalls()[call];
```
rarajes2: 1. We can use `Object.entries` and access key and value both, that way we won't have to call the `getActiveCalls()` in loop, it will be called once
2. Using async-await syntax will make code more readable
3. We are not returning anything from this method. Based on the method name it should return a boolean

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+          });
+          /*
+           * Media Renegotiation Possibility if call keepalive succeeds,
+           * for cases like WebRTC disconnect and media inactivity.
+           */
+        })
+        .catch((err) => {
+          log.warn(`Call Keepalive failed: ${err}`, {
+            file: CALLING_CLIENT_FILE,
+            method: METHODS.NETWORK_ONLINE,
```
rarajes2: This can be created as context object since being used twice

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+          });
+
+          callObj.sendCallStateMachineEvt({type: 'E_SEND_CALL_DISCONNECT'});
+        });
+    }
+  }
+
+  private handleNetworkOffline = async () => {
+    this.networkDownTimestamp = new Date(Date.now()).toISOString();
```
rarajes2: No need to pass `Date.now()`

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+    this.networkDown = !(await this.checkNetworkReachability());
+    log.warn(`Network has flapped, wait for it to come back up`, {
```
rarajes2: `Network has flapped` -> `Network is down`

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+
+  // Wondering if we should keep this for timestamp recording purpose
+  private handleNetworkOnline = async () => {
     log.info(METHOD_START_MESSAGE, {
       file: CALLING_CLIENT_FILE,
-      method: METHODS.DETECT_NETWORK_CHANGE,
+      method: METHODS.NETWORK_ONLINE,
     });
-    let retry = false;
+    this.networkUpTimestamp = new Date(Date.now()).toISOString();
```
rarajes2: Same here

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+  private handleNetworkOffline = async () => {
+    this.networkDownTimestamp = new Date(Date.now()).toISOString();
+    this.networkDown = !(await this.checkNetworkReachability());
+    log.warn(`Network has flapped, wait for it to come back up`, {
+      file: CALLING_CLIENT_FILE,
+      method: METHODS.NETWORK_OFFLINE,
+    });
+
+    if (this.networkDown) {
+      const line = Object.values(this.lineDict)[0];
```
rarajes2, line 302: This approach is error prone. `Object.values` does not guarantee the order of keys/values. Since we are using the first element from the array, it could be any of the values in the list

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-        log.warn(`Network has flapped, waiting for mercury connection to be up`, {
-          file: CALLING_CLIENT_FILE,
-          method: METHODS.DETECT_NETWORK_CHANGE,
-        });
+  private handleMercuryOffline = async () => {
+    log.warn(`Mercury down, waiting for connection to be up`, {
+      file: CALLING_CLIENT_FILE,
+      method: METHODS.MERCURY_OFFLINE,
+    });
+    this.mercuryDownTimestamp = new Date(Date.now()).toISOString();
```
rarajes2: Same here

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-      if (
-        !this.webex.internal.mercury.connected &&
-        !retry &&
-        !Object.keys(this.callManager.getActiveCalls()).length
-      ) {
-        log.warn(`Network has flapped, waiting for mercury connection to be up`, {
-          file: CALLING_CLIENT_FILE,
-          method: METHODS.DETECT_NETWORK_CHANGE,
-        });
+  private handleMercuryOffline = async () => {
```
rarajes2: This does not have to be an async method

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+    );
+  };
 
-        line.registration.clearKeepaliveTimer();
+  private handleMercuryOnline = async () => {
+    log.info(METHOD_START_MESSAGE, {
+      file: CALLING_CLIENT_FILE,
+      method: METHODS.MERCURY_ONLINE,
+    });
+    this.mercuryUpTimestamp = new Date(Date.now()).toISOString();
```
rarajes2: Same here

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+    if (this.networkDown) {
+      const callCheckInterval = setInterval(async () => {
+        if (!Object.keys(this.callManager.getActiveCalls()).length) {
+          clearInterval(callCheckInterval);
+          const line = Object.values(this.lineDict)[0];
```
rarajes2, line 341: Same here. This approach is error prone

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+  };
 
-        line.registration.clearKeepaliveTimer();
+  private handleMercuryOnline = async () => {
+    log.info(METHOD_START_MESSAGE, {
+      file: CALLING_CLIENT_FILE,
+      method: METHODS.MERCURY_ONLINE,
+    });
+    this.mercuryUpTimestamp = new Date(Date.now()).toISOString();
+    if (this.networkDown) {
```
rarajes2: If mercury comes online, shouldn't this already be true ?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -191,48 +205,165 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+            );
+          } else {
+            this.networkDown = false;
+          }
+        }
+      }, NETWORK_FLAP_TIMEOUT);
 
-        retry = true;
+      if (Object.keys(this.callManager.getActiveCalls()).length) {
+        await this.isCallActive();
```
rarajes2: We are not doing anything with this. I am not able to understand purpose of this call

packages/calling/src/Metrics/types.ts
```diff
@@ -47,6 +48,12 @@ export type SERVER_TYPE = 'PRIMARY' | 'BACKUP' | 'UNKNOWN';
 
 export const UPLOAD_LOGS_ACTION = 'upload_logs';
 
+export enum CONNECTION_ACTION {
+  NETWORK_FLAP = 'network_flap',
```
rarajes2, line 59: `network flap` does not explain clearly the state of network. We could have used `Network up` / `Network down` the same way we have for Mercury or vice versa to keep it consistent

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: This seems to be specific to Firefox browser. We will need to check it's working in the samples app running on the Firefox browser

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: This will only be true when network goes down and will remain true until we set it back to false after restoring registration. We are adding this condition to separate the case where only network goes down. This will also be false, if network is up but mercury is down, in that case we don't want to restore registration

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: We are sending the call keepalives using this method. If there are active calls, we will start sending keepalives and clear stale calls or if calls are active just log it and continue waiting for call to be manually disconnected

packages/calling/src/Metrics/types.ts
Kesari3008: We are only sending metric once after network has come up. So by then network is already up and there is no differentiation between the network down and network up. So I used network flap as event name to track that network went down and came back up and we submit the accurate timestamp for when it went offline and came back up

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Removed

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Updated the context

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Removed from all the places

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: We do not want to return any value from this method because all we want to ensure here is that we send call keepalives for calls that are present from SDK perspective and clear if there are any stale calls.
I used .then approach because we want to catch all failures and clear the call if it's not active anymore.

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: I have tested this in Chrome browser as well as Firefox browser, works in both.
@Shreyas281299 When you got NS_Binding_Aborted, did it set networkDown to true. It can be checked using calling.callingClient.networkDown in console. Based on that we can figure out if this logic is intact

Shreyas281299: Since we have tested then its okay, im resolving this

packages/calling/src/CallingClient/CallingClient.ts
Shreyas281299: Since we wont have alot of calls so this should not affect the performance.

Shreyas281299 (approved):
Looks good.

packages/calling/src/CallingClient/CallingClient.ts
Shreyas281299: Im no expert in this method but I searched and it does mention that it will gaurantee the sequence as long as we populate the dict in the same sequence everytime, also when we will add support for multiline, a user would have to choose which line to proceed with so this code will anyways be overwritten and we would need a lineId to select the line

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: Maybe we can rename the method `isCallActive` to something else.

rarajes2 (approved):

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Discussed this, doesn't require change for now

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: updated

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: addressed

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Discussed this, doesn't require change

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Addressed

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Updated now

PR #4531, opened 2025-10-13
fix(calling): fixed final error handling for keepalives

Kesari3008:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6661 

Error being thrown from webworker on keepalive failure was not in correct format due to which errorCode match within handleRegistrationErrors() failed and hence all the keepalive failures were going to default case: Unknown Error and hence keepalives kept getting sent even for final error cases.

Fixed the error format in webworker file.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Register the device one one browser and let the first keepalive be successful. Register with the same user in another window and next keepalive in existing window will fail with 404 and attempt device registration to restore the registration.

  - [ ] Github Copilot
  - [x] Other - Please Specify
- [x] This PR is related to
  - [x] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -824,12 +852,14 @@ export class Registration implements IRegistration {
                     METRIC_TYPE.BEHAVIORAL,
                     KEEPALIVE_UTIL,
                     serverType,
-                    error.headers?.trackingid ?? '',
+                    event.data.err.trackingId ?? '',
```
adhmenon: Are we sure that this value will be populated.

packages/calling/src/CallingClient/registration/webWorker.ts
```diff
@@ -46,11 +48,19 @@ const messageHandler = (event: MessageEvent) => {
             } as KeepaliveStatusMessage);
           }
           keepAliveRetryCount = 0;
-        } catch (err: unknown) {
+        } catch (err: any) {
```
adhmenon, line 49: Any way we can avoid using `any`?

adhmenon (commented):
Changes look fine, is it possible to show that this does not break agent desktop?
Maybe a vidcast.

adhmenon (approved):

Shreyas281299 (approved):
Code is good, I just see alot of type casting to `as any`, if we can minimize them that would be great

packages/calling/src/CallingClient/registration/webWorker.ts
Kesari3008: It can be made unknown. This error is of type Response which in webwork file we are unable to import

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: I have updated the code for trackingid to be read from header and in headers we are populating this value. If this value isn;t present then it will simply be an empy string

Kesari3008:
> Code is good, I just see alot of type casting to `as any`, if we can minimize them that would be great

Removed the any from the test file.

PR #4555, opened 2025-10-30
fix(calling): moves fetching of mobius clusters to service layer

Kesari3008:
# COMPLETES #SPARK-700838 

Accommodate fetching different services urls using V1Catalog as well as V2 catalog.

1. Fetching Mobius cluster information for V1Catalog and V2Catalog from services layer itself.
2. Removed /myIP request as it is not needed.
3. Updated other urls being fetched directly fro serviceUrl for V1 catalog to fetch entries from V2 catalog.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

Logs for v1Catalog and v2Catalog are uploaded in Jira, please check.

Screenshot 2025-11-06 at 3 59 15 PM

  - [ ] Github Copilot
  - [x] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
+++ b/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
@@ -14,6 +14,7 @@ import {
   Service,
   ServiceHostmap,
   ServiceGroup,
+  ServiceHost,
 } from './types';
 
 const trailingSlashes = /(?:^\/)|(?:\/$)/;
@@ -104,40 +105,34 @@ const Services = WebexPlugin.extend({
     return catalog.markFailedServiceUrl(url);
   },
 
-  /*
-   *  Get all Mobius cluster host entries from the v2 services list.
+  /**
+   * Get all Mobius cluster host entries from the v2 services list.
+   * @returns {Array<ServiceHost>} - An array of `ServiceHost` objects.
    */
-  getMobiusClusters(): Array<{
-    host: string;
-    ttl: number;
-    priority: number;
-    id: string;
-    homeCluster?: boolean;
-  }> {
-    const clusters: Array<{
-      host: string;
-      ttl: number;
-      priority: number;
-      id: string;
-      homeCluster?: boolean;
-    }> = [];
+  getMobiusClusters(): Array<ServiceHost> {
+    const clusters: Array<ServiceHost> = [];
     const services: Array<Service> = this._services || [];
 
-    services.forEach((service) => {
-      if (service?.serviceName === 'mobius' && Array.isArray(service.serviceUrls)) {
+    services
+      .filter(
+        (service) =>
+          service?.serviceName === 'mobius' &&
+          Array.isArray(service.serviceUrls) &&
+          service.serviceUrls.length > 0
+      )
+      .forEach((service) => {
         service.serviceUrls.forEach((serviceUrl) => {
-          if (!clusters.find((c) => c && c.host === serviceUrl.baseUrl)) {
-            serviceUrl.baseUrl = serviceUrl.baseUrl.replace('https://', '').replace('/api/v1', '');
+          const modifiedHost = serviceUrl.baseUrl.replace('https://', '').replace('/api/v1', '');
+          if (!clusters.find((c) => c && c.host === modifiedHost)) {
             clusters.push({
-              host: serviceUrl.baseUrl,
+              host: modifiedHost,
               priority: serviceUrl.priority,
               id: service.id,
               ttl: 0,
             });
           }
         });
-      }
-    });
+      });
 
     return clusters;
   },
--- a/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
+++ b/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
@@ -104,6 +104,41 @@ const Services = WebexPlugin.extend({
     return catalog.markFailedServiceUrl(url);
   },
 
+  /*
+   * Get all Mobius cluster host entries from the v2 services list
+   */
+  getMobiusClusters(): Array<{
+    host: string;
+    ttl: number;
+    priority: number;
+    id: string;
+    homeCluster?: boolean;
+  }> {
+    const clusters: Array<{
+      host: string;
+      ttl: number;
+      priority: number;
+      id: string;
+      homeCluster?: boolean;
+    }> = [];
+    const services: Array<Service> = this._services || [];
+
+    services.forEach((service) => {
+      if (service?.serviceName === 'mobius' && Array.isArray(service.serviceUrls)) {
+        service.serviceUrls.forEach((serviceUrl) => {
+          clusters.push({
+            host: serviceUrl.host,
+            priority: serviceUrl.priority,
+            id: service.id,
+            ttl: 0,
+          });
+        });
+      }
+    });
+
+    return clusters;
+  },
+
   /**
    * saves all the services from the pre and post catalog service
    * @param {ActiveServices} activeServices
--- a/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
+++ b/packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
@@ -105,7 +105,7 @@ const Services = WebexPlugin.extend({
   },
 
   /*
-   * Get all Mobius cluster host entries from the v2 services list
+   *  Get all Mobius cluster host entries from the v2 services list.
    */
   getMobiusClusters(): Array<{
     host: string;
@@ -126,12 +126,15 @@ const Services = WebexPlugin.extend({
     services.forEach((service) => {
       if (service?.serviceName === 'mobius' && Array.isArray(service.serviceUrls)) {
         service.serviceUrls.forEach((serviceUrl) => {
-          clusters.push({
-            host: serviceUrl.host,
-            priority: serviceUrl.priority,
-            id: service.id,
-            ttl: 0,
-          });
+          if (!clusters.find((c) => c && c.host === serviceUrl.baseUrl)) {
+            serviceUrl.baseUrl = serviceUrl.baseUrl.replace('https://', '').replace('/api/v1', '');
+            clusters.push({
+              host: serviceUrl.baseUrl,
+              priority: serviceUrl.priority,
+              id: service.id,
+              ttl: 0,
+            });
+          }
         });
       }
     });
--- a/packages/@webex/webex-core/src/lib/services/services.js
+++ b/packages/@webex/webex-core/src/lib/services/services.js
@@ -61,6 +61,9 @@ const Services = WebexPlugin.extend({
 
   _hostCatalog: null,
 
+  // Map of active cluster ids per service, e.g. { wdm: 'urn:TEAM:ap-southeast-2_m:wdm' }
+  _activeServices: {},
+
   /**
    * Get the registry associated with this webex instance.
    *
@@ -103,6 +106,21 @@ const Services = WebexPlugin.extend({
    */
   get(name, priorityHost, serviceGroup) {
     const catalog = this._getCatalog();
+
+    // Prefer active cluster selection when available (e.g., in-region WDM)
+    const clusterId = this._activeServices && this._activeServices[name];
+
+    if (clusterId) {
+      const service = catalog.findServiceFromClusterId({
+        clusterId,
+        priorityHost,
+        serviceGroup,
+      });
+
+      if (service && service.url) {
+        return service.url;
+      }
+    }
 
     return catalog.get(name, priorityHost, serviceGroup);
   },
@@ -158,6 +176,39 @@ const Services = WebexPlugin.extend({
     const catalog = this._getCatalog();
 
     return catalog.markFailedUrl(url, noPriorityHosts);
+  },
+
+  /**
+   * Get all Mobius cluster host entries from the legacy host catalog.
+   * @returns {Array<Object>}
+   */
+  getMobiusClusters() {
+    const clusters = [];
+    const hostCatalog = this._hostCatalog || {};
+
+    Object.entries(hostCatalog).forEach(([host, entries]) => {
+      (entries || []).forEach((entry) => {
+        if (typeof entry?.id === 'string' && entry.id.endsWith(':mobius')) {
+          // Ensure host is included; prefer entry.host if present, else use the map key
+          const withHost = entry.host ? entry.host : host;
+          // Skip duplicates for the same host
+          if (!clusters.find((c) => c && c.host === withHost)) {
+            clusters.push({...entry, host: withHost});
+          }
+        }
+      });
+    });
+
+    return clusters;
+  },
+
+  /**
+   * Merge provided active cluster mappings into current state.
+   * @param {Record<string,string>} activeServices
+   * @returns {void}
+   */
+  _updateActiveServices(activeServices) {
+    this._activeServices = {...this._activeServices, ...activeServices};
   },
 
   /**
@@ -498,6 +549,30 @@ const Services = WebexPlugin.extend({
         // On failure, reject with error from **License**.
         .catch((error) => Promise.reject(error))
     );
+  },
+
+  /**
+   * Update cluster id via mercury/service update. Allows selecting in-region URLs (e.g., WDM).
+   * If a provided clusterId cannot be resolved to a service, it will be ignored.
+   * @param {Record<string,string>} newActiveClusters - e.g. { wdm: 'urn:TEAM:ap-southeast-2_m:wdm' }
+   * @returns {Promise<void>}
+   */
+  switchActiveClusterIds(newActiveClusters) {
+    this.logger.info('services: switching active cluster ids');
+
+    if (!newActiveClusters || typeof newActiveClusters !== 'object') {
+      this.logger.warn('services: invalid newActiveClusters provided to switchActiveClusterIds');
+
+      return Promise.resolve();
+    }
+
+    // Do not validate against current catalog readiness; persist desired mapping immediately.
+    // When catalogs are ready, `get()` will resolve URLs via clusterId.
+    this._updateActiveServices(newActiveClusters);
+
+    this.logger.info('services: active cluster ids updated successfully');
+
+    return Promise.resolve();
   },
 
   /**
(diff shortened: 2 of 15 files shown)


packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
```diff
@@ -104,6 +104,41 @@ const Services = WebexPlugin.extend({
...
+      host: string;
+      ttl: number;
+      priority: number;
+      id: string;
+      homeCluster?: boolean;
+    }> = [];
+    const services: Array<Service> = this._services || [];
+
+    services.forEach((service) => {
+      if (service?.serviceName === 'mobius' && Array.isArray(service.serviceUrls)) {
```
adhmenon: Nit: Can we have a constant for this.

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -104,6 +107,21 @@ const Services = WebexPlugin.extend({
   get(name, priorityHost, serviceGroup) {
     const catalog = this._getCatalog();
 
+    // Prefer active cluster selection when available (e.g., in-region WDM)
+    const clusterId = this._activeServices && this._activeServices[name];
```
adhmenon: We use this method in the meetings plugin as well... have we tested that to see if it is affected?

Maybe we should have a separate method with this logic?

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
```diff
@@ -104,6 +104,41 @@ const Services = WebexPlugin.extend({
...
+    id: string;
+    homeCluster?: boolean;
+  }> {
+    const clusters: Array<{
+      host: string;
+      ttl: number;
+      priority: number;
+      id: string;
+      homeCluster?: boolean;
+    }> = [];
```
rarajes2: 1. We can create an interface for this, being used multiple times.
2. Please add logs. Use the marker as `servicesV2`

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
rarajes2: 1. Q. Can we have different type other than an array for `service.serviceUrls` ?
2. We could have used filter and map for better readability

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
rarajes2, line 36: This could be changed to `ServicesV2` since we already have same namespace for V1

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -104,6 +107,21 @@ const Services = WebexPlugin.extend({
...
+      const service = catalog.findServiceFromClusterId({
+        clusterId,
+        priorityHost,
+        serviceGroup,
+      });
+
+      if (service && service.url) {
+        return service.url;
+      }
+    }
```
rarajes2: Can we try using the early return design? Helps keep the code less indented

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -160,6 +178,39 @@ const Services = WebexPlugin.extend({
     return catalog.markFailedUrl(url, noPriorityHosts);
   },
 
+  /**
+   * Get all Mobius cluster host entries from the legacy host catalog.
+   * @returns {Array<Object>}
```
rarajes2, line 168: Can we add the actual type instead of `Object` here ?

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -500,6 +551,30 @@ const Services = WebexPlugin.extend({
...
   },
 
+  /**
+   * Update cluster id via mercury/service update. Allows selecting in-region URLs (e.g., WDM).
+   * If a provided clusterId cannot be resolved to a service, it will be ignored.
+   * @param {Record<string,string>} newActiveClusters - e.g. { wdm: 'urn:TEAM:ap-southeast-2_m:wdm' }
+   * @returns {Promise<void>}
+   */
+  switchActiveClusterIds(newActiveClusters) {
+    this.logger.info('services: switching active cluster ids');
```
rarajes2: Should we use the marker as `servicesV1` instead of `services` ?

packages/calling/src/CallHistory/CallHistory.ts
```diff
@@ -102,7 +102,9 @@ export class CallHistory extends Eventing<CallHistoryEventTypes> implements ICal
       SDKConnector.setWebex(webex);
     }
     this.webex = this.sdkConnector.getWebex();
-    this.janusUrl = this.webex.internal.services._serviceUrls.janus;
+    this.janusUrl =
+      this.webex.internal.services._serviceUrls.janus ||
+      this.webex.internal.services.get(this.webex.internal.services._activeServices.janus);
```
rarajes2, line 107: Q. What is this janus service? Do we use it directly or via the Mobius for calling ?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -157,32 +151,10 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-      const mobiusObject = this.webex.internal.services._services.find(
-        // @ts-ignore
-        (item) => item.serviceName === 'mobius'
-      );
-      this.mobiusClusters = [mobiusObject.serviceUrls[0].baseUrl];
-    }
+    // Use unified helper to fetch all mobius hosts from catalog (legacy or v2)
+    // Returns an array of ServiceHost-like entries: {host, priority, id, ttl}
+    // @ts-ignore - services may be v1 or v2, both expose getMobiusClusters()
+    this.mobiusClusters = this.webex.internal.services.getMobiusClusters();
```
rarajes2: See if we can typecast the value from `getMobiusClusters()` instead of using `// @ts-ignore`

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -401,109 +373,71 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-        });
+      const clientRegionInfo = response.body as ClientRegionInfo;
 
-        log.log(`Response trackingId: ${temp?.headers?.trackingid}`, {
-          file: CALLING_CLIENT_FILE,
-          method: METHODS.GET_CLIENT_REGION_INFO,
-        });
-
-        const myIP = (temp.body as IpInfo).ipv4;
+      regionInfo.clientRegion = clientRegionInfo?.clientRegion ? clientRegionInfo.clientRegion : '';
```
rarajes2: `clientRegionInfo?.clientRegion ? clientRegionInfo.clientRegion : '';` -> `clientRegionInfo?.clientRegion || '';`

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -401,109 +373,71 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-      } else {
-        this.mobiusHost = mobius as unknown as string;
-      }
+    try {
+      const response = <WebexRequestPayload>await this.webex.request({
+        uri: `${DISCOVERY_URL}`,
+        method: HTTP_METHODS.GET,
+        addAuthHeader: false,
+        headers: {
+          [SPARK_USER_AGENT]: null,
```
rarajes2, line 379: Q. Why are we using `null` here ?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -401,109 +373,71 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-        // eslint-disable-next-line no-await-in-loop
-        const response = <WebexRequestPayload>await this.webex.request({
-          uri: `${DISCOVERY_URL}/${myIP}`,
-          method: HTTP_METHODS.GET,
-          addAuthHeader: false,
-          headers: {
-            [SPARK_USER_AGENT]: null,
-          },
-        });
+      regionInfo.countryCode = clientRegionInfo?.countryCode ? clientRegionInfo.countryCode : '';
```
rarajes2: Same here

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -561,79 +497,89 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-            [SPARK_USER_AGENT]: CALLING_USER_AGENT,
-          },
-          service: ALLOWED_SERVICES.MOBIUS,
-        });
+      for (const mobius of this.mobiusClusters) {
+        if (mobius.host) {
+          this.mobiusHost = `https://${mobius.host}${API_V1}`;
+        } else {
+          this.mobiusHost = mobius as unknown as string;
+        }
```
rarajes2, line 502: Good case for using ternary operator

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -561,79 +497,89 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+          // eslint-disable-next-line no-await-in-loop
+          const abort = await handleCallingClientErrors(
+            err as WebexRequestPayload,
+            (clientError) => {
+              this.metricManager.submitRegistrationMetric(
+                METRIC_EVENT.REGISTRATION_ERROR,
+                REG_ACTION.REGISTER,
+                METRIC_TYPE.BEHAVIORAL,
+                GET_MOBIUS_SERVERS_UTIL,
+                'UNKNOWN',
```
rarajes2, line 563: Is it possible to avoid `UNKNOWN` here ?

packages/calling/src/CallSettings/WxCallBackendConnector.ts
```diff
@@ -67,7 +67,9 @@ export class WxCallBackendConnector implements IWxCallBackendConnector {
 
     this.webex = this.sdkConnector.getWebex();
     /* eslint no-underscore-dangle: 0 */
-    this.hydraEndpoint = this.webex.internal.services._serviceUrls.hydra;
+    this.hydraEndpoint =
+      this.webex.internal.services._serviceUrls.hydra ||
+      this.webex.internal.services.get(this.webex.internal.services._activeServices.hydra);
```
rarajes2, line 72: Q. Why was this change required ?

packages/calling/src/Voicemail/UcmBackendConnector.ts
```diff
@@ -250,7 +250,10 @@ export class UcmBackendConnector implements IUcmBackendConnector {
 
     return new Promise((resolve, reject) => {
       const voicemailContentUrl = `${this.vgVoiceMessageURI}${VOICEMAILS}/${messageId}/${CONTENT}`;
-      const mercuryApi = `${this.webex.internal.services._serviceUrls.mercuryApi}`;
+      const mercuryUrl =
+        this.webex.internal.services._serviceUrls.mercuryApi ||
+        this.webex.internal.services.get(this.webex.internal.services._activeServices.mercuryApi);
+      const mercuryApi = `${mercuryUrl}`;
```
rarajes2: Do we need this line? Why can't we directly use `mercuryUrl` variable ?

rarajes2 (commented):
Are we not adding unit tests for services (v1 and v2) ?

packages/calling/src/CallHistory/CallHistory.ts
Kesari3008: We use it directly, it's a microservice used directly for fetching call history information

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: This code might get removed, I added it during investigation for in-region WDM URL

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
Kesari3008: It is intentionally kept same. At a moment, either v1 or v catalog will be used but they are present within the webex object as services object only. 
Team implementing the v2 catalog implemented it in that manner I believe

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: This will be removed, this is not needed

packages/calling/src/CallSettings/WxCallBackendConnector.ts
Kesari3008: We are only pulling all these urls from v1 catalog so for cases where v2 catalog gets used, urls return undefined and   api start to fail

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: With the new metric added, this code won't be there only

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: Unused code, not required. Removed now

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: This code is in js, not using ts. No types files even exists so creating a type file with one type created would be unnecessary I felt
For service.v2, I can add that

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: I am not sure, it's an existing code only. But I am assuming, it might because region discovery url(ds.ciscospark.com) doesn't require this param

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: This is addressed

packages/calling/src/Voicemail/UcmBackendConnector.ts
Kesari3008: Updated

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: updated

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: updated

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
Kesari3008: ServiceUrls are received from backend itself in the form of array and this is how the type for that is created so I believe changing that won't be possible. I will try to see if readability can be improved

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
Kesari3008: Using that marke won;t be ideal as we not want to showcase for user if it's v1 or v2

packages/@webex/webex-core/src/lib/services/services.js
```diff
@@ -104,6 +107,21 @@ const Services = WebexPlugin.extend({
   get(name, priorityHost, serviceGroup) {
     const catalog = this._getCatalog();
 
+    // // Prefer active cluster selection when available (e.g., in-region WDM)
```
adhmenon: Nit: We can remove this.

adhmenon (approved):

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: I thought i removed it but I only commented. Thanks for noticing. Removed now

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
rarajes2: Are we not adding logs at all for V2 ?

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
rarajes2: > ServiceUrls are received from backend itself in the form of array

Do we need to check the type in that case ?

packages/@webex/webex-core/src/lib/services/services.js
rarajes2: I see this code is removed

packages/@webex/webex-core/src/lib/services/services.js
rarajes2: I meant we can add the object shape in the jsdoc, that way it helps the developer to understand what is being returned. We can't add typescript type but jsdoc can have the object's shape

packages/calling/src/CallSettings/WxCallBackendConnector.ts
rarajes2: So for V2 cases we can only access the url using the `get` method ?

packages/calling/src/Voicemail/BroadworksBackendConnector.ts
```diff
@@ -167,9 +167,14 @@ export class BroadworksBackendConnector implements IBroadworksCallBackendConnect
     };
 
     log.info(METHOD_START_MESSAGE, loggerContext);
+    const broadworksIdpProxyUrl =
+      this.webex.internal.services._serviceUrls?.broadworksIdpProxy ||
+      this.webex.internal.services.get(
+        this.webex.internal.services._activeServices.broadworksIdpProxy
+      );
```
rarajes2, line 174: nitpick: We can de-structure `this.webex.internal.services` -> `const { services } = this.webex.internal`

With destructuring it will look cleaner

```
const { services } = this.webex.internal;

const broadworksIdpProxyUrl = services._serviceUrls?.broadworksIdpProxy || services.get(services._activeServices.broadworksIdpProxy)
```


This pattern is repeated.

packages/calling/src/CallSettings/WxCallBackendConnector.ts
Kesari3008: Yes

packages/@webex/webex-core/src/lib/services/services.js
Kesari3008: Yes this was added by me only for in-region mercury investigation. This is not needed

packages/@webex/webex-core/src/lib/services-v2/services-v2.ts
Kesari3008: It's a check to ensure that we do not perform that for loop on something which is not array and if catalog team makes some changes, we might get into the error so I kept the check

Kesari3008:
> Are we not adding unit tests for services (v1 and v2) ?

Added

rarajes2 (approved):

PR #4557, opened 2025-10-31
fix(calling): update call keepalive retry logic

rarajes2:
# COMPLETES #< [CAI-7258](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7258) >

Adds the call keepalive error handling scenarios for 4xx and 5xx cases where for 4xx cases call cleanup happens and for 5xx keepalive retries are done based on the value received in the response header `retry-after` followed by fallback value if it's not present.

Normal keepalive for call will be **10 minutes** but if receives 5xx response then it will schedule the re-tries based on the **after-retry** header value and fallback to **30 seconds**

Vidcast - https://app.vidcast.io/share/670a12d0-1b16-4561-8d86-f34df10e3a63

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested the following with the burp suite (to update the response header)

- For 4xx cases the call was being cleaned up
- For 5xx cases with retry-after, it honors the value if provided else fallback to 30 sec retry interval
- Tries max of 4 retries and if all fails it stops the retries
- Failed 3 keepalives and then let 1 pass which did reset the counter and again letting the keepalive fail for 4 times stops sending further keepalive request

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -1072,6 +1072,11 @@ describe('State Machine handler tests', () => {
   });
 
   it('keepalive ends after reaching max retry count', async () => {
+    const resolvePromise = async () => {
+      await Promise.resolve();
+      await Promise.resolve();
+    };
+
     const errorPayload = <WebexRequestPayload>(<unknown>{
       statusCode: 500,
       headers: {
@@ -1081,18 +1086,29 @@ describe('State Machine handler tests', () => {
 
     jest.spyOn(global, 'clearInterval');
     const warnSpy = jest.spyOn(log, 'warn');
-    jest.spyOn(call, 'postStatus').mockRejectedValue(errorPayload);
+    const postStatusSpy = jest.spyOn(call, 'postStatus').mockRejectedValue(errorPayload);
 
-    // Manually set the retry count to 3 so it becomes 4 after increment
-    call['callKeepaliveRetryCount'] = 3;
+    // Put the call in the S_CALL_ESTABLISHED state and set it as connected
+    call['callStateMachine'].state.value = 'S_CALL_ESTABLISHED';
+    call['connected'] = true;
 
     // Call handleCallEstablished which will setup interval
     call['handleCallEstablished']({} as CallEvent);
 
-    // Advance timer to trigger the failure
+    // Advance timer to trigger the first failure (uses DEFAULT_SESSION_TIMER)
     jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
-    await Promise.resolve();
-    await Promise.resolve();
+    await resolvePromise();
+
+    // Now advance by 1 second for each of the 3 more retry attempts (retry-after: 1 second each)
+    // Need to do this separately to allow state machine to process and create new intervals
+    jest.advanceTimersByTime(1000);
+    await resolvePromise();
+
+    jest.advanceTimersByTime(1000);
+    await resolvePromise();
+
+    jest.advanceTimersByTime(1000);
+    await resolvePromise();
 
     // The error handler should detect we're at max retry count and stop
     expect(warnSpy).toHaveBeenCalledWith(
@@ -1102,6 +1118,7 @@ describe('State Machine handler tests', () => {
         method: 'handleCallEstablished',
       }
     );
+    expect(postStatusSpy).toHaveBeenCalledTimes(4);
     expect(call['callKeepaliveRetryCount']).toBe(0);
     expect(call['sessionTimer']).toBeUndefined();
   });
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -977,8 +977,38 @@ describe('State Machine handler tests', () => {
     );
   });
 
-  it('session refresh failure', async () => {
+  it('session refresh 401 emits token error and ends the call', async () => {
     expect.assertions(4);
+    const statusPayload = <WebexRequestPayload>(<unknown>{
+      statusCode: 401,
+    });
+
+    webex.request.mockReturnValue(statusPayload);
+    jest.spyOn(global, 'clearInterval');
+
+    const endSpy = jest.spyOn(call as any, 'end');
+
+    call.on(CALL_EVENT_KEYS.CALL_ERROR, (errObj) => {
+      expect(errObj.type).toStrictEqual(ERROR_TYPE.TOKEN_ERROR);
+    });
+
+    const funcSpy = jest.spyOn(call, 'postStatus').mockRejectedValue(statusPayload);
+
+    if (call['sessionTimer'] === undefined) {
+      call['handleCallEstablished']({} as CallEvent);
+    }
+    call['handleCallEstablished']({} as CallEvent);
+
+    jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
+    await Promise.resolve();
+    await Promise.resolve();
+
+    expect(clearInterval).toHaveBeenCalled();
+    expect(funcSpy).toBeCalledTimes(1);
+    expect(endSpy).toHaveBeenCalled();
+  });
+
+  it('session refresh 403 ends the call without emitting error', async () => {
     const statusPayload = <WebexRequestPayload>(<unknown>{
       statusCode: 403,
     });
@@ -986,33 +1016,72 @@ describe('State Machine handler tests', () => {
     webex.request.mockReturnValue(statusPayload);
     jest.spyOn(global, 'clearInterval');
 
-    call.on(CALL_EVENT_KEYS.CALL_ERROR, (errObj) => {
-      expect(errObj.type).toStrictEqual(ERROR_TYPE.FORBIDDEN_ERROR);
-      expect(errObj.message).toStrictEqual(
-        'An unauthorized action has been received. This action has been blocked. Please contact the administrator if this persists.'
-      );
-    });
+    const endSpy = jest.spyOn(call as any, 'end');
+    const errorListener = jest.fn();
+    call.on(CALL_EVENT_KEYS.CALL_ERROR, errorListener);
 
     const funcSpy = jest.spyOn(call, 'postStatus').mockRejectedValue(statusPayload);
 
     if (call['sessionTimer'] === undefined) {
-      /* In cases where this test is run independently/alone, there is no sessionTimer initiated
-      Thus we will check and initialize the timer when not present by calling handleCallEstablish() */
       call['handleCallEstablished']({} as CallEvent);
     }
     call['handleCallEstablished']({} as CallEvent);
 
     jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
-
-    /* This is to flush all the promises from the Promise queue so that
-     * Jest.fakeTimers can advance time and also clear the promise Queue
-     */
-
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(clearInterval).toHaveBeenCalledTimes(2); // check this
+    expect(clearInterval).toHaveBeenCalled();
     expect(funcSpy).toBeCalledTimes(1);
+    expect(endSpy).toHaveBeenCalled();
+    expect(errorListener).not.toHaveBeenCalled();
+  });
+
+  it('session refresh 500 schedules retry via retry-after or default interval', async () => {
+    const errorPayload = <WebexRequestPayload>(<unknown>{
+      statusCode: 500,
+      headers: {
+        'retry-after': 1,
+      },
+    });
+
+    const okPayload = <WebexRequestPayload>(<unknown>{statusCode: 200, body: {}});
+
+    jest.spyOn(global, 'setTimeout');
+    const sendEvtSpy = jest.spyOn(call as any, 'sendCallStateMachineEvt');
+    const postStatusSpy = jest
+      .spyOn(call as any, 'postStatus')
+      .mockRejectedValueOnce(errorPayload)
+      .mockResolvedValueOnce(okPayload);
+
+    if (call['sessionTimer'] === undefined) {
+      call['handleCallEstablished']({} as CallEvent);
+    }
+
+    jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
+    await Promise.resolve();
+
+    jest.advanceTimersByTime(1005);
+
+    expect(setTimeout).toHaveBeenCalled();
+    expect(postStatusSpy).toHaveBeenCalledTimes(2);
+    expect(sendEvtSpy).toHaveBeenCalledWith({type: 'E_CALL_ESTABLISHED'});
+  });
+
+  it('keepalive ends after reaching max retry count', async () => {
+    jest.spyOn(global, 'clearInterval');
+    const endSpy = jest.spyOn(call as any, 'end');
+
+    call['sessionTimer'] = setInterval(() => {}, 1000);
+
+    call['handleCallEstablished']({} as CallEvent);
+    call['handleCallEstablished']({} as CallEvent);
+    call['handleCallEstablished']({} as CallEvent);
+    call['handleCallEstablished']({} as CallEvent);
+    call['handleCallEstablished']({} as CallEvent);
+
+    expect(endSpy).toHaveBeenCalled();
+    expect(call['callKeepaliveRetryCount']).toBe(0);
   });
 
   it('state changes during successful incoming call', async () => {
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -994,21 +994,19 @@ describe('State Machine handler tests', () => {
 
     const funcSpy = jest.spyOn(call, 'postStatus').mockRejectedValue(statusPayload);
 
-    if (call['sessionTimer'] === undefined) {
-      call['handleCallEstablished']({} as CallEvent);
-    }
     call['handleCallEstablished']({} as CallEvent);
 
     jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(clearInterval).toHaveBeenCalled();
+    expect(clearInterval).toHaveBeenCalledTimes(1);
     expect(funcSpy).toBeCalledTimes(1);
     expect(emitSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.DISCONNECT, call.getCorrelationId());
   });
 
-  it('session refresh 403 ends the call without emitting error', async () => {
+  it('session refresh failure', async () => {
+    expect.assertions(4);
     const statusPayload = <WebexRequestPayload>(<unknown>{
       statusCode: 403,
     });
@@ -1016,25 +1014,33 @@ describe('State Machine handler tests', () => {
     webex.request.mockReturnValue(statusPayload);
     jest.spyOn(global, 'clearInterval');
 
-    const emitSpy = jest.spyOn(call, 'emit');
-    const errorListener = jest.fn();
-    call.on(CALL_EVENT_KEYS.CALL_ERROR, errorListener);
+    call.on(CALL_EVENT_KEYS.CALL_ERROR, (errObj) => {
+      expect(errObj.type).toStrictEqual(ERROR_TYPE.FORBIDDEN_ERROR);
+      expect(errObj.message).toStrictEqual(
+        'An unauthorized action has been received. This action has been blocked. Please contact the administrator if this persists.'
+      );
+    });
 
     const funcSpy = jest.spyOn(call, 'postStatus').mockRejectedValue(statusPayload);
 
     if (call['sessionTimer'] === undefined) {
+      /* In cases where this test is run independently/alone, there is no sessionTimer initiated
+      Thus we will check and initialize the timer when not present by calling handleCallEstablish() */
       call['handleCallEstablished']({} as CallEvent);
     }
     call['handleCallEstablished']({} as CallEvent);
 
     jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
+
+    /* This is to flush all the promises from the Promise queue so that
+     * Jest.fakeTimers can advance time and also clear the promise Queue
+     */
+
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(clearInterval).toHaveBeenCalled();
+    expect(clearInterval).toHaveBeenCalledTimes(2); // check this
     expect(funcSpy).toBeCalledTimes(1);
-    expect(emitSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.DISCONNECT, call.getCorrelationId());
-    expect(errorListener).not.toHaveBeenCalled();
   });
 
   it('session refresh 500 schedules retry via retry-after or default interval', async () => {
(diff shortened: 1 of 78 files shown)


packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1520,6 +1524,16 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
         method: METHODS.HANDLE_CALL_ESTABLISHED,
       });
       clearInterval(this.sessionTimer);
+
+      // If we have reached the max retry count, do not attempt to refresh the session
+      if (this.callKeepaliveRetryCount === MAX_CALL_KEEPALIVE_RETRY_COUNT) {
+        this.end();
```
Kesari3008: Would be good to use state mahcine to trigger discponnect instead of invoking the method directly

packages/calling/src/common/Utils.ts
```diff
@@ -584,6 +587,9 @@ export async function handleCallErrors(
 
   log.warn(`Status code: ->${errorCode}`, loggerContext);
 
+  const isKeepalive =
+    caller === 'handleCallEstablished' || caller === METHODS.HANDLE_CALL_ESTABLISHED;
```
Kesari3008: second one is nothing but the same string s why do we need to explicitly add it

packages/calling/src/common/Utils.ts
```diff
@@ -728,11 +752,25 @@ export async function handleCallErrors(
       );
 
       emitterCb(callError);
+
+      if (isKeepalive && retryCb) {
+        if (err.headers && 'retry-after' in err.headers) {
```
Kesari3008: based on this condition we can just update retryInterval value and only invoke retyrCB once

packages/calling/src/common/Utils.ts
```diff
@@ -572,7 +574,8 @@ export async function handleCallErrors(
   correlationId: CorrelationId,
   err: WebexRequestPayload,
   caller: string,
-  file: string
+  file: string,
+  endCall?: () => void
```
Kesari3008: Instead of this method, pass a callback for cleanup logic and also have a common place to execute either cleanup logic or retry logic for final and non-final errors. Here we are calling the method again and again so same code repetetion can be avoided

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
     );
   });
 
-  it('session refresh failure', async () => {
+  it('session refresh 401 emits token error and ends the call', async () => {
     expect.assertions(4);
     const statusPayload = <WebexRequestPayload>(<unknown>{
-      statusCode: 403,
+      statusCode: 401,
```
Kesari3008, line 983: Why did we change this, result is same for both error codes right

Kesari3008 (changes requested):
Haven't looked at the test files for now, because there are comments for code changes.

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
     if (call['sessionTimer'] === undefined) {
-      /* In cases where this test is run independently/alone, there is no sessionTimer initiated
-      Thus we will check and initialize the timer when not present by calling handleCallEstablish() */
       call['handleCallEstablished']({} as CallEvent);
     }
     call['handleCallEstablished']({} as CallEvent);
 
     jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
+    await Promise.resolve();
+    await Promise.resolve();
```
adhmenon, line 1001: Why are there 2 promise resolves here?

adhmenon (commented):
Is it possible to test this with the agent desktop?
Just want to ensure it does not break the flow.

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Done

packages/calling/src/common/Utils.ts
rarajes2: Fixed

packages/calling/src/common/Utils.ts
rarajes2: Done

packages/calling/src/common/Utils.ts
rarajes2: > Instead of this method, pass a callback for cleanup logic 

Will `end` method not clean the call state ?

> and also have a common place to execute either cleanup logic or retry logic for final and non-final errors.

Need more info with some example

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: We need to add these two Promise.resolve() because the processNextTick() is used internally by Jest.fakeTimers to flush the Promise queue. The processNextTick() is basically a macro for setTimeout(fn, 0) which means it will be added to the end of the Promise queue. So, to ensure that the Promise queue is flushed completely, we need to call Promise.resolve() twice.

adhmenon (approved):
Approved.
But let us test once with agent desktop just to be safe.

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Right, was just testing this and missed to revert back. It does not make any difference. Let me know if you want me to revert this.

packages/calling/src/common/Utils.ts
rarajes2: As discussed, returning the `abort` flag from the `handleCallError` and based on that clearing the call state. Not passing the `call.end` as callback anymore.

packages/calling/src/CallingClient/calling/call.test.ts
Kesari3008: Either revert the test as it is or add the removed expect statements back

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
 
-    /* This is to flush all the promises from the Promise queue so that
-     * Jest.fakeTimers can advance time and also clear the promise Queue
-     */
+    expect(clearInterval).toHaveBeenCalled();
+    expect(funcSpy).toBeCalledTimes(1);
+    expect(emitSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.DISCONNECT, call.getCorrelationId());
+  });
+
+  it('session refresh 403 ends the call without emitting error', async () => {
```
rarajes2: 403 case is update here.

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Ah my bad. This is a new test case. For 403 the test was updated. Please check [this](https://github.com/webex/webex-js-sdk/pull/4557/files#r2523576979)

packages/calling/src/common/Utils.ts
```diff
@@ -597,10 +605,20 @@ export async function handleCallErrors(
...
+        abort = true;
+      }
+
       break;
     }
 
     case ERROR_CODE.FORBIDDEN:
+      if (isKeepalive) {
+        abort = true;
+        break;
```
Kesari3008, line 628: break can be outside the if

packages/calling/src/common/Utils.ts
```diff
@@ -619,18 +637,22 @@ export async function handleCallErrors(
         );
         emitterCb(callError);
 
-        return;
+        return abort;
```
Kesari3008, line 649: We don't have to return abort here, just don;t update update for non-final error cases and retun once at the end should suffice

packages/calling/src/common/Utils.ts
```diff
@@ -619,18 +637,22 @@ export async function handleCallErrors(
...
       }
 
       /* Handle retry-after cases */
 
+      // Common for keepalive and normal scenarios
       if (err.headers && 'retry-after' in err.headers && retryCb) {
         const retryInterval = Number(err.headers['retry-after'] as unknown);
 
         log.warn(`Retry Interval received: ${retryInterval}`, loggerContext);
         retryCb(retryInterval);
```
Kesari3008, line 659: We can move out invoking retryCb outside the if and even the return abort. Once the retry interval is updated we can just call retyr cb at once

packages/calling/src/common/Utils.ts
```diff
@@ -728,13 +755,29 @@ export async function handleCallErrors(
...
+        retryCb(retryInterval);
+      }
+
       break;
     }
 
     default: {
       log.warn(`Unknown Error`, loggerContext);
+
+      if (isKeepalive && retryCb) {
```
Kesari3008: We shouldn't invoke call keepalive in 30 seconds for default cases

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1520,6 +1524,16 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
         method: METHODS.HANDLE_CALL_ESTABLISHED,
       });
       clearInterval(this.sessionTimer);
+
+      // If we have reached the max retry count, do not attempt to refresh the session
+      if (this.callKeepaliveRetryCount === MAX_CALL_KEEPALIVE_RETRY_COUNT) {
```
Kesari3008: Shouldn't this block be inside the catch block where we have the error handling

packages/calling/src/CallingClient/calling/call.ts
Kesari3008, line 1547: This block should be removed we should not clear the interval now that we call cleanup logic

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2537,7 +2558,7 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
           },
           ERROR_LAYER.CALL_CONTROL,
           // eslint-disable-next-line @typescript-eslint/no-unused-vars
-          /* istanbul ignore next */ (interval: number) => undefined,
+          /* istanbul ignore next */ (interval?: number) => undefined,
```
Kesari3008: Why did we add ? here

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
     });
 
     webex.request.mockReturnValue(statusPayload);
     jest.spyOn(global, 'clearInterval');
 
+    const emitSpy = jest.spyOn(call, 'emit');
+
     call.on(CALL_EVENT_KEYS.CALL_ERROR, (errObj) => {
-      expect(errObj.type).toStrictEqual(ERROR_TYPE.FORBIDDEN_ERROR);
-      expect(errObj.message).toStrictEqual(
```
Kesari3008, line 991: Please add this back

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
+      call['handleCallEstablished']({} as CallEvent);
+    }
+    call['handleCallEstablished']({} as CallEvent);
+
+    jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(clearInterval).toHaveBeenCalledTimes(2); // check this
+    expect(clearInterval).toHaveBeenCalled();
```
Kesari3008: we should have check for this as how many times its has been invoked

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
+    const statusPayload = <WebexRequestPayload>(<unknown>{
+      statusCode: 403,
+    });
 
+    webex.request.mockReturnValue(statusPayload);
+    jest.spyOn(global, 'clearInterval');
+
+    const emitSpy = jest.spyOn(call, 'emit');
+    const errorListener = jest.fn();
+    call.on(CALL_EVENT_KEYS.CALL_ERROR, errorListener);
```
Kesari3008: We need not to overwrite the listener, we should be able to spy on the 'on' event itself and check for it;s invocation for Call_Error event

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,42 +977,111 @@ describe('State Machine handler tests', () => {
...
+    expect(sendEvtSpy).toHaveBeenCalledWith({type: 'E_CALL_ESTABLISHED'});
+  });
+
+  it('keepalive ends after reaching max retry count', async () => {
+    jest.spyOn(global, 'clearInterval');
+    const emitSpy = jest.spyOn(call, 'emit');
+
+    call['sessionTimer'] = setInterval(() => {}, 1000);
+
+    call['handleCallEstablished']({} as CallEvent);
```
Kesari3008: Is it possible to add checks for keepalive count incrementing and post status being tried step by steps and then eventualy disconnect happening

Kesari3008 (changes requested):

packages/calling/src/common/Utils.ts
rarajes2: As discussed, this logic is fine. Added additional error handling inside the if block

packages/calling/src/common/Utils.ts
rarajes2: Discussed. This is fine.

packages/calling/src/common/Utils.ts
rarajes2: As discussed, this logic handles both call errors and call keepalive errors. No change is required here.

packages/calling/src/common/Utils.ts
rarajes2: Removed

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Fixed

packages/calling/src/CallingClient/calling/call.ts
rarajes2: This is to clear the existing interval so, this is fine. Discussed

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Fixed

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Done

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Done

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: This UT is removed. Not required

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: This UT is removed. Not required

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1544,34 +1546,55 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
           },
           ERROR_LAYER.CALL_CONTROL,
           (interval: number) => {
-            setTimeout(() => {
-              /* We first post the status and then recursively call the handler which
-               * starts the timer again
-               */
-              this.postStatus();
-              this.sendCallStateMachineEvt({type: 'E_CALL_ESTABLISHED'});
-            }, interval * 1000);
```
rarajes2, line 1560: This `setTimeout` is not required as it was causing the `this.postStatus()` call to be made twice - once here and next on invoking the `handleCallEstablished` through state machine. 

Now we are only calling `handleCallEstablished` through state machine immediately as the `handleCallEstablished` makes the keepalive call inside the `setInterval`  which adds the required delay and keeps the counter consistent.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1544,34 +1546,55 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+          'handleCallEstablished',
           CALL_FILE
         );
 
+        if (abort) {
+          this.sendCallStateMachineEvt({type: 'E_SEND_CALL_DISCONNECT'});
+          this.emit(CALL_EVENT_KEYS.DISCONNECT, this.getCorrelationId());
+          this.callKeepaliveRetryCount = 0;
+
+          return;
```
Kesari3008: We shouldn't skip logs uploading for final error cases, this return would cause that

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1544,34 +1547,55 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
           ERROR_LAYER.CALL_CONTROL,
           (interval: number) => {
-            setTimeout(() => {
-              /* We first post the status and then recursively call the handler which
-               * starts the timer again
-               */
-              this.postStatus();
-              this.sendCallStateMachineEvt({type: 'E_CALL_ESTABLISHED'});
-            }, interval * 1000);
+            this.callKeepaliveRetryCount += 1;
```
Kesari3008, line 1558: We didn't need to remove the setTimeout. Retry logic was implemented in a way that if it fails, we were immediately trying one more time and schedule the timeout to try it again after specific interval. Even if we wanted to avoid the postStatus being sent twice, we should just remove postStatus and keep setting the state machine to established with a interval timeout to trigger the keepalive retry

Kesari3008, line 1558: With this logic with the first keepalive retry itself, count will start increasing

packages/calling/src/common/Utils.test.ts
```diff
@@ -106,6 +112,216 @@ describe('Mobius service discovery tests', () => {
...
+      ERROR_LAYER.CALL_CONTROL,
+      retrySpy,
+      dummyCorrelationId,
+      payload,
+      'handleCallEstablished',
+      logObj.file,
+      endSpy
+    );
+
+    expect(emitted).toBe(true);
```
Kesari3008, line 233: Add check for for abort being false for non-final error cases

packages/calling/src/common/Utils.test.ts
```diff
@@ -106,6 +112,216 @@ describe('Mobius service discovery tests', () => {
...
+      'handleCallEstablished',
+      logObj.file
+    );
+
+    expect(emitted).toBe(true);
+    expect(abort).toBe(true);
+    expect(retrySpy).not.toHaveBeenCalled();
+  });
+
+  it('503 during keepalive with retry-after does not emit and retries with interval', async () => {
```
Kesari3008: it should say does not emit what

packages/calling/src/common/Utils.test.ts
```diff
@@ -106,6 +112,216 @@ describe('Mobius service discovery tests', () => {
...
+      },
+      ERROR_LAYER.CALL_CONTROL,
+      retrySpy,
+      dummyCorrelationId,
+      payload,
+      'handleCallEstablished',
+      logObj.file
+    );
+
+    expect(emitted).toBe(true);
```
Kesari3008, line 176: we should be check for accurate emitter callback where we emit error event instead of plain setting it true or false. Or just pass jest.fn() and check for it's invocation

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -977,6 +977,34 @@ describe('State Machine handler tests', () => {
...
+
+    jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
+    await Promise.resolve();
+    await Promise.resolve();
+
+    expect(clearInterval).toHaveBeenCalledTimes(1);
+    expect(funcSpy).toBeCalledTimes(1);
+    expect(emitSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.DISCONNECT, call.getCorrelationId());
```
Kesari3008, line 1005: we should add expect to check for end API being invoked, media connection getting closed and call getting cleared from callCollection

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -1015,6 +1043,69 @@ describe('State Machine handler tests', () => {
...
+    const postStatusSpy = jest
+      .spyOn(call as any, 'postStatus')
+      .mockRejectedValueOnce(errorPayload)
+      .mockResolvedValueOnce(okPayload);
+
+    if (call['sessionTimer'] === undefined) {
+      call['handleCallEstablished']({} as CallEvent);
+    }
+
+    jest.advanceTimersByTime(DEFAULT_SESSION_TIMER);
```
Kesari3008, line 1066: we should only advance here as per retyr after and not default session timer right

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -1015,6 +1043,69 @@ describe('State Machine handler tests', () => {
...
+        'retry-after': 1,
+      },
+    });
+
+    jest.spyOn(global, 'clearInterval');
+    const warnSpy = jest.spyOn(log, 'warn');
+    jest.spyOn(call, 'postStatus').mockRejectedValue(errorPayload);
+
+    // Manually set the retry count to 3 so it becomes 4 after increment
+    call['callKeepaliveRetryCount'] = 3;
```
Kesari3008: We should let the code run 4 times and count getting increased on its own. And add expect for postStatus being invoked 4 times too

Kesari3008 (approved):
Approving the PR but there are still significant changes needed. Please address these

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Done

packages/calling/src/CallingClient/calling/call.ts
rarajes2: > Even if we wanted to avoid the postStatus being sent twice, we should just remove postStatus and keep setting the state machine to established with a interval timeout to trigger the keepalive retry

This logic is already there. I have removed the unnecessary setTimeout and postStatus, instead I am directly invoking the logic with the state machine. Please check the new logic. If required we can discuss over a call

rarajes2: This is only being increased in case of error.

packages/calling/src/common/Utils.test.ts
rarajes2: Done

packages/calling/src/common/Utils.test.ts
rarajes2: Changed to -> `503 during keepalive with retry-after does not invoke emitterCb and retries with interval`

packages/calling/src/common/Utils.test.ts
rarajes2: This check ensures that the passed `emitterCb` is being invoked. We don't need to make it complicated.

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Wouldn't that make it an integration test? There would be a separate unit test for that already present

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: First time invocation of postStatus will only happen after the `DEFAULT_SESSION_TIMER`

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Done

PR #4559, opened 2025-11-02
fix(calling): fix 429 re-try logic

Shreyas281299:
# COMPLETES # [JIRA](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-715146)

- When a 429 fails we dont honor the retry-after values
- If the retry-after value is huge then we shouldnt wait and connect to the secondary (if connected to primary)
- If we get 429 with secondary then we should just unregister

- When 429 happens we honor the retry-after value and connect to secondary

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Unit tests
- Failed with 404, 403, 429
- Failed Primary with 429 ensured it conects to backup immediately
- Failed Primary with 429 with small retyr-after value - connects to Primary
- Failed Backup with 429 ensured it connects to backup -2 immediately

### 429- on primary and backup - moved to 2nd backup
429- on primary and backup - moved to 2nd backup

### 403-101 handling
403-101 handling

### 429-small interval - connect to primary
429-small interval

### 429-big internal- fallback
429-big internal- fallback

### 429-big internal fallback
429-big internal fallback

  - [ ] Github Copilot
  - [x] Other - Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/registration/register.test.ts
+++ b/packages/calling/src/CallingClient/registration/register.test.ts
@@ -752,8 +752,6 @@ describe('Registration Tests', () => {
       });
       webex.request
         .mockRejectedValueOnce(failurePayload429Small)
-        .mockResolvedValueOnce(successPayload)
-        .mockRejectedValueOnce(failurePayload429Small)
         .mockResolvedValueOnce(successPayload);
 
       await reg.reconnectOnFailure(RECONNECT_ON_FAILURE_UTIL); // This call is being used to set the retry-after value
@@ -782,8 +780,6 @@ describe('Registration Tests', () => {
       });
       webex.request
         .mockRejectedValueOnce(failurePayload429Small)
-        .mockResolvedValueOnce(successPayload)
-        .mockRejectedValueOnce(failurePayload429Small)
         .mockResolvedValueOnce(successPayload);
 
       await reg.reconnectOnFailure(RECONNECT_ON_FAILURE_UTIL); // This call is being used to trigger the retry
@@ -807,6 +803,7 @@ describe('Registration Tests', () => {
       );
       expect(restartSpy).not.toHaveBeenCalledTimes(1);
       expect(restartSpy).not.toHaveBeenCalledWith(RECONNECT_ON_FAILURE_UTIL);
+      expect(reg.retryAfter).toEqual(undefined); // Clear retryAfter after 429 retry
     });
     it('should restart registration with primary if we get 429 while on backup', async () => {
       // Setup: Register successfully with primary first
--- a/packages/calling/src/CallingClient/registration/register.test.ts
+++ b/packages/calling/src/CallingClient/registration/register.test.ts
@@ -31,6 +31,7 @@ import {
   REG_429_RETRY_UTIL,
   REG_TRY_BACKUP_TIMER_VAL_IN_SEC,
   SEC_TO_MSEC_MFACTOR,
+  RECONNECT_ON_FAILURE_UTIL,
 } from '../constants';
 import {ICall} from '../calling/types';
 import {LINE_EVENTS} from '../line/types';
@@ -1277,9 +1278,9 @@ describe('Registration Tests', () => {
 
       expect(reg.getStatus()).toEqual(RegistrationStatus.INACTIVE);
       expect(lineEmitter).toHaveBeenCalledWith(LINE_EVENTS.UNREGISTERED);
-      expect(reconnectSpy).toBeCalledOnceWith(KEEPALIVE_UTIL);
-      expect(restoreSpy).toBeCalledOnceWith(KEEPALIVE_UTIL);
-      expect(restartRegSpy).toBeCalledOnceWith(KEEPALIVE_UTIL);
+      expect(reconnectSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
+      expect(restoreSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
+      expect(restartRegSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
 
       jest.useRealTimers();
 
@@ -1311,7 +1312,7 @@ describe('Registration Tests', () => {
       await flushPromises();
 
       expect(reg.webWorker).toBeUndefined();
-      expect(reconnectSpy).toBeCalledOnceWith(reg.startKeepaliveTimer.name);
+      expect(reconnectSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
 
       webex.request.mockResolvedValueOnce(successPayload);
       await reg.triggerRegistration();
@@ -1329,8 +1330,8 @@ describe('Registration Tests', () => {
       expect(reg.getStatus()).toEqual(RegistrationStatus.ACTIVE);
       // reconnectSpy should have been called only once.
       expect(reconnectSpy).toBeCalledTimes(1);
-      expect(restoreSpy).toBeCalledOnceWith(reg.startKeepaliveTimer.name);
-      expect(restartSpy).toBeCalledOnceWith(reg.startKeepaliveTimer.name);
+      expect(restoreSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
+      expect(restartSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
       // Active Mobius URL should remain unchanged.
       expect(reg.getActiveMobiusUrl()).toStrictEqual(url);
     });
@@ -1422,7 +1423,7 @@ describe('Registration Tests', () => {
 
       // Verify that the keepalive timer was cleared and reconnectOnFailure was triggered
       expect(clearKeepaliveSpy).toHaveBeenCalled();
-      expect(reconnectSpy).toHaveBeenCalledWith(reg.startKeepaliveTimer.name);
+      expect(reconnectSpy).toHaveBeenCalledWith(RECONNECT_ON_FAILURE_UTIL);
 
       // Verify that the active Mobius URL has been updated to the backup server and registration is active
       expect(reg.getActiveMobiusUrl()).toEqual(mobiusUris.backup[0]);
@@ -1462,7 +1463,7 @@ describe('Registration Tests', () => {
       expect(lineEmitter).lastCalledWith(LINE_EVENTS.UNREGISTERED);
       expect(reg.keepaliveTimer).toStrictEqual(undefined);
       expect(reg.failbackTimer).toStrictEqual(undefined);
-      expect(reconnectSpy).toBeCalledOnceWith(KEEPALIVE_UTIL);
+      expect(reconnectSpy).toBeCalledOnceWith(RECONNECT_ON_FAILURE_UTIL);
       expect(restoreSpy).not.toBeCalled();
       expect(restartRegSpy).not.toBeCalled();
       expect(reg.reconnectPending).toStrictEqual(true);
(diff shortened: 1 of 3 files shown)


packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -207,9 +207,32 @@ export class Registration implements IRegistration {
    */
   private async restorePreviousRegistration(caller: string): Promise<boolean> {
     let abort = false;
-
+    // Testing:
+    // 4XX
+    // 429 big n small value
+    // 404
+    // 500
```
rarajes2: We can remove this comment or make it more verbose

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -207,9 +207,32 @@ export class Registration implements IRegistration {
...
+          this.backupMobiusUris.length > 0
+        ) {
+          // If we are using primary and got 429, switch to backup
+          abort = await this.attemptRegistrationWithServers(caller, this.backupMobiusUris);
+        } else {
+          // If we are using backup and got 429, restart registration
+          this.restartRegistration(caller);
+        }
+
+        return true;
```
rarajes2, line 232: Are we supposed to return `abort` here ? If not, assigning a value to abort won't make any difference in this logic

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -207,9 +207,32 @@ export class Registration implements IRegistration {
...
+        } else if (
+          this.primaryMobiusUris.includes(this.activeMobiusUrl) &&
+          this.backupMobiusUris.length > 0
+        ) {
+          // If we are using primary and got 429, switch to backup
+          abort = await this.attemptRegistrationWithServers(caller, this.backupMobiusUris);
+        } else {
+          // If we are using backup and got 429, restart registration
+          this.restartRegistration(caller);
+        }
```
rarajes2, line 229: Why are we adding this whole logic here? It seems bit confusing here, this whole logic could be moved to a different function named something like "handling429RegResponse" and then call this inside some catch block

Kesari3008: @rarajes2  We already have a callback function created for 429 handling and we are updating the retyr after value there. Why we cannot have separate function and handle retries separately is because we need to keep our retries consolidated.

If you see in this approach, we are not scheduling any retry as such but we are calling existing functions only and ensuring that retry after value is honored while invoking those functions. It is to avoid cases where mutliple retries are scheduled from different flows.

Let's discuss this over call if it's still confusing

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Wherever we are calling restore function, after that immediatety we are invoking restart which we wanna avopid hence we are returning true here in hardcoded manner but I am thinking if there is a better wway to handle this probably

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +727,110 @@ describe('Registration Tests', () => {
...
+      jest.clearAllMocks();
+      jest.useRealTimers();
+    });
+
+    it('should schedule retry when 429 with retry-after < 60 seconds during reconnect', async () => {
+      restartSpy = jest.spyOn(reg, 'restartRegistration');
+      reg.setActiveMobiusUrl(mobiusUris.primary[0]);
+      const failurePayload429Small = <WebexRequestPayload>(<unknown>{
+        statusCode: 429,
+        body: 'SOMETHING RANDOM',
```
Kesari3008: We can use proper caller name from wherever reconnectONFailure is being invoked

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +727,110 @@ describe('Registration Tests', () => {
...
+          'retry-after': 30,
+        },
+      });
+      webex.request
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload)
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload);
+
+      await reg.reconnectOnFailure('SOMETHING RANDOM'); // This call is being used to set the retry-after value
```
Kesari3008: Why do we have to invoke reconnectOnFailure twice ?

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +727,110 @@ describe('Registration Tests', () => {
...
+
+      await reg.reconnectOnFailure('SOMETHING RANDOM'); // This call is being used to trigger the retry
+      jest.advanceTimersByTime(40 * SEC_TO_MSEC_MFACTOR);
+      await flushPromises();
+
+      expect(restartSpy).toHaveBeenCalledTimes(1);
```
Kesari3008, line 766: We should have more expect statements as to check restore being invoked then restrat being invoked with retyr after value

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +727,110 @@ describe('Registration Tests', () => {
...
+
+      expect(attemptRegistrationWithServersSpy).toHaveBeenCalledTimes(2);
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(1, 'SOMETHING RANDOM', [
+        mobiusUris.backup[0],
+      ]);
+      // Immediately try primary servers when retry-after >= 60 seconds on backup
+      expect(restartSpy).toHaveBeenCalledTimes(1);
+      expect(restartSpy).toHaveBeenCalledWith('SOMETHING RANDOM');
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(2, 'SOMETHING RANDOM', [
+        mobiusUris.primary[0],
```
Kesari3008: We should add expect statement for all functions invocation in sequence, Restore being invoked, which fails with 429 then if it goes to restart from there or directly to attemmptRegistration or to failoverTmer function also

Kesari3008 (changes requested):

packages/calling/src/CallingClient/registration/register.ts
Shreyas281299: Removed

packages/calling/src/CallingClient/registration/register.ts
Shreyas281299: Since its just a boolean being returned Im not changing it right now. Please let me know if this change is required and we can discuss and make the change

packages/calling/src/CallingClient/registration/register.ts
Shreyas281299: Also, to make sure any existing flow is not broken because of our changes here

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: The issue is that retryAfter value get sets only after we have ran the 'restorePreviousRegistration' function. So the first time I call this method the retry after value is undefined and it gets set. And when I call it the second time retryAfter is valid and we actually invoke 'restorePreviousRegistration' method. Its just for testing - I ensured that we are actually testing the logic in 'restorePreviousRegistration', so I think this should be fine

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Updated

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Ive added more expects below

packages/calling/src/CallingClient/registration/register.ts
rarajes2: This logic seems confusing, if the change requires further discussion/refactoring then we should create a ticket for this.

packages/calling/src/CallingClient/registration/register.ts
rarajes2: Went through the code and I see that `restorePreviousRegistration` is being invoked inside the `handle429Retry` method which is a callback to the `handleRegistrationErrors`.

rarajes2 (approved):
Approving the PR with [this](https://github.com/webex/webex-js-sdk/pull/4559#discussion_r2510700628) comment. Let's discuss this among us.

packages/calling/src/CallingClient/registration/register.ts
rarajes2: I see that `abort` is being returned below at the end of the function. My bad

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,122 @@ describe('Registration Tests', () => {
...
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload);
+
+      await reg.reconnectOnFailure(TEST_RECONNECT_CALLER); // This call is being used to set the retry-after value
+      // Verify restore is invoked first and retry-after captured before scheduling
+      expect(restoreSpy).toBeCalledOnceWith(TEST_RECONNECT_CALLER);
+      expect(restartSpy).not.toBeCalled();
+      expect(reg.retryAfter).toEqual(30);
+
+      await reg.reconnectOnFailure(TEST_RECONNECT_CALLER); // This call is being used to trigger the retry
```
Kesari3008: Why are we invoking this again here, based on the code: attemptregistationWithServers will be invoked which will fail with 429 and code should automatically go invoking restartRegistration with timeout of smaller value of  retryaAfter

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -210,6 +211,25 @@ export class Registration implements IRegistration {
...
+
+        return true;
```
Kesari3008, line 232: Where are we resetting the retryAfter value ?

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,122 @@ describe('Registration Tests', () => {
...
+
+      expect(attemptRegistrationWithServersSpy).toHaveBeenCalledTimes(2);
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(1, TEST_RECONNECT_CALLER, [
+        mobiusUris.primary[0],
+      ]);
+      // Immediately try backup servers when retry-after >= 60 seconds on primary
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(
+        2,
+        TEST_RECONNECT_CALLER,
+        mobiusUris.backup
```
Kesari3008, line 802: You can add one more expect here to show that restart flow was not invoked here at all

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,122 @@ describe('Registration Tests', () => {
...
+          'retry-after': 100,
+        },
+      });
+      webex.request
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload)
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload);
+
+      await reg.reconnectOnFailure(TEST_RECONNECT_CALLER); // This call is being used to trigger the retry
```
Kesari3008: We should also invoke restore and restart flow as part of handleConnectionRestoration in some UT

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,122 @@ describe('Registration Tests', () => {
...
+        mobiusUris.primary[0],
+      ]);
+      // Immediately try backup servers when retry-after >= 60 seconds on primary
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(
+        2,
+        TEST_RECONNECT_CALLER,
+        mobiusUris.backup
+      );
+    });
+    it('should restart registration with primary if we get 429 while on backup', async () => {
```
Kesari3008, line 808: We should also have a testcase where device is registered with backup and we get 429 with smaller value

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -43,6 +44,7 @@ const MockServiceData = {
   indicator: ServiceIndicator.CALLING,
   domain: '',
 };
+const TEST_RECONNECT_CALLER = 'TEST_RECONNECT_CALLER';
```
Kesari3008: We can use the actual caller name instead of dummy one

Kesari3008 (changes requested):

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Removed this

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: This is not required, ive removed this and checked the tests

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Added the expect to ensure restart flow is not triggered

packages/calling/src/CallingClient/registration/register.ts
Shreyas281299: I forgot about that. Ive added that

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Added 2 more tests where I invoke handleConnectionRestoration and our new handling is triggered

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,206 @@ describe('Registration Tests', () => {
...
+        statusCode: 429,
+        body: RECONNECT_ON_FAILURE_UTIL,
+        headers: {
+          'retry-after': 30,
+        },
+      });
+      webex.request
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload)
+        .mockRejectedValueOnce(failurePayload429Small)
```
Kesari3008, line 754: here extra 2 mockRejectedValue and MockResolvedValue may not be required. In total we are calling attemptRegistrationWithServer 2 times after which we are return success payload. It should stop after that

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,206 @@ describe('Registration Tests', () => {
...
+        statusCode: 429,
+        body: RECONNECT_ON_FAILURE_UTIL,
+        headers: {
+          'retry-after': 100,
+        },
+      });
+      webex.request
+        .mockRejectedValueOnce(failurePayload429Small)
+        .mockResolvedValueOnce(successPayload)
+        .mockRejectedValueOnce(failurePayload429Small)
```
Kesari3008, line 782: Same comment here

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -727,6 +729,206 @@ describe('Registration Tests', () => {
...
+        [mobiusUris.primary[0]]
+      );
+      // Immediately try backup servers when retry-after >= 60 seconds on primary
+      expect(attemptRegistrationWithServersSpy).toHaveBeenNthCalledWith(
+        2,
+        RECONNECT_ON_FAILURE_UTIL,
+        mobiusUris.backup
+      );
+      expect(restartSpy).not.toHaveBeenCalledTimes(1);
+      expect(restartSpy).not.toHaveBeenCalledWith(RECONNECT_ON_FAILURE_UTIL);
```
Kesari3008, line 805: Add check for retyr after being reset here as well

packages/calling/src/CallingClient/registration/register.test.ts
```diff
@@ -1479,6 +1687,58 @@ describe('Registration Tests', () => {
         })
       );
     });
+
+    it('ensure retryAfter is set when 429 occurs during failover retry', async () => {
```
Kesari3008, line 1688: Why did we need these tests ?

Kesari3008 (approved):
Few small comments on test improvements

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: This is added for better coverage, we are only checking that retryAfter value is set to undefined, here we can confirming that it gets set tothe correct value.

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Added

packages/calling/src/CallingClient/registration/register.test.ts
Shreyas281299: Removed

PR #4578, opened 2025-11-19
fix(calling): fix the timer issue for call keepalive

rarajes2:
# COMPLETES #< Ad-hoc >

Fixes the timer issue for the handleCallEstablished function. 

This is continuation of the PR - #4557 

Clearing the interval in success case and rescheduling a new interval. Logic is bit altered to use `setTimeout` instead of `setInterval` now.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Tested the following with the burp suite (to update the response header)

- For 4xx cases the call was being cleaned up
- For 5xx cases with retry-after, it honors the value if provided else fallback to 30 sec retry interval
- Tries max of 4 retries after first failure and if all fails it stops the retries
- Failed 3 keepalives and then let 1 pass which did reset the counter and again letting the keepalive fail for 4 times stops sending further keepalive request

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -948,7 +948,7 @@ describe('State Machine handler tests', () => {
     call['callStateMachine'].state.value = 'S_SEND_CALL_CONNECT';
 
     webex.request.mockReturnValue(statusPayload);
-    jest.spyOn(global, 'setTimeout');
+    jest.spyOn(global, 'setInterval');
 
     const funcSpy = jest.spyOn(call, 'postStatus').mockResolvedValue(statusPayload);
     const logSpy = jest.spyOn(log, 'info');
@@ -962,12 +962,12 @@ describe('State Machine handler tests', () => {
      */
     await flushPromises(3);
 
-    expect(setTimeout).toHaveBeenCalledTimes(2);
-    expect(setTimeout).toHaveBeenLastCalledWith(expect.any(Function), DEFAULT_SESSION_TIMER);
+    expect(setInterval).toHaveBeenCalledTimes(1);
+    expect(setInterval).toHaveBeenCalledWith(expect.any(Function), DEFAULT_SESSION_TIMER);
     expect(funcSpy).toBeCalledTimes(1);
     expect(logSpy).toBeCalledWith('Session refresh successful', {
       file: 'call',
-      method: 'handleCallEstablished',
+      method: 'scheduleCallKeepaliveInterval',
     });
     expect(logSpy).toHaveBeenCalledWith(
       `${METHOD_START_MESSAGE} with: ${call.getCorrelationId()}`,
@@ -985,7 +985,7 @@ describe('State Machine handler tests', () => {
     });
 
     webex.request.mockReturnValue(statusPayload);
-    jest.spyOn(global, 'clearTimeout');
+    jest.spyOn(global, 'clearInterval');
 
     const emitSpy = jest.spyOn(call, 'emit');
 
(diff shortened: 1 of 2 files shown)


adhmenon (commented):
Have we tested with agent desktop? 
I am a bit worried about that case as any change in timer logic might cause issues on their side since AD isn't exactly a browser app...

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1524,15 +1524,19 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     if (this.sessionTimer) {
       log.log('Resetting session timer', loggerContext);
 
-      clearInterval(this.sessionTimer);
+      clearTimeout(this.sessionTimer);
     }
 
-    this.sessionTimer = setInterval(async () => {
+    this.sessionTimer = setTimeout(async () => {
```
Kesari3008: Let's discuss this once you are back. We should not remove setInterval itself in my opinion

rarajes2:
> Have we tested with agent desktop? 

Tried testing on the AD but there seems to be some breaking change in the desktop login flow with the latest calling sdk. Please verify once and let me know if you are able to perform the desktop login with the latest calling SDK.


> I am a bit worried about that case as any change in timer logic might cause issues on their side since AD isn't exactly a browser app...

Not sure what do you mean by `AD isn't exactly a browser app...` 

@adhmenon

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Sure, we can discuss this.

rarajes2: Updated as discussed.

rarajes2:
@adhmenon - Testing on the AD is not required for this.

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1524,78 +1604,10 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     if (this.sessionTimer) {
       log.log('Resetting session timer', loggerContext);
 
-      clearInterval(this.sessionTimer);
+      clearTimeout(this.sessionTimer);
```
Kesari3008: This should be clearInterval I believe, also we may not need this block at all now because we are never calling Established handled in recurring manner anymore

packages/calling/src/CallingClient/calling/call.test.ts
```diff
@@ -1068,7 +1069,14 @@ describe('State Machine handler tests', () => {
...
     expect(postStatusSpy).toHaveBeenCalledTimes(1);
-    expect(sendEvtSpy).toHaveBeenCalledWith({type: 'E_CALL_ESTABLISHED'});
+
+    // Now advance by 1 second for the retry-after interval
+    jest.advanceTimersByTime(1000);
+    await Promise.resolve();
+    await Promise.resolve();
+
+    expect(postStatusSpy).toHaveBeenCalledTimes(2);
+    expect(scheduleKeepaliveSpy).toHaveBeenCalled();
```
Kesari3008: How many times this has been called, add that please

Kesari3008 (approved):

Kesari3008:
Please add manual test results in the PR description. I would like to take a look

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Right, this is not needed anymore. Removed.

packages/calling/src/CallingClient/calling/call.test.ts
rarajes2: Done

adhmenon (approved):

PR #4579, opened 2025-11-19
fix(contact-center): fixed-outdial-fail-issue

adhmenon:
# COMPLETES #< AD-HOC >

* EPIC team reported an issue that was happening with the widgets.
* Whenever a user got the `AgentOutboundFailed` event, the widgets was not properly cleaning up the call and so the user was not able to do further outdials.
* Further triaging found that this was because the SDK was clearing the `task` object on receiving this event and hence subsequent wrapup was not being shown.

* Now we simply emit TASK_REJECT on receiving the AgentOutboundFailed event and on subsequent events from the backend, we end task and show wrapup.
* Code was also added to ensure that rejecting the outdial event clears up the task and older flows remain unaffected.
* Other task clear ups also work.

Vidcast (EXTENSION) - https://app.vidcast.io/share/604f2f25-55ea-4446-950b-68777ce8f6a3
Vidcast (BROWSER) - https://app.vidcast.io/share/25857bee-9aa8-4539-b411-5b4c4808d1d3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested outdial on both EXTENSION and BROWSER
* Tested for both working outdial and failing.
* Tested user rejecting outdial call.
* Tested basic call to see all flows worked.

  - [ ] Github Copilot
  - [x] Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/contact-center/src/services/task/TaskManager.ts
+++ b/packages/@webex/contact-center/src/services/task/TaskManager.ts
@@ -214,9 +214,9 @@ export default class TaskManager extends EventEmitter {
             LoggerProxy.log(`Agent outbound failed for task`, {
               module: TASK_MANAGER_FILE,
               method: METHODS.REGISTER_TASK_LISTENERS,
-              interactionId: payload.data?.interactionId,
+              interactionId: payload.data.interactionId,
             });
-            task.emit(TASK_EVENTS.TASK_REJECT, payload?.data?.reason);
+            task.emit(TASK_EVENTS.TASK_REJECT, payload.data.reason ?? 'UNKNOWN_REASON');
             break;
           case CC_EVENTS.AGENT_CONTACT_ASSIGNED:
             task = this.updateTaskData(task, payload.data);
@@ -560,9 +560,9 @@ export default class TaskManager extends EventEmitter {
       this.webCallingService.cleanUpCall();
     }
 
-    const isOutdial = task?.data?.interaction?.outboundType === 'OUTDIAL';
-    const isNew = task?.data?.interaction?.state === 'new';
-    const isTerminated = task?.data?.interaction?.isTerminated;
+    const isOutdial = task.data.interaction.outboundType === 'OUTDIAL';
+    const isNew = task.data.interaction.state === 'new';
+    const isTerminated = task.data.interaction.isTerminated;
 
     // For OUTDIAL: only remove if NOT terminated (user-declined, no wrap-up follows)
     // If terminated, keep task for wrap-up flow (CONTACT_ENDED → AGENT_WRAPUP)
--- a/packages/@webex/contact-center/src/services/task/TaskManager.ts
+++ b/packages/@webex/contact-center/src/services/task/TaskManager.ts
@@ -201,15 +201,22 @@ export default class TaskManager extends EventEmitter {
             this.emit(TASK_EVENTS.TASK_OFFER_CONTACT, task);
             break;
           case CC_EVENTS.AGENT_OUTBOUND_FAILED:
-            // We don't have to emit any event here since this will be result of promise.
-            if (task.data) {
-              this.removeTaskFromCollection(task);
-            }
+            task = this.updateTaskData(task, payload.data);
+            this.metricsManager.trackEvent(
+              METRIC_EVENT_NAMES.TASK_OUTDIAL_FAILED,
+              {
+                ...MetricsManager.getCommonTrackingFieldForAQMResponse(payload.data),
+                taskId: payload.data.interactionId,
+                reason: payload.data.reasonCode || payload.data.reason,
+              },
+              ['behavioral', 'operational']
+            );
             LoggerProxy.log(`Agent outbound failed for task`, {
               module: TASK_MANAGER_FILE,
               method: METHODS.REGISTER_TASK_LISTENERS,
               interactionId: payload.data?.interactionId,
             });
+            task.emit(TASK_EVENTS.TASK_REJECT, payload?.data?.reason);
             break;
           case CC_EVENTS.AGENT_CONTACT_ASSIGNED:
             task = this.updateTaskData(task, payload.data);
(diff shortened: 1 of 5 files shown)


packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -201,15 +201,22 @@ export default class TaskManager extends EventEmitter {
             this.emit(TASK_EVENTS.TASK_OFFER_CONTACT, task);
             break;
           case CC_EVENTS.AGENT_OUTBOUND_FAILED:
-            // We don't have to emit any event here since this will be result of promise.
-            if (task.data) {
-              this.removeTaskFromCollection(task);
```
adhmenon, line 206: This here was the root cause of the issue, have removed it - it was old code.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -201,15 +201,22 @@ export default class TaskManager extends EventEmitter {
...
-            }
+            task = this.updateTaskData(task, payload.data);
```
adhmenon, line 204: Here we simply update the task, post metrics and then emit the `TASK_REJECTED` event so users can handle it. Agent Desktop also does the same (shows a modal).
I preferred re-using this event as opposed to simply sending another event from the SDK.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -553,9 +560,15 @@ export default class TaskManager extends EventEmitter {
       this.webCallingService.cleanUpCall();
     }
 
-    if (task.data.interaction.state === 'new' || isSecondaryEpDnAgent(task.data.interaction)) {
-      // Only remove tasks in 'new' state or isSecondaryEpDnAgent immediately. For other states,
-      // retain tasks until they complete wrap-up, unless the task disconnected before being answered.
+    const isOutdial = task?.data?.interaction?.outboundType === 'OUTDIAL';
```
adhmenon: These are new conditions. Open to suggestions if I can do it it a better way but this is the cleanest way I saw.

Let me explain - basically we only need to clear the task for outdial cases where the user rejects it (i.e agent rejects it). In the case where we get `AgentOutboundFailed` - we **ALWAYS** get `ContactEnded` AND `AgentWrapup` in the same sequence. Hence once the user wraps up the task clears up automatically.
The change I made ensures these cases are handled with as minimal changes to the SDK flow. Otherwise we might need more checks and I wanted to do this as a pure SDK change and no widgets change needed.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -553,9 +560,15 @@ export default class TaskManager extends EventEmitter {
...
+    const isNew = task?.data?.interaction?.state === 'new';
+    const isTerminated = task?.data?.interaction?.isTerminated;
+
+    // For OUTDIAL: only remove if NOT terminated (user-declined, no wrap-up follows)
+    // If terminated, keep task for wrap-up flow (CONTACT_ENDED → AGENT_WRAPUP)
+    // For non-OUTDIAL: remove if state is 'new'
+    // Always remove if secondary EpDn agent
+    if ((isNew && !(isOutdial && isTerminated)) || isSecondaryEpDnAgent(task.data.interaction)) {
```
adhmenon, line 571: Another point - I've tested this with widgets against older flows and nothing breaks.
All old calls will follow the same flow and the check specifcially checks for both `isOutdial` and `isTerminated`.

packages/@webex/contact-center/test/unit/spec/services/task/TaskManager.ts
```diff
@@ -673,29 +673,100 @@ describe('TaskManager', () => {
     expect(taskUpdateTaskDataSpy).toHaveBeenCalledWith(payload.data);
   });
 
-  it('should remove currentTask from taskCollection on AGENT_OUTBOUND_FAILED event', () => {
+  it('should NOT remove OUTDIAL task from taskCollection on AGENT_OUTBOUND_FAILED when terminated (wrap-up flow)', () => {
```
adhmenon, line 676: Added some tests for checking the clearing of the task and also the case where we do not want to clear the task object.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -201,15 +201,22 @@ export default class TaskManager extends EventEmitter {
...
+                reason: payload.data.reasonCode || payload.data.reason,
+              },
+              ['behavioral', 'operational']
+            );
             LoggerProxy.log(`Agent outbound failed for task`, {
               module: TASK_MANAGER_FILE,
               method: METHODS.REGISTER_TASK_LISTENERS,
               interactionId: payload.data?.interactionId,
             });
+            task.emit(TASK_EVENTS.TASK_REJECT, payload?.data?.reason);
```
rsarika: Why TASK_REJECT why not a new event TASK_OUTDIAL_FAILED ?

rsarika (approved):

packages/@webex/contact-center/src/services/task/TaskManager.ts
adhmenon: So Agent Desktop also just shows a modal showing an error... it was a similar flow like RONA.
I can add a new event but I felt it would be too many events for the user to listen to...

Kesari3008: ? is not really needed after payload data as you can see in previous lines it's not used

Kesari3008: For failure case, Task_Reject doesn't seem like the appropriate event as it sends the wrong message to UI that task has been rejected. For cases where we are declining the outdial call or it goes to RONA, it is ok to use this but for cases where it fails while even attempting to answer is different right

packages/@webex/contact-center/src/services/task/TaskManager.ts
Kesari3008: Same comment here about ? after task and after data also

Kesari3008 (changes requested):

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -553,9 +560,15 @@ export default class TaskManager extends EventEmitter {
       this.webCallingService.cleanUpCall();
     }
 
-    if (task.data.interaction.state === 'new' || isSecondaryEpDnAgent(task.data.interaction)) {
-      // Only remove tasks in 'new' state or isSecondaryEpDnAgent immediately. For other states,
-      // retain tasks until they complete wrap-up, unless the task disconnected before being answered.
+    const isOutdial = task?.data?.interaction?.outboundType === 'OUTDIAL';
+    const isNew = task?.data?.interaction?.state === 'new';
+    const isTerminated = task?.data?.interaction?.isTerminated;
```
Kesari3008: Why are we checking for Outdial and isTerminated specifically ?

packages/@webex/contact-center/src/services/task/TaskManager.ts
adhmenon: Updated and added new `TASK_OUTBOUND_FAILED` event.
Also added it in the sample app.

Kesari3008 (approved):

PR #4542, opened 2025-10-16
fix(plugin-meetings): saving subnet ip details in reachability with p…

venky-mediboina:
# COMPLETES #[SPARK-610518](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610518)

## This pull request addresses

Saving the subnet ip details which includes subnet IPs, latency, etc..

## by making the following changes

Saved the subnet ip details in reachability results and sending to web-client, implements each UDP URL checks for Reachability.

### Change Type

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios were tested

< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

## The GAI Coding Policy And Copyright Annotation Best Practices ##
 

 
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [ ] GAI was used to create a draft that was subsequently customized or modified
- [ ] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [ ] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [X] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

- [X] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [X] I discussed changes with code owners prior to submitting this pull request
- [X] I have not skipped any automated checks
- [X] All existing and new tests passed
- [x] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #4580, opened 2025-11-20
fix(plugin-meetings): moved cluster reachability functionality to new…

venky-mediboina:
… class pr1

# COMPLETES # [SPARK-610518](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610518)

As changes are somewhat big, we split into 3 PRs. Old combined PR: https://github.com/webex/webex-js-sdk/pull/4542.

This is the first PR of 3 PRs to add subnet details which consists of each subnet IP in the details section including whether its reachable or not. 

As part of this PR-1, created new class(ReachabilityPeerConnection) to handle each cluster as instance in ClusterReachabilty class for now.

```
PR-1: Create ReachabilityPeerConnection class and refactor existing code in ClusterReachability to use it - there should be no logic change at all, no behavioral change, the code should do exactly all the same as it currently does, but ClusterReachability should not have any code that deals with PeerConnection APIs or candidates. It should just create 1 instance of ReachabilityPeerConnection and ReachabilityPeerConnection should do all that work.

PR-2: Add support for per URL checks for UDP- in this PR we'll introduce the new config flag and when it's set, ClusterReachability will pass only TCP and TLS URLs to its single ReachabilityPeerConnection instance and will create additional array of ReachabilityPeerConnection instances - one for each UDP URL.

PR-3: Add all the extra stuff like getAllClustersInfo() API, populating the details and sending them to the backend etc.
```

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -6,7 +6,56 @@ import {convertStunUrlToTurn, convertStunUrlToTurnTls} from './util';
...
+ * @param {string|null} [serverIp] - Server IP address (subnet)
+ * @param {Set<string>} [reachedSubnets] - Optional set to track reached subnets
+ * @returns {boolean} true if a new IP was added, false otherwise
+ */
+function processIceCandidateResult(
+  result: TransportResult,
+  latency: number,
+  publicIp?: string | null,
+  serverIp?: string | null,
+  reachedSubnets?: Set<string>
```
marcin-bazyl: why is `reachedSubnets` optional? this function is called from only 1 place and with `reachedSubnets` always provided.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -6,7 +6,56 @@ import {convertStunUrlToTurn, convertStunUrlToTurnTls} from './util';
...
+ * Handles IP deduplication and latency tracking.
+ *
+ * @param {TransportResult} result - The protocol result object to update (e.g., result.udp)
+ * @param {number} latency - Latency in milliseconds
+ * @param {string|null} [publicIp] - Public IP address from ICE candidate
+ * @param {string|null} [serverIp] - Server IP address (subnet)
+ * @param {Set<string>} [reachedSubnets] - Optional set to track reached subnets
+ * @returns {boolean} true if a new IP was added, false otherwise
+ */
+function processIceCandidateResult(
```
marcin-bazyl: Some questions about this:

1. I don't really understand the purpose of the refactor of existing ClusterReachability.saveResult() into the new ReachabilityPeerConnection.saveResult() and the free function processIceCandidateResult(). Can you explain the reasoning behind it?
2. Why is processIceCandidateResult() not just a method inside ReachabilityPeerConnection?
3. Why ReachabilityPeerConnection has reachedSubnets as a property? it's being added to, but never read from.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -413,21 +422,149 @@ export class ClusterReachability extends EventsScope {
...
+        result: 'untested',
+      },
+      xtls: {
+        result: 'untested',
+      },
+    };
+
+    this.reachabilityPeerConnection = new ReachabilityPeerConnection(clusterInfo, name);
+
+    this.reachabilityPeerConnection.on('resultReady', (data) => {
```
marcin-bazyl: probably would be better to have an enum for all the ReachabilityPeerConnection events

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -413,21 +422,149 @@ export class ClusterReachability extends EventsScope {
...
+ * A class that handles reachability checks for a single cluster.
+ * Creates and orchestrates a ReachabilityPeerConnection instance.
+ * Listens to events and emits them to consumers.
+ */
+export class ClusterReachability extends EventsScope {
+  private reachabilityPeerConnection?: ReachabilityPeerConnection;
+  public readonly isVideoMesh: boolean;
+  public readonly name;
+  public readonly reachedSubnets: Set<string> = new Set();
+  private result: ClusterReachabilityResult;
```
marcin-bazyl: I don't think we need the `result` in `ClusterReachability` now. We have it inside `ReachabilityPeerConnection` and for now we only have a single instance of `ReachabilityPeerConnection` in every `ClusterReachability`, so there is no need to have a duplication of the result in the 2 classes. In one of your future PRs, maybe we'll have to introduce the result in `ClusterReachability` so that it will aggregate the results from all `ReachabilityPeerConnection` instances used, but right now `ClusterReachability.getResult()` can just return the result from `this.reachabilityPeerConnection`, right?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -35,31 +84,29 @@ export const Events = {
...
- * It emits events from Events enum
+ * Handles RTCPeerConnection lifecycle and ICE candidate gathering for reachability checks.
+ * Does ALL the work: PeerConnection lifecycle, candidate processing, result management, and event emission.
  */
-export class ClusterReachability extends EventsScope {
-  private numUdpUrls: number;
-  private numTcpUrls: number;
-  private numXTlsUrls: number;
-  private result: ClusterReachabilityResult;
+class ReachabilityPeerConnection extends EventsScope {
```
marcin-bazyl: we usually try to have a single class in each file, so it's probably better to move this new class into a new file

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -162,10 +159,6 @@ describe('ClusterReachability', () => {
       assert.calledOnceWithExactly(fakePeerConnection.createOffer, {offerToReceiveAudio: true});
       assert.calledOnce(fakePeerConnection.setLocalDescription);
 
-      // Make sure that gatherIceCandidates is called before setLocalDescription
-      // as setLocalDescription triggers the ICE gathering process
-      assert.isTrue(gatherIceCandidatesSpy.calledBefore(fakePeerConnection.setLocalDescription));
```
marcin-bazyl, line 167: why has this check been removed?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Used enum for ReachabilityPeerConnection events.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Yeah, we can return the `this.reachabilityPeerConnection`'s result in `clusterReachability`. Avoided using results inside `ClusterReachability`.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Created new file and moved `ReachabilityPeerConnection` class into new file.

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: Incorrectly assumed delegation tests were sufficient, so removed, added back now,

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Removed the function as we don't need as of now.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Added this function to process the result, may be for future PR-2 with more usage. No reason for adding outside as its free function. But for now removed and consolidated inside `saveResult()` function itself.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+      if (!this.emittedSubnets.has(serverIp)) {
+        this.emittedSubnets.add(serverIp);
+        this.emit(
+          {
+            file: 'reachabilityPeerConnection',
+            function: 'saveResult',
+          },
+          ReachabilityPeerConnectionEvents.reachedSubnets,
+          {
+            subnets: [serverIp],
```
marcin-bazyl, line 279: why are we sending this as an array here? are we expecting to have multiple `serverIp`s sent with this event in the future?

venky-mediboina: yes, for now single IP in array. but for next PR, it will be batch of multiple IPs. We can also send single serverIp string, but as its minor change for PR-2, emitting as array.

marcin-bazyl (approved):

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -60,374 +55,74 @@ export class ClusterReachability extends EventsScope {
...
-    });
-
-    const turnTlsIceServers = cluster.xtls.map((urlString: string) => {
-      return {
-        username: 'webexturnreachuser',
-        credential: 'webexturnreachpwd',
-        urls: [convertStunUrlToTurnTls(urlString)],
-      };
-    });
+    this.reachabilityPeerConnection = new ReachabilityPeerConnection(clusterInfo, name);
```
rarajes2: nitpick: Param order can be same as this constructor.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -60,374 +55,74 @@ export class ClusterReachability extends EventsScope {
...
-
-      if (ipAdded)
+    this.reachabilityPeerConnection.on(
+      ReachabilityPeerConnectionEvents.clientMediaIpsUpdated,
+      (data) => {
         this.emit(
           {
             file: 'clusterReachability',
-            function: 'addPublicIP',
+            function: 'onClientMediaIpsUpdated',
```
rarajes2: Does this function name represent an action or is it a literal function name ? If it's the name of the function emitting this event, I don't see the function name anywhere in the diff. Applicable everywhere in this file

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -60,374 +55,74 @@ export class ClusterReachability extends EventsScope {
...
-              file: 'clusterReachability',
-              function: 'determineNatType',
-            },
-            Events.natTypeUpdated,
-            {
-              natType: NatType.SymmetricNat,
-            }
-          );
-        }
+    this.reachabilityPeerConnection.on(ReachabilityPeerConnectionEvents.reachedSubnets, (data) => {
```
rarajes2, line 105: nitpick: All the listeners could be grouped inside a separate method and called once inside the constructor. 

Question: Do we also need to un-register these listener for some cases ?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+ * Events emitted by ReachabilityPeerConnection
+ */
+export const ReachabilityPeerConnectionEvents = {
+  resultReady: 'resultReady', // emitted when successfully reached over a protocol
+  clientMediaIpsUpdated: 'clientMediaIpsUpdated', // emitted when new public IPs are found
+  natTypeUpdated: 'natTypeUpdated', // emitted when NAT type is determined
+  reachedSubnets: 'reachedSubnets', // emitted when server IP (subnet) is discovered
+} as const;
+
+export type ReachabilityPeerConnectionEvents = Enum<typeof ReachabilityPeerConnectionEvents>;
```
rarajes2: nitpick: Can move these to respective constant and type files

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+      urls: [url],
+    }));
+
+    // STUN servers are contacted only using UDP, so in order to test TCP reachability
+    // we pretend that Linus is a TURN server, because we can explicitly say "transport=tcp" in TURN urls.
+    // We then check for relay candidates to know if TURN-TCP worked (see registerIceCandidateListener()).
+    const tcpIceServers = cluster.tcp.map((urlString: string) => {
+      return {
+        username: 'webexturnreachuser',
+        credential: 'webexturnreachpwd',
```
rarajes2, line 86: Are we supposed to hardcode the credentials here ?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+      const peerConnection = new RTCPeerConnection(config);
+
+      return peerConnection;
+    } catch (peerConnectionError) {
+      LoggerProxy.logger.warn(
+        `Reachability:ReachabilityPeerConnection#createPeerConnection --> Error creating peerConnection:`,
+        peerConnectionError
+      );
+
+      return undefined;
```
rarajes2: nitpick: should we return `null` instead of `undefined`?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+    }
+  }
+
+  /**
+   * Adds public IP (client media IPs)
+   * @param {string} protocol
+   * @param {string} publicIp
+   * @returns {void}
+   */
+  private addPublicIp(protocol: 'udp' | 'tcp' | 'xtls', publicIp?: string | null) {
```
rarajes2: Should we create a type for `'udp' | 'tcp' | 'xtls'` ?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+    const result = this.result[protocol];
+
+    if (publicIp) {
```
rarajes2: nitpick: Can use early-return pattern, will reduce the indentation.

```js
if (!publicIp) return;

...

```

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+   * @param {number} latency
+   * @param {string|null} [publicIp]
+   * @param {string|null} [serverIp]
+   * @returns {void}
+   */
+  private saveResult(
+    protocol: 'udp' | 'tcp' | 'xtls',
+    latency: number,
+    publicIp?: string | null,
+    serverIp?: string | null
```
rarajes2, line 239: Do we need to use both `undefined` and `null` in the type ?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+        );
+      }
+    }
+  }
+
+  /**
+   * Determines NAT Type.
+   * @param {RTCIceCandidate} candidate
+   * @returns {void}
+   */
```
rarajes2, line 290: Description could be a bit verbose

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+  private determineNatType(candidate: RTCIceCandidate) {
+    this.srflxIceCandidates.push(candidate);
```
rarajes2, line 292: Question: How are we determining the passed candidate is server reflexive ? 

If the method `determingNatType` is specifically for the `srflx` type, maybe we can rename it to `determineNatTypeForSrflxCandidate` or something similar

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+
+      this.srflxIceCandidates.forEach((c) => {
+        const key = `${c.address}:${c.relatedPort}`;
+        if (!portsFound[key]) {
+          portsFound[key] = new Set();
+        }
+        portsFound[key].add(c.port);
+      });
+
+      Object.entries(portsFound).forEach(([, ports]) => {
```
rarajes2, line 305: nitpick: Can use `for-in` loop instead

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+   *
+   * @returns {void}
+   */
+  private registerIceCandidateListener() {
+    this.pc.onicecandidate = (e) => {
+      const TURN_TLS_PORT = 443;
+      const CANDIDATE_TYPES = {
+        SERVER_REFLEXIVE: 'srflx',
+        RELAY: 'relay',
+      };
```
rarajes2, line 334: Can be moved out from the listener callback, better to a constant file

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+
+      const latencyInMilliseconds = this.getElapsedTime();
+
+      if (e.candidate) {
+        if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
+          let serverIp = null;
+          if ('url' in e.candidate) {
+            const stunServerUrlRegex = /stun:([\d.]+):\d+/;
```
rarajes2, line 342: This regex could be re-used, if possible move it out to constant file

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+
+            const match = (e.candidate as any).url.match(stunServerUrlRegex);
+            if (match) {
+              // eslint-disable-next-line prefer-destructuring
+              serverIp = match[1];
+            }
```
rarajes2: nitpick:

```js
serverIp = match && match[1];
```

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -0,0 +1,425 @@
...
+            {
+              file: 'reachabilityPeerConnection',
+              function: 'determineNatType',
+            },
+            ReachabilityPeerConnectionEvents.natTypeUpdated,
+            {
+              natType: NatType.SymmetricNat,
+            }
+          );
+        }
```
rarajes2, line 318: What about the other type? I don't see `else` case

rarajes2 (commented):
**Question**: We are performing all the WebRTC stuff such as creating offer, setting answer and gathering ice candidates just to ensure the `reachability`, these offer/answer/candidates are not being used to establish the media connection, right ? 

Actual webrtc connection is established using the `media-core`. Correct me if my understanding is wrong.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
marcin-bazyl: there is no other type that we care about, we only have a type for "unknown" which is the default and we try to detect the symmetric nat type, if we do then we emit the event, we don't want to emit the event for "unknown". BTW, this functionality is used purely for logging and metrics.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
marcin-bazyl: yes, this is fine, the credentials here are not for security, they are just required by browser when using TURN, but in our case the TURN allocate requests don't allocate any resources on the server, backend only pretends to be a TURN server

marcin-bazyl:
> **Question**: We are performing all the WebRTC stuff such as creating offer, setting answer and gathering ice candidates just to ensure the `reachability`, these offer/answer/candidates are not being used to establish the media connection, right ?
> 
> Actual webrtc connection is established using the `media-core`. Correct me if my understanding is wrong.

yes, that's correct

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
marcin-bazyl: these are used only here, there is no need for them to be exposed more broadly

BTW, just to explain the context of these changes a bit better: @venky-mediboina is trying to add a feature required for network checker to list all server IPs that have been successfully reached. In order to do this he needs to create a separate RTCPeerConnection for each UDP url being tested, but he can still use a single RTCPeerConnection for all TCP and TLS urls together. So, he needs to re-use all the RTCPeerConnection stuff, that's why he's moving this into a new class - the ReachabilityPeerConnection so that ClusterReachability class will be able to instantiate either just 1 instance of ReachabilityPeerConnection for most cases (web app and all other SDK clients) and many instances (1 for each UDP url plus 1 for the rest of all TCP/TLS urls together) when network checker uses SDK - this will be controlled through SDK config. This is a bigger set of changes, so I've asked him to split it into a series of smaller PRs - this PR is the first one and its purpose is to just extract the existing code into the new class. Therefore, it's better if he just moves the code with minimum changes and avoids various small refactors like the one proposed in this comment and some other comments - this is so that it's easier to review the PR and to be sure that we're not actually changing and not breaking anything, we just want to move the code.

rarajes2 (approved):

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: As of now we are using only here, so let's keep using here directly. Yeah, maybe we might use at multiple places in next PRs. If yes, will take from constants file.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Tried to use `for-in`, but eslint giving errors. So, reverted to use same `Object.entries()`.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Modified.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Moved.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: created a separate function and used.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: It's an action, changed to use function containing all actions.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: Changed.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Modified to much verbose.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Used above early-return.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Created type and used.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Modified to use `null` and changed accordingly.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Changed function name.

We're checking e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE (which is 'srflx'). This is the WebRTC API's built-in property that tells us the candidate type.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: We can remove optional, but this is consistent throughout the SDK and is the preferred approach over using default values like = null. For now, keeping same as earlier.

PR #4583, opened 2025-11-21
fix(meetings): initial support for Locus hash trees (for 5k webinars)

marcin-bazyl:
# COMPLETES #[SPARK-697765](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-697765)

## This pull request addresses
For 5k webinars Locus is completely changing the way they signal Locus DTO changes to the clients. It will now be done using hash trees. More details here:
https://sqbu-github.cisco.com/pages/WebExSquared/locus/guides/convergedWebinar5k.html
NOTE: that Locus wiki is not fully up-to-date, there are various aspects of Locus API that are still being discussed and modified.

## by making the following changes
This PR is part 1 of 2. The whole set of changes for supporting Locus hash trees is available here:
https://github.com/webex/webex-js-sdk/pull/4427
and that can be tested using co.dmz.webex.com - it allows you to join 5k webinars, receive all the updates etc. but that PR is quite big and doesn't have all the necessary unit tests, with the tests it would be even bigger. So I've decided to split it into 2 parts:
1. Modifications to existing SDK code to connect it up with the new HashTreeParser
2. New code for hash trees - implementation of HashTreeParser and HashTree classes etc.

This PR is the part 1 of above list. It contains code copied from PR#4427 plus more unit tests for that code. The HashTreeParser remains empty in this PR. The full implementation of HashTreeParser and hash trees themselves will be in the next PR.

Here's a wiki explaining the design behind the changes in more detail:
https://confluence-eng-gpk2.cisco.com/conf/x/lo_uLQ

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios where tested

unit tests, manual e2e testing with the web app on co.dmz.webex.com (done with code from PR#4427)

## The GAI Coding Policy And Copyright Annotation Best Practices ##
  
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [ ] GAI was used to create a draft that was subsequently customized or modified
- [x] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [ ] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [x] Github Copilot
  - [ ] Other - Please Specify
[...]

PR #4427, opened 2025-08-06
feat(meetings): support for Locus hash trees

marcin-bazyl:
# COMPLETES #[SPARK-697765](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-697765)

## This pull request addresses
Adds support for Locus Hash Trees (needed for Webinar 5K)

This PR contains changes from a branch that was used for integration with Locus during the workshop. It's not meant to be merged to next as is, hence I've kept it as a draft. For actual merging into master it has been split into 2 PRs
1. https://github.com/webex/webex-js-sdk/pull/4583
2. https://github.com/webex/webex-js-sdk/pull/4587

These 2 PRs have same code changes as this PR, but extra unit tests added

PR #4587, opened 2025-11-24
feat(meetings): Hash trees implementation for Locus syncs

marcin-bazyl:
# COMPLETES #[SPARK-697765](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-697765)

Locus is using hash trees for sending DTO updates in 5k webinars.

The work for making web client work with Locus hash tree DTOs was done on a branch visible in this PR:
https://github.com/webex/webex-js-sdk/pull/4427

In order to merge these changes into next branch, above changes were split into 2 PRs:
1. https://github.com/webex/webex-js-sdk/pull/4583 - focuses on connecting the existing SDK LocusInfo code with new HashTreeParser
2. This PR - implements HashTreeParser and HashTree classes

So, this PR depends on PR#4583 and in terms of production code, it has the same code as PR#4427, but has more unit tests.

HashTree class implementation was taken from Colin's initial implementation done a few months ago and modified to use a different library for the hashing algorithm plus some bug fixes done.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, e2e manual testing of code from PR#4427 with Locus and web app

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
+++ b/packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
@@ -1,0 +1,424 @@
+import {XXHash128} from 'xxhash-addon';
+import {EMPTY_HASH} from './constants';
+
+type LeafDataItem = {
+  type: string;
+  id: number;
+  version: number;
+};
+
+/**
+ * HashTree is a data structure that organizes items into leaves based on their IDs,
+ */
+class HashTree {
+  leaves: Array<Record<string, Record<number, LeafDataItem>>>;
+
+  leafHashes: Array<string>;
+  readonly numLeaves: number;
+
+  /**
+   * Constructs a new HashTree.
+   * @param {LeafDataItem[]} leafData Initial data to populate the tree.
+   * @param {number} numLeaves The number of leaf nodes in the tree. Must be 0 or a power of 2.
+   * @throws {Error} If numLeaves is not 0 or a power of 2.
+   */
+  constructor(leafData: LeafDataItem[], numLeaves: number) {
+    // eslint-disable-next-line no-bitwise
+    if ((numLeaves & (numLeaves - 1)) !== 0) {
+      throw new Error(`Number of leaves must be a power of 2, saw ${numLeaves}`);
+    }
+
+    this.numLeaves = numLeaves;
+    this.leafHashes = new Array(numLeaves).fill(EMPTY_HASH);
+
+    this.leaves = new Array(numLeaves).fill(null).map(() => {
+      return {};
+    });
+
+    if (leafData) {
+      this.putItems(leafData);
+    }
+  }
+
+  /**
+   * Internal logic for adding or updating an item, without computing the leaf hash.
+   * @param {LeafDataItem} item The item to add or update.
+   * @returns {{put: boolean, index: (number|null)}} Object indicating if put and the leaf index.
+   * @private
+   */
+  private _putItemInternal(item: LeafDataItem): {put: boolean; index: number | null} {
+    if (this.numLeaves === 0) {
+      return {put: false, index: null}; // Cannot add to a tree with 0 leaves
+    }
+
+    const index = item.id % this.numLeaves;
+
+    if (!this.leaves[index][item.type]) {
+      this.leaves[index][item.type] = {};
+    }
+
+    const existingItem = this.leaves[index][item.type][item.id];
+
+    if (!existingItem || existingItem.version < item.version) {
+      this.leaves[index][item.type][item.id] = item;
+
+      return {put: true, index};
+    }
+
+    return {put: false, index: null};
+  }
+
+  /**
+   * Adds or updates a single item in the hash tree.
+   * @param {LeafDataItem} item The item to add or update.
+   * @returns {boolean} True if the item was added or updated, false otherwise (e.g., older version or tree has 0 leaves).
+   */
+  putItem(item: LeafDataItem): boolean {
+    const {put, index} = this._putItemInternal(item);
+
+    if (put && index !== null) {
+      this.computeLeafHash(index);
+    }
+
+    return put;
+  }
+
+  /**
+   * Adds or updates multiple items in the hash tree.
+   * @param {LeafDataItem[]} items The array of items to add or update.
+   * @returns {boolean[]} An array of booleans indicating success for each item.
+   */
+  putItems(items: LeafDataItem[]): boolean[] {
+    if (this.numLeaves === 0 && items.length > 0) {
+      // Cannot add items to a tree with 0 leaves.
+      return items.map(() => false);
+    }
+    const results: boolean[] = [];
+    const changedLeafIndexes = new Set<number>();
+
+    items.forEach((item) => {
+      const {put, index} = this._putItemInternal(item);
+      results.push(put);
+      if (put && index !== null) {
+        changedLeafIndexes.add(index);
+      }
+    });
+
+    changedLeafIndexes.forEach((index) => {
+      this.computeLeafHash(index);
+    });
+
+    return results;
+  }
+
+  /**
+   * Internal logic for removing an item, without computing the leaf hash.
+   * @param {LeafDataItem} item The item to remove.
+   * @returns {{removed: boolean, index: (number|null)}} Object indicating if removed and the leaf index.
+   * @private
+   */
+  private _removeItemInternal(item: LeafDataItem): {removed: boolean; index: number | null} {
+    if (this.numLeaves === 0) {
+      return {removed: false, index: null};
+    }
+
+    const index = item.id % this.numLeaves;
+
+    if (
+      this.leaves[index] &&
+      this.leaves[index][item.type] &&
+      this.leaves[index][item.type][item.id]
+    ) {
+      const existingItem = this.leaves[index][item.type][item.id];
+      if (
+        existingItem.id === item.id &&
+        existingItem.type === item.type &&
+        existingItem.version <= item.version
+      ) {
+        delete this.leaves[index][item.type][item.id];
+        if (Object.keys(this.leaves[index][item.type]).length === 0) {
+          delete this.leaves[index][item.type];
+        }
+
+        return {removed: true, index};
+      }
+    }
+
+    return {removed: false, index: null};
+  }
+
+  /**
+   * Removes a single item from the hash tree.
+   * The removal is based on matching type, id, and the provided item's version
+   * being greater than or equal to the existing item's version.
+   * @param {LeafDataItem} item The item to remove.
+   * @returns {boolean} True if the item was removed, false otherwise.
+   */
+  removeItem(item: LeafDataItem): boolean {
+    const {removed, index} = this._removeItemInternal(item);
+
+    if (removed && index !== null) {
+      this.computeLeafHash(index);
+    }
+
+    return removed;
+  }
+
+  /**
+   * Removes multiple items from the hash tree.
+   * @param {LeafDataItem[]} items The array of items to remove.
+   * @returns {boolean[]} An array of booleans indicating success for each item.
+   */
+  removeItems(items: LeafDataItem[]): boolean[] {
+    if (this.numLeaves === 0 && items.length > 0) {
+      return items.map(() => false);
+    }
+
+    const results: boolean[] = [];
+    const changedLeafIndexes = new Set<number>();
+
+    items.forEach((item) => {
+      const {removed, index} = this._removeItemInternal(item);
+      results.push(removed);
+      if (removed && index !== null) {
+        changedLeafIndexes.add(index);
+      }
+    });
+
+    changedLeafIndexes.forEach((index) => {
+      this.computeLeafHash(index);
+    });
+
+    return results;
+  }
+
+  /**
+   * Computes the hash for a specific leaf.
+   * The hash is based on the sorted items within the leaf.
+   * @param {number} index The index of the leaf to compute the hash for.
+   * @returns {void}
+   */
+  computeLeafHash(index: number) {
+    const leafContent = this.leaves[index];
+
+    // create a hasher
+    const hasher = new XXHash128(Buffer.from([0, 0, 0, 0]));
+
+    // iterate through the item types lexicographically
+    const itemTypes = Object.keys(leafContent).sort();
+    itemTypes.forEach((type) => {
+      // iterate through the items of this type in ascending order of ID
+      const items = Object.values(leafContent[type]).sort(
+        (a: LeafDataItem, b: LeafDataItem) => a.id - b.id
+      );
+
+      // add all the items id and version to the hasher
+      items.forEach((item: LeafDataItem) => {
+        const idBuffer = Buffer.alloc(8);
+        idBuffer.writeBigInt64LE(BigInt(item.id));
+
+        const versionBuffer = Buffer.alloc(8);
+        versionBuffer.writeBigInt64LE(BigInt(item.version));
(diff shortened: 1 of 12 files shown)


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -478,12 +480,19 @@ export default class LocusInfo extends EventsScope {
...
+        );
+        // continuing as we can still manage without responseBody.dataSets, but this is very suspicious
+      }
       LoggerProxy.logger.info(
-        'Locus-info:index#handleLocusAPIResponse: skipping handling of API http response with hashTreeParser'
+        'Locus-info:index#handleLocusAPIResponse --> passing Locus API response to HashTreeParser: ',
+        responseBody
       );
+      // update the data in our hash trees
+      this.hashTreeParser.handleLocusUpdate(responseBody);
```
mickelr, line 495: Is that necessary to wrap the function call in try/catch to avoid potentially breaking?

marcin-bazyl: hmm I'm not sure.... the code inside HashTreeParser.handleLocusUpdate() is not supposed to throw in any expected edge case and the existing code for non-hash trees also doesn't catch any exceptions here, so feels like any exception that happens should be just thrown and caught higher up in the code that actually invokes the Locus API for which the response is parsed here. Only that code in the upper layer will know really the full context and what API was called so it can decide what to do and how important the failure is. If I was to catch the exception here, I don't see what I could do with it, all I could do is log it and I would still probably have to re-throw it.

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
```diff
@@ -0,0 +1,459 @@
...
+
+  /**
+   * Constructs a new HashTree.
+   * @param {LeafDataItem[]} leafData Initial data to populate the tree.
+   * @param {number} numLeaves The number of leaf nodes in the tree. Must be 0 or a power of 2.
+   * @throws {Error} If numLeaves is not 0 or a power of 2.
+   */
+  constructor(leafData: LeafDataItem[], numLeaves: number) {
+    // eslint-disable-next-line no-bitwise
+    if ((numLeaves & (numLeaves - 1)) !== 0) {
```
jor-row, line 28: Nice simple check!

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTree.ts
```diff
@@ -0,0 +1,449 @@
...
+      
+      const diff1_2 = tree1.diffHashes(tree2.getHashes());
+      expect(diff1_2).to.deep.equal([1]); // Leaf 1 should differ
+
+      const diff2_1 = tree2.diffHashes(tree1Hashes);
+      expect(diff2_1).to.deep.equal([1]);
+    });
+  });
+
+  describe('computeTreeHashes', () => {
```
jor-row, line 590: a test case for numLeaves = 0 could be added here

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
```diff
@@ -0,0 +1,459 @@
...
+    return results;
+  }
+
+  /**
+   * Computes the hash for a specific leaf.
+   * The hash is based on the sorted items within the leaf.
+   * @param {number} index The index of the leaf to compute the hash for.
+   * @returns {void}
+   */
+  computeLeafHash(index: number) {
```
jor-row, line 240: Would it be worth checking that the index is a valid lead index at the start of this function?

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
```diff
@@ -0,0 +1,459 @@
...
+    changedLeafIndexes.forEach((index) => {
+      this.computeLeafHash(index);
+    });
+
+    return results;
+  }
+
+  /**
+   * Updates multiple items in the hash tree.
+   * This method can handle both adding and removing items based on the `operation` flag.
```
jor-row: I think this comment should be "both updating and removing"

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
```diff
@@ -0,0 +1,459 @@
...
+   *
+   * @param {object[]} itemUpdates An array of objects containing `operation` flag and the `item` to update.
+   * @returns {boolean[]} An array of booleans indicating success for each operation.
+   */
+  updateItems(itemUpdates: {operation: 'update' | 'remove'; item: LeafDataItem}[]): boolean[] {
```
jor-row, line 203: Unit tests for this function would be great since it is stubbed out in the tests for hashTreeParser

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
```diff
@@ -0,0 +1,459 @@
...
+   */
+  getLeafCount(): number {
+    return this.numLeaves;
+  }
+
+  /**
+   * Calculates the total number of items stored in the tree.
+   * @returns {number} The total number of items.
+   */
+  getTotalItemCount(): number {
```
jor-row, line 342: I know this is used in the unit tests quite a bit, but do you think it be worth adding some explicit/"dumb" ```getTotalItemCount``` unit tests?

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTree.ts
jor-row, line 1: Great test coverage!

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -117,7 +243,195 @@ class HashTreeParser {
...
+          `HashTreeParser#initializeDataSets --> ${this.debugId} initializing dataset "${name}" (${debugText})`
+        );
+
+        this.dataSets[name] = {
+          ...dataSet,
+        };
+      } else {
+        LoggerProxy.logger.info(
+          `HashTreeParser#initializeDataSets --> ${this.debugId} dataset "${name}" already exists (${debugText})`
+        );
```
jor-row, line 287: question: just checking my understanding here

so if a dataSet already exists, then a hash tree for that dataSet will already exist as well and will have it's own handling of getting the most up to date information so that is why we do nothing here? There is no need to send an update/sync request to locus because of that

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -117,7 +243,195 @@ class HashTreeParser {
...
+      if (Array.isArray(currentLocusPart)) {
+        for (const item of currentLocusPart) {
+          findAndStoreMetaData(item);
+        }
+      } else {
+        for (const key of Object.keys(currentLocusPart)) {
+          if (Object.prototype.hasOwnProperty.call(currentLocusPart, key)) {
+            findAndStoreMetaData(currentLocusPart[key]);
+          }
+        }
```
jor-row, line 376: silly q: can the currentLocusPart only be either an array or an object? Would it be problematic to type it as such? There are technically some other types in js that would satisfy  the ```typeof currentLocusPart !== 'object' || currentLocusPart === null``` check such as a set/map (although I understand we won't receive these kinds of types from locus)

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -128,7 +442,424 @@ class HashTreeParser {
...
+
+  /**
+   * Adds entries to the passed in updateObjects array
+   * for the changes that result from adding and removing visible data sets.
+   *
+   * @param {HashTreeMessage} message - The hash tree message that triggered the visible data set changes.
+   * @param {string[]} addedDataSets - The list of added data sets.
+   * @returns {Promise<void>}
+   */
+  private async initializeNewVisibleDataSets(
```
jor-row, line 691: Would it be possible for a dataSet to be removed from the visible list before we get around to doing this? I guess it isn't worth putting a check in here to see if they are still visible because the next call to processVisibleDataSetChanges will clean everything up anyway (just thinking "aloud" here)

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -128,7 +442,424 @@ class HashTreeParser {
...
+
+    LoggerProxy.logger.info(
+      `HashTreeParser#parseMessage --> ${this.debugId} received message ${debugText || ''}:`,
+      message
+    );
+    if (message.locusStateElements?.length === 0) {
+      LoggerProxy.logger.warn(
+        `HashTreeParser#parseMessage --> ${this.debugId} got empty locusStateElements!!!`
+      );
+      // todo: send a metric
```
jor-row, line 740: How come we want to send a metric here ? Is this something that you've seen happening in testing that shouldn't? Wondering if there are any other places we should send metrics

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -139,7 +870,292 @@ class HashTreeParser {
...
+      LoggerProxy.logger.info(
+        `HashTreeParser#runSyncAlgorithm --> ${this.debugId} Data set "${dataSet.name}" has no hash tree, skipping sync algorithm`
+      );
+
+      return;
+    }
+
+    dataSet.hashTree.resize(receivedDataSet.leafCount);
+
+    // temporary log for the workshop // todo: remove
```
jor-row, line 943: I'm guessing we want to keep these temp logs for debugging?

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
jor-row: Could the size of the dataTree and version etc be out of date until the next message arrives?

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -128,7 +442,424 @@ class HashTreeParser {
    * @returns {void}
    */
   handleLocusUpdate(update: {dataSets?: Array<DataSet>; locus: any}): void {
-    // todo
+    const {dataSets, locus} = update;
```
jor-row, line 445: should we be running this dataSets object through updateDataSetInfo here?

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: as discussed offline, I prefer to keep the type as "any" as we might easily miss something if we try to list all possible types, but at the same time the type will never be any JS data structure like set/map, because it's a Locus DTO that's a result of a call to JSON.parse()

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: as discussed offline, yes if a hash tree exists, it will be kept up-to-date via the heartbeat mechanism, but there is one things we can add here to handle a case if we never get any message at all: we can setup the sync timer (call `runSyncAlgorithm()`) for each of the data sets here

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: yes, it was happening during workshop when we had a bug in hash calculation - we were incorrectly thinking that we're out of sync and that triggered Locus to send us a message with a list of out of sync objects, but because we were actually in-sync (only our hash calculation was wrong) that message resulted in being empty. 
There may be some other places we want to send metrics - I've raised SPARK-753139 to track this

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: yeah, probably while 5k webinars are only testable in integration then it's worth keeping them, we'll remove them later on

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: yes, I think we should add this here

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: yeah I guess theoretically it is possible, but the datasets we get from Locus in line 692 are only supposed to be the visible ones at that moment in time (because the url we're using is `visibleDataSetsUrl`) and also as you say it would get cleaned up on next message, so I think we're OK here

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
marcin-bazyl: this line of code and the rest of initial implementation of the hash tree class was done by Colin

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTree.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
marcin-bazyl: fixed

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
marcin-bazyl: added unit tests

packages/@webex/plugin-meetings/src/hashTree/hashTree.ts
marcin-bazyl: added tests

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: actually, I've found that runSyncAlgorithm() is already called - it's called because we call sendInitializationSyncRequestToLocus() a few lines above and that calls parseMessage() which calls runSyncAlgorithm(), so I've only updated the relevant unit tests to add a check that the timers are set.
There is potentially 1 edge case where the timers won't be set - that's if the sync request we send to Locus in sendInitializationSyncRequestToLocus() fails. I'll check with Locus team what we should do in that case, but let's take this out of scope of this PR.

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: done - call added and test updated

jor-row (approved):

mickelr (approved):

PR #4592, opened 2025-11-30
feat(calling): failover cache logic

Kesari3008:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7420 

Storing the failover state everytime a retry is scheduled so that if browser refresh occurs in between of these inyervals, we can retain the existing timer interval and use that for next retry.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Testcase: Block the primary url and let the registration fail. In one of the retry intervals, refresh the browser and verify that failover isn't started from scratch.

  - [ ] Github Copilot
  - [x] Other - Please Specify
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -68,6 +68,8 @@ export class Registration implements IRegistration {
 
   private webex: WebexSDK;
 
+  private static readonly FAILOVER_CACHE_PREFIX = 'webex-calling-failover-state';
```
rarajes2: nitpick: 
1. This can be moved out to a constant file
2. We can shorten it to `wxc-failover` or `wxc-failover-state` since we are also adding `userId`, it will make the key very long

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -129,6 +131,36 @@ export class Registration implements IRegistration {
     this.backupMobiusUris = [];
   }
 
+  private getFailoverCacheKey(): string {
+    return `${Registration.FAILOVER_CACHE_PREFIX}.${this.userId || 'unknown'}`;
```
rarajes2: Let's use `??` instead of `||`

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -129,6 +131,36 @@ export class Registration implements IRegistration {
...
+  }
+
+  private saveFailoverState(failoverState: FailoverCacheState): void {
+    try {
+      localStorage.setItem(this.getFailoverCacheKey(), JSON.stringify(failoverState));
```
rarajes2, line 139: - Let's not use the `localStorage` API directly here. We can create util methods in a separate file and if any change is required with the storage API, the change can be made in just one place and it will be reflected everywhere.
- Question: We were supposed to use bounded storage, right ?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -129,6 +131,36 @@ export class Registration implements IRegistration {
...
+        method: 'saveFailoverState',
+      });
+    }
+  }
+
+  private clearFailoverState(): void {
+    try {
+      localStorage.removeItem(this.getFailoverCacheKey());
+    } catch {
+      // ignore
```
rarajes2: We can add log

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -691,6 +738,21 @@ export class Registration implements IRegistration {
...
+        this.resumeFailover(failoverState);
+
+        return;
+      }
+    } catch (error) {
+      log.warn(`No failover state found in cache`, {
+        file: REGISTRATION_FILE,
+        method: 'triggerRegistration',
+      });
+    }
```
rarajes2: This whole logic can be moved inside the `resumeFailover()` method. 

`resumeFailover()` can return a boolean and based on that we can return from here or proceed.

e.g.

```ts
if (resumeFailover()) return;
```

packages/calling/src/SDKConnector/types.ts
```diff
@@ -66,6 +67,11 @@ export type Logger = {
 // TODO: is there a way to import bindings from the Webex JS SDK without having to redefine expected methods and structure?
 // This defines the shape for the webex SDK, if a typing doesn't exist, it should be added here
 export interface WebexSDK {
+  boundedStorage: {
+    get: (namespace: string, key: string) => Promise<FailoverCacheState>;
+    put: (namespace: string, key: string, value: FailoverCacheState) => Promise<void>;
+    del: (namespace: string, key: string) => Promise<void>;
+  };
```
rarajes2, line 74: Are we planning to use it ?

rarajes2 (changes requested):
All the failover state management related methods can be moved to a util file.

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: Bounded storage by default uses InMemoryStoreAdapter which is not persistent. So eventually we have to overwrite the bounded storage to use some sort of persistent storage which is evntually local storage so I just used that directly instead of going round about way

Kesari3008: To overwrite, it needs to be done from sample app so we will have to instruct users to do it which I feel shouldn't be needed. We should control it internally.

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: addressed

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: added

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: done

packages/calling/src/CallingClient/registration/register.ts
rarajes2: > We can create util methods in a separate file and if any change is required with the storage API, the change can be made in just one place and it will be reflected everywhere.

Still good to do this

packages/calling/src/SDKConnector/types.ts
rarajes2: If we are not planning to use it, we can remove the type

rarajes2 (approved):
Please address remaining comments before merging the PR

packages/calling/src/SDKConnector/types.ts
rarajes2: Discussed, we already have the boundedStorage available. We are adding the type

rsarika (approved):
we can remove

 boundedStorage: {
    get: (namespace: string, key: string) => Promise<FailoverCacheState>;
    put: (namespace: string, key: string, value: FailoverCacheState) => Promise<void>;
    del: (namespace: string, key: string) => Promise<void>;
  };
  
  
  and add when we actually implement this.

PR #4639, opened 2026-01-05
feat(contact-center): added-mpc

adhmenon:
# COMPLETES #< [CAI-7286](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7286)>

* Adds most features of the MPC (Multi party conference)
* Fixed and updated unit tests of consult flow

* Added code in state machine and actions for MPC.
* Added sample app changes.

Vidcast - https://app.vidcast.io/share/ce3d2942-79d9-499f-9678-90ca4dda9d13?playerMode=vidcast

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

* Tested manually as per the widgets test spec.
* Tested 3PC and MPC only.
* Some issues remain like agent list & merge - will be handled in future PRs

  - [ ] Github Copilot
  - [x] Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [x] Tech Debt
  - [ ] Automation

### I certified that

---


adhmenon (commented):
Some flags are redundant, working on removing them and optimising the guard cases.

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+  MEDIA_TYPE_MAIN_CALL,
+} from './state-machine/constants';
+
+// Re-export for backward compatibility
+export {MAX_PARTICIPANTS_IN_MULTIPARTY_CONFERENCE};
+
+/**
+ * Media entry type from interaction.media
+ */
+type MediaEntry = {
```
Kesari3008: Shouldn't this and below be present in types file ?

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+  return {isHeld: mediaEntry.isHold === true, mediaResourceId};
+};
+
+/**
+ * Gets the hold status of the consult call leg
+ * @param interaction - The interaction object
+ * @param interactionId - The main interaction ID
+ * @returns true if consult call is held
+ */
+export const getConsultCallHeld = (interaction: Interaction, interactionId: string): boolean => {
```
Kesari3008: I feel these 2 methods are unncessary because anyway when to pass the the Media type as consult and media type as main call is decided based on where you call this method so directly we can call findHoldStatus() from those places and pass the right media type

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+ * Checks if the agent is on the consult leg and the consult is NOT held
+ * (meaning the main call is effectively held from the agent's perspective)
+ * Per widgets-util-logic.md: isConsultOnHoldMPC implementation
+ *
+ * @param interaction - The interaction object
+ * @param interactionId - The main interaction ID (kept for API compatibility)
+ * @returns true if agent is on consult (and thus main call is held from their POV)
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const isConsultOnHoldMPC = (interaction: Interaction, interactionId: string): boolean => {
```
Kesari3008: We seem to be passing interactionId in all these methods but is isn't really getting used

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+  return findHoldStatus(interaction, interactionId, MEDIA_TYPE_CONSULT).isHeld;
+};
+
+/**
+ * Gets the hold status of the main call leg
+ * @param interaction - The interaction object
+ * @param interactionId - The main interaction ID
+ * @returns true if main call is held
+ */
+export const getMainCallHeld = (interaction: Interaction, interactionId: string): boolean => {
```
Kesari3008: Above 4 methods can be combined into one single method and and return isHeld status directly based on per call leg

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+  return findHoldStatus(interaction, interactionId, MEDIA_TYPE_MAIN_CALL).isHeld;
+};
+
+/**
+ * Checks if the customer is still in the call (not left)
+ * Per widgets-util-logic.md: getIsCustomerInCall implementation
```
Kesari3008: Please remove this comment from everywhere

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+ *
+ * @param interaction - The interaction object
+ * @param interactionId - The main interaction ID
+ * @returns true if customer is in the call
+ */
+export const getIsCustomerInCall = (interaction: Interaction, interactionId: string): boolean => {
+  if (!interaction?.media || !interaction?.participants) {
```
Kesari3008: I believe these checks in all the utility method are inncessary because interaction seems to be a mandatory param and ideally we should not even be invoking these util method if interaction object itself can be optional

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+    return false;
+  }
+
+  // Get the main call media
+  const mainCallMedia = interaction.media[interactionId];
```
Kesari3008, line 16: Why is this extra check needed ?

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+
+/**
+ * Checks if the maximum number of participants in a conference has been reached
+ * Per conference-spec.md: Max 7 counted participants + 1 customer
+ *
+ * @param interaction - The interaction object
+ * @param interactionId - The main interaction ID
+ * @returns true if max participants reached
+ */
+export const hasReachedMaxParticipants = (
```
Kesari3008: This method is not adding much value

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,280 @@
...
+  const conferenceParticipantsCount = getConferenceParticipantsCount(interaction, interactionId);
+  const isConsultInProgress = getIsConsultInProgress(interaction);
+  const isCustomerInCall = getIsCustomerInCall(interaction, interactionId);
+
+  // Per widgets-util-logic.md gating logic:
+  // 1. Must not be at max participants
+  // 2. Must not have consult already in progress
+  // 3. Customer must be in call
+  // 4. Special case: if held AND conference in progress AND consult not completed, disable
+  if (conferenceParticipantsCount >= MAX_PARTICIPANTS_IN_MULTIPARTY_CONFERENCE) {
```
Kesari3008: These ifs can be combined into one single condition

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -71,7 +71,8 @@ export default abstract class Task extends EventEmitter implements ITask {
     super();
     this.contact = contact;
     this.data = data;
-    this.uiControlConfig = uiControlConfig;
+    // Include agentId in the config for ownership checks (transfer conference)
```
Kesari3008, line 76: We can remove this comment

packages/@webex/contact-center/src/services/task/state-machine/constants.ts
```diff
@@ -3,6 +3,43 @@
...
 
+// ============================================
+// Conference Constants
+// ============================================
+
+/**
+ * Maximum number of participants allowed in a multi-party conference.
+ * Max 7 counted agents + 1 customer.
+ */
+export const MAX_PARTICIPANTS_IN_MULTIPARTY_CONFERENCE = 7;
```
Kesari3008, line 14: We can just name it MAX_PARTICIPANTS

packages/@webex/contact-center/src/services/task/state-machine/constants.ts
```diff
@@ -3,6 +3,43 @@
...
+
+// ============================================
+// Participant Type Constants
+// ============================================
+
+/** Customer participant type */
+export const PARTICIPANT_TYPE_CUSTOMER = 'Customer';
```
Kesari3008: These are categories for the same thing so they can be created as type: PARTICIPANT_TYPE and in that you have mutliple values

packages/@webex/contact-center/src/services/task/state-machine/types.ts
```diff
@@ -150,17 +133,36 @@ interface TaskEventPayloadMap {
...
   [TaskEvent.CONFERENCE_START]: BaseEvent<TaskEvent.CONFERENCE_START> & {
     participants?: ConferenceParticipant[];
+    taskData?: TaskData;
   };
   [TaskEvent.CONFERENCE_FAILED]: BaseEvent<TaskEvent.CONFERENCE_FAILED> & {
     reason?: string;
+    taskData?: TaskData;
   };
-  [TaskEvent.CONFERENCE_END]: BaseEvent<TaskEvent.CONFERENCE_END>;
+  [TaskEvent.CONFERENCE_END]: BaseEvent<TaskEvent.CONFERENCE_END> & {taskData?: TaskData};
```
Kesari3008: Why is taskData optional in these conference events?

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -4,35 +4,575 @@
...
+  },
+
+  // ============================================
+  // Consult Guards
+  // ============================================
+
+  /**
+   * Check if a consult is currently in progress
+   */
+  consultInProgress: ({context}: GuardParams): boolean => {
```
Kesari3008: We don't need guards for positive and negative both usecases, just one condition should suffice both. Let's discuss the guards created in this file if they are really needed and what all would be needed

Kesari3008 (changes requested):

adhmenon:
Tested everything - most cases work fine. Only switch and some minor UI issues are left.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -722,6 +747,270 @@ export default class Voice extends Task implements IVoice {
...
+      throw detailedError;
+    }
+  }
+
+  /**
+   * Check if conference is in progress from task data
+   * Uses same logic as guards.isConferenceInProgress but directly on instance data
+   * @returns true if 2+ agents are in the main call
+   */
+  private checkIsConferenceInProgress(): boolean {
```
adhmenon: Nit: Will be moving this to utils - have done the cleanup locally but will push after Arun's changes are merged and probably along with other review comments.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -177,11 +177,13 @@ export default class TaskManager extends EventEmitter {
...
    * @param ccEvent - The CC_EVENT type from WebSocket
    * @param payload - The event payload
+   * @param agentId - Optional agent ID for state detection (needed for HYDRATE)
    * @returns TaskEventPayload for state machine or null if no mapping
    */
   private static mapEventToTaskStateMachineEvent(
     ccEvent: CC_EVENTS,
-    payload: WebSocketPayload
+    payload: WebSocketPayload,
+    agentId?: string
```
adhmenon, line 147: Needed because otherwise some `AgentConsultEnded` events were not getting mapped properly - they were being cleared by the state machine during `clearConsultState`.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -412,7 +439,40 @@ export default class TaskManager extends EventEmitter {
       return null;
     }
 
-    const task = this.taskCollection[message.data.interactionId];
+    // Some consult/conference events can arrive on a child interactionId while the task we track
+    // is keyed by the main/parent interactionId. Resolve to an existing task when possible.
+    const interactionId = message.data.interactionId;
```
adhmenon, line 386: Nit: Will revert this - added by mistake.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -247,16 +217,18 @@ export default class TaskManager extends EventEmitter {
...
           type: TaskEvent.TRANSFER_SUCCESS,
           taskData: payload,
         };
 
-      case CC_EVENTS.AGENT_VTEAM_TRANSFERRED:
       case CC_EVENTS.AGENT_WRAPUP:
+        return {type: TaskEvent.TASK_WRAPUP, taskData: {...payload, wrapUpRequired: true}};
       case CC_EVENTS.AGENT_CONTACT_UNASSIGNED:
-        return {type: TaskEvent.END, taskData: {...payload, wrapUpRequired: true}};
+        return null; // Add WRAPUP if needed
```
rsarika, line 232: `AGENT_CONTACT_UNASSIGNED` now maps to `null`, so the state machine never sees a terminal/wrapup event. Previously this flowed to END with `wrapUpRequired: true`. If this event still means “agent unassigned from task”, skipping the state machine can leave the task in collection and skip cleanup/wrapup. Should this map to `TASK_WRAPUP` or `CONTACT_ENDED` instead of `null`?

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -4,46 +4,311 @@
...
+      ? getIsConferenceInProgress(context.taskData)
+      : false;
+
+    return conferenceInEvent || conferenceInContext;
+  },
+
+  shouldDowngradeConference: ({context, event}: GuardParams): boolean => {
+    const eventTaskData = getTaskDataFromEvent(event);
+    const taskData = eventTaskData ?? context.taskData;
+    if (!taskData?.interaction || !taskData?.interactionId) return false;
```
rsarika: `shouldDowngradeConference()` uses `taskData.interactionId` when counting participants. For consulted agents, `interactionId` may differ from the main call; other helpers now use `mainInteractionId`. Should this use `taskData.interaction.mainInteractionId || taskData.interactionId` to avoid miscounting and incorrect downgrade/termination?

rsarika (commented):
Posting inline comments based on verified diffs.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
adhmenon: This is valid, however I was using this before but felt it might not have been needed.
As of now, things do seem to be working with the current interactionId since the rest of our guards are adequate here.

packages/@webex/contact-center/src/services/task/TaskManager.ts
adhmenon: This should be fine for now Ravi, since backend will always send `AGENT_WRAPUP` after this event, if needed. I believe in case of queue consult-transfer this happens.
In that case our state machine moves to `WRAPPING_UP`.
If not, we just go to `TERMINATED`. But we can test these edge cases, if needed. Have already tested it out.

adhmenon:
Full MPC + Edge Cases - https://app.vidcast.io/share/d66170ec-06cd-4e21-9279-dcbcd5358613

MPC + Downgrade - https://app.vidcast.io/share/19a105d2-5c7a-4f48-8bf0-68739d1f7cb3

packages/@webex/contact-center/src/services/task/state-machine/constants.ts
```diff
@@ -3,10 +3,46 @@
...
+
+// ============================================
+// Participant Type Constants
+// ============================================
+
+/** Participant types for conference filtering */
+export const PARTICIPANT_TYPE = {
+  CUSTOMER: 'Customer',
+  SUPERVISOR: 'Supervisor',
+  VVA: 'VVA',
```
Kesari3008, line 24: Which participant type is this? Are we using it ?

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -597,74 +550,239 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
-          // AgentConferenceEnded
-          [TaskEvent.CONFERENCE_END]: {
-            target: TaskState.CONNECTED,
-            actions: ['clearConsultState'],
+          [TaskEvent.EXIT_CONFERENCE_FAILED]: {
+            actions: ['handleExitConferenceFailed'],
+          },
+
+          [TaskEvent.TRANSFER_CONFERENCE]: {
+            actions: ['emitTaskTransferConference'],
```
Kesari3008: If we are emitting the event towards client on this transition itself then what value Transfer_ConferenceSuccess is adding in state machine

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -597,74 +550,239 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+                'emitTaskParticipantLeft',
+                'emitTaskConferenceEnded',
+              ],
             },
+            {actions: ['updateTaskData', 'handleParticipantLeft', 'emitTaskParticipantLeft']},
           ],
-          // AgentConferenceTransferred
-          [TaskEvent.TRANSFER_CONFERENCE]: [
+          [TaskEvent.EXIT_CONFERENCE]: {
+            actions: ['setExitingConference', 'emitTaskExitConference'],
```
Kesari3008: Same question here

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -597,74 +550,239 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+                'updateTaskData',
+                'markEnded',
+                'clearConsultState',
+                'handleExitConferenceSuccess',
+                'emitTaskWrapup',
+              ],
+            },
+            {
+              // Agent exited → TERMINATED
+              guard: guards.isExitingConference,
```
Kesari3008: Why is this guard needed? What is it helping in terms of decision making?

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -597,74 +550,239 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+          [TaskEvent.CONFERENCE_END]: [
+            {
+              // Owner/primary who should wrap up → WRAPPING_UP (must be first!)
+              guard: guards.shouldWrapUp,
+              target: TaskState.WRAPPING_UP,
+              actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
+            },
+            {
+              // Non-owner who triggered exit → TERMINATED
+              guard: guards.isExitingConference,
```
Kesari3008: There is no difference in action here and the one without guard down, transition state is also Terminated for both so why is the guard needed at all and why 2 blocks. Single section for it without the guard should suffice

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -417,39 +321,60 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
             actions: ['updateTaskData', 'handleConsultFailed'],
           },
-          // AgentConsultCreated
-          [TaskEvent.CONSULT_CREATED]: {
-            target: TaskState.CONSULTING,
-            actions: ['updateTaskData', 'setConsultInitiator', 'emitTaskConsultCreated'],
-          },
+          [TaskEvent.CONSULT_CREATED]: [
+            {
+              guard: guards.isConsultingAgentOrBeingConsulted,
```
Kesari3008: This guard should not be required since Consult Created is only received by the initiator and also user will only be in CONSULT_Initiating state who actually pressed the consult button so this transition can happen without the guard

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -417,39 +321,60 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
             {
-              guard: isConsultQueueFlow,
+              guard: guards.isConsultQueueFlow,
               target: TaskState.CONNECTED,
               actions: ['updateTaskData', 'handleConsultFailed'],
             },
             {
-              guard: serverReportsHeld,
+              // If this consult originated from conference, always return to CONFERENCING.
+              guard: ({context}) => context.consultFromConference === true,
```
Kesari3008, line 342: If we can use context directly here for decision making then all the one liner guadrs can be avoided and context can directly be used. Also why do we need to check on this and in all the transitions to Conferencing state. There is no dofference in target state or actions invoked so why are we using 3 different guards and doing it 3 times

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -461,17 +386,17 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           // AgentCtqCancelled Event
           [TaskEvent.CTQ_CANCEL]: [
             {
-              guard: isConsultQueueFlow,
+              guard: guards.isConsultQueueFlow,
```
Kesari3008: This guard is also not needed

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -417,39 +321,60 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+              actions: ['updateTaskData', 'handleConsultFailed'],
+            },
+            {
+              guard: guards.serverReportsHeld,
               target: TaskState.HELD,
               actions: ['updateTaskData', 'handleConsultFailed'],
             },
             {
-              guard: serverReportsConsulting,
+              guard: guards.serverReportsConsulting,
```
Kesari3008: If consult failure is received in consult initiating state, what case can be there where we would want to transition to consulting state. This doesn't seem to be a valid case

packages/@webex/contact-center/src/services/task/TaskUtils.ts
```diff
@@ -2,6 +2,126 @@
 import {Interaction, ITask, TaskData, MEDIA_CHANNEL} from './types';
 import {OUTDIAL_DIRECTION, OUTDIAL_MEDIA_TYPE, OUTBOUND_TYPE} from '../../constants';
 import {LoginOption} from '../../types';
+import {PARTICIPANT_TYPE, MEDIA_TYPE_MAIN_CALL} from './state-machine/constants';
```
Kesari3008, line 5: Now that we have action, guards, context and ui controls, how relevant this file is ? Do we need it ? Can util methods being used here not be combined with state machine logic ? Having both appraches are redundant

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1811,3 +1849,44 @@ export interface IWebRTC extends IVoice {
...
+/**
+ * Context for processing an event
+ *
+ * Contains all information needed to process a WebSocket event:
+ * - Event type and payload from the backend
+ * - Task instance (if exists)
+ * - Pre-mapped state machine event (if applicable)
+ * - Task state flags (e.g., was this a consulted task)
+ */
+export interface EventContext {
```
Kesari3008, line 1874: Having only this property: wasConsultedTask here in this type looks little odd as oher objects are bigger and high level objects ? Is there no other type to fit this in

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1782,6 +1806,20 @@ export type Participant = {
  */
 export type TaskAccessorParticipant = Participant;
 
+/**
+ * Media entry type from interaction.media
+ * Used for media state tracking in consult and conference scenarios
+ */
+export type MediaEntry = {
```
Kesari3008: Where is this getting used ? Couldn't find it getting used

Kesari3008 (changes requested):
Let's discuss the statemahcine transitions, guards and action in the meeting for now. Rest of the files are mostly reviewed except tests

adhmenon:
Tested after changes, everything works fine - https://app.vidcast.io/share/41b49e60-fcf8-4089-b630-82b8c5bd8b43

packages/@webex/contact-center/src/services/task/types.ts
adhmenon: Addressed.

packages/@webex/contact-center/src/services/task/state-machine/constants.ts
adhmenon: Talked with Ravi, I think we should keep this for the virtual agent. Might be useful in the future.

packages/@webex/contact-center/src/services/task/TaskUtils.ts
adhmenon: I have combined as many utils as possible with the guards and ui controls.. did not want to go further but we can take it up as an effort with the next PR.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika, line 232: **Critical:** `AGENT_CONTACT_UNASSIGNED` previously mapped to `END` with `wrapUpRequired: true`. Now it returns `null`, so the state machine never sees a terminal event. If the backend doesn't always follow up with `AGENT_WRAPUP` or `CONTACT_ENDED`, the task will stay in the collection indefinitely with no cleanup.

Is there a guarantee from the backend that a follow-up event always arrives? If not, this should map to `TASK_WRAPUP` or `CONTACT_ENDED`.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -4,46 +4,248 @@
...
+  /**
+   * Conference downgrade check specifically for transitioning back to CONNECTED.
+   *
+   * Returns true only when:
+   * - conference has downgraded (fewer than 2 active agent participants in main call)
+   * - customer is still in the call
+   * - current agent is still in the main call
+   */
+  shouldDowngradeConferenceToConnected: ({context, event}: GuardParams): boolean => {
+    const eventTaskData = getTaskDataFromEvent(event);
```
rsarika, line 146: **Note:** For consulted agents, `taskData.interactionId` may differ from the main call's ID. If `mainInteractionId` is not populated and `interactionId` is a child interaction, `getConferenceParticipantsCount` will look up the wrong media entry, return 0, and incorrectly trigger a downgrade.

Consider logging a warning when `mainInteractionId` is missing so this edge case is surfaced during debugging.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -672,6 +693,13 @@ export default class Voice extends Task implements IVoice {
         data: paramsDataForConferenceV2.data,
       });
 
+      // Send success event to transition to CONFERENCING
+      if (this.stateMachineService) {
+        this.stateMachineService.send({
+          type: TaskEvent.CONFERENCE_START,
+        });
```
rsarika, line 700: **Minor:** `CONFERENCE_START` is sent locally here after the API resolves, but the backend will also send `AGENT_CONSULT_CONFERENCED` which maps to `CONFERENCE_START` in TaskManager (line 274). This may cause a duplicate conference-started emission to consumers.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -259,67 +190,50 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
-            actions: ['updateTaskData'],
           },
           // Click of the consult button
           [TaskEvent.CONSULT]: {
             target: TaskState.CONSULT_INITIATING,
             actions: ['setConsultInitiator', 'setConsultDestination'],
           },
-          // AgentConsultCreated event confirms the consult request
-          [TaskEvent.CONSULT_CREATED]: {
-            target: TaskState.CONSULTING,
```
rsarika, line 279: **Note:** `CONSULT_CREATED` handling was removed here, but `TaskManager.mapEventToTaskStateMachineEvent` still maps `AGENT_CONSULT_CREATED` → `TaskEvent.CONSULT_CREATED` (TaskManager line 194). The event is now silently dropped by the state machine. If this is intentional, consider also removing the mapping in TaskManager to keep things consistent.

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -326,6 +402,9 @@ export default abstract class Task extends EventEmitter implements ITask {
 
   private getCommonActionOverrides(): Partial<TaskActionsMap> {
     return {
+      syncTaskDataFromEvent: ({event}: {event: TaskEventPayload}) => {
+        this.updateTaskFromEvent(event);
```
rsarika, line 406: **Minor:** `CONTACT_UPDATED` and `CONTACT_OWNER_CHANGED` now run both `updateTaskData` (updates XState context.taskData) and `syncTaskDataFromEvent` (updates Task instance `this.data`). This dual-path update works but is worth being aware of — if a future refactor splits these, the two data stores could diverge.

rsarika (approved):
Approving with inline comments on a few items to address or track. The MPC state machine design is solid — clean separation between TaskManager lifecycle and state machine transitions, well-organized guards, and thorough UI controls computation. Nice work on the refactoring.

adhmenon:
Merging the PR now. Ravi, I'll handle the comments with the switch PR.

PR #4651, opened 2026-01-09
feat(meetings): added handling for WebrtcApiNotAvailableError

marcin-bazyl:
# COMPLETES #[SPARK-753012](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-753012)

Some users have browser extensions installed that remove RTCPeerConnection. Web app needs to detect this and show a specific error message

- isWebrtcApiNotAvailableError() to call diagnostics metrics util - this is used by web app to check the error they will receive when addMedia() fails
- added handling of a new error WebrtcApiNotAvailableError to joinWithMedia() - to avoid doing a retry
- added pre-check in joinWithMedia() so that we don't even try to join the Locus if we know that RTCPeerConnection is not available
- added checks also to reachability code - without this reachability would fail and web app would show "your network doens't meet webex requirements" modal

depends on this PR from internal-media-core:
https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/363

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual test with web app and internal-media-core

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---


szotrabh (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
```diff
@@ -179,6 +179,14 @@ export const isSdpOfferCreationError = (rawError: any) => {
   return false;
 };
 
+export const isWebrtcApiNotAvailableError = (rawError: any) => {
```
antsukanova: I see that in other places there is already `any` here, but maybe at least it would be good to have something like `Error | {  code: number;
      message: string;
      name: string}`

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -196,6 +197,14 @@ export default class Reachability extends EventsScope {
     if (!this.webex.config.meetings.enableReachabilityChecks) {
       throw new Error('enableReachabilityChecks is disabled in config');
     }
+
+    if (WebCapabilities.supportsRTCPeerConnection() === CapabilityState.NOT_CAPABLE) {
```
antsukanova: How have you decided that here should be  `NOT_CAPABLE` and in `packages/@webex/plugin-meetings/src/meeting/index.ts` is `if (WebCapabilities.supportsRTCPeerConnection() !== CapabilityState.CAPABLE) {`

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
marcin-bazyl: hmm I'm not convinced... the function is called with any error that is caught during meeting join/media establishment, so there are tons of potential things that could be thrown there. In theory it should be just errors, but I can't guarantee that, I've seen SDK code before where it would just throw strings or other random things, so putting a more specific type here would make you falsely think that the object is actually of that type. This is a util function that's meant to take anything and tell you if it's a WebrtcApiNotAvailableError, so restricting the type of objects it takes would not be helpful for the user of this function. In the end the code that calls it would probably force a cast anyway, so we wouldn't really have any more type safety, just a false impression of having type safety.

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: no special reason... the function only ever returns CAPABLE or NOT_CAPABLE, it won't return UNKNOWN, so it doesn't matter much, but if we want to be very thorough then probably it would be better the other way around:
- in join we should check === NOT_CAPABLE, because only if we are 100% sure that we don't have RTCPeerConnection then we want to fail early, otherwise if it's UNKNOWN it's OK to continue and we'll find out if RTCPeerConnection is actually available or not when media connection is attempted - if it throws, that will be handled
- in reachability code we should do !== CAPABLE, because we don't want to risk starting reachability and then failing as we would end up with the reachability result being "unreachable" for some clusters then and that could cause the app to incorrectly display a warning about the user's network not meeting the webex requirements

I'll change this now.

marcin-bazyl: done

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
antsukanova: Then I would say you can do `{ code: number; message: string; name: string} | unknown` because you are expecting and checking exactly the error for WebrtcApi and you know its type, but all others are unknown to you. What do you think?
I know that types are not a strong part of the SDK currently, but overall it does not feel hard to help us make code clearer with a bit more precise expected types

packages/@webex/plugin-meetings/src/reachability/index.ts
antsukanova: This logic lgtm. Thanks for updating it. It’s much cleaner now.

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics.util.ts
antsukanova: lgtm with updates

antsukanova (approved):

arun3528 (approved):

PR #4609, opened 2025-12-11
Performing reachability check of each UDP URL when newly added flag(`enablePerUdpUrlReachability`) enabled else existing flow (combined reachability check).

venky-mediboina:
# COMPLETES [SPARK-610518](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610518)

## This pull request addresses

As changes are somewhat big, we split into 3 PRs. Old combined PR: https://github.com/webex/webex-js-sdk/pull/4542.

This is the 2nd PR of 3 PRs to add subnet details which consists of each subnet IP in the details section including whether its reachable or not.

As part of this PR-2, added a flag in config, based on this flag each UDP URL will be checked or follows existing all with  single peer connection
```
PR-1: Create ReachabilityPeerConnection class and refactor existing code in ClusterReachability to use it - there should be no logic change at all, no behavioral change, the code should do exactly all the same as it currently does, but ClusterReachability should not have any code that deals with PeerConnection APIs or candidates. It should just create 1 instance of ReachabilityPeerConnection and ReachabilityPeerConnection should do all that work.

PR-2: Add support for per URL checks for UDP- in this PR we'll introduce the new config flag and when it's set, ClusterReachability will pass only TCP and TLS URLs to its single ReachabilityPeerConnection instance and will create additional array of ReachabilityPeerConnection instances - one for each UDP URL.

PR-3: Add all the extra stuff like getAllClustersInfo() API, populating the details and sending them to the backend etc.
```

## by making the following changes

Added a flag in reachability config `enablePerUdpUrlReachability`, and based on this flag each UDP URL will be checked or follows existing all with  single peer connection.

### Change Type

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios were tested

MANUAL Tests performed  with unit test cases

## The GAI Coding Policy And Copyright Annotation Best Practices ##
 

 
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [X] GAI was used to create a draft that was subsequently customized or modified
- [ ] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [ ] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [X] Github Copilot
[...]

PR #4662, opened 2026-01-21
fix(plugin-meetings): add subnet details in reachability, getAllClust…

venky-mediboina:
# COMPLETES [SPARK-610518](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-610518)

Changes flow documented in this [wiki ](https://confluence-eng-gpk2.cisco.com/conf/spaces/MediaDebugTools/pages/806988909/Overview+of+3rd+PR+changes+for+Subnet+IP+details+in+Reachability+Results) page.

1st PR [link](https://github.com/webex/webex-js-sdk/pull/4580).
2nd PR [link](https://github.com/webex/webex-js-sdk/pull/4609).

Old combined PR: https://github.com/webex/webex-js-sdk/pull/4542. As changes are somewhat big, we split into 3 PRs. 

Description for 3PRs:
```
PR-1: Create ReachabilityPeerConnection class and refactor existing code in ClusterReachability to use it - there should be no logic change at all, no behavioral change, the code should do exactly all the same as it currently does, but ClusterReachability should not have any code that deals with PeerConnection APIs or candidates. It should just create 1 instance of ReachabilityPeerConnection and ReachabilityPeerConnection should do all that work.

PR-2: Add support for per URL checks for UDP- in this PR we'll introduce the new config flag and when it's set, ClusterReachability will pass only TCP and TLS URLs to its single ReachabilityPeerConnection instance and will create additional array of ReachabilityPeerConnection instances - one for each UDP URL.

PR-3: Add all the extra stuff like getAllClustersInfo() API, populating the details and sending them to the backend etc.
```

This is the last and 3rd PR of 3 PRs series to add subnet details which consists of each subnet IP in the details section including whether its reachable or not.

As part of this PR-3, added subnet details to reachability results and a function to return all URLs.

Added subnet details to reachability results based on this flag and a function to return all URLs.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

MANUAL Tests performed  with unit test cases

- [X] GAI was used to create a draft that was subsequently customized or modified
  - [X] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [X] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/package.json
```diff
@@ -1,96 +1,97 @@
...
-    "lodash": "^4.17.21",
-    "uuid": "^3.3.2",
-    "webrtc-adapter": "^8.1.2",
-    "xxh3-ts": "^2.0.1"
-  },
-  "//": [
-    "TODO: upgrade jwt-decode when moving to node 18"
-  ]
-}
+{
```
marcin-bazyl: there are lots of whitespace changes in this file (on every line) - can you check this, please?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -231,6 +245,61 @@ export class ClusterReachability extends EventsScope {
...
+    const minLatency = reachableLatencies.length > 0 ? Math.min(...reachableLatencies) : undefined;
+
+    // Determine overall result
+    if (firstReachableResult) {
+      return {
+        result: 'reachable',
+        latencyInMilliseconds: firstReachableResult.latencyInMilliseconds,
+        clientMediaIPs: firstReachableResult.clientMediaIPs,
+        details: allDetails,
+        minLatency,
```
marcin-bazyl: shouldn't `minLatency` be the same as `latencyInMilliseconds`?

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
```diff
@@ -231,6 +245,61 @@ export class ClusterReachability extends EventsScope {
...
+    const reachableLatencies = allDetails
+      .filter((d) => d.answeredTx === 1 && d.latencies.length > 0)
+      .map((d) => Math.min(...d.latencies));
+    const minLatency = reachableLatencies.length > 0 ? Math.min(...reachableLatencies) : undefined;
+
+    // Determine overall result
+    if (firstReachableResult) {
+      return {
+        result: 'reachable',
+        latencyInMilliseconds: firstReachableResult.latencyInMilliseconds,
```
marcin-bazyl: not sure if this is correct. `firstReachableResult` is just the first found result in the array, it doesn't necessarily mean that it was actually the one that generated the "reachable" result as first in time, right? so I think this should be set to a minimum value of `latencyInMilliseconds` from all peer connections, shouldn't it?

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -922,11 +943,27 @@ export default class Reachability extends EventsScope {
...
+          result: cluster.udp.length > 0 ? 'unreachable' : 'untested',
+          details: udpDetails,
+        },
+        tcp: {
+          result: cluster.tcp.length > 0 ? 'unreachable' : 'untested',
+          details: tcpDetails,
+        },
+        xtls: {
+          result: cluster.xtls.length > 0 ? 'unreachable' : 'untested',
+          details: cluster.xtls.length > 0 ? xtlsDetails : [],
```
marcin-bazyl: why only xtls has the check for length > 0 and not udp nor tcp?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -285,6 +330,69 @@ export class ReachabilityPeerConnection extends EventsScope {
...
+    // Update minLatency
+    this.updateMinLatency(protocol);
+  }
+
+  /**
+   * Calculates and updates minLatency for a protocol based on reachable subnets
+   * @param {Protocol} protocol - the protocol to update
+   * @returns {void}
+   */
+  private updateMinLatency(protocol: Protocol): void {
```
marcin-bazyl: do we really need this function and maintaining of `minLatency` at all? we do 1 check for each url and so we only ever have 1 latency result for each url, so when we send results to Orpheus, `latencyInMilliseconds` will always be the same as `minLatency` so instead of trying to maintain both in our internal data structures, we only need 1 field `latencyInMilliseconds` and only need to create `minLatency` by copying `latencyInMilliseconds` when we prepare the data format for the backend in mapTransportResultToBackendDataFormat(), right?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -252,18 +308,7 @@ export class ReachabilityPeerConnection extends EventsScope {
...
-          file: 'reachabilityPeerConnection',
-          function: 'saveResult',
-        },
-        ReachabilityPeerConnectionEvents.resultReady,
-        {
-          protocol,
-          ...result,
-        }
-      );
+      // Note: resultReady event is emitted when ICE gathering completes in emitResultsForAllProtocols()
```
marcin-bazyl: this is a big change - before only "reachable" results were ever emitted, now "unreachable" ones are also emitted. This will require changes in places where this event is handled in reachability/index.ts and in clusterReachability.ts and I don't see them being done in this PR. Can you explain a bit more why do we need this change?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -339,22 +462,190 @@ export class ReachabilityPeerConnection extends EventsScope {
 
       if (e.candidate) {
         if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
-          let serverIp = null;
+          // In standard WebRTC mode (not per-URL), we may receive multiple srflx candidates
+          // but we only need to mark ONE pre-populated entry as reachable (the first one)
+          // For subsequent candidates, just collect public IPs for NAT detection
```
marcin-bazyl: the comment should say "and do NAT detection" instead of "for NAT detection"

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -339,22 +462,190 @@ export class ReachabilityPeerConnection extends EventsScope {
...
+            }
+          }
 
-            const match = (e.candidate as any).url.match(stunServerUrlRegex);
-            serverIp = match && match[1];
+          // Fallback: Use first pre-populated entry if URL extraction failed
+          // In standard mode (not per-URL), we don't know which server responded,
+          // so we mark the first one as reachable
+          if (!serverIp && this.result.udp.details && this.result.udp.details.length > 0) {
+            const firstDetail = this.result.udp.details[0];
```
marcin-bazyl: I think it would be better to just not have the details set at all for the non per-URL mode - we don't really need them and it's better to not have them then to have incorrect information

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -339,22 +462,190 @@ export class ReachabilityPeerConnection extends EventsScope {
 
       if (e.candidate) {
         if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
-          let serverIp = null;
+          // In standard WebRTC mode (not per-URL), we may receive multiple srflx candidates
+          // but we only need to mark ONE pre-populated entry as reachable (the first one)
+          // For subsequent candidates, just collect public IPs for NAT detection
+          if (this.udpFirstCandidateProcessed) {
```
marcin-bazyl: are you sure we need that flag? wasn't the whole point of doing separate peer connection for each url, because when we had multiple urls we would only ever get 1 candidate?

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -339,22 +462,190 @@ export class ReachabilityPeerConnection extends EventsScope {
...
+
+          // Determine protocol first, before fallback handling
+          let protocol: Protocol = 'tcp';
+          if ('relayProtocol' in e.candidate) {
+            const relayProto = (e.candidate as any).relayProtocol;
+            if (relayProto === 'tls') {
+              protocol = 'xtls';
+            }
+          } else if ('url' in e.candidate) {
+            // Fallback: check URL scheme (turns: indicates TLS)
```
marcin-bazyl: why do we need these other methods of detecting the protocol (by using `relayProtocol` or the `url` from e.candidate)? the method we've been using so far (checking the port) always works, doesn't it? why can't we just keep using it as is? it's better to have less code than more

packages/@webex/plugin-meetings/package.json
venky-mediboina: Removed extra spaces. Got added while fixing linting errors.

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: yeah, removed instances of minLatency from other places. Now copying while transforming into backend compatible format.

packages/@webex/plugin-meetings/src/reachability/index.ts
venky-mediboina: modified, now populating for all protocols.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: You are right. removed from other places and copying in `mapTransportResultToBackendDataFormat()` only.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: fixed.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: yeah, better to not carry the details for non per-URL mode. did the changes accordingly. Earlier tried to catch the details for both per-URL mode and standard mode, now changed.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Removed it. The flag is replaced by checking the current result state.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: reverted to old approach as it was simple. got some over cautious.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: `resultReady` still emits both reachable and unreachable (needed for per-URL UDP aggregation in `ClusterReachability`, but the `firstResultAvailable` guard in `index.ts` now filters with `&& result === 'reachable'` so unreachable results never trigger early calls."

packages/@webex/plugin-meetings/src/reachability/clusterReachability.ts
venky-mediboina: yeah, firstReachableResult was just the first in array order (creation order), not the fastest. Fixed it with renaming to bestResult and now picks the one with the minimum latencyInMilliseconds across all per-URL peer connections.

venky-mediboina:
@claude review

venky-mediboina:
@codex review

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -43,3 +86,40 @@ export function convertStunUrlToTurnTls(stunUrl: string) {
...
+ * @param {boolean} [includeDomains=false] - Whether to include domain names in details
+ * @returns {SubnetDetail[]} subnet details marked as unreachable
+ */
+export function prepopulateSubnetDetails(urls: string[], includeDomains = false): SubnetDetail[] {
+  const details: SubnetDetail[] = [];
+  const seenHostPorts = new Set<string>();
+
+  urls.forEach((url) => {
+    const {host, port, isIp} = parseIceServerUrl(url);
+    if (host && port) {
```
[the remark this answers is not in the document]
venky-mediboina: We won't get any port-less domain URLs. The guard `if (host && port)` will skip port-less URLs, so a server like stun:server.example.com would get no unreachable detail entry. However, Orpheus always returns URLs with explicit ports (e.g., :5004, :443), so this path is never hit in real usage. And since SubnetDetail.port is a required number field, there's no safe value to assign for a missing port anyway. So don't need to change anything.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -252,18 +313,6 @@ export class ReachabilityPeerConnection extends EventsScope {
       if (publicIp) {
         result.clientMediaIPs = [publicIp];
       }
-
-      this.emit(
```
marcin-bazyl, line 256: the resultReady event was emitted immediately and now it is only emitted at the end. As a result the "reachability:firstResultAvailable" event will only be emitted at the end and not immediately when the first good result is found - the web app relies on this event being emitted immediately, so we can't have that change of behaviour

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -44,15 +53,27 @@ export class ReachabilityPeerConnection extends EventsScope {
...
       udp: {
         result: 'untested',
+        details: enablePerUdpUrlReachability
+          ? prepopulateSubnetDetails(clusterInfo.udp, true)
+          : undefined,
       },
       tcp: {
         result: 'untested',
+        details: enablePerUdpUrlReachability
+          ? prepopulateSubnetDetails(clusterInfo.tcp)
```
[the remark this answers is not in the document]
venky-mediboina: We are not tracking hostname-based URLs for TCP and xTLS. So its fine for now.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -214,32 +237,72 @@ export class ReachabilityPeerConnection extends EventsScope {
...
-   * @param {string|null} [publicIp]
-   * @param {string|null} [serverIp]
+   * Emits resultReady events for all protocols that have URLs.
+   * @returns {void}
+   */
+  private emitResultsForAllProtocols(): void {
+    PROTOCOLS_LIST.forEach((protocol) => {
+      const result = this.result[protocol];
+      if (result.result === 'untested' || this.emittedProtocols.has(protocol)) {
+        return;
```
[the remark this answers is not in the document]
venky-mediboina: The concern is that early emission might cause partial TCP/xTLS details to be stored. But this doesn't happen because: `areAllResultsReady` requires all clusters/protocols to report before the final store, and the final re-read at that point calls `getResult()` which picks up any details added after the early emission. So, no fix needed.

marcin-bazyl:
I've asked AI to check if the existing flows (with enablePerUdpUrlReachability = false in config) are affected by this PR. This is what it came back with:

| # | Difference | Severity | Affects given config? |
|---|-----------|----------|----------------------|
| 1 | Unreachable `resultReady` events now emitted (via `emitResultsForAllProtocols`) | **High** | Yes — changes result counting, timing of everything |
| 2 | `areAllResultsReady()` completes with unreachable clusters | **High** | Yes — `'reachability:done'` fires earlier |
| 3 | Potential double `'reachability:done'` + double `sendMetric` on overallTimer | **High (Bug)** | Yes — when some PCs haven't completed ICE when timer fires |
| 4 | `minLatency` added to backend data format | **Medium** | Yes — always added for reachable results |
| 5 | IPv6/domain name parsed from STUN URLs → more reachedSubnets | **Medium** | Yes — different serverIp values + subnet events |
| 6 | `'reachability:firstResultAvailable'` condition change | **Low** | Functionally same timing, semantic change only |
| 7 | Extra `storeResults()` write | **Low** | Yes — redundant I/O |
| 8 | New `getAllClustersInfo()` method | **Low** | Additive only |


I think we can ignore points 4, 6 7 and 8, but please check the other points.

venky-mediboina:
Fixed abouve issue:
Points 1 & 2: Guarded emitResultsForAllProtocols() with if (this.enablePerUdpUrlReachability) at both call sites (abort + ICE gathering complete) in reachabilityPeerConnection.ts. Default mode no longer emits unreachable resultReady events, preserving the existing timer-based completion flow.
Point 3: Added a reachabilityDoneEmitted boolean flag in index.ts to guard all 3 reachability:done emission sites, preventing double-emit from the abort → timer race.
Point 5: Gated serverPort (UDP srflx) and relay candidate.port (TCP/xTLS) behind enablePerUdpUrlReachability in reachabilityPeerConnection.ts, so subnet details are not populated in default mode.

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -802,15 +819,19 @@ export default class Reachability extends EventsScope {
...
-        {
-          file: 'reachability',
-          function: 'overallTimer timeout',
-        },
-        'reachability:done',
-        {}
-      );
-      this.sendMetric();
+      // Guard: abort may have triggered resultReady events that already emitted reachability:done
+      if (!this.reachabilityDoneEmitted) {
```
marcin-bazyl: since we have to make sure reachabilityDoneEmitted is set to true whenever we emit that event and we already have 3 places where we do this, it would be good to have a method like emitReachabilityDoneEvent() that would set the flag and emit the event. This avoids any risk of someone adding another place where they emit the event and not set the flag.

Also, shouldn't we be checking that reachabilityDoneEmitted flag is false in all the places where we emit the event? it's only checked here. That check should also just be in the new method emitReachabilityDoneEvent()

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -999,11 +1037,31 @@ export default class Reachability extends EventsScope {
...
+              results[clusterKey].udp.details = finalResult.udp.details;
+            }
+            if (finalResult.tcp.details) {
+              results[clusterKey].tcp.details = finalResult.tcp.details;
+            }
+            if (finalResult.xtls.details) {
+              results[clusterKey].xtls.details = finalResult.xtls.details;
+            }
+          });
+          await this.storeResults(results);
```
marcin-bazyl: nitpick: storeResults() is already called in line 1044, the code could be written in a way that we call it only once instead of twice for the `areAllResultsReady` case

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -334,28 +458,60 @@ export class ReachabilityPeerConnection extends EventsScope {
...
-            const match = (e.candidate as any).url.match(stunServerUrlRegex);
-            serverIp = match && match[1];
+      if (e.candidate.type === CANDIDATE_TYPES.SERVER_REFLEXIVE) {
+        // Extract server info from candidate URL (if available)
+        const candidateWithUrl = e.candidate as RTCIceCandidate & {url?: string};
+        const {host: serverIp, port: serverPort} = candidateWithUrl.url
+          ? parseIceServerUrl(candidateWithUrl.url)
+          : {host: undefined, port: undefined};
+
+        const isPerUrlMode = this.numUdpUrls === 1;
```
marcin-bazyl: this is not safe, we might have just a single udp url in web app and still not want do have the "separate peer connection per udp url" functionality. Better to use `this.enablePerUdpUrlReachability`

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -334,28 +458,60 @@ export class ReachabilityPeerConnection extends EventsScope {
...
+        // Only pass serverPort when enablePerUdpUrlReachability is true —
+        // in standard mode there are no subnet details to update.
+        const effectiveServerPort = this.enablePerUdpUrlReachability ? serverPort : undefined;
+        if (isPerUrlMode || this.result.udp.result !== 'reachable') {
```
marcin-bazyl: I don't understand this if condition, previously we always called saveResult. Can you explain the purpose of this condtion?

packages/@webex/plugin-meetings/src/reachability/index.ts
marcin-bazyl: issue: this may update the resultsCount for unreachable results, right? and the resultsCount is only supposed to count reachable results

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: The condition tried to avoid calling `saveResult` redundantly after UDP was already reachable, but it was unnecessary — `saveResult` already handles this internally, since its own guard only marks reachable on the first call and routes all subsequent calls to addPublicIp automatically. So the this was removed and saveResult is now called unconditionally.

packages/@webex/plugin-meetings/src/reachability/index.ts
venky-mediboina: yeah, we can use a common `emitReachabilityDoneEvent()` fn. thanks for the suggestion, used same.

packages/@webex/plugin-meetings/src/reachability/index.ts
venky-mediboina: consolidated to call storeResults() at only one place.

packages/@webex/plugin-meetings/src/reachability/index.ts
venky-mediboina: Earlier thought for each connection i.e even its reachable or not reachable, counting it. Changed with condition.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Right, used `enablePerUdpUrlReachability`, removed isPerUrlMode  var.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -334,28 +459,50 @@ export class ReachabilityPeerConnection extends EventsScope {
...
+          serverIp,
+          effectiveServerPort
+        );
 
-          this.saveResult('udp', latencyInMilliseconds, e.candidate.address, serverIp);
+        this.determineNatTypeForSrflxCandidate(e.candidate);
+      } else if (e.candidate.type === CANDIDATE_TYPES.RELAY) {
+        // Use candidate.url (Chrome) for protocol detection; fall back to ephemeral e.candidate.port in Firefox (best-effort).
+        const relayCandidateWithUrl = e.candidate as RTCIceCandidate & {url?: string};
+        const configuredPort = relayCandidateWithUrl.url
```
marcin-bazyl: I don't understand what is the benefit of getting the port from `relayCandidateWithUrl.url`. The existing code that uses `e.candidate.port` to detect TLS vs TCP has been working fine in all browsers and is simpler. Is there any edge case where network checker needs to use the port from the url because `e.candidate.port` doesn't work? If not, then I'd rather keep existing code and minimize the changes in this PR.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
```diff
@@ -375,14 +522,18 @@ export class ReachabilityPeerConnection extends EventsScope {
 
     // Initialize this.result as saying that nothing is reachable.
     // It will get updated as we go along and successfully gather ICE candidates.
+    // Preserve existing details (pre-populated or undefined based on enableSubnetDetails flag).
```
marcin-bazyl: I don't see any "enableSubnetDetails" flag anywhere in the code, is this comment valid?

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -1,4 +1,47 @@
...
+  const [, ipv6, hostOrIpv4, portStr] = match;
+  const port = portStr ? parseInt(portStr, 10) : undefined;
+
+  // IPv6 in brackets - validate and return without brackets
+  if (ipv6) {
+    if (!Address6.isValid(ipv6)) {
+      return {isIp: false};
+    }
+
+    return port ? {host: ipv6, port, isIp: true} : {host: ipv6, isIp: true};
```
marcin-bazyl: wouldn't a simpler line like this do the same?
`return {host: ipv6, port, isIp: true}`

same in line 39

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -1,4 +1,47 @@
...
+import {STUN_TURN_URL_REGEX} from '../constants';
+import {SubnetDetail} from './reachability.types';
+
+/**
+ * Parses a STUN/TURN URL to extract host and port information.
+ * Returns host with brackets removed for IPv6, and isIp flag.
+ * @param {string} url - The STUN/TURN URL to parse (e.g., 'stun:server.com:3478', 'stun:[2402:2500::1]:5004')
+ * @returns {object} Object containing host, port, and isIp flag
+ */
+export function parseIceServerUrl(url: string): {
```
marcin-bazyl, line 12: a new exported util function like this should have its own unit tests

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -43,3 +86,40 @@ export function convertStunUrlToTurnTls(stunUrl: string) {
...
+
+/**
+ * Pre-populates subnet details from STUN URLs.
+ * By default, only includes IP addresses (IPv4 or IPv6), skips domain names.
+ * When includeDomains is true, also includes domain names (used for UDP per-URL mode).
+ * @param {string[]} urls - STUN URLs to extract hosts from
+ * @param {boolean} [includeDomains=false] - Whether to include domain names in details
+ * @returns {SubnetDetail[]} subnet details marked as unreachable
+ */
+export function prepopulateSubnetDetails(urls: string[], includeDomains = false): SubnetDetail[] {
```
marcin-bazyl, line 98: a new exported util function like this should have its own unit tests

packages/@webex/plugin-meetings/src/reachability/reachability.types.ts
```diff
@@ -6,19 +6,38 @@ export type Protocol = 'udp' | 'tcp' | 'xtls';
  * Events emitted by ReachabilityPeerConnection
  */
 export const ReachabilityPeerConnectionEvents = {
-  resultReady: 'resultReady', // emitted when successfully reached over a protocol
+  resultReady: 'resultReady', // emitted when a reachability result is available for a protocol (reachable or unreachable)
```
marcin-bazyl: this comment is not entirely correct, right? when per url mode is disabled, this event is only emitted for reachable urls, right?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -210,6 +214,12 @@ describe('ClusterReachability', () => {
       const udpRpc1 = (perUdpClusterReachability as any).reachabilityPeerConnectionsForUdp[0];
       const udpRpc2 = (perUdpClusterReachability as any).reachabilityPeerConnectionsForUdp[1];
 
+      // Stub getResult so aggregateUdpResults() sees the correct state
+      sinon.stub(udpRpc1, 'getResult').returns({
```
marcin-bazyl, line 214: why do we need this? aggregateUdpResults() is not supposed to be called in default mode and these tests use the defaultmode, right?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -225,7 +235,7 @@ describe('ClusterReachability', () => {
       });
 
       assert.equal(udpEvents.length, 1);
-      assert.equal(udpEvents[0].latencyInMilliseconds, 50);
+      assert.equal(udpEvents[0].protocol, 'udp');
```
marcin-bazyl: why the existing check has been replaced?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -528,22 +542,20 @@ describe('ClusterReachability', () => {
       await clock.tickAsync(100);
       fakePeerConnection.onicecandidate({candidate: {type: 'srflx', address: 'somePublicIp'}});
 
-      // check the right event was emitted
+      // resultReady is emitted immediately for the first reachable protocol (UDP)
       assert.equal(emittedEvents[Events.resultReady].length, 1);
-      assert.deepEqual(emittedEvents[Events.resultReady][0], {
```
marcin-bazyl, line 533: why this check has been removed?

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -639,29 +652,22 @@ describe('ClusterReachability', () => {
       await clock.tickAsync(10);
       fakePeerConnection.onicecandidate({candidate: {type: 'srflx', address: 'somePublicIp1'}});
 
-      // check events emitted: there should be a resultReady and no clientMediaIpsUpdated
+      // resultReady is emitted immediately for the first reachable protocol (UDP)
       assert.equal(emittedEvents[Events.resultReady].length, 1);
-      assert.deepEqual(emittedEvents[Events.resultReady][0], {
```
marcin-bazyl, line 644: same as above

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
```diff
@@ -639,29 +652,22 @@ describe('ClusterReachability', () => {
...
       assert.equal(emittedEvents[Events.clientMediaIpsUpdated].length, 0);
-      resetEmittedEvents();
 
       await clock.tickAsync(10);
       fakePeerConnection.onicecandidate({candidate: {type: 'srflx', address: 'somePublicIp1'}});
 
-      // no new event was emitted
-      assert.equal(emittedEvents[Events.resultReady].length, 0);
+      // no new event was emitted (duplicate IP)
+      assert.equal(emittedEvents[Events.resultReady].length, 1);
```
marcin-bazyl: why this change? we should minimize any changes to existing tests so that we can easily see that changes in this PR are not breaking them

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: Yeah, reverted to using `e.candidate.port` for tcp/xtls protocol detection. The `relayCandidateWithUrl.url` parsing has been removed to keep the change minimal. Done.

packages/@webex/plugin-meetings/src/reachability/reachabilityPeerConnection.ts
venky-mediboina: The comment was stale. Updated it to "Preserve existing details (pre-populated when enablePerUdpUrlReachability is true)." Done. Thanks.

packages/@webex/plugin-meetings/src/reachability/util.ts
venky-mediboina: Yes, simplified both branches. When `port` is undefined the property is just set to undefined, which is equivalent for callers. Thanks.

packages/@webex/plugin-meetings/src/reachability/util.ts
venky-mediboina: Done, added a dedicated #parseIceServerUrl() describe block in util.ts tests file

packages/@webex/plugin-meetings/src/reachability/util.ts
venky-mediboina: Done, added a #prepopulateSubnetDetails() describe block in the same file.

packages/@webex/plugin-meetings/src/reachability/reachability.types.ts
venky-mediboina: yeah, differentiated for both cases PerURLCheck and Non-PerURLCheck.

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: Reverted, restored to the original assertion shape.

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: Reverted, restored the original.

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: reverted

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: reverted all such instances/changes to keep minimal chages.

packages/@webex/plugin-meetings/test/unit/spec/reachability/clusterReachability.ts
venky-mediboina: This test uses per‑URL mode (3rd arg = true), not default mode. The stub is needed because aggregateUdpResults() reads from rpc.getResult() — without it, udpRpc1 would still be 'untested' and the latency check would fail. Updated the inline comment to make this clearer.

PR #4673, opened 2026-01-30
feat: added allowPublishMediaInLobby to AddMediaOptions

marcin-bazyl:
# COMPLETES #[SPARK-681282](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-681282)

There is a possible race condition:
When the app calls Meeting.joinWithMedia() and passes some local streams, it is possible that SDK is notified about user being placed in the lobby before the media connection is actually established. This means that when SDK forwards the "entered lobby" event to the app, the app doesn't do anything about local streams, because there is no media connection yet created so it thinks that there is nothing to unpublish. However, the media creation triggered by joinWithMedia() call continues in parallel to all this and eventually will publish the streams that were given in the arguments. So we end up with local streams published while we are in a lobby. This is not the end of the world, as backend does remote mute on us anyway while in lobby, but when the user is admitted to the meeting, sometimes we have big delays in getting Locus event about it (or even the event is missed completely) and then if we have published streams, remote participants will see our video while our app thinks we're still in the lobby.

Sequence of things:
1. Meeting.joinWithMedia() called with some local streams - at this point we don't know if we'll be in the lobby or not
3. Locus tells SDK that we're in lobby
4. SDK tells the app that we're in the lobby -> app ignores it, because there is no media connection yet, so nothing to do with local streams
5. Meeting.addMediaInternal() is called from Meeting.joinWithMedia()
6. local streams from point 1 are published

Result: we're in the lobby with published local streams

Added a new parameter to AddMediaOptions to indicate that the client app doesn't want to have streams published in a lobby - this allows SDK to have logic to just not do anything with passed in local streams if we are in the lobby at the point of media connection creation.

To test the fix manually with the web app:
- you need to hack joinWithMedia() and add a delay before calling addMediaInternal()
- hack SDK or web app to ignore or delay the event about being admitted from lobby into the meeting

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual testing with web app

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -250,6 +250,7 @@ export type AddMediaOptions = {
   remoteMediaManagerConfig?: RemoteMediaManagerConfiguration; // applies only to multistream meetings
   bundlePolicy?: BundlePolicy; // applies only to multistream meetings
   allowMediaInLobby?: boolean; // allows adding media when in the lobby
+  allowPublishMediaInLobby?: boolean; // allows publishing media when in the lobby, if not specified, default value false is used
   additionalMediaOptions?: AdditionalMediaOptions; // allows adding additional options like send/receive audio/video
 };
 
@@ -623,6 +624,13 @@ export default class Meeting extends StatelessWebexPlugin {
   keepAliveTimerId: NodeJS.Timeout;
   lastVideoLayoutInfo: any;
   locusInfo: any;
+  // this group of properties is populated via updateMeetingObject() that's registered as a callback with LocusInfo
+  isUserUnadmitted?: boolean;
+  joinedWith?: any;
+  selfId?: string;
+  roles: any[];
+  // ... there is more ... see SelfUtils.parse()
+  // end of the group
   locusMediaRequest?: LocusMediaRequest;
   mediaProperties: MediaProperties;
   mediaRequestManagers: {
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -623,6 +624,13 @@ export default class Meeting extends StatelessWebexPlugin {
   keepAliveTimerId: NodeJS.Timeout;
   lastVideoLayoutInfo: any;
   locusInfo: any;
+  // this group of properties is populated via updateMeetingObject() that's registered as a callback with LocusInfo
+  isUserUnadmitted?: boolean;
```
marcin-bazyl, line 628: the existing `@ts-ignore` in line 8049 was really annoying me, especially that I needed to use isUserUnadmitted in the new code so would have to add another ts-ignore, so I've added the type declaration for isUserUnadmitted and moved declarations of a few other props that come from the same source (SelfUtils.parse()) here - I really don't like how they are set via updateMeetingObject() but to fix this would require a bigger refactor, out of scope for this PR

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -8026,6 +8031,7 @@ export default class Meeting extends StatelessWebexPlugin {
       remoteMediaManagerConfig,
       bundlePolicy = 'max-bundle',
       additionalMediaOptions = {},
+      allowPublishMediaInLobby = false,
```
marcin-bazyl, line 8034: by having the default `false` we're changing existing API behaviour, but I think it's worth as I don't think many other SDK clients would want to have the streams published while in the lobby, but in theory they might, that's why I've added the flag, so they can still override this behavior

maxinteger (approved):

robstax (approved):

PR #4684, opened 2026-02-02
fix(calling): new method getDevices added

Kesari3008:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7541 

Needed to fetch device list for a user if needed to be cleared during reload or refreshes.

Added new API: getDevices() under CallingClient module to fetch list of registered devices for a user. It helps deleting registrations if not required or handle duplicate registration cases more gracefully.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Using SDK Kitchen Sink App tested it.
Screenshot 2026-02-04 at 9 46 37 PM
Network Logs are attached in the Jira

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -718,6 +721,67 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+   * Resolve the base Mobius URL for device APIs.
+   */
+  private getMobiusUrl(): string {
+    const activeMobiusUrl = Object.values(this.lineDict)[0]?.getActiveMobiusUrl?.();
+
+    if (activeMobiusUrl) {
+      return activeMobiusUrl;
+    }
+
+    if (this.primaryMobiusUris.length) {
```
arun3528: @Kesari3008  why do do these check , from what i read  we have to get active mobius from all of the mobius server then do a delete

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -718,6 +721,67 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      return `${this.mobiusHost}${URL_ENDPOINT}`;
+    }
+
+    throw new Error('Mobius URL is not available to fetch devices');
+  }
+
+  /**
+   * Fetches the list of devices for a given user from Mobius.
+   */
+  public async getDevices(userId: string): Promise<DeviceType[]> {
```
rsarika: Do we need this param?
why not this.webex.internal.device.userId ?

rsarika (approved):

rsarika:
Now If we have to de-register user/developer need to run through this list(DeviceType[]) and call de-register.

Why not give this as a simple one more method deRegisterAllDevices() ? just thinking it out loud.

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Updated the approach now

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: Updated it

Kesari3008:
> Now If we have to de-register user/developer need to run through this list(DeviceType[]) and call de-register.
> 
> Why not give this as a simple one more method deRegisterAllDevices() ? just thinking it out loud.

This is somehting we can consider for future. At the moment, multiple devices are not supported for one user. But registration is also somehting not implemented so allowing to deregiter for multiple devices would be something not required

arun3528 (approved):
@Kesari3008  @rsarika  i am ok with this approach of saving the the url in local storage and then cleaning it up .

but if its active why not reclaim it , why do we have to delete it 

my idea of the whole cleanup was that the user does not know which mobius they are active on so they wil have to do feat all then see if there is a active one available then do a delete on that , when we do automation local storage is not shared across PRs

PR #4700, opened 2026-02-10
fix(calling): clears a call keepalive interval before creating one

rarajes2:
# COMPLETES #< Ad-hoc >

After the investigation throughout the SDK and backend services we have found out that during the optimization of SDK we removed the extra interval clearing logic which was being invoked every time a re-invite was being sent from the backend with or without the change in the OFFER sdp. This caused to create an extra call keepalive interval running every 10 mins and for a longer call going beyond 30 mins where SDK receives a re-invite every 30 mins from the backend services so it started having multiple keepalives with the same callId. Now once the current call is ended and a new call gets established sdk has previous call keepalives with the old callId along with the new keepalive interval running. While the current call is still ongoing, the old keepalive request goes to the backend and receives a 404 response and as a result sdk clears the current call, causing it to unexpectedly drop the ongoing call.

Here Mobius is receiving RE_INVITE from SSE every 30 mins which it is forwarding to the SDK.

Clearing the existing call keepalive interval before create a new one. This fixes any unexpected invite/re-invite causing to create a new call keepalive intervals.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/calling/call.ts
+++ b/packages/calling/src/CallingClient/calling/call.ts
@@ -1428,7 +1428,7 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
       this.mediaConnection.close();
       log.info('Closing media channel', {
         file: CALL_FILE,
-        method: METHODS.HANDLE_OUTGOING_CALL_DISCONNECT,
+        method: METHODS.HANDLE_INCOMING_CALL_DISCONNECT,
       });
     }
 
@@ -1567,6 +1567,8 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
       file: CALL_FILE,
       method: 'scheduleCallKeepaliveInterval',
     };
+
+    clearInterval(this.sessionTimer);
 
     this.sessionTimer = setInterval(async () => {
       try {
--- a/packages/calling/src/CallingClient/calling/call.test.ts
+++ b/packages/calling/src/CallingClient/calling/call.test.ts
@@ -1002,7 +1002,9 @@ describe('State Machine handler tests', () => {
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(clearInterval).toHaveBeenCalledTimes(1);
+    // clearInterval is called twice: once in scheduleCallKeepaliveInterval (at start)
+    // and once in handleCallKeepaliveError when clearing timer due to error
+    expect(clearInterval).toHaveBeenCalledTimes(2);
     expect(funcSpy).toBeCalledTimes(1);
     expect(emitSpy).toHaveBeenCalledWith(CALL_EVENT_KEYS.DISCONNECT, call.getCorrelationId());
   });
@@ -1034,7 +1036,9 @@ describe('State Machine handler tests', () => {
      */
     await flushPromises(2);
 
-    expect(clearInterval).toHaveBeenCalledTimes(1);
+    // clearInterval is called twice: once in scheduleCallKeepaliveInterval (at start)
+    // and once in handleCallKeepaliveError when clearing timer due to error
+    expect(clearInterval).toHaveBeenCalledTimes(2);
     expect(funcSpy).toBeCalledTimes(1);
   });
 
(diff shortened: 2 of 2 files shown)


packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -1568,6 +1568,8 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
       method: 'scheduleCallKeepaliveInterval',
     };
 
+    clearInterval(this.sessionTimer);
```
adhmenon, line 1571: Question - do we need a null check `if` case for the session timer?
I remember it was there in the older code, but yeah might be redundant.

adhmenon (approved):
LGTM. One comment on the check.

adhmenon (approved):
LGTM. One comment on the check.

packages/calling/src/CallingClient/calling/call.ts
rarajes2: It's not required. `clearInterval(undefined)` does not throw any error

Kesari3008, line 1571: What happens if there is no existing timer running  ?
We should check if there is an existing timer running and if there is then we should skip creating new atleast for this case. Mid call clearing existing timer and running a new timer feels unnecessary or even if want to clear and start new, it should be done as below:

  if (this.sessionTimer) {
      clearInterval(this.sessionTimer);
    }

Same is followed everywhere else in code

rarajes2: we don't have to add the if condition here, it won't make any difference. if there is no value in the `this.sessionTimer` then clearInterval will not clear anything and it will also not throw any error.

Kesari3008: OK in that case it's ok not to add null check but what I am trying to say is we should not be clearing the interval at all. In a call there can be renegotiation happening multliple times so everytime we finish the renegotiation we will be destroying a  valid running timer and creating new one. We should not create new timer for renegotiations happening on the same call.

rarajes2: In the old code we had the same logic  - https://github.com/webex/webex-js-sdk/blob/wxc-disconnect/packages/calling/src/CallingClient/calling/call.ts#L1523

mkesavan13: @rarajes2 - Since we were discussing in the office and I also had a chat with Priya before that, I think I just understood. What if an actual renegotiation happened instead of the routine "30-min no-sdp change reinvite" during the call? If that's the case, let's say the timer has run for 5 mins and then a legitimate media renegotiation happens? At that point you'll be clearing 5 min interval and add a 10 min one which means now, the keepalive interval would become 15 min instead of the originally remaining 5 mins leading to a keepalive miss for Mobius.

Kesari3008 (approved):

packages/calling/src/CallingClient/calling/call.ts
rarajes2: **Here is what actually happens:**

When re-invite comes leading to the re-negotiation, the sdk triggers the `handleCallEstablished()` which signifies the call is established which then further triggers the keepalive interval logic. Now at this point it is confirm that the call is active which implies the result of call keepalive check, meaning the call keepalive check has already happened at that moment making the remaining time of the keepalive obsolete and from there we create another interval which starts checking every 10 mins from that moment.

@Kesari3008 @mkesavan13 - Correct me if I am wrong.

mkesavan13: @rarajes2 - Isn't that for backend to decide? Backend keeps track of the timer as well and we do not know if it is resetting the timer when a media renegotiation happens. 

If it is not resetting, whatever I mentioned in the earlier comment is true but, if the reset happens, then this logic is good to go. Can we check with Mobius once on this and close this out soon?

rarajes2: Discussed with Mobius team. Mobius runs a timer expecting keepalive from the SDK and if it does not receive any keepalive for 15 mins, it clears the call.

During re-negotiation it **resets** the timer making our logic work in sync.

I think we can go ahead and merge the PR.

PR #4761, opened 2026-03-11
chore(contact-center): add spec-drift detection commands and pre-commit hook

ciscoRankush:
# COMPLETES N/A — Internal tooling addition

Adding automated SDD (Spec-Driven Development) documentation drift detection tooling to ensure ai-docs stay in sync with source code. This provides commands and a pre-commit hook for validating documentation accuracy against the actual codebase.

- **`.claude/commands/spec-drift.md`** — Full scan command that auto-discovers all ai-docs folders, spawns parallel checker agents, and validates 7 drift categories (file tree, method/API, types, events, architecture, links, code examples)
- **`.claude/commands/spec-drift-changed.md`** — Incremental scan that only validates ai-docs affected by staged/unstaged file changes, with verification marker creation
- **`.claude/hooks/check-ai-docs-drift.sh`** — Pre-commit blocking hook that prevents committing when ai-docs drift is unverified

### Change Type

- [x] Tooling change

- Verified spec-drift command discovers all ai-docs folders and produces a categorized report
- Verified spec-drift-changed detects changed files and maps them to their ai-docs
- Verified pre-commit hook blocks commits with unverified ai-docs changes (exit code 2)
- Verified hook passes after spec-drift-changed creates verification marker
- Verified hook is transparent for commits without ai-docs-related changes

  - [x] Other - Claude Code (Anthropic)
- [x] This PR is related to
  - [x] Automation

### I certified that

### DIFF
--- a/.claude/hooks/check-ai-docs-drift.sh
+++ b/.claude/hooks/check-ai-docs-drift.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# PreToolUse hook: Block git commit if contact-center code/docs changed without spec-drift verification
+#
+# Flow:
+#   1. Read stdin to get the Bash command (JSON at tool_input.command)
+#   2. If command is NOT git commit -> exit 0 (allow immediately)
+#   3. Check if ANY staged files are under packages/@webex/contact-center/
+#   4. If none -> exit 0 (allow)
+#   5. Check for verification marker (created by /spec-drift-changed)
+#   6. If marker exists -> delete it, exit 0 (allow, one-time pass)
+#   7. If no marker -> exit 2 (BLOCK, instruct to run /spec-drift-changed)
+
+CC_PKG="packages/@webex/contact-center"
+
+# Read tool input from stdin (JSON with tool_input.command)
+INPUT=$(cat)
+COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_input',{}).get('command',''))" 2>/dev/null)
+
+# Only gate git commit commands
+case "$COMMAND" in
+  git\ commit*) ;;  # Continue to check
+  *) exit 0 ;;      # Not a commit, allow immediately
+esac
+
+# Get staged files under the contact-center package
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^${CC_PKG}/")
+
+if [ -z "$STAGED_CC" ]; then
+  exit 0  # No contact-center files staged, allow commit
+fi
+
+# Compute hash of all staged contact-center files
+HASH=$(echo "$STAGED_CC" | sort | shasum | cut -d' ' -f1)
+MARKER="/tmp/.spec-drift-verified-${HASH}"
+
+if [ -f "$MARKER" ]; then
+  rm -f "$MARKER"  # One-time use
+  exit 0           # Verified, allow commit
+fi
+
+# Block the commit
+echo "BLOCKED: contact-center files are staged but ai-docs have not been verified for spec drift."
+echo ""
+echo "Staged contact-center files:"
+echo "$STAGED_CC" | sed 's/^/  - /'
+echo ""
+echo "Run /spec-drift-changed to validate ai-docs against source code before committing."
+echo "The command will check documentation accuracy and create a verification marker."
+exit 2  # Exit code 2 = blocking error in Claude Code hooks
--- a/.claude/hooks/check-ai-docs-drift.sh
+++ b/.claude/hooks/check-ai-docs-drift.sh
@@ -3,23 +3,37 @@
 #
 # Flow:
 #   1. Read stdin to get the Bash command (JSON at tool_input.command)
-#   2. If command is NOT git commit -> exit 0 (allow immediately)
-#   3. Check if ANY staged files are under packages/@webex/contact-center/
-#   4. If none -> exit 0 (allow)
-#   5. Check for verification marker (created by /spec-drift-changed)
-#   6. If marker exists -> delete it, exit 0 (allow, one-time pass)
-#   7. If no marker -> exit 2 (BLOCK, instruct to run /spec-drift-changed)
+#   2. If python3 unavailable or JSON parse fails -> exit 2 (fail-closed)
+#   3. If command is NOT git commit -> exit 0 (allow immediately)
+#   4. Check if ANY staged files are under packages/@webex/contact-center/
+#   5. If none -> exit 0 (allow)
+#   6. Check for verification marker (created by /spec-drift-changed)
+#   7. If marker exists -> exit 0 (allow — marker stays until content changes)
+#   8. If no marker -> exit 2 (BLOCK, instruct to run /spec-drift-changed)
 
 CC_PKG="packages/@webex/contact-center"
+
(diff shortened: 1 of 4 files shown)


.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+CC_PKG="packages/@webex/contact-center"
+
+# Read tool input from stdin (JSON with tool_input.command)
+INPUT=$(cat)
+COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_input',{}).get('command',''))" 2>/dev/null)
+
+# Only gate git commit commands
+case "$COMMAND" in
+  git\ commit*) ;;  # Continue to check
+  *) exit 0 ;;      # Not a commit, allow immediately
```
Shreyas281299: ### HIGH: Fail-Open Bypass When Python3 Fails or JSON is Malformed

The hook extracts the command string via `python3 -c "..." 2>/dev/null` (line 17). If `python3` is unavailable or the JSON input is malformed, `COMMAND` becomes empty. The `case` statement then falls through to `*) exit 0 ;;` here, **silently allowing unverified commits**.

**Fix:** The hook should fail-closed — if it cannot determine the command, it should block (`exit 2`) rather than allow (`exit 0`). Add a check after the python3 call to verify `COMMAND` was set, and default to blocking if parsing fails.

```bash
COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_input',{}).get('command',''))" 2>/dev/null)
if [ -z "$COMMAND" ]; then
  echo "ERROR: Could not parse tool input. Blocking commit as a safety measure."
  exit 2
fi
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Get staged files under the contact-center package
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^${CC_PKG}/")
+
+if [ -z "$STAGED_CC" ]; then
+  exit 0  # No contact-center files staged, allow commit
+fi
+
+# Compute hash of all staged contact-center files
+HASH=$(echo "$STAGED_CC" | sort | shasum | cut -d' ' -f1)
```
Shreyas281299: ### HIGH: Verification Marker Based Only on File Paths, Not Content

The marker hash is computed from staged file **paths** only (`sort | shasum`), not file content. This means:
1. Run `/spec-drift-changed` → marker created
2. Modify and re-stage the same files with different content
3. Commit succeeds because paths haven't changed → marker still matches
4. The drift check was run against **stale content**

Additionally, markers have **no TTL** — a marker created hours or days ago still passes validation.

**Fix:** Hash should incorporate content (e.g., `git diff --cached -- "$CC_PKG" | shasum` or blob hashes). Consider adding a timestamp check or TTL mechanism.

```bash
# Hash content instead of just paths
HASH=$(git diff --cached -- "$CC_PKG" | shasum | cut -d' ' -f1)
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+MARKER="/tmp/.spec-drift-verified-${HASH}"
+
+if [ -f "$MARKER" ]; then
+  rm -f "$MARKER"  # One-time use
```
Shreyas281299: ### HIGH: Marker Consumed Before Commit Completes

This is a PreToolUse hook — it runs **before** `git commit` executes. The marker is deleted here immediately upon verification. If the subsequent `git commit` fails (e.g., a lint pre-commit hook rejects it, merge conflict), the marker is already consumed. The developer must re-run `/spec-drift-changed` just to retry the same commit.

**Fix:** Clean up the marker after a successful commit (e.g., via a PostToolUse hook), or use a mechanism that survives failed attempts. For example:
- Add a companion PostToolUse hook that deletes the marker after successful commit
- Or keep the marker and let it be invalidated naturally when the staged content changes (if using content-based hashing from issue #2)

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# PreToolUse hook: Block git commit if contact-center code/docs changed without spec-drift verification
```
Shreyas281299, line 2: ### HIGH: Hook is Dead Code Without Settings Configuration

This hook file exists but there is no corresponding `settings.json` entry to register it as a Claude Code PreToolUse hook. Without this configuration, the hook does nothing — the "pre-commit blocking hook" described here doesn't actually block anything.

**Fix:** Either:
1. Include the `settings.json` change in this PR to register the hook, e.g.:
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "hooks": [".claude/hooks/check-ai-docs-drift.sh"]
      }
    ]
  }
}
```
2. Or add explicit documentation in the PR description with the exact JSON snippet needed for manual configuration.

Shreyas281299 (changes requested):
## Review Summary

The concept is sound and addresses a real need for keeping ai-docs in sync with source code. However, the pre-commit hook has several reliability issues that should be addressed before merging.

| Severity | Count |
|----------|-------|
| HIGH | 4 |
| MEDIUM | 4 |
| LOW | 3 |

Posting the 4 HIGH severity issues as inline comments below. The MEDIUM and LOW issues are documented in detail in the full review report.

Shreyas281299 (commented):
## Code Review

Thorough review of the spec-drift detection tooling. Found 1 high-severity issue around hash coupling, 3 medium-severity robustness/design concerns, and 3 low-severity nitpicks. Overall the approach is sound and well-structured.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in ed3a6502. Added `.claude/settings.json` with the PreToolUse hook registration for Bash commands.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in ed3a6502. Added `python3` availability check (`command -v python3`) and a fail-closed guard — if `COMMAND` is empty after parsing, the hook now exits with code 2 (block) instead of 0 (allow).

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in ed3a6502. Hash now uses `git diff --cached -- "$CC_PKG"` (staged content) instead of just file paths. Also added `shasum || sha256sum` fallback for cross-platform portability. The matching hash logic in `spec-drift-changed.md` Step 5 was updated to match.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in ed3a6502. The marker is no longer deleted in the PreToolUse hook — it stays on disk and is naturally invalidated when staged content changes (since the content-based hash will differ). Failed commits no longer consume the marker.

Shreyas281299 (approved):

PR #4768, opened 2026-03-12
feat(changelog): add commit history section for package comparisons and enhance styling

vivekv1504:
# COMPLETES #< [SPARK-772395](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-772395) >

Vidcast Link:-https://app.vidcast.io/share/40aa3aaa-51cb-470a-850c-10015be09353

This pull request addresses
The Webex JS SDK Changelog portal had no way to view the commit history of a specific package between two SDK stable releases. When a user wanted to know what changed in, say, @webex/calling between 3.6.0-next.5 and 3.10.0-next.7, they had to manually open each version log file and scan through commits — across multiple intermediate stable versions (3.7.0, 3.8.0, 3.8.1, 3.9.0) — with no tooling to aggregate them.

Implemented a cross-stable commit history collection engine in docs/changelog/assets/js/app.js that:
Identifies all stable versions that sit between a user-selected base version (stableA) and target version (stableB) using semver-aware sorting
Fetches each intermediate stable's log file in sequence, reusing already-loaded changelogs for stableA and stableB to avoid redundant network requests
Applies position-based rules for which pre-release entries to include per stable:
start — from the user's base pre-release (e.g. 3.6.0-next.5) through all remaining pre-releases of that stable
middle — skips the exact stable entry (e.g. 3.7.0); collects ALL pre-releases of that intermediate stable
end — from next.1 up to the user's target pre-release (e.g. 3.10.0-next.7) inclusive
only — when both versions fall within the same stable file, collects the range within that single file
Renders results in a table with short-linked commit hash (linked to GitHub), commit message (with PR number auto-linked), and the pre-release version the commit belongs to
Displays a summary of how many commits were collected and which stable log files were traversed

Breaking behavior change: The full-version comparison mode (comparing all packages between two stable versions without selecting a specific package) has been removed

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [x] Other - cursor 
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


rarajes2:
@codex review

vivekv1504:
@codex review

docs/changelog/assets/js/app.js
```diff
@@ -1812,349 +1871,264 @@ const handleStableVersionChange = async () => {
...
-    
+
     // Toggle visibility
     if (searchForm) searchForm.classList.add('hide');
     if (comparisonForm) comparisonForm.classList.remove('hide');
     if (searchResults) searchResults.classList.add('hide');
     if (helperSection) helperSection.classList.add('hide');
-    
+
     populateComparisonVersions();
```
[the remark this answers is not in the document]
vivekv1504: done changes updated

docs/changelog/assets/js/app.js
```diff
@@ -1812,349 +1876,268 @@ const handleStableVersionChange = async () => {
...
+            alert('When comparing within the same stable version, please select both pre-release versions.');
+            return false;
+        }
+        if (versionASpecific === versionBSpecific) {
+            alert('Please select two different versions to compare.');
+            return false;
+        }
+        return true;
+    }
+    if (stableA > stableB) {
```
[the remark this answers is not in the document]
vivekv1504: done changes updated with code 
`const allSorted = sortStableVersions([stableA, stableB]);
if (allSorted[0] !== stableA) {
    alert(`Base version (${stableA}) must be older than target version (${stableB}). Please swap.`);
    return false;
}`

docs/changelog/index.html
```diff
@@ -78,15 +78,6 @@
...
             <form id="comparison-form" class="hide" method="GET">
-              <!-- Package Selection Row (first, required) -->
-              <div class="form-row" id="comparison-package-row">
-                <div class="form-group">
-                  <label for="comparison-package-select">Select Package <span class="required-asterisk">*</span></label>
-                  <select id="comparison-package-select" class="full-width" required>
-                    <option value="">Select a Package</option>
-                  </select>
-                </div>
-              </div>
```
Shreyas281299, line 89: Why did we remove this?

docs/changelog/index.html
```diff
@@ -104,11 +94,18 @@
                     <select id="version-b-select" class="full-width" required disabled>
                       <option value="">Select target version</option>
                     </select>
-                    <button type="button" id="clear-version-b-btn" class="btn btn-clear-circle" title="Clear target version" aria-label="Clear target version"><svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round"><path d="M18 6L6 18M6 6l12 12"/></svg></button>
```
Shreyas281299, line 107: Why did we remove this clear buttons?

docs/changelog/assets/js/app.js
```diff
@@ -33,15 +33,11 @@ const versionBPrereleaseSelect = document.getElementById('version-b-prerelease-s
 const prereleaseRow = document.getElementById('comparison-prerelease-row');
 const compareButton = document.getElementById('compare-button');
 const clearComparisonButton = document.getElementById('clear-comparison-button');
-const clearVersionABtn = document.getElementById('clear-version-a-btn');
-const clearVersionBBtn = document.getElementById('clear-version-b-btn');
```
Shreyas281299, line 37: Why did we remove these 2 buttons?

docs/changelog/assets/js/app.js
```diff
@@ -601,190 +597,9 @@ window.onhashchange = () => {
...
-// Global state for comparison mode
-let comparisonMode = false;
-
-/**
- * Extract all packages from a version changelog.
- * @param {Object} changelog - The changelog JSON for a version
- * @param {Object} specificVersions - Optional map of {packageName: specificVersion}
- * @returns {Object} - Map of {packageName: version}
- */
-const extractPackagesFromVersion = (changelog, specificVersions = null) => {
```
Shreyas281299, line 621: Why did we remov ethis functionality?

docs/changelog/assets/js/app.js
```diff
@@ -800,232 +615,17 @@ const showComparisonLoading = () => {
...
-   ============================================ */
-
-/**
- * DATA LAYER: Fetch and compare versions (Pure data logic, no DOM manipulation).
- * @param {string} versionA - Base version
- * @param {string} versionB - Target version
- * @returns {Promise<Object>} Comparison data with versionA, versionB, and comparisonData
- * @throws {Error} If fetch fails or comparison fails
- */
-const fetchAndCompareVersions = async (versionA, versionB) => {
```
Shreyas281299, line 825: Same here why are we removing this?

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1465,260 @@ const handleStableVersionChange = async () => {
...
     );
-    updateCompareButtonState();
-    setPackageSelection(enhancedParams.packageName);
 };
-
 /**
- * Handle standard comparison URL parameters on page load.
- * Version A/B are pre-filled but disabled until user selects a package (package is required).
- */
-const loadStandardComparisonFromURL = async (urlParams) => {
```
Shreyas281299, line 2102: Why are ew remove this?

Shreyas281299 (changes requested):
Code cahnges look good for the commit data but I dont understand why have we refactored the code and removed so much code. We can do the cleanup latter. Lets do any cleanup right now, just add the commit data functionality.

vivekv1504:
> Code cahnges look good for the commit data but I dont understand why have we refactored the code and removed so much code. We can do the cleanup latter. Lets do any cleanup right now, just add the commit data functionality.

why am remove that full version comparison related code  becuse that is not recommended this time 
i had discusion with rajesh regarding this full version comparison ,he said this not part of this ticket we can follow what we mentioned in the confluence page that only fine ,that's why i am removing that code .

docs/changelog/index.html
vivekv1504: If the clear buttons needs i will add

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1465,260 @@ const handleStableVersionChange = async () => {
...
+
     updateCompareButtonState();
 };
 
 /**
- * Switch to single view mode.
+ * Switch to single view mode
  */
 const switchToSingleViewMode = () => {
     comparisonMode = false;
```
Shreyas281299, line 1519: **[Medium] `comparisonMode` variable undeclared — refactoring oversight**

The `let comparisonMode = false;` declaration was accidentally removed along with the old comparison code block. The variable is still referenced in `switchToSingleViewMode` (line 1506) and `switchToComparisonViewMode` (line 1528), making it an implicit global variable. This would cause a `ReferenceError` in strict mode.

**Suggested fix:** Add `let comparisonMode = false;` near the top of the file with other state variables (around line 2-3).

docs/changelog/assets/js/app.js
```diff
@@ -1037,13 +637,22 @@ const switchToComparisonMode = (versionA = null, versionB = null) => {
  * @returns {Array<string>} - Array of all package names (union)
  */
 const getUnionPackages = (changelogA, changelogB) => {
```
Shreyas281299, line 640: **[Low] Removed null safety in `getUnionPackages`**

The old code had `Object.keys(changelogA || {})` with null fallbacks. The new code uses `Object.keys(changelogA)` directly. While callers are wrapped in try/catch and this is unlikely to trigger, restoring the fallback is trivial and adds defense in depth.

**Suggested fix:** Restore `Object.keys(changelogA || {})` and `Object.keys(changelogB || {})`.

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1465,260 @@ const handleStableVersionChange = async () => {
...
- * Order: package first, then enable and set Base/Target versions, then pre-release and run comparison.
+ * Handle enhanced comparison URL parameters on page load
  */
 const loadEnhancedComparisonFromURL = async (enhancedParams) => {
-    switchToComparisonMode();
-    await populateComparisonPackagesInitial();
+    switchToComparisonViewMode();
 
-    await new Promise(resolve => setTimeout(resolve, 100));
+    await new Promise(resolve => setTimeout(resolve, 300));
```
Shreyas281299: **[Low] Fragile `setTimeout` delays (pre-existing)**

`loadEnhancedComparisonFromURL` uses three hardcoded `setTimeout(resolve, 300)` delays. This pattern pre-dates this PR (previously 100ms), and the increase to 300ms actually improves reliability. However, the underlying approach is brittle on slow connections.

**Suggested fix:** Consider event-driven patterns in a future cleanup (e.g., MutationObserver or promise-based DOM readiness). Not blocking for this PR.

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1465,260 @@ const handleStableVersionChange = async () => {
...
     populateComparisonVersions();
-    await populateComparisonPackagesInitial();
-    updateCompareButtonState();
 };
 
 /**
- * Validate comparison form inputs. Package is required for package-level comparison; full version comparison only needs Base and Target.
+ * Validate comparison form inputs
  */
 const validateComparisonInputs = (stableA, stableB, selectedPackage, versionASpecific, versionBSpecific) => {
```
Shreyas281299, line 1561: **[Low] Indentation error in `validateComparisonInputs`**

Misaligned closing brace at column 0 instead of column 4 in the `if (allSorted[0] !== stableA)` block.

**Suggested fix:** Fix indentation to align the closing brace with the `if` statement.

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +901,339 @@ const calculateComparisonStats = (packages) => {
...
+ * Rules (from normal-text.txt):
+ *  'start'  → from versionA (inclusive) through ALL remaining pre-releases
+ *             Special: if versionA === stableVersion → only stable entry
+ *  'middle' → skip exact stable entry; ALL pre-releases of this stable
+ *  'end'    → ALL pre-releases from next.1 up to versionB (inclusive)
+ *             Special: if versionB === stableVersion → only stable entry
+ *  'only'   → stableA === stableB; from versionA to versionB within same file
  */
-const generatePackageComparisonData = (packageName, versionASpecific, versionBSpecific, changelogA, changelogB) => {
+const collectCommitsFromStable = (packageData, stableVersion, versionA, versionB, position) => {
```
Shreyas281299, line 1005: **[Low] Cross-tag pre-release inclusion — intentional but under-documented**

In `collectCommitsFromStable`, the 'start' position logic includes ALL pre-releases from different tags (e.g., `multipleLLM` when comparing `next` versions). The inline comment says "Different tag: include all" but doesn't explain *why*.

**Suggested fix:** Expand the inline comment to explain the rationale: commits from alternate pre-release streams also ship in the final stable release.

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +901,339 @@ const calculateComparisonStats = (packages) => {
...
+    return new RegExp(`^${escaped}-`).test(version);
+};
+
+// Is this an exact stable version (no pre-release suffix)?
+// e.g. isExactStable("3.6.0")        → true
+//      isExactStable("3.6.0-next.1") → false
+const isExactStable = (version) => /^\d+\.\d+\.\d+$/.test(version);//
+
+// Extract numeric suffix: "3.5.0-next.5" → 5,  "3.5.0-multipleLLM.3" → 3
+const getPreReleaseNum = (version) => {
```
Shreyas281299, line 982: **[Low] `getPreReleaseNum` regex matches stable versions**

`/\.(\d+)$/` would match `3.6.0` and return `0`. Currently safe because callers filter with `isPreRelease` first, but could cause subtle bugs if called in a different context.

**Suggested fix:** Consider making the regex more specific: `/\-[a-zA-Z]+\.(\d+)$/`.

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1465,260 @@ const handleStableVersionChange = async () => {
...
+    comparisonPackageSelect.value = enhancedParams.packageName;
     handlePackageChange();
-    setPackageSelection(enhancedParams.packageName);
-    if (versionAPrereleaseSelect) versionAPrereleaseSelect.value = enhancedParams.versionA;
-    if (versionBPrereleaseSelect) versionBPrereleaseSelect.value = enhancedParams.versionB;
 
-    await new Promise(resolve => setTimeout(resolve, 100));
+    await new Promise(resolve => setTimeout(resolve, 300));
+
+    versionAPrereleaseSelect.value = enhancedParams.versionA;
```
Shreyas281299: **[Low] Missing null guard on prerelease selects in `loadEnhancedComparisonFromURL`**

The old code used `if (versionAPrereleaseSelect) versionAPrereleaseSelect.value = ...` but the new code directly accesses `.value` without the null check. If the DOM element doesn't exist, this throws.

**Suggested fix:** Restore the null guards.

docs/changelog/assets/js/app.js
vivekv1504: add the state variable `let comparisonMode=false;`

docs/changelog/assets/js/app.js
vivekv1504: added null fallbacks for avoiding errors

docs/changelog/assets/js/app.js
vivekv1504: current code is good For now, the 300ms delays are fine. No action needed.

docs/changelog/assets/js/app.js
vivekv1504: align the indendation for if block

docs/changelog/assets/js/app.js
vivekv1504: add the detailed comments for previous inline comments

docs/changelog/assets/js/app.js
vivekv1504: `/-[a-zA-Z]+\.(\d+)$/` update thye regex becuase it taking wrons in sometime ,how they are if pre-release version as stable version selects (3.6.0--> it takes 0)as for previous regex ,actually that is wrong .

docs/changelog/assets/js/app.js
vivekv1504: restoring the null guard checks like  a ClearComparisonForm checks guards

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+        } else {
+            const tagA = getPreReleaseTag(versionA, stableVersion);
+            const numA = getPreReleaseNum(versionA);
+            versionsToUse = all.filter(v => {
+                if (!isPreRelease(v, stableVersion)) return false;
+                const tag = getPreReleaseTag(v, stableVersion);
+                const num = getPreReleaseNum(v);
+                // Same tag (e.g. "next"): include if num >= numA
+                // Different tag (e.g. "multipleLLM"): include all — alternate pre-release streams also ship in the final stable
+                console.log('tag', tag, 'tagA', tagA);
```
bhabalan: Looks like some debug logs got left in here. These fire on every iteration of the filter so they'll spam the console pretty heavily during any comparison. There are a few more at L1051-1052 and L1223 as well — would clean all of these up before merging.

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+ * Returns a flat, deduplicated array of commit objects.
+ */
+const collectCommitsAcrossStables = async (stableA, stableB, packageName, versionA, versionB, changelogA, changelogB) => {
+    const stables = getStableVersionsBetween(stableA, stableB);
+    if (stables.length === 0) return { commitsBetween: [], stableVersionsTraversed: [] };
+
+    const all = new Map();
+    const traversed = [];
+
+    for (let i = 0; i < stables.length; i++) {
```
bhabalan, line 1116: These intermediate changelogs are fetched one at a time in the loop — if someone compares 3.6.0 to 3.15.0 that's roughly 10 sequential network requests each waiting on the previous one. Could we batch these with `Promise.all` instead? Something like:

```js
const intermediateStables = stables.filter(s => s !== stableA && s !== stableB);
const fetched = await Promise.all(
    intermediateStables.map(async (stable) => {
        try {
            const res = await fetch(versionPaths[stable]);
            return [stable, res.ok ? await res.json() : null];
        } catch { return [stable, null]; }
    })
);
const changelogMap = Object.fromEntries(fetched);
```

Then just look up `changelogMap[stable]` in the loop instead of awaiting inside it.

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1472,260 @@ const handleStableVersionChange = async () => {
...
             selectedPackage,
             finalVersionA,
             finalVersionB,
             comparisonState.cachedChangelogA,
             comparisonState.cachedChangelogB
         );
-    } else {
-        // Full version comparison
-        performVersionComparison(stableA, stableB);
     }
```
bhabalan, line 1638: The old code had an `else` branch here that called `performVersionComparison(stableA, stableB)` — that was the path for comparing all packages between two versions without picking a specific one. With that removed, and `updateCompareButtonState` (L1422) now disabling the button unless a package is selected, the full-version comparison flow is gone entirely.

Was this intentional? If so might be worth calling it out in the PR description since it's a user-facing behavior change.

docs/changelog/assets/js/app.js
```diff
@@ -1668,141 +1344,125 @@ const comparisonState = {
...
+    if (stableA && stableB) {
         try {
-            const [changelogA, changelogB] = await Promise.all([
-                fetch(versionPaths[stableA]).then(res => res.json()),
-                fetch(versionPaths[stableB]).then(res => res.json())
-            ]);
+            let changelogA, changelogB;
+            if (stableA === stableB) {
+                // Same stable — fetch once and reuse for both sides
+                changelogA = await fetch(versionPaths[stableA]).then(res => res.json());
```
bhabalan: This doesn't check `res.ok` before calling `res.json()`. If the server returns a 404 or 500 you'll get a confusing `SyntaxError` from trying to parse HTML as JSON instead of a useful error message. Same thing on L1459-1460.

The fetch in `collectCommitsAcrossStables` (L1125) already does this check — same pattern would work here:

```js
const res = await fetch(versionPaths[stableA]);
if (!res.ok) throw new Error(`Failed to load changelog for ${stableA} (${res.status})`);
changelogA = await res.json();
```

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+    if (iA === -1 || iB === -1) return [];
+    return all.slice(Math.min(iA, iB), Math.max(iA, iB) + 1);
+};
+
+// --- Regex helpers for pre-release version identification ---
+
+// Is this version a pre-release of the given stable?
+// e.g. isPreRelease("3.5.0-next.1", "3.5.0") → true
+//      isPreRelease("3.5.0",         "3.5.0") → false
+const isPreRelease = (version, stableVersion) => {
```
bhabalan: This creates a new `RegExp` on every call, and it gets called quite a bit inside the filter callbacks. Since we're really just checking if the version starts with the stable prefix + `'-'`, `startsWith` would do the same thing without the regex overhead:

```js
const isPreRelease = (version, stableVersion) =>
    version.startsWith(stableVersion + '-');
```

docs/changelog/assets/js/app.js
```diff
@@ -1812,364 +1472,260 @@ const handleStableVersionChange = async () => {
...
- * Order: package first, then enable and set Base/Target versions, then pre-release and run comparison.
+ * Handle enhanced comparison URL parameters on page load
  */
 const loadEnhancedComparisonFromURL = async (enhancedParams) => {
-    switchToComparisonMode();
-    await populateComparisonPackagesInitial();
+    switchToComparisonViewMode();
 
-    await new Promise(resolve => setTimeout(resolve, 100));
+    await new Promise(resolve => setTimeout(resolve, 300));
```
bhabalan: Three hardcoded 300ms waits in a row feels a bit fragile — if the network or DOM is slower than that, the downstream code could end up operating on stale state (e.g., setting a dropdown value before its options have been populated, which would silently fail). Could we at least verify the expected state is ready before proceeding? Something like checking that `comparisonState.cachedChangelogA` is populated before trying to set the package selection.

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+        const p = v => v.split('.').map(Number);
+        const [aMaj, aMin, aPatch] = p(a);
+        const [bMaj, bMin, bPatch] = p(b);
+        return aMaj !== bMaj ? aMaj - bMaj : aMin !== bMin ? aMin - bMin : aPatch - bPatch;
+    });
+
+// Get all stable versions (from versionPaths) that sit between stableA and stableB (inclusive)
+const getStableVersionsBetween = (stableA, stableB) => {
+    const all = sortStableVersions(Object.keys(versionPaths));
+    const iA = all.indexOf(stableA), iB = all.indexOf(stableB);
```
bhabalan, line 963: This handles either argument order with `Math.min/max`, which is nice, but it can cause issues downstream. `collectCommitsAcrossStables` assigns `'start'`/`'end'` positions by checking `stable === stableA` vs `stable === stableB`. If stableA is actually the *later* version, those positions get swapped and the commit collection logic inverts.

`validateComparisonInputs` catches this for form submissions, but `loadEnhancedComparisonFromURL` bypasses validation entirely — so a hand-crafted URL with the versions flipped would hit this. Worth either validating in `loadEnhancedComparisonFromURL` or normalizing the order at the top of `collectCommitsAcrossStables`.

docs/changelog/index.html
```diff
@@ -128,20 +125,20 @@
...
+                  <button type="button" id="copy-comparison-link" class="btn btn-info hide" title="Copy shareable link to clipboard" >
+                     📋 Copy Permalink
                   </button>
                 </div>
               </div>
               <!-- Comparison Helper Text -->
               <div class="comparison-helper hide" id="comparison-helper">
                 <p class="comparison-helper-tip">
-                  <strong>💡 Tip:</strong> After comparing, you can share the URL from your browser's address bar, or use the "Copy Permalink & share link" button above. 
+                  <strong>💡 Tip:</strong> After comparing, you can share the URL from your browser's address bar, or use the "Copy Permalink & share link" button above.
```
bhabalan: The button on L134 says "Copy Permalink" but this tip text still says "Copy Permalink & share link" — should probably match.

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+            versionsToUse = [stableVersion];
+        } else {
+            const tagB = getPreReleaseTag(versionB, stableVersion);
+            const numB = getPreReleaseNum(versionB);
+            versionsToUse = all.filter(v => {
+                if (!isPreRelease(v, stableVersion)) return false;
+                const tag = getPreReleaseTag(v, stableVersion);
+                const num = getPreReleaseNum(v);
+                // Same tag: include if num <= numB
+                // Different tag: include all-— alternate pre-release streams also ship in the final stable
```
bhabalan: nit: extra dash before the em dash — `all-—` should be `all —`

docs/changelog/assets/js/app.js
```diff
@@ -1405,166 +902,345 @@ const calculateComparisonStats = (packages) => {
...
+//      isPreRelease("3.5.0",         "3.5.0") → false
+const isPreRelease = (version, stableVersion) => {
+    const escaped = stableVersion.replace(/\./g, '\\.');
+    return new RegExp(`^${escaped}-`).test(version);
+};
+
+// Is this an exact stable version (no pre-release suffix)?
+// e.g. isExactStable("3.6.0")        → true
+//      isExactStable("3.6.0-next.1") → false
+const isExactStable = (version) => /^\d+\.\d+\.\d+$/.test(version);//
```
bhabalan: nit: stray `//` at the end of this line

docs/changelog/index.html
```diff
@@ -290,12 +287,51 @@ <h3>Version Comparison: {{versionA}} vs {{versionB}}</h3>
...
+                <th style="width:100px;">Hash</th>
+                <th>Commit Message</th>
+                <th style="width:155px;">Version</th>
+              </tr>
+            </thead>
+            <tbody>
+              {{#each commitsBetween}}
+              <tr>
+                <td><code><a href="{{this.url}}" target="_blank">{{this.shortHash}}</a></code></td>
+                <td>{{{github_linking this.message 'message'}}}</td>
```
bhabalan, line 320: This uses triple-brace Handlebars (unescaped output). I see the existing template on L248 does the same thing so it's not new, but worth double-checking that `github_linking` sanitizes its input — a commit message with something like `` would get rendered as-is otherwise.

Shreyas281299:
Screenshot 2026-04-01 at 9 34 53 AM

When I do this comparision the commits or the packages changed both are incorrect. Please fix this, we dont show any commits for 3.5.0-next.X, 3.6.0-next.X, 3.7.0-next.X,

docs/changelog/assets/js/app.js
vivekv1504: i am using promise.all fetching at a time files becuase previous it takes 3 to 4 sec take for fetching all files one by one , now its takes 300ms for fetching files

docs/changelog/assets/js/app.js
vivekv1504: "Yes, this is intentional. The full-version comparison (comparing all packages between two stables without selecting a specific package) has been removed. It was rarely useful and could be misleading — the package-level comparison with commit history is the primary workflow now. I 'll call this out in the PR description."

docs/changelog/assets/js/app.js
vivekv1504: removed console logs

docs/changelog/assets/js/app.js
vivekv1504: add the checks for data fetching are not (Add a res.ok check before parsing JSON. ),Now if the server returns a 404 or 500, the user sees "Error loading version data. Failed to fetch 3.8.0 (404)" instead of a cryptic SyntaxError.

docs/changelog/assets/js/app.js
vivekv1504: clear the regex because it creates new regExp on every call ,added this code instead of regex
`version.startsWith(stableVersion + '-');`

docs/changelog/assets/js/app.js
vivekv1504: Replaced the three setTimeout waits with proper async/await — handleStableVersionChange is already async, so awaiting it ensures changelogs are fetched before we proceed. Also added an early return if the changelogs failed to load. handlePackageChange and switchToComparisonViewMode are synchronous, so no wait is needed for those.

docs/changelog/assets/js/app.js
vivekv1504: "The UI already validates version order before comparison. loadEnhancedComparisonFromURL only runs from permalinks generated by our own code, which always produces correctly ordered parameters. I'll add a comment noting this assumption."

docs/changelog/index.html
vivekv1504: updated  that tip message matched with copy permalink button  name

docs/changelog/assets/js/app.js
vivekv1504: Fixed the typo — removed the extra dash.

docs/changelog/assets/js/app.js
vivekv1504: Removed the stray comment marker.

docs/changelog/index.html
vivekv1504: Good catch — added HTML escaping before link injection in github_linking for both the hash and message cases.

vivekv1504:
> Screenshot 2026-04-01 at 9 34 53 AM
> When I do this comparision the commits or the packages changed both are incorrect. Please fix this, we dont show any commits for 3.5.0-next.X, 3.6.0-next.X, 3.7.0-next.X,

Screenshot 2026-04-01 at 23 54 51
now ,we can compare with both stable version instead of pre-release versions ,and gets  in between version with commits collected

Shreyas281299 (commented):
Looks good to me.

docs/changelog/assets/js/app.js
vivekv1504: this functionaliy used in full version comparison code  ,now that full version comoparison code deleted no need this time .
Reference message 
https://github.com/webex/webex-js-sdk/pull/4768#issuecomment-4109145847

docs/changelog/assets/js/app.js
vivekv1504: this functionaliy used in full version comparison code ,now that full version comoparison code deleted no need this time .
Reference message
https://github.com/webex/webex-js-sdk/pull/4768#issuecomment-4109145847

docs/changelog/assets/js/app.js
vivekv1504: this functionaliy used in full version comparison code ,now that full version comoparison code deleted no need this time .
Reference message
https://github.com/webex/webex-js-sdk/pull/4768#issuecomment-4109145847

docs/changelog/index.html
vivekv1504: i removed these lines and add the after selecting stable version  dropdowns

Shreyas281299 (approved):

bhabalan (approved):

PR #4769, opened 2026-03-12
feat: use new set custom codec api

edvujic:
# COMPLETES [SPARK-784314](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-784314)

The `SendSlotManager.setCodecParameters` and `SendSlotManager.deleteCodecParameters` methods use a legacy WCME API that does not require specifying a codec MIME type. WCME has since introduced `setCustomCodecParameters` and `markCustomCodecParametersForDeletion`, which scope parameter changes to a specific codec (e.g. OPUS, H264, AV1). This ensures SDP/codec handling stays consistent with WCME.

- **Deprecated** `setCodecParameters` and `deleteCodecParameters` in `SendSlotManager` with `@deprecated` JSDoc annotations.
- **Added** `setCustomCodecParameters(mediaType, codecMimeType, codecParameters)` to `SendSlotManager`, which delegates to WCME's `SendSlot.setCustomCodecParameters`.
- **Added** `markCustomCodecParametersForDeletion(mediaType, codecMimeType, parameters)` to `SendSlotManager`, which delegates to WCME's `SendSlot.markCustomCodecParametersForDeletion`.
- **Migrated** the `enableMusicMode` caller in `meeting/index.ts` from the deprecated API to the new API, using `MediaCodecMimeType.OPUS` for the OPUS codec parameters (`maxaveragebitrate`, `maxplaybackrate`).
- **Updated** unit tests in `sendSlotManager.ts` to cover the new methods.
- **Updated** unit tests in `meeting/index.js` to match the migrated `enableMusicMode` implementation.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

- Unit tests for `setCustomCodecParameters`: verifies parameters are forwarded to the WCME `SendSlot` with correct `codecMimeType` and `codecParameters`
- Unit tests for `markCustomCodecParametersForDeletion`: verifies parameters are forwarded to the WCME `SendSlot` with correct `codecMimeType` and parameter keys
- Unit tests for error cases: both new methods throw when the slot does not exist
- Unit tests for deprecated `setCodecParameters`: verifies it still works but emits a deprecation warning
- Unit tests for deprecated `deleteCodecParameters`: verifies it still works but emits a deprecation warning
- Unit tests for `enableMusicMode(true)`: verifies `setCustomCodecParameters` is called with `MediaCodecMimeType.OPUS` and the correct parameters
- Unit tests for `enableMusicMode(false)`: verifies `markCustomCodecParametersForDeletion` is called with `MediaCodecMimeType.OPUS` and the correct parameter keys

  - [ ] Github Copilot
  - [x] Other - Cursor
- [x] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [x] Tech Debt
  - [ ] Automation

### I certified that

---

### Files Modified

| File | Change |
|------|--------|
| `packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts` | Deprecated old methods, added `setCustomCodecParameters` and `markCustomCodecParametersForDeletion` |
| `packages/@webex/plugin-meetings/src/meeting/index.ts` | Migrated `enableMusicMode` to use new API with `MediaCodecMimeType.OPUS` |
| `packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts` | Added tests for new methods, updated deprecated method tests to assert warnings |
| `packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js` | Updated `enableMusicMode` tests to match new API |


packages/@webex/media-helpers/package.json
```diff
@@ -22,7 +22,7 @@
     "deploy:npm": "yarn npm publish"
   },
   "dependencies": {
-    "@webex/internal-media-core": "2.22.1",
+    "@webex/internal-media-core": "link:../../../../webrtc-media-core",
```
edvujic: Temporal local linking for testing purposes.

packages/calling/package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@types/platform": "1.3.4",
-    "@webex/internal-media-core": "2.22.1",
+    "@webex/internal-media-core": "link:../../../../webrtc-media-core",
```
edvujic: Temporal local linking for testing purposes.

packages/@webex/plugin-meetings/package.json
```diff
@@ -62,7 +62,7 @@
   },
   "dependencies": {
     "@webex/common": "workspace:*",
-    "@webex/internal-media-core": "2.22.1",
+    "@webex/internal-media-core": "link:../../../../webrtc-media-core",
```
edvujic: Temporal local linking for testing purposes.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
fnowakow, line 233: Maybe we should use new API here instead of old one. We did the same in WCME so old `setCodecParameters` is just for backward compability but inside it calls new API

fnowakow, line 260: Same as in https://github.com/webex/webex-js-sdk/pull/4769/changes#r2930416325

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
fnowakow: These expressions evaluate the sinon boolean but never chain `.to.be.true`, so they silently pass
regardless of the outcome. The new tests correctly use `.to.be.true`. This is a pre-existing issue
but worth fixing while touching the file.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -9220,27 +9220,29 @@ describe('plugin-meetings', () => {
...
         );
       });
 
-      it('should set the codec parameters when shouldEnableMusicMode is false', async () => {
+      it('should mark custom codec parameters for deletion when shouldEnableMusicMode is false', async () => {
         await meeting.enableMusicMode(false);
         assert.calledOnceWithExactly(
-          meeting.sendSlotManager.getSlot(MediaType.AudioMain).deleteCodecParameters,
+          meeting.sendSlotManager.getSlot(MediaType.AudioMain).markCustomCodecParametersForDeletion,
+          'audio/opus',
```
fnowakow: Can we use here `MediaCodecMimeType.OPUS`?

fnowakow (commented):
I think PR description is out-of-date with code

> Deprecated `setCodecParameters` and `deleteCodecParameters` … with runtime deprecation
> warnings via `LoggerProxy.logger.warn`.

However, no `logger.warn` call was added

> Updated unit tests in `sendSlotManager.ts` … to assert deprecation warnings on the old methods.

No such test modifications appear in the diff.

fnowakow (approved):

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -224,14 +228,21 @@ export default class SendSlotManager {
       throw new Error(`Slot for ${mediaType} does not exist`);
     }
 
-    await slot.setCodecParameters(codecParameters);
+    const codecMimeType =
+      mediaType === MediaType.AudioMain || mediaType === MediaType.AudioSlides
+        ? MediaCodecMimeType.OPUS
+        : MediaCodecMimeType.H264;
```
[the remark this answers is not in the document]
edvujic: valid concern, done.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -217,40 +223,121 @@ export default class SendSlotManager {
...
     }
   ): Promise<void> {
-    // These codec parameter changes underneath are SDP value changes that are taken care by WCME automatically. So no need for any change in streams from the web sdk side
+    this.LoggerProxy.logger.warn(
+      'SendSlotsManager->setCodecParameters --> [DEPRECATION WARNING]: setCodecParameters has been deprecated, use setCustomCodecParameters instead'
+    );
+
+    Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.DEPRECATED_SET_CODEC_PARAMETERS_USED, {
+      mediaType,
+    });
```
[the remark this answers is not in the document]
edvujic: The metric calls are already positioned after the slot work, so the core functionality completes regardless. `SendSlotManager` is only used within active meetings where `Metrics.initialSetup()` has already been called so send slots are created during `addMedia()`, which is behind `Meetings.onReady()`. 

This is consistent with all other `sendBehavioralMetric` usages in the codebase (86+ calls, none with initialization guards). Adding a guard here would introduce an inconsistent pattern.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -244,10 +259,89 @@ export default class SendSlotManager {
...
-      `SendSlotsManager->deleteCodecParameters#Deleted the following codec parameters -> ${parameters} for ${mediaType}`
+      `SendSlotsManager->setCustomCodecParameters#Set custom codec parameters for ${mediaType} (codec: ${codecMimeType}) to ${JSON.stringify(
+        parameters
+      )}`
+    );
+
+    Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.SET_CUSTOM_CODEC_PARAMETERS_USED, {
+      mediaType,
+      codecMimeType,
+      parameters,
```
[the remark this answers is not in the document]
edvujic: Both the deprecated and new methods have the metric calls positioned after the slot work, so the core codec functionality always completes first. See https://github.com/webex/webex-js-sdk/pull/4769#discussion_r2965105201

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -206,32 +210,38 @@ export default class SendSlotManager {
...
-   * @param {Object} codecParameters
+   * @param {CodecParameters} parameters Key-value pairs of codec parameters to set
    * @returns {Promise<void>}
    */
   public async setCodecParameters(
     mediaType: MediaType,
-    codecParameters: {
-      [key: string]: string | undefined; // As per ts-sdp undefined is considered as a valid value to be used for codec parameters
-    }
+    parameters: CodecParameters
```
k-wasniowski: I think we shouldnt change signature of this method

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -189,28 +197,33 @@ describe('SendSlotsManager', () => {
     describe('setCodecParameters', () => {
         let mediaConnection;
         const mediaType = MediaType.AudioMain;
-        const codecParameters = {};
```
k-wasniowski, line 192: Any particular reason for renaming here?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -246,8 +256,82 @@ export default class SendSlotManager {
...
+    await slot.setCustomCodecParameters(codecMimeType, parameters);
+
+    this.LoggerProxy.logger.info(
+      `SendSlotsManager->setCustomCodecParameters#Set custom codec parameters for ${mediaType} (codec: ${codecMimeType}) to ${JSON.stringify(
+        parameters
+      )}`
+    );
+
+    Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.SET_CUSTOM_CODEC_PARAMETERS_USED, {
+      mediaType,
```
antsukanova: Why do we need a metric for the new methods?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -246,8 +256,82 @@ export default class SendSlotManager {
...
+    codecMimeType: MediaCodecMimeType,
+    parameters: CodecParameters
+  ): Promise<void> {
+    const slot = this.slots.get(mediaType);
+
+    if (!slot) {
+      throw new Error(`Slot for ${mediaType} does not exist`);
+    }
+
+    await slot.setCustomCodecParameters(codecMimeType, parameters);
```
antsukanova: why there is no try catch for that?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -246,8 +256,82 @@ export default class SendSlotManager {
...
+    Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.DEPRECATED_DELETE_CODEC_PARAMETERS_USED, {
+      mediaType,
+      parameters,
+    });
+  }
+
+  /**
+   * Sets custom codec parameters for the sendSlot of the given mediaType, scoped to a specific codec MIME type.
+   * Delegates to WCME's setCustomCodecParameters API.
+   * @param {MediaType} mediaType MediaType of the sendSlot (AUDIO_MAIN/VIDEO_MAIN/AUDIO_SLIDES/VIDEO_SLIDES)
```
antsukanova: TS type `MediaType` already constrains what values are accepted, and any ide will show those values via autocomplete and hover, so I would recommend removing that here.

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -29,13 +37,13 @@ describe('SendSlotsManager', () => {
...
 
-            expect(mediaConnection.createSendSlot.calledWith(mediaType, true));
+            expect(mediaConnection.createSendSlot.calledWith(mediaType, true)).to.be.true;
         });
 
         it('should create a slot for the given mediaType & active state', () => {
             sendSlotsManager.createSlot(mediaConnection, mediaType, false);
 
-            expect(mediaConnection.createSendSlot.calledWith(mediaType, false));
+            expect(mediaConnection.createSendSlot.calledWith(mediaType, false)).to.be.true;
```
antsukanova: I think such output could be misleading `AssertionError: expected false to be true`
could we try to impove this?  Like for example I see an option to do something like that: `expect(mediaConnection.createSendSlot).to.have.been.calledWithExactly(mediaType, false);`

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -227,20 +240,99 @@ describe('SendSlotsManager', () => {
...
+            expect((Metrics.sendBehavioralMetric as sinon.SinonStub).calledWith(
+                BEHAVIORAL_METRICS.DEPRECATED_DELETE_CODEC_PARAMETERS_USED,
+                { mediaType, parameters: [] }
+            )).to.be.true;
+        });
+
+        it('should throw an error if a slot for the given mediaType does not exist', (done) => {
+            sendSlotsManager.deleteCodecParameters(mediaType, []).catch((error) => {
+                expect(error.message).to.equal(`Slot for ${mediaType} does not exist`);
+                done();
```
antsukanova: why do you need this code here with `done()`?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
edvujic: Good point!

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
edvujic: This is consistent with all other async methods in SendSlotManager. `publishStream`, `unpublishStream`, `setCodecParameters`, and `deleteCodecParameters` all let exceptions propagate to the caller without try-catch. Why do you think this should be added here?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
edvujic: The idea is: 

- We add metrics for both `setCodecParameters` and `setCustomCodecParametars`
- Observe in Amplitude the usage of both methods
- Once the usage of `setCodecParameters` becomes insignificant, fully remove the deprecated method.

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
edvujic: No :)

edvujic: Reverting...

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
edvujic: sinon-chai is not installed in this project, so the `expect(stub).to.have.been.calledWithExactly(...)` syntax isn't available. But we have the pattern used in other parts of the code is `assert.calledWith(stub, ...args)` which gives a much better error message showing actual vs expected args. Will change to that.

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
edvujic: The `done` callback is used here because the test relies on .catch() to assert the rejected promise. Without `done()`, Mocha would consider the test passed immediately (before the `.catch()` runs), meaning a missing rejection would silently pass. Which is not correct.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
edvujic: Done.

k-wasniowski (approved):

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
antsukanova: Got you, but then why you don't want to try use `async/await` + `rejectedWith` here?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: I see, then it seems only metric about `setCodecParameters` should be enough?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: I do not think your metric code will work if you receive a rejection here for `setCustomCodecParameters`. Could you check if it will work with this flow?

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
k-wasniowski: The idea behind having a metric for the new method was not to track if users have already migrated to the new method, but rather to be able to track what kind of modifications they are doing. That's why we are populating there mimeType and params.

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
edvujic: The metric here is intended to track successful usage of the new API (it's named `SET_CUSTOM_CODEC_PARAMETERS_USED`, not `_ATTEMPTED`). If we want to also track failures, we could add a try-catch that sends a separate failure metric, but I think tracking successful usage is sufficient for the purpose of this metric. The same pattern applies to the deprecated methods, their metrics also only fire on success.

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
edvujic: Good point!

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: @edvujic sorry, I didn't het your answer here, could you confirm that this flow without try catch will work if `setCustomCodecParameters` failed and metric will be sent?

antsukanova: If we are talking about `SET_CUSTOM_CODEC_PARAMETERS_USED` logic as metric I believe we need make sure that with failue `SET_CUSTOM_CODEC_PARAMETERS_USED` will be fired, this metric is about __used__, so you are using this method here and it's failing

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
antsukanova: I see, good, thanks for info)

packages/@webex/plugin-meetings/test/unit/spec/multistream/sendSlotManager.ts
```diff
@@ -290,6 +291,20 @@ describe('SendSlotsManager', () => {
             await expect(sendSlotsManager.setCustomCodecParameters(mediaType, codecMimeType, parameters))
                 .to.be.rejectedWith(`Slot for ${mediaType} does not exist`);
         });
+
+        it('should throw and log error when setCustomCodecParameters fails', async () => {
```
antsukanova, line 295: this test also should  called with check for `assert.calledWith(Metrics.sendBehavioralMetric as sinon.SinonStub,
                BEHAVIORAL_METRICS.SET_CUSTOM_CODEC_PARAMETERS_USED,
                { mediaType, codecMimeType, parameters }
            );`

antsukanova (approved):
lgtm, but would be good to investigate and improve https://github.com/webex/webex-js-sdk/pull/4769#discussion_r3020572689

packages/@webex/plugin-meetings/src/multistream/sendSlotManager.ts
```diff
@@ -246,8 +259,89 @@ export default class SendSlotManager {
...
+        )}`
+      );
+    } catch (error) {
+      this.LoggerProxy.logger.error(
+        `SendSlotsManager->setCustomCodecParameters#Failed to set custom codec parameters for ${mediaType} (codec: ${codecMimeType}): ${error}`
+      );
+      throw error;
+    } finally {
+      Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.SET_CUSTOM_CODEC_PARAMETERS_USED, {
+        mediaType,
```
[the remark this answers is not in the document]
edvujic: Valid point on the error-masking scenario, if both the slot operation and the metric call fail, the finally block would replace the original codec error with `TypeError` which would be confusing. 

That said, the risk is low: `sendBehavioralMetric` is used 70+ times across the codebase without `try`/`catch` wrapping, and `Metrics.initialSetup(webex)` runs during SDK bootstrap well before any codec operations are possible.

Still, wrapping the `finally` metric call in a `try`/`catch` is not a pattern in this codebase.

Shreyas281299 (approved):

PR #4790, opened 2026-03-18
test(calling): add Playwright E2E base setup and SDK initialization tests

eigengravy:
# COMPLETES https://jira-eng-gpk2.cisco.com/jira/browse/CAI-7736

Adding Playwright E2E test infrastructure for the Calling SDK kitchen sink sample app, starting with SDK initialization test cases (INIT-001 through INIT-003).

- Add Playwright config (`playwright.config.ts`) with Chrome WebRTC flags, web server setup, and project configuration
- Add OAuth global setup (`playwright/global.setup.ts`) that fetches personal access tokens from developer.webex.com with parallel token fetching for caller, callee, and transfer users
- Add shared test utilities (`playwright/Utils/callingUtils.ts`) for navigation, SDK init, and service indicator/domain selection (with commented-out future utils for registration, calls, hold, etc.)
- Add constants (`playwright/constants.ts`) for element selectors and timeouts
- Add SDK initialization test cases (`playwright/tests/01-sdk-initialization.spec.ts`)
- Add `@playwright/test` devDependency and ESLint override for playwright files

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] INIT-001: Normal Calling — init with `calling` service indicator
- [x] INIT-002: Contact Center — init with `contactcenter` service indicator and service domain
- [x] INIT-003: Guest Calling — generate guest token and init with `guestcalling` service indicator
- [x] All 4 tests passing locally (OAuth + 3 init tests, ~1.5m)

Vidcast Link: https://app.vidcast.io/share/a69273ae-fa0d-4025-a67a-55bd06c16a23

  - [ ] Github Copilot
  - [x] Other - Claude Code
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [x] Automation

### I certified that


playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
+import {test as setup} from '@playwright/test';
```
rarajes2: Since this is for getting the tokens for different types - calling, guest, cc

We can rename the file to something like `tokenUtil.ts` or `OAuth.ts` or something else. Please use whatever feels best

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  verifySDKInitialized,
+  setServiceIndicator,
+  setServiceDomain,
+} from './callingUtils';
+import {SELECTORS, SDK_INIT_TIMEOUT, AWAIT_TIMEOUT} from './constants';
+
+const getCallerToken = (): string => {
+  const token = process.env.CALLER_ACCESS_TOKEN;
+  if (!token) {
+    throw new Error('CALLER_ACCESS_TOKEN not set. Run OAuth setup first.');
```
rarajes2: `global.setup.ts` can be named `OAuthSetup` or something similar --- matching this message.

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  return token;
+};
+
+test.describe('SDK Initialization', () => {
+  test('Normal Calling - init with calling service indicator', async ({page}) => {
```
rarajes2: Should we create 3 directories for each type of user - `normal calling`, `guest` and `cc` ? And keep all the tests related to them in the respective folder ?

Let's discuss this with Priya

playwright/calling/utils/constants.ts
```diff
@@ -0,0 +1,61 @@
+export const SAMPLE_APP_PATH = '/samples/calling/';
+
+// Timeouts
+export const AWAIT_TIMEOUT = 10000;
+export const SDK_INIT_TIMEOUT = 60000;
+export const REGISTRATION_TIMEOUT = 40000;
+export const OPERATION_TIMEOUT = 30000;
```
rarajes2: Let's ensure these numbers are as per the sdk values, can differ slightly.

playwright/tests/calling/callingUtils.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  // Fill in the access token
+  await page.locator(SELECTORS.ACCESS_TOKEN_INPUT).fill(accessToken, {timeout: AWAIT_TIMEOUT});
+
+  // Click "Initialize Calling" (submits the credentials form)
+  await page.locator(SELECTORS.INITIALIZE_CALLING_BTN).click({timeout: AWAIT_TIMEOUT});
+
+  // Wait for SDK to initialize - status changes to "Saved access token!" on ready
+  await expect(page.locator(SELECTORS.AUTH_STATUS)).toHaveText('Saved access token!', {
```
rarajes2: For utils, should we add the `expect` or the `expect` should be inside the actual test file ?

playwright/tests/calling/callingUtils.ts
```diff
@@ -0,0 +1,192 @@
...
+  const hasCallingClient = await page.evaluate(() => !!(window as any).callingClient);
+  expect(hasCallingClient).toBe(true);
+};
+
+/**
+ * Set service indicator before initialization (calling, contactcenter, guestcalling)
+ */
+export const setServiceIndicator = async (
+  page: Page,
+  service: 'calling' | 'contactcenter' | 'guestcalling'
```
rarajes2: We can create a type for service

playwright/tests/calling/callingUtils.ts
```diff
@@ -0,0 +1,192 @@
...
+};
+
+export const unregisterLine = async (page: Page): Promise<void> => {
+  await page.locator(SELECTORS.UNREGISTER_BTN).click({timeout: AWAIT_TIMEOUT});
+  await expect(page.locator(SELECTORS.REGISTRATION_STATUS)).toContainText('Un registering', {
+    timeout: REGISTRATION_TIMEOUT,
+  });
+};
+
+export const fullSetup = async (
```
rarajes2: `fullSetup` --> `initAndRegister` 

Or something more meaningful.

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+    return '';
+  }
+
+  return fs.readFileSync(ENV_PATH, 'utf8');
+};
+
+const escapeRegExp = (value: string): string => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+const upsertEnvVariables = (updates: EnvUpdateMap): void => {
+  let envContent = readEnvFile();
```
rarajes2: If we are only storing the token in the env, maybe instead of going through the env flow we can simply return the token value to caller. We can discuss this

playwright.config.ts
```diff
@@ -0,0 +1,66 @@
...
+      '--disable-site-isolation-trials',
+      '--disable-web-security',
+      '--no-sandbox',
+      '--disable-features=WebRtcHideLocalIpsWithMdns',
+      '--allow-file-access-from-files',
+      '--use-fake-ui-for-media-stream',
+      '--use-fake-device-for-media-stream',
+      '--disable-extensions',
+      '--disable-plugins',
+      '--ignore-certificate-errors',
```
rarajes2: We should add some inline comments explaining these flags

rarajes2 (commented):
As discussed the structure can be something like this

```
playwright
  - calling
    - tests
    - utils
  - foo
```

playwright/calling/utils/oauth.setup.ts
eigengravy: We store the tokens in `.env` so that on subsequent local runs you can set `SKIP_AUTH=true` and skip the OAuth flow entirely — dotenv picks them up before the tests start. This avoids re-authenticating every time while iterating locally. Pattern is copied from the widgets repo (`widgets/playwright/global.setup.ts`).

playwright/calling/utils/oauth.setup.ts
eigengravy: Done — renamed to `oauth.setup.ts` and moved to `playwright/calling/utils/`.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Done — renamed to `oauth.setup.ts`.

playwright/tests/calling/callingUtils.ts
eigengravy: Done — added `ServiceIndicator` type in `constants.ts`, now used in `setServiceIndicator` and `initAndRegister`.

playwright/tests/calling/callingUtils.ts
eigengravy: Done — renamed to `initAndRegister`.

playwright.config.ts
eigengravy: Done — added inline comments for each Chrome launch flag.

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (transferEmail && transferPassword) {
+    tokenFetches.push(
+      fetchAccessToken(browser, transferEmail, transferPassword).then((token) => {
+        tokenUpdates.TRANSFER_ACCESS_TOKEN = token;
+      })
+    );
+  }
+
+  await Promise.all(tokenFetches);
+  upsertEnvVariables(tokenUpdates);
```
akulakum: **P1 — Token handoff between OAuth setup project and test project is unreliable**

`dotenv.config()` in `playwright.config.ts` runs **once** at config-parse time before any projects execute. The dependent "Calling SDK E2E" project runs in a separate worker process, so `process.env` mutations made here (`process.env[key] = value`) do **not** propagate to the test worker.

On a clean run (no pre-existing `.env` tokens), the OAuth project will succeed and write to `.env`, but `process.env.CALLER_ACCESS_TOKEN` in `01-sdk-initialization.spec.ts` will be `undefined` in the test worker, causing all tests to fail with "CALLER_ACCESS_TOKEN not set."

**Recommendation:** Use Playwright's [storageState](https://playwright.dev/docs/auth#basic-shared-account-in-all-tests) or write a JSON file (e.g., `playwright/.auth/tokens.json`) that the test `beforeAll`/fixture reads at runtime, rather than relying on `.env` rewrite + `process.env` mutation across project boundaries.

Alternatively, if you want to keep `.env`, the test file needs to explicitly call `dotenv.config()` at the top before reading tokens, rather than relying on the config-time load.

playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const BASE_URL = process.env.PW_BASE_URL || 'https://localhost:8000';
+
+const chromeOptions = {
+  ...devices['Desktop Chrome'],
+  channel: 'chrome' as const,
+  launchOptions: {
+    args: [
+      '--disable-site-isolation-trials', // Allow cross-origin iframes in the same process
+      '--disable-web-security', // Bypass CORS for local dev server
```
akulakum: **P1 — `--disable-web-security` bypasses CORS entirely — masks real integration bugs**

Disabling web security in Chrome removes **all** CORS enforcement. If the SDK or sample app has a CORS misconfiguration, these E2E tests will never catch it. This is especially risky since the test hits real Webex endpoints during OAuth.

**Recommendation:** Remove `--disable-web-security`. If the local dev server needs CORS headers, configure the webpack dev server proxy instead. If it's only needed for the sample app's self-signed cert, `--ignore-certificate-errors` (already present on line 23) suffices.

package.json
```diff
@@ -131,6 +132,7 @@
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-module-resolver": "^5.0.2",
     "babelify": "^10.0.0",
+    "baseline-browser-mapping": "latest",
```
akulakum: **P2 — `baseline-browser-mapping: "latest"` is unpinned and appears unused**

Using `"latest"` violates dependency pinning best practices — it will resolve to a different version on every `yarn install`/`npm ci`, making builds non-deterministic and potentially introducing breaking changes silently.

Also, this dependency doesn't appear to be imported or used anywhere in this PR. Is it needed?

**Recommendation:** Either pin to a specific version (e.g., `"^2.10.8"`) or remove if unused.

package.json
```diff
@@ -107,6 +107,7 @@
     "@babel/types": "^7.14.9",
     "@commitlint/cli": "^17.1.2",
     "@commitlint/config-conventional": "^17.1.0",
+    "@playwright/test": "^1.58.2",
```
akulakum, line 111: **P2 — Consider tighter version pinning for test tooling**

Caret ranges (`^1.58.2`) can introduce drift when Playwright publishes minor versions with behavioral changes. For root-level test tooling, a tighter range reduces CI surprises.

**Recommendation:** Consider `"~1.58.2"` or exact pinning.

---

**P2 — No CI integration — tests will never run automatically**

There's no update to any CI workflow (`.github/workflows/`) or `package.json` script to run these Playwright tests. The new tests exist but are invisible to automation.

**Recommendation:** Add a `test:e2e` or `test:e2e:calling` script to `package.json` and wire it into the CI workflow (even if gated behind a manual trigger initially). E.g.:
```json
"test:e2e:calling": "npx playwright test --project='Calling SDK E2E'"
```

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  const tokenUpdates: EnvUpdateMap = {};
+
+  const callerEmail = process.env.CALLER_EMAIL;
+  const callerPassword = process.env.CALLER_PASSWORD;
+
+  if (!callerEmail || !callerPassword) {
+    throw new Error('CALLER_EMAIL and CALLER_PASSWORD must be set in .env');
+  }
+
```
akulakum: **P2 — `SKIP_AUTH` fast-path requires all three tokens but only caller is needed**

This guard requires `CALLEE_ACCESS_TOKEN` and `TRANSFER_ACCESS_TOKEN` even though the initialization test suite only consumes `CALLER_ACCESS_TOKEN`. A developer running just INIT tests with `SKIP_AUTH=true` and only a cached caller token will be forced through the full OAuth flow, which then throws at lines 125-129 if callee/transfer credentials aren't configured.

**Recommendation:** Make `SKIP_AUTH` validation conditional — only require the tokens that the current test suite needs. At minimum:
```typescript
if (
  process.env.SKIP_AUTH === 'true' &&
  process.env.CALLER_ACCESS_TOKEN
) {
  return;
}
```
Or check each token individually before adding its fetch to the queue.

playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+      testMatch: /oauth\.setup\.ts/,
+    },
+    {
+      name: 'Calling SDK E2E',
+      dependencies: ['Calling: OAuth Setup'],
+      testDir: './playwright/calling/tests',
+      use: chromeOptions,
+    },
+    {
+      name: 'Contact Center SDK E2E',
```
akulakum: **P2 — Contact Center project references a non-existent directory**

`./playwright/contact-center` does not exist in this PR. Playwright will either error or silently skip this project.

**Recommendation:** Remove this project block until actual Contact Center tests exist, or add a comment that it's a placeholder. An empty/missing `testDir` can cause confusing errors for other developers.

playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+      '--disable-web-security', // Bypass CORS for local dev server
+      '--no-sandbox', // Required for CI containers without root
+      '--disable-features=WebRtcHideLocalIpsWithMdns', // Expose real local IPs for WebRTC ICE candidates
+      '--allow-file-access-from-files', // Allow file:// protocol access
+      '--use-fake-ui-for-media-stream', // Auto-grant camera/mic permissions without prompt
+      '--use-fake-device-for-media-stream', // Use synthetic audio/video instead of real hardware
+      '--disable-extensions', // Prevent extensions from interfering with tests
+      '--disable-plugins', // Prevent plugins from interfering with tests
+      '--ignore-certificate-errors', // Accept self-signed certs from local dev server
+      '--auto-open-devtools-for-tabs', // Open DevTools for debugging test runs
```
akulakum: **P2 — `--auto-open-devtools-for-tabs` should not be in CI-ready config**

This flag opens Chrome DevTools for every test tab. In CI this is unnecessary overhead and may interfere with screenshots/traces/headless mode.

**Recommendation:** Gate behind an environment variable or remove:
```typescript
...(process.env.DEBUG_TESTS ? ['--auto-open-devtools-for-tabs'] : []),
```

packages/calling/playwright/utils/setup.ts
```diff
@@ -0,0 +1,95 @@
...
+  page: Page,
+  accessToken: string,
+  options: {registerLine?: boolean; service?: ServiceIndicator} = {}
+): Promise<void> => {
+  const {registerLine, verifyLineRegistered} = await import('./registration');
+
+  await navigateToCallingApp(page);
+  if (options.service) {
+    await setServiceIndicator(page, options.service);
+  }
```
akulakum, line 172: **P3 — Unnecessary dynamic `import()` — use a static import instead**

`registration.ts` is a sibling module in the same directory. Dynamic import adds complexity and defeats static analysis/tree-shaking.

**Recommendation:** Replace with a regular static import at the top of the file:
```typescript
import {registerLine, verifyLineRegistered} from './registration';
```

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  test('Contact Center - init with contactcenter service indicator', async ({page}) => {
+    await navigateToCallingApp(page);
+    await setServiceIndicator(page, 'contactcenter');
+    await setServiceDomain(page, 'rtw.prod-us1.rtmsprod.net');
+
+    const accessToken = getCallerToken();
+    await initializeCallingSDK(page, accessToken);
+    await verifySDKInitialized(page);
+  });
+
```
akulakum: **P3 — Guest Calling test duplicates logic from utility functions**

This test manually clicks "Initialize Calling", waits for auth status, and then calls `verifySDKInitialized` — but `verifySDKInitialized` already re-checks the same auth status text. Meanwhile `initializeCallingSDK` does the same fill+click+wait pattern.

Consider extracting a variant of `initializeCallingSDK` that skips the token fill step (since the guest token is already populated in the input), or at minimum remove the duplicate `AUTH_STATUS` check on lines 59-61 since `verifySDKInitialized` already covers it.

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (!fs.existsSync(ENV_PATH)) {
+    return '';
+  }
+
+  return fs.readFileSync(ENV_PATH, 'utf8');
+};
+
+const escapeRegExp = (value: string): string => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+const upsertEnvVariables = (updates: EnvUpdateMap): void => {
```
akulakum: **P3 — `.env` file manipulation via regex is fragile**

The `upsertEnvVariables` function does regex-based replacement on the `.env` file. This is brittle for edge cases (values containing `=`, special characters, multi-line values, access tokens with regex-significant characters).

Since this is test infrastructure, consider using a dedicated library like `dotenv-cli`, or writing a simple JSON file (`playwright/.auth/tokens.json`) that can be read back without parsing ambiguity. This would also help solve the P1 token handoff issue.

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+const getCallerToken = (): string => {
+  const token = process.env.CALLER_ACCESS_TOKEN;
+  if (!token) {
+    throw new Error('CALLER_ACCESS_TOKEN not set. Run OAuth setup first.');
+  }
+
+  return token;
+};
+
+test.describe('SDK Initialization', () => {
```
akulakum: **P3 — Consider adding `test.describe.configure({ mode: 'serial' })`**

The tests are implicitly serial because `workers: 1` and `fullyParallel: false` are set in the config. However, explicitly annotating serial mode within the test file documents the intent and protects against config changes:
```typescript
test.describe('SDK Initialization', () => {
  test.describe.configure({ mode: 'serial' });
  // ...
});
```

packages/calling/playwright/constants/selectors.ts
```diff
@@ -0,0 +1,63 @@
...
+  LOCAL_AUDIO: '#local-audio',
+  REMOTE_VIDEO: '#remote-video',
+  REMOTE_AUDIO: '#remote-audio',
+
+  // Guest Calling
+  GUEST_CONTAINER: '#guest-container',
+  JWT_TOKEN_FOR_DEST: '#jwt-token-for-dest',
+  GUEST_NAME: '#guest-name',
+  GENERATE_GUEST_TOKEN_BTN: 'button:has-text("Generate Guest Token")',
+
```
akulakum, line 46: **Nit — Text-based selector is fragile**

`button:has-text("Generate Guest Token")` will break if the button label changes. The sample app HTML doesn't have an `id` on this button. Adding an `id` attribute to the button in `index.html` (e.g., `id="generate-guest-token"`) would make this more resilient.

If changing the sample app is out of scope, at least note this fragility with a comment.

playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+  timeout: 120000,
+  webServer: {
+    command: 'yarn samples:serve',
+    url: BASE_URL,
+    ignoreHTTPSErrors: true,
+    reuseExistingServer: true,
+    stdout: 'ignore',
+    stderr: 'pipe',
+  },
+  retries: 0,
```
akulakum: **Nit — `retries: 0` is fine for local dev but may cause CI flakiness**

Consider making this configurable for CI environments:
```typescript
retries: process.env.CI ? 1 : 0,
```

akulakum (changes requested):
## PR Review

Good foundation for E2E testing infrastructure. The separation of concerns (utils, constants, setup, registration) is clean, the `ServiceIndicator` type is a nice pattern, and the Chrome flags are well-documented. However, there are a couple of critical issues that need to be addressed before merge.

### Summary

**Critical (P1):** 2 issues — token handoff across Playwright projects is broken on clean runs; `--disable-web-security` masks real CORS bugs.

**High (P2):** 6 issues — unpinned dependency, overly strict `SKIP_AUTH` guard, phantom Contact Center project, DevTools flag in CI config, no CI integration, caret range on Playwright dep.

**Medium (P3):** 5 issues — unnecessary dynamic import, duplicated logic in guest test, fragile `.env` manipulation, missing serial mode annotation, text-based selector fragility.

See inline comments for details and recommendations.

playwright/calling/utils/oauth.setup.ts
eigengravy: This is the same pattern used in the widgets repo (`widgets/playwright/global.setup.ts` lines 30-46, 157-177). The setup project writes tokens to `.env` via `upsertEnvVariables` and also sets `process.env[key]` directly — dependent projects pick them up from `process.env` in the same run. On subsequent runs, `dotenv.config()` loads them from the persisted `.env` file (used with `SKIP_AUTH=true` for local iteration).

playwright.config.ts
eigengravy: Same flag is used in the widgets repo (`widgets/playwright.config.ts` line 48). Required for the calling sample app which makes cross-origin requests to Webex APIs from a local dev server.

package.json
eigengravy: Done — removed. It was unused.

package.json
eigengravy: Keeping `^1.58.2` to stay consistent with the widgets repo which also uses caret ranges (`^1.51.1`).

playwright/calling/utils/oauth.setup.ts
eigengravy: More test suites (device registration, call flows) are being added on feature branches that require all 3 tokens (caller, callee, transfer). The guard is intentionally checking all 3.

playwright.config.ts
eigengravy: Done — commented out as a placeholder until Contact Center tests are added.

playwright.config.ts
eigengravy: Done — gated behind `\!process.env.CI` so DevTools only open in local runs.

packages/calling/playwright/utils/setup.ts
eigengravy: Done — switched to static import.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Done — removed the duplicate auth status check.

playwright/calling/utils/oauth.setup.ts
eigengravy: Same pattern used in the widgets repo (`widgets/playwright/global.setup.ts` lines 28-46). Works reliably for access tokens which are base64-encoded strings without special regex characters.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Added `test.describe.configure({mode: 'parallel'})` — the 3 init tests are independent (normal calling, contact center, guest) and now use separate credentials so they can run in parallel.

packages/calling/playwright/constants/selectors.ts
eigengravy: Done — added `id="generate-guest-token"` to the sample app button and updated the selector to `#generate-guest-token`.

playwright.config.ts
eigengravy: Done — set retries to 3.

playwright/tests/calling/callingUtils.ts
eigengravy: Keeping the utils as-is based on 1-1 discussion.

package.json
eigengravy: CI workflow will be added once all the tests are in the feature branch and CI credentials for testing will be setup.

playwright/calling/utils/constants.ts
eigengravy: Done — aligned test timeouts with SDK constants + consistent 5s buffer:
- `SDK_INIT_TIMEOUT`: 65s (`RETRY_TIMER_UPPER_LIMIT` 60s + 5s)
- `REGISTRATION_TIMEOUT`: 35s (`BASE_REG_RETRY_TIMER_VAL_IN_SEC` 30s + 5s)
- `OPERATION_TIMEOUT`: 15s (`SUPPLEMENTARY_SERVICES_TIMEOUT` 10s + 5s)

akulakum (approved):

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Discussed with Priya, since only a couple of test cases need to be repeated for `cc` I think we wouldn't need to split tests into folders.

docs/samples/calling/index.html
```diff
@@ -44,7 +44,7 @@ <h2 class="collapsible">Authentication</h2>
                   <div id="guest-container" class="hidden">
                     <input id="jwt-token-for-dest" name="jwtToken" placeholder="JWT token for destination" value="" type="text" style="margin: 0.5rem 0 0.5rem 0;">
                     <input id="guest-name" name="guestName" placeholder="Guest name" value="" type="text" style="margin: 0.5rem 0 0.5rem 0;">
-                    <button type="button" onclick="generateGuestToken()">Generate Guest Token [Prod only]</button>
+                    <button id="generate-guest-token" type="button" onclick="generateGuestToken()">Generate Guest Token [Prod only]</button>
```
Kesari3008, line 47: Why is this change required ?

package.json
Kesari3008, line 111: Do we not need the command(script) to run these tests here ?

packages/calling/playwright/utils/setup.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Set service domain before initialization (needed for contactcenter)
+ */
+export const setServiceDomain = async (page: Page, domain: string): Promise<void> => {
+  await page.locator(SELECTORS.SERVICE_DOMAIN).fill(domain, {timeout: AWAIT_TIMEOUT});
+};
+
+/**
+ * Navigate, init SDK, verify, and optionally set service and register line.
+ */
+export const initAndRegister = async (
```
Kesari3008, line 164: Where are we using this and when we have individually testcases calling the below methods anyway then why we need this method

navigateToCallingApp(page) anyway then 
  if (options.service) {
    await setServiceIndicator(page, options.service);
  } 
  await initializeCallingSDK(page, accessToken);
  await verifySDKInitialized(page);
  if (options.registerLine) {
    await registerLine(page);
    await verifyLineRegistered(page);

playwright/calling/utils/constants.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Transfer
+  TRANSFER_TARGET_INPUT: '#transfer_target',
+  TRANSFER_OPTIONS: '#transfer-options',
+  TRANSFER_BTN: '#transfer',
+  END_SECOND_CALL_BTN: '#end-second',
+  TRANSFER_STATUS: '#transfer-call',
+
+  // Media
+  GET_MEDIA_STREAMS_BTN: '#sd-get-media-streams',
+  LOCAL_VIDEO: '#local-video',
```
Kesari3008: Why have we added constants for local video and remote video, we do not have video mode supported in WebRTC ?

playwright/calling/utils/constants.ts
```diff
@@ -0,0 +1,63 @@
+export type ServiceIndicator = 'calling' | 'contactcenter' | 'guestcalling';
+
+export const SAMPLE_APP_PATH = '/samples/calling/';
+
+// Timeouts — SDK timeout + 5s buffer for network/UI overhead
+export const AWAIT_TIMEOUT = 10000; // General UI interactions
+export const SDK_INIT_TIMEOUT = 65000; // RETRY_TIMER_UPPER_LIMIT (60s) + 5s
+export const REGISTRATION_TIMEOUT = 35000; // BASE_REG_RETRY_TIMER_VAL_IN_SEC (30s) + 5s
```
Kesari3008: How did we come up with this timeout value ?

packages/calling/playwright/constants/selectors.ts
Kesari3008, line 1: Is this folder where we would be placing all our spec.ts file ? Can we have more structured folders? Like a separate folder saying constants.ts under which separate files can be created for CSS selector constants, then code related constants. Similarly, anything setup related can be in separate folder. Keeping everything under utils folkder would become a lot when we have all our flows testcases created

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
+import {test as setup} from '@playwright/test';
+import fs from 'fs';
+import path from 'path';
+
+const ENV_PATH = path.resolve(__dirname, '../../../.env');
```
Kesari3008: If we are having a dedicated constants file then we should keep these in those files

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+  browser: import('@playwright/test').Browser,
+  email: string,
+  password: string
+): Promise<string> => {
+  const context = await browser.newContext({ignoreHTTPSErrors: true});
+  const page = await context.newPage();
+
+  try {
+    // 1. Navigate to widgets sample app
+    await page.goto(WIDGETS_URL, {waitUntil: 'load', timeout: 30000});
```
Kesari3008: Shouldn't we be testing with Chrome extension or Calling SDK kitchend sin app, why are we using CC Widgets sample app

packages/calling/playwright/utils/setup.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Fill in the access token and initialize the Calling SDK.
+ * The sample app flow:
+ * 1. Enter token in #access-token
+ * 2. Click "Initialize Calling" (#access-token-save) which triggers initCalling()
+ * 3. Calling.init() is called, then on 'ready' event:
+ *    - authStatus shows "Saved access token!"
+ *    - calling.register() is called automatically
+ *    - After register, registerElm is enabled and callingClient + line are set up
+ */
+export const initializeCallingSDK = async (page: Page, accessToken: string): Promise<void> => {
```
Kesari3008, line 34: Are we only testing in prod, what about choosing Int environment  and fedramp ?

packages/calling/playwright/utils/setup.ts
```diff
@@ -0,0 +1,94 @@
...
+ * - Register button is enabled
+ * - window.callingClient is set (Calling object exists)
+ */
+export const verifySDKInitialized = async (page: Page): Promise<void> => {
+  await expect(page.locator(SELECTORS.AUTH_STATUS)).toHaveText('Saved access token!', {
+    timeout: SDK_INIT_TIMEOUT,
+  });
+  await expect(page.locator(SELECTORS.REGISTER_BTN)).toBeEnabled({timeout: SDK_INIT_TIMEOUT});
+
+  const hasCallingClient = await page.evaluate(() => !!(window as any).callingClient);
```
Kesari3008, line 67: We should also consider checking the config that is passed during initialization and based on that we should be checking which clients are active. There can be cases where we have CallingClient, CallHistoryClient, Voicemail etc instantiated too. I know we are focusing on CallingCleint specific tests for now but initialization should cover all scenarios or if we do not want to do that, add a TODO here to add expect statements for other clients

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    throw new Error(`${envVar} not set. Run OAuth setup first.`);
+  }
+
+  return token;
+};
+
+test.describe('SDK Initialization', () => {
+  test.describe.configure({mode: 'parallel'});
+
+  test('Normal Calling - init with calling service indicator', async ({page}) => {
```
Kesari3008: We should also have testcase where we are passing region and country during initialization. And verify mobius servers discovered based on that. HAve helper methods created to pass those values.

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    await setServiceIndicator(page, 'calling');
+
+    await initializeCallingSDK(page, getToken('CALLER_ACCESS_TOKEN'));
+    await verifySDKInitialized(page);
+  });
+
+  test('Contact Center - init with contactcenter service indicator', async ({page}) => {
+    await navigateToCallingApp(page);
+    await setServiceIndicator(page, 'contactcenter');
+    await setServiceDomain(page, 'rtw.prod-us1.rtmsprod.net');
```
Kesari3008: Create a constant for this domain string

playwright/calling/tests/01-sdk-initialization.spec.ts
Kesari3008, line 1: Also do we have options to test from different browser, we should be providing that

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  initializeCallingSDK,
+  verifySDKInitialized,
+  setServiceIndicator,
+  setServiceDomain,
+} from '../utils/setup';
+import {SELECTORS, SDK_INIT_TIMEOUT, AWAIT_TIMEOUT} from '../utils/constants';
+
+const getToken = (envVar: string): string => {
+  const token = process.env[envVar];
+  if (!token) {
```
Kesari3008: We are using 3 different tokens based on different scenarios, can that cause errors here? Maybe adding test.skip guards for optional users

Kesari3008 (changes requested):
1. Why are we using feature branch to merge E2E tests changes, how does this affect any working code files if we merge it next branch ?
2. If we are planning to submit this in separate branch then we should be making change in pul request yml to make the pipeline run?
3. Also do we have plan to add playwright.yml to run it in the pipeline. Maybe we should add that in initial PRs itself where we are setting the playwright config up

playwright/calling/utils/registration.ts
```diff
@@ -0,0 +1,24 @@
+import {Page, expect} from '@playwright/test';
+import {SELECTORS, AWAIT_TIMEOUT, REGISTRATION_TIMEOUT} from './constants';
+
+export const registerLine = async (page: Page): Promise<void> => {
```
Kesari3008: This method and below method can be combined

packages/calling/playwright/utils/setup.ts
Kesari3008, line 34: Same comment here, below and this can be combined and we should also be checking that after initialization init button and token textbox getting disabled

playwright/calling/tests/01-sdk-initialization.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return token;
+};
+
+test.describe('SDK Initialization', () => {
+  test.describe.configure({mode: 'parallel'});
+
+  test('Normal Calling - init with calling service indicator', async ({page}) => {
+    await navigateToCallingApp(page);
```
Kesari3008: Consider looking into approach of creating Page Object Model and using the page class objects to do verification directly under tests

eigengravy:
@Kesari3008 

> 1. Why are we using feature branch to merge E2E tests changes, how does this affect any working code files if we merge it next branch ?
> 2. If we are planning to submit this in separate branch then we should be making change in pul request yml to make the pipeline run?
> 3. Also do we have plan to add playwright.yml to run it in the pipeline. Maybe we should add that in initial PRs itself where we are setting the playwright config up

The idea was to first merge everything into a feature branch and then open a PR to next with the playwright test added to the pipeline once the CI testing accounts are setup as well. Currently I am testing with the Bifrost Testing Accounts.

package.json
eigengravy: Done — added `test:e2e:calling` script: `"test:e2e:calling": "yarn playwright test --project='Calling SDK E2E'"`

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Done — extracted into `CC_SERVICE_DOMAIN` constant in `constants.ts`.

playwright/calling/utils/oauth.setup.ts
eigengravy: CC Widgets Sample app is only used to login and get the OAuth token which is later used in the Calling SDK Kitchen Sink App.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Adding config for firefox, edge and safari and ability to set broswer an env variable. Should we run tests across multiple browsers in the test command?

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: I am using the 3 available accounts to run the tests (that only require one account) in parallel sets. Here the 3 SDK Init tests run in parallel and in the following Device Registration tests,  I have grouped them into sets with each account mapped to a set.

docs/samples/calling/index.html
eigengravy: The id="generate-guest-token" was added to the button in the sample app HTML so the Playwright test can target it with a stable #generate-guest-token CSS selector. Previously the test used a fragile text-based selector (button:has-text("Generate Guest Token")) which would break if the button label ever changed. This was flagged by Akula in the review as a nit — adding an id makes the selector resilient to label changes.

packages/calling/playwright/utils/setup.ts
eigengravy: `initAndRegister` is used in the device registration tests (next PR) where each account's `beforeAll` needs to navigate → init → register. Avoids repeating the same 5-step sequence in every test group.

playwright/calling/utils/constants.ts
eigengravy: Done — removed `LOCAL_VIDEO` and `REMOTE_VIDEO` selectors since calling doesn't have video.

playwright/calling/utils/constants.ts
eigengravy: Each timeout is derived from the corresponding SDK constant + 5s buffer for network/UI overhead:

- `SDK_INIT_TIMEOUT` (65s) = [`RETRY_TIMER_UPPER_LIMIT`](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/constants.ts#L111) (60s) + 5s
- `REGISTRATION_TIMEOUT` (35s) = [`BASE_REG_RETRY_TIMER_VAL_IN_SEC`](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/constants.ts#L105) (30s) + 5s
- `OPERATION_TIMEOUT` (15s) = [`SUPPLEMENTARY_SERVICES_TIMEOUT`](https://github.com/webex/webex-js-sdk/blob/next/packages/calling/src/CallingClient/constants.ts#L57) (10s) + 5s

packages/calling/playwright/constants/selectors.ts
eigengravy: Done — split into `playwright/calling/constants/` with separate files:
- `selectors.ts` — CSS selectors for sample app elements
- `timeouts.ts` — SDK-aligned timeout values
- `index.ts` — types, app paths, OAuth URLs, and re-exports

playwright/calling/utils/oauth.setup.ts
eigengravy: Done — moved `ENV_PATH` and `WIDGETS_URL` from `oauth.setup.ts` into `constants/index.ts`.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Added `setRegion` and `setCountry` helpers in `setup.ts`, along with `REGION_INPUT`/`COUNTRY_INPUT` selectors and `REGION`/`COUNTRY` constants. The test case itself will be added in the device registration PR where we have the TestManager for proper account isolation across parallel tests.

playwright/calling/utils/registration.ts
eigengravy: Keeping them separate — in the device registration tests (next PR), `verifyLineRegistered` is used standalone to assert auto-re-registration after network disconnect (REG-008) where the SDK re-registers automatically without us calling `registerLine`.

packages/calling/playwright/utils/setup.ts
eigengravy: Currently we are only testing in Prod. What all tests should be repeated in Int and FedRAMP?

packages/calling/playwright/utils/setup.ts
eigengravy: Added a TODO comment in `verifySDKInitialized` to check which clients are active based on the initialization config. Currently we only verify `callingClient` exists — the TODO notes that different configs can also instantiate CallHistoryClient, Voicemail, etc. and we should add expect statements for each.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: We're using the same utility-function pattern as the widgets repo. I did a feasibility analysis on POM and it doesn't work well with our test patterns — we rely on `context.route()` for network interception, `browserContext.newPage()` for multi-page scenarios, and shared browser context across chained tests (e.g. failover where one test's backup state feeds into the next). These are context-centric patterns that don't map cleanly to page class objects. The current utility-function approach keeps things flexible for route mocking and shared context.

package.json
```diff
@@ -67,6 +67,7 @@
     "samples:serve": "webpack serve --color --env NODE_ENV=development",
     "samples:serve:integration": "webpack serve --color --env NODE_ENV=test",
     "samples:test": "NODE_ENV=test npx wdio run wdio.conf.js",
+    "test:e2e:calling": "yarn playwright test --project='Calling SDK E2E'",
```
Shreyas281299: I think we should have something different here, the existing scripts here are all designed in the following manner

yarn worksapce --forEach run <SOME_COMMAND>

So maybe we can have a similar design here for e2e tests

```suggestion
    "test:e2e": "yarn workspace --foreach run test:e2e,
```

And then we can add the specific test:e2e command in the calling pack. This way if we add e2e tests to more packages we dont have to change the root scripts.

playwright.config.ts
```diff
@@ -0,0 +1,90 @@
...
+      dependencies: ['Calling: OAuth Setup'],
+      testDir: './playwright/calling/tests',
+      use: browserOptions[PW_BROWSER],
+    },
+    // TODO: Uncomment when Contact Center tests are added
+    // {
+    //   name: 'Contact Center SDK E2E',
+    //   testDir: './playwright/contact-center',
+    //   use: browserOptions[PW_BROWSER],
+    // },
```
Shreyas281299: Do we really need this?

packages/calling/playwright/constants/selectors.ts
Shreyas281299: good call

playwright/calling/constants/selectors.ts
```diff
@@ -0,0 +1,53 @@
+// Element selectors (from the calling sample app)
+export const SELECTORS = {
```
Shreyas281299: NIT: Does it make sense to call this CALLING_SELECTORS?

playwright/calling/tests/01-sdk-initialization.spec.ts
Shreyas281299: IMO we should only run 1 browser while testing locally, MAYBE in the pipeline we can have all the browsers running, Locally we just need an indicator if our tests are passing or not, so running just 1 browser is enough. 

@Kesari3008

playwright/calling/tests/01-sdk-initialization.spec.ts
Shreyas281299: We are doing something similar in cc-widget's tests as well, just to speed up the process, otherwise the time taken to run these tests increases drastically (historically 15mins -> 32 mins, ccwidgets)

Shreyas281299: Marking this as resolved

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,146 @@
...
+type EnvUpdateMap = Record<string, string>;
+
+const readEnvFile = (): string => {
+  if (!fs.existsSync(ENV_PATH)) {
+    return '';
+  }
+
+  return fs.readFileSync(ENV_PATH, 'utf8');
+};
+
```
Shreyas281299: can we not use process.env here? or maybe dotenv?

playwright/calling/utils/oauth.setup.ts
Shreyas281299: Yeah, calling-sdk doesnt have oAuth so that wont be possible, as for chrome extensino it would be a tedious process of getting the token from the storage. I think we can move ahead with this.

packages/calling/playwright/utils/setup.ts
Shreyas281299: The same set of tests should be in INT as the functionality is same in INT and PROD. As for Fedramp we have callHistory, callsetting and voicemail client related functionality

Shreyas281299 (approved):

packages/calling/playwright/utils/setup.ts
Shreyas281299: Dont know if this should be a TODO or this can be taken as part of this ticekt @eigengravy  you can take a call here.

Shreyas281299:
@eigengravy  what about this comment from @Kesari3008 
> Why are we using feature branch to merge E2E tests changes, how does this affect any working code files if we merge it next branch ?

Shreyas281299 (changes requested):
Still need a few clarification around the feature branch.

Shreyas281299 (approved):
Missed the comment around feature branch- Approving the PR

playwright/calling/utils/oauth.setup.ts
Kesari3008: Why aren't we trying to use developer.webex.com to fetch the token? Did we try that ?

packages/calling/playwright/utils/setup.ts
Kesari3008: To test in INT, only change that happens is change of credentials and selecting the environment before SDK initialization so I believe for INT we can support that here itself. Fedramp tests can be accommodated separately

playwright/calling/tests/01-sdk-initialization.spec.ts
Kesari3008: I am not talking about running here locally. I meant these tests run should happen in different browsers in pipeline only. We should setup the config for it

playwright/calling/tests/01-sdk-initialization.spec.ts
Kesari3008: I believe my point is missed here. What I meant is in a test if other 2 tokens are not needed, add skip guards to avoid running into the error which we are thrown for the condition (if !token)

Kesari3008:
> @Kesari3008
> 
> > 1. Why are we using feature branch to merge E2E tests changes, how does this affect any working code files if we merge it next branch ?
> > 2. If we are planning to submit this in separate branch then we should be making change in pul request yml to make the pipeline run?
> > 3. Also do we have plan to add playwright.yml to run it in the pipeline. Maybe we should add that in initial PRs itself where we are setting the playwright config up
> 
> The idea was to first merge everything into a feature branch and then open a PR to next with the playwright test added to the pipeline once the CI testing accounts are setup as well. Currently I am testing with the Bifrost Testing Accounts.

We should discuss this, because this is not a change in code and we have anyway not added it in pipeline which will block any of our upcoming PRs from merging so merging to next directly should be ok

playwright/calling/tests/01-sdk-initialization.spec.ts
Kesari3008: These testcases should be added here because region discovery happens during initialization itself

playwright/calling/utils/oauth.setup.ts
eigengravy: Yup tried developer.webex.com but it needed to rely on clipboard copy mechanism since the token is not show in plain text on the website. The tests were failing intermittently and required a large timeout, so I switched to using the widgets URL.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: Added a testcase with US and US-EAST.

playwright/calling/tests/01-sdk-initialization.spec.ts
eigengravy: I have already added configs for the major browsers which can be set via an env variable when running in the CI like so.

```
PW_BROWSER=chrome yarn test:e2e:calling
PW_BROWSER=firefox yarn test:e2e:calling
PW_BROWSER=edge yarn test:e2e:calling
```

packages/calling/playwright/utils/setup.ts
eigengravy: Added support to test in INT as well.

playwright/calling/utils/oauth.setup.ts
eigengravy: Got developer.webex.com working through intercept the get token network request. This has also allowed support to login to INT environment easily/

Kesari3008 (approved):

PR #4799, opened 2026-03-24
test(calling): add Playwright E2E tests for SDK init and device registration

eigengravy:
# COMPLETES https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7737
# COMPLETES https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7738

Playwright E2E tests for the Calling SDK covering SDK initialization and device registration flows, using a phased parallel execution architecture that runs 3 accounts across 3 parallel workers.

## Architecture: Phased Parallel Execution

The test suite uses a **phased parallel execution** model to work within the constraint that the same Webex account must never be in two browsers simultaneously:

- **Phase 1 — OAuth**: Fetch tokens for USER_1, USER_2, USER_3 accounts (dependency project, PROD + INT)
- **Phase 2 — Test Sets**: 3 parallel workers (SET_1, SET_2, SET_3), each owning one account and running a subset of tests

Each phase fully completes before the next starts. Playwright project dependencies enforce ordering. Projects are generated dynamically from a `USER_SETS` config — each set gets a PROD and INT variant automatically.

### Test Distribution Across Workers

| Worker (Suite) | Tests | Pattern |
|---|---|---|
| `SET_1` (set-1.spec.ts) | SDK init (5 tests) + REG-001, 003, 008, 010, 011 | SDK initialization, registration lifecycle, error flows |
| `SET_2` (set-2.spec.ts) | REG-006, 017, 007 | Primary/backup failover & failback (chained shared context) |
| `SET_3` (set-3.spec.ts) | REG-004, 005, 015, 016 | Keepalive & 429 retry flows (fresh context per test) |

### New test infrastructure
- `playwright.config.ts` — Config with dynamic project generation from `USER_SETS` and phased dependencies
- `playwright/test-manager.ts` — TestManager pattern (modeled after widgets) for shared browser context lifecycle
- `playwright/test-data.ts` — Account/set definitions, token helpers, project name utilities
- `playwright/utils/registration.ts` — Registration action helpers (register, unregister, status checks)
- `playwright/utils/setup.ts` — SDK init helpers (navigation, init, discovery capture)
- `playwright/constants/` — Selectors, timeouts, Mobius URL constants

### Test groups (shared logic, not spec files)
- `sdk-init.ts` — SDK initialization: normal calling, contact center, guest calling, explicit region/country, negative case
- `registration-lifecycle.ts` — Serial tests with shared browser via TestManager (REG-001, 003, 008, 010)
- `registration-failover.ts` — Serial chained tests for failover/failback (REG-006 → REG-017 → REG-007)
- `registration-keepalive.ts` — Fresh context per test for keepalive and 429 retry flows (REG-004, 005, 015, 016)
- `registration-errors.ts` — Invalid token registration (REG-011)

### Suite entry points (thin spec files)
- `suites/set-1.spec.ts` — SDK init + registration lifecycle + errors
- `suites/set-2.spec.ts` — Failover & failback
- `suites/set-3.spec.ts` — Keepalive flows

### OAuth setup
- `utils/oauth.setup.ts` — Data-driven OAuth for all unique roles derived from `USER_SETS`

### Techniques used
- Playwright route interception to count/mock Mobius registration, keepalive, and delete requests
- Response modification to shorten `keepaliveInterval` and `rehomingInterval` for faster test execution
- `context.setOffline(true/false)` for network disruption simulation (REG-008)
- `page.evaluate()` to read SDK internals (registration status, active Mobius URL, device info)
- Route-based request counting to assert negative cases (no requests made) instead of long hard sleeps
- Timestamp tracking to verify Retry-After backoff timing (REG-005, REG-015)

### SET_1 — SDK Init + Registration Lifecycle + Errors (~8 min)

| Test | Approx Time | Description |
|---|---|---|
| Normal Calling | ~30s | Init with `calling` service indicator, verify auto-discovery |
| Contact Center | ~20s | Init with `contactcenter` indicator and service domain |
| Guest Calling | ~20s | Generate guest token and init (prod-only, skipped on INT) |
| Region/Country | ~30s | Init with explicit region, verify discovery params and server regions |
| Negative (empty token) | ~15s | Empty token init blocked, no Mobius requests made, buttons stay disabled |
| REG-001 | ~30s | Initial registration success with primary Mobius server |
| REG-003 | ~25s | Keepalive requests sent after registration |
| REG-008 | ~3 min | Connection restoration re-registers when no active calls |
| REG-010 | ~20s | Deregistration success and cleanup |
| REG-011 | ~30s | Registration fails with invalid token (401) |

### SET_2 — Failover & Failback (~10 min)

| Test | Approx Time | Description |
|---|---|---|
| REG-006 | ~4 min | Primary-to-backup failover on repeated 503 failure |
| REG-017 | ~4 min | 429 during failback exhausts retry budget, stays on backup |
| REG-007 | ~2 min | Failback to primary from backup |

### SET_3 — Keepalive & 429 Retry Flows (~12 min)

| Test | Approx Time | Description |
|---|---|---|
| REG-004 | ~3 min | Keepalive 404 triggers deregister→register cycle, keepalive resumes |
| REG-005 | ~2 min | 429 Retry-After honored on keepalive (timing gap verified) |
| REG-015 | ~3 min | 429 on initial registration honors Retry-After before retrying |
| REG-016 | ~4 min | 429 with high Retry-After triggers immediate backup failover |

### Deferred

| ID | Description | Reason |
|---|---|---|
| REG-009 | Connection restoration deferred when active calls exist | Blocked on CAI-7739 (call tests) |

**Total: 17 tests (5 SDK init + 12 registration) across 3 parallel workers. All 3 sets run concurrently, so wall-clock time is ~12 min (longest set).**

  - [ ] Github Copilot
  - [x] Other - Claude Code
- [x] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [x] Automation

### I certified that


playwright/calling/suites/sdk-init.spec.ts
```diff
@@ -0,0 +1,3 @@
+import {sdkInitTests} from '../test-groups/sdk-init';
```
Kesari3008: Why do we need this layer to invoke the specific test flows ?

playwright/calling/suites/set-transfer.spec.ts
```diff
@@ -0,0 +1,3 @@
+import {registrationKeepaliveTests} from '../test-groups/registration-keepalive';
+
+registrationKeepaliveTests('TRANSFER');
```
Kesari3008: Why are we passing CALLER, CALLEE and TRANSFER for registration tests

playwright/calling/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+      );
+      expect(isRegistered).toBe(true);
+
+      const activeMobiusUrl = await page.evaluate(() =>
+        (
+          Object.values((window as any).callingClient.getLines())[0] as any
+        ).registration.getActiveMobiusUrl()
+      );
+      expect(activeMobiusUrl).toBeTruthy();
+      expect(activeMobiusUrl).toContain('mobius');
```
Kesari3008: Please compare complete Mobius URL

packages/calling/playwright/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      // Track Mobius registration and delete requests across all tests
+      await context.route(/\/calling\/web\/device$/, async (route) => {
+        if (route.request().method() === 'POST') {
+          registrationPosts += 1;
+        }
+        await route.continue();
+      });
+
+      await context.route(/\/calling\/web\/devices\/[^/]+$/, async (route) => {
```
Kesari3008, line 67: Why are we deleting registration right away here and if we are doing so to test it then add some expect statements for delete flow as well

playwright/calling/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+
+    test.afterAll(async () => {
+      await tm.cleanup();
+    });
+
+    test('REG-001: Initial registration success', async () => {
+      const page = tm.page;
+
+      expect(registrationPosts).toBeGreaterThanOrEqual(1);
```
Kesari3008: This would be equal to one right for this testcase

playwright/calling/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+      const activeMobiusUrl = await page.evaluate(() =>
+        (
+          Object.values((window as any).callingClient.getLines())[0] as any
+        ).registration.getActiveMobiusUrl()
+      );
+      expect(activeMobiusUrl).toBeTruthy();
+      expect(activeMobiusUrl).toContain('mobius');
+
+      const deviceInfo = await page.evaluate(() => {
+        const ln = Object.values((window as any).callingClient.getLines())[0] as any;
```
Kesari3008: This seems to be used in repetition so maybe we can fetch it once at the start, store it in a variable and keep using that. During CallingClient initialization line should already be created so it should be available for registration testcases

packages/calling/playwright/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+      const deviceInfo = await page.evaluate(() => {
+        const ln = Object.values((window as any).callingClient.getLines())[0] as any;
+
+        return JSON.parse(JSON.stringify(ln.registration.getDeviceInfo()));
+      });
+      expect(deviceInfo.device).toBeTruthy();
+      expect(deviceInfo.device.deviceId).toBeTruthy();
+    });
+
+    test('REG-002: Repeated register calls are idempotent', async () => {
```
Kesari3008: What are we trying to test here ? I believe register button would be disabled so do we still need this ?

playwright/calling/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+      expect(deletePosts).toBeGreaterThanOrEqual(1);
+
+      const isRegistered = await page.evaluate(() =>
+        (
+          Object.values((window as any).callingClient.getLines())[0] as any
+        ).registration.isDeviceRegistered()
+      );
+      expect(isRegistered).toBe(false);
+
+      await expect(page.locator(SELECTORS.REGISTER_BTN)).toBeEnabled({timeout: AWAIT_TIMEOUT});
```
Kesari3008: We can add one more expect here for checking the string that is printed: 'Unregistered' on the sample app

Kesari3008 (changes requested):
Partial review done, will finish rest tomorrow

playwright/calling/test-groups/registration-lifecycle.ts
eigengravy: Done — changed to `toBe(1)` since `beforeAll` does exactly one registration POST before this test runs. Done in d48b687.

playwright/calling/test-groups/registration-lifecycle.ts
eigengravy: Done — extracted three shared helpers (`isLineRegistered`, `getActiveMobiusUrl`, `getDeviceInfo`) into `registration.ts` and replaced all 19 inline occurrences across the 4 test-group files. Done in d48b687.

playwright/calling/test-groups/registration-lifecycle.ts
eigengravy: Done — replaced the loose `.toContain('mobius')` with exact URL assertions using `PRIMARY_MOBIUS_URL` and `BACKUP_MOBIUS_URL` constants defined in `constants.ts`. Now verified in:
- **REG-001**: active Mobius URL === `PRIMARY_MOBIUS_URL`
- **REG-006**: active Mobius URL after failover === `BACKUP_MOBIUS_URL`
- **REG-007**: backup URL carried from REG-006 === `BACKUP_MOBIUS_URL`, active URL after failback === `PRIMARY_MOBIUS_URL`

Done in 38b3a13.

packages/calling/playwright/test-groups/registration-lifecycle.ts
eigengravy: Yup it is disabled in the UI. Since it was in the test plan, I am invoking the SDK method directly to test idempotency handled by the SDK. I can remove this if not needed. This is handled by the SDK but is not tested in the unit tests.

playwright/calling/suites/sdk-init.spec.ts
rsarika: do we really need suites here? as we are running everything parallelly?

for cc sdk we have to design that. does it really a[ply here as well?

packages/calling/playwright/test-groups/registration-failover.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    test.beforeAll(async ({browser}, testInfo) => {
+      tm = new TestManager(testInfo.project.name);
+      const {context} = await tm.setupContext(browser, 0, {
+        initSDK: true,
+        service: 'calling',
+      });
+
+      // Intercept registration POST — fail primary during failover, allow during failback
+      await context.route(/\/calling\/web\/device$/, async (route) => {
```
rsarika, line 46: this will be websocket message in the new mobius socket effort. how to tackle that.
keep that design in mind to avoid rework

playwright/calling/utils/oauth.setup.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  envContent = envContent.replace(/\n{3,}/g, '\n\n');
+  fs.writeFileSync(ENV_PATH, envContent, 'utf8');
+};
+
+/**
+ * Login via widgets.webex.com OAuth flow and grab the access token.
+ *
+ * Flow:
+ * 1. Navigate to widgets.webex.com sample app
```
rsarika: IS this ok? to fetch token using widgets ui?

playwright/calling/suites/set-transfer.spec.ts
eigengravy: Moved to automatic accounts to set mapping similar to widgets repo.

playwright/calling/utils/oauth.setup.ts
eigengravy: Moved to developer portal login to support INT login as well.

packages/calling/playwright/test-groups/registration-lifecycle.ts
eigengravy: It's only intercepting and counting DELETE requests that the SDK makes on its own.
The deletePosts counter is then asserted in two places:
  - REG-008 (line 134): verifies that reconnection triggers a delete (cleanup before re-register)
  - REG-010 (line 152): verifies that explicit deregistration sends a DELETE request

playwright/calling/suites/sdk-init.spec.ts
eigengravy: Since different tests need 1,2 and 3 users, to effectively parallelize them, suites were the only way to create a dependency chain between groups of tests.

packages/calling/playwright/test-groups/registration-failover.ts
eigengravy: Yup, will handle the reading of ws messages and apply similar test logic once its ready in a later ticket.

packages/calling/playwright/test-groups/registration-lifecycle.ts
```diff
@@ -0,0 +1,166 @@
...
+    });
+
+    test.afterAll(async () => {
+      await tm.cleanup();
+    });
+
+    test('REG-001: Initial registration success', async () => {
+      const page = tm.page;
+
+      expect(registrationPosts).toBeGreaterThanOrEqual(1);
```
Kesari3008: This should be ideally Equal to 1, why are we checking here for more than 1

packages/calling/playwright/test-groups/registration-lifecycle.ts
Kesari3008: Checking that once registration is succesful and button is diabled and not clikable anymore should be enough and that can be checked in the first textcase itself. We do not need separate testcase I believe

packages/calling/playwright/suites/set-3.spec.ts
```diff
@@ -0,0 +1,4 @@
+import {registrationKeepaliveTests} from '../test-groups/registration-keepalive';
+
+// Account role is resolved from testInfo.project.name → USER_SETS.
+registrationKeepaliveTests();
```
Kesari3008, line 4: We do not need separate suite for keepalives, this actually can be combined right after lifecycle tests. 
We can create testuites in this manner:

Suite1: Initialization followed by initial registration success followed by keepalive
Suite2: Failover followed by Failback tests, even error cases can be combined here.
Suite3: Inbound/outbound call followed by call keeaplive
Suite 4: Supplementary services 
Suite 5: Call specific error handling

packages/calling/playwright/test-groups/registration-errors.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      const hasCallingClient = await page.evaluate(() => !!(window as any).callingClient);
+      expect(hasCallingClient).toBe(false);
+
+      await expect(page.locator(CALLING_SELECTORS.REGISTER_BTN)).toBeDisabled({
+        timeout: AWAIT_TIMEOUT,
+      });
+    });
+
+    test('REG-013: Registration not possible before SDK init', async ({page}) => {
```
Kesari3008: This is not really and error case, it is just not allowed and can be a part of lifecycle tests itself

packages/calling/playwright/test-groups/registration-errors.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      if (registrationPosts > 0) {
+        expect(registrationStatus).toBe(401);
+      }
+
+      const status = await page.locator(CALLING_SELECTORS.REGISTRATION_STATUS).textContent();
+      expect(status).not.toMatch(/Registered, deviceId:/);
+    });
+
+    test('REG-012: SDK init fails with empty token', async ({page}) => {
```
Kesari3008: Same comment applies here, if sdk init fails. This is not a registration error but initlaization failure case and should be part of sdk-init.ts test group

packages/calling/playwright/test-groups/registration-failover.ts
```diff
@@ -0,0 +1,153 @@
...
+  PRIMARY_MOBIUS_URL,
+  BACKUP_MOBIUS_URL,
+} from '../constants';
+
+/**
+ * Failover & failback tests: REG-006, REG-007.
+ * Run serially in a shared browser context — REG-007 chains from REG-006's
+ * backup-registered state to test failback without repeating the failover phase.
+ */
+export function registrationFailoverTests() {
```
Kesari3008, line 19: There are 429 error handling done in different manner, would be good to check if there is something we can add for 429 cases

Kesari3008 (commented):
Update to the latest branch

packages/calling/playwright/test-groups/sdk-init.ts
```diff
@@ -0,0 +1,105 @@
...
+  COUNTRY,
+} from '../constants';
+
+/**
+ * SDK initialization tests: Normal Calling, Contact Center, Guest Calling, Region/Country.
+ * All tests use a single account from the owning set.
+ */
+export function sdkInitTests() {
+  test.describe('SDK Initialization', () => {
+    test('Normal Calling - init with calling service indicator', async ({page}, testInfo) => {
```
Kesari3008, line 34: There should be expect statements to verify all available options for service indicator

packages/calling/playwright/test-groups/sdk-init.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      const mobiusDiscoveryRequest = waitForMobiusDiscoveryRequest(page, {
+        region: REGION,
+        country: COUNTRY,
+      });
+
+      await initializeCallingSDK(page, getToken(role, isInt));
+      await verifySDKInitialized(page);
+
+      await expect(mobiusDiscoveryRequest).resolves.toContain(
```
Kesari3008, line 143: These expect would pass even for case where no region and country are passed. Expect statements should verify that there was no region discovery done and mobius url used match the region or country specified or matches the one based on region discovery. Let's discuss this over call to avoid confusion

packages/calling/playwright/test-groups/registration-lifecycle.ts
eigengravy: Done, checking the button is disabled after successful registration in REG-001. Removed the REG-002 testcase.

packages/calling/playwright/test-groups/registration-errors.ts
eigengravy: Moved to SDK Init as an error test case.

packages/calling/playwright/test-groups/registration-errors.ts
eigengravy: Moved to SDK Init as an error test case.

packages/calling/playwright/test-groups/sdk-init.ts
eigengravy: Without the region passed there is a region request to `https://ds.ciscospark.com/v1/region` followed by
`https://mobius-us-east-2.prod.infra.webex.com/api/v1/calling/web/?regionCode=AP-SOUTHEAST&countryCode=IN`.

When the region is passed there is no region request and the mobius request is to `https://mobius-us-east-2.prod.infra.webex.com/api/v1/calling/web/?regionCode=US-EAST&countryCode=US`. Right now I'm only checking the region and country code in this mobius url.

So I'll add an extra expect to confirm that no request was made to the region discovery endpoint?

packages/calling/playwright/test-groups/registration-failover.ts
eigengravy: 429 handling is covered by REG-005 (registration-keepalive.ts:157). That test intercepts the first keepalive with a 429 + Retry-After: 10 header and verifies the SDK honors the delay before resuming keepalives.

Is there any other case that needs be tested for 429?

packages/calling/playwright/suites/set-3.spec.ts
eigengravy: The current suite breakup is to use 3 accounts in parallel with a suite each since set-1, set-2, set-3 need only a single account each. However the call tests would need 2/3 accounts at a time.

packages/calling/playwright/test-groups/registration-lifecycle.ts
eigengravy: Fixed this to check equal to 1.

rsarika (approved):

packages/calling/playwright/test-groups/sdk-init.ts
```diff
@@ -0,0 +1,217 @@
...
+
+      // Verify discovery response has distinct primary and backup server regions.
+      // Auto-discovered region depends on the runner's location, so we only check
+      // that both regions are present and differ from each other.
+      const discoveryResponse = await discoveryResponsePromise;
+      expect(discoveryResponse.primary?.uris?.length).toBeGreaterThan(0);
+      expect(discoveryResponse.backup?.uris?.length).toBeGreaterThan(0);
+      expect(discoveryResponse.primary.region).toBeTruthy();
+      expect(discoveryResponse.backup.region).toBeTruthy();
+      expect(discoveryResponse.primary.region).not.toBe(discoveryResponse.backup.region);
```
Kesari3008: This may not always be the case so we can avoid this expect

Kesari3008 (approved):

Kesari3008:
Ensure to add validated label in each PR else pipeline checks won't run

packages/calling/playwright/test-groups/sdk-init.ts
eigengravy: Done - removed the expect statement.

PR #4816, opened 2026-03-27
feat(plugin-meetings): add AV1 codec support and related constants

fnowakow:
# COMPLETES [SPARK-797573](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-797573)

Multistream meetings currently only support H.264 codec for video slides. This PR adds AV1 codec support for video slides in multistream media connections, enabling better compression efficiency and quality at lower bitrates when the AV1 codec is available and negotiated.

- **New AV1 codec constants**: Added `codec/constants.ts` with resolution-based AV1 encoding parameters (90p through 1080p) including `levelIdx`, `tier`, `maxWidth`, `maxHeight`, `maxPicSize`, and `maxDecodeRate`. Extracted H264 codec defaults into the same constants file.

- **Dynamic ingress payload type resolution**: Introduced a `getIngressPayloadTypeCallback` that queries the active `webrtcMediaConnection` for the correct payload type per codec/media type at runtime, replacing the previously hardcoded `0x80` H264 payload type.

- **AV1 codec info in media requests**: Extended `MediaRequestManager` to build and include AV1 `WcmeCodecInfo` entries alongside H264 when `enableAv1` is set. AV1 encoding parameters are resolved from the requested frame size by mapping to the appropriate resolution bucket.

- **Configuration and API surface**: Added `enableAv1SlidesSupport` config flag (default `false`), wired it through `Media.createMediaConnection` into the multistream connection config, and added `_toggleEnableAv1SlidesSupport()` on the `Meetings` class to toggle it before joining.

- **Degradation support**: AV1 codec parameters degrade in lockstep with H264 when the macroblock limit is exceeded, ensuring consistent quality reduction across codecs.

- **Test coverage**: Comprehensive unit tests for AV1 resolution mapping, codec info generation, degradation behavior with AV1 enabled, payload type callback integration, and the new toggle API.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Verified AV1 codec parameters are correctly mapped for all resolution buckets (90p, 180p, 360p, 540p, 720p, 1080p)
- Verified frame sizes between breakpoints map to the correct next-higher resolution bucket
- Verified frame sizes exceeding 1080p default to 1080p AV1 parameters
- Verified AV1 codec info is included in both active-speaker and receiver-selected stream requests when enabled
- Verified AV1 codec info is **not** included when `enableAv1` is `false` (default behavior unchanged)
- Verified AV1 codec info is **not** included when the AV1 ingress payload type is unavailable (callback returns `undefined`)
- Verified degradation correctly adjusts AV1 parameters alongside H264 when macroblock limits are exceeded
- Verified degradation with mixed resolutions only downgrades the largest streams
- Verified no degradation occurs when receive slot sources are not live
- Verified `_toggleEnableAv1SlidesSupport` correctly updates config and rejects non-boolean values
- Verified `enableAV1SlidesSupport` flag is passed through to `MultistreamConnectionConfig`
- All existing unit tests continue to pass

  - [ ] Github Copilot
  - [x] Other - Please Specify
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -298,6 +321,36 @@ export class MediaRequestManager {
     // map all the client media requests to wcme stream requests
     Object.values(clientRequests).forEach((mr) => {
       if (mr.receiveSlots.length > 0) {
+        const codecInfos: WcmeCodecInfo[] = [];
+
+        if (mr.codecInfo) {
+          const h264CodecInfo = WcmeCodecInfo.fromH264(
+            0x80,
```
fnowakow: [getIngressPayloadType](https://sqbu-github.cisco.com/WebExSquared/webrtc-media-core/pull/367/files#diff-104955bcfdc25946796b8be8baabafd28f5222781eb2cbe59593aa41c39f563f) is available in multistream connection, that we don't have access in mediaRequestManager

Should we delegate creating `WcmeCodecInfo.fromAv1` and `fromH264` in WCME/media-core? 

Currently do update this values we either need to pass mediaconnection to mediaRequestManager (or proxy it like sendMediaRequestCallback - smth like `getIngressPayloadTypeSupplier()`) or update it when we have access to media connection (current changes)


Currently we also need to have imports for
`CodecInfo` - js-sdk encoding params for buildingAv1/H264Codec
`[Av1/H264]Codec` - jmp encoding params for building WcmeCodecInfo
`WcmeCodecInfo` - json-multistream class to build codecinfo in jmp message

This should be simplified in wcme so at this level (js-sdk) we don't need to use json-multistream classes and calculate payload type

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -965,6 +941,21 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      codecMimeType: MediaCodecMimeType
+    ) => {
+      if (this.mediaProperties?.webrtcMediaConnection instanceof MultistreamRoapMediaConnection) {
+        return this.mediaProperties.webrtcMediaConnection.getIngressPayloadType(
+          mediaType,
+          codecMimeType
+        );
+      }
+
+      return 0;
```
k-wasniowski: payload type 0 is a valid payload type. In this context you consider it as "invalid", but we should rather return "undefined" or "null"

k-wasniowski: payload type 0 is "PCMU" codec standard FYI

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -965,6 +941,25 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      mediaType: MediaType,
+      codecMimeType: MediaCodecMimeType
+    ) => {
+      if (this.isMultistream) {
+        try {
+          return this.mediaProperties.webrtcMediaConnection.getIngressPayloadType(
+            mediaType,
+            codecMimeType
+          );
+        } catch (error) {
```
k-wasniowski, line 979: Shall we log an error here?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1042,16 +1042,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
           );
         },
+        getIngressPayloadTypeCallback,
         {
           // @ts-ignore - config coming from registerPlugin
           degradationPreferences: this.config.degradationPreferences,
           kind: 'video',
           trimRequestsToNumOfSources: false,
+          // @ts-ignore - config coming from registerPlugin
+          enableAV1: this.config.enableAV1VideoSlides,
```
k-wasniowski: I think we should use `enableAv1`? and across all other places.
Look that we have for example `backendIpv6NativeSupport` not `backendIPV6NativeSupport`. We don't use `HTML`, but rather `Html` semantic

k-wasniowski: refers to all occurences

packages/@webex/plugin-meetings/src/config.ts
```diff
@@ -88,11 +88,12 @@ export default {
       enableAdhocMeetings: true,
       enableTcpReachability: false,
       enableTlsReachability: false,
-      storeLocusHashTreeEventsForDebugging: false,
     },
     degradationPreferences: {
       maxMacroblocksLimit: 8192,
     },
+    // !! TODO: this is a temporary config to enable AV1 video slides for testing purposes
```
k-wasniowski: We don't need this comment. I believe it's not for testing only, but we should keep this for  longer period of time

packages/@webex/plugin-meetings/src/meeting/index.ts
edvujic: good point :)

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
edvujic: @k-wasniowski, any thoughts on this?

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,198 @@ describe('MediaRequestManager', () => {
...
+          codecInfos: sinon.match((codecInfos) =>
+            codecInfos.length === 1 &&
+            codecInfos[0].payloadType === FAKE_H264_PAYLOAD_TYPE
+          ),
+        }),
+      ]);
+    });
+  });
+
+  describe('AV1 encoding parameters (resolution mapping)', () => {
```
edvujic: Nit: The other describe blocks in this file use short noun-phrases:

`maxPayloadBitsPerSecond`
`maxMbps`
`trimming of requested receive slots`
`getIngressPayloadTypeCallback`
Could we simplify this to something like `AV1 resolution mapping` or `getAv1EncodingParams` to stay consistent?

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,198 @@ describe('MediaRequestManager', () => {
...
+    ];
+
+    resolutionCases.forEach(({maxFs, expectedRes, levelIdx, maxWidth, maxHeight}) => {
+      it(`maps maxFs=${maxFs} to ${expectedRes} AV1 parameters (levelIdx=${levelIdx}, ${maxWidth}x${maxHeight})`, () => {
+        addReceiverSelectedRequest(100, fakeReceiveSlots[0], maxFs, true);
+
+        assert.calledOnce(sendMediaRequestsCallback);
+        const codecInfos = sendMediaRequestsCallback.getCall(0).args[0][0].codecInfos;
+
+        expect(codecInfos).to.have.length(2);
```
edvujic: Nit: The rest of this file uses assert from `@webex/test-helper-chai` exclusively. Could we stick with assert for the new tests too and drop this import? That keeps a single assertion style across the file.

e.g. `expect(codecInfos).to.have.length(2)` -> `assert.lengthOf(codecInfos, 2)`

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,198 @@ describe('MediaRequestManager', () => {
...
+      {maxFs: MAX_FS_VALUES['720p'], expectedRes: '720p', levelIdx: 5, maxWidth: 1280, maxHeight: 720},
+      {maxFs: MAX_FS_VALUES['1080p'], expectedRes: '1080p', levelIdx: 8, maxWidth: 1920, maxHeight: 1080},
+    ];
+
+    resolutionCases.forEach(({maxFs, expectedRes, levelIdx, maxWidth, maxHeight}) => {
+      it(`maps maxFs=${maxFs} to ${expectedRes} AV1 parameters (levelIdx=${levelIdx}, ${maxWidth}x${maxHeight})`, () => {
+        addReceiverSelectedRequest(100, fakeReceiveSlots[0], maxFs, true);
+
+        assert.calledOnce(sendMediaRequestsCallback);
+        const codecInfos = sendMediaRequestsCallback.getCall(0).args[0][0].codecInfos;
```
edvujic: The rest of the file validates sent requests either via `checkMediaRequestsSent()` or via `assert.calledWith` + `sinon.match`. Reaching into `.getCall(0).args[0][0].codecInfo` is more fragile so if the callback signature changes, these tests break silently in a different way than the others.

We could:

- Use `assert.calledWith(sendMediaRequestsCallback, [sinon.match({ codecInfos: sinon.match(...) })])` like the tests above in `getIngressPayloadTypeCallback`.
- Or extend `checkMediaRequestsSent()` to optionally verify AV1 codec info so all tests share the same validation path

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,198 @@ describe('MediaRequestManager', () => {
...
+
+        expect(codecInfos).to.have.length(2);
+
+        const av1Info = codecInfos[1];
+        expect(av1Info.payloadType).to.equal(FAKE_AV1_PAYLOAD_TYPE);
+        expect(av1Info.av1.levelIdx).to.equal(levelIdx);
+        expect(av1Info.av1.maxWidth).to.equal(maxWidth);
+        expect(av1Info.av1.maxHeight).to.equal(maxHeight);
+
+        sendMediaRequestsCallback.resetHistory();
```
edvujic: Nit: The `beforeEach` at line 1489 already calls `sendMediaRequestsCallback.resetHistory()` before each test. This trailing reset is redundant and none of the other tests in the file do it. Safe to remove.

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,198 @@ describe('MediaRequestManager', () => {
...
+      assert.calledWith(sendMediaRequestsCallback, [
+        sinon.match({
+          codecInfos: sinon.match((codecInfos) =>
+            codecInfos.length === 2 &&
+            codecInfos[0].payloadType === FAKE_H264_PAYLOAD_TYPE &&
+            codecInfos[1].payloadType === FAKE_AV1_PAYLOAD_TYPE
+          ),
+        }),
+      ]);
+    });
```
edvujic: Would this work?

```suggestion
codecInfos: [
  sinon.match({ payloadType: FAKE_H264_PAYLOAD_TYPE }),
  sinon.match({ payloadType: FAKE_AV1_PAYLOAD_TYPE }),
]
```

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
```diff
@@ -0,0 +1,58 @@
+export const AV1_CODEC_PARAMETERS = {
```
edvujic, line 1: Does the `AV1EncodingParams` type from `@webex/internal-media-core` still expect a profile field? If so, this could be a type error or silent omission downstream. Worth confirming against the 2.23.3 type definitions.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -966,6 +965,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      codecMimeType: MediaCodecMimeType
+    ) => {
+      if (this.isMultistream) {
+        try {
+          return this.mediaProperties.webrtcMediaConnection.getIngressPayloadType(
+            mediaType,
+            codecMimeType
+          );
+        } catch (error) {
+          LoggerProxy.logger.error(
```
k-wasniowski: Not sure if we should treat it as error. There could be a browser version which doesnt support AV1, but we still enable a feature toggle for it. So this will get called each time when we send a media request, and we will log an error for every occurence. Not sure if we should treat it as an error for such cases. Would lead more into info/debug.

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
```diff
@@ -0,0 +1,58 @@
...
+    tier: 0,
+    maxWidth: 1920,
+    maxHeight: 1080,
+    maxPicSize: 2_073_600,
+    maxDecodeRate: 70_778_880,
+  },
+};
+
+export const CODEC_DEFAULTS = {
+  h264: {
```
k-wasniowski: Shouldnt we call it H264_CODEC_DEFAULTS, and since it doenst have av1, and flatten the h264 field?

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -298,6 +337,50 @@ export class MediaRequestManager {
...
+            if (av1PayloadType !== undefined) {
+              const av1EncodingParams = this.getAv1EncodingParams(mr);
+              const av1CodecInfo = WcmeCodecInfo.fromAv1(
+                av1PayloadType,
+                new AV1Codec(
+                  av1EncodingParams.levelIdx,
+                  av1EncodingParams.tier,
+                  av1EncodingParams.maxWidth,
+                  av1EncodingParams.maxHeight,
+                  av1EncodingParams.maxPicSize,
```
k-wasniowski, line 375: Shouldnt we use here `maxWidth` and `maxHeight` from the `codecInfo`?

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -54,34 +60,33 @@ export interface MediaRequest {
...
+  mediaType: MediaType,
+  codecMimeType: MediaCodecMimeType
+) => number | undefined;
 type Kind = 'audio' | 'video';
 
 type Options = {
   degradationPreferences: DegradationPreferences;
   kind: Kind;
   trimRequestsToNumOfSources: boolean; // if enabled, AS speaker requests will be trimmed based on the calls to setNumCurrentSources()
+  enableAv1?: boolean;
```
k-wasniowski, line 80: Should it be optional? Or simple bool value with default to false? Optional boolean seems a little bit weird

packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts
```diff
@@ -1361,8 +1366,199 @@ describe('MediaRequestManager', () => {
       assert.throws(() => mediaRequestManager.commit(), 'a mix of active-speaker groups with different values for preferLiveVideo is not supported');
     })
   })
-});
-function assertEqual(arg0: any, arg1: string) {
-  throw new Error('Function not implemented.');
-}
 
+  describe('getIngressPayloadTypeCallback', () => {
+    beforeEach(() => {
```
k-wasniowski, line 1394: Either i'm not seeing, or we dont have it yet. Would be great to have unit tests verifying degradation functionality for AV1 and H264. Would be great to assert there specific values, which we expect. It makes reasoning about the changes simpler

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
SomeBody16: we cannot make a default in type `Options` this would still require providing `enableAv1` everywhere or modyfing type at constructor with `Partial<T>`

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
SomeBody16: It does not at 2.23.3

```typescript
export type AV1EncodingParams = BaseVideoEncodingParams & {
  levelIdx?: number;
  tier?: number;
  maxPicSize?: number;
  maxDecodeRate?: number;
};
```

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
SomeBody16: I've updated it to
```typescript
mr.codecInfo.maxWidth || av1EncodingParams.maxWidth,
mr.codecInfo.maxHeight || av1EncodingParams.maxHeight,
```

Because it makes sense, but from the actual behavior:
maxWidth/ maxHeight are **never** populated

There are only two places where codecInfo objects are constructed — in `remoteMedia.ts` (line 225) and `remoteMediaGroup.ts` (line 236). Both only set codec and maxFs:
```typescript
        codecInfo: maxFs && {
          codec: 'h264',
          maxFs,
        },
```
```typescript
        codecInfo: effectiveMaxFs && {
          codec: 'h264',
          maxFs: effectiveMaxFs,
        },
````

File: packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts (hunk 16 above)
Comment (SomeBody16) [reply packages/@webex/plugin-meetings/test/unit/spec/multistream/mediaRequestManager.ts:1394]: There are tests for:
`re-sends media requests after degradation preferences are set`
`should not degrade max-fs if receive slot sources are not live`
`can degrade max-fs once when request exceeds max macroblocks limit`
`can degrade max-fs multiple times when request exceeds max macroblocks limit`
`can degrade only the largest max-fs when request exceeds max macroblocks limit`
`does trimming first and applies degradationPreferences after that`

___
I've added separate tests for checking if degradation also affect av1

**Comment by k-wasniowski** [comment]
@codex do the code review

File: packages/@webex/plugin-meetings/src/multistream/codec/constants.ts [hunk 17]
```diff
@@ -0,0 +1,56 @@
...
+    maxPicSize: 921_600,
+    maxDecodeRate: 31_950_720,
+  },
+  '1080p': {
+    levelIdx: 8,
+    tier: 0,
+    maxWidth: 1920,
+    maxHeight: 1080,
+    maxPicSize: 2_073_600,
+    maxDecodeRate: 70_778_880,
```
k-wasniowski: maxDecodeRate per AV1 spec for level Idx 8 is 77,856,768.
https://aomediacodec.github.io/av1-spec/#levels

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
```diff
@@ -0,0 +1,56 @@
+export const AV1_CODEC_PARAMETERS = {
+  '90p': {
+    levelIdx: 0,
+    tier: 0,
+    maxWidth: 160,
+    maxHeight: 90,
+    maxPicSize: 15_360,
```
k-wasniowski: maxPicSze should be 14_400 -> 160 × 90.
Current 15_360 is 160x96

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
SomeBody16: I fixed all the values
- for maxPicSize -> calculations of maxWidth * maxHeight so it's more visible what this value represents
- for maxDecodeRate - updated all values because when creating this I copied values from `MaxDisplayRate` by mistake

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
SomeBody16: https://github.com/webex/webex-js-sdk/pull/4816#discussion_r3098802602

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -966,6 +965,30 @@ export default class Meeting extends StatelessWebexPlugin {
       },
       (csi: CSI) => (this.members.findMemberByCsi(csi) as any)?.id
     );
+
+    const getIngressPayloadTypeCallback = (
```
antsukanova: Quick question on the try/catch here: was the empty catch intentional, or would it be worth logging when `getIngressPayloadType` throws? I'm wondering if it could hide bugs later

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -966,6 +965,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      mediaType: MediaType,
```
antsukanova: Curious why you went with an inline callback inside the constructor instead of a small private method on `Meeting` (e.g. private `getIngressPayloadType`(...) and pass it bound). The constructor is already pretty big, and a method would be easier to unit test and document. Was there a reason it had to stay inline?

packages/@webex/plugin-meetings/src/config.ts
```diff
@@ -93,6 +93,7 @@ export default {
     degradationPreferences: {
       maxMacroblocksLimit: 8192,
     },
+    enableAv1SlidesSupport: false,
```
antsukanova, line 96: why we have in some place `enableAv1SlidesSupport` and in some `enableAV1SlidesSupport`?

packages/@webex/plugin-meetings/src/multistream/codec/constants.ts
```diff
@@ -0,0 +1,56 @@
+export const AV1_CODEC_PARAMETERS = {
```
antsukanova: Was leaving `AV1_CODEC_PARAMETERS` untyped intentional here?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -927,6 +927,27 @@ export default class Meetings extends WebexPlugin {
...
+  /**
+   * API to toggle AV1 codec support for video slides in multistream,
+   * needs to be called before webex.meetings.joinWithMedia()
+   *
+   * @param {Boolean} newValue
+   * @private
+   * @memberof Meetings
+   * @returns {undefined}
+   */
+  private _toggleEnableAv1SlidesSupport(newValue: boolean) {
```
antsukanova, line 939: The `JSDoc` says this must be called before `joinWithMedia`(), but nothing enforces it. If a caller toggles after a connection is up, the config silently changes with no effect. Should we log a warn when a multistream connection already exists?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -927,6 +927,27 @@ export default class Meetings extends WebexPlugin {
...
+    if (typeof newValue !== 'boolean') {
```
antsukanova, line 940: The early return on a non-boolean swallows bad input silently. Worth a `LoggerProxy.logger.warn` so misuse is visible?

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -927,6 +927,27 @@ export default class Meetings extends WebexPlugin {
...
+      return;
+    }
+
+    // @ts-ignore
```
antsukanova, line 944: why do we need here @ts-ignore?

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: Not sure to which part you referring to
In `getIngressPayloadTypeCallback` there is catch block with error logging
```typescript
} catch (error) {
    LoggerProxy.logger.info(
        `Meeting:index#mediaRequestManager --> failed to get ingress payload type for mediaType=${mediaType}, codecMimeType=${codecMimeType}`,
        error
    );

    return undefined;
}
```

packages/@webex/plugin-meetings/src/meetings/index.ts
SomeBody16: The only use of this method is in web-client when enabling/ disabling feature. Otherwise this method is completely private and internal.

Currently it follows all other feature toggle flow so I don't think that it should have any checks and warnings

packages/@webex/plugin-meetings/src/meetings/index.ts
SomeBody16: https://github.com/webex/webex-js-sdk/pull/4816/changes#r3115926846

packages/@webex/plugin-meetings/src/meetings/index.ts
SomeBody16: current `this` is `Meetings` which doesn't have config property so we need `@ts-ignore`

packages/@webex/plugin-meetings/src/config.ts
SomeBody16: in js-sdk we use `Av1` for consistency. But during development of wcme and media-core, this property was named `AV1`

So in js-sdk we `Av1`
But when passing arguments to media-core/ wcme we map it to `AV1` 

This should be updated during refactor and av1 for video-main
I've created task for it:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-802843

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: Updated to be private function

packages/@webex/plugin-meetings/src/meeting/index.ts
antsukanova: thanks, all good, I had earlier version without this logger locally

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: Got you, thanks! but why we are interetsed to have it called only after `joinWithMedia`?

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: I see. Just out of curiosity and understanding when this prop will be there, could you clarify?

packages/@webex/plugin-meetings/src/meetings/index.ts
edvujic: Small clarification: the doc says before `joinWithMedia()`, not after.

We read `enableAv1SlidesSupport` when we create the multistream setup (for example when we build the slide `MediaRequestManager`) and when we create the media connection. Those paths copy the value once, they don’t watch the config forever. So changing the flag after join doesn’t change the running call; it only changes what we’d use on a later join. That’s why we document “set it before join.”

packages/@webex/plugin-meetings/src/meetings/index.ts
edvujic: `config` isn’t missing at runtime for normal flows. `Meetings` extends `WebexPlugin`, which is implemented in JS and adds config as a getter. That getter returns `webex.config.meetings` (see `webex-plugin.js`). Our TS class never declares that property, so TypeScript complains unless we use @ts-ignore or add a type. The prop is there whenever the client has been set up with the usual meetings config.

edvujic (approved):

k-wasniowski (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: Got you, all clear here, thank you

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: Yeah, it seems that it will require large refactor touching many files across the plugin, do not feel we need to do this here, thanks for clarification

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
antsukanova, line 94: I would suggest improving the type here to restrict to pass `enableAv1` to the audio type, like this:
```
new MediaRequestManager(..., { kind: 'audio', enableAv1: true }) 
```

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -192,12 +204,39 @@ export class MediaRequestManager {
...
+   * Returns the AV1 encoding parameters for a media request
+   * @param mediaRequest - The media request to get the AV1 encoding parameters for
+   * @returns {AV1EncodingParams} The AV1 encoding parameters
+   */
+  // eslint-disable-next-line class-methods-use-this
+  private getAv1EncodingParams(mediaRequest: MediaRequest): AV1EncodingParams {
+    const frameSize = mediaRequest.codecInfo.maxFs || H264_CODEC_PARAMETERS.maxFs;
+    let resolution: SupportedResolution;
+
+    if (frameSize <= MAX_FS_VALUES['90p']) {
```
antsukanova: Let's make it cleaner here with array like:

```
const RESOLUTION_BUCKETS: Array<[SupportedResolution, number]> = [
  ['90p',  MAX_FS_VALUES['90p']],
  ['180p', MAX_FS_VALUES['180p']],
  ['360p', MAX_FS_VALUES['360p']],
  ['540p', MAX_FS_VALUES['540p']],
  ['720p', MAX_FS_VALUES['720p']],
];

private getAv1EncodingParams(mediaRequest: MediaRequest): AV1EncodingParams {
  const frameSize = mediaRequest.codecInfo.maxFs || H264_CODEC_PARAMETERS.maxFs;
  const resolution = RESOLUTION_BUCKETS.find(([, maxFs]) => frameSize <= maxFs)?.[0] ?? '1080p';

  return AV1_CODEC_PARAMETERS[resolution];
}
```

packages/@webex/plugin-meetings/src/multistream/mediaRequestManager.ts
```diff
@@ -298,6 +337,50 @@ export class MediaRequestManager {
...
+              h264PayloadType,
+              new H264Codec(
+                mr.codecInfo.maxFs,
+                mr.codecInfo.maxFps || H264_CODEC_PARAMETERS.maxFps,
+                this.getH264MaxMbps(mr),
+                mr.codecInfo.maxWidth,
+                mr.codecInfo.maxHeight
+              )
+            );
+            codecInfos.push(h264CodecInfo);
```
antsukanova: WDYT to have reformatted this part as next:

```
if (mr.codecInfo) {
  const codecInfos = [
    this.buildH264CodecInfo(mr),
    this.buildAv1CodecInfo(mr),
  ].filter((info): info is WcmeCodecInfo => info !== undefined);
}
```

and have two private builder methods that return undefined on failure? Just for readability, it feels better.

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
```diff
@@ -505,6 +512,7 @@ describe('createMediaConnection', () => {
       {
         iceServers: [],
         disableAudioTwcc: true,
+        enableAV1SlidesSupport: false,
```
antsukanova, line 535: As well, maybe you have already covered that, but not sure about the plan with naming inconsistency `enableAv1SlidesSupport` and `enableAV1SlidesSupport`, will it be aligned?

As well, just for the future, do you feel that such an object will be better?
```
  av1Support?: {
    slides: boolean;
    mainVideo: boolean;
  };
```

You won't scatter av1 props across configs separately.

antsukanova (commented):
From the tests I see that:
`media/index.ts` — only `enableAV1SlidesSupport: false` is asserted
`mediaRequestManager.ts` — AV1 width/height override not tested
would as well suggest to check 
`webrtcMediaConnection.getIngressPayloadType`. It is never stubbed in `meeting/index.js` tests but it contains important code flow worth checking

packages/@webex/plugin-meetings/test/unit/spec/media/index.ts
edvujic: What the tests in `test/unit/spec/media/index.ts` do: they assert the object passed into `MultistreamRoapMediaConnection`, which comes from `internal-media-core` and uses the property name `enableAV1SlidesSupport` (acronym spelled AV1). The public option on `Media.createMediaConnection` is `enableAv1SlidesSupport`.

For the 2nd part YAGNI - the product path is slides-only AV1, we should consider this for the feature. Though I can see the benefits of one namespace for AV1, easier to extend.

Shreyas281299:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Shreyas281299 (approved):

antsukanova (approved):

szotrabh (approved):

PR #4847, opened 2026-04-10
feat(calling): add utils to conditionally make socket or http request

rarajes2:
# COMPLETES #<[CAI-7727](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7727) >

Adds utils to handle conditional request to mobius based on socket connection.

Added the required utils to handle the seamless functionality without altering too much of existing code.

**Note**: This PR does not add UTs, logs and metrics to the code. These things will be handled after integration testing with actual mobius server

**For reference:**
Screenshot 2026-04-10 at 2 26 53 PM

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Value for `isMobiusSocketEnabled` is false by default so expectation in this PR is that all the existing scenarios should work as it used to be. Following scenarios were tested

- Registration and Deregistration
- Call flows: establishing and ending call, hold, resume, consult and blind transfer

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/CallingClient.ts
+++ b/packages/calling/src/CallingClient/CallingClient.ts
@@ -63,6 +63,7 @@ import {
 } from '../Metrics/types';
 import {getMetricManager} from '../Metrics';
 import windowsChromiumIceWarmup from './windowsChromiumIceWarmupUtils';
+import {APIRequest} from './utils/request';
 
 /**
  * The `CallingClient` module provides a set of APIs for line registration and calling functionalities within the SDK.
@@ -99,6 +100,8 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
   public mediaEngine: typeof Media;
 
   private lineDict: Record<string, ILine> = {};
+
+  private apiRequest: APIRequest;
 
   private isNetworkDown = false;
 
@@ -156,6 +159,8 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
     this.backupMobiusUris = [];
     this.mobiusClusters = this.webex.internal.services.getMobiusClusters();
     this.mobiusHost = '';
+
+    this.apiRequest = APIRequest.getInstance({webex: this.webex});
 
     this.registerSessionsListener();
 
@@ -744,8 +749,9 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
         userid
       )}`;
       try {
+        // TODO: Add Mobius Socket support
         // eslint-disable-next-line no-await-in-loop
-        const response = <WebexRequestPayload>await this.webex.request({
+        const response = (await this.apiRequest.makeRequest({
           uri,
           method: HTTP_METHODS.GET,
           service: ALLOWED_SERVICES.MOBIUS,
@@ -753,7 +759,7 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
             [CISCO_DEVICE_URL]: this.webex.internal.device.url,
             [SPARK_USER_AGENT]: CALLING_USER_AGENT,
           },
-        });
+        })) as WebexRequestPayload;
 
         if (response.statusCode !== 200) {
           throw new Error(`API call failed with ${response.statusCode}`);
--- a/packages/calling/src/CallingClient/utils/constants.ts
+++ b/packages/calling/src/CallingClient/utils/constants.ts
@@ -1,35 +1,8 @@
 /**
- * Mobius WebSocket namespace prefixes for channel topic paths
- */
-export enum MOBIUS_SOCKET_NAMESPACE {
-  REGISTRATION = 'registration',
-  CALLS = 'calls',
-}
-
-/**
- * Mobius WebSocket channel topic paths
- * Each channel maps to a request/response event pair
- */
-export enum MOBIUS_SOCKET_CHANNEL {
-  REGISTER = 'registration/register',
-  UNREGISTER = 'registration/unregister',
-  DEVICE_STATUS = 'registration/device-status',
-  DEVICE_GET = 'registration/device-get',
-  DEVICE_LIST = 'registration/device-list',
-  CALL_SETUP = 'calls/setup',
-  CALL_STATE = 'calls/state',
-  CALL_STATUS = 'calls/status',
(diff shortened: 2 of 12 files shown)


packages/calling/src/CallingClient/utils/constants.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Mobius WebSocket namespace prefixes for channel topic paths
+ */
+export enum MOBIUS_SOCKET_NAMESPACE {
```
rarajes2: This will be removed. Not required

packages/calling/src/CallingClient/utils/constants.ts
```diff
@@ -0,0 +1,68 @@
...
+  REGISTRATION = 'registration',
+  CALLS = 'calls',
+}
+
+/**
+ * Mobius WebSocket channel topic paths
+ * Each channel maps to a request/response event pair
+ */
+export enum MOBIUS_SOCKET_CHANNEL {
```
rarajes2: This will be removed. Not required

packages/calling/src/CallingClient/utils/request.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * APIRequest class provides a unified interface for making requests
+ * that can be routed through either HTTP (webex.request) or WebSocket
+ * (mobiusSocketRequest) based on configuration.
+ */
+export class APIRequest {
+  // eslint-disable-next-line no-use-before-define
+  private static instance: APIRequest | undefined;
+  private isMobiusSocketEnabled: boolean;
```
rsarika, line 21: create a method to set this so that it can be used to handle fallback scenarios may be

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -744,16 +749,17 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+        const response = (await this.apiRequest.makeRequest({
           uri,
           method: HTTP_METHODS.GET,
           service: ALLOWED_SERVICES.MOBIUS,
           headers: {
             [CISCO_DEVICE_URL]: this.webex.internal.device.url,
             [SPARK_USER_AGENT]: CALLING_USER_AGENT,
           },
-        });
+        })) as WebexRequestPayload;
```
rsarika, line 762: is this type cast required?  is there any better pattern?

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -100,6 +101,8 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
 
   private lineDict: Record<string, ILine> = {};
 
+  private apiRequest: APIRequest;
```
rsarika, line 104: if its singleton why do we need to keep a copy as variable in all the plcaes/classes?

  is there a better pattern directly calling it?

rsarika (approved):

rsarika:
fix unit tests also

packages/calling/src/CallingClient/utils/constants.ts
rarajes2: Done

packages/calling/src/CallingClient/utils/constants.ts
rarajes2: Done

packages/calling/src/CallingClient/utils/request.ts
rarajes2: This is the flag to handle the fallback scenario where if it's false sdk will use http/mercury approach. We will see in later PR if method to set this value is required

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: Yes this required for now. Earlier it was passed as type arg to webex.request method. This will be improved in later PR

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: Since this is singleton, this is how it is expected to work. Can't think of a better approach as of now, will see if we can find something better.

rarajes2:
### Changes to fix the 4 failing tests:

**1. `packages/calling/src/CallingClient/utils/request.ts`**
- Added `static resetInstance()` method to `APIRequest` class for proper test cleanup of the singleton between test suites.

**2. `packages/calling/src/CallingClient/registration/register.test.ts`**
- Added `import {APIRequest} from '../utils/request'`
- Added `APIRequest.resetInstance()` in `beforeEach` to ensure a fresh singleton per test
- Updated the "verify failure registration 403-101" test:
  - Added a third mock for `webex.request` (for the DELETE call that now goes through `apiRequest.makeRequest` instead of `fetch`)
  - Changed `expect(webex.request).toBeCalledTimes(2)` to `toBeCalledTimes(3)`
  - Replaced the `global.fetch` assertion with an assertion that `webex.request` was called with the DELETE parameters
  - Removed the now-unnecessary `global.fetch` mock

**3. `packages/calling/src/CallingClient/calling/call.test.ts`**
- Added `import {APIRequest} from '../utils/request'`
- Added `APIRequest.resetInstance()` in all 3 `beforeEach` blocks (`Call Tests`, `State Machine handler tests`, `Supplementary Services tests`)
- Added extra `await Promise.resolve()` calls in the 3 failing tests to account for the additional microtask level introduced by the `async makeRequest` wrapper:
  - "state changes during unsuccessful outgoing call due to error in call setup"
  - "state changes during unsuccessful outgoing call due to error in media ok"
  - "incoming call: failing ROAP_ANSWER posts error path and tears down"

rarajes2:
Merging the PR. Any new PR comments will be addressed in a separate PR.

mkesavan13 (commented):
* The most nagging thing is the return statements that comes with an `as {payloadType}`
* Second, the discovery of socket message based on the URL. Today it can be absolute but could become a pain when we want to remove the HTTP route all-together

Was there any other way to avoid it?

Rest of the code looks good to me.

PR #4853, opened 2026-04-13
fix: display two items for one webinar in OBTP list

WeijuanShao:
# COMPLETES #

1.For hash tree, no controls return in loci array. Using the info.isBreakout to determine it's bo or not.
2. getReplaceInfoFromSelf's change - the account may start webinar and join bo in device1, but same account in device2 can't find the replace info is just use the current device url to find. So just change to use self.deviceUrl to find the replace info.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -103,7 +103,7 @@ export type HashTreeParserEntry = {
  */
 function getReplaceInfoFromSelf(self: any, deviceUrl: string): ReplacesInfo | undefined {
   if (self) {
-    const device = MeetingsUtil.getThisDevice({self}, deviceUrl);
+    const device = MeetingsUtil.getThisDevice({self}, self?.deviceUrl);
```
marcin-bazyl: `deviceUrl` argument is not used now so can be removed, also from findMeetingForHashTreeMessage()

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1950,9 +1951,18 @@ export default class Meetings extends WebexPlugin {
     this.breakoutLocusForHandleLater = [];
     const lociToUpdate = [...mainLoci];
     breakoutLoci.forEach((breakoutLocus) => {
-      const associateMainLocus = mainLoci.find(
-        (mainLocus) => mainLocus.controls?.breakout?.url === breakoutLocus.controls?.breakout?.url
-      );
+      const associateMainLocus = mainLoci.find((mainLocus) => {
+        if (mainLocus.controls?.breakout?.url === breakoutLocus.controls?.breakout?.url) {
```
marcin-bazyl: if both sides are undefined, they will be equal and we return true, which is probably not want we want

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1950,9 +1951,18 @@ export default class Meetings extends WebexPlugin {
     this.breakoutLocusForHandleLater = [];
     const lociToUpdate = [...mainLoci];
     breakoutLoci.forEach((breakoutLocus) => {
-      const associateMainLocus = mainLoci.find(
-        (mainLocus) => mainLocus.controls?.breakout?.url === breakoutLocus.controls?.breakout?.url
-      );
+      const associateMainLocus = mainLoci.find((mainLocus) => {
```
marcin-bazyl: there is a race condition so sometimes the fix from this PR doesn't work - it happens when checkHandleBreakoutLocus() is called before the HashTreeParser instance is created, I'll send you more details 1:1

packages/@webex/plugin-meetings/src/locus-info/index.ts
WeijuanShao: I'd remove the ```deviceUrl``` argument from ```findMeetingForHashTreeMessage```, but ```getReplaceInfoFromSelf``` also called by ```handleHashTreeParserSwitch``` which I think should use this device url. So I just keep the parameter for ```getReplaceInfoFromSelf```

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1954,9 +1950,21 @@ export default class Meetings extends WebexPlugin {
...
+        if (
```
marcin-bazyl: this block of lines 1954-1966 is the same as 2005-2018, maybe it would be worth extracting it into a util function?

packages/@webex/plugin-meetings/src/locus-info/index.ts
marcin-bazyl: ok, but now the parameter in getReplaceInfoFromSelf() has been made optional, but in all the places where getReplaceInfoFromSelf() is called we actually pass it, so there is no need for it to be optional and no need for the ` || self?.deviceUrl` fallback, or if you want to keep it optional and with that fallback, then we don't need to pass `self?.deviceUrl` when calling it from findMeetingForHashTreeMessage

packages/@webex/plugin-meetings/src/meetings/index.ts
WeijuanShao: OK, updated

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -155,7 +157,7 @@ export function findMeetingForHashTreeMessage(
   // if we haven't found anything, it may mean that message has a new locusUrl
   // check if it indicates that it replaces some existing current locusUrl (this is indicated in "self")
   const self = message.locusStateElements?.find((el) => isSelf(el))?.data;
-  const replaces = getReplaceInfoFromSelf(self, deviceUrl);
+  const replaces = getReplaceInfoFromSelf(self, self?.deviceUrl);
```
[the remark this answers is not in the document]
WeijuanShao: If ```self.deviceUrl``` is not available, then replaceInfo shouldn’t be found. Using the current device URL as a fallback might return a result, but it wouldn’t be correct.

packages/@webex/plugin-meetings/src/locus-info/index.ts
WeijuanShao: OK, I also think removing the optional handling and ``` || self?.deviceUrl``` makes this method cleaner

packages/@webex/plugin-meetings/src/locus-info/index.ts
marcin-bazyl: deviceUrl argument is no longer there, so we should remove this comment

marcin-bazyl (approved):

mickelr (approved):

PR #4859, opened 2026-04-14
feat(calling): add mobius socket connection logic

rarajes2:
# COMPLETES #<Partially [CAI-7823](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7823)>

This adds logic to connect to the mobius wss using the `internal-plugin-mobius-socket`

#### 1. `src/common/types.ts` (+1 line)
- Added optional `wss?: string[]` field to the `ServerInfo` type to model the new WSS endpoints returned by Mobius discovery.

#### 2. `src/CallingClient/types.ts` (+1 line)
- Added optional `isMobiusSocketEnabled?: boolean` to `CallingClientConfig`, providing a feature flag (default `false`) to gate the entire socket feature.

#### 3. `src/common/Utils.ts` (+35 lines, −3 lines)
- Extended `filterMobiusUris()` to also extract and deduplicate WSS URIs from the `primary.wss` and `backup.wss` arrays in the Mobius discovery response.
- Return type expanded from `{primary, backup}` to `{primary, backup, primaryWss, backupWss}` — fully backward-compatible.

#### 4. `src/CallingClient/constants.ts` (+1 line)
- Added `CONNECT_TO_MOBIUS_SOCKET: 'connectToMobiusSocket'` to the `METHODS` constant object for logger context.

#### 5. `src/CallingClient/CallingClient.ts` (+111 lines)
- **Import**: Added `require('@webex/internal-plugin-mobius-socket')` for `createMobiusSocket`.
- **New fields**: `primaryWssMobiusUris`, `backupWssMobiusUris`, `isMobiusSocketEnabled`, `mobiusSocket`.
- **Constructor**: Initializes WSS URI arrays, reads the feature flag from config, and conditionally creates a `MobiusSocket` instance via `createMobiusSocket(webex)`.
- **`getMobiusServers()`**: Now stores `primaryWss` and `backupWss` from the updated `filterMobiusUris` return.
- **`init()`**: Calls `connectToMobiusSocket()` after `getMobiusServers()` when `isMobiusSocketEnabled` is `true`.
- **`connectToMobiusSocket()` (new private method)**: Tries each primary WSS URI sequentially; if all fail, tries each backup WSS URI. Logs clearly distinguish primary vs backup attempts, successes, and failures. If all URIs are exhausted, logs a warning and continues without a socket connection.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/CallingClient.ts
+++ b/packages/calling/src/CallingClient/CallingClient.ts
@@ -65,6 +65,9 @@ import {getMetricManager} from '../Metrics';
 import windowsChromiumIceWarmup from './windowsChromiumIceWarmupUtils';
 import {APIRequest} from './utils/request';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires, global-require
+const {createMobiusSocket} = require('@webex/internal-plugin-mobius-socket');
+
 /**
  * The `CallingClient` module provides a set of APIs for line registration and calling functionalities within the SDK.
  *
@@ -102,6 +105,15 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
   private lineDict: Record<string, ILine> = {};
 
   private apiRequest: APIRequest;
+
+  private primaryWssMobiusUris: string[];
+
+  private backupWssMobiusUris: string[];
+
+  private isMobiusSocketEnabled: boolean;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private mobiusSocket: any;
 
   private isNetworkDown = false;
 
@@ -157,8 +169,17 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
 
     this.primaryMobiusUris = [];
     this.backupMobiusUris = [];
+    this.primaryWssMobiusUris = [];
+    this.backupWssMobiusUris = [];
     this.mobiusClusters = this.webex.internal.services.getMobiusClusters();
     this.mobiusHost = '';
+
+    // MOBIUS SOCKET TODO: Fix this when feature flag is implemented
+    this.isMobiusSocketEnabled = config?.isMobiusSocketEnabled ?? false;
+
+    if (this.isMobiusSocketEnabled) {
+      this.mobiusSocket = createMobiusSocket(this.webex);
+    }
 
     this.apiRequest = APIRequest.getInstance({webex: this.webex});
 
@@ -207,6 +228,11 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
     }
 
     await this.getMobiusServers();
+
+    if (this.isMobiusSocketEnabled) {
+      await this.connectToMobiusSocket();
+    }
+
     await this.createLine();
 
     this.setupNetworkEventListeners();
@@ -543,6 +569,8 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
           const mobiusUris = filterMobiusUris(mobiusServers, this.mobiusHost);
           this.primaryMobiusUris = mobiusUris.primary;
           this.backupMobiusUris = mobiusUris.backup;
+          this.primaryWssMobiusUris = mobiusUris.primaryWss;
+          this.backupWssMobiusUris = mobiusUris.backupWss;
 
           log.log(
             `Final list of Mobius Servers, primary: ${mobiusUris.primary} and backup: ${mobiusUris.backup}`,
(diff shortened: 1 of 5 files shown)


packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -65,6 +65,9 @@ import {getMetricManager} from '../Metrics';
 import windowsChromiumIceWarmup from './windowsChromiumIceWarmupUtils';
 import {APIRequest} from './utils/request';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires, global-require
+const {createMobiusSocket} = require('@webex/internal-plugin-mobius-socket');
```
rarajes2: This will be changed to import syntax

packages/calling/src/common/Utils.ts
```diff
@@ -187,7 +205,21 @@ export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl:
     }
   }
 
-  return {primary: primaryUris, backup: backupUris};
+  /* Remove duplicates from primary wss by keeping the order intact */
+  for (let i = 0; i < wssArrayPrimary.length; i += 1) {
+    if (primaryWss.indexOf(wssArrayPrimary[i]) === -1) {
+      primaryWss.push(wssArrayPrimary[i]);
+    }
+  }
```
rarajes2, line 213: Check if this is required.

packages/calling/src/common/Utils.ts
```diff
@@ -152,13 +154,27 @@ export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl:
     }
   }
 
+  if (mobiusServers?.primary?.wss) {
+    log.info('Adding Primary wss uris', logContext);
+    for (const wssUri of mobiusServers.primary.wss) {
+      wssArrayPrimary.push(wssUri);
```
rarajes2, line 160: See if we need to run this loop. This was required in the http uri to append the URL_ENDPOINT

rsarika (approved):

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -207,6 +228,11 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
     }
 
     await this.getMobiusServers();
+
+    if (this.isMobiusSocketEnabled) {
```
Kesari3008, line 231: Above code and this code can be combined. We can create the socket and connect at once

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -608,6 +636,89 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+   */
+  private async connectToMobiusSocket(): Promise<void> {
+    const loggerContext = {
+      file: CALLING_CLIENT_FILE,
+      method: METHODS.CONNECT_TO_MOBIUS_SOCKET,
+    };
+
+    log.info(METHOD_START_MESSAGE, loggerContext);
+
+    if (!this.primaryWssMobiusUris.length && !this.backupWssMobiusUris.length) {
```
Kesari3008, line 651: This check is something should be performed before even create socket object, If we don't have the URL itelf then creating the object would be of no use. By the time we reach the connect state, there should be 2 handling patch here success or failure and on failure we should throw error

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -608,6 +636,89 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+            loggerContext
+          );
+
+          return;
+        } catch (err: unknown) {
+          log.warn(`Primary WSS URI connection failed for ${wssUri}: ${err}`, loggerContext);
+        }
+      }
+
+      log.warn('All primary WSS URI connection attempts failed', loggerContext);
```
Kesari3008, line 682: These should not be warnings, we should throw error here for failures where nothing more can be done because after this registration n all can not be attempted anyway so createLine n all should be blocked

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -608,6 +636,89 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      log.log(
+        `Attempting Mobius socket connection using primary WSS URIs (${this.primaryWssMobiusUris.length} available)`,
+        loggerContext
+      );
+
+      for (const wssUri of this.primaryWssMobiusUris) {
+        try {
+          log.log(`Trying primary WSS URI: ${wssUri}`, loggerContext);
+          // eslint-disable-next-line no-await-in-loop
+          await this.mobiusSocket.connect(wssUri);
```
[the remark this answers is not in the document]
rarajes2: @rsarika This is what we were talking about. We have to provide some default value

packages/calling/src/CallingClient/CallingClient.ts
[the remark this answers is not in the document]
rarajes2: Fixed

rarajes2: Done

packages/calling/src/common/Utils.ts
rarajes2: Keeping for now, will update when confirmed the wss url in the region discovery response

packages/calling/src/common/Utils.ts
rarajes2: Keeping for now, will update when confirmed the wss url in the region discovery response

Kesari3008, line 160: We may not need to maintain 2 different arrays in our SDK because from what I saw on the sample catalog payload shared, Mobius host is same. Only difference was the additional endpoint parameter being added to uri to do different actions. Maintaing single array for primary and one for backup should be enough and the entries sould just be updated as it is of how we receive it

Kesari3008 (changes requested):

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: We should keep the object creation at the constructor level only. The connection part may move to registration logic later based on further development. Not changing this logic as of now, will be updated in upcoming PR if required

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: This does not make any difference to the functional behaviour. If it breaks anything, will update the code. Upcoming PRs will anyway go through a few refactor, this will be updated if required

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: There is a TODO added for this in the code already

packages/calling/src/common/Utils.ts
rarajes2: Created separate variables because the connecting and registration flows will be different. This makes the code readable and does not add much complexity to the existing code or consume too much memory.

rarajes2:
Merging the PR. Further comments will be addressed in upcoming PRs

Issue #4588, opened 2025-11-24
Unhandled promise rejection somewhere in people.get


brettdh:
**Describe the bug**
When an SDK method (such as `webex.people.get(personId)`) fails with an error, I expect the promise returned by the method to reject, which it does. However, some other promise inside the SDK is also rejecting, as shown in the browser console log and also in the dev server's built-in error handling, which detects the unhandled promise rejection and displays the error (in this case, just the person ID with no other context) in the full-screen overlay.

The reason I know that this is an issue inside the SDK and not in my code is that if I replace this call in my code:
```
await webex.people.get(personId)
```
with this:
```
throw new Error('whoopsie')
```
then the unhandled rejections go away. This indicates that the error handling in my app is working correctly and that the problem is in the SDK, perhaps with a [broken chain of promises](https://martindzejky.medium.com/why-your-promise-still-throws-unhandled-rejection-even-with-a-catch-6b37ddc819a7) somewhere.

**To Reproduce**

1. Create a browser app that calls `webex.people.get(personId)` with a `personId that doesn't exist
2. Run the app in a browser
3. Observe the unhandled promise rejection in the browser console

**Expected behavior**
No unhandled rejections from within the SDK

**Platform (please complete the following information):**
 - OS: macOS Tahoe 26.1
 - Browser: Chrome
 - Version: 142.0.7444.176
 - Device Type: Desktop

sandiban:
Hi @brettdh,
Could you kindly share us the complete browser HAR logs for this item.
Also wanted to understand before that, were you able to reproduce the issue over the JS SDK sample - https://web-sdk.webex.com/samples/browser-plugin-meetings/, if not could you kindly try once from the sample App please and if the unhandled exception still comes, kindly share the corresponding HAR logs for it.
And one more thing, are you using for personal Bearer Access token or, JWT token for authorization? Kindly confirm please. Thanks!

brettdh:
As far as I can tell, there's no sample app at https://web-sdk.webex.com/samples that demonstrates broad use of the messaging SDK - specifically, the call to `webex.people.get(personId)` that I mentioned above.

I am using the bearer access token.

I cannot share a HAR or the source code for this app, but here is a minimal repro using `webex-node` (though I reported this bug with the browser SDK, the result is the same in NodeJS):

```javascript
import Webex from 'webex-node';

const webex = Webex.init({
  credentials: {
    access_token: process.env.WEBEX_ACCESS_TOKEN
  },
[...]

PR #4873, opened 2026-04-17
fix(webex-core): handle broken chain of promises

mkesavan13:
# COMPLETES [SPARK-747471](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-747471)

Issue #4588

Basically, the SDK was internally having an unhandled promise rejection that affected the flow of external apps.

* Cleaner promise chain handling on Batcher Requests
* Ensuring to reject defers in case of handleHttpError where the request doesn't have a body. Right now the promise is rejected but the defer itself isn't. With this change, that case is handled
* Not propagating caught rejects so that it doesn't again land in unhandledRejection

Restricted Cisco Only Vidcast for the fix: https://app.vidcast.io/share/69133bb4-84c1-4b08-a4cf-b0db6f05a233

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Ran this snippet in Meetings Sample to not see the `unhandledRejection` handler being called in the following cases:
1. API succeeded but the personId was incorrect
2. API itself failed (Let's say the token is wrong)
```
// 1) Watch global unhandled rejections
window.addEventListener('unhandledrejection', (event) => {
  console.error('[UNHANDLED]', event.reason);
});

// 2) Trigger people.get() with guaranteed-invalid personId
(async () => {
  const uuid = crypto.randomUUID();
  const personId = "some_random_id";

  console.log('personId:', personId);
  try {
    await window.webex.people.get(personId);
  } catch (e) {
    console.error('[CAUGHT]', e);
  }
})();
```

  - [ ] Github Copilot
  - [x] Other - Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/webex-core/src/lib/batcher.js
```diff
@@ -59,13 +59,12 @@ const Batcher = WebexPlugin.extend({
         this.deferreds.set(idx, defer);
         this.prepareItem(item)
           .then((req) => {
-            defer.promise = defer.promise
-              .then(tap(() => this.deferreds.delete(idx)))
```
mkesavan13, line 63: Here's the definition of tap:
```
function tap(fn) {
  return function (r) {
    return new _promise.default(function (resolve) {
      resolve(fn(r));
    }).then(function () {
      return r;
    }).catch(function () {
      return r;
    });
  };
}
```
It does nothing but return a curried function that returns a promise which executes and tries to resolve passed `fn` which in this case is `() => this.deferreds.delete(idx)`

If that's resolved, resolves with `idx` as value or when rejected, rejects with `idx` as value again.

Per the old flow, this was required to catch again to do the deletion. 

Now, we have a cleaner code with promise.then receiving methods to resolve/reject directly

packages/@webex/webex-core/src/lib/batcher.js
```diff
@@ -116,7 +115,19 @@ const Batcher = WebexPlugin.extend({
...
+              // queued request to ensure caller promises settle.
+              return this.handleHttpError(reason).catch(() =>
+                Promise.all(
+                  queue.map((item) =>
+                    this.getDeferredForRequest(item).then((defer) => {
+                      defer.reject(reason);
+                    })
+                  )
+                )
+              );
```
mkesavan13, line 130: Although `handleHttpError` is part of batcher and could have made this change directly there, the `handleHttpError` is overridden by downstream classes like avatar or metrics plugin. Therefore, this fix would get lost for such plugins.

If we add it in the `executeQueue`'s catch block, the fix is propagated to all the downstream classes as executeQueue is nowhere overridden.

mkesavan13 (commented):
Leaving notes of my findings with Cursor as co-pilot.

akulakum (approved):
This looks like a solid fix for the unhandledRejection / unsettled-queue behavior described in the issue and demo: cleaning up the deferred promise handling, rejecting queued work when handleHttpError fails (including GET-style cases without a body), and avoiding a re-thrown rejection on a debounced path that nothing awaits. The unit tests for unhandledRejection and for failing queued deferreds are a good addition.

Before merge, could we confirm two things for maintainers and downstream consumers? First, with the final executeQueue catch now swallowing the rejection after logging, are we certain that every path that reaches that catch has already failed the relevant deferred promises, so we do not risk leaving a caller hanging? Second, for batcher subclasses that override handleHttpError, does the new fallback in executeQueue still guarantee that the full queue is rejected when appropriate?

Coread (approved):

mkesavan13:
> Before merge, could we confirm two things for maintainers and downstream consumers? First, with the final executeQueue catch now swallowing the rejection after logging, are we certain that every path that reaches that catch has already failed the relevant deferred promises, so we do not risk leaving a caller hanging?

Yes. Every queued deferred is rejected in the inner promises

> Second, for batcher subclasses that override handleHttpError, does the new fallback in executeQueue still guarantee that the full queue is rejected when appropriate?

Yes. The overrides properly reject so that the new fallback in executeQueue is reached in any case.

Thanks for the questions to double check, @akulakum. Will merge the PR when the CI passes with recent updates from next branch.

Also, thanks for the approval @Coread

PR #4882, opened 2026-04-20
fix: no meeting title on OBTP when the account start webinar on another device

WeijuanShao:
# COMPLETES #<https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-799460>

For this issue, we receive a ```state_message``` which only have self to create meeting, during the meeting creation will ```fetchMeetingInfo``` base on the ```info?.webexSite``` but missing, so will caused the issue. To fix it, defer to ```fetchMeetingInfo``` after ```initialSetup``` is done

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4659,6 +4659,34 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   * @param {Boolean} needFetchMeetingInfo
+   * @returns {Promise<void>}
+   */
+  public async finalizeMeetingAfterInitialLocusSetup(needFetchMeetingInfo: boolean): Promise<void> {
+    const syncedLocus = this?.locusInfo?.parsedLocus;
+
+    if (syncedLocus && this?.destinationType === DESTINATION_TYPE.LOCUS_ID) {
+      // destination is initialized from the initial locus snapshot in constructor,
+      // so refresh it after locus sync to avoid stale partial hash-tree data.
+      this.destination = syncedLocus;
```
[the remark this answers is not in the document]
marcin-bazyl: yeah I agree with AI, normally this.destination is set to the raw Locus DTO we get from Locus and here we're setting it to point to a parsed one, so they will be different and can cause some issues

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -594,13 +596,19 @@ export default class Meetings extends WebexPlugin {
               locus: data.locus,
               hashTreeMessage: data.stateElementsMessage,
             });
+            if (data.eventType !== LOCUSEVENT.SDK_LOCUS_FROM_SYNC_MEETINGS) {
```
marcin-bazyl: maybe it would be cleaner if we just always call `finalizeMeetingAfterInitialLocusSetup()` without any condition here and maybe even without `shouldFetchMeetingInfoAfterInitialSetup` argument, so that all the logic of deciding what needs to be done is encapsulated in one place - inside `finalizeMeetingAfterInitialLocusSetup()`? and does the condition for fetching info really need to depend on the data.eventType? maybe we should just always fetch meeting info if we don't have it at this point (but we would have to check not just the info from Locus but also if we maybe already have it from appapi)

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
WeijuanShao: Yes, I’ve considered this as well. However, if we want to use the raw locus, we’d need to cache a copy of it in locus-info, or pass a callback into initSetUp to construct a full locus once all locus data is available and then assign destination. I’ll try the callback approach with initSetUp and see how it goes.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4659,6 +4659,33 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      // so refresh it after locus sync to avoid stale partial hash-tree data.
+      this.destination = locus;
+    }
+    if (
+      (!this.meetingInfo || isEmpty(this.meetingInfo)) &&
+      (this.destination as LocusDTO)?.info &&
+      !this.fetchMeetingInfoTimeoutId
+    ) {
+      try {
+        this.fetchMeetingInfo({});
```
marcin-bazyl: await is missing here

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1773,8 +1777,12 @@ export default class Meetings extends WebexPlugin {
             waitingTime
           );
           meeting.parseMeetingInfo(undefined, destination);
-        } else {
+        } else if (!shouldDeferMeetingInfoFetch) {
           await meeting.fetchMeetingInfo(meetingInfoOptions);
+        } else {
+          LoggerProxy.logger.info(
+            'Meetings:index#createMeeting --> defer fetchMeetingInfo for incomplete locus, will retry after locus initialSetup'
```
marcin-bazyl: nitpick: the log should probably say "will do it after...." instead of "will retry after..." because we're not retrying it, we're just doing it later

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -11353,6 +11353,93 @@ describe('plugin-meetings', () => {
...
+
+          assert.notCalled(fetchMeetingInfoStub);
+        });
+
+        it('swallows sync fetchMeetingInfo errors and logs info', () => {
+          const error = new Error('fetch failed');
+
+          meeting.meetingInfo = {};
+          meeting.destination = {url: 'https://locus.example.com/locus/123', info: {topic: 'x'}};
+          sinon.stub(meeting, 'fetchMeetingInfo').throws(error);
```
marcin-bazyl: I think this test would better simulate the failure if it returned a rejected promise rather than throwing synchronously - then it would have caught the missing "await" in the code (see the other comment)

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -592,17 +592,21 @@ export default class LocusInfo extends EventsScope {
           locus: LocusDTO;
           dataSets?: DataSet[];
           metadata?: Metadata;
+          onLocusSynced?: (locus: LocusDTO) => void;
```
marcin-bazyl: nitpick: feels like onLocusSynced doesn't really belong inside `data` - I think `data` is supposed to be related to the data we received from Locus and this is just a callback specific to how we want this function to work, so kind of feels like the callback should be a separate argument rather than being inside data, do you agree?

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -672,12 +677,47 @@ export default class LocusInfo extends EventsScope {
...
+  /**
+   * Builds a full locus DTO snapshot from current internal locus state.
+   *
+   * @returns {LocusDTO}
+   */
+  private getCurrentLocusSnapshot(): LocusDTO {
+    const locus: Record<string, any> = {};
+
+    LocusDtoTopLevelKeys.forEach((key) => {
+      const value = (this as Record<string, any>)[key];
```
marcin-bazyl, line 706: just an observation: I'm not sure if we store everything from LocusDTO in LocusInfo class unaltered, this code is very messy, it's hard to tell without checking a lot of places manually, I guess you could ask AI to check that, but I can see we store at least the main things like controls, self, info, fullState, so I guess it should be good enough

marcin-bazyl (approved):
overall looks good, I've made some minor comments, it would be great if you could address them before merging, but I'm approving anyway, just so that you're not blocked waiting for me because of the time zone differences

JudyZhuHz (approved):

packages/@webex/plugin-meetings/src/locus-info/index.ts
WeijuanShao: Yes, can we just cache the whole locus instead of specific attributes?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4659,6 +4659,33 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      // so refresh it after locus sync to avoid stale partial hash-tree data.
+      this.destination = locus;
+    }
+    if (
+      (!this.meetingInfo || isEmpty(this.meetingInfo)) &&
+      (this.destination as LocusDTO)?.info &&
+      !this.fetchMeetingInfoTimeoutId
+    ) {
+      try {
+        await this.fetchMeetingInfo({});
```
marcin-bazyl, line 4680: while testing I've noticed that this gets called for example when I receive an incoming 1-1 call, so maybe we need to improve the condition to avoid it, but I guess it hasn't caused any issue, I managed to answer the call, so it's OK to merge this PR and think about more changes in another PR

PR #4883, opened 2026-04-20
fix(jmt): stayLobbyTimeCappedBy should return 0 if there's no lobby for the meeting

gabrielchl:

an earlier change changed how we use the stayLobbyTime in some JMT calculations (introducing the new stayLobbyTimeCappedBy function, but I forgot that some meetings don't have a lobby and therefore stayLobbyTime could be undefined

the previous implementation of those metrics that uses stayLobbyTime would treat lobbyTime as 0
the new implementation could potentially use an overly-long time period (up to the cap) as the "lobby time" for meetigns without lobby

treat lobbyTime as 0 in calculations if the start point (client.lobby.entered, we use this to determine if the meeting has a lobby or not) doesn't exist

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
+++ b/packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
@@ -314,12 +314,16 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
    * @returns - latency
    */
   public getStayLobbyTimeCappedBy(endTimestampKey: MetricEventNames) {
-    const lobbyStartTimestamp = this.latencyTimestamps.get('client.locus.join.response'); // must exist
-    const lobbyEndTimestamp = this.latencyTimestamps.get('client.lobby.exited'); // might not exist
+    const lobbyStartTimestamp = this.latencyTimestamps.get('client.locus.join.response'); // might not exist (some meetings don't have lobby)
+    const lobbyEndTimestamp = this.latencyTimestamps.get('client.lobby.exited'); // might not exist (some meetings don't have lobby / user still in lobby at the time of measurement)
     const maximumEndTimestamp = this.latencyTimestamps.get(endTimestampKey); // must exist
 
-    if (typeof lobbyStartTimestamp !== 'number' || typeof maximumEndTimestamp !== 'number') {
+    if (typeof maximumEndTimestamp !== 'number') {
       return undefined;
+    }
+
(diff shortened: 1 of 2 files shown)


nataliadelmar (approved):

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -426,7 +426,7 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
       'internal.client.interstitial-window.click.joinbutton',
       'client.ice.end'
     );
-    const stayLobbyTimeCappedByIceEnd = this.getStayLobbyTimeCappedBy('client.ice.end');
```
marcin-bazyl, line 429: wouldn't it be better to change getStayLobbyTimeCappedBy() to always return a number, so return 0 instead of undefined? after all if there was no lobby then it's true to say that user stayed 0 seconds in the lobby, right? and it avoids anyone making the same mistake when using getStayLobbyTimeCappedBy() in more places in the future (plus existing comment for getStayLobbyTimeCappedBy() doesn't say it can return undefined)

packages/@webex/internal-plugin-metrics/src/call-diagnostic/call-diagnostic-metrics-latencies.ts
```diff
@@ -314,14 +314,18 @@ export default class CallDiagnosticLatencies extends WebexPlugin {
    * @returns - latency
    */
   public getStayLobbyTimeCappedBy(endTimestampKey: MetricEventNames) {
-    const lobbyStartTimestamp = this.latencyTimestamps.get('client.locus.join.response'); // must exist
-    const lobbyEndTimestamp = this.latencyTimestamps.get('client.lobby.exited'); // might not exist
+    const lobbyStartTimestamp = this.latencyTimestamps.get('client.locus.join.response'); // might not exist (some meetings don't have lobby)
```
marcin-bazyl: the comment doesn't seem correct.... we're getting a timestamp for 'client.locus.join.response' so join response always exists (if we joined) no matter if a meeting has a lobby or not

marcin-bazyl (approved):

nataliadelmar (approved):

PR #4886, opened 2026-04-20
fix(contact-center): consult flow state-machine fixes for task-refactor

akulakum:
# COMPLETES #<https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7849>

The CC Widgets task-refactor migration exposed eight issues in the SDK state machine. These affect page refresh during consult, consult hold state synchronization, early consult termination handling, hold/resume event emission during the CONSULTING state, missing auto-wrapup object creation after call end and consult transfer, incorrect state restoration on page refresh during wrapup, and incorrect HELD state restoration when the backend reports media-level hold without updating interaction state, and mute button disappearing during held state in desktop mode — all of which cause incorrect UI behavior in the widget layer.

**1. Hydrate consult context on page refresh (Issue 1)**

- Extended deriveTaskDataUpdates in actions.ts to reconstruct consultInitiator, consultDestinationAgentJoined, and consultCallHeld context flags from persisted taskData during rehydration — reusing the existing action instead of adding a new hydrateConsultState action
- Added taskData.isConsulted === false fallback (strict equality) for consultInitiator derivation to avoid misclassifying consulted agents when determineConsultInitiator returns undefined
- Gated consultCallHeld derivation behind consultInitiator check so only the initiating agent computes consult hold state 
- Without this, page refresh during an active consult left context flags empty, causing uiControls to compute no consult controls

**2. Sync consultCallHeld on hold/unhold events (Issue 2)**

- Enhanced existing setHoldState action in actions.ts to update context.consultCallHeld when HOLD_SUCCESS / UNHOLD_SUCCESS events target the consult media, gated behind context.consultInitiator — reusing the existing action instead of adding a new syncConsultCallHeld action
- Without this, hold/unhold events during consult did not update consultCallHeld, so consult-on-hold controls did not reflect the call switch

**3. Handle CONSULT_END in OFFERED state (Issue 3)**

- Added CONSULT_END handler to the OFFERED state in TaskStateMachine.ts, transitioning to TERMINATED
- Without this, when the initiator ended the consult before the consulted agent accepted, the consulted agent's task remained stuck in OFFERED state indefinitely

**4. Emit TASK_HOLD / TASK_RESUME during CONSULTING state (Issue 4)**

- Added emitTaskHold action to the HOLD_SUCCESS handler and emitTaskResume action to the UNHOLD_SUCCESS handler within the CONSULTING state in TaskStateMachine.ts
- Without this, hold/unhold during consulting did not emit TASK_HOLD / TASK_RESUME events, causing the widget to miss hold state changes and not display consult-on-hold timers

**5. Fix auto-wrapup object not created after call end and consult transfer (Issue 5)**

- Updated emitTaskWrapup in Task.ts to call updateTaskFromEvent(event) before evaluating wrapUpRequired, ensuring Task.data is populated with wrapUpRequired: true so setupAutoWrapupTimer creates the autoWrapup object correctly
- Without this, emitTaskWrapup checked a local variable extracted from the event but never persisted it to Task.data, so setupAutoWrapupTimer() saw wrapUpRequired as undefined and never created the autoWrapup object

**6. Restore WRAPPING_UP state on page refresh (Issue 6)**

- Enhanced isInteractionTerminated guard in guards.ts to also check participant[selfAgentId].isWrapUp === true for the current agent during the HYDRATE event — consolidated into the existing guard instead of adding a separate isAgentInWrapUp guard, per review feedback
- Without this, page refresh while an agent was in wrapup (e.g., after consult transfer) fell through to the isInteractionConnected guard and restored the task to CONNECTED state instead of WRAPPING_UP

**7. Restore HELD state on page refresh when media is on hold (Issue 7)**

- Enhanced isInteractionHeld guard in guards.ts to cross-check the primary media's isHold flag via mainInteractionId when interaction.state is not 'hold'
- The backend may report interaction.state: "connected" while media[mainInteractionId].isHold: true, causing the state machine to hydrate into CONNECTED instead of HELD
- Without this, page refresh during a held call required two clicks on Resume — the first incorrectly sending a hold request (AgentContactHeld) because the state machine was in CONNECTED

**8. Keep mute button visible (disabled) during held state in desktop mode (Issue 8)**

- In uiControlsComputer.ts, the mute visibility gate checked state === TaskState.CONNECTED || isConferencing, which excluded the HELD state entirely
- When the state machine transitioned to HELD, the mute logic fell through to return DISABLED (hidden), even though the existing isHeld ? VISIBLE_DISABLED : VISIBLE_ENABLED line was designed to handle this case — it was unreachable
- Changed the condition to isConnected || isHeld || isConferencing, keeping the mute button visible but disabled during hold

Incoming task not cleared at Agent 2 side when Agent 1 ends consult before acceptance

image

**Agent 2 consults → transfers to Agent 1 → Agent 2 sees wrapup → refresh page → should now show wrapup screen**

WrapUp Screen shouold be visible instead displaying call control on page refresh

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Normal call: Agent ends call → auto-wrapup banner appears with countdown → task auto-wraps up
- Consult transfer: Agent 1 consults Agent 2 → Agent 2 accepts → Agent 1 switches consult to main → Agent 1 transfers → Agent 1 sees auto-wrapup banner and task auto-wraps up
- Page refresh during wrapup: Agent 2 consults Agent 1 → Agent 2 transfers → Agent 2 sees wrapup screen → page refresh → Agent 2 still sees wrapup screen (not call controls)
- Page refresh during consult: Agent in active consult → page refresh → consult controls restored correctly
- Multi-login hold/unhold sync: Hold/unhold from second browser session reflects correctly in first session
- Early consult end: Agent 1 ends consult before Agent 2 accepts → Agent 2's task transitions to TERMINATED

- Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [ ] Github Copilot
  - [x] Other - Cursor AI
- This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

- [x] I have read and followed contributing guidelines

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md) before submitting.

### DIFF
--- a/packages/@webex/contact-center/src/services/task/Task.ts
+++ b/packages/@webex/contact-center/src/services/task/Task.ts
@@ -458,11 +458,9 @@ export default abstract class Task extends EventEmitter implements ITask {
         this.emit(TASK_EVENTS.TASK_REJECT, reason);
       },
       emitTaskWrapup: ({event}: {event?: TaskEventPayload}) => {
-        const wrapUpRequiredFromEvent =
-          event && typeof event === 'object' && 'taskData' in event
-            ? (event as {taskData?: TaskData}).taskData?.wrapUpRequired
-            : undefined;
-        const shouldEmitWrapup = Boolean(wrapUpRequiredFromEvent ?? this.data.wrapUpRequired);
+        this.updateTaskFromEvent(event);
+
+        const shouldEmitWrapup = Boolean(this.data.wrapUpRequired);
         if (!shouldEmitWrapup) {
           LoggerProxy.info(`Skipping task:wrapup event - wrapUpRequired is false`, {
             module: TASK_FILE,
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -95,7 +95,7 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             {
               guard: guards.isInteractionConsulting,
               target: TaskState.CONSULTING,
-              actions: ['updateTaskData', 'emitTaskHydrate'],
+              actions: ['updateTaskData', 'hydrateConsultState', 'emitTaskHydrate'],
             },
             {
               guard: guards.isInteractionHeld,
@@ -212,6 +212,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           // AgentConsultFailed - when consulted agent (Agent 2) doesn't answer (RONA or decline)
           // Clears the incoming consult notification by transitioning to TERMINATED
           [TaskEvent.CONSULT_FAILED]: {
+            target: TaskState.TERMINATED,
+            actions: ['updateTaskData', 'clearConsultState', 'emitTaskReject'],
+          },
+          // AgentConsultEnded - when consult initiator (Agent 1) ends the consult before
+          // the consulted agent (Agent 2) accepts. Clears the incoming notification.
+          [TaskEvent.CONSULT_END]: {
             target: TaskState.TERMINATED,
             actions: ['updateTaskData', 'clearConsultState', 'emitTaskReject'],
           },
@@ -482,10 +488,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             actions: ['handleSwitchToConsult', 'emitTaskSwitchCall'],
           },
           [TaskEvent.HOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState'],
+            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskHold'],
           },
           [TaskEvent.UNHOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState'],
+            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskResume'],
           },
 
           [TaskEvent.TRANSFER_SUCCESS]: [
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -103,6 +103,11 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
               actions: ['updateTaskData', 'emitTaskHydrate'],
             },
             {
+              guard: guards.isAgentInWrapUp,
+              target: TaskState.WRAPPING_UP,
+              actions: ['updateTaskData', 'markEnded', 'emitTaskHydrate'],
+            },
+            {
               guard: guards.isInteractionConnected,
               target: TaskState.CONNECTED,
               actions: ['updateTaskData', 'emitTaskHydrate'],
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -95,17 +95,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             {
               guard: guards.isInteractionConsulting,
               target: TaskState.CONSULTING,
-              actions: ['updateTaskData', 'hydrateConsultState', 'emitTaskHydrate'],
+              actions: ['updateTaskData', 'emitTaskHydrate'],
             },
             {
               guard: guards.isInteractionHeld,
               target: TaskState.HELD,
               actions: ['updateTaskData', 'emitTaskHydrate'],
-            },
-            {
-              guard: guards.isAgentInWrapUp,
-              target: TaskState.WRAPPING_UP,
-              actions: ['updateTaskData', 'markEnded', 'emitTaskHydrate'],
             },
             {
               guard: guards.isInteractionConnected,
@@ -493,10 +488,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             actions: ['handleSwitchToConsult', 'emitTaskSwitchCall'],
           },
           [TaskEvent.HOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskHold'],
+            actions: ['updateTaskData', 'setHoldState', 'emitTaskHold'],
           },
           [TaskEvent.UNHOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskResume'],
+            actions: ['updateTaskData', 'setHoldState', 'emitTaskResume'],
           },
 
           [TaskEvent.TRANSFER_SUCCESS]: [
--- a/packages/@webex/contact-center/src/services/task/state-machine/actions.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/actions.ts
@@ -259,10 +259,82 @@ export const actions: TaskActionsMap = {
   setTransferConferenceRequested: assign({transferConferenceRequested: true}),
   clearTransferConferenceRequested: assign({transferConferenceRequested: false}),
 
+  /**
+   * Reconstruct consult context flags from hydrated task data on page refresh.
+   * During live flows these are set incrementally by setConsultInitiator,
+   * setConsultAgentJoined, and handleSwitchToMainCall/handleSwitchToConsult.
+   * After HYDRATE those actions never fire, so we derive the same flags from
+   * the backend snapshot.
+   */
+  hydrateConsultState: assign(({context, event}: TaskActionArgs) => {
+    const taskData = getTaskDataFromEvent(event);
+    if (!taskData?.interaction) return {};
+
+    const interaction = taskData.interaction;
+    const selfAgentId = context.uiControlConfig?.agentId ?? taskData?.agentId;
+
+    const derivedInitiator = determineConsultInitiator(taskData, selfAgentId);
+    const consultInitiator = derivedInitiator ?? taskData.isConsulted !== true;
+
+    const consultDestinationAgentJoined = Boolean(
+      interaction.participants &&
+        Object.values(interaction.participants).some(
+          (p: any) => p?.isConsulted === true && !p?.hasLeft
+        )
+    );
+
+    const consultMediaId = taskData.consultMediaResourceId;
+    const consultMedia: any = consultMediaId
+      ? interaction.media?.[consultMediaId]
+      : Object.values(interaction.media ?? {}).find((m: any) => m?.mType === 'consult');
+    const consultCallHeld = Boolean(consultMedia?.isHold);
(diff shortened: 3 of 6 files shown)


packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -102,6 +102,16 @@ export const guards = {
     return taskData?.interaction?.state === 'hold';
   },
 
+  isAgentInWrapUp: ({context, event}: GuardParams): boolean => {
```
Kesari3008: Why do we need this guard ? It looks very similar to what we are trying to do in isInteractionTerminated so keep only one: either this or that

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -95,13 +95,18 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
-              actions: ['updateTaskData', 'emitTaskHydrate'],
+              actions: ['updateTaskData', 'hydrateConsultState', 'emitTaskHydrate'],
             },
             {
               guard: guards.isInteractionHeld,
               target: TaskState.HELD,
               actions: ['updateTaskData', 'emitTaskHydrate'],
             },
+            {
+              guard: guards.isAgentInWrapUp,
```
Kesari3008: This is similar to very first transitions, please check if this is really required

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -482,10 +493,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             actions: ['handleSwitchToConsult', 'emitTaskSwitchCall'],
           },
           [TaskEvent.HOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState'],
+            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskHold'],
```
Kesari3008: I don't think syncConsultCallHeld action is needed as we already have consultCallheld being set to true/false everytime switch is attempted so it's only a matter of updating taskdata with latest context value which will have updated consultCallHeld and we also have setHoldState so if any remaining logic is needed to be fixed should be done there

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -482,10 +493,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
           },
           [TaskEvent.UNHOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState'],
+            actions: ['updateTaskData', 'setHoldState', 'syncConsultCallHeld', 'emitTaskResume'],
```
Kesari3008: Even for emission of events, these events will only be received in consult state when we attempt switch and for that we are emitting an event so try to align things  with the help og existing events being listened for instead of adding new ones

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -259,10 +259,82 @@ export const actions: TaskActionsMap = {
...
   clearTransferConferenceRequested: assign({transferConferenceRequested: false}),
 
+  /**
+   * Reconstruct consult context flags from hydrated task data on page refresh.
+   * During live flows these are set incrementally by setConsultInitiator,
+   * setConsultAgentJoined, and handleSwitchToMainCall/handleSwitchToConsult.
+   * After HYDRATE those actions never fire, so we derive the same flags from
+   * the backend snapshot.
+   */
+  hydrateConsultState: assign(({context, event}: TaskActionArgs) => {
```
Kesari3008: Even for this action, we do have some existing actions which update consult related data in context, see if we can use those instead of creating this new one

Kesari3008 (changes requested):

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
akulakum: Agreed. Merged the isAgentInWrapUp logic into the existing isInteractionTerminated guard. It now checks for interaction.isTerminated === true OR participant[selfAgentId].isWrapUp === true, covering both terminated interactions and agent-level wrapup state in a single guard. The standalone isAgentInWrapUp guard has been removed.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Removed the separate isAgentInWrapUp transition. The existing isInteractionTerminated transition (which is the very first guard in the HYDRATE chain) now handles both cases — it targets WRAPPING_UP with actions ['updateTaskData', 'markEnded', 'emitTaskHydrate'] for both terminated interactions and agents in wrapup. No additional transition entry needed.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Agreed. Removed syncConsultCallHeld entirely. The consultCallHeld update logic is now handled within the existing setHoldState action — it checks if the held/unheld media is the consult media and if the current agent is the consultInitiator, then updates consultCallHeld accordingly. Retained emitTaskHold in the CONSULTING HOLD_SUCCESS handler as it is needed to emit the TASK_HOLD event so the widget receives the hold state change notification for displaying the consult-on-hold timer.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Removed syncConsultCallHeld as noted above. Regarding emitTaskHold/emitTaskResume — these are not new events or actions, they are existing actions (emitTaskHold and emitTaskResume) already defined in the actions map. They were just not included in the CONSULTING state's HOLD_SUCCESS/UNHOLD_SUCCESS handlers. Without them, the widget does not receive TASK_HOLD/TASK_RESUME event callbacks during consult, so the consult-on-hold timer is not triggered. These are the same actions used in the CONNECTED and HELD states for the same purpose.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: Agreed. Removed hydrateConsultState entirely. The consult context reconstruction logic is now handled within the existing deriveTaskDataUpdates action, which already runs during hydration via updateTaskData. It now derives consultInitiator (using determineConsultInitiator with a taskData.isConsulted === false strict fallback), consultDestinationAgentJoined, and consultCallHeld (gated behind consultInitiator) from the hydration payload — no new action required.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -99,7 +106,14 @@ export const guards = {
   isInteractionHeld: ({event}: GuardParams): boolean => {
     const taskData = getTaskDataFromEvent(event);
 
-    return taskData?.interaction?.state === 'hold';
+    if (taskData?.interaction?.state === 'hold') return true;
```
Kesari3008, line 109: Why is this change required ?

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -91,7 +91,39 @@ const deriveTaskDataUpdates = (context: TaskContext, taskData: TaskData | undefi
...
+            updates.consultInitiator = true;
+          }
+        }
+
+        if (taskData.interaction?.state === 'consulting') {
+          if (!context.consultDestinationAgentJoined) {
+            const hasJoinedConsultee = Boolean(
+              taskData.interaction.participants &&
+                Object.values(taskData.interaction.participants).some(
+                  (p: any) => p?.isConsulted === true && !p?.hasLeft
```
Kesari3008, line 109: We should not use any and we have actions to set consultInitiator, consultDestinationAgentJoined and consultCallHeld so why don't we invoke those at hydrate

Kesari3008 (approved):
Approving but would be good to really make these changes keeping in mind that there might be something existing which already take care of it but just needs to be used

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -458,11 +458,9 @@ export default abstract class Task extends EventEmitter implements ITask {
...
       },
       emitTaskWrapup: ({event}: {event?: TaskEventPayload}) => {
-        const wrapUpRequiredFromEvent =
-          event && typeof event === 'object' && 'taskData' in event
-            ? (event as {taskData?: TaskData}).taskData?.wrapUpRequired
-            : undefined;
-        const shouldEmitWrapup = Boolean(wrapUpRequiredFromEvent ?? this.data.wrapUpRequired);
+        this.updateTaskFromEvent(event);
+
+        const shouldEmitWrapup = Boolean(this.data.wrapUpRequired);
```
rsarika, line 463: as per old logic we are doing below, is it not applicable here also?

  payload.data.interaction?.participants?.[this.agentId]?.isWrapUp || false,

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -323,6 +356,17 @@ export const actions: TaskActionsMap = {
         },
       },
     };
+
+    const consultMediaId = context.taskData?.consultMediaResourceId;
```
rsarika, line 360: cross check consultMediaResourceId logic in old widgets util's findMediaResourceId

rsarika (approved):

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: Checked against the old widgets findMediaResourceId in task-utils.ts. The old logic iterates over interaction.media and matches by mType. Our SDK logic follows the same approach — it first tries taskData.consultMediaResourceId for a direct lookup, and if unavailable, falls back to finding the media entry with mType === 'consult' via Object.values(media).find(m => m.mType === 'consult'). This is functionally equivalent to findMediaResourceId(task, 'consult'). The same pattern is used in setHoldState for identifying consult media during hold/unhold events. Consistent with old widgets behavior.

packages/@webex/contact-center/src/services/task/Task.ts
akulakum: Good callout. The current logic works because this.updateTaskFromEvent(event) is called first, which merges the full event payload into Task.data including wrapUpRequired. By the time we evaluate this.data.wrapUpRequired, it's already populated from the backend event. This is working correctly across all tested scenarios — normal call end, consult transfer, and page refresh. The participant.isWrapUp fallback would only help if the backend sends an event where wrapUpRequired is missing but isWrapUp is true, which hasn't been observed. Can add it as a defensive fallback in a follow-up if edge cases arise.

PR #4899, opened 2026-04-22
feat(plugin-meetings): abandon sync

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-780408

syncAbortController on InternalDataSet — Each dataset gets an AbortController when performSync starts. Incoming messages call .abort() on it to signal cancellation.

cancelPendingSyncsForDataSets() method — Removes matching entries from syncQueue AND calls .abort() on any active syncAbortController. Called from two places:

parseMessage() — when a message with locusStateElements arrives
handleRootHashHeartBeatMessage() — when a heartbeat arrives
Abort checks in performSync() — After /hashtree returns and before /sync is sent, checks abortController.signal.aborted. If true, logs and returns without sending /sync.

Cleanup in stop(), deleteHashTree() — Aborts any in-flight sync and clears the controller.

AI is used to write tests

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1243,6 +1261,15 @@ class HashTreeParser {
       }
       // request sync for mismatched leaves
       if (Object.keys(leavesData).length > 0) {
+        // check if sync was aborted before sending /sync request
+        if (abortController.signal.aborted) {
```
marcin-bazyl: this feels like unnecessary duplication of code from lines 1245-1252. there is only 1 async thing we do before calling sendSyncRequestToLocus() so we should only need one check like this just before calling sendSyncRequestToLocus()

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1262,6 +1289,40 @@ class HashTreeParser {
           error
         );
       }
+    } finally {
+      if (dataSet.syncAbortController === abortController) {
```
marcin-bazyl: why this `if` here? it assumes that there may be other abortControllers set? that could happen if performSync() is called again before previous one is finished, which shouldn't happen because we use a queue, but the queueing logic is outside of performSync() so it's fair enough to assume that performSync() might be called more than once. In that case I feel that it would be better to change the code where it creates the abort controller to not create one if there is one already there, so that we always have just one abort controller per dataset and if performSync() is ever called more than once in parallel for the same dataset, all of them should be aborted if that single abort controller is aborted

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1188,6 +1194,9 @@ class HashTreeParser {
       return;
     }
 
+    const abortController = dataSet.syncAbortController ?? new AbortController();
```
shnaaz, line 1265: Reuse existing controller: if one already exists (from a hypothetical parallel call), all calls share it and a single abort cancels them all.Removed the guard from finally

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -965,6 +970,7 @@ class HashTreeParser {
     // first, update our metadata about the datasets with info from the message
     this.visibleDataSetsUrl = visibleDataSetsUrl;
     dataSets.forEach((dataSet) => this.updateDataSetInfo(dataSet));
+    this.cancelPendingSyncsForDataSets(dataSets.map((ds) => ds.name));
```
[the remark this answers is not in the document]
marcin-bazyl: this is fine, this comment is invalid

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTreeParser.ts
```diff
@@ -4320,6 +4320,245 @@ describe('HashTreeParser', () => {
     });
   });
 
+  describe('#performSync abort controller', () => {
+    it('should reuse an existing syncAbortController if one is already set on the dataset', async () => {
```
marcin-bazyl, line 4739: I don't see this test doing what it says - it doesn't check that `existingController` is reused

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTreeParser.ts
```diff
@@ -4320,6 +4320,245 @@ describe('HashTreeParser', () => {
...
+
+      // Resolve GET hashtree with mismatched hashes so the code would normally proceed to /sync
+      resolveGetHashtree({
+        body: {
+          hashes: new Array(16).fill('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1'),
+          dataSet: createDataSet('main', 16, 1100),
+        },
+      });
+
+      await clock.tickAsync(0);
```
marcin-bazyl: nitpick: here we should probably use `await testUtils.flushPromises();`? same for other places where we call tickAsync(0)

packages/@webex/plugin-meetings/test/unit/spec/hashTree/hashTreeParser.ts
```diff
@@ -4320,6 +4320,245 @@ describe('HashTreeParser', () => {
...
+          locusUrl,
+        },
+        'trigger self sync'
+      );
+
+      // Fire the timer to start the sync
+      await clock.tickAsync(1000);
+
+      // POST sync should NOT have been called because the controller was already aborted
+      assert.neverCalledWith(webexRequest, sinon.match({method: 'POST', uri: `${selfUrl}/sync`}));
```
marcin-bazyl, line 4853: would be useful to also check that /hashtree is not called (same might apply to some of the other tests too)

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1263,6 +1280,38 @@ class HashTreeParser {
           error
         );
       }
+    } finally {
+      dataSet.syncAbortController = undefined;
```
marcin-bazyl, line 1356: thinking more about that AI comment that I replied to saying it's invalid, there is one aspect of it that would be good to fix - it is true that parseMessage() is called from performSync() (via handleMessage) so each time we do a sync, at the point when parseMessage() calls cancelPendingSyncsForDataSets() the abort controller will still be set on that data set, so we will see a log saying that we're abort a sync, which isn't really true. It would be good to fix that, one way would be to set `syncAbortController` to undefined before handleMessage() is called (but keep it here in finally too just in case if something fails) - it's not a super clean solution, so maybe you'll come up with a better one, but either way we need to do something about this

marcin-bazyl (approved):

jor-row (approved):

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1250,6 +1267,10 @@ class HashTreeParser {
       // sync API may return nothing (in that case data will arrive via messages)
       // or it may return a response in the same format as messages
       if (syncResponse) {
+        // clear the abort controller before processing the response so that
+        // parseMessage() -> cancelPendingSyncsForDataSets() doesn't log a
+        // misleading "aborting sync" message for this already-completed sync
+        dataSet.syncAbortController = undefined;
         this.handleMessage(syncResponse, 'via sync API');
```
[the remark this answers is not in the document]
marcin-bazyl: this is fine

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -546,6 +547,21 @@ class HashTreeParser {
...
+        dataSets.map(({name, root, leafCount, version}) => ({
+          name,
+          root,
+          leafCount,
+          version,
+        }))
+      )}`
+    );
+
+    this.cancelPendingSyncsForDataSets(dataSets.map((ds) => ds.name));
```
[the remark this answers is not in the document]
marcin-bazyl: this is fine

PR #4907, opened 2026-04-23
fix(meetings): fetching the correct locus url

Kesari3008:
# COMPLETES #

BEMS01973104 [New] EXTENDED CARE  ZD: #146772 - Breakout room issues with the JS SDK

Created new method to fetch the correct/accurate local url instead of using meeting.selfUrl which can be stale.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [x] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -810,8 +810,6 @@ export default class Meeting extends StatelessWebexPlugin {
   private logUploadIntervalIndex: number;
   private mediaServerIp: string;
   private llmHealthCheckTimer?: ReturnType<typeof setTimeout>;
-  private returnToMainRoapLocus?: LocusDTO;
-  private returnToMainSelfUrlPromise?: Promise<string>;
 
   /**
    * @param {Object} attrs
@@ -4068,108 +4066,6 @@ export default class Meeting extends StatelessWebexPlugin {
       Object.keys(object).forEach((key) => {
         this[key] = object[key];
       });
-    }
-  }
-
-  /**
-   * Marks that ROAP should resolve the main session selfUrl before sending because
-   * the current breakout locus has ended and may no longer accept /media requests.
-   * @param {LocusDTO} closedBreakoutLocus closed breakout locus
-   * @returns {void}
-   * @public
-   * @memberof Meeting
-   */
-  public updateClosedBreakoutLocus(closedBreakoutLocus: LocusDTO) {
-    this.returnToMainRoapLocus = closedBreakoutLocus;
-  }
-
-  /**
-   * Returns the selfUrl that ROAP /media should target.
-   * In the normal case this is just meeting.selfUrl. During a confirmed
-   * breakout-ended transition we first resolve the main session participant
-   * selfUrl so ROAP is not sent to the inactive breakout locus.
-   * @returns {Promise<string>}
-   * @public
-   * @memberof Meeting
-   */
-  public async resolveRoapMediaSelfUrl(): Promise<string> {
-    const closedBreakoutLocus = this.returnToMainRoapLocus;
-
-    if (!closedBreakoutLocus) {
-      if (!this.selfUrl) {
-        throw new Error('Cannot send ROAP: meeting selfUrl is not available');
-      }
-
-      return this.selfUrl;
-    }
-
-    if (!this.returnToMainSelfUrlPromise) {
-      this.returnToMainSelfUrlPromise = (async () => {
-        const cachedMainLocus = this.locusInfo?.mainSessionLocusCache;
-        let mainLocus =
-          cachedMainLocus?.self?.url &&
-          cachedMainLocus?.controls?.breakout?.sessionType !== BREAKOUTS.SESSION_TYPES.BREAKOUT &&
-          !cachedMainLocus?.self?.removed
-            ? cachedMainLocus
-            : undefined;
-
-        if (!mainLocus) {
-          const mainLocusUrl =
-            this.breakouts?.mainLocusUrl ||
-            cachedMainLocus?.url ||
-            MeetingsUtil.getThisDevice(closedBreakoutLocus, this.deviceUrl)?.replaces?.[0]
-              ?.locusUrl;
-
-          if (!mainLocusUrl) {
-            throw new Error('Cannot send ROAP: main locus URL is not available');
-          }
-
-          const response = await this.meetingRequest.getLocusDTO({url: mainLocusUrl});
-
-          mainLocus = response.body;
-        }
-
-        if (
-          !mainLocus?.self?.url ||
-          mainLocus?.controls?.breakout?.sessionType === BREAKOUTS.SESSION_TYPES.BREAKOUT ||
-          mainLocus?.self?.removed
-        ) {
-          throw new Error('Cannot send ROAP: main locus does not contain usable selfUrl');
-        }
-
-        const currentLocusUrl = this.locusUrl;
-
-        if (
-          currentLocusUrl &&
-          closedBreakoutLocus.url &&
-          mainLocus.url &&
-          currentLocusUrl !== closedBreakoutLocus.url &&
-          currentLocusUrl !== mainLocus.url
-        ) {
-          throw new Error(
-            'Cannot send ROAP: meeting moved to another locus during breakout return'
-          );
-        }
-
-        this.locusInfo.updateMainSessionLocusCache(mainLocus);
-        this.locusInfo.onFullLocus('return-to-main media selfUrl before ROAP', mainLocus);
-
-        if (this.selfUrl !== mainLocus.self.url) {
-          this.updateMeetingObject({selfUrl: mainLocus.self.url});
-        }
-
-        return mainLocus.self.url;
-      })();
-    }
-
-    try {
-      const selfUrl = await this.returnToMainSelfUrlPromise;
-
-      this.returnToMainRoapLocus = undefined;
-
-      return selfUrl;
-    } finally {
-      this.returnToMainSelfUrlPromise = undefined;
     }
   }
 
@@ -8663,7 +8559,7 @@ export default class Meeting extends StatelessWebexPlugin {
           regionCode: this.webex.meetings.geoHintInfo?.regionCode,
         },
         preferTranscoding: !this.isMultistream,
-        resolveMediaSelfUrl: () => this.resolveRoapMediaSelfUrl(),
+        getCurrentSelfUrl: () => this.selfUrl,
       },
       {
         // @ts-ignore
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -740,6 +740,7 @@ export default class Meeting extends StatelessWebexPlugin {
   isMoveToInProgress = false;
   registrationIdStatus: string;
   brbState: BrbState;
+  private promisesWaitingForPropUpdate: Record<string, Defer> = {};
 
   voiceaListenerCallbacks: object = {
     [VOICEAEVENTS.VOICEA_ANNOUNCEMENT]: (payload: Transcription['languageOptions']) => {
@@ -4065,6 +4066,11 @@ export default class Meeting extends StatelessWebexPlugin {
     if (object && Object.keys(object).length) {
       Object.keys(object).forEach((key) => {
         this[key] = object[key];
+
+        if (this.promisesWaitingForPropUpdate[key]) {
+          this.promisesWaitingForPropUpdate[key].resolve();
+          delete this.promisesWaitingForPropUpdate[key];
+        }
       });
     }
   }
@@ -8560,12 +8566,36 @@ export default class Meeting extends StatelessWebexPlugin {
         },
         preferTranscoding: !this.isMultistream,
         getCurrentSelfUrl: () => this.selfUrl,
+        waitForSelfUrlChange: () => this.waitForSelfUrlChange(),
       },
       {
         // @ts-ignore
         parent: this.webex,
       }
     );
+  }
+
+  /**
+   * Waits for LocusInfo to update meeting.selfUrl, with a timeout fallback.
+   * @returns {Promise<void>}
+   * @private
+   * @memberof Meeting
+   */
+  private waitForSelfUrlChange(): Promise<void> {
+    if (!this.promisesWaitingForPropUpdate.selfUrl) {
+      const pendingSelfUrlUpdate = new Defer();
+
+      this.promisesWaitingForPropUpdate.selfUrl = pendingSelfUrlUpdate;
+
+      setTimeout(() => {
+        if (this.promisesWaitingForPropUpdate.selfUrl === pendingSelfUrlUpdate) {
+          delete this.promisesWaitingForPropUpdate.selfUrl;
+          pendingSelfUrlUpdate.resolve();
+        }
+      }, 5000);
+    }
+
+    return this.promisesWaitingForPropUpdate.selfUrl.promise;
   }
 
   /**
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -4065,9 +4065,15 @@ export default class Meeting extends StatelessWebexPlugin {
     // is not changed by any delta event
     if (object && Object.keys(object).length) {
       Object.keys(object).forEach((key) => {
+        const previousValue = this[key];
+
         this[key] = object[key];
 
-        if (this.promisesWaitingForPropUpdate[key]) {
+        const shouldResolveWaiter =
+          key !== 'selfUrl' ||
+          (this.promisesWaitingForPropUpdate[key] && previousValue !== object[key]);
+
+        if (this.promisesWaitingForPropUpdate[key] && shouldResolveWaiter) {
           this.promisesWaitingForPropUpdate[key].resolve();
           delete this.promisesWaitingForPropUpdate[key];
         }
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -810,6 +810,8 @@ export default class Meeting extends StatelessWebexPlugin {
   private logUploadIntervalIndex: number;
   private mediaServerIp: string;
   private llmHealthCheckTimer?: ReturnType<typeof setTimeout>;
+  private returnToMainRoapLocus?: LocusDTO;
+  private returnToMainSelfUrlPromise?: Promise<string>;
 
   /**
    * @param {Object} attrs
@@ -4066,6 +4068,108 @@ export default class Meeting extends StatelessWebexPlugin {
       Object.keys(object).forEach((key) => {
         this[key] = object[key];
       });
+    }
+  }
+
+  /**
+   * Marks that ROAP should resolve the main session selfUrl before sending because
+   * the current breakout locus has ended and may no longer accept /media requests.
+   * @param {LocusDTO} closedBreakoutLocus closed breakout locus
+   * @returns {void}
+   * @public
+   * @memberof Meeting
+   */
+  public updateClosedBreakoutLocus(closedBreakoutLocus: LocusDTO) {
+    this.returnToMainRoapLocus = closedBreakoutLocus;
+  }
+
+  /**
+   * Returns the selfUrl that ROAP /media should target.
+   * In the normal case this is just meeting.selfUrl. During a confirmed
+   * breakout-ended transition we first resolve the main session participant
+   * selfUrl so ROAP is not sent to the inactive breakout locus.
+   * @returns {Promise<string>}
+   * @public
+   * @memberof Meeting
+   */
+  public async resolveRoapMediaSelfUrl(): Promise<string> {
+    const closedBreakoutLocus = this.returnToMainRoapLocus;
+
+    if (!closedBreakoutLocus) {
+      if (!this.selfUrl) {
+        throw new Error('Cannot send ROAP: meeting selfUrl is not available');
+      }
+
+      return this.selfUrl;
+    }
+
+    if (!this.returnToMainSelfUrlPromise) {
+      this.returnToMainSelfUrlPromise = (async () => {
+        const cachedMainLocus = this.locusInfo?.mainSessionLocusCache;
+        let mainLocus =
+          cachedMainLocus?.self?.url &&
+          cachedMainLocus?.controls?.breakout?.sessionType !== BREAKOUTS.SESSION_TYPES.BREAKOUT &&
+          !cachedMainLocus?.self?.removed
+            ? cachedMainLocus
+            : undefined;
+
+        if (!mainLocus) {
+          const mainLocusUrl =
+            this.breakouts?.mainLocusUrl ||
+            cachedMainLocus?.url ||
+            MeetingsUtil.getThisDevice(closedBreakoutLocus, this.deviceUrl)?.replaces?.[0]
+              ?.locusUrl;
+
+          if (!mainLocusUrl) {
+            throw new Error('Cannot send ROAP: main locus URL is not available');
+          }
+
+          const response = await this.meetingRequest.getLocusDTO({url: mainLocusUrl});
+
+          mainLocus = response.body;
+        }
+
+        if (
+          !mainLocus?.self?.url ||
+          mainLocus?.controls?.breakout?.sessionType === BREAKOUTS.SESSION_TYPES.BREAKOUT ||
+          mainLocus?.self?.removed
+        ) {
+          throw new Error('Cannot send ROAP: main locus does not contain usable selfUrl');
+        }
+
+        const currentLocusUrl = this.locusUrl;
+
+        if (
+          currentLocusUrl &&
+          closedBreakoutLocus.url &&
+          mainLocus.url &&
+          currentLocusUrl !== closedBreakoutLocus.url &&
+          currentLocusUrl !== mainLocus.url
+        ) {
+          throw new Error(
+            'Cannot send ROAP: meeting moved to another locus during breakout return'
+          );
+        }
+
+        this.locusInfo.updateMainSessionLocusCache(mainLocus);
+        this.locusInfo.onFullLocus('return-to-main media selfUrl before ROAP', mainLocus);
+
+        if (this.selfUrl !== mainLocus.self.url) {
+          this.updateMeetingObject({selfUrl: mainLocus.self.url});
+        }
+
+        return mainLocus.self.url;
+      })();
+    }
+
+    try {
+      const selfUrl = await this.returnToMainSelfUrlPromise;
+
+      this.returnToMainRoapLocus = undefined;
+
+      return selfUrl;
+    } finally {
+      this.returnToMainSelfUrlPromise = undefined;
     }
   }
 
@@ -8559,6 +8663,7 @@ export default class Meeting extends StatelessWebexPlugin {
           regionCode: this.webex.meetings.geoHintInfo?.regionCode,
         },
         preferTranscoding: !this.isMultistream,
+        resolveMediaSelfUrl: () => this.resolveRoapMediaSelfUrl(),
       },
       {
         // @ts-ignore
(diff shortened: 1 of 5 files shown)


packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -101,7 +120,7 @@ export default class Roap extends StatelessWebexPlugin {
       return this.roapRequest
         .sendRoap({
           roapMessage,
-          locusSelfUrl: meeting.selfUrl,
+          locusSelfUrl: this.getLocusSelfUrl(meeting),
```
marcin-bazyl: I don't think this is the right approach. If you think that we end up in a situation where `meeting.selfUrl` is stale, then the fix should be to fix that and make sure that `meeting.selfUrl` has always up-to-date value and not to introduce a new getLocusSelfUrl() method. What's the point of having the public `selfUrl` property on a meeting if we know that it may be out-of-date?

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -77,6 +77,25 @@ export default class Roap extends StatelessWebexPlugin {
...
 
+  /**
+   * Returns the latest self URL to target Locus /media updates.
+   * During session moves (for example breakout -> main), meeting.selfUrl can lag behind
+   * the latest locus self snapshot for a short period.
+   * @param {Meeting} meeting
+   * @returns {string | undefined}
+   */
+  private getLocusSelfUrl(meeting: Meeting) {
+    const latestSelfUrl = meeting?.locusInfo?.parsedLocus?.self?.selfUrl;
```
marcin-bazyl: I can't see how `meeting?.locusInfo?.parsedLocus?.self?.selfUrl` can ever end up with a newer value than `meeting.selfUrl`. Can you explain this a bit more? I see `meeting.selfUrl` is updated via the call to `this.updateMeeting()` from `LocusInfo.updateSelf()` right before `meeting.locusInfo.parsedLocus.self` is updated, so they both should always have the same values.

packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
```diff
@@ -351,7 +351,15 @@ ControlsUtils.getSessionSwitchStatus = (oldLocus: any, newLocus: any) => {
 };
 
 ControlsUtils.isMainSessionDTO = (locus: any) => {
-  return locus?.controls?.breakout?.sessionType !== BREAKOUTS.SESSION_TYPES.BREAKOUT;
+  const sessionType = locus?.controls?.breakout?.sessionType;
+
+  // If breakout controls are present but sessionType is missing, treat as unknown and avoid
+  // misclassifying it as MAIN due to partial/in-flight breakout deltas.
+  if (locus?.controls?.breakout && !sessionType) {
```
marcin-bazyl: can you explain a bit more when can we end up in a situation where controls.breakout are present but without the sessionType? as far as I'm aware (and also according to Locus [docs](https://sqbu-github.cisco.com/pages/WebExSquared/locus/dto/locus/controls/breakout.html#breakoutcontrolsdto)) sessionType is not optional so whenever we have controls.breakout, we should always have the sessionType too

Also, I'm struggling to understand the connection between this fix here and the issue that this PR is trying to fix, can you explain a bit more?

packages/@webex/plugin-meetings/src/roap/index.ts
```diff
@@ -135,7 +154,7 @@ export default class Roap extends StatelessWebexPlugin {
 
     return this.roapRequest.sendRoap({
       roapMessage,
-      locusSelfUrl: meeting.selfUrl,
+      locusSelfUrl: this.getLocusSelfUrl(meeting),
```
marcin-bazyl: I don't think that by making sure that we use the latest url here we will fully fix the BEMS case. The process of exchanging SDP offer/answer with Homer and the signalling of breakout move from Locus happen in parallel and I don't think we can guarantee that we will always receive the update from Locus in time before we send the answer. Mercury messages can be delayed, can be missed and can come out-of-order. You might want to confirm that with Locus team, but I think we need to be prepared that in some edge cases we will get 409 from Locus and we then have to assume that we just haven't received the Locus url update yet, but it's in progress so will receive a new Locus url shortly and once we do, we need to retry the /media roap request

packages/@webex/plugin-meetings/src/roap/index.ts
Kesari3008: Updated the approach now. Based on Bartosz analysis, this was just very initial change which I couldn't verify much in detail. Went through the logs, analysis and code flow.

packages/@webex/plugin-meetings/src/roap/index.ts
Kesari3008: Updated approach now

packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
Kesari3008: Ignore the previous change, please check the new approach

packages/@webex/plugin-meetings/src/roap/index.ts
marcin-bazyl: can you explain the design for the new approach? from what I can see we're trying to detect that we're in a transition between breakouts when we have a ROAP message to send and if we are in a transition, we make a call to `getLocusDTO({url: this.locusUrl})` to get latest Locus DTO. I don't think this will work, because we're using the `this.locusUrl` - so if `this.locusUrl` is old, we will get the old Locus DTO in the response not the new one. If `this.locusUrl` is already new, then this call is not needed, because the ROAP message would have already been sent to the right Locus url with existing code.

Kesari3008: @marcin-bazyl So after the breakout session end, Meetings.handleLocusEvent() is invoked in the flow and it seems it skip updating data with locus event as we have a block stating doesn't need to process locus event.

This can cause scenarios where Locus Data is not updated hence when ROAP message is attempted, it is attempted with stale URL because locus data is yet to be updated.

With these changes, all we are trying to ensure before sending ROAP message, we refresh the locus data and also we are trying to ensure that if meeting is in breakout transition state, it will try to avoid this block where we skip updating locus data.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4013,6 +4026,49 @@ export default class Meeting extends StatelessWebexPlugin {
         this[key] = object[key];
       });
     }
+
+    // Keep selfUrl aligned with the latest parsed locus self snapshot to avoid stale values
+    // during fast session transitions (e.g. breakout/main moves).
+    const latestLocusSelfUrl = this.locusInfo?.parsedLocus?.self?.selfUrl;
```
marcin-bazyl: this block of code doesn't belong here and it doesn't do anything useful. The purpose of updateMeetingObject() is that it is a callback passed to LocusInfo and is called with updated object already containing parsed props from Locus DTO. So, `this.selfUrl` is already being updated like this:  when we get a new Locus DTO `updateSelf()` calls:
1. `parsedSelves = getSelves()` - which calls SelfUtils.parse() and returns `{current: { selfUrl: ..., ...}, ...}`
2. `this.updateMeeting(parsedSelves.current)` - updateMeeting is a callback, so actually Meeting.updateMeetingObject() gets called

packages/@webex/plugin-meetings/src/roap/index.ts
marcin-bazyl: > Meetings.handleLocusEvent() is invoked in the flow and it seems it skip updating data with locus event

The purpose of that code in handleLocusEvent() is to skip the updates for the old locus (the one we are moving from), You can talk to Web App Marmot team, who implemented most of this if you need more details

> With these changes, all we are trying to ensure before sending ROAP message, we refresh the locus data

but you're using the old locus url so you won't get any new data, you will still get the old data

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -735,6 +735,7 @@ export default class Meeting extends StatelessWebexPlugin {
   turnServerUsed: boolean;
   areVoiceaEventsSetup = false;
   isMoveToInProgress = false;
+  private isBreakoutTransitioning = false;
```
marcin-bazyl: This flag doesn't belong here. There are existing classes `Breakouts` and `Breakout`, so if we ever need such a flag, it should be in there. Maybe even those classes already contain this kind of information.

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -18,6 +18,7 @@ export type RoapRequest = {
   reachability: any;
   clientMediaPreferences: ClientMediaPreferences;
   sequence?: any;
+  __selfUrlRetryCount?: number;
```
marcin-bazyl: instead of putting this here and in LocalMuteRequest (line 33) it would be better to put the retry counter inside InternalRequestInfo - this way it would stay private within the locusMediaRequest module and you wouldn't need to use __ prefix in the name

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -264,6 +272,17 @@ export class LocusMediaRequest extends WebexPlugin {
         return result;
       })
       .catch((e) => {
+        if (
+          (e?.statusCode === 409 || e?.code === 409) &&
```
marcin-bazyl: do we really need to check `e.code` too? I see lots of existing code in SDK only checking `statusCode`, so why is a check of `code` needed here?

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +320,56 @@ export class LocusMediaRequest extends WebexPlugin {
     return promise;
   }
 
+  /**
+   * Returns the latest selfUrl for this meeting, falling back to the value
+   * captured when the request was enqueued.
+   */
+  private getCurrentSelfUrl(fallbackUrl: string): string {
+    // @ts-ignore
+    const meeting = this.webex.meetings.meetingCollection.getByKey(
```
marcin-bazyl: nitpick: I'm not a huge fan of accessing `this.webex`, because it opens up the floodgates for potentially accessing anything and in practice creates a hidden dependency and can cause circular dependencies. We could just have a callback that would be implemented in setupLocusMediaRequest() and would return the current `selfUrl` from the Meeting object. I think that would be cleaner.

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +320,56 @@ export class LocusMediaRequest extends WebexPlugin {
...
+      'correlationId',
+      this.config.correlationId
+    );
+
+    if (meeting?.selfUrl && meeting.selfUrl !== fallbackUrl) {
+      LoggerProxy.logger.info(
+        `Meeting:LocusMediaRequest#getCurrentSelfUrl --> selfUrl updated since enqueue, using ${meeting.selfUrl}`
```
marcin-bazyl: maybe it would be useful to send a metric here so we can see if this edge case is actually happening in the field?

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -264,6 +272,17 @@ export class LocusMediaRequest extends WebexPlugin {
         return result;
       })
       .catch((e) => {
+        if (
+          (e?.statusCode === 409 || e?.code === 409) &&
+          this.shouldRetryOnSelfUrlChange(request)
+        ) {
+          // In-flight race: selfUrl rotated after we left the queue but
```
marcin-bazyl, line 283: maybe it would be useful to send a metric here too

marcin-bazyl (commented):
I've only reviewed the changes in commit 343f135 - I think that approach is fine, just made some minor comments. It won't fix all the edge cases, because it just relies on a few retries instead of waiting (with a timeout) for selfUrl change, but it might be good enough as an initial fix if the customer needs something quickly. Also, changing retries to a callback to the Meeting object that would resolve once selfUrl changes shouldn't be too difficult, if we want to cover all edge cases better.

I don't like the approach in the second commit (7476073) it feels very fragile and overcomplicated, it's also hardcoded to handled only the specific case of a move from breakout to main, but the same issue can exist also during moves from main to breakout and between breakouts. Therefore I haven't reviewed it.

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Updated now

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Done

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Updated

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: added

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Added metric

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -91,6 +91,7 @@ export type Config = {
   correlationId: string;
   meetingId: string;
   preferTranscoding: boolean;
+  getCurrentSelfUrl?: () => string | undefined;
```
marcin-bazyl: I don't think this should be optional - we always need to have it, so let's remove the `?`

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -263,7 +270,34 @@ export class LocusMediaRequest extends WebexPlugin {
...
+          e?.statusCode === 409 &&
+          (await this.shouldRetryOnSelfUrlChange(request, selfUrlRetryCount))
+        ) {
+          const roapRequest = request.type === 'RoapMessage' ? request : undefined;
+
+          // @ts-ignore
+          this.webex.internal.newMetrics.submitClientEvent({
+            name: 'client.locus.media.retry',
+            options: {
+              meetingId: this.config.meetingId,
```
marcin-bazyl: having a correlationId instead of meetingId in the metric would be more useful

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +335,62 @@ export class LocusMediaRequest extends WebexPlugin {
...
+
+  /**
+   * Decides whether a 409 warrants a retry against the meeting's latest
+   * selfUrl.
+   */
+  private async shouldRetryOnSelfUrlChange(
+    request: Request,
+    selfUrlRetryCount: number
+  ): Promise<boolean> {
+    if (request.type !== 'RoapMessage') {
```
marcin-bazyl: I don't think we need this check here, I think this functionality would be useful for any request type that we send to Locus as any of them can fail this way

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -263,7 +270,34 @@ export class LocusMediaRequest extends WebexPlugin {
...
+      .catch(async (e) => {
+        if (
+          e?.statusCode === 409 &&
+          (await this.shouldRetryOnSelfUrlChange(request, selfUrlRetryCount))
+        ) {
+          const roapRequest = request.type === 'RoapMessage' ? request : undefined;
+
+          // @ts-ignore
+          this.webex.internal.newMetrics.submitClientEvent({
+            name: 'client.locus.media.retry',
```
marcin-bazyl: there is no such CA event defined in https://sqbu-github.cisco.com/WebExSquared/event-dictionary/tree/main
I think it would be easier to just send a behavioral metric

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +335,62 @@ export class LocusMediaRequest extends WebexPlugin {
...
+    const currentSelfUrl = this.config.getCurrentSelfUrl?.();
+
+    if (request.type === 'RoapMessage' && currentSelfUrl && currentSelfUrl !== request.selfUrl) {
+      LoggerProxy.logger.info(
+        `Meeting:LocusMediaRequest#getCurrentSelfUrl --> resolved updated selfUrl, using ${currentSelfUrl}`
+      );
+
+      // @ts-ignore
+      this.webex.internal.newMetrics.submitClientEvent({
+        name: 'client.locus.selfUrlUpdated',
```
marcin-bazyl: there is no such CA event defined in https://sqbu-github.cisco.com/WebExSquared/event-dictionary/tree/main so we cannot send it like this. Anyway, I think it's more useful and easier to just send a behavioral metric

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +335,62 @@ export class LocusMediaRequest extends WebexPlugin {
...
   }
 
+  /**
+   * Returns the latest selfUrl for this meeting, falling back to the value
+   * captured when the request was enqueued.
+   */
+  private getCurrentSelfUrl(request: Request): string {
+    const currentSelfUrl = this.config.getCurrentSelfUrl?.();
+
+    if (request.type === 'RoapMessage' && currentSelfUrl && currentSelfUrl !== request.selfUrl) {
```
marcin-bazyl: we shouldn't be checking for `request.type === 'RoapMessage'` here, all request types need to use the most up-to-date self url

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +335,62 @@ export class LocusMediaRequest extends WebexPlugin {
...
+   */
+  private async shouldRetryOnSelfUrlChange(
+    request: Request,
+    selfUrlRetryCount: number
+  ): Promise<boolean> {
+    if (request.type !== 'RoapMessage') {
+      return false;
+    }
+
+    if (selfUrlRetryCount >= 2) {
```
marcin-bazyl: nitpick: a const for max retries would be nice

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -73,7 +74,7 @@ import {HashTreeObject} from '../hashTree/types';
 import {isSelf} from '../hashTree/utils';
 
 import {createLocusFromHashTreeMessage, findMeetingForHashTreeMessage} from '../locus-info';
-import {LocusDTO} from '../locus-info/types';
+import {EndMeetingReason, LocusDTO} from '../locus-info/types';
```
marcin-bazyl: EndMeetingReason is not used

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -34,6 +34,7 @@ import {
   _JOIN_,
   _INCOMING_,
   LOCUS,
+  BREAKOUTS,
```
marcin-bazyl: BREAKOUTS is not used

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -91,6 +91,8 @@ export type Config = {
   correlationId: string;
   meetingId: string;
   preferTranscoding: boolean;
+  getCurrentSelfUrl?: () => string | undefined;
+  waitForSelfUrlChange?: () => Promise<void>;
```
marcin-bazyl: same here, no need for it to be optional

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -263,7 +275,27 @@ export class LocusMediaRequest extends WebexPlugin {
...
       })
-      .catch((e) => {
+      .catch(async (e) => {
+        if (
+          e?.statusCode === 409 &&
+          (await this.shouldRetryOnSelfUrlChange(request, selfUrlRetryCount))
+        ) {
+          const roapRequest = request.type === 'RoapMessage' ? request : undefined;
+
+          Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.ADD_MEDIA_RETRY, {
```
marcin-bazyl: it would be better to add a new metric specific for this, we should not be using ADD_MEDIA_RETRY here

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +333,65 @@ export class LocusMediaRequest extends WebexPlugin {
...
+   */
+  private getCurrentSelfUrl(request: Request): string {
+    const currentSelfUrl = this.config.getCurrentSelfUrl();
+
+    if (currentSelfUrl && currentSelfUrl !== request.selfUrl) {
+      LoggerProxy.logger.info(
+        `Meeting:LocusMediaRequest#getCurrentSelfUrl --> resolved updated selfUrl, using ${currentSelfUrl}`
+      );
+
+      Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.ADD_MEDIA_RETRY, {
```
marcin-bazyl: same as above

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4065,6 +4066,11 @@ export default class Meeting extends StatelessWebexPlugin {
     if (object && Object.keys(object).length) {
       Object.keys(object).forEach((key) => {
         this[key] = object[key];
+
+        if (this.promisesWaitingForPropUpdate[key]) {
+          this.promisesWaitingForPropUpdate[key].resolve();
```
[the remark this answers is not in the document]
marcin-bazyl: AI is right, we should check if the value is actually different

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +333,65 @@ export class LocusMediaRequest extends WebexPlugin {
...
+    selfUrlRetryCount: number
+  ): Promise<boolean> {
+    if (selfUrlRetryCount >= MAX_SELF_URL_RETRY_COUNT) {
+      return false;
+    }
+
+    const currentSelfUrl = this.getCurrentSelfUrl(request);
+
+    if (!currentSelfUrl || currentSelfUrl === request.selfUrl) {
+      await this.config.waitForSelfUrlChange?.();
```
marcin-bazyl: no need for `?.` here - the callback should not be optional

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +333,65 @@ export class LocusMediaRequest extends WebexPlugin {
...
+
+      const latestSelfUrl = this.getCurrentSelfUrl(request);
+
```
marcin-bazyl, line 368: maybe it would be better to have the metric here instead of the one inside the catch()? because the one in catch() will always be sent together with the metric from getCurrentSelfUrl() and here we want a metric to know if the wait from line 374 resulted in a new self url or not.

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: done

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: done

packages/@webex/plugin-meetings/src/meeting/index.ts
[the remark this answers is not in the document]
Kesari3008: done

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Added new metric

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: updated

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
Kesari3008: Updated

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -4065,9 +4065,15 @@ export default class Meeting extends StatelessWebexPlugin {
     // is not changed by any delta event
     if (object && Object.keys(object).length) {
       Object.keys(object).forEach((key) => {
+        const previousValue = this[key];
+
         this[key] = object[key];
 
-        if (this.promisesWaitingForPropUpdate[key]) {
+        const shouldResolveWaiter =
+          key !== 'selfUrl' ||
```
marcin-bazyl, line 4073: why are we doing this special treatment of selfUrl here? for any property we should resolve the promise only when the value actually changes, right?

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -371,20 +362,32 @@ export class LocusMediaRequest extends WebexPlugin {
...
-      await this.config.waitForSelfUrlChange?.();
+      await this.config.waitForSelfUrlChange();
 
       const latestSelfUrl = this.getCurrentSelfUrl(request);
 
       if (!latestSelfUrl || latestSelfUrl === request.selfUrl) {
+        Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.LOCUS_MEDIA_REQUEST_RETRY, {
+          correlation_id: this.config.correlationId,
+          reason: 'selfUrlNotChangedAfterWait',
+          retryAttempt: selfUrlRetryCount + 1,
```
marcin-bazyl, line 373: not sure if the retryAttempt value should be here with +1 here, because when we fall into this case we don't do any more retries, right?

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
```diff
@@ -301,6 +324,77 @@ export class LocusMediaRequest extends WebexPlugin {
...
+      await this.config.waitForSelfUrlChange();
+
+      const latestSelfUrl = this.getCurrentSelfUrl(request);
+
+      if (!latestSelfUrl || latestSelfUrl === request.selfUrl) {
+        Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.LOCUS_MEDIA_REQUEST_RETRY, {
+          correlation_id: this.config.correlationId,
+          reason: 'selfUrlNotChangedAfterWait',
+          retryAttempt: selfUrlRetryCount + 1,
+          roapMessageType: roapRequest?.roapMessage?.messageType,
```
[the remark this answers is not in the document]
rarajes2: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
rarajes2: Done

packages/@webex/plugin-meetings/src/meeting/locusMediaRequest.ts
rarajes2: Done

rarajes2:
Closing this PR as I have raised a new PR to address the review comments in absence of @Kesari3008 

New PR - https://github.com/webex/webex-js-sdk/pull/5081

rarajes2:
@marcin-bazyl Can you please verify the changes in this PR - https://github.com/webex/webex-js-sdk/pull/5081

PR #4919, opened 2026-04-24
fix: stop destroying meeting on TERMINATING

marcin-bazyl:
# COMPLETES #[SPARK-798051](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-798051)

After this fix for SPARK-798051 was merged:
https://github.com/webex/webex-js-sdk/pull/4892

I've realised that there may still be edge cases where it won't work, because of the exising SDK code in LocusInfo.isMeetingActive() which destroys the meeting when it receives TERMINATING state.

So the flow is:
when we leave a breakout we should receive self=LEFT with reason=MOVED or fullState=INACTIVE with endMeetingReason=BREAKOUT_ENDED - this causes SDK to ignore those events (see LocusInfo.updateLocusInfo()) and not destroy the meeting, because we know there will be more events for a new breakout coming, so we don't want to destroy that meeting.
However, before fullState=INACTIVE we receive fullState=TERMINATING which causes meeting to be destroyed - this was reported as a problem by the devices team and when I looked at our code I could see it causing us the same problem.

When testing I found that TERMINATING is actually not causing the meeting to be destroyed for us, but only because we have another workaround in `LocusInfo.updateLocusFromHashTreeObject()` where if we receive a participant for self with state=LEFT, reason=MOVED, we copy it into self and Locus happens to send us the 2 changes (the self participant LEFT, MOVED and the fullState=TERMINATING) in the same LLM event, so then the check for self LEFT in updateLocusInfo() works, the event is ignored and the LocusInfo class is not updated, so fullState is not updated to TERMINATING and so the code in LocusInfo.isMeetingActive() doesn't destroy the meeting. However, I think it's still worth fixing this and removing the code that checks for fullState==TERMINATING, because there is no guarantee that Locus will always send these 2 changes together in a single event.

TLDR: this fixes a potential case where we would destroy a meeting when being moved from a breakout, but right now this edge case is not happening (but theoretically could happen)

Locus team said in the 5k webinar space that fullState TERMINATING is not something clients should really react to, it is always followed by INACTIVE and INACTIVE is the one we should listen for.

link to discussion in the space:
webexteams://im?space=ac0ff360-8226-11f0-84cb-4fc47ab024cc&message=caff6db0-3f35-11f1-8766-e34b93ef394e

removed the check for fullState TERMINATING when deciding to destroy the meeting, added also a check endMeetingReason !== _BREAKOUT_ENDED_

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/constants.ts
+++ b/packages/@webex/plugin-meetings/src/constants.ts
@@ -103,7 +103,6 @@ export const _JOINED_ = 'JOINED';
 export const _LOCUS_ID_ = 'LOCUS_ID';
 export const _LEFT_ = 'LEFT';
 export const _MOVED_ = 'MOVED';
-export const _BREAKOUT_ENDED_ = 'BREAKOUT_ENDED';
 export const _ON_HOLD_LOBBY_ = 'ON_HOLD_LOBBY';
 export const _MEETING_LINK_ = 'MEETING_LINK';
 export const _MEETING_UUID_ = 'MEETING_UUID';
@@ -418,7 +417,7 @@ export const HEADERS = {
 // Meeting actually ended
 export const MEETING_REMOVED_REASON = {
   SELF_REMOVED: 'SELF_REMOVED', // server or host removed you from the meeting
-  MEETING_INACTIVE_TERMINATING: 'MEETING_INACTIVE_TERMINATING', // Meeting got ended or everyone left the meeting
+  MEETING_INACTIVE_TERMINATING: 'MEETING_INACTIVE_TERMINATING', // Meeting got ended or everyone left the meeting (historically was sent on Locus TERMINATING or INACTIVE, but now only on INACTIVE as that's the final state)
   CLIENT_LEAVE_REQUEST: 'CLIENT_LEAVE_REQUEST', // You triggered leave meeting
   CLIENT_LEAVE_REQUEST_TAB_CLOSED: 'CLIENT_LEAVE_REQUEST_TAB_CLOSED', // You triggered leave meeting, such as closing the browser tab directly
   USER_ENDED_SHARE_STREAMS: 'USER_ENDED_SHARE_STREAMS', // user triggered stop share
--- a/packages/@webex/plugin-meetings/src/locus-info/index.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/index.ts
@@ -19,7 +19,6 @@ import {
   RECORDING_STATE,
   Enum,
   SELF_ROLES,
-  _BREAKOUT_ENDED_,
 } from '../constants';
 
 import InfoUtils from './infoUtils';
@@ -1782,13 +1781,9 @@ export default class LocusInfo extends EventsScope {
         );
       }
     } else if (this.parsedLocus.fullState?.type === _MEETING_) {
-      if (
-        this.fullState &&
-        this.fullState.state === LOCUS.STATE.INACTIVE &&
-        this.fullState.endMeetingReason !== _BREAKOUT_ENDED_
-      ) {
+      if (this.fullState && MeetingsUtil.isWholeMeetingEnded(this.fullState)) {
         LoggerProxy.logger.warn(
-          'Locus-info:index#isMeetingActive --> Meeting is ending due to inactive or terminating'
+          'Locus-info:index#isMeetingActive --> Meeting is ending due to inactive'
         );
 
         // @ts-ignore
--- a/packages/@webex/plugin-meetings/src/locus-info/index.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/index.ts
@@ -19,6 +19,7 @@ import {
   RECORDING_STATE,
   Enum,
   SELF_ROLES,
+  _BREAKOUT_ENDED_,
 } from '../constants';
 
 import InfoUtils from './infoUtils';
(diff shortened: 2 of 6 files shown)


packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -4765,7 +4769,7 @@ describe('plugin-meetings', () => {
         });
       });
 
-      it('sends client event correctly for state = FULLSTATE_REMOVED', () => {
```
marcin-bazyl, line 4809: this pre-existing test was completely wrong and passing by luck, so I updated it

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -1783,9 +1784,8 @@ export default class LocusInfo extends EventsScope {
     } else if (this.parsedLocus.fullState?.type === _MEETING_) {
       if (
         this.fullState &&
-        (this.fullState.state === LOCUS.STATE.INACTIVE ||
-          // @ts-ignore
-          this.fullState.state === LOCUS.STATE.TERMINATING)
+        this.fullState.state === LOCUS.STATE.INACTIVE &&
```
chrisadubois: nitpick: I would explicitly factor this out into a util function so it gets clearly located in one spot and has a pure test associated with it. Not necessary.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -4743,7 +4747,7 @@ describe('plugin-meetings', () => {
         });
       });
 
-      it('sends client event correctly for state = MEETING_INACTIVE_TERMINATING', () => {
```
chrisadubois, line 4787: question: Shouldn't we also delete the constant? https://github.com/webex/webex-js-sdk/blob/4be4cb70624a5a5fd2422a1ef948f9cd263c5f7a/packages/%40webex/plugin-meetings/src/constants.ts#L421

chrisadubois: nitpick: or atleast mark that with a comment that says... it was historically also TERMINATING, but now it's just INACTIVE

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
```diff
@@ -4765,7 +4769,7 @@ describe('plugin-meetings', () => {
         });
       });
 
-      it('sends client event correctly for state = FULLSTATE_REMOVED', () => {
+      it('does not send client event when state = INACTIVE and endMeetingReason = BREAKOUT_ENDED', () => {
```
chrisadubois, line 4812: question: Just to be clear... there's no timing issue with Locus where it could *blip* one of these and THEN the other property would become consistent?

I doubt this is an issue since they are both on the `fullState` property but it popped to mind as a question

chrisadubois (approved):
Just some questions and nitpicks.

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
marcin-bazyl: no, we don't have to worry about that, Locus will always set the endMeetingReason together with the state=INACTIVE

packages/@webex/plugin-meetings/test/unit/spec/locus-info/index.js
marcin-bazyl: this constant is passed as the reason and eventually gets propagated into the "meeting:removed" event as reason, so it's a public API. I'm worried that changing it could break some existing SDK clients. I can add a comment

packages/@webex/plugin-meetings/src/locus-info/index.ts
marcin-bazyl: done

packages/@webex/plugin-meetings/src/locus-info/types.ts
```diff
@@ -1,15 +1,33 @@
+import {Enum} from '../constants';
 import {HtMeta} from '../hashTree/types';
 
+export const EndMeetingReason = {
```
marcin-bazyl, line 4: I've taken this list from Locus code: https://sqbu-github.cisco.com/WebExSquared/locus/blob/cf4f1578a9e4690a9f4c4c72e803a22c99b14b51/common/src/main/java/com/cisco/wx2/dto/locus/EndMeetingRequest.java#L13

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -103,7 +103,6 @@ export const _JOINED_ = 'JOINED';
 export const _LOCUS_ID_ = 'LOCUS_ID';
 export const _LEFT_ = 'LEFT';
 export const _MOVED_ = 'MOVED';
-export const _BREAKOUT_ENDED_ = 'BREAKOUT_ENDED';
 export const _ON_HOLD_LOBBY_ = 'ON_HOLD_LOBBY';
```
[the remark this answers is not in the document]
marcin-bazyl: _BREAKOUT_ENDED_ has only been added very recently (like last week), web app is not using it and it's very unlikely that any other client would be using it

packages/@webex/plugin-meetings/src/locus-info/types.ts
chrisadubois: stuff like this makes me wish we had a types generator like we do for event-dictionary

mickelr (approved):

PR #4921, opened 2026-04-27
fix(meetings): joinWithMedia retry logic

maxinteger:
# COMPLETES 
- SPARK-804808
- SPARK-804811

- Meeting.joinWithMedia() returns the wrong error sometimes
- The retry in Meeting.joinWithMedia() sometimes doesn't work correctly

In joinWithMedia
- retry calls join when the last error was UserNotJoinedError
- retry preserve the previous error and report both when it fails

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5660,7 +5660,7 @@ export default class Meeting extends StatelessWebexPlugin {
 
       const mediaResponse = await this.addMediaInternal(
         () => {
-          return this.joinWithMediaRetryInfo.retryCount === JOIN_WITH_MEDIA_RETRY_MAX_COUNT
+          return this.joinWithMediaRetryInfo.retryCount >= 1
             ? 'JOIN_MEETING_FINAL'
             : 'JOIN_MEETING_RETRY';
         },
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5712,8 +5712,8 @@ export default class Meeting extends StatelessWebexPlugin {
         });
       }
 
-      // we only want to call leave if join was successful and this was a retry or we won't be doing any more retries
-      if (joined && (retryCount > 0 || !shouldRetry)) {
+      // we only want to call leave if join was successful, and we won't be doing any more retries
+      if (joined && !shouldRetry) {
         try {
           await this.leave({resourceId: joinOptions?.resourceId, reason: 'joinWithMedia failure'});
         } catch (e) {
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -5576,7 +5576,7 @@ export default class Meeting extends StatelessWebexPlugin {
     } = {}
   ) {
     const {mediaOptions, joinOptions = {}} = options;
-    const {isRetry, prevJoinResponse, prevError} = this.joinWithMediaRetryInfo;
+    const {isRetry, prevJoinResponse} = this.joinWithMediaRetryInfo;
 
     if (!mediaOptions?.allowMediaInLobby) {
       return Promise.reject(
@@ -5602,7 +5602,7 @@ export default class Meeting extends StatelessWebexPlugin {
       );
     }
 
-    const shouldJoin = !joinResponse || (prevError && prevError instanceof UserNotJoinedError);
+    const shouldJoin = !joinResponse || joinResponse instanceof Error;
 
     try {
       let turnServerInfo;
@@ -5731,15 +5731,15 @@ export default class Meeting extends StatelessWebexPlugin {
         return this.joinWithMedia(options);
       }
 
-      const previousError = this.joinWithMediaRetryInfo?.prevError;
+      const {prevError} = this.joinWithMediaRetryInfo;
 
-      this.joinWithMediaRetryInfo = {isRetry: false, prevJoinResponse: undefined};
+      this.joinWithMediaRetryInfo = {
+        isRetry: false,
+        prevJoinResponse: undefined,
+        prevError: undefined,
+      };
 
-      if (isRetry && previousError) {
-        error.message = `First attempt error: ${previousError.message}\n\nSecond attempt error: ${error.message}`;
-      }
-
-      throw error;
+      throw prevError ?? error;
     }
   }
 
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -795,7 +795,7 @@ export default class Meeting extends StatelessWebexPlugin {
   private deferSDPAnswer?: Defer; // used for waiting for a response
   private sdpResponseTimer?: ReturnType<typeof setTimeout>;
   private hasMediaConnectionConnectedAtLeastOnce: boolean;
-  private joinWithMediaRetryInfo?: {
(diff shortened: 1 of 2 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5608,12 +5610,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        this.joinWithMediaRetryInfo.prevError = error;
 
         return this.joinWithMedia(options);
       }
 
+      const previousError = this.joinWithMediaRetryInfo?.prevError;
+
       this.joinWithMediaRetryInfo = {isRetry: false, prevJoinResponse: undefined};
 
+      if (isRetry && previousError) {
```
marcin-bazyl: hmm not sure about this.... we're still throwing the old error with a different message. I think Cantina has code checking for error being instance of specific class or checking other props of the error like `code` etc. I think we would want to throw the `previousError` and maybe only log the 2nd  error

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5486,12 +5486,14 @@ export default class Meeting extends StatelessWebexPlugin {
       );
     }
 
+    const shouldJoin = !joinResponse || (prevError && prevError instanceof UserNotJoinedError);
```
marcin-bazyl: I don't think this will work as you intend it unless I'm missing something. We do only 1 retry and we want to do a join if adding media fails on the retry, so in that case prevError is not going to be set to UserNotJoinedError, it will be set to some other error that caused the retry. Only on the retry we will get the UserNotJoinedError.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5608,13 +5610,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+      const {prevError} = this.joinWithMediaRetryInfo;
 
-      throw error;
+      this.joinWithMediaRetryInfo = {
+        isRetry: false,
+        prevJoinResponse: undefined,
+        prevError: undefined,
+      };
+
+      throw prevError ?? error;
```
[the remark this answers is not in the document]
maxinteger: works as expected

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5648,14 +5660,16 @@ export default class Meeting extends StatelessWebexPlugin {
 
       const mediaResponse = await this.addMediaInternal(
         () => {
-          return this.joinWithMediaRetryInfo.isRetry ? 'JOIN_MEETING_FINAL' : 'JOIN_MEETING_RETRY';
+          return this.joinWithMediaRetryInfo.retryCount >= 1
```
marcin-bazyl, line 5664: this is a callback that's used only in case of failure, so I think the intention is to send JOIN_MEETING_FINAL only if this is the final attempt and there won't be any more retries and this code will send JOIN_MEETING_FINAL twice if we do 2 retries.

EDIT: ok, after looking at the new tests from this PR I see that in the case where we do 2 retries (and that's only possible if first media error is not UserNotJoinedError and 2nd one is UserNotJoinedError) the callback is not called when UserNotJoinedError is thrown, so it's OK, but maybe worth adding some comment to explain this?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5722,15 +5738,26 @@ export default class Meeting extends StatelessWebexPlugin {
 
       if (shouldRetry) {
         LoggerProxy.logger.warn('Meeting:index#joinWithMedia --> retrying call to joinWithMedia');
-        this.joinWithMediaRetryInfo.isRetry = true;
+        this.joinWithMediaRetryInfo.retryCount = retryCount + 1;
```
marcin-bazyl, line 5742: there are warnings about `this.joinWithMediaRetryInfo` being possibly undefined, this is not strictly related to this PR, but I think it's an easy fix - just remove the `?` in the type declaration, because `joinWithMediaRetryInfo` is always defined, right?

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1445,6 +1598,101 @@ describe('plugin-meetings', () => {
...
+            .rejects(userNotJoinedError)
+            .onSecondCall()
+            .resolves(test4);
+
+          const result = await meeting.joinWithMedia({joinOptions, mediaOptions});
+
+          assert.deepEqual(result, {join: fakeJoinResult, media: test4, multistreamEnabled: true});
+
+          // join() should be called twice — once for the first attempt, once for the re-join after UserNotJoinedError
+          assert.calledTwice(meeting.join);
```
marcin-bazyl, line 1723: would be useful to also check that leave() is not called

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1445,6 +1598,101 @@ describe('plugin-meetings', () => {
...
+            .rejects(addMediaError)
+            .onSecondCall()
+            .resolves(test4);
+
+          const result = await meeting.joinWithMedia({joinOptions, mediaOptions});
+
+          assert.deepEqual(result, {join: fakeJoinResult, media: test4, multistreamEnabled: true});
+
+          // join() should be called twice — once for the first attempt, once because the user is in left state
+          assert.calledTwice(meeting.join);
```
marcin-bazyl, line 1749: same as above, would be useful to check that leave() is not called

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5696,7 +5713,7 @@ export default class Meeting extends StatelessWebexPlugin {
       }
 
       // we only want to call leave if join was successful and this was a retry or we won't be doing any more retries
-      if (joined && (isRetry || !shouldRetry)) {
+      if (joined && (retryCount > 0 || !shouldRetry)) {
```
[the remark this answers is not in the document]
marcin-bazyl: I think AI is right, the conditions for calling leave() need to be changed and also we should first modify the tests to make sure that a test catches this case of an unnecessary leave() call between 1st and 2nd retry.

packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -1188,6 +1261,85 @@ describe('plugin-meetings', () => {
...
+          const secondError = new Error('second attempt error');
+
+          meeting.addMediaInternal = sinon
+            .stub()
+            .onFirstCall()
+            .rejects(firstError)
+            .onSecondCall()
+            .rejects(secondError);
+
+          sinon.stub(meeting, 'leave').resolves();
```
marcin-bazyl: this is suspicious - why are we stubbing leave() here and not checking that leave was not called in this test (same in the next test) - this is probably related to the AI comment about the conditions for the call to leave()

maxinteger: Stub is necessary to check how many times the function called. I fixed the P2 comment and I refactored the tests to make sure `leave` called only after the last re-try

marcin-bazyl (approved):

jor-row (approved):

PR #4923, opened 2026-04-27
feat(calling): auto-fetch RTMS domain from service catalog

Kesari3008:
# COMPLETES #AD-HOC

Adding new method to fetch RTMS domain from catalog

Adding new method to fetch RTMS domain from catalog

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Screenshot 2026-04-28 at 9 03 39 PM

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [x] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/CallingClient.test.ts
+++ b/packages/calling/src/CallingClient/CallingClient.test.ts
@@ -17,7 +17,7 @@ import {
 /* eslint-disable dot-notation */
 import {CALLING_CLIENT_EVENT_KEYS, CallSessionEvent, MOBIUS_EVENT_KEYS} from '../Events/types';
 import log from '../Logger';
-import {createClient} from './CallingClient';
+import {CallingClient, createClient} from './CallingClient';
 import {ICallingClient} from './types';
 import * as utils from '../common/Utils';
 import {getCallManager} from './calling/callManager';
@@ -212,19 +212,16 @@ describe('CallingClient Tests', () => {
       }).not.toThrow(Error);
     });
 
-    /**
-     * Input sdk config to callingClient with serviceData carrying valid value for indicator
-     * 'contactcenter', but an empty string for domain field in it.
-     *
-     * It should throw error and abort execution as domain value is invalid.
-     *
-     * DOMAIN field for service type 'contactcenter' must carry a non-empty valid domain type string.
-     */
-    it('ContactCenter: verify empty domain is allowed (auto-fetch from catalog)', async () => {
-      const serviceDataObj = {indicator: ServiceIndicator.CONTACT_CENTER, domain: ''};
+    it('ContactCenter: uses config domain and does not fetch RTMS domain from catalog', async () => {
+      const serviceDataObj = {
+        indicator: ServiceIndicator.CONTACT_CENTER,
+        domain: 'test.example.com',
+      };
 
+      webex.internal.services.get = jest.fn();
       callingClient = await createClient(webex, {serviceData: serviceDataObj});
       expect(callingClient).toBeTruthy();
+      expect(webex.internal.services.get).not.toHaveBeenCalled();
     });
 
     /**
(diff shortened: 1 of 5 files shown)


akulakum (approved):
Thanks for the change — auto-filling CC RTMS domain from U2C is useful. A few things to align on before merge:
(1) Confirm post-getMobiusServers() we always have catalog data for wcc-calling-rtms-domain, or add a bounded wait / document fail-fast behavior.
(2) getCallManager is a process singleton; deferring validateServiceData until init() means an invalid first indicator could cache bad state — consider validating non-catalog fields before constructing managers or handling failed init.
(3) Extend tests so the catalog success path is covered without mocking validateServiceData.
(4) Minor: align getRTMSDomain return type with actual returns.

Kesari3008:
> Thanks for the change — auto-filling CC RTMS domain from U2C is useful. A few things to align on before merge: (1) Confirm post-getMobiusServers() we always have catalog data for wcc-calling-rtms-domain, or add a bounded wait / document fail-fast behavior. (2) getCallManager is a process singleton; deferring validateServiceData until init() means an invalid first indicator could cache bad state — consider validating non-catalog fields before constructing managers or handling failed init. (3) Extend tests so the catalog success path is covered without mocking validateServiceData. (4) Minor: align getRTMSDomain return type with actual returns.

Addressed the return type and validating service data before singleton instance creations

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -202,11 +207,60 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+      this.serviceData.indicator === ServiceIndicator.CONTACT_CENTER &&
+      !this.serviceData.domain
+    ) {
+      const rtmsDomain = this.getRTMSDomain();
+
+      this.serviceData.domain = rtmsDomain;
+      if (this.sdkConfig?.serviceData) {
+        this.sdkConfig.serviceData.domain = this.serviceData.domain;
+      }
+    }
```
bhabalan, line 222: If `getRTMSDomain()` returns `''` (catalog down, bad URL, etc.), we mutate `sdkConfig.serviceData.domain` to empty string here before the validation check on line 224 throws. That leaves the caller's config object in a dirty state even though init failed.

Can we move the sdkConfig sync to after validation passes?

```suggestion
      this.serviceData.domain = rtmsDomain;
    }

    if (!isValidServiceDomain(this.serviceData)) {
      throw new Error('Invalid service domain.');
    }

    if (this.sdkConfig?.serviceData) {
      this.sdkConfig.serviceData.domain = this.serviceData.domain;
    }
```

packages/calling/src/common/Utils.ts
```diff
@@ -1607,10 +1607,6 @@ export function validateServiceData(serviceData: ServiceData) {
   if (!isValidServiceIndicator(serviceData.indicator)) {
     throw new Error(`Invalid service indicator, Allowed values are: ${formattedValues}`);
   }
```
bhabalan, line 1609: Now that domain validation is gone from here, the JSDoc above (line 1596) still says this validates "indicator & domain". Worth a quick update so the next person reading this doesn't assume domain checking still happens in this function.

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -239,16 +232,37 @@ describe('CallingClient Tests', () => {
...
-    it('ContactCenter: verify valid service domain', async () => {
-      const serviceDataObj = {
-        indicator: ServiceIndicator.CONTACT_CENTER,
-        domain: 'test.example.com',
-      };
+    it('ContactCenter: fetches RTMS domain from catalog when config domain is empty', async () => {
+      const serviceDataObj = {indicator: ServiceIndicator.CONTACT_CENTER, domain: ''};
+      const validateServiceDataSpy = jest
+        .spyOn(utils, 'validateServiceData')
+        .mockImplementation(() => undefined);
```
bhabalan: Is the `validateServiceData` mock still needed? Since domain validation was removed from that function, the constructor should pass fine without it. Keeping the mock means we're not actually testing the real constructor path here.

```suggestion
    it('ContactCenter: fetches RTMS domain from catalog when config domain is empty', async () => {
      const serviceDataObj = {indicator: ServiceIndicator.CONTACT_CENTER, domain: ''};

```

packages/calling/src/CallingClient/CallingClient.test.ts
```diff
@@ -239,16 +232,37 @@ describe('CallingClient Tests', () => {
      *
      * DOMAIN field for service type 'contactcenter' must carry a non-empty valid domain type string.
```
bhabalan: This comment block still says "DOMAIN field for service type 'contactcenter' must carry a non-empty valid domain type string" but the test below now validates the opposite (empty domain triggers catalog fetch). Should update to match the new behavior.

packages/calling/src/CallingClient/CallingClient.ts
Kesari3008: We only fetch the domain from the catalog when there is no value for domain is passed in the config. In that case this.serviceData and thisSdkConfig.serviceData both will have empty string which would be the case even when catalog fetch fails and both values will get updated with empty string only

packages/calling/src/common/Utils.ts
Kesari3008: Removed the domain

packages/calling/src/CallingClient/CallingClient.test.ts
Kesari3008: Removed it

packages/calling/src/CallingClient/CallingClient.test.ts
Kesari3008: Updated

bharath063 (approved):

bhabalan (approved):

PR #4926, opened 2026-04-27
feat(plugin-meetings): service app recording fields 

Coread:

Two new fields are being added to the locus DTO for recording:
* modifiedByServiceAppName - the name of the service app that started/stopped/paused the recording
* modifiedByServiceAppId - the id of the service app that started/stopped/paused the recording

Added the two new fields to the existing meta data parsing

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
@@ -27,6 +27,8 @@ ControlsUtils.parse = (controls: any) => {
       paused: controls.record.paused ? controls.record.paused : false,
       recording: controls.record.recording,
       lastModified: controls.record.meta.lastModified,
+      modifiedByServiceAppName: controls.record.meta?.modifiedByServiceAppName || undefined,
+      modifiedByServiceAppId: controls.record.meta?.modifiedByServiceAppId || undefined,
     };
   }
 
--- a/packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
+++ b/packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
@@ -26,9 +26,9 @@ ControlsUtils.parse = (controls: any) => {
       modifiedBy: ControlsUtils.getId(controls),
       paused: controls.record.paused ? controls.record.paused : false,
       recording: controls.record.recording,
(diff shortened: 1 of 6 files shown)


packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
```diff
@@ -26,7 +26,9 @@ ControlsUtils.parse = (controls: any) => {
       modifiedBy: ControlsUtils.getId(controls),
       paused: controls.record.paused ? controls.record.paused : false,
       recording: controls.record.recording,
-      lastModified: controls.record.meta.lastModified,
+      lastModified: controls.record.meta?.lastModified,
```
jor-row, line 29: Is the addition of the  ```?``` needed here?

jor-row: Will locus ever give us any of these as undefined?

jor-row: I can see you've added tests for this case so I imagine it's possible

jor-row (approved):

packages/@webex/plugin-meetings/src/locus-info/controlsUtils.ts
Coread: I was just being cautious with the additions and then added it for consistency. I think it is always there, but it doesn't hurt

chrisadubois: not a blocker: We never had it before, I don't think it's needed but I don't think it's a problem to have. Hypothetically doesn't that mean we should add it in all of the other places? 

It's not a blocker but I feel like we should know if locus is going to ever not send us this otherwise we might end up with hard to track errors if this is missing but modifiedByServiceAppName is present?

This is not the only place where these are mismatched, so this is not a blocker

chrisadubois (approved):

PR #4930, opened 2026-04-29
feat(calling): add registration down event handling

rarajes2:
# COMPLETES #< [CAI-7866](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7866) >

Handles `registration.down` event.

- **New `Registration.handleRegistrationDownEvent(event?)`** (`register.ts`, `types.ts`):
  - If any active calls exist, sets `registrationDownPending = true` and returns early (defers cleanup).
  - Otherwise, runs cleanup immediately.
  - Safe to re-invoke: idempotent across repeated calls.
- **New `Registration.isRegistrationDownPending()`**: exposes the pending flag so `CallingClient` can poll it.
- **New private `performRegistrationDownCleanup(caller)`** (runs under the shared mutex):
  - Stops `failbackTimer` and keepalive web worker.
  - Resets transient flags: `reconnectPending`, `scheduled429Retry`, `failoverImmediately`, `retryAfter`, `registerRetry`.
  - Clears the localStorage failover cache and sets status to `INACTIVE`.
  - If Mobius WebSocket is enabled, disconnects with code `3050` / reason `'done (permanent)'` (errors are logged but don’t block cleanup).
  - Emits `LINE_EVENTS.UNREGISTERED` to the SDK consumer.
  - Clears `registrationDownPending`.
  - No `DELETE /devices/{id}` — Mobius already signaled registration is gone.
- **`CallingClient.handleMobiusAsyncEvent`** (`CallingClient.ts`):
  - `MobiusEventType.REGISTRATION_DOWN` now invokes `line.registration.handleRegistrationDownEvent(event)` (replacing the prior TODO stub).
  - If the call leaves `isRegistrationDownPending() === true`, schedules a 30-second polling interval (`registrationDownInterval`) that re-invokes `handleRegistrationDownEvent` with the same event and self-clears once the pending flag goes `false`.
  - Prior interval (if any) is cleared on each new `REGISTRATION_DOWN` event to avoid stacking.
- **Events/metrics**: no new events or metrics; continues to emit existing `LINE_EVENTS.UNREGISTERED` and the pre-existing `MOBIUS_SOCKET_ERROR` / `REGISTRATION_DOWN` metric.
- **Docs + tests**: `registration/ai-docs/AGENTS.md` documents the polling protocol; 5 new unit tests in `register.test.ts` cover immediate cleanup, deferred cleanup, re-invocation after calls clear, socket-enabled teardown, and graceful handling of socket-disconnect failure.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Unit Tests. Further verification will be done when Mobius tests this by explicitly sending `registration.down` event.

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/calling/src/CallingClient/CallingClient.ts
+++ b/packages/calling/src/CallingClient/CallingClient.ts
@@ -120,6 +120,14 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
   private mercuryUpTimestamp = '';
 
   /**
+   * Polling interval used to retry a deferred `REGISTRATION_DOWN` cleanup.
+   * Set when `Registration.handleRegistrationDownEvent` defers cleanup because
+   * active calls are present; cleared once the registration-down cleanup
+   * completes (`Registration.isRegistrationDownPending()` returns `false`).
+   */
+  private registrationDownInterval?: NodeJS.Timeout;
+
+  /**
    * @ignore
    */
   constructor(webex: WebexSDK, config?: CallingClientConfig) {
@@ -743,14 +751,23 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
     }
 
     if (eventType === MobiusEventType.REGISTRATION_DOWN) {
-      // const line = Object.values(this.lineDict)[0];
-      // line.registration.handleRegistrationDownEvent(event);
+      log.warn(`Received ${eventType} event from Mobius.`, loggerContext);
+      const line = Object.values(this.lineDict)[0];
 
-      // TODO: check active calls and clean up the state and emit disconnect event
-      log.warn(
-        'Received REGISTRATION_DOWN event from Mobius; teardown handling pending (TODO)',
-        loggerContext
-      );
+      await line.registration.handleRegistrationDownEvent(event);
+      clearInterval(this.registrationDownInterval);
+      this.registrationDownInterval = undefined;
+
+      if (line.registration.isRegistrationDownPending()) {
+        this.registrationDownInterval = setInterval(async () => {
+          await line.registration.handleRegistrationDownEvent(event);
+
+          if (!line.registration.isRegistrationDownPending()) {
+            clearInterval(this.registrationDownInterval);
+            this.registrationDownInterval = undefined;
+          }
+        }, 1000 * 30); // 30 seconds
+      }
 
       this.metricManager.submitMobiusSocketMetric(
         METRIC_EVENT.MOBIUS_SOCKET_ERROR,
--- a/packages/calling/src/CallingClient/CallingClient.ts
+++ b/packages/calling/src/CallingClient/CallingClient.ts
@@ -754,6 +754,12 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
       log.warn(`Received ${eventType} event from Mobius.`, loggerContext);
       const line = Object.values(this.lineDict)[0];
 
+      if (!line) {
+        log.warn('No line found, skipping registration down event', loggerContext);
+
+        return;
+      }
+
       await line.registration.handleRegistrationDownEvent(event);
       clearInterval(this.registrationDownInterval);
       this.registrationDownInterval = undefined;
--- a/packages/calling/src/CallingClient/CallingClient.ts
+++ b/packages/calling/src/CallingClient/CallingClient.ts
@@ -120,14 +120,6 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements 
   private mercuryUpTimestamp = '';
 
   /**
-   * Polling interval used to retry a deferred `REGISTRATION_DOWN` cleanup.
-   * Set when `Registration.handleRegistrationDownEvent` defers cleanup because
-   * active calls are present; cleared once the registration-down cleanup
-   * completes (`Registration.isRegistrationDownPending()` returns `false`).
-   */
-  private registrationDownInterval?: NodeJS.Timeout;
-
-  /**
    * @ignore
    */
   constructor(webex: WebexSDK, config?: CallingClientConfig) {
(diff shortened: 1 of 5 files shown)


packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -743,14 +751,23 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
-      // line.registration.handleRegistrationDownEvent(event);
+      log.warn(`Received ${eventType} event from Mobius.`, loggerContext);
+      const line = Object.values(this.lineDict)[0];
 
-      // TODO: check active calls and clean up the state and emit disconnect event
-      log.warn(
-        'Received REGISTRATION_DOWN event from Mobius; teardown handling pending (TODO)',
-        loggerContext
-      );
+      await line.registration.handleRegistrationDownEvent(event);
```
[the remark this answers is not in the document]
rarajes2: Added null check

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1339,6 +1340,110 @@ export class Registration implements IRegistration {
...
+    };
+
+    log.info(
+      `Registration down received - trackingId: ${event?.trackingId ?? 'unknown'}, eventId: ${
+        event?.eventId ?? 'unknown'
+      }`,
+      loggerContext
+    );
+
+    if (Object.keys(this.callManager.getActiveCalls()).length > 0) {
```
Kesari3008: As per my understanding of this event, if we receive this event from Mobius that means device has been marked inactive. I don't think Mobius will maintain any active call for such scenario so this block is not needed. Also they have mentioned SIP Register failing would trigger sending this event so in that case SIP Call Signalling won't work anyway. Did we check with Mobius that can there be calls up an running when we receive this event ?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1339,6 +1340,110 @@ export class Registration implements IRegistration {
...
+   * Stops timers, resets transient flags, clears failover cache, sets status to
+   * INACTIVE, tears down the Mobius WebSocket (when enabled), and finally emits
+   * `LINE_EVENTS.UNREGISTERED` so the SDK consumer is notified.
+   *
+   * Runs under the shared mutex to avoid racing with other registration flows.
+   * Idempotent: `registrationDownPending` is reset before emitting.
+   *
+   * @param caller - Identifier of the caller, used for logs.
+   */
+  private async performRegistrationDownCleanup(caller: string): Promise<void> {
```
Kesari3008, line 1379: This elaborate handling is not required, we should just do straightforward cleanup and notify user with UNREGISTERED event

packages/calling/src/CallingClient/CallingClient.ts
```diff
@@ -743,14 +751,29 @@ export class CallingClient extends Eventing<CallingClientEventTypes> implements
...
+        log.warn('No line found, skipping registration down event', loggerContext);
+
+        return;
+      }
+
+      await line.registration.handleRegistrationDownEvent(event);
+      clearInterval(this.registrationDownInterval);
+      this.registrationDownInterval = undefined;
+
+      if (line.registration.isRegistrationDownPending()) {
```
Kesari3008: This won't be needed too as per previous comment

packages/calling/src/CallingClient/registration/register.ts
rarajes2: This was implemented after discussing with Mobius team. Please check the message in space

packages/calling/src/CallingClient/registration/register.ts
rarajes2: Same as before. This is handling the worst case scenario.

packages/calling/src/CallingClient/CallingClient.ts
rarajes2: Same here

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: I still feel this is a lot of code just for a defensive check. And even if we really wanna rely on defensive check then also using call keepalive with Mobius directly and getting the answer right away to know if there is any call or not would be better and quicker way to deal with this. This check is local to the SDK which may contain stale calls at times and as a result of that we are setting up an interval and running it repeatedly just for a simple  registration cleanup. This also causes delay for the consumers where their device is inactive and they won't be notified to attempt re-registration.

rsarika (approved):

packages/calling/src/CallingClient/registration/register.ts
rarajes2: **Fixed** --- Discussed again with Mobius team about the active call check. We can now end/clear the call state if any and move on the clearing the registration state and emitting the `unregistered` event.

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1339,6 +1339,82 @@ export class Registration implements IRegistration {
...
+
+    log.info(
+      `Registration down received - trackingId: ${event?.trackingId ?? 'unknown'}, eventId: ${
+        event?.eventId ?? 'unknown'
+      }`,
+      loggerContext
+    );
+
+    const [activeCall] = Object.values(this.callManager.getActiveCalls());
+    activeCall?.end();
```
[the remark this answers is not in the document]
rarajes2: There is only one call at a time for now. This is fine

PR #4941, opened 2026-04-30
fix(data-channel): connect failure with multiple meetings

Tianhui-Han:

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-809708

When multiple meetings coexist, the shared refresh handler gets overwritten and PS token refresh targets the wrong meeting (401/403). 

Fix: interceptor now resolves the owning meeting from the request URL at refresh time, falling back to the singleton handler if unresolved.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [x] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -45,6 +45,7 @@ export const LOCI = 'loci';
 export const LOCUS_URL = 'locusUrl';
 export const END = 'end';
 export const LLM_PRACTICE_SESSION = 'llm-practice-session';
+export const LLM_DEFAULT_SESSION = 'llm-default-session';
```
marcin-bazyl: we already have LLM_DEFAULT_SESSION defined here: https://github.com/webex/webex-js-sdk/blob/next/packages/%40webex/internal-plugin-llm/src/constants.ts#L4
it would be better to use that instead of creating a copy

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -42,11 +43,56 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+            // @ts-ignore
+            this.meetings?.getMeetingByType?.(LOCUS_URL, sessionLocusUrl)) ||
+          undefined;
+
+        let result;
+        if (meeting) {
+          result = await meeting.refreshDataChannelToken();
+        } else {
+          // @ts-ignore
+          result = await this.internal.llm.refreshDataChannelToken();
```
marcin-bazyl: looking at this code makes me realize that existing LLM module APIs are a bit flawed.... LLM plugin's refreshDataChannelToken() and setDatachannelToken() methods should take session ids as arguments, shouldn't they?

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -42,11 +43,56 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+      // matching against each session's stored datachannelUrl. Then looks up the
+      // session's tracked locusUrl and finds the matching Meeting in the
+      // meetings collection. Falls back to the LLM plugin's singleton handler
+      // when the lookup cannot resolve a Meeting (preserves prior behavior).
+      refreshDataChannelToken: async (requestUrl?: string) => {
+        let sessionId: string | undefined;
+
+        if (typeof requestUrl === 'string') {
+          // @ts-ignore
+          const psDatachannelUrl = this.internal.llm.getDatachannelUrl?.(LLM_PRACTICE_SESSION);
```
marcin-bazyl: this code will stop working when any meeting object creates an LLM session with a non-default session id or if we ever have more than 1 active LLM connection. Wouldn't it be better to get the list of all LLM sessions and among them look for the locus url? or maybe even don't bother with checking the llm sessions and just look up the meeting by locus url and call `meeting.refreshDataChannelToken()` on it?

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
Tianhui-Han: done

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
Tianhui-Han: done

packages/@webex/plugin-meetings/src/constants.ts
Tianhui-Han: done

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -233,26 +247,33 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
+    tokenKey: DataChannelTokenType | string = DataChannelTokenType.Default
   ): void => {
-    this.datachannelTokens[dataChannelTokenType] = datachannelToken;
+    this.datachannelTokens[tokenKey] = datachannelToken;
+
+    const sessionData = this.connections.get(tokenKey);
+
+    if (sessionData) {
+      sessionData.datachannelToken = datachannelToken;
+      this.connections.set(tokenKey, sessionData);
```
marcin-bazyl: I don't think this set() call is needed - the get() returns a reference to the object, so when you modify it, it's already modified in the map.
see this js console example for proof:

Image

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -233,26 +247,33 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
-    dataChannelTokenType: DataChannelTokenType = DataChannelTokenType.Default
+    tokenKey: DataChannelTokenType | string = DataChannelTokenType.Default
   ): void => {
-    this.datachannelTokens[dataChannelTokenType] = datachannelToken;
+    this.datachannelTokens[tokenKey] = datachannelToken;
+
+    const sessionData = this.connections.get(tokenKey);
+
+    if (sessionData) {
+      sessionData.datachannelToken = datachannelToken;
```
marcin-bazyl: I'm a bit confused, why do we store datachannelToken in 2 places in `this.datachannelTokens` and in sessionData inside `this.connections`?

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -66,14 +79,15 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
+  private datachannelTokens: Record<string, string> = {
     [DataChannelTokenType.Default]: undefined,
     [DataChannelTokenType.PracticeSession]: undefined,
   };
 
-  private refreshHandler?: () => Promise<{
-    body: {datachannelToken: string; datachannelTokenType: DataChannelTokenType};
-  }>;
+  private refreshHandlers: Map<
+    string,
```
marcin-bazyl: would be nice to have a comment explaining what this string is. I guess it's the sessionId, right? in that case, do we actually need this separate `refreshHandlers` map? wouldn't it be better to store the refreshHander inside the session data in `this.connections`?

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -12,6 +12,19 @@ import {
 } from './constants';
 import {ILLMChannel, DataChannelTokenType} from './llm.types';
 
+const getRefreshHandlersMap = (
```
marcin-bazyl: why is this not a method on the LLMChannel class? Also I don't think this is needed at all, see my other comments

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3471,7 +3472,10 @@ export default class Meeting extends StatelessWebexPlugin {
         this.controlsOptionsManager.setLocusUrl(this.locusUrl, !!isMainLocus);
         this.webinar.locusUrlUpdate(url);
         // @ts-ignore
-        this.webex.internal.llm.setRefreshHandler(() => this.refreshDataChannelToken());
+        this.webex.internal.llm.setRefreshHandler(
+          () => this.refreshDataChannelToken(),
+          LLM_DEFAULT_SESSION
```
marcin-bazyl: this doesn't look right. Why is LLM_DEFAULT_SESSION hardcoded here? what if we are in a practice session? will this work correctly then?
Also if we have multiple meeting objects, then one meeting object can override the refresh handler from another meeting here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6526,14 +6530,14 @@ export default class Meeting extends StatelessWebexPlugin {
 
     if (datachannelToken) {
       // @ts-ignore
-      this.webex.internal.llm.setDatachannelToken(datachannelToken, DataChannelTokenType.Default);
+      this.webex.internal.llm.setDatachannelToken(datachannelToken, LLM_DEFAULT_SESSION);
```
marcin-bazyl: similar to previous comment - this won't work when we have multiple meeting objects. We can probably get away with it for now, because we don't join more than 1 meeting in parallel and so we don't have more than 1 llm active, but overall feels like this whole design of how data channel tokens are handled is flawed.

packages/@webex/plugin-meetings/src/constants.ts
```diff
@@ -45,6 +45,7 @@ export const LOCI = 'loci';
 export const LOCUS_URL = 'locusUrl';
 export const END = 'end';
 export const LLM_PRACTICE_SESSION = 'llm-practice-session';
+export {LLM_DEFAULT_SESSION} from '@webex/internal-plugin-llm';
```
marcin-bazyl: shouldn't the same be done for LLM_PRACTICE_SESSION from line above?

packages/@webex/internal-plugin-llm/src/llm.ts
Tianhui-Han: done

packages/@webex/internal-plugin-llm/src/llm.ts
Tianhui-Han: We don’t store datachannelToken in two places anymore.

It now lives only in datachannelTokens; connections/sessionData keeps only socket metadata + owner/refreshHandler.

packages/@webex/internal-plugin-llm/src/llm.ts
Tianhui-Han: Done

packages/@webex/internal-plugin-llm/src/llm.ts
Tianhui-Han: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: LLM_DEFAULT_SESSION is only used for the main/default meeting data channel path.
Practice session uses LLM_PRACTICE_SESSION in the webinar practice-session path, so refresh/token routing is session-specific and still correct.

The cross-meeting override risk is now guarded.
We now claim ownership immediately when setting each session’s refresh handler, and all later updates check ownership before rewriting/disconnecting.
So one meeting should no longer override another meeting’s refresh handler for the same session.

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: You’re right, and we already added mitigations for this in this PR: owner-aware guards for refresh handler/token writes, explicit session/token keys, and token cache decoupled from connection lifecycle to avoid cross-meeting wipes.  

So the immediate multi-meeting overwrite risk is addressed, though the singleton LLM model is still technical debt for true parallel meetings.

packages/@webex/plugin-meetings/src/constants.ts
Tianhui-Han: Done

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -231,28 +231,88 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
+
+  /**
+   * Clears a single session's data channel token.
+   * @param {DataChannelTokenType|string} tokenKey
+   * @returns {void}
+   */
+  public clearDatachannelToken = (
+    tokenKey: DataChannelTokenType | string = DataChannelTokenType.Default
+  ): void => {
+    this.datachannelTokens[tokenKey] = undefined;
```
marcin-bazyl: shouldn't we be checking ownership rules here in the same way like we do in setDatachannelToken() to prevent a meeting that's not an owner from deleting the token?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6460,8 +6483,9 @@ export default class Meeting extends StatelessWebexPlugin {
     throwOnError?: boolean;
   } = {}): Promise<void> => {
     // @ts-ignore - Fix type
-    const currentOwner = this.webex.internal.llm.getOwnerMeetingId();
-    const isOwner = !currentOwner || currentOwner === this.id;
+    const llmPlugin = this.webex.internal.llm;
+    // @ts-ignore - Fix type
+    const {currentOwner, isOwner} = llmPlugin.resolveSessionOwnership(this.id, LLM_DEFAULT_SESSION);
```
marcin-bazyl: what if we are in a practice session when cleanupLLMConneciton() is called? this code only handles cleanup of the main session

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6505,13 +6529,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
    * @returns {void}
    */
   clearDataChannelToken(): void {
     // @ts-ignore
-    this.webex.internal.llm.resetDatachannelTokens();
+    this.webex.internal.llm.clearDatachannelToken(LLM_DEFAULT_SESSION);
+
+    // Avoid wiping another meeting's practice-session token cache.
+    // @ts-ignore - Fix type
+    const {isOwner: isPracticeOwner} = this.webex.internal.llm.resolveSessionOwnership(
```
marcin-bazyl: why are we doing this `resolveSessionOwnership()` call and ownership check only for practice session and not the main session few lines above?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3470,8 +3471,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        // its own handler in webinar/index.ts (LLM_PRACTICE_SESSION).
+        // @ts-ignore - Fix type
+        const {isOwner} = this.webex.internal.llm.resolveSessionOwnership(
+          this.id,
+          LLM_DEFAULT_SESSION
+        );
+
+        if (isOwner) {
+          // @ts-ignore
+          this.webex.internal.llm.setRefreshHandler(
```
marcin-bazyl: question: I'm slightly confused why this call is needed here. We're in a listener for locus url change and the refresh handler is set for a given session id not a locus url, so why do we need to make this call here? if isOwner is true, doesn't that mean that we already have our refresh handler set for this session?

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -171,20 +170,40 @@ const Webinar = WebexPlugin.extend({
...
-      {
-        code: 3050,
-        reason: 'done (permanent)',
-      },
-      LLM_PRACTICE_SESSION
-    );
+    try {
+      if (isOwner) {
+        // @ts-ignore - Fix type
+        await this.webex.internal.llm.disconnectLLM(
```
marcin-bazyl: wouldn't it be better if the check for ownership was done inside disconnectLLM() so that no one can accidentally call disconnectLLM() without checking if they have ownership in the future?

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: This call is needed for ownership handoff safety, not because the session ID changes.

`LLM_DEFAULT_SESSION` is a shared slot across meeting instances. On locus URL updates, ownership can shift (or be reclaimed) even though the session key is the same. Re-setting `setRefreshHandler(...)` and immediately `setOwnerMeetingId(...)` ensures the default session is routed to the current owner early, before `registerAndConnect()` completes, and avoids stale handler races from another meeting instance.

This part of the change is based on the git codex review (title: Keep default-session handler aligned with ownership changes). You can check the details, and if you think it's unnecessary, I can also remove this part of the logic.

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: `cleanupLLMConneciton()` only cleans the default/main LLM session.  
Practice-session cleanup is done separately in `webinar.cleanupPSDataChannel()`, and both run in the normal meeting cleanup flow.

packages/@webex/plugin-meetings/src/webinar/index.ts
Tianhui-Han: You are absolutely right, and I updated it that way.

Ownership validation is now enforced inside disconnectLLM itself, so callers cannot bypass it by mistake. 
The method now requires explicit sessionId and ownerMeetingId, performs the ownership check internally, and skips disconnect when the caller is not the owner.

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: We have now encapsulated ownership checks for both sessions inside index.ts:6541, so it is no longer “main outside, practice inside.”

For main session, we now call resolveSessionOwnership with LLM_DEFAULT_SESSION before clearing the default token.

For practice session, we do the same with LLM_PRACTICE_SESSION.
index.ts:10010 now simply delegates to index.ts:6541, so the ownership logic lives in one place and is consistent for both tokens.

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: Ownership will not change with locus url update - the meeting object stays the same when locus url is changed, so the same meeting remains the owner,

I'm still confused - this code is in a locus url change handler and the AI comment you referenced is talking about the call to registerAndConnect() from updateLLMConnection() - what does this have to do with the locus url update?

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -212,9 +234,7 @@ const Webinar = WebexPlugin.extend({
     }
 
     // @ts-ignore
-    const cachedToken = this.webex.internal.llm.getDatachannelToken(
-      DataChannelTokenType.PracticeSession
-    );
+    const cachedToken = this.webex.internal.llm.getDatachannelToken(LLM_PRACTICE_SESSION);
```
marcin-bazyl: just a thought: I'm wondering if maybe getDatachannelToken() should also take an ownerMeetingId parameter so that you don't ever get a data channel token that you don't own? that token wouldn't work for you anyway, but maybe it would be easier to understand why if we checked the ownership and added a corresponding log?

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -231,28 +231,105 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
+    ownerMeetingId?: string
+  ): void => {
+    const resolvedTokenKey = tokenKey ?? DataChannelTokenType.Default;
+
+    const {currentOwner, canAssertOwnership, isOwner} = this.resolveSessionOwnership(
+      ownerMeetingId,
+      resolvedTokenKey
+    );
+
+    if (!isOwner && canAssertOwnership) {
```
marcin-bazyl: looking at resolveSessionOwnership() implementation, it is not possible at all to ever have `isOwner===false` AND `canAssertOwnership === false`, so that second part `&& canAssertOwnership` of this if condition is redundant - if isOwner is false, then canAssertOwnership must be true. The same applies to all other places with similar if condition. Furthermore, if we change all these conditions to just `if (!isOwner)` then `canAssertOwnership` is not needed as an output of `resolveSessionOwnership()` and we can remove it to keep the code more simple

packages/@webex/plugin-meetings/src/webinar/index.ts
Tianhui-Han: Yes, great thought. I made that change in this PR.

`getDatachannelToken()` now accepts an optional `ownerMeetingId` and performs the same ownership check used by other LLM session operations. If the caller is not the owner, it logs a clear skip message and returns `undefined` instead of returning a token owned by another meeting.

packages/@webex/internal-plugin-llm/src/llm.ts
Tianhui-Han: I agree with the reasoning.

I simplified this in the latest changes:

Replaced all if (!isOwner && canAssertOwnership) checks with if (!isOwner).
Removed canAssertOwnership from resolveSessionOwnership() return shape.

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -271,33 +329,71 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
+   * @param {string} sessionId - Connection identifier
+   * @param {string | undefined} ownerMeetingId - Meeting id asserting refresh-handler ownership
    * @returns {void}
    */
   public setRefreshHandler(
     handler: () => Promise<{
       body: {datachannelToken: string; datachannelTokenType: DataChannelTokenType};
-    }>
+    }>,
+    sessionId: string,
```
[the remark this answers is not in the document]
Tianhui-Han: This regression would be valid if setRefreshHandler wrote directly using the raw sessionId.
In the current implementation it normalizes with sessionId ?? LLM_DEFAULT_SESSION, so one-argument callers still register under the default session, not an undefined key.
refreshDataChannelToken also defaults to the same default session, so legacy one-argument behavior is preserved.

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -42,13 +39,54 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+            }
+          }
+        }
+
+        let result;
+        if (meeting) {
+          result = await meeting.refreshDataChannelToken();
+        } else {
+          // @ts-ignore
+          result = await this.internal.llm.refreshDataChannelToken(sessionId);
```
marcin-bazyl, line 115: I've just tried to test this PR - I joined a webinar as host, started breakouts and joined a breakout session - the code ended up here in this branch with sessionId being undefined and I ended up in a state with no LLM connection

marcin-bazyl (changes requested):

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
marcin-bazyl: this is 100% reproducible, you don't need to have any other meetings, just have a webinar, start it and start breakouts and join a breakout

Tianhui-Han: Done

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: we've talked about this 1-1 in webex, but still my question remains unanswered - why is this needed here? Please give me an exact example flow where things go wrong if we don't have this code here.

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -315,16 +432,38 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
     options: {code: number; reason: string},
-    sessionId: string = LLM_DEFAULT_SESSION
-  ): Promise<void> =>
-    this.disconnect(options, sessionId).then(() => {
+    sessionId: string,
+    ownerMeetingId: string
+  ): Promise<boolean> => {
+    if (!ownerMeetingId) {
+      throw new Error('llm#disconnectLLM --> ownerMeetingId is required');
+    }
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6687,6 +6722,15 @@ export default class Meeting extends StatelessWebexPlugin {
...
       .then((registerAndConnectResult) => {
+        // Re-bind the default-session token refresh path after each successful
+        // (re)connect so ownership handoffs cannot keep a stale Meeting-bound
+        // refresh handler.
+        // @ts-ignore - Fix type
+        this.webex.internal.llm.setRefreshHandler(
+          () => this.refreshDataChannelToken(),
+          LLM_DEFAULT_SESSION,
+          this.id
+        );
```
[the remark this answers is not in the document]
Tianhui-Han: In the current implementation, we pre-bind setRefreshHandler before connect, and in the stale-owner/disconnected path it is intentionally bound with ownerMeetingId = undefined, then ownership is reassigned only after a successful reconnect. 
 
 So we should not end up with owner/refresh-handler mismatch

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6636,7 +6673,8 @@ export default class Meeting extends StatelessWebexPlugin {
 
     // @ts-ignore
     const datachannelToken = this.webex.internal.llm.getDatachannelToken(
-      DataChannelTokenType.Default
+      LLM_DEFAULT_SESSION,
+      this.id
     );
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -257,20 +281,47 @@ const Webinar = WebexPlugin.extend({
...
+    this.webex.internal.llm.setRefreshHandler(
+      () => meeting.refreshDataChannelToken(),
+      LLM_PRACTICE_SESSION,
+      this.meetingId
+    );
+    // Claim ownership immediately after setting refresh routing so a
+    // concurrent meeting instance cannot override this practice-session
+    // refresh handler before registerAndConnect() finishes.
+    // @ts-ignore - Fix type
+    this.webex.internal.llm.setOwnerMeetingId?.(this.meetingId, LLM_PRACTICE_SESSION);
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -3470,8 +3471,30 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        const {isOwner} = this.webex.internal.llm.resolveSessionOwnership(
+          this.id,
+          LLM_DEFAULT_SESSION
+        );
+
+        if (isOwner) {
+          // @ts-ignore
+          this.webex.internal.llm.setRefreshHandler(
+            () => this.refreshDataChannelToken(),
+            LLM_DEFAULT_SESSION,
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -280,6 +315,14 @@ const Webinar = WebexPlugin.extend({
     }
     // @ts-ignore - Fix type
     if (this.webex.internal.llm.isConnected(LLM_PRACTICE_SESSION)) {
+      if (!isOwner && canAssertOwnership) {
+        LoggerProxy.logger.info(
+          `Webinar:index#updatePSDataChannel --> skipping; practice-session LLM owned by meeting ${currentOwner}, not ${this.meetingId}`
```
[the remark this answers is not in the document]
Tianhui-Han: We already guard ownership before any reconnect path in `updatePSDataChannel`, so a non-owner exits early even when practice-session LLM is disconnected.  
That means it cannot re-register, cannot reclaim ownership, and cannot overwrite refresh routing.  
This is also covered by tests for both connected and disconnected “owned-by-other-meeting” cases.

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: You are right — after re-checking the flow, this call in the locus URL listener is not the one that protects ownership handoff.

There is no concrete failure path where removing this listener call alone breaks refresh routing:

Ownership does not change just because locus URL changes on the same meeting object.
The handoff/race protection is handled in updateLLMConnection around registerAndConnect (pre-bind + post-claim alignment), not in the locus URL listener.
So my previous explanation here was imprecise. I can remove this listener-level setRefreshHandler/setOwnerMeetingId block in a follow-up cleanup PR to keep the logic focused in updateLLMConnection only.

Tianhui-Han:
> ### 💡 Codex Review
> https://github.com/webex/webex-js-sdk/blob/d52f4f16888ad13f2c2e7bb8746aca35f6be2061/packages/@webex/plugin-meetings/src/meeting/index.ts#L6547-L6549
> 
> 
> **P2 Badge Re-check ownership before clearing the LLM owner**
> If another meeting claims the default LLM session while this cleanup is awaiting `disconnectLLM`, the stale `isOwner` snapshot can still clear that new `ownerMeetingId` in `finally`. This leaves a live session ownerless, so later cleanup or reconnect calls from other meetings can pass the ownership guard and disconnect/reconfigure the wrong socket. Re-read the current owner before clearing, and only clear when it still belongs to this meeting.
> 
> ℹ️ About Codex in GitHub

done

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -356,6 +409,11 @@ const Webinar = WebexPlugin.extend({
         LLM_PRACTICE_SESSION
       )
       .then((registerAndConnectResult) => {
+        if (this.meetingId && isOwner) {
+          // @ts-ignore - Fix type
+          this.webex.internal.llm.setOwnerMeetingId?.(this.meetingId, LLM_PRACTICE_SESSION);
```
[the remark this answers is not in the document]
Tianhui-Han: done

Tianhui-Han:
> **P1 Badge Recheck ownership before clearing default-session owner tag**
> `cleanupLLMConneciton` computes `isOwner` before awaiting `disconnectLLM`, but the `finally` block always clears the default-session owner when that stale flag is true. In concurrent multi-meeting handoffs, ownership can change during the async disconnect path (where `disconnectLLM` may skip and return `false` for a new owner), and this unconditional clear removes the active owner tag, allowing other meetings to later hijack or disconnect a session they do not own.

done

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -347,6 +384,30 @@ const Webinar = WebexPlugin.extend({
...
+    // This avoids claiming ownership in flows that return early (e.g. missing
+    // practiceSessionDatachannelUrl or waiting for default session online).
+    // @ts-ignore - Fix type
+    this.webex.internal.llm.setRefreshHandler(
+      () => meeting.refreshDataChannelToken(),
+      LLM_PRACTICE_SESSION,
+      this.meetingId
+    );
+    // @ts-ignore - Fix type
+    this.webex.internal.llm.setOwnerMeetingId?.(this.meetingId, LLM_PRACTICE_SESSION);
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -42,13 +39,47 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
```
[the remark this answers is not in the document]
Tianhui-Han: done

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -60,6 +60,33 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+            // Before registerAndConnect resolves, LLM may not yet map requestUrl
+            // to a session/locus. Fall back to active meetings' locusInfo URLs.
+            // @ts-ignore
+            const allMeetings = this.meetings?.getAllMeetings?.() || {};
+
+            meeting = Object.values(allMeetings).find((activeMeeting: any) => {
+              const info = activeMeeting?.locusInfo?.info || {};
+
+              return (
+                info.practiceSessionDatachannelUrl === requestUrl ||
```
marcin-bazyl: why is it OK to just do a simple string comparison here and in getLocusUrlByDatachannelUrl() we use  `LLMChannel.matchesDatachannelRequestUrl()`? shouldn't we use `LLMChannel.matchesDatachannelRequestUrl()` here too?

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -60,6 +60,33 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+                info.datachannelUrl === requestUrl
+              );
+            });
+
+            if (!sessionId) {
+              // @ts-ignore
+              const info = meeting?.locusInfo?.info || {};
+
+              if (info.practiceSessionDatachannelUrl === requestUrl) {
```
marcin-bazyl: same as above, don't we need to use `matchesDatachannelRequestUrl()`?

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -60,6 +60,33 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
               meeting = this.meetings?.getMeetingByType?.(LOCUS_URL, locusUrl);
             }
           }
+
+          if (!meeting) {
+            // Before registerAndConnect resolves, LLM may not yet map requestUrl
+            // to a session/locus. Fall back to active meetings' locusInfo URLs.
```
marcin-bazyl: this is not great.... I'm not going to block this PR over this, but it feels very hacky. Shouldn't we store the urls in `connections` before registerAndConnect resolves? we might need to add a state to `connections` to indicate if the entry is already connected or connecting....

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6585,6 +6565,28 @@ export default class Meeting extends StatelessWebexPlugin {
...
+   */
+  private getDatachannelTokenWriteOwnerMeetingId(
+    tokenKey: DataChannelTokenType
+  ): string | undefined {
+    // @ts-ignore - Fix type
+    const {currentOwner} = this.webex.internal.llm.resolveSessionOwnership(this.id, tokenKey);
+    // @ts-ignore - Fix type
+    const isConnected = this.webex.internal.llm.isConnected(tokenKey);
+
+    if (!isConnected && currentOwner && currentOwner !== this.id) {
```
marcin-bazyl: question: why is this check done only here and getDatachannelTokenWriteOwnerMeetingId() is called only from a couple of places. If a situation described from the function description "When the session is disconnected but a foreign owner tag lingers" is something that can happen then shouldn't this be taken care of inside resolveSessionOwnership()? Also, how is it possible to get to the state of "foreign owner lingering after disconnect" - wouldn't it be better to avoid this in the first place?

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
Tianhui-Han: done

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
Tianhui-Han: done

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
Tianhui-Han: Agree this fallback is hacky.
I updated the main path so registerAndConnect() now pre-populates locusUrl/datachannelUrl in connections before register() runs, which fixes refresh routing during the registration window.

So getAllMeetings() here is now only a defensive fallback for rare timing gaps, not the normal path.

packages/@webex/plugin-meetings/src/meeting/index.ts
Tianhui-Han: You’re right. This shouldn’t be handled only at a few call sites; it should be prevented in the core layer.

I changed it so disconnectLLM clears the owner tag on successful disconnect, preventing “foreign owner lingering” at the source.
Because of that, the Meeting-side special helper was removed and we now use normal ownership-scoped writes.

marcin-bazyl (approved):
I don't like this PR, it's making the code very complicated and hard to read and maintain. I'm only approving it on the assumption that we will rewrite the whole llm and mercury plugins completely soon and we won't need all these "owner" workarounds after that. I'm on PTO for next 2 days and Monday is public holiday in the UK, so approving it only so that you're not blocked by my absence, but I haven't reviewed the tests and it's a big and complicated PR so needs a thorough 2nd review from someone.

mickelr (approved):

PR #4961, opened 2026-05-07
fix(plugin-meetings): reset stale share state during whiteboard handoff

mickelr:
## Summary

Fixes a webinar share-render failure during a whiteboard -> file/content share handoff.

When Locus reports `whiteboard.disposition = RELEASED` and `content.disposition = ACCEPTED`, `Meeting.setUpLocusMediaSharesListener()` used to keep the previous `whiteboard_share_active` state. That caused the SDK to re-emit whiteboard-start behavior for an already released whiteboard before the real content share was granted.

This PR defaults the computed share status to `NO_SHARE`, allowing only `GRANTED` content/whiteboard states to override it. Intermediate/non-active floor states such as `ACCEPTED` no longer carry stale share state forward.

## Root cause

Before this change:

```ts
let newShareStatus = this.shareStatus;
```

If content was `ACCEPTED` and whiteboard was `RELEASED`, none of the existing `GRANTED`/`RELEASED`/`null` branches matched, so `newShareStatus` remained stale as `whiteboard_share_active`.

## Expected behavior after fix

For the failing sequence:

```text
whiteboard_share_active + whiteboard RELEASED + content ACCEPTED -> no_share
no_share + content GRANTED -> remote_share_active
```

This avoids the spurious `MEETING_STARTED_SHARING_WHITEBOARD` event and gives consumers a clean whiteboard-stop then remote-share-start handoff.

## Tests

Temporarily focused the relevant `setUpLocusMediaSharesListener` block per plugin-meetings guidance, ran the targeted meeting test file, and removed `.only` afterward.

```zsh
cd /Users/mickelr/Workspace/git/webex-js-sdk
source ~/.nvm/nvm.sh
nvm install 22.14
nvm use 22.14
yarn workspace @webex/plugin-meetings test:unit --targets meeting/index.js
```

Result:

```text
26 passing (116ms)
```

Also ran:

```zsh
git --no-pager diff --check -- packages/@webex/plugin-meetings/src/meeting/index.ts packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```

Result: no whitespace/diff-check errors.


packages/@webex/plugin-meetings/test/unit/spec/meeting/index.js
```diff
@@ -14399,13 +14399,6 @@ describe('plugin-meetings', () => {
                     }
                   }
 
-                  if (isAccepting) {
```
marcin-bazyl, line 14402: the fix seems OK, but I don't think it will fix the actual issue with sharing. The issue it's fixing is just an unnecessary notification about whiteboard started when it's already started, so it won't change anything in the app state. It will stop SDK from sending an incorrect CA event about whiteboard, but I don't think that will have any effect on CA either.
However, any fixes in this area are high risk as they will affect whiteboards and screen sharing for all meetings and I don't understand the changes in the unit tests - the existing code in these unit tests seem overly complicated and these changes are hard to read. Are you able to explain all these changes in the tests?

PR #4962, opened 2026-05-07
fix(task-refactor): consult flow bug bash fixes

akulakum:
# COMPLETES #<https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7897>

Issues in the contact center SDK task-refactor state machine related to the consult flow lifecycle and 3-party conference controls. The uiControlsComputer, state machine guards, actions, task manager, and voice service did not correctly handle: (1) post-call lifecycle when the customer leaves a 3-way consult interaction, (2) conference UI controls during and after merge, (3) EP_DN conference API payload resolution, and (4) pending vs real conference state distinction.

**1. Fix extra controls visible after customer ends call (uiControlsComputer.ts)**

- When `customerPresent` is false during an active consult, `transfer`, `conference`, `mergeToConference`, and `switch` controls now return `VISIBLE_DISABLED` on the consult leg and `DISABLED` (hidden) on the main leg.
- The `recording` control on the main leg returns `DISABLED` when `hasParallelConsultLeg && !customerPresent`.

**2. Emit post-call activity event to Widgets (TaskManager.ts)**

- When `CC_EVENTS.PARTICIPANT_POST_CALL_ACTIVITY` is received, the task manager now explicitly emits `TASK_EVENTS.TASK_POST_CALL_ACTIVITY` on the task object so Widgets can render the "Post Call" timer label.
- Added null check before `removeTaskFromCollection` in `handleContactMergedEvent` to prevent `cancelAutoWrapupTimer` TypeError.

**3. Fix hydration failure during post-call state (guards.ts, actions.ts)**

- Enhanced `isInteractionConsulting` guard to return true when `interaction.state === 'post_call'` if self-agent's `consultState === 'consulting'` and consult media exists.
- Introduced `isActiveConsultState` helper to correctly gate hydration of consult context fields during post-call refresh.

**4. Fix incorrect wrap-up transition and conference state actions (TaskStateMachine.ts)**

- Added guarded transition: when `consultInitiator === true`, `interaction.isTerminated === true`, and `shouldWrapUpForThisAgent` is true, transitions directly from `CONSULTING → WRAPPING_UP`.
- Added `updateTaskData`/`syncTaskDataFromEvent` actions to `CONFERENCE_START` and `UNHOLD_SUCCESS` transitions to ensure task data is current when entering conference.
- Added `clearConsultState` action to `CONF_INITIATING → CONFERENCING` transition to reset stale consult context.

**5. Fix conference Exit Conference button for pending state (uiControlsComputer.ts)**

- `exitConference` returns `DISABLED` when `participantCount <= 1` (pending conference where only initiating agent is present).

**6. Fix EP_DN conference 400 Bad Request (Voice.ts)**

- Reordered resolution priority in `consultConference()` to use `derivedDestAgentId`/`derivedDestType` (computed from live interaction data via `calculateDestAgentId`/`calculateDestType`) as highest priority, ensuring EP_ID UUID is used instead of phone number.

**7. Fix conference controls for pending vs real conference (uiControlsComputer.ts)**

- Uses `participantCount` to distinguish pending conference (only self agent in mainCall) vs real multi-agent conference:
  - **Pending (participantCount <= 1):** Transfer/Recording enabled, Consult disabled, Exit Conference hidden
  - **Real (participantCount > 1):** Transfer/Recording hidden, Consult enabled, Exit Conference visible

**8. Fix missing wrapup screen after end call from pending conference (TaskManager.ts, TaskStateMachine.ts)**

- Added `AgentConsultConferencing` to the `CONFERENCE_START` event mapping in TaskManager so the state machine correctly transitions from `CONF_INITIATING` to `CONFERENCING` when Agent 1 merges before Agent 2 accepts.
- Added `CONSULT_END`, `CONTACT_ENDED`, and `TASK_WRAPUP` handlers to the `CONF_INITIATING` state for defense-in-depth, ensuring wrapup transitions are never dropped.
- Enhanced the `CONFERENCING` state's `CONSULT_END` handler to transition to `WRAPPING_UP` when `isTerminated === true`.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -619,7 +619,7 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
 
       [TaskState.CONF_INITIATING]: {
         on: {
-          // AgentConsultConferenced, ParticipantJoinedConference
+          // AgentConsultConferenced, AgentConsultConferencing, ParticipantJoinedConference
           [TaskEvent.CONFERENCE_START]: {
             target: TaskState.CONFERENCING,
             actions: [
@@ -633,6 +633,37 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           [TaskEvent.CONFERENCE_FAILED]: {
             target: TaskState.CONSULTING,
             actions: ['handleConferenceFailed', 'emitTaskConferenceFailed'],
+          },
+          // AgentConsultEnded while conference is initiating (end call before conference completes)
+          [TaskEvent.CONSULT_END]: [
+            {
+              guard: ({event}) => {
+                const taskData = getTaskDataFromEvent(event);
+
+                return taskData?.interaction?.isTerminated === true;
+              },
+              target: TaskState.WRAPPING_UP,
+              actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
+            },
+            {
+              target: TaskState.CONNECTED,
+              actions: ['updateTaskData', 'clearConsultState'],
+            },
+          ],
+          // ContactEnded while conference is initiating
+          [TaskEvent.CONTACT_ENDED]: {
+            target: TaskState.WRAPPING_UP,
+            actions: [
+              'updateTaskData',
+              'markEnded',
+              'clearConsultState',
+              'emitTaskWrapup',
+              'requestCleanup',
+            ],
+          },
+          [TaskEvent.TASK_WRAPUP]: {
+            target: TaskState.WRAPPING_UP,
+            actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
           },
         },
       },
@@ -660,9 +691,23 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           ],
 
           // Needed as all agents in conference get this event, hence we need to clear the consult state
-          [TaskEvent.CONSULT_END]: {
-            actions: ['updateTaskData', 'clearConsultState'],
-          },
+          [TaskEvent.CONSULT_END]: [
+            {
+              guard: ({context, event}) => {
+                const taskData = getTaskDataFromEvent(event);
+
+                return (
+                  taskData?.interaction?.isTerminated === true &&
+                  shouldWrapUpForThisAgent(context, taskData)
+                );
+              },
+              target: TaskState.WRAPPING_UP,
+              actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
+            },
+            {
+              actions: ['updateTaskData', 'clearConsultState'],
+            },
+          ],
 
           [TaskEvent.HOLD_SUCCESS]: {
             actions: ['updateTaskData', 'setHoldState', 'emitTaskHold'],
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -12,7 +12,7 @@ import {setup} from 'xstate';
 import {TaskContext, TaskEventPayload, UIControlConfig, TaskActionsMap} from './types';
 import {TaskState, TaskEvent} from './constants';
 import {actions, createInitialContext} from './actions';
-import {guards} from './guards';
+import {guards, shouldWrapUpForThisAgent, getTaskDataFromEvent} from './guards';
 import {getIsCustomerInCall} from '../TaskUtils';
 
 type TaskActionConfigMap = {[K in keyof typeof actions]: undefined};
@@ -466,6 +466,26 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
                 context.consultInitiator === true && context.consultCallHeld === true,
               target: TaskState.CONNECTED,
               actions: ['updateTaskData', 'clearConsultState', 'emitTaskConsultEnd'],
+            },
+            {
+              // Interaction terminated during consult (customer left) → WRAPPING_UP
+              guard: ({context, event}) => {
+                if (context.consultInitiator !== true) return false;
+                const taskData = getTaskDataFromEvent(event);
+
+                return (
+                  taskData?.interaction?.isTerminated === true &&
+                  shouldWrapUpForThisAgent(context, taskData)
+                );
+              },
+              target: TaskState.WRAPPING_UP,
+              actions: [
+                'updateTaskData',
+                'markEnded',
+                'clearConsultState',
+                'emitTaskWrapup',
+                'requestCleanup',
+              ],
             },
             {
               // Initiator (no conference) → HELD
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -607,7 +607,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           // AgentConsultConferenced, ParticipantJoinedConference
           [TaskEvent.CONFERENCE_START]: {
             target: TaskState.CONFERENCING,
-            actions: ['handleConferenceStarted', 'clearConsultState'],
+            actions: [
+              'updateTaskData',
+              'syncTaskDataFromEvent',
+              'handleConferenceStarted',
+              'clearConsultState',
+            ],
           },
         },
       },
@@ -617,7 +622,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           // AgentConsultConferenced, ParticipantJoinedConference
           [TaskEvent.CONFERENCE_START]: {
             target: TaskState.CONFERENCING,
-            actions: ['handleConferenceStarted'],
+            actions: [
+              'updateTaskData',
+              'syncTaskDataFromEvent',
+              'handleConferenceStarted',
+              'clearConsultState',
+            ],
           },
           // AgentConsultConferenceFailed
           [TaskEvent.CONFERENCE_FAILED]: {
@@ -630,7 +640,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
       [TaskState.CONFERENCING]: {
         on: {
           [TaskEvent.CONFERENCE_START]: {
-            actions: ['updateTaskData', 'clearConsultState', 'emitTaskConferenceStarted'],
+            actions: [
+              'updateTaskData',
+              'syncTaskDataFromEvent',
+              'clearConsultState',
+              'emitTaskConferenceStarted',
+            ],
           },
           [TaskEvent.EXIT_CONFERENCE_SUCCESS]: [
             {
@@ -653,7 +668,7 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             actions: ['updateTaskData', 'setHoldState', 'emitTaskHold'],
           },
           [TaskEvent.UNHOLD_SUCCESS]: {
-            actions: ['updateTaskData', 'setHoldState', 'emitTaskResume'],
+            actions: ['updateTaskData', 'syncTaskDataFromEvent', 'setHoldState', 'emitTaskResume'],
           },
 
           // Start a new consult from within an active conference
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -468,15 +468,12 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
               actions: ['updateTaskData', 'clearConsultState', 'emitTaskConsultEnd'],
             },
             {
-              // Interaction terminated during consult (customer left) → WRAPPING_UP
+              // Customer left during consult → WRAPPING_UP
               guard: ({context, event}) => {
                 if (context.consultInitiator !== true) return false;
                 const taskData = getTaskDataFromEvent(event);
 
-                return (
-                  taskData?.interaction?.isTerminated === true &&
-                  shouldWrapUpForThisAgent(context, taskData)
-                );
+                return shouldWrapUpForThisAgent(context, taskData, {requireCustomerLeft: true});
               },
               target: TaskState.WRAPPING_UP,
               actions: [
--- a/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
@@ -470,10 +470,11 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             {
               // Customer left during consult → WRAPPING_UP
               guard: ({context, event}) => {
-                if (context.consultInitiator !== true) return false;
                 const taskData = getTaskDataFromEvent(event);
+                const cpd = taskData?.interaction?.callProcessingDetails;
+                if (cpd?.hasCustomerLeft !== 'true') return false;
 
-                return shouldWrapUpForThisAgent(context, taskData, {requireCustomerLeft: true});
+                return shouldWrapUpForThisAgent(context, taskData);
               },
               target: TaskState.WRAPPING_UP,
               actions: [
@@ -647,21 +648,6 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
               actions: ['updateTaskData', 'clearConsultState'],
             },
           ],
-          // ContactEnded while conference is initiating
-          [TaskEvent.CONTACT_ENDED]: {
-            target: TaskState.WRAPPING_UP,
-            actions: [
-              'updateTaskData',
-              'markEnded',
-              'clearConsultState',
-              'emitTaskWrapup',
-              'requestCleanup',
-            ],
-          },
-          [TaskEvent.TASK_WRAPUP]: {
-            target: TaskState.WRAPPING_UP,
-            actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
-          },
         },
       },
 
--- a/packages/@webex/contact-center/src/services/task/state-machine/actions.ts
+++ b/packages/@webex/contact-center/src/services/task/state-machine/actions.ts
@@ -74,12 +74,6 @@ const deriveRecordingState = (taskData?: TaskData | null): RecordingStateUpdate 
   return update;
 };
 
-const isActiveConsultState = (taskData: TaskData | undefined, selfAgentId?: string): boolean => {
-  if (taskData?.interaction?.state === INTERACTION_STATE.CONSULTING) return true;
-
-  return hasActiveConsultInPostCall(taskData, selfAgentId);
-};
-
 const deriveTaskDataUpdates = (context: TaskContext, taskData: TaskData | undefined) =>
   taskData
     ? (() => {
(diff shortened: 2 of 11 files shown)


packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -73,6 +73,19 @@ const deriveRecordingState = (taskData?: TaskData | null): RecordingStateUpdate
   return update;
 };
 
+const isActiveConsultState = (taskData: TaskData | undefined, selfAgentId?: string): boolean => {
+  if (taskData?.interaction?.state === 'consulting') return true;
```
rsarika: are we not using constants for 'consulting' and 'post_call' etc. if not consider doing it

rsarika (approved):

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: Good catch! Added INTERACTION_STATE and CONSULT_STATE constants in constants.ts and updated all usages in actions.ts and guards.ts to use them instead of raw strings.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -467,6 +467,26 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
             },
+            {
+              // Interaction terminated during consult (customer left) → WRAPPING_UP
+              guard: ({context, event}) => {
+                if (context.consultInitiator !== true) return false;
+                const taskData = getTaskDataFromEvent(event);
+
+                return (
+                  taskData?.interaction?.isTerminated === true &&
+                  shouldWrapUpForThisAgent(context, taskData)
```
Kesari3008: Here using shouldWrapUpForThisAgent should suffice right why do we need extra conditions here ? We already have matches present for cases where initiator has to go back to HELD after end consult and consulted agent goes to terminated. We do not need this elaborated guard

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -625,15 +691,29 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
           ],
 
           // Needed as all agents in conference get this event, hence we need to clear the consult state
-          [TaskEvent.CONSULT_END]: {
-            actions: ['updateTaskData', 'clearConsultState'],
-          },
+          [TaskEvent.CONSULT_END]: [
```
Kesari3008, line 677: Which case this new transitions would be handling ?

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -587,30 +607,76 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+              'clearConsultState',
+            ],
           },
           // AgentConsultConferenceFailed
           [TaskEvent.CONFERENCE_FAILED]: {
             target: TaskState.CONSULTING,
             actions: ['handleConferenceFailed', 'emitTaskConferenceFailed'],
           },
+          // AgentConsultEnded while conference is initiating (end call before conference completes)
+          [TaskEvent.CONSULT_END]: [
```
Kesari3008, line 636: I would like to understand these 2 cases as well, Could you please elaborate these for me

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -73,6 +79,19 @@ const deriveRecordingState = (taskData?: TaskData | null): RecordingStateUpdate
   return update;
 };
 
+const isActiveConsultState = (taskData: TaskData | undefined, selfAgentId?: string): boolean => {
+  if (taskData?.interaction?.state === INTERACTION_STATE.CONSULTING) return true;
+  if (taskData?.interaction?.state === INTERACTION_STATE.POST_CALL && selfAgentId) {
+    const selfParticipant = taskData.interaction?.participants?.[selfAgentId] as any;
```
Kesari3008: Using any should be avoided as much as possible

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -81,27 +100,30 @@ const deriveTaskDataUpdates = (context: TaskContext, taskData: TaskData | undefi
           ...deriveRecordingState(taskData),
         };
 
+        const selfAgentId = context.uiControlConfig.agentId ?? taskData?.agentId;
+        const consultingActive = isActiveConsultState(taskData, selfAgentId);
```
Kesari3008: Isn't there a simpler way to figure out if there is an active consult ? State machine state itself would be consulting for active consult. We are using isActiveConsult only in one place and it looks like unncessary complicated logic to just figure out if there is an active consult.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -230,7 +252,10 @@ export const actions: TaskActionsMap = {
     const taskData = getTaskDataFromEvent(event);
     const consultDestinationType =
       'destinationType' in event ? event.destinationType ?? null : null;
-    const consultDestinationAgentId = 'destAgentId' in event ? event.destAgentId ?? null : null;
+    const consultDestinationAgentId =
```
Kesari3008, line 239: Why was this change required ?

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -81,27 +100,30 @@ const deriveTaskDataUpdates = (context: TaskContext, taskData: TaskData | undefi
           ...deriveRecordingState(taskData),
         };
 
+        const selfAgentId = context.uiControlConfig.agentId ?? taskData?.agentId;
+        const consultingActive = isActiveConsultState(taskData, selfAgentId);
+
         if (taskData.destAgentId) {
-          updates.consultDestinationAgentId = taskData.destAgentId;
+          const isEpDnWithStoredId =
```
Kesari3008, line 91: Please explain this change

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -97,17 +97,35 @@ export const guards = {
...
-    if (cpd?.relationshipType === 'consult' && taskData?.interaction?.state === 'connected') {
+    if (
+      cpd?.relationshipType === 'consult' &&
+      taskData?.interaction?.state === INTERACTION_STATE.CONNECTED
+    ) {
       return true;
     }
 
+    // Customer left during consult: interaction state is "post_call" but consult
+    // between agents is still active. Detect via agent's consultState + consult media.
```
Kesari3008, line 133: How was this guard impacting the flow and this change was needed here ? and this looks very similar to what we are doing in isActiveConsult as well

packages/@webex/contact-center/test/unit/spec/services/task/state-machine/uiControlsComputer.ts
```diff
@@ -27,7 +27,7 @@ function createConsultTaskData() {
           currentState: 'consulting',
           isConsulted: true,
         },
-        'customer-1': {id: 'customer-1', pType: 'CUSTOMER', hasLeft: false},
+        'customer-1': {id: 'customer-1', pType: 'Customer', hasLeft: false},
```
Kesari3008, line 30: Did we not need any other unit test changes ?

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -710,17 +710,22 @@ export default class Voice extends Task implements IVoice {
...
-      destinationType:
-        this.getStateMachineSnapshot()?.context?.consultDestinationType ||
-        this.data.destinationType ||
-        derivedDestType ||
-        'agent',
+      destinationType: resolvedDestinationType,
+      // derivedDestAgentId is most reliable as it resolves epId for EP_DN
+      // and agent ID for regular agents from live interaction data
       destAgentId:
+        derivedDestAgentId ||
```
Kesari3008, line 726: Is it always present ?

Kesari3008 (changes requested):

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: shouldWrapUpForThisAgent alone would match even when the consult ends normally (where the initiator should return to HELD or CONFERENCING). The isTerminated check distinguishes between: (1) normal consult end — Agent 2 hangs up, main call still active → initiator goes to HELD, and (2) interaction terminated — customer left during consult → initiator should wrap up. Without isTerminated, this guard fires prematurely during a normal consult end and swallows transitions meant for the subsequent guards below it.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: This handles: Agent 1 merges EP_DN consult → enters CONFERENCING → clicks End Call → backend sends AgentConsultEnded with isTerminated: true. Previously, CONSULT_END in CONFERENCING only cleared consult state without transitioning, so the agent stayed stuck in CONFERENCING instead of moving to WRAPPING_UP. The guard checks isTerminated && shouldWrapUpForThisAgent to only transition when the call has actually ended.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: These are defense-in-depth handlers for race conditions. After Agent 1 clicks merge, the state machine enters CONF_INITIATING while awaiting CONFERENCE_START. If the call ends or consult ends before the conference confirmation arrives (e.g., network delay, Agent 1 immediately clicks End Call), these events would be dropped since CONF_INITIATING previously only handled CONFERENCE_START and CONFERENCE_FAILED.

Two cases:

1. isTerminated === true → interaction has ended, transition to WRAPPING_UP
2. Not terminated → consult ended normally, transition back to CONNECTED

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: Fixed. Removed as any cast on selfParticipant (direct property access from participants map) and used explicit type assertion (media as {mType?: string})?.mType for media objects instead of as any.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: This function is used specifically during hydration (page refresh) in deriveTaskDataUpdates. At that point, the state machine state hasn't been restored yet — we're inspecting raw TaskData from the backend to determine which context fields to populate. We can't rely on the state machine's current state because it's being rebuilt from scratch. The post_call case is needed because when the customer leaves during a consult, interaction.state becomes 'post_call' but the consult between agents is still active. Without this check, consult context fields wouldn't be populated on refresh, and the UI would lose the consulting section.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: For EP_DN consults, event.destAgentId may not be present in the event payload, but event.destination contains the Entry Point UUID needed for conference API calls. This change ensures we store the EP_DN UUID in consultDestinationAgentId so that Voice.consultConference() can resolve the correct destAgentId when initiating a conference, preventing the 400 Bad Request error.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
akulakum: During hydration, taskData.destAgentId can contain the EP_DN phone number (e.g., +13159998059) instead of the Entry Point UUID. If the context already has the correct UUID stored from the live consult flow (consultDestinationType === 'entryPoint' and consultDestinationAgentId exists), we skip overwriting it with the phone number. Without this, after a page refresh, the conference API would send the phone number instead of the UUID, causing a 400 Bad Request.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
akulakum: The isInteractionConsulting guard determines which state the machine hydrates into on page refresh. Without this post_call check, when the customer leaves during an active consult, the guard returns false (since interaction.state is 'post_call', not 'consulting'), and the state machine hydrates to CONNECTED instead of CONSULTING — causing the agent to lose the entire consulting section and controls. The similarity with isActiveConsultState is by design: the guard decides the target state (CONSULTING), while isActiveConsultState in actions decides whether to populate consult context fields. Both need the same detection logic at different layers. We can extract a shared utility if preferred.

packages/@webex/contact-center/test/unit/spec/services/task/state-machine/uiControlsComputer.ts
akulakum: Added 11 new unit tests covering: (1) conference controls — pending vs real conference (transfer/recording/consult/exitConference behavior based on participant count), (2) hold visibility in conference, (3) post-call consult controls when customer left, (4) CONF_INITIATING state transitions (CONSULT_END, CONTACT_ENDED, TASK_WRAPUP), (5) CONFERENCING → WRAPPING_UP on CONSULT_END with isTerminated, and (6) isInteractionConsulting guard for EP_DN and post_call hydration scenarios.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
akulakum: No, it's not always present — it's derived from calculateDestAgentId() which inspects live interaction participants. It returns empty string if no consulted agent or EP_DN participant is found. That's why we have a fallback chain: derivedDestAgentId || context.consultDestinationAgentId || this.data.destAgentId. The derivedDestAgentId is the most reliable source because it resolves the correct EP_ID UUID for EP_DN from the interaction's participant epId field, but we fall back to context/taskData if the interaction data isn't available yet.

akulakum:
1. Removed isActiveConsultState function from actions.ts — inlined its logic into deriveTaskDataUpdates using the shared hasActiveConsultInPostCall utility
2. Extracted hasActiveConsultInPostCall shared utility in guards.ts — eliminates duplicate post_call + consult detection logic between actions.ts and guards.ts
3. Simplified isInteractionConsulting guard to use hasActiveConsultInPostCall instead of repeating the same checks
4. Moved hasCustomerLeft check inside shouldWrapUpForThisAgent via requireCustomerLeft option — simplifies Guard 3 in TaskStateMachine.ts (removes inline callProcessingDetails access from the guard)
5. Replaced magic strings 'consulting', 'post_call', 'consult' with INTERACTION_STATE, CONSULT_STATE, MEDIA_TYPE_CONSULT constants

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -587,30 +604,76 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+              target: TaskState.WRAPPING_UP,
+              actions: ['updateTaskData', 'markEnded', 'clearConsultState', 'emitTaskWrapup'],
+            },
+            {
+              target: TaskState.CONNECTED,
+              actions: ['updateTaskData', 'clearConsultState'],
+            },
+          ],
+          // ContactEnded while conference is initiating
+          [TaskEvent.CONTACT_ENDED]: {
```
Kesari3008: We discussed these, these cases are not really hitting  a match

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -467,6 +467,23 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
               target: TaskState.CONNECTED,
               actions: ['updateTaskData', 'clearConsultState', 'emitTaskConsultEnd'],
             },
+            {
+              // Customer left during consult → WRAPPING_UP
+              guard: ({context, event}) => {
+                if (context.consultInitiator !== true) return false;
```
Kesari3008: This initiator check is also not needed

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -467,6 +467,23 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+                const taskData = getTaskDataFromEvent(event);
+
+                return shouldWrapUpForThisAgent(context, taskData, {requireCustomerLeft: true});
```
Kesari3008: Why we need to add a flag here? Would there be an issue to directly check hasCustomerleft from event payload.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -44,14 +44,41 @@ export const isSelfConsultingAgent = (context: TaskContext, taskData?: TaskData)
...
+  context: TaskContext,
+  taskData: TaskData,
+  options?: {requireCustomerLeft?: boolean}
+): boolean => {
   const selfAgentId = getSelfAgentId(context, taskData);
   if (!selfAgentId) return false;
 
+  if (options?.requireCustomerLeft) {
+    const cpd = taskData?.interaction?.callProcessingDetails;
+    if (cpd?.hasCustomerLeft !== 'true') return false;
```
Kesari3008: Is hasCustomerLeft is true then it doesn't seem to be returning true here so won;t it exit this block and still match old conditions

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Agreed. Removed both CONTACT_ENDED and TASK_WRAPUP handlers from CONF_INITIATING. The existing CONSULT_END handler in this state is sufficient to cover edge cases where the consult ends before conference completes.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Removed.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
akulakum: Removed the requireCustomerLeft flag entirely and reverted shouldWrapUpForThisAgent back to its original signature. The hasCustomerLeft check is now directly inline in the guard:

guard: ({context, event}) => {
  const taskData = getTaskDataFromEvent(event);
  const cpd = taskData?.interaction?.callProcessingDetails;
  if (cpd?.hasCustomerLeft !== 'true') return false;
  return shouldWrapUpForThisAgent(context, taskData);
},

The reason hasCustomerLeft is kept as a pre-condition in the guard rather than inside shouldWrapUpForThisAgent is that shouldWrapUpForThisAgent is a shared utility — it's also called by CONTACT_ENDED and EXIT_CONFERENCE_SUCCESS transitions, whose payloads don't contain hasCustomerLeft. Moving it inside would either break those flows or require the same conditional flag we just removed. This way the guard owns the scenario-specific check ("did the customer leave?") and shouldWrapUpForThisAgent stays focused on the shared eligibility check ("is this agent pending wrap-up?"). Updated in latest commit.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
akulakum: You're right — the flag-based approach was confusing and error-prone. Removed the entire requireCustomerLeft block from shouldWrapUpForThisAgent. The function is now back to its original clean form with no conditional branches. The hasCustomerLeft check lives directly in Guard 3 as a pre-condition (see reply above). If hasCustomerLeft !== 'true', the guard returns false immediately and the state machine falls through to the next guard which handles normal consult end (back to CONNECTED). Updated in latest commit.

Kesari3008 (approved):

PR #4968, opened 2026-05-08
feat(auth): oauth helper methods

robstax:
new SSO OAuth helper functions

Adds Credentials.prototype.buildThirdPartyLoginUrl({oauth2provider, returnURL}) which constructs the IdBroker /idb/ThirdPartyLogin URL (mirrors the existing buildLoginUrl/buildLogoutUrl style). Adds Authorization.prototype.initiateThirdPartyLogin and initiateThirdPartyLoginRedirect to wrap the URL build + window.location assignment, mirroring initiateAuthorizationCodeGrant's shape so the web client can override initiateThirdPartyLoginRedirect for iframed targetOrigin/postMessage handling. Behaviour-preserving lift of logic that today lives in webex-web-client's createThirdPartyLoginURI; no consumers wired up yet.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -212,6 +212,47 @@ const Credentials = WebexPlugin.extend({
...
+   */
+  buildThirdPartyLoginUrl({oauth2provider, returnURL} = {}) {
+    if (!oauth2provider) {
+      throw new Error('`options.oauth2provider` is required');
+    }
+    if (!returnURL) {
+      throw new Error('`options.returnURL` is required');
+    }
+
+    const baseUrl = this.webex.internal.services.get('idbroker') || this.config.idbroker.url;
```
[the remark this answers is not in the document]
robstax: changed to match other build* functions

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -212,6 +212,38 @@ const Credentials = WebexPlugin.extend({
...
+   * @instance
+   * @memberof Credentials
+   * @param {Object} options
+   * @param {string} options.oauth2provider - Provider name (`google`,
+   *   `microsoft`, `apple`, ...). Required.
+   * @param {string} options.returnURL - URL IdBroker should send the user
+   *   back to after the third-party hand-off. Required.
+   * @returns {string}
+   */
+  buildThirdPartyLoginUrl({oauth2provider, returnURL} = {}) {
```
martroos: suggestion: Other methods in this file don't destructure arguments, instead they just accept `options` - this may be preferable since it is more flexible (in case someone wanted to provide an `email` for example, which appears to be another valid parameter to this API).

On further thought, these aren't really options, they are required, so maybe they should just be positional arguments, with an options object at the end, i.e.:

```js
buildThirdPartyLoginUrl(oauth2provider, returnURL, options={}) {
```

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -212,6 +212,38 @@ const Credentials = WebexPlugin.extend({
...
+    if (!oauth2provider) {
+      throw new Error('`options.oauth2provider` is required');
+    }
+    if (!returnURL) {
+      throw new Error('`options.returnURL` is required');
+    }
```
martroos, line 241: nitpick: Whilst I appreciate these are required, there are other validation rules on these (I suspect `oauth2provider` has to be a known value, and `returnURL` should be a valid URL) which is being done on the server side, I'm not sure how much value there is to doing this basic validation on the client side, especially if we make these positional arguments. That being said, these aren't doing any harm, so not worried about removing them.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -232,6 +232,47 @@ describe('webex-core', () => {
       });
     });
 
+    describe('#buildThirdPartyLoginUrl()', () => {
+      it('throws if `oauth2provider` is missing', () => {
```
martroos, line 247: nitpick: For completeness, a test where both are missing would be good.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -232,6 +232,47 @@ describe('webex-core', () => {
...
+        const credentials = new Credentials(undefined, {parent: webex});
+
+        webex.trigger('change:config');
+
+        assert.throws(() => {
+          credentials.buildThirdPartyLoginUrl({oauth2provider: 'google'});
+        }, /`options.returnURL` is required/);
+      });
+
+      skipInBrowser(it)('generates the third-party login url', () => {
```
martroos, line 269: question: Why `skipInBrowser`? Not suggesting it's wrong, just wanting to understand.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -232,6 +232,47 @@ describe('webex-core', () => {
...
+        const webex = new MockWebex();
+        const credentials = new Credentials(undefined, {parent: webex});
+
+        webex.trigger('change:config');
+
+        assert.equal(
+          credentials.buildThirdPartyLoginUrl({
+            oauth2provider: 'google',
+            returnURL: 'https://web.webex.com',
```
martroos, line 278: nitpick: I'd like to have at least one other test that used different values for the parameters to sanity check that we aren't just hardcoding these.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -343,6 +343,59 @@ const Authorization = WebexPlugin.extend({
...
 
+  /**
+   * Initiates third-party (social provider) login by redirecting the user
+   * to IdBroker's `/idb/ThirdPartyLogin` endpoint.
+   *
+   * Mirrors `initiateLogin`'s role for the standard `/authorize` flow.
+   * Delegates to `initiateThirdPartyLoginRedirect` so any pre-redirect
+   * security plumbing (CSRF / `state`) can be added here without changing
+   * the navigation method, which the web client overrides for
+   * `targetOrigin` / postMessage handling in iframed contexts.
```
martroos: suggestion: Not sure we want to include web client specific details in this repo.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -343,6 +343,59 @@ const Authorization = WebexPlugin.extend({
...
+  },
+
+  /**
+   * Performs the navigation step of the third-party login flow. Builds the
+   * IdBroker URL via `Credentials#buildThirdPartyLoginUrl` and assigns it
+   * to `getWindow().location`.
+   *
+   * Mirrors `initiateAuthorizationCodeGrant` for the `/authorize` flow. The
+   * web client overrides this method to add `targetOrigin`/postMessage
+   * branching for iframed (crosslauncher / MJP) contexts.
```
martroos: suggestion: Again, not sure we want to include web client specific details in this repo.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -343,6 +343,59 @@ const Authorization = WebexPlugin.extend({
...
+   * tick and provide no useful signal.
+   *
+   * @instance
+   * @memberof AuthorizationBrowserFirstParty
+   * @param {Object} options
+   * @param {string} options.oauth2provider
+   * @param {string} options.returnURL
+   * @returns {Promise<void>}
+   */
+  initiateThirdPartyLoginRedirect(options = {}) {
```
martroos, line 382: question: Are you expecting this to _always_ return a Promise?

Currently, if the call to `buildThirdPartyLoginUrl`, it will throw an error instead of returning the promise, which couldn't be caught with a `.catch` (since no promise is available). Should this instead either be an async function or catch the error here and return a rejected promise instead?

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -560,6 +560,72 @@ describe('plugin-authorization-browser-first-party', () => {
...
+          assert.calledOnceWithExactly(stub, options);
+        });
+      });
+
+      it('returns the promise from #initiateThirdPartyLoginRedirect', () => {
+        const webex = makeWebex();
+        const expected = Promise.resolve();
+        sinon.stub(webex.authorization, 'initiateThirdPartyLoginRedirect').returns(expected);
+
+        assert.equal(
```
martroos: question: Is there a chaijs equivalent of jest's `toBe`? Maybe `strictEqual` would be closer, maybe it doesn't matter, but in theory this should be the _exact_ same promise object being returned.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -560,6 +560,72 @@ describe('plugin-authorization-browser-first-party', () => {
...
+          oauth2provider: 'google',
+          returnURL: 'https://web.webex.com',
+        };
+
+        return webex.authorization.initiateThirdPartyLogin(options).then(() => {
+          assert.calledOnceWithExactly(stub, options);
+        });
+      });
+
+      it('returns the promise from #initiateThirdPartyLoginRedirect', () => {
```
martroos: suggestion: Combine this test with the previous one - I don't think there is benefit to keeping them separate, both should be true at the same time.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -560,6 +560,72 @@ describe('plugin-authorization-browser-first-party', () => {
...
+          .then(() => {
+            assert.calledOnceWithExactly(webex.credentials.buildThirdPartyLoginUrl, {
+              oauth2provider: 'google',
+              returnURL: 'https://web.webex.com',
+            });
+            assert.equal(webex.getWindow().location, builtUrl);
+          });
+      });
+
+      it('returns a Promise', () => {
```
martroos: suggestion: This is covered by the previous test - if it didn't return a promise, the previous test would fail.

martroos (changes requested):
Functionally works well, just some questions about the implementation I'd like addressing before approval.

packages/@webex/webex-core/src/lib/credentials/credentials.js
robstax: good call out. i changed per your suggestion and to more closely align with the other functions in the file.

packages/@webex/webex-core/src/lib/credentials/credentials.js
robstax: ack. i left as is. backend does verify too.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
robstax: added

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
robstax: two reasons (both a little tenous)
1. other tests in this file do that
2. my understanding is that these tests run in `node` and `browser`. these functions use `process.env` which is not present when these tests are run in `browser` mode

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: yep. i deleted. probably should have just messaged why :P

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: ack. removed

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: good call out. wrapped in try/catch and returns a `Promise.reject(err)`. felt like this best aligned with other patterns in the plugin.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: my understanding is that chai's `toEqual` internally uses `===`. also, almost nothing else in the SDK uses `strictEqual`

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: yep done.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: yep. changed to reflect the new function behavior/return values.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
martroos: According to [Chai docs](https://www.chaijs.com/api/assert/#method_equal), `.equal` uses non-strict equality (`==`), which is _why_ a `strictEqual` exists. Not too concerned, I just tried it and it does fail/error if a different resolved promise to the expected one is returned, so the test will catch this with `.equal`.

martroos (approved):
Not sure how easy it would be to avoid, but the auto-formatting in these files made it a little trickier to see only the changes you made (i.e. not the auto-formatter), but fortunately this is a relatively small PR!

chrisadubois (approved):

PR #4986, opened 2026-05-13
feat(plugin-meetings): added Locus hash trees back-off syncs when LLM is connected

marcin-bazyl:
# COMPLETES #[SPARK-804241](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-804241)

New requirement from Locus team:
- when we connect a new LLM or after we reconnect LLM/Mercury, we need to do a sync, but we should not do it immediately, we should first wait the back-off period (without idleMs, just the back-off) and then do the sync (so /hashtree and /sync API calls)

We already do syncs after Mercury reconnections, but we don't trigger them currently after new LLM connection, so now doing this:
1. calling syncAllDatasets() after new LLM connection is connected or LLM goes online
2. doing a back-off wait before actually starting doing the syncs in HashTreeParser.syncAllDatasets()

We also need to take into account that messages might come in while we're doing the back-off and in that case we need to cancel the sync for those datasets - introduced `dataSetsSyncedDuringBackoff` prop for that.
We also need to distinguish between doing syncAllDatasets() for all datasets or just the ones that are updated over LLM. So far I've been trying to avoid having any knowledge in SDK about which data sets come over LLM vs Mercury, but now there is no way around it, SDK needs to know to only sync LLM based datasets when LLM is connected,

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual testing with web app and Locus on convergedats

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/hashTree/constants.ts
+++ b/packages/@webex/plugin-meetings/src/hashTree/constants.ts
@@ -17,3 +17,10 @@ export const DataSetNames = {
 // participant object for webinar attendees. If SELF were initialized first, locus.info
 // would not yet be populated and the attendee participant would be skipped.
 export const DATA_SET_INIT_PRIORITY: string[] = [DataSetNames.MAIN, DataSetNames.SELF];
+
+// Data sets for which we normally receive events over LLM connection
+export const LLM_DATASET_NAMES = [
+  DataSetNames.MAIN,
+  DataSetNames.ATD_ACTIVE,
+  DataSetNames.ATD_UNMUTED,
+];
--- a/packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
+++ b/packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
@@ -4,10 +4,16 @@ import LoggerProxy from '../common/logs/logger-proxy';
 import Metrics from '../metrics';
 import BEHAVIORAL_METRICS from '../metrics/constants';
 import {Enum, HTTP_VERBS} from '../constants';
-import {DataSetNames, DATA_SET_INIT_PRIORITY, EMPTY_HASH} from './constants';
+import {DataSetNames, DATA_SET_INIT_PRIORITY, EMPTY_HASH, LLM_DATASET_NAMES} from './constants';
 import {ObjectType, HtMeta, HashTreeObject} from './types';
 import {LocusDTO, LocusErrorCodes} from '../locus-info/types';
-import {deleteNestedObjectsWithHtMeta, isMetadata, sortByInitPriority} from './utils';
+import {deleteNestedObjectsWithHtMeta, isMetadata, sleep, sortByInitPriority} from './utils';
+
+export enum SyncAllBackoffType {
+  NONE = 'none',
+  ONLY_LLM = 'onlyLLM',
+  ALL = 'all',
+}
 
 export interface DataSet {
   url: string;
@@ -122,7 +128,10 @@ class HashTreeParser {
   state: 'active' | 'stopped';
   private syncQueue: Array<{dataSetName: string; reason: string; isInitialization?: boolean}> = [];
   private isSyncInProgress = false;
-  private isSyncAllInProgress = false;
+  // tracks whether syncAllDatasets is currently in its backoff delay phase and with what scope
+  private syncAllBackoffType: SyncAllBackoffType = SyncAllBackoffType.NONE;
+  // datasets that received messages during the syncAllDatasets backoff sleep and should be skipped
+  private dataSetsSyncedDuringBackoff: Set<string> = new Set();
   private syncQueueProcessingPromise: Promise<void> = Promise.resolve();
 
   /**
@@ -1342,8 +1351,14 @@ class HashTreeParser {
         // parseMessage() -> cancelPendingSyncsForDataSets() doesn't log a
         // misleading "aborting sync" message for this already-completed sync
         dataSet.syncAbortController = undefined;
+
         // the format of sync response is the same as messages, so we can reuse the same handler
-        this.handleMessage(syncResponse, 'via sync API');
+        this.handleMessage(
+          syncResponse,
+          `via sync API (${
+            isInitialization ? 'init' : `${Object.keys(leavesData).length} mismatched leaves`
+          })`
+        );
       }
     } catch (error) {
       if (!this.handleSyncErrors(error)) {
@@ -1375,6 +1390,13 @@ class HashTreeParser {
           previousLength - this.syncQueue.length
         } entries from sync queue for data sets: ${dataSetNames.join(', ')}`
       );
+    }
+
+    // if we're in the syncAllDatasets backoff sleep, remember these datasets so we skip them after the sleep
+    if (this.syncAllBackoffType !== SyncAllBackoffType.NONE) {
+      for (const name of dataSetNames) {
+        this.dataSetsSyncedDuringBackoff.add(name);
+      }
     }
 
     for (const name of dataSetNames) {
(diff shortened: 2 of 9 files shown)


packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1453,37 +1475,102 @@ class HashTreeParser {
...
+    await sleep(delay);
+
+    // read the (possibly upgraded) scope and clear the backoff flag
+    const effectiveBackoffType = this.syncAllBackoffType;
+    const skippedDataSets = this.dataSetsSyncedDuringBackoff;
+
+    this.syncAllBackoffType = SyncAllBackoffType.NONE;
+    this.dataSetsSyncedDuringBackoff = new Set();
+
+    if ((this.state as string) === 'stopped') return;
```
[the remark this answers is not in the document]
marcin-bazyl: it's not worth adding code to handle this edge case

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1453,37 +1475,102 @@ class HashTreeParser {
...
 
+    // delay the start of the syncs - this is a Locus requirement to avoid thundering herd issues
+    await sleep(delay);
+
+    // read the (possibly upgraded) scope and clear the backoff flag
+    const effectiveBackoffType = this.syncAllBackoffType;
+    const skippedDataSets = this.dataSetsSyncedDuringBackoff;
+
+    this.syncAllBackoffType = SyncAllBackoffType.NONE;
+    this.dataSetsSyncedDuringBackoff = new Set();
```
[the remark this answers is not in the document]
marcin-bazyl: this is not an issue

jor-row (approved):

packages/@webex/plugin-meetings/src/hashTree/constants.ts
```diff
@@ -17,3 +17,10 @@ export const DataSetNames = {
 // participant object for webinar attendees. If SELF were initialized first, locus.info
 // would not yet be populated and the attendee participant would be skipped.
 export const DATA_SET_INIT_PRIORITY: string[] = [DataSetNames.MAIN, DataSetNames.SELF];
+
+// Data sets for which we normally receive events over LLM connection
+export const LLM_DATASET_NAMES = [
```
chrisadubois, line 22: 1, I prefer enums
2, can we add a comment on each of these here? 

generally, for constants I want to know what each of them are with some form of an explanation. At first glance, I have NO idea what ATD_ACTIVE is

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -4,10 +4,16 @@ import LoggerProxy from '../common/logs/logger-proxy';
...
-import {DataSetNames, DATA_SET_INIT_PRIORITY, EMPTY_HASH} from './constants';
+import {DataSetNames, DATA_SET_INIT_PRIORITY, EMPTY_HASH, LLM_DATASET_NAMES} from './constants';
 import {ObjectType, HtMeta, HashTreeObject} from './types';
 import {LocusDTO, LocusErrorCodes} from '../locus-info/types';
-import {deleteNestedObjectsWithHtMeta, isMetadata, sortByInitPriority} from './utils';
+import {deleteNestedObjectsWithHtMeta, isMetadata, sleep, sortByInitPriority} from './utils';
+
+export enum SyncAllBackoffType {
+  NONE = 'none',
+  ONLY_LLM = 'onlyLLM',
```
chrisadubois, line 14: this one cracks me up.

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1377,6 +1392,13 @@ class HashTreeParser {
       );
     }
 
+    // if we're in the syncAllDatasets backoff sleep, remember these datasets so we skip them after the sleep
+    if (this.syncAllBackoffType !== SyncAllBackoffType.NONE) {
```
chrisadubois, line 1407: Can we put this (and the lower loop) into their own individual functions and call them individually? it makes testing their behavior quite a bit more straightforward

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1453,37 +1475,102 @@ class HashTreeParser {
...
     if (this.state === 'stopped') return;
-    if (this.isSyncAllInProgress) return;
 
-    this.isSyncAllInProgress = true;
-    try {
-      const dataSetsWithHashTrees = Object.values(this.dataSets)
-        .filter((dataSet) => dataSet?.hashTree)
-        .map((dataSet) => ({name: dataSet.name}));
+    // if we're already in the backoff delay phase, try to upgrade the scope instead of starting a new one
+    if (this.syncAllBackoffType !== SyncAllBackoffType.NONE) {
```
chrisadubois: Can this be split out to it's own individual function? we could call it `if (shortCircuitSyncAllDatasets(...))`

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1453,37 +1475,102 @@ class HashTreeParser {
...
+
+      return;
+    }
+
+    const dataSetsToSync = this.getSortedDataSetsWithHashTrees(onlyLLM);
 
-      const sorted = sortByInitPriority(dataSetsWithHashTrees, DATA_SET_INIT_PRIORITY);
+    if (dataSetsToSync.length === 0) return;
+
+    this.syncAllBackoffType = onlyLLM ? SyncAllBackoffType.ONLY_LLM : SyncAllBackoffType.ALL;
```
chrisadubois: can you make this a function? `setSyncAllbackOffTypeFromProperty(onlyLLM)`

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1453,37 +1475,102 @@ class HashTreeParser {
...
+    await this.syncQueueProcessingPromise;
+  }
+
+  /**
+   * Returns the list of data sets that have hash trees, sorted by the priority order provided by sortByInitPriority().
+   *
+   * @param {boolean} onlyLLM - Whether to include only LLM based data sets
+   * @returns {Array<{name: string, backoff: {maxMs: number, exponent: number}}>} The sorted list of data sets with their backoff configurations
+   */
+  private getSortedDataSetsWithHashTrees(onlyLLM: boolean) {
```
chrisadubois, line 1608: 👍

packages/@webex/plugin-meetings/src/hashTree/utils.ts
```diff
@@ -77,3 +77,22 @@ export function sortByInitPriority<T extends {name: string}>(items: T[], priorit
...
   return [...prioritized, ...rest];
 }
+
+/**
+ * Sleeps for the specified amount of milliseconds
+ *
+ * @param {number} ms amount of milliseconds to sleep
+ * @returns {Promise<void>} A promise that resolves after the specified delay
+ */
+export function sleep(ms: number): Promise<void> {
```
chrisadubois, line 87: `export const sleep = (ms: number) => ms > 0 ? new Promise<void>(r => setTimeout(r, ms)) : Promise.resolve();`

chrisadubois (approved):
please see comments, nothing blocking

packages/@webex/plugin-meetings/src/hashTree/constants.ts
marcin-bazyl: 1. Enum is not the right type here. Enum is a type you use when you want to narrow down a set of possible values to a predefined list. Here we have a list of dataset names that we want to search against. I agree that the already existing `const DataSetNames` could be an enum, but not the `LLM_DATASET_NAMES`
2. the comments are already there for each dataset name where they are defined:
image

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: done, but I don't agree that this approach makes testing always more straightforward - the new methods are private and tests shouldn't know about their existence, tests should only call and stub public methods

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: done, but I named the new method `tryUpgradePendingBackoff()`.

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: done, although I don't feel like this makes the code easier to read as you have to jump to the function to see what it does and it's just a very simple 1 line

packages/@webex/plugin-meetings/src/hashTree/utils.ts
marcin-bazyl: changing it to a 1-liner like this causes a warning:
`Return values from promise executor functions cannot be read.eslint[no-promise-executor-return](https://eslint.org/docs/latest/rules/no-promise-executor-return)`
so we have to wrap the call to setTimeout() with `{ }` and due to formatting this then ends up being multi-line function anyway, so the whole thing is not worth it and that's why I've ended up with it in the current form

PR #4906, opened 2026-04-23
feat(calling): mobius socket 

rsarika:
# COMPLETES #< [CAI-7692](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7692) >

## This pull request addresses

Adds the `mobius-socket` transport layer for signaling/communication with `Mobius`.

## by making the following changes

- Adds a socket layer for connection with mobius socket, keeps this feature behind a feature toggle.
- Maintains the backward compatibility with http/mercury

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

## The following scenarios were tested

Tested the registration and call features with both http/mercury and mobius-socket transport.

## The GAI Coding Policy And Copyright Annotation Best Practices ##
 

 
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [ ] GAI was used to create a draft that was subsequently customized or modified
- [x] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [ ] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [ ] Github Copilot
  - [x] Other - Claude, Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

- [ ] I have read and followed [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request)
- [ ] I discussed changes with code owners prior to submitting this pull request
- [ ] I have not skipped any automated checks
- [ ] All existing and new tests passed
- [ ] I have updated the documentation accordingly

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

PR #4995, opened 2026-05-20
test(calling): add mobius socket WebSocket E2E test infrastructure

eigengravy:
# COMPLETES [CAI-7883](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7883)

The Calling SDK Playwright E2E suite needs to exercise flows over the new mobius WebSocket transport (introduced in #4906 / #4982) in addition to the existing HTTP transport. Without a way to intercept and mock mobius WS frames, the existing registration, keepalive, call lifecycle, and error-flow tests cannot be re-run against the WS transport, leaving the new socket path uncovered by E2E.

- Added `packages/calling/playwright/utils/mobius-ws.ts` — a reusable WebSocket route interceptor for mobius frames (auth/register/unregister/device_status/call_*) with hooks for request/response mocking, plus tracking-id correlation.
- Extended `playwright/utils/setup.ts` and `test-manager.ts` to support running suites in mobius WS mode via env, including conditional WS route installation per worker.
- Updated all calling test groups (`registration-lifecycle`, `registration-failover`, `registration-keepalive`, `registration-errors`, `call-lifecycle`, `call-controls`, `call-keepalive`, `call-errors`) to drive the same scenarios over the WS transport using the new interceptor.
- Added the necessary selectors / test-data plumbing for the WS-mode runs and tweaked `oauth.setup.ts` to be transport-agnostic.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [x] Internal code refactor

- Ran the calling Playwright suites locally in mobius WS mode (`MOBIUS_TRANSPORT=ws` via env) covering:
  - SDK init project (CALLER + CALLEE + TRANSFER) in parallel.
  - SET_CALLER lifecycle (REG-001/002/008/010/014) and error flows (REG-011/012/013).
  - SET_CALLEE failover chain (REG-006 → REG-007).
  - SET_TRANSFER keepalive (REG-003/004/005).
  - Call lifecycle, controls, keepalive, and error flows.
- Re-ran the same suites in HTTP mode to confirm no regression on the existing transport path.
- Manually exercised the mobius WS switch in the calling sample app to validate the interceptor's frame shapes against real traffic.

  - [ ] Github Copilot
  - [x] Other - Please Specify
    - Claude Code (Anthropic)
- [x] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [x] Automation

### I certified that

---


packages/calling/playwright/test-groups/call-controls.ts
```diff
@@ -12,6 +12,130 @@ import {
...
+      if (!tm.page.isClosed()) {
+        await tm.page.waitForTimeout(3000);
+      }
+    });
+
+    test.afterAll(async () => {
+      await tm.cleanup();
+    });
+
+    test('CALL-030: Mute and unmute - toggle mute during active call', async () => {
```
Shreyas281299: What this CALL-030? I see each test has some CALL-XXX.

packages/calling/playwright/test-groups/call-controls.ts
```diff
@@ -255,12 +256,28 @@ export function callHoldErrorTests() {
...
+            onRequest: (frame) =>
+              failResume && frame.type === MOBIUS_WS_MESSAGE.CALL_RESUME
+                ? {
+                    statusCode: 500,
+                    statusMessage: 'Internal Server Error',
+                    data: {message: 'Resume failed'},
+                  }
+                : undefined,
+          })
+        : undefined;
```
rarajes2: Let's avoid using nested ternary operator. Applicable to other places as well

packages/calling/playwright/test-groups/call-keepalive.ts
```diff
@@ -280,17 +351,20 @@ export function callKeepaliveTests() {
...
-      let statusRequestCount = 0;
-      await callerPage.route('**/calls/**/status', (route) => {
-        statusRequestCount += 1;
-        route.fulfill({
-          status: 500,
-          headers: {'retry-after': '1'},
-          contentType: 'application/json',
-          body: JSON.stringify({error: 'Internal Server Error'}),
+      if (mobiusWsMode) {
+        failStatus = true;
```
rarajes2, line 361: Q. How are we using this value ?

packages/calling/playwright/test-groups/call-lifecycle.ts
```diff
@@ -40,7 +40,7 @@ export function callLifecycleTests() {
           media: true,
         }),
       ]);
-      calleeNumber = getPhoneNumber(tm.userSet.accounts[1]);
+      calleeNumber = getPhoneNumber(tm.userSet.accounts[1], tm.isInt);
```
rarajes2: nitpick: Instead of using `tm`, will be better to use `timeManager` or `taskManager` --- you can see how `tm` gives two different meaning in our context

packages/calling/playwright/test-groups/registration-errors.ts
```diff
@@ -12,17 +14,41 @@ export function registrationErrorTests() {
...
+              return {
+                statusCode: 401,
+                statusMessage: 'Unauthorized',
+                data: {message: 'Unauthorized'},
+              };
+            }
+
+            return undefined;
+          },
+        }).install(context);
```
rarajes2: The callback `onRequest` can be defined outside. Defining this inside makes it less readable

please see if this pattern is repeated.

packages/calling/playwright/test-groups/registration-failover.ts
```diff
@@ -30,58 +36,111 @@ export function registrationFailoverTests() {
...
+    const mobiusWsMode = isMobiusWsMode();
 
     test.beforeAll(async ({browser}, testInfo) => {
       const isInt = isIntProject(testInfo.project.name);
       expectedPrimaryUrl = isInt ? PRIMARY_MOBIUS_URL.INT : PRIMARY_MOBIUS_URL.PROD;
       expectedBackupUrl = isInt ? BACKUP_MOBIUS_URL.INT : BACKUP_MOBIUS_URL.PROD;
 
       tm = new TestManager(testInfo.project.name);
+      const interceptor = mobiusWsMode
+        ? new MobiusWsInterceptor({
```
rarajes2: Let's not use ternary operator for this type of code, makes difficult to read

packages/calling/playwright/test-groups/registration-lifecycle.ts
```diff
@@ -32,44 +39,68 @@ export function registrationLifecycleTests() {
...
     let expectedPrimaryUrl: string;
+    let mobiusWsInterceptor: MobiusWsInterceptor | undefined;
+    const mobiusWsMode = isMobiusWsMode();
 
     test.beforeAll(async ({browser}, testInfo) => {
       const isInt = isIntProject(testInfo.project.name);
       expectedPrimaryUrl = isInt ? PRIMARY_MOBIUS_URL.INT : PRIMARY_MOBIUS_URL.PROD;
       tm = new TestManager(testInfo.project.name);
+      if (mobiusWsMode) {
+        mobiusWsInterceptor = new MobiusWsInterceptor({
```
rarajes2, line 50: Do you think making this interceptor a singleton will make it easy and performant to use ? 

We can create the interceptor once and then we can add the `onResponse()` through a setter based on the test. After each test we can reset the interceptor (clear the existing `onResponse()` )

packages/calling/playwright/utils/mobius-ws.ts
```diff
@@ -0,0 +1,199 @@
...
+      route.onMessage(async (message) => {
+        const frame = parseFrame(message);
+
+        if (!frame?.type) {
+          server.send(message);
+
+          return;
+        }
+
+        const requestCount = MobiusWsInterceptor.increment(this.requestCounts, frame.type);
```
rarajes2, line 109: Q. What is the purpose of increment and count? How are we utilizing this ? Just trying to understand the logic behind request/response

packages/calling/playwright/utils/mobius-ws.ts
```diff
@@ -0,0 +1,199 @@
...
+  constructor(options: MobiusWsInterceptorOptions = {}) {
+    this.options = options;
+  }
+
+  async install(context: BrowserContext): Promise<void> {
+    await context.routeWebSocket(MOBIUS_WS_ROUTE, async (route: WebSocketRoute) => {
+      const server = route.connectToServer();
+      const url = route.url();
+
+      route.onMessage(async (message) => {
```
rarajes2, line 100: Q. What is the difference between `route.onMessage()` and `server.onMessage()` ?

packages/calling/src/mobius-socket/errors.ts
```diff
@@ -29,6 +29,9 @@ export class ConnectionError extends Exception {
    * @returns The reason string from the event
    */
   parse(event: SocketCloseEvent = {}) {
+    this.code = event.code;
+    this.reason = event.reason;
+
     return event.reason;
   }
```
rarajes2, line 33: This change was already done earlier I think. @rsarika Please verify if this is alright

Shreyas281299 (commented):
Submitting existing pending review to allow fresh inline comments.

packages/calling/playwright/test-groups/call-controls.ts
```diff
@@ -12,6 +12,130 @@ import {
   cleanupActiveCalls,
 } from '../utils/call';
 import {CALLING_SELECTORS, AWAIT_TIMEOUT} from '../constants';
+import {MOBIUS_WS_MESSAGE, MobiusWsInterceptor} from '../utils/mobius-ws';
+
+/**
+ * Call control tests: mute/unmute, DTMF, network flap.
+ * Fresh contexts — isolated from hold and error groups.
+ */
+export function callControlTests() {
```
Shreyas281299: **[Critical / Bug]** `export function callControlTests()` is declared **twice** in this file — here at line 21 and again at line 528. TypeScript rejects duplicate function implementations (TS2393), and at runtime the second declaration overwrites the first, so the three new tests added here (`CALL-030` mute, `CALL-008` DTMF, `CALL-016` network-flap) at lines 21–141 never register with the runner.

Either delete the second `callControlTests` block (lines 528–end) or merge the two into one (and pick a single set of test IDs). Run `npx tsc --noEmit` against the calling playwright tsconfig to confirm no duplicate-symbol errors remain.

packages/calling/playwright/test-groups/registration-errors.ts
```diff
@@ -12,17 +14,41 @@ export function registrationErrorTests() {
...
+            const response = await route.fetch();
+            registrationStatus = response.status();
+            await route.fulfill({response});
+          } else {
+            await route.continue();
+          }
+        });
+      }
 
       await navigateToCallingApp(page);
```
Shreyas281299, line 34: **[High / Test correctness]** REG-011 in WS mode passes vacuously. The interceptor is installed above (lines 19–39), but `setMobiusWebSocket(page, true)` is never called and this test deliberately bypasses `initializeCallingSDK`. The sample-app's `localStorage` flag stays unset, the SDK falls back to HTTP, the interceptor sees zero traffic, and `registrationPosts` stays at `0`. The only WS-validating assertion (`if (registrationPosts > 0) expect(registrationStatus).toBe(401)`) is therefore skipped, and the remaining assertions ("not registered") pass regardless of transport with an invalid token — so `MOBIUS=ws` runs give zero coverage of the WS 401 path.

Suggestion: enable WS explicitly in the WS branch, then make the assertion unconditional:

```ts
await navigateToCallingApp(page);
await setServiceIndicator(page, 'calling');
if (mobiusWsMode) {
  await setMobiusWebSocket(page, true);
}
// ...
if (mobiusWsMode) {
  expect(registrationPosts).toBeGreaterThan(0);
  expect(registrationStatus).toBe(401);
} else if (registrationPosts > 0) {
  expect(registrationStatus).toBe(401);
}
```

packages/calling/playwright/utils/mobius-ws.ts
```diff
@@ -0,0 +1,199 @@
...
+          route.send(message);
+
+          return;
+        }
+
+        const responseType = frame.subtype || frame.type || 'unknown';
+        const responseCount = MobiusWsInterceptor.increment(this.responseCounts, responseType);
+        this.responses.push({...frame, url});
+
+        const transformedFrame = await this.options.onResponse?.(frame, {
```
Shreyas281299: **[Medium / Reliability]** `await this.options.onRequest?.(...)` (line 106) and `await this.options.onResponse?.(...)` (line 139) have no try/catch. If a test's callback throws — e.g., a buggy expectation in a closure or an unexpected frame shape — the WebSocketRoute handler rejects, the proxy stops forwarding, and the SDK's pending-response timer eventually fires with an opaque "Mobius socket request timeout" instead of the original assertion error. Tests become flaky and very hard to debug.

Wrap both callbacks and fall through to passthrough on throw:

```ts
let mockedResponse;
try {
  mockedResponse = await this.options.onRequest?.(frame, ctx);
} catch (err) {
  console.error('MobiusWsInterceptor onRequest threw, passing frame through', err);
  server.send(message);
  return;
}
```

Same pattern for `onResponse`, with `route.send(message)` as the passthrough.

packages/calling/playwright/test-groups/call-errors.ts
```diff
@@ -56,9 +71,11 @@ export function callErrorTests() {
       await verifyLineRegistered(page);
       await getMediaStreams(page);
 
-      await context.route(/\/devices\/[^/]+\/call$/, (route) => {
-        route.abort('failed');
-      });
+      if (!mobiusWsMode) {
+        await context.route(/\/devices\/[^/]+\/call$/, (route) => {
+          route.abort('failed');
```
Shreyas281299: **[Medium / Test correctness]** CALL-013 exercises different failure modes in the two transports under one ID. The WS branch returns a structured `{statusCode: 500, ...}` response for `CALL_SETUP` — the SDK gets a server-side error. The HTTP branch uses `route.abort('failed')` — a transport-level network error (no response, fetch rejection). These hit different code paths in the SDK (`response error` vs `transport error`). The "no stuck call" assertion is true in both cases, but the test isn't validating equivalent behavior across transports.

Pick one failure mode and mirror it: either drop the WS connection / never reply on the WS side, or use `route.fulfill({status: 500, ...})` on the HTTP side.

packages/calling/playwright/utils/mobius-ws.ts
```diff
@@ -0,0 +1,199 @@
...
+
+        if (!frame) {
+          route.send(message);
+
+          return;
+        }
+
+        const responseType = frame.subtype || frame.type || 'unknown';
+        const responseCount = MobiusWsInterceptor.increment(this.responseCounts, responseType);
+        this.responses.push({...frame, url});
```
Shreyas281299: **[Medium / Security]** `this.requests.push({...frame, url})` (line 104) and `this.responses.push({...frame, url})` (line 137) capture the entire frame, including `metadata.authorization` — Mobius supplementary-service WS messages put the user's access token there (see `packages/calling/src/CallingClient/registration/request.ts`). The Playwright config uses `trace: 'retain-on-failure'`, so failed-test traces would persist live tokens, and CI typically uploads those artifacts.

Redact authorization before pushing:

```ts
const redact = (frame) => {
  if (!frame?.metadata?.authorization) return frame;
  return {...frame, metadata: {...frame.metadata, authorization: '[REDACTED]'}};
};
this.requests.push({...redact(frame), url});
```

packages/calling/playwright/test-data.ts
```diff
@@ -71,11 +91,16 @@ export const getToken = (role: AccountRole, isInt = false): string => {
   return token;
 };
 
+/** Env var for E.164 (or test) phone number, production vs integration Playwright projects. */
+export const phoneEnvVar = (role: AccountRole, isInt = false): string =>
```
Shreyas281299, line 95: **[Medium / Breaking change]** `phoneEnvVar(role, true)` resolves to `USER_N_INT_PHONE_NUMBER` — a brand-new env var name. The old code used only `${role}_PHONE_NUMBER`. INT projects are still active (per `playwright.config.ts:116` — `SET_CALL - INT`), so existing CI envs and developer `.env` files that only set `USER_N_PHONE_NUMBER` will hit `Error: USER_2_INT_PHONE_NUMBER not set.` immediately on the first INT run after this lands. The new var is also undocumented in any README or `.env.example`.

Suggestion: fall back to the prod var when the INT-specific one isn't set, and add the new vars to the package README / `.env.example`:

```ts
export const getPhoneNumber = (role: AccountRole, isInt = false): string => {
  const intVar = `${role}_INT_PHONE_NUMBER`;
  const prodVar = `${role}_PHONE_NUMBER`;
  const number = (isInt && process.env[intVar]) || process.env[prodVar];
  if (!number) {
    throw new Error(`${isInt ? intVar : prodVar} not set.`);
  }
  return number;
};
```

packages/calling/playwright/test-groups/call-controls.ts
eigengravy: Done — replaced all nested ternary interceptor creations with if/else blocks across call-controls, call-keepalive, and registration-failover. Also unwrapped the nested ternary inside onRequest callbacks into explicit if/return statements.

packages/calling/playwright/test-groups/registration-failover.ts
eigengravy: Done — converted to if/else block.

packages/calling/playwright/utils/mobius-ws.ts
eigengravy: Addressed — both onRequest and onResponse callbacks are now wrapped in try/catch. On throw, the error is logged and the frame passes through to the server/client unchanged.

packages/calling/playwright/utils/mobius-ws.ts
eigengravy: Addressed — added a redactFrame() helper that strips metadata.authorization before pushing to the requests/responses arrays. Frames are redacted in all capture sites.

packages/calling/playwright/test-groups/call-controls.ts
eigengravy: It's just test numbers with respect to the test plan document.

packages/calling/playwright/test-groups/call-lifecycle.ts
eigengravy: Done — renamed `tm` to `testManager` across all test group files.

packages/calling/playwright/test-groups/registration-errors.ts
eigengravy: Done — extracted the onRequest callback into a named variable before passing it to MobiusWsInterceptor.

packages/calling/playwright/test-groups/call-keepalive.ts
eigengravy: This flag gates the interceptor's onRequest callback — when `failStatus = true`, any `CALL_STATUS` frame hitting the interceptor returns a 500 (simulating a keepalive failure). The value is toggled from the test body after the call is established, so the interceptor only starts failing keepalives at the right moment.

packages/calling/playwright/utils/mobius-ws.ts
eigengravy: `requestCount`/`responseCount` track how many frames of each type have been seen so far. They're exposed to test callbacks via the `RouteContext` parameter so tests can make decisions based on ordinal position — e.g. "fail only the 1st REGISTER request" or "return 500 on the 2nd CALL_STATUS". They're also readable via `interceptor.getRequestCount(type)` for assertions like `expect(interceptor.getRequestCount('register')).toBeGreaterThan(initialCount)`.

packages/calling/playwright/utils/mobius-ws.ts
eigengravy: `route.onMessage()` handles messages from the **browser → server** direction (i.e. the SDK sending requests like register/unregister/call_setup). `server.onMessage()` handles messages from the **server → browser** direction (i.e. Mobius sending responses/events back to the SDK). Together they form a bidirectional proxy — we can intercept/mock in either direction.

packages/calling/playwright/test-groups/call-controls.ts
eigengravy: This was from an earlier merge commit that caused the duplicate. Already fixed.

packages/calling/src/mobius-socket/errors.ts
eigengravy: Yup this was already fixed in commit efaad8bc .

packages/calling/playwright/test-data.ts
eigengravy: The INT env var separation isn't just about phone numbers — the INT accounts use entirely different credentials (login/tokens), so a simple fallback to the prod var wouldn't work. The INT-specific vars are intentionally distinct and need to be set up in CI/`.env` alongside the INT account credentials.

packages/calling/playwright/test-groups/registration-errors.ts
eigengravy: Investigated — in WS mode, the SDK fails at HTTP service discovery (region/server lookup returns 401) before it ever opens a WebSocket connection. The invalid token `'invalid-token-12345'` gets rejected at the HTTP layer, so the WS interceptor was dead code that never fired.

Removed the WS interceptor, kept the test running in both modes (it still validates the "not registered" end state via UI assertions), and scoped the network-level `registrationPosts`/401 assertion to HTTP mode only. Added a comment documenting why.

packages/calling/playwright/test-groups/call-errors.ts
eigengravy: Valid — aligned the HTTP branch to use `route.fulfill({status: 500, body: ...})` instead of `route.abort('failed')` so both transports now exercise the same failure mode (server-side 500 on call setup). Both still validate the same end state: no stuck call objects after cleanup.

rarajes2 (approved):
LGTM. We can discuss the singleton pattern for the interceptor in person, it's not a blocker for this PR.

Shreyas281299 (approved):

PR #5005, opened 2026-05-26
feat(auth): refactor some auth helpers to handle CSRF and state from thirdparty flows

robstax:
new endpoint from idbroker now accepts state and mirrors it back. these functions have been added/refactored to handle that new state and also manage the CSRF token.

refactor and add auth helpers

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor
 

 
  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+   *
+   * @instance
+   * @memberof AuthorizationBrowserFirstParty
+   * @param {string} [search] - Optional query string (defaults to the
+   *   window's current `location.search`). Leading `?` is tolerated.
+   * @returns {Promise<{idToken: string|undefined, email: string|undefined,
+   *   error: string|undefined, state: Object}>}
+   */
+  handleThirdPartyCallback(search) {
+    return new Promise((resolve) => {
```
martroos: question: Why does this return a promise? It doesn't appear to do anything asynchronous that would need `resolve`-ing.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+
+      let parsedState = {};
+      if (query.state) {
+        try {
+          parsedState = JSON.parse(base64.decode(query.state));
+        } catch (err) {
+          throw new Error('Unable to decode third-party callback state');
+        }
+        query.state = parsedState;
+      }
```
martroos: nitpick: Might be worth having a common function for doing this, almost identical code exists in `initialize`.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+      const rawSearch =
+        typeof search === 'string' ? search : (win.location && win.location.search) || '';
+      const query = querystring.parse(rawSearch.replace(/^\?/, ''));
+
+      let parsedState = {};
+      if (query.state) {
+        try {
+          parsedState = JSON.parse(base64.decode(query.state));
+        } catch (err) {
+          throw new Error('Unable to decode third-party callback state');
```
martroos: suggestion: This completely masks the original error - just allow the error to happen, we don't expect it to happen, and throwing a different error can make it harder to debug if it did happen.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+        }
+        query.state = parsedState;
+      }
+
+      this._verifySecurityToken(query, {requireMatch: true});
+
+      const location = url.parse(win.location.href, true);
```
martroos: nitpick: `location` is a global keyword in the browser... other places in this do the exact same thing so not too concerned about changing this though.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+      // Replace freshly-parsed state with already-decoded version so
+      // _cleanUrl re-encodes correctly without csrf_token.
+      location.query.state = parsedState;
+      Reflect.deleteProperty(location.query, 'id_token');
+      Reflect.deleteProperty(location.query, 'email');
```
martroos: suggestion: Add these to the `_cleanUrl` method - I expect these should always be removed, and if they don't exist already, it won't do any harm.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -843,19 +895,28 @@ const Authorization = WebexPlugin.extend({
...
-   * If no stored token (e.g., user navigated directly), silently returns.
+   * If no stored token (e.g., user navigated directly), silently returns
+   * unless `options.requireMatch` is `true`, in which case absence of a
+   * stored token is treated as a CSRF failure.
    *
    * @instance
    * @memberof AuthorizationBrowserFirstParty
    * @param {Object} query - Parsed query (location.query)
+   * @param {Object} [options]
+   * @param {boolean} [options.requireMatch=false] - When true, throws if
```
martroos, line 891: question: Is there a reason not to make this true always (i.e. not have the option for it at all)? Do we have cases where we want to ignore csrf if we haven't got one stored?

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+   * @param {string} [search] - Optional query string (defaults to the
+   *   window's current `location.search`). Leading `?` is tolerated.
+   * @returns {Promise<{idToken: string|undefined, email: string|undefined,
+   *   error: string|undefined, state: Object}>}
+   */
+  handleThirdPartyCallback(search) {
+    return new Promise((resolve) => {
+      const win = this.webex.getWindow();
+      const rawSearch =
+        typeof search === 'string' ? search : (win.location && win.location.search) || '';
```
martroos: issue: There are no unit tests covering the case where the search ends up being `''` - all tests have either search passed in, or one in the window location.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -393,6 +395,56 @@ const Authorization = WebexPlugin.extend({
...
+      const query = querystring.parse(rawSearch.replace(/^\?/, ''));
+
+      let parsedState = {};
+      if (query.state) {
+        try {
+          parsedState = JSON.parse(base64.decode(query.state));
```
martroos: nitpick: I suspect there are ways of making the `base64.decode` fail, which isn't covered by unit tests, but the general error handling is covered, so not too concerned.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -627,6 +663,131 @@ describe('plugin-authorization-browser-first-party', () => {
...
+            state: {popUpSignIn: true, mode: 'meeting'},
+          });
+          // Stored token consumed
+          assert.calledWith(webex.getWindow().sessionStorage.removeItem, 'oauth2-csrf-token');
+          // URL scrubbed: no id_token/email/csrf_token left
+          assert.called(webex.getWindow().history.replaceState);
+          const cleaned = webex.getWindow().location.href;
+
+          assert.notInclude(cleaned, 'id_token');
+          assert.notInclude(cleaned, 'email=');
```
martroos: nitpick: Why not `'email'` here or why not add the `'='` for others?

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -627,6 +663,131 @@ describe('plugin-authorization-browser-first-party', () => {
...
+          assert.notInclude(cleaned, 'csrf_token');
```
martroos: nitpick: This would probably be better if the query string was parsed, to avoid issues of substring matching in future (if we added `redacted_email` in future for example, not sure if this is an actual use case of course!).

martroos: suggestion: I don't think `location.href` ever contain `'csrf_token'` directly. It's base64 encoded in the state, so I wouldn't expect this to fail, even if no cleaning was done.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -627,6 +663,131 @@ describe('plugin-authorization-browser-first-party', () => {
...
+        const encodedState = buildState({csrf_token: storedToken});
+        const search = `?id_token=id-1&state=${encodedState}`;
+        const href = `http://example.com/${search}`;
+        const webex = makeWebex(href, storedToken);
+        webex.getWindow().sessionStorage.getItem = sinon.stub().returns(storedToken);
+        webex.getWindow().location.search = search;
+
+        return webex.authorization.handleThirdPartyCallback().then((result) => {
+          assert.equal(result.idToken, 'id-1');
+        });
```
martroos: suggestion: We should be doing the same checks here as for the argument supplied version of the test.

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -627,6 +663,131 @@ describe('plugin-authorization-browser-first-party', () => {
...
+
+        return assert.isRejected(
+          webex.authorization.handleThirdPartyCallback(search),
+          /Unable to decode third-party callback state/
+        );
+      });
+
+      it('returns error value in error field and leaves it in the URL, with empty state', () => {
+        const storedToken = 'csrf-abc';
+        const encodedState = buildState({csrf_token: storedToken});
```
martroos: suggestion: I think it would be worth adding some other state to prove that this case does still return state correctly (whilst omitting the csrf token)

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -627,6 +663,131 @@ describe('plugin-authorization-browser-first-party', () => {
       });
     });
 
+    describe('#_verifySecurityToken() requireMatch', () => {
+      it('silently returns when no stored token and requireMatch is false', () => {
+        const webex = makeWebex();
+        webex.getWindow().sessionStorage.getItem = sinon.stub().returns(null);
+
+        assert.doesNotThrow(() => {
+          webex.authorization._verifySecurityToken({});
```
martroos: nitpick: Would be nice to check the return value here also.

packages/@webex/webex-core/src/lib/credentials/credentials.js
```diff
@@ -228,22 +228,35 @@ const Credentials = WebexPlugin.extend({
...
 
-    return `${this.config.thirdPartyLoginUrl}?${querystring.stringify({
+    const query = {
       oauth2provider,
       returnURL,
-    })}`;
+    };
+
+    if (state && !isEmpty(state)) {
+      query.state = base64.toBase64Url(JSON.stringify(state));
```
martroos: thought: It's unfortunate that the encoding is in a different place to the decoding... not the end of the world, but there is the possibility for these to be incompatible.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -299,6 +299,61 @@ describe('webex-core', () => {
...
+
+        webex.trigger('change:config');
+
+        const result = credentials.buildThirdPartyLoginUrl({
+          oauth2provider: 'google',
+          returnURL: 'https://web.webex.com',
+          state: {},
+        });
+
+        assert.notInclude(result, 'state=');
```
martroos: nitpick: As with the other test, it might be worth parsing the query string and checking it all.

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
```diff
@@ -299,6 +299,61 @@ describe('webex-core', () => {
...
+      });
+
+      skipInBrowser(it)('base64url-encodes a non-empty `state` and emits it as a top-level query param', () => {
+        const webex = new MockWebex();
+        const credentials = new Credentials(undefined, {parent: webex});
+
+        webex.trigger('change:config');
+
+        const state = {csrf_token: 'abc', popUpSignIn: true};
+        const encoded = base64.toBase64Url(JSON.stringify(state));
```
martroos: issue: This is testing the implementation with the implementation, it doesn't prove it did the right thing. Ideally this should be a literal expected string, assuming that JSON.stringify is stable between different machines. If not, we should at least try to spy on these to make sure they are being called, and then to use the returned value in the check below.

martroos (changes requested):

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -344,22 +344,24 @@ const Authorization = WebexPlugin.extend({
...
    * @param {string} options.oauth2provider
    * @param {string} options.returnURL
+   * @param {Object} [options.state] - Caller-supplied state object. Merged
+   *   with the generated `csrf_token`.
    * @returns {Promise<void>}
    */
   initiateThirdPartyLogin(options = {}) {
+    options = cloneDeep(options);
+    options.state = options.state || {};
+    options.state.csrf_token = this._generateSecurityToken();
```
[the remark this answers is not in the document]
robstax: fixed this

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: just bad on my part. i apologize.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: good suggestion. i made a new helper `packages/@webex/common/src/oauth-state.js` which centralizes the encode/decode logic

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: agreed. changed.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: ack - probably not going to change based on other examples in the repo

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: done! originally i had them there and then for some reason second guessed myself. put it back in cleanUrl

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: it's a good question. i can't think of a realistic example. that being said, i am a little bit apprehensive of changing the behavior to always expect one because that would be breaking change.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: i changed this implementation a bit and also added a test.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: ack

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: thanks. addressed and changed

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: ack. changed

robstax: yep. removed

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: i changed this test

packages/@webex/webex-core/src/lib/credentials/credentials.js
robstax: i added some helper functions to at least make it a bit more cohesive and consistent

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: nice suggestion. done

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
robstax: done

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
robstax: makes sense. changed

packages/@webex/webex-core/test/unit/spec/credentials/credentials.js
robstax: good call out. i think it should make the same expected string every time so i am trying that

packages/@webex/common/test/unit/spec/oauth-state.js
```diff
@@ -0,0 +1,34 @@
+/*!
+ * Copyright (c) 2015-2024 Cisco Systems, Inc. See LICENSE file.
+ */
+
+import {assert} from '@webex/test-helper-chai';
+import {decodeState, encodeState} from '@webex/common';
+
+describe('oauth-state', () => {
+  describe('encodeState() / decodeState()', () => {
+    it('round-trips a typical state object', () => {
```
martroos, line 36: suggestion: Whilst this is a useful test, it would also pass if both functions did nothing but return the input (in fact, most of the tests pass like this, the only ones that fail do so because of a side effect of what is returned / not thrown, not because the transformation didn't take place). I think it would be good to add in some encode / decode tests that use string literals.

packages/@webex/common/test/unit/spec/oauth-state.js
```diff
@@ -0,0 +1,34 @@
...
+      const state = {csrf_token: 'abc123', provider: 'google', returnURL: '/app'};
+
+      assert.deepEqual(decodeState(encodeState(state)), state);
+    });
+
+    it('round-trips an empty object', () => {
```
martroos, line 46: suggestion: Same comment here about checking encode and decode separately.

packages/@webex/common/test/unit/spec/oauth-state.js
```diff
@@ -0,0 +1,34 @@
...
+    });
+
+    it('produces a url-safe encoding (no +, /, or = padding)', () => {
+      // Inputs chosen so the resulting bytes would include +, / and padding under
+      // standard base64.
+      const encoded = encodeState({a: '???>>>'});
+
+      assert.notInclude(encoded, '+');
+      assert.notInclude(encoded, '/');
+      assert.notInclude(encoded, '=');
```
martroos, line 62: nitpick: These could be checked in all tests since they should always be web safe.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -396,53 +399,41 @@ const Authorization = WebexPlugin.extend({
...
+  handleThirdPartyCallback() {
+    const location = url.parse(this.webex.getWindow().location.href, true);
 
-      const location = url.parse(win.location.href, true);
-      // Replace freshly-parsed state with already-decoded version so
-      // _cleanUrl re-encodes correctly without csrf_token.
+    let parsedState = {};
+    if (location.query.state) {
+      parsedState = decodeState(location.query.state);
       location.query.state = parsedState;
```
martroos: nitpick: This could be replaced with:

```js
location.query.state = decodeState(location.query.state || 'e30');
```

Since `_cleanUrl` will remove an empty state anyway. This would mean that later we could get all the values  in one destructure (no need to store the parsed state):

```js
const {id_token: idToken, email, error, state: {csrf_token, ...state}} = location.query;
```

So the whole method boils down to:

```js
handleThirdPartyCallback() {
  const location = url.parse(this.webex.getWindow().location.href, true);

  location.query.state = decodeState(location.query.state || 'e30');

  this._verifySecurityToken(location.query, {requireMatch: true});
  this._cleanUrl(location);

  const {id_token: idToken, email, error, state: {csrf_token, ...state}} = location.query;

  return {idToken, email, error, state};
},
```

I realise this is just refactoring at this point, so completely up to you!

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -396,53 +399,41 @@ const Authorization = WebexPlugin.extend({
...
-
-      this._verifySecurityToken(query, {requireMatch: true});
+  handleThirdPartyCallback() {
+    const location = url.parse(this.webex.getWindow().location.href, true);
 
-      const location = url.parse(win.location.href, true);
-      // Replace freshly-parsed state with already-decoded version so
-      // _cleanUrl re-encodes correctly without csrf_token.
+    let parsedState = {};
+    if (location.query.state) {
```
martroos: suggestion: I don't think there is a success case that has a false-y state, might be nice to add this.

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
```diff
@@ -920,11 +912,7 @@ const Authorization = WebexPlugin.extend({
       return;
     }
 
-    if (!query.state) {
-      throw new Error(`Expected CSRF token ${sessionToken}, but not found in redirect query`);
-    }
-
-    if (!query.state.csrf_token) {
+    if (!query.state || !query.state.csrf_token) {
```
martroos: nitpick:

```js
if (!query.state?.csrf_token) {
```

packages/@webex/plugin-authorization-browser-first-party/test/unit/spec/authorization.js
```diff
@@ -1256,7 +1287,33 @@ describe('plugin-authorization-browser-first-party', () => {
...
+          },
+        };
+
+        webex.authorization._cleanUrl(location);
+        assert.called(webex.getWindow().history.replaceState);
+        const {href} = webex.getWindow().location;
+
+        assert.notInclude(href, 'id_token');
+        assert.notInclude(href, 'email');
+        assert.notInclude(href, 'code');
```
martroos, line 1326: nitpick: I think parsing the location would be better here, but this is overly brittle (i.e. it would fail even if we did expect `masked_email` or similar), so the test is still valid.

martroos (approved):

packages/@webex/common/test/unit/spec/oauth-state.js
robstax: good idea! done!

packages/@webex/common/test/unit/spec/oauth-state.js
robstax: done

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: i like it! done

packages/@webex/plugin-authorization-browser-first-party/src/authorization.js
robstax: done!

robstax:
int tests always fail on next even without code changes. https://github.com/webex/webex-js-sdk/actions/runs/26764247021/job/78887934260?pr=4966

chrisadubois (approved):

PR #5012, opened 2026-05-29
test(calling): added playwright tests for call settings

vivekv1504:
# COMPLETES #< https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7878 >

The Calling SDK CallSettingsClient APIs (Do Not Disturb, Call Forwarding, Voicemail, and Call Waiting) did not have end-to-end Playwright coverage against the calling sample app. Without automated tests, regressions in settings fetch/save, UI wiring, or live call behavior (e.g. DND blocking incoming calls, call forwarding routing) could go unnoticed.

This PR adds a dedicated SET_CALL_SETTINGS Playwright project and a full E2E suite that exercises these flows through the sample app UI and validates real call behavior with multiple registered users.

<h2 class="mt-6 mb-2 font-semibold text-2xl" data-streamdown="heading-2" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; --tw-font-weight: 600; font-weight: 600; line-height: 1.42; font-size: 1.428em; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">by making the following changes</h2><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Playwright test infrastructure</h3><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Added<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">SET_CALL_SETTINGS</code></span><span> </span>user set in<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">test-data.ts</span></code><span> </span>with three accounts:<ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">USER_3</code><span> </span>— settings owner / callee</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">USER_2</code><span> </span>— primary caller</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">USER_1</code><span> </span>— second caller (for multi-user scenarios)</li></ul></li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Registered<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">SET_CALL_SETTINGS - PROD</code></span><span> </span>and<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">SET_CALL_SETTINGS - INT</code></span><span> </span>projects in<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">playwright.config.ts</span></code><span> </span>(with OAuth dependencies).</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Added suite entry point:<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-prefix" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">playwright/suites/</span><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">set-call-settings.spec.ts</span></code>.</li></ul><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Reusable test utilities (<code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-prefix" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">utils/</span><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">call-settings.ts</span></code>)</h3><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">loadSettings()</code></span><span> </span>— clicks “Get Settings” and waits for both<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">callForwarding</code><span> </span>and<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">voicemail</code><span> </span>GET responses plus DND UI settle (avoids saving CF/VM before server state is loaded).</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">DND helpers</span><span> </span>—<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">clickDnd()</code>,<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">ensureDndState()</code><span> </span>with PUT wait +<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">expect.poll()</code><span> </span>to handle eventual consistency between write and read.</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Call Forward helpers</span><span> </span>—<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">setCallForwardAlways/Busy/NoAnswer/NotReachable()</code><span> </span>and<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">saveCfSettings()</code><span> </span>with PUT response waits.</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Voicemail helpers</span><span> </span>—<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">setVoicemailSendAllCalls/BusyCalls/UnansweredCalls()</code><span> </span>and<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">saveVoicemailSettings()</code>.</li></ul><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Selectors (<code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-prefix" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">constants/</span><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">selectors.ts</span></code>)</h3><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Added call-settings selectors for DND, Call Waiting, Call Forward, and Voicemail forms.</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Scoped selectors to<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">#callForwardForm</code><span> </span>and<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">#voicemailForm</code><span> </span>to avoid duplicate ID collisions in the sample app.</li></ul><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">UI tests (<code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">callSettingsTests</code>) — single user, serial</h3><div class="ui-scroll-area" data-scroll-padding="4" data-visibility="hover" data-direction="horizontal" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; --scrollbar-size: 6px; --scrollbar-inset: 0px; --scrollbar-thumb-top-offset: 6px; --scroll-area-scroll-padding: 4px; position: relative; overflow: hidden; display: grid; grid-template: 1fr / 1fr; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; margin: 1em 0px; border-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; border-radius: 6px; box-sizing: border-box; color: rgb(216, 216, 216); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="ui-scroll-area__viewport" style="grid-area: 1 / 1; min-height: 0px; max-height: 100%; border-radius: inherit; overflow: auto hidden; scroll-padding: 0px 4px; overscroll-behavior: contain auto; scrollbar-width: none !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); box-sizing: border-box;"><div class="ui-scroll-area__content" style="box-sizing: border-box; width: auto; min-width: 100%; max-width: 100%; min-height: 100%; display: flex; flex-direction: row; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; height: fit-content;">
Test ID | Coverage
-- | --
CS-GET-001 | Get Settings — all UI fields settle and become enabled
CS-DND-002 | DND enable/disable round-trip
CS-CF-003–006 | Call Forward Always, Busy, No Answer (+ rings), Not Reachable round-trips
CS-CFA-007 | Get Call Forward Always by directory number
CS-ERR-008 | DND PUT 500 — button text reverts
CS-VM-009–015 | Voicemail form load, toggles, numberOfRings, MWI, enabled, email notification, email copy

</div></div></div><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Each live-call test resets settings in<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">beforeAll</code><span> </span>/<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">finally</code><span> </span>blocks and uses<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">INCOMING_CALL_ANSWER_BTN</code><span> </span>to assert whether<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">line:incoming_call</code><span> </span>fired.</p><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Reliability fixes</h3><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Improved DND toggle handling to poll server state after PUT instead of relying on immediate UI updates.</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Tests restore original settings after each case to avoid polluting shared test accounts.</li></ul><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 600; margin-top: 16px; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Sample app</h3><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgb(216, 216, 216); scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box; font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 418; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.72); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">Minor fix in<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(123, 175, 233); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; cursor: pointer; box-sizing: border-box;"><span class="md-inline-path-prefix" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">docs/samples/calling/</span><span class="md-inline-path-filename" style="scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">index.html</span></code>: corrected voicemail “When not answered” rings input ID (<code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.847059 0.847059 0.847059 / 0.08); color: rgb(216, 216, 216); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all; scrollbar-color: rgba(228, 228, 228, 0.14) rgba(0, 0, 0, 0); scrollbar-width: thin; box-sizing: border-box;">vmNotAnsweredRings</code>).</li></ul><br class="Apple-interchange-newline">

Screenshot 2026-05-29 at 13 04 00
Screenshot 2026-05-29 at 13 04 26

### Change Type

- [ x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [x] Other - Cursor
- [x] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [x] Automation

### I certified that

---


packages/calling/playwright/test-data.ts
```diff
@@ -121,6 +121,14 @@ export const USER_SETS: Record<string, UserSet> = {
...
   },
 
+  // Call Settings E2E tests:
+  //   accounts[0] = USER_3 — settings owner / callee
+  //   accounts[1] = USER_2 — primary caller (also used for busy-state first call)
+  //   accounts[2] = USER_1 — second caller (places call while USER_3 is already busy)
+  SET_CALL_SETTINGS: {
+    name: 'SET_CALL_SETTINGS',
+    accounts: ['USER_3', 'USER_2', 'USER_1'],
+    testSuite: 'set-call-settings.spec.ts',
```
eigengravy, line 138: **CRITICAL: Missing closing `},` — this file won't parse**

The `SET_CALL_SETTINGS` object is missing its closing brace+comma. Playwright test discovery will fail immediately with a syntax error.

```diff
  SET_CALL_SETTINGS: {
    name: 'SET_CALL_SETTINGS',
    accounts: ['USER_3', 'USER_2', 'USER_1'],
    testSuite: 'set-call-settings.spec.ts',
+ },
  // Single-user Contacts supplementary service E2E tests
  SET_CONTACTS: {
```

packages/calling/playwright.config.ts
```diff
@@ -139,5 +139,36 @@ export default defineConfig({
...
+    // Single-user Call Settings tests
+    {
+      name: 'SET_CALL_SETTINGS - PROD',
+      dependencies: ['OAuth - PROD'],
+      testDir: './playwright/suites',
+      testMatch: USER_SETS.SET_CALL_SETTINGS.testSuite,
+      use: browserOptions[PW_BROWSER],
+    },
+    {
+      name: 'SET_CALL_SETTINGS - INT',
```
eigengravy, line 164: **Account isolation violation: SET_CALL_SETTINGS shares USER_1/2/3 with registration tests**

`SET_CALL_SETTINGS` uses accounts USER_1, USER_2, and USER_3 but only depends on `OAuth - PROD`. Since `SET_REGISTRATION_1/2/3` also depend only on OAuth and use these same accounts, Playwright can schedule them concurrently — violating the constraint at the top of test-data.ts:

> the same Webex account must NEVER be active in two browser sessions simultaneously

This needs `dependencies: ['SET_REGISTRATION_1 - PROD', 'SET_REGISTRATION_2 - PROD', 'SET_REGISTRATION_3 - PROD']` at minimum, or should wait until after `SET_CALL` (which already serializes through the registration projects).

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    // Call Forward Always — callee must not ring
+    // -----------------------------------------------------------------------
+
+    test('CS-CALL-101: CF Always enabled — callee does not receive the incoming call', async () => {
+      await loadSettings(calleePage);
+      await setCallForwardAlways(calleePage, true, cfDestination);
+      // Allow the CF setting to propagate to the server before placing the call.
+      await calleePage.waitForTimeout(5000);
+
+      try {
```
eigengravy, line 716: **Test logic gap: no caller-side verification that the call was actually placed**

CS-CALL-101 (and CS-CALL-102 below) only asserts the callee's answer button stays disabled. But if `makeCall()` silently fails (network error, invalid number, SDK exception), no call is created *and* the callee never rings — the test still passes.

You need a caller-side assertion to confirm the outbound call was actually established. For example, wait for the caller's call-state indicator to show an active/ringing call before starting the observation window on the callee side.

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    });
+    // Test for Call Forward When Busy: setting is saved and first call still rings.
+
+    test('CS-CALL-103: CF When Busy — setting persists and first call still reaches callee', async () => {
+      await loadSettings(calleePage);
+      await setCallForwardBusy(calleePage, true, cfDestination);
+
+      // Reload and verify the setting was persisted on the server.
+      await loadSettings(calleePage);
+      await expect(calleePage.locator(CALLING_SELECTORS.CF_BUSY_CB)).toBeChecked({
```
eigengravy, line 763: **Incomplete test: CS-CALL-103 never actually tests CF Busy forwarding**

The test name says "CF When Busy" but it only verifies:
1. The setting persists after reload
2. The *first* call rings normally

It never:
- Answers the first call (to make the callee busy)
- Places a second call from `secondCallerPage` (USER_1)
- Verifies the second call gets forwarded instead of ringing the callee

As written, this test passes even if CF Busy forwarding is completely broken. The title should either be renamed to reflect what it actually tests ("CF Busy setting persists and doesn't affect non-busy calls"), or the test should be completed to exercise the actual busy-forward path.

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    // CS-VM-014: Get notifications via email — toggle and email ID persist
+    // -----------------------------------------------------------------------
+
+    test('CS-VM-014: Voicemail — email notification toggle and email ID save and reload correctly', async () => {
+      await loadSettings(page);
+      const originalChecked = await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_CB).isChecked();
+
+      // Enable the checkbox so the email input is accessible.
+      if (!originalChecked) {
+        await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_CB).check({timeout: AWAIT_TIMEOUT});
```
eigengravy: **Incomplete restoration: if checkbox was originally ON, the test email is left in the account**

The restoration block only unchecks if the checkbox was originally off:
```typescript
if (!originalChecked) {
  await page.locator(...).uncheck(...);
}
await saveVoicemailSettings(page);
```

If the checkbox was already enabled, the test overwrites the real email with `autotest-notification@example.com` and never restores it. The account is left with a test email address as the configured notification destination.

Fix: capture `originalEmail` before the test, and always restore both the checkbox state *and* the email value:
```typescript
const originalEmail = await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_ID).inputValue();
// ... test ...
// Restore:
if (originalChecked) {
  await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_ID).fill(originalEmail || '');
} else {
  await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_CB).uncheck(...);
}
```

Same issue exists in CS-VM-015 below.

eigengravy (changes requested):
E2E test review — there's a parse error that blocks the entire suite, an account isolation violation, and several test logic gaps. Details inline.

packages/calling/playwright/test-data.ts
vivekv1504: Done chnages applied ,rebase the branch and appiled the changes

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: restoration now captures originalEmail via inputValue() before the test overwrites it. If the checkbox was originally on, we restore the prior email and save; if it was off, we uncheck and save. Applied the same fix to CS-VM-015.

packages/calling/playwright.config.ts
```diff
@@ -139,5 +139,36 @@ export default defineConfig({
...
+    //   testMatch: USER_SETS.SET_3USER.testSuite,
+    //   use: browserOptions[PW_BROWSER],
+    // },
+    // {
+    //   name: 'SET_3USER - INT',
+    //   dependencies: ['SET_2USER - INT'],
+    //   testDir: './playwright/suites',
+    //   testMatch: USER_SETS.SET_3USER.testSuite,
+    //   use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
+    // },
```
Shreyas281299: Please remove these commented lines

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Added waitForCallerOutboundCall() after makeCall() in CS-CALL-101 and CS-CALL-102. It waits for the caller’s SDK to report an active outbound call via getActiveCalls() before asserting the callee doesn’t ring, so the test fails if the dial never happened.

Shreyas281299 (commented):
Submitting accidental empty pending review.

packages/calling/playwright.config.ts
```diff
@@ -139,5 +139,36 @@ export default defineConfig({
       testMatch: USER_SETS.SET_CALL_TRANSFER_CONSULT.testSuite,
       use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
     },
+    // Single-user Call Settings tests
+    {
+      name: 'SET_CALL_SETTINGS - PROD',
```
Shreyas281299, line 146: SET_CALL_SETTINGS currently reuses USER_1/USER_2/USER_3, which are also used by SET_REGISTRATION_1/2/3. Since these projects only depend on OAuth and run in parallel, this violates the single-active-session rule in test-data and creates account-collision flakiness. Could we assign a dedicated user set to SET_CALL_SETTINGS (or add explicit project dependencies) so those sessions never overlap?

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    // Voicemail sendAllCalls — callee must not ring
+    // -----------------------------------------------------------------------
+
+    test('CS-CALL-105: VM sendAllCalls enabled — callee does not receive the incoming call', async () => {
+      await loadSettings(calleePage);
+      await setVoicemailSendAllCalls(calleePage, true);
+      // Allow the setting to propagate before placing the call.
+      await calleePage.waitForTimeout(5000);
+
+      try {
```
Shreyas281299, line 919: `cfDestination = getPhoneNumber('USER_4')` can collide with `SET_CALL - PROD`, where USER_4 is actively registered in parallel. That means forwarded calls from CS-CALL-101 may ring in the SET_CALL suite and perturb both projects. Can we route CF tests to a destination number not owned by any concurrently registered test account?

packages/calling/playwright/utils/call-settings.ts
```diff
@@ -0,0 +1,270 @@
...
+export async function setCallForwardAlways(
+  page: Page,
+  enable: boolean,
+  destination?: string
+): Promise<void> {
+  const cb = page.locator(CALLING_SELECTORS.CF_ALWAYS_CB);
+  const dest = page.locator(CALLING_SELECTORS.CF_ALWAYS_DEST);
+  const isChecked = await cb.isChecked();
+
+  if (enable && !isChecked) {
```
Shreyas281299, line 110: In `setCallForwardAlways` (and similarly at the Busy/NoAnswer/NotReachable helpers), when `enable === true` and the toggle is already checked, we skip filling `destination` and still save. That can silently persist a stale number from previous state. Consider adding an `enable && isChecked && destination` branch to always refresh the destination before `saveCfSettings`.

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    });
+
+    // -----------------------------------------------------------------------
+    // CS-VM-012: MWI notification badge toggle persists
+    // -----------------------------------------------------------------------
+
+    test('CS-VM-012: Voicemail — MWI notification badge toggle saves and reloads correctly', async () => {
+      await loadSettings(page);
+      const original = await page.locator(CALLING_SELECTORS.VM_MWI_CB).isChecked();
+
```
Shreyas281299, line 471: `page.unroute('**/features/doNotDisturb*', routeHandler)` is currently outside a `finally` block. If an assertion throws before the end, the route stub can leak and affect retries/subsequent tests in the same page context. Could we wrap the route setup in `try/finally` and always unroute in `finally`?

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+        await page.locator(CALLING_SELECTORS.VM_UNANSWERED_CB).uncheck({timeout: AWAIT_TIMEOUT});
+      } else {
+        await page.locator(CALLING_SELECTORS.VM_UNANSWERED_CB).check({timeout: AWAIT_TIMEOUT});
+      }
+      await saveVoicemailSettings(page);
+      await loadSettings(page);
+
+      if (original) {
+        await expect(page.locator(CALLING_SELECTORS.VM_UNANSWERED_CB)).not.toBeChecked({
+          timeout: OPERATION_TIMEOUT,
```
Shreyas281299, line 447: `expect(cfaDataElem).not.toBeEmpty()` is too weak here because both success and error paths populate the element (`callSetting` or `error`), so this can pass on regressions. Can we assert a semantic value instead (e.g., expected destination substring or parsed JSON fields) to prove the call-forward update actually succeeded?

packages/calling/playwright/utils/call-settings.ts
```diff
@@ -0,0 +1,270 @@
...
+
+// ---------------------------------------------------------------------------
+// Call Forward helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Click the Call Forward save button and block until the PUT is acknowledged
+ * by the server.  More reliable than a fixed waitForTimeout.
+ */
+export async function saveCfSettings(page: Page): Promise<void> {
```
Shreyas281299, line 87: The `waitForResponse` predicate here only checks URL + method, so 4xx/5xx responses still satisfy the waiter and appear as success. Would you consider asserting success status in the predicate (e.g. `r.status() < 400`) or checking `response.ok()` immediately after resolving?

Shreyas281299 (commented):
Posting inline findings from `/tmp/pr5012-review.md`.

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+        });
+      } finally {
+        await cleanupActiveCalls(callerPage);
+        await loadSettings(calleePage).catch(() => {});
+        await setVoicemailSendUnansweredCalls(calleePage, false);
+        await calleePage.waitForTimeout(3000);
+      }
+    });
+  });
+}
```
Shreyas281299, line 976: CS-CALL-104 currently enables CF No Answer without controlling `numberOfRings`, so behavior depends on account defaults and can exceed the 45s timeout before forwarding happens. Could we extend `setCallForwardNoAnswer` to accept rings and pass a deterministic value (for example `'2'`) in this test?

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+      if (!originalChecked) {
+        await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_CB).check({timeout: AWAIT_TIMEOUT});
+      }
+
+      // Fill a test email. The input may be hidden until the checkbox is checked;
+      // force:true bypasses the visibility check (same pattern as vmNotAnsweredRings).
+      const testEmail = 'autotest-notification@example.com';
+      await page.locator(CALLING_SELECTORS.VM_NOTIF_EMAIL_ID).fill(testEmail, {force: true});
+      await saveVoicemailSettings(page);
+      await loadSettings(page);
```
Shreyas281299: This test sequence toggles/saves/asserts/restores in a straight line without `try/finally`, so any assertion failure can skip restore and leak mutated account state into downstream tests. Can we adopt the same `try/finally` cleanup pattern used in the CS-CALL tests to guarantee restoration?

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+      }
+
+      // Restore
+      if (original) {
+        await page.locator(CALLING_SELECTORS.VM_ENABLED_CB).check({timeout: AWAIT_TIMEOUT});
+      } else {
+        await page.locator(CALLING_SELECTORS.VM_ENABLED_CB).uncheck({timeout: AWAIT_TIMEOUT});
+      }
+      await saveVoicemailSettings(page);
+    });
```
Shreyas281299, line 532: In CS-VM-009 step 3, `vmNotAnsweredRings` is overwritten to `'4'` but only partially restored depending on original toggle state. If `originalUnanswered` is true, the rings value may persist after test completion. Could we capture and restore the original rings value unconditionally to avoid polluting shared test accounts?

packages/calling/playwright/test-groups/call-settings.ts
Shreyas281299, line 976: CS-CALL-102 appears to place the call immediately after DND state enforcement, unlike sibling CS-CALL tests that include a short propagation wait. Since backend routing can lag briefly, adding the same stabilization delay here would likely reduce flakiness.

packages/calling/playwright/utils/oauth.setup.ts
```diff
@@ -75,7 +75,9 @@ const fetchAccessToken = async (
     await page.getByRole('textbox', {name: 'name@example.com'}).press('Enter');
 
     // 4. Enter password and click Sign In
-    await page.getByPlaceholder('Password').fill(password, {timeout: 15000});
+    await page
+      .locator('input#IDToken2[name="IDToken2"][type="password"]')
```
Shreyas281299, line 79: Switching to a strict `input#IDToken2[name="IDToken2"][type="password"]` selector improves specificity but removes the previous placeholder-based fallback. Since IdP attribute churn is common, can we keep this selector as primary and fall back to `getByPlaceholder('Password')` if it doesn't resolve?

packages/calling/playwright/constants/selectors.ts
```diff
@@ -52,4 +52,47 @@ export const CALLING_SELECTORS = {
...
+  CF_SAVE_BTN: '#callForwardForm #cfButton',
+  CF_ALWAYS_CB: '#callForwardForm #alwaysCb',
+  CF_ALWAYS_DEST: '#alwaysDest',
+  CF_BUSY_CB: '#callForwardForm #busyCb',
+  CF_BUSY_DEST: '#busyDest',
+  CF_NO_ANSWER_CB: '#notAnsweredCb',
+  CF_NO_ANSWER_DEST: '#notAnsweredDest',
+  CF_NOT_REACHABLE_CB: '#notReachableCb',
+  CF_NOT_REACHABLE_DEST: '#notReachableDest',
+  CF_DIRECTORY_NUMBER: '#directoryNumber',
```
Shreyas281299, line 74: `div#incomingsection #answer` relies on the container tag name to disambiguate duplicate `incomingsection` IDs. If markup changes (e.g., fieldset -> div), this may become ambiguous and fail in strict mode. Would it be safer to fix duplicate IDs in sample HTML and keep selectors semantically unique?

packages/calling/playwright/test-groups/call-settings.ts
```diff
@@ -0,0 +1,839 @@
...
+    // -----------------------------------------------------------------------
+    // Call Forward No Answer — call arrives, then gets forwarded after timeout
+    // -----------------------------------------------------------------------
+
+    test('CS-CALL-104: CF No Answer — call is forwarded when callee does not answer', async () => {
+      await loadSettings(calleePage);
+      await setCallForwardNoAnswer(calleePage, true, cfDestination);
+      // Allow the setting to propagate before placing the call.
+      await calleePage.waitForTimeout(5000);
+
```
Shreyas281299, line 884: `secondCallerPage` appears to be fully initialized but unused in `callSettingsCallTests`, while still adding setup cost and an extra registered account session. Is this context still required? If not, removing it may both reduce runtime and eliminate one source of account collision.

packages/calling/playwright.config.ts
```diff
@@ -139,5 +139,36 @@ export default defineConfig({
...
+    },
+    {
+      name: 'SET_CALL_SETTINGS - INT',
+      dependencies: ['OAuth - INT'],
+      testDir: './playwright/suites',
+      testMatch: USER_SETS.SET_CALL_SETTINGS.testSuite,
+      use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
+    },
+
+    // 3-user transfer tests — waits for 2-user (shared USER_4+USER_5)
```
Shreyas281299: Noting the commented `SET_3USER` block references identifiers/projects that don't exist (`USER_SETS.SET_3USER`, `SET_2USER - PROD/INT`). Could we either remove this stale block or update it to valid identifiers to avoid confusion for future maintainers?

packages/calling/playwright/utils/call.ts
```diff
@@ -31,7 +31,9 @@ export const makeCall = async (page: Page, destination: string): Promise<void> =
 };
 
 export const waitForIncomingCall = async (page: Page): Promise<void> => {
-  await expect(page.locator(CALLING_SELECTORS.INCOMING_ANSWER_BTN)).toBeEnabled({timeout: 30000});
+  await expect(page.locator(CALLING_SELECTORS.INCOMING_CALL_ANSWER_BTN)).toBeEnabled({
+    timeout: 30000,
+  });
 };
 
 export const answerCall = async (page: Page): Promise<void> => {
```
Shreyas281299, line 60: `waitForIncomingCall` and `answerCall` now rely on different incoming-answer selectors (`INCOMING_CALL_ANSWER_BTN` vs `INCOMING_ANSWER_BTN`). That split can cause partial breakages if one locator changes independently. Can we standardize both flows on a single selector constant/invariant?

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: added waitForCallerOutboundCall() on CS-CALL-101/102/104/105/106; implemented CS-CALL-103 (answer first call, second call while busy, call-waiting-aware); hardened CS-CALL-104/106 with setting reload/verify. Full SET_CALL_SETTINGS - PROD suite passes locally.

packages/calling/playwright.config.ts
vivekv1504: Done!  Removed commented lines

packages/calling/playwright.config.ts
vivekv1504: SET_CALL_SETTINGS currently shares USER_1/2/3 with SET_REGISTRATION_* and SET_CALL_TRANSFER_CONSULT while only depending on OAuth, which breaks the single-active-session rule in test-data.ts. I’ll fix this in a follow-up by [adding project dependencies | assigning a dedicated user set] so those sessions never overlap.

packages/calling/playwright.config.ts
vivekv1504: added dependencies on SET_REGISTRATION_1/2/3 and SET_CALL_TRANSFER_CONSULT for both PROD and INT so SET_CALL_SETTINGS won’t run concurrently with other suites using USER_1/2/3.

packages/calling/playwright.config.ts
vivekv1504: Already removed the stale commented SET_3USER / SET_2USER block in the latest push — it was leftover from an older naming scheme. The live config now only uses SET_CALL and SET_CALL_TRANSFER_CONSULT

packages/calling/playwright/utils/oauth.setup.ts
vivekv1504: restored the fallback. Password entry now uses input#IDToken2[name="IDToken2"][type="password"] as the primary locator and falls back to getByPlaceholder('Password') via .or() if the strict selector doesn’t resolve. This keeps specificity when the IdP markup is stable while still tolerating attribute churn.

packages/calling/playwright/utils/call.ts
vivekv1504: Standardized on a single INCOMING_ANSWER_BTN constant. Both waitForIncomingCall and answerCall use it, and call-settings.ts was updated. Removed the duplicate incoming-call markup from the sample HTML so the selector is #incomingsection #answer without tag-based disambiguation.

packages/calling/playwright/constants/selectors.ts
vivekv1504: Addressed by removing duplicate IDs from the sample HTML and standardizing on INCOMING_ANSWER_BTN: '#incomingsection #answer'. No tag-based disambiguation needed anymore.

packages/calling/playwright/utils/call-settings.ts
vivekv1504: Updated all waitForResponse predicates in call-settings.ts to require r.ok() in addition to URL and method, so 4xx/5xx no longer satisfy the waiter.

packages/calling/playwright/utils/call-settings.ts
vivekv1504: Fixed — added enable && isChecked && destination handling to all CF helpers so the destination is always refreshed before save when enabling with an explicit number.

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Thanks — agreed that not.toBeEmpty() was too weak since both success (callSetting) and error paths populate #callforwardalways-data.

Updated CS-CFA-007 to assert on the network response instead: we wait for a successful GET callForwarding, then verify the response body includes callForwarding.always with a boolean enabled field. This ensures the lookup used valid API data rather than only checking that the element had any text (which could be an error string).

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Wrapped the DND route stub in try/finally and pass the same routeHandler to unroute so the stub is always removed, even when assertions fail.

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Captured originalRings before Step 3 and restore both the checkbox and rings value unconditionally after the assertion, so shared test accounts are not left with numberOfRings = 4.

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Wrapped mutate/assert steps in try/finally so checkbox and email are always restored on the server, even when assertions fail — same cleanup pattern as the CS-CALL tests.

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: secondCallerPage (USER_1) is intentionally provisioned and is used — it's the second concurrent caller in CS-CALL-103 (CF When Busy). CF Busy / call-waiting can only be exercised when USER_3 is already on an established call with USER_2 (callerPage) and a second inbound call arrives from a different registered line:
```

// Second call while callee is busy.
await makeCall(secondCallerPage, calleeNumber);
await waitForCallerOutboundCall(secondCallerPage);
```
It's also cleaned up in afterEach and in CS-CALL-103's finally. Removing it would break CS-CALL-103, so I've kept it.

Agreed that the other live-call tests (CS-CALL-101/102/104/105/106) don't dial from it — if reducing the always-on 3-account setup becomes a concern, I'm happy to follow up by lazy-initializing it or splitting CS-CALL-103 into its own 3-user describe block.

packages/calling/playwright/test-groups/call-settings.ts
vivekv1504: Good catch on the potential for collision. In practice these suites don't run in parallel — SET_CALL_SETTINGS already runs after SET_CALL because of the transitive dependency chain (SET_CALL_SETTINGS → SET_CALL_TRANSFER_CONSULT → SET_CALL), so USER_4 is deregistered by the time the CF tests forward to it. That said, relying on a transitive ordering is fragile. I've made the dependency explicit by adding SET_CALL directly to the SET_CALL_SETTINGS project dependencies (both PROD and INT) and documented the reasoning in both the config and the test, so the guarantee survives future refactors.

eigengravy (approved):

Shreyas281299 (approved):

PR #5018, opened 2026-06-03
fix(plugin-meetings): handling of changes to "view participant list" meeting controls

marcin-bazyl:
# COMPLETES #[SPARK-819661](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-819661)

See the jira description.

SDK needs to send current values for the remaining props from viewTheParticipantList control (the ones that client sets to undefined). 
First I've done it by caching the current values in ControlsOptionsManager, this is more inline with other existing code in SDK where we listen for locus info changes in meeting object and update any other objects with the changes - that's the 1st commit. 
Then I decided that I don't really like having the same property values copies in multiple places and changed it to use a callback that gets the values directly from locusInfo.controls - that's the 2nd commit.
Finally, while testing I realised that we should not be sending the 3 props for normal meetings, so now ControlsOptionsManager also needs to know if it's a webinar or not and that also comes from locus info, so I ended up with 1 more callback - see 3rd commit.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests, manual testing with web app

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/controls-options-manager/index.ts
+++ b/packages/@webex/plugin-meetings/src/controls-options-manager/index.ts
@@ -56,11 +56,7 @@ export default class ControlsOptionsManager {
    */
   private mainLocusUrl: string;
 
-  private currentControls: {
-    viewTheParticipantList: {enabled: boolean; panelistEnabled: boolean; attendeeCount: boolean};
-  } = {
-    viewTheParticipantList: {enabled: false, panelistEnabled: false, attendeeCount: false},
-  };
+  private getControls: () => Record<string, any> = () => ({});
 
   /**
    * @param {MeetingRequest} request
@@ -73,6 +69,7 @@ export default class ControlsOptionsManager {
     options?: {
       locusUrl: string;
       displayHints?: Array<string>;
+      getControls: () => Record<string, any>;
     }
   ) {
     this.initialize(request);
@@ -95,7 +92,11 @@ export default class ControlsOptionsManager {
    * @public
    * @memberof ControlsOptionsManager
    */
-  public set(options?: {locusUrl: string; displayHints?: Array<string>}) {
+  public set(options?: {
+    locusUrl: string;
+    displayHints?: Array<string>;
+    getControls?: () => Record<string, any>;
+  }) {
     this.extract(options);
   }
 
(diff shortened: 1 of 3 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -1567,6 +1567,8 @@ export default class Meeting extends StatelessWebexPlugin {
     this.controlsOptionsManager = new ControlsOptionsManager(this.meetingRequest, {
       locusUrl: this.locusInfo?.url,
       displayHints: [],
+      getControls: () => this.locusInfo?.controls,
+      isWebinar: () => this.locusInfo?.info?.isWebinar,
```
marcin-bazyl, line 1571: I had my doubts if it's worth doing the callback for isWebinar too - it feels a bit like an overkill, I mean a meeting is always either a webinar or not, it won't change dynamically, but here at the point of constructing ControlsOptionsManager we might still not have locus info, so we still need to support it changing at least once, so decided to keep it.

packages/@webex/plugin-meetings/src/controls-options-manager/index.ts
```diff
@@ -89,7 +95,12 @@ export default class ControlsOptionsManager {
    * @public
    * @memberof ControlsOptionsManager
    */
-  public set(options?: {locusUrl: string; displayHints?: Array<string>}) {
+  public set(options?: {
+    locusUrl: string;
+    displayHints?: Array<string>;
+    getControls?: () => Record<string, any>;
+    isWebinar?: () => boolean;
```
chrisadubois, line 102: I was going to say. I really dont like this. but i guess locus returns it to us...

packages/@webex/plugin-meetings/src/meeting/index.ts
chrisadubois: yeah, that's what is smelly about values like this. `isWebinar` smells to begin with.

chrisadubois (approved):

packages/@webex/plugin-meetings/src/controls-options-manager/index.ts
marcin-bazyl: yes, the flag `isWebinar` comes straight from Locus and Locus docs say that all 3 params for the /controls API to change the "view participant list" are required for webinars and I assumed we shouldn't be sending them for non-webinars as they don't make sense in non-webinar context

packages/@webex/plugin-meetings/src/controls-options-manager/index.ts
```diff
@@ -172,8 +194,22 @@ export default class ControlsOptionsManager {
         );
       }
 
+      let {properties} = control;
+
+      if (control.scope === Control.viewTheParticipantList) {
+        const props = properties as ViewTheParticipantListProperties;
+        const current = this.getControls()?.viewTheParticipantList;
+        properties = {
+          enabled: props.enabled ?? current?.enabled ?? false,
```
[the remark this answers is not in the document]
marcin-bazyl: I think this is fine, in theory we could say the same about all 3 of these props, but that's the point of this PR to use the current values and if we don't have the current values than there is nothing we can do - if we send undefined, Locus will fail. Anyway, the web app won't be able to show any UI for changing these settings unless it has received Locus info already, so this edge case is not really possible in real life with our web app.

jor-row (approved):

PR #5023, opened 2026-06-05
fix(breakout): keep history csis for meeting members

WeijuanShao:

Cache history csis information for case that: host end breakouts and host return to main earlier than other panelists, there may exist a time gap between homer video layout users and locus users.
Csis are meeting-level identifiers. When a user switches between a BO session and the main session, their CSIs are not regenerated. So, to bridge this time gap, we should still be able to use the existing CSIs to match the member in ```findMemberByCsi```, allowing media users to resolve the correct member even while the member data in the main session is temporarily incomplete.

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -598,6 +607,21 @@ export default class Members extends StatelessWebexPlugin {
...
+          // Skip re-extracting when devices haven't changed - existing history already covers them.
+          const devicesUnchanged = isEqual(
+            existingMember.participant?.devices,
+            member.participant?.devices
+          );
+          if (!devicesUnchanged) {
+            const history = new Set<number>();
+            MemberUtil.extractCsis(existingMember.participant).forEach((csi) => history.add(csi));
+            if (history.size > 0) {
+              this.historyCsisByMemberId.set(member.id, history);
```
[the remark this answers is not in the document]
WeijuanShao: CSIs is meeting level, should not changed during the meeting. If csis changed, means the user already leave meeting and join again, don't need remember the old csis

packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -569,6 +578,17 @@ export default class Members extends StatelessWebexPlugin {
...
     removedMembers.forEach((memberId) => {
+      // capture CSIs before removal so findMemberByCsi can still resolve them
+      // if the member is later re-added (e.g. when leaving a breakout session)
+      // with a participant payload that no longer contains the original CSIs
+      const existingMember = this.membersCollection.get(memberId);
+      if (existingMember) {
+        const history = new Set<number>();
+        MemberUtil.extractCsis(existingMember.participant).forEach((csi) => history.add(csi));
+        if (history.size > 0) {
+          this.historyCsisByMemberId.set(memberId, history);
```
marcin-bazyl: we're overwriting the history entry here, instead we should be adding new CSIs to it

packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -598,6 +618,21 @@ export default class Members extends StatelessWebexPlugin {
...
+          // Skip re-extracting when devices haven't changed - existing history already covers them.
+          const devicesUnchanged = isEqual(
+            existingMember.participant?.devices,
+            member.participant?.devices
+          );
+          if (!devicesUnchanged) {
+            const history = new Set<number>();
+            MemberUtil.extractCsis(existingMember.participant).forEach((csi) => history.add(csi));
+            if (history.size > 0) {
+              this.historyCsisByMemberId.set(member.id, history);
```
marcin-bazyl: same as prev comment - we don't want to overwrite the entry in historyCsisByMemberId, but add to it

packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -99,6 +100,14 @@ export default class Members extends StatelessWebexPlugin {
   selfId: any;
   type: any;
 
+  /**
+   * Map of memberId -> CSIs that the member used in previous Locus updates.
+   * Kept here (rather than on each Member) so it survives members being removed
+   * and re-added (e.g. when entering/leaving a breakout session).
+   * @private
+   */
+  private historyCsisByMemberId: Map<string, Set<number>> = new Map();
```
marcin-bazyl: suggestion: we're only using this to find a member for a specific csi, so maybe instead of having a map of sets we could just have a map where the keys are CSI and values are memberId? It would also simplify the code:

for storing it, we would just call:
`historyCsisByMemberId.set(csi, member.id)`
and that's it no other checks or code needed

and for reading it:
`memberId=historyCsisByMemberId.get(csi)`
and that's it, no for loop or anything else

what do you think?

packages/@webex/plugin-meetings/src/members/index.ts
```diff
@@ -1161,18 +1196,38 @@ export default class Members extends StatelessWebexPlugin {
...
       )
     );
+
+    if (currentMatch) {
+      return currentMatch;
+    }
+
+    for (const [memberId, history] of this.historyCsisByMemberId) {
+      if (history.has(csi)) {
+        const member = this.membersCollection.get(memberId);
```
marcin-bazyl: question: when you see the issue that this PR is trying to fix, are the participants completely disappearing or they are still there, but with state LEFT? I thought they were completely disappearing and in that case just having the `historyCsisByMemberId` cache won't be enough because this lookup here in `membersCollection` will still fail, right?

packages/@webex/plugin-meetings/src/members/index.ts
WeijuanShao: Wouldn't this introduce redundant data? If a signed-in user joins the meeting again, the CSI generated during the first join would already be invalid. I'm not sure whether an expired CSI could later be reused by another participant.

packages/@webex/plugin-meetings/src/members/index.ts
WeijuanShao: I considered that approach as well. Personally, I feel that using memberId as the key provides a more member-centric view of the data, which makes the overall data model clearer.

That said, I agree that using the data can be more convenient when csi is the key. If you think using csi as the key is the better approach, I'm happy to switch to that instead.

packages/@webex/plugin-meetings/src/members/index.ts
WeijuanShao: For this specific fix, based on testing on the largeco test site, after ending a BO, Mercury sends  all data,  including participants in the LEFT state. So the LEFT participants will exist in the membersCollection, but their CSI information is empty, which is why we were previously unable to match them.

With the historyCSI approach, we keep track of CSIs that a participant has used before. This allows us to match the participant using a previously known CSI. As long as that participant has been in the main session at some point, there is a good chance that we can find a match.

That said, the historyCSI solution is only a workaround. It improves the likelihood of finding the correct member, but it does not guarantee that a match can always be found.

packages/@webex/plugin-meetings/src/members/index.ts
marcin-bazyl: yes, actually, you're right, if CSIs change we shouldn't really need the old ones for anything.

packages/@webex/plugin-meetings/src/members/index.ts
marcin-bazyl: The main downside of current approach is that we have to loop over all the entries in `findMemberByCsi()` - when we have a 1k meeting or in the future 5k meetings in theory we may have to be doing a loop over thousands of members and `findMemberByCsi()` is called whenever the person shown in one of video panes changes, so that can be quite often in a large meeting where more than 6 people are speaking (although maybe in practice not all members will have entries in historyCsisByMemberId?), so that's the main reason why I think using csi as key would be better.

marcin-bazyl (approved):

xinhyao (approved):

JudyZhuHz (approved):

PR #5036, opened 2026-06-16
feat: llmrefactor and mercury revert

shnaaz:

This PR refactors the LLM, adds LLM and Mercury plugins to support multiple simultaneous LLM sessions (default + practice session) and wraps Mercury in a plugin pattern for consistency.

< DESCRIBE YOUR CHANGES >
Created LLMPlugin wrapper that maintains Map<sessionId, LLMChannel>

Refactored llm.ts: 
- Simplified to single-session LLMChannel class
- Moved multi-session management to plugin layer
- Each channel has its own connectingPromise

Added mercury-plugin.ts:
- Wraps existing Mercury class in plugin pattern
- Exposes: connected, socket, localClusterServiceUrls, hasEverConnected
- Methods: connect(), disconnect(), logout(), processRegistrationStatusEvent()
- Event re-emission from internal _mercury to plugin

Reverted: mercury.js
- Removed session-id changes that were causing issues
- Back to single Mercury connection model

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
+++ b/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
@@ -28,6 +28,12 @@ export class LLMPlugin extends (WebexPlugin as any) {
     if (!channel) {
       // @ts-ignore — WebexPlugin children require {parent: this.webex}
       channel = new LLMChannel({parent: this.webex});
+      // Forward all events emitted by the channel up through the plugin so that
+      // callers doing llm.on('event:relay.event', ...) or llm.on('online', ...)
+      // receive events from whichever session channel emits them.
+      channel.on('all', (eventName: string, ...args: any[]) => {
+        this.trigger(eventName, ...args);
+      });
       this.sessions.set(sessionId, channel);
     }
 
@@ -53,6 +59,7 @@ export class LLMPlugin extends (WebexPlugin as any) {
     datachannelToken?: string,
     sessionId: string = LLM_DEFAULT_SESSION
   ): Promise<void> {
+    // Deduplicate concurrent calls for the same session while a connection is in-flight.
     const inProgress = this.connectingPromises.get(sessionId);
 
     if (inProgress) {
@@ -60,6 +67,17 @@ export class LLMPlugin extends (WebexPlugin as any) {
     }
 
     const channel = this.getOrCreateSession(sessionId);
+
+    // If the channel is already connected to the exact same datachannel URL,
+    // there is nothing to do — avoid triggering a reconnect that would cause
+    // the server to replace the existing connection with 4000 Replaced.
+    if (
+      channel.isConnected() &&
+      channel.getDatachannelUrl() === datachannelUrl &&
+      channel.getLocusUrl() === locusUrl
+    ) {
+      return Promise.resolve();
+    }
 
     const promise = channel
       .registerAndConnect(locusUrl, datachannelUrl, datachannelToken)
--- a/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
+++ b/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
@@ -1,0 +1,271 @@
+/* eslint-disable require-jsdoc */
+import {WebexPlugin} from '@webex/webex-core';
+import LLMChannel, {config} from './llm';
+import {DATA_CHANNEL_WITH_JWT_TOKEN, LLM_DEFAULT_SESSION} from './constants';
+import {DataChannelTokenType} from './llm.types';
+
+/**
+ * LLMPlugin — registered as `webex.internal.llm`.
+ *
+ * Maintains a Map<sessionId, LLMChannel> so multiple simultaneous LLM
+ * connections (e.g. default session + practice session) can coexist.
+ * All existing callers continue to work unchanged via the session-keyed API.
+ *
+ * sessionId values are the LLM_DEFAULT_SESSION / LLM_PRACTICE_SESSION constants,
+ * which match the DataChannelTokenType enum values so token keys and session keys
+ * are the same namespace.
+ */
+export class LLMPlugin extends (WebexPlugin as any) {
+  namespace = 'llm';
+
+  private sessions = new Map<string, LLMChannel>();
+
+  private getOrCreateSession(sessionId: string): LLMChannel {
+    let channel = this.sessions.get(sessionId);
+
+    if (!channel) {
+      // @ts-ignore — WebexPlugin children require {parent: this.webex}
+      channel = new LLMChannel({parent: this.webex});
+      this.sessions.set(sessionId, channel);
+    }
+
+    return channel;
+  }
+
+  private getSession(sessionId: string): LLMChannel | undefined {
+    return this.sessions.get(sessionId);
+  }
+
+  // Before webex.internal.llm was LLChannel instance which extended Mercury so it was directly available
+  get hasEverConnected(): boolean {
+    for (const ch of this.sessions.values()) {
+      if (ch.hasEverConnected) return true;
+    }
+
+    return false;
+  }
+
+  public registerAndConnect(
+    locusUrl: string,
+    datachannelUrl: string,
+    datachannelToken?: string,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): Promise<void> {
+    const channel = this.getOrCreateSession(sessionId);
+
+    return channel.registerAndConnect(locusUrl, datachannelUrl, datachannelToken);
+  }
+
+  public disconnectLLM(
+    options: {code: number; reason: string},
+    ownerMeetingId?: string,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): Promise<boolean | void> {
+    const channel = this.getSession(sessionId);
+
+    if (!channel) return Promise.resolve();
+
+    const {isOwner} = this.resolveSessionOwnership(ownerMeetingId, sessionId);
+
+    if (!isOwner) {
+      this.logger.info(`llm#disconnectLLM --> skipping, not owner of session ${sessionId}`);
+
+      return Promise.resolve(false);
+    }
+
+    return channel.disconnectLLM(options).then(() => {
+      this.sessions.delete(sessionId);
+
+      return true;
+    });
+  }
+
+  public disconnectAllLLM(options?: {code: number; reason: string}): Promise<void> {
+    const promises = Array.from(this.sessions.entries()).map(([sessionId, channel]) =>
+      channel.disconnectLLM(options).then(() => this.sessions.delete(sessionId))
+    );
+
+    return Promise.all(promises).then(() => undefined);
+  }
+
+  public isConnected(sessionId: string = LLM_DEFAULT_SESSION): boolean {
+    return this.getSession(sessionId)?.isConnected() ?? false;
+  }
+
+  public getBinding(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
+    return this.getSession(sessionId)?.getBinding();
+  }
+
+  public getLocusUrl(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
+    return this.getSession(sessionId)?.getLocusUrl();
+  }
+
+  public getDatachannelUrl(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
+    return this.getSession(sessionId)?.getDatachannelUrl();
+  }
+
+  // tokenKey IS the sessionId (DataChannelTokenType enum values equal LLM_*_SESSION constants)
+
+  public getDatachannelToken(
+    // eslint-disable-next-line default-param-last
+    tokenKey: string = LLM_DEFAULT_SESSION,
+    ownerMeetingId?: string
+  ): string | undefined {
+    const channel = this.getSession(tokenKey);
+
+    if (!channel) return undefined;
+
+    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
+
+    if (!isOwner) {
+      this.logger.info(
+        `llm#getDatachannelToken --> skip read for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
+      );
+
+      return undefined;
+    }
+
+    return channel.getDatachannelToken();
+  }
+
+  public setDatachannelToken(
+    datachannelToken: string,
+    ownerMeetingId?: string,
+    tokenKey: string = LLM_DEFAULT_SESSION
+  ): void {
+    const channel = this.getOrCreateSession(tokenKey);
+    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
+
+    if (!isOwner) {
+      this.logger.info(
+        `llm#setDatachannelToken --> skip write for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
+      );
+
+      return;
+    }
+
+    channel.setDatachannelToken(datachannelToken);
+  }
+
+  public clearDatachannelToken(tokenKey: string, ownerMeetingId: string): void {
+    const channel = this.getSession(tokenKey);
+
+    if (!channel) return;
+
+    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
+
+    if (!isOwner) {
+      this.logger.info(
+        `llm#clearDatachannelToken --> skip clear for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
+      );
+
+      return;
+    }
+
+    channel.clearDatachannelToken();
+  }
+
+  public setRefreshHandler(
+    handler: () => Promise<{
+      body: {datachannelToken: string; datachannelTokenType: DataChannelTokenType};
+    }>,
+    ownerMeetingId?: string,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): void {
+    const channel = this.getOrCreateSession(sessionId);
+    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, sessionId);
+
+    if (!isOwner) {
+      this.logger.info(
+        `llm#setRefreshHandler --> skip write for session ${sessionId}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
+      );
+
+      return;
+    }
+
+    channel.setRefreshHandler(handler);
+  }
+
+  public refreshDataChannelToken(sessionId: string = LLM_DEFAULT_SESSION): Promise<any> {
+    const channel = this.getSession(sessionId);
+
+    if (!channel) {
+      this.logger.warn(`llm#refreshDataChannelToken --> no channel for session ${sessionId}`);
+
+      return Promise.resolve(null);
+    }
+
+    return channel.refreshDataChannelToken();
+  }
+
+  public setOwnerMeetingId(
+    ownerMeetingId: string | undefined,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): void {
+    const channel = this.getSession(sessionId);
+
+    if (channel) channel.ownerMeetingId = ownerMeetingId;
+  }
+
+  public getOwnerMeetingId(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
+    return this.getSession(sessionId)?.ownerMeetingId;
+  }
+
+  public resolveSessionOwnership(
+    ownerMeetingId?: string,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): {currentOwner: string | undefined; isOwner: boolean} {
+    const currentOwner = this.getOwnerMeetingId(sessionId);
+    const isOwner = !currentOwner || !ownerMeetingId || currentOwner === ownerMeetingId;
+
+    return {currentOwner, isOwner};
+  }
+
+  public getConnectionByDatachannelUrl(url: string): LLMChannel | undefined {
+    for (const channel of this.sessions.values()) {
+      const datachannelUrl = channel.getDatachannelUrl();
+
+      if (datachannelUrl && LLMChannel.matchesDatachannelRequestUrl(url, datachannelUrl)) {
+        return channel;
+      }
+    }
+
+    return undefined;
+  }
+
+  public getLocusUrlByDatachannelUrl(requestUrl: string): string | undefined {
+    for (const channel of this.sessions.values()) {
+      const datachannelUrl = channel.getDatachannelUrl();
+
+      if (datachannelUrl && LLMChannel.matchesDatachannelRequestUrl(requestUrl, datachannelUrl)) {
+        return channel.getLocusUrl();
+      }
+    }
+
+    return undefined;
+  }
+
+  public getSessionIdByDatachannelUrl(requestUrl: string): string | undefined {
+    for (const [sessionId, channel] of this.sessions.entries()) {
+      const datachannelUrl = channel.getDatachannelUrl();
+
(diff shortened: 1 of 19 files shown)


packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -44,50 +41,40 @@ export const config = {
 };
 
 /**
- * LLMChannel to provide socket connections
+ * LLMChannel — a single WebSocket connection to the LLM data channel.
+ * Session-level routing (multiple connections keyed by sessionId) is handled
+ * by LLMPlugin, which owns a Map<sessionId, LLMChannel>.
  */
-export default class LLMChannel extends (Mercury as any) implements ILLMChannel {
+export default class LLMChannel extends (Mercury as any) {
```
marcin-bazyl: why is `implements ILLMChannel` removed?

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -0,0 +1,316 @@
...
+ * sessionId values are the LLM_DEFAULT_SESSION / LLM_PRACTICE_SESSION constants,
+ * which match the DataChannelTokenType enum values so token keys and session keys
+ * are the same namespace.
+ */
+export class LLMPlugin extends (WebexPlugin as any) {
+  namespace = 'llm';
+
+  private sessions = new Map<string, LLMChannel>();
+
+  private connectingPromises = new Map<string, Promise<void>>();
```
marcin-bazyl: question: can we have the connectingPromise inside each LLMChannel instead of the Map here?

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -117,309 +101,107 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
-    }
-
-    return this.register(datachannelUrl, datachannelToken, sessionId).then(async () => {
-      if (!locusUrl || !datachannelUrl) return undefined;
-
-      // locusUrl and datachannelUrl were pre-populated before register(); here
-      // we only need to read the existing session data to get webSocketUrl/binding
-      // that register() filled in.
-      const sessionData = this.connections.get(sessionId) || {};
+    // Store directly on the instance — no session map needed
```
marcin-bazyl: nitpick: this comment is not really needed here

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -429,122 +211,21 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
...
-   * @param {{code: number, reason: string}} options - The disconnect option object with code and reason
+   * Disconnects the WebSocket and clears all connection state.
+   * @param {object} [options]
    * @returns {Promise<void>}
    */
-  public disconnectAllLLM = (options?: {code: number; reason: string}): Promise<void> =>
-    this.disconnectAll(options).then(() => {
-      // Clean up all connection data
-      this.connections.clear();
+  public disconnectLLM = (options?: {code: number; reason: string}): Promise<void> => {
```
marcin-bazyl, line 237: longer term we would probably want this method to just be called `disconnect` and overrride the Mercury's `disconnect` method and still call the Mercury's one via `super.disconnect(options)` - but this doesn't need to be done in this PR, can be in one of the next ones

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
```diff
@@ -315,7 +315,11 @@ export default class Socket extends EventEmitter {
    */
   onmessage(event) {
     try {
-      const data = JSON.parse(event.data);
+      const rawData =
+        event.data instanceof ArrayBuffer
+          ? new TextDecoder('utf-8').decode(event.data)
```
marcin-bazyl: why is this now needed and wasn't previously?

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
```diff
@@ -419,19 +423,27 @@ export default class Socket extends EventEmitter {
...
-          this.removeListener('message', waitForBufferState);
-          this._ping();
-          resolve();
-        }
-      };
+      // Some channels (e.g. LLM data-channel) never send mercury.buffer_state.
+      // For those, the socket signals skipBufferState: true via open() options,
+      // so we start ping/pong and resolve immediately after sending auth.
+      if (this.skipBufferState) {
+        this._ping();
```
marcin-bazyl: just out of curiosity: how was this working before? where was this logic implemented?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -208,180 +178,62 @@ const Mercury = WebexPlugin.extend({
...
-      });
-
-    this._connectPromises.set(sessionId, connectPromise);
-
-    return connectPromise;
+      return this._connectWithBackoff(webSocketUrl);
+    });
   },
 
   logout() {
```
marcin-bazyl: feels like the logout method only really belongs to the plugin, I don't think we need it in Mercury class, do you agree? Plugin's logout would just call disconnect on the only Mercury instance it has.
However, there is a bigger issue here: old code was doing a disconnect of all sockets (including LLMs) when Mercury plugins's logout was happening, now Mercury's logout will only disconnect the Mercury connection, so we probably need a similar logout method on the LLM plugin to disconnect all LLM connections, unless we decide that it's the LLM connections owner's responsibility to disconnect it on logout - then we would have to change the code in plugin-meetings to do that.

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -473,29 +325,22 @@ const Mercury = WebexPlugin.extend({
...
         return url.format(webSocketUrl);
       });
   },
 
-  _attemptConnection(socketUrl, sessionId, callback, options = {}) {
+  _attemptConnection(socketUrl, callback, options = {}) {
     const {isShutdownSwitchover = false, onSuccess = null} = options;
 
     const socket = new Socket();
-    socket.connecting = true;
```
marcin-bazyl, line 486: how come we don't need to set some "connnecting" flag now anymore?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -691,45 +537,29 @@ const Mercury = WebexPlugin.extend({
           this.logger.info(
             `${this.namespace}: ${msg}; log statement about next retry was inaccurate; ${err}`
           );
-          if (sessionSocket) {
-            sessionSocket.connecting = false;
```
marcin-bazyl, line 695: same as above: are we sure we don't need "connecting" flag at all? what was it used for before?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -782,55 +603,34 @@ const Mercury = WebexPlugin.extend({
...
-        const sessionId = first;
-        const eventName = second;
-        const suffix = sessionId === this.defaultSessionId ? '' : `:${sessionId}`;
-
-        this.trigger(`${eventName}${suffix}`, ...rest);
-      } else {
-        // Old usage: _emit(eventName, ...args)
-        this.trigger(...args);
-      }
+      this.trigger(...args);
```
marcin-bazyl, line 634: the old code was adding the event suffix - see lines 806-808, this new code is not doing that - I think this breaks the practice session code in plugin-meetings that registers specificallly for practice session events with the suffix. I think for now we need to keep the suffix and fix this in one of later PRs when we actually change plugin meetings to use the right LLM instances instead of using session ids

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: this is removed as part of revert, we removed all sessionId related stuff from mercury

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
shnaaz: Had error in console after refactor: -  'received non-string data, type=[object ArrayBuffer], domain=llm2-connection-r.wbx2.com'
The LLM server sends binary frames, but socket-base.js blindly calls JSON.parse(event.data) on everything

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
shnaaz: Think this.connect() used to hang before as well but may be we got away with it, this refactor has exposed the issue

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: yes makes sense, I will move logout mercury plugin level. for llm LLMPlugin already has disconnectAllLLM, so can do 
```
// packages/@webex/internal-plugin-llm/src/index.ts
WebexCore.registerInternalPlugin('llm', LLMPlugin, {
  config,
  onBeforeLogout() {
    return this.disconnectAllLLM();
  },
});
```

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: in multi session code each socket needed it's own connecting state. now we have mercury's own this.connecting

packages/@webex/internal-plugin-llm/src/llm.ts
shnaaz: fixed

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
shnaaz: done

packages/@webex/internal-plugin-llm/src/llm.ts
shnaaz: not in this pr but as suggested in future prs

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: Taken care of the issue

shnaaz: There was another issue where params were swapped, that has been taken care

packages/@webex/internal-plugin-llm/src/llm.ts
marcin-bazyl: ok, how are we going to track things like that? shall we just add jiras under the SPARK-811005 epic?

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
marcin-bazyl: interesting.... so previously these binary frames were ignored and now we're handling them? what do you see in those binary frames? is it even json at all or something else?

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
marcin-bazyl: I've just tried it on "next" branch and I am receiving the mercury buffer state message on the LLM connection and the promise is getting resolved:
image
so we should understand this better and we should not remove the check for LLM - it is part of authorization so it's important

packages/@webex/internal-plugin-mercury/src/mercury.js
marcin-bazyl: ok, the "connecting" property on the socket is probably redundant and I don't like it so good to remove it, but looks like you've just reverted the whole of Mickel's PR? there are some things in it that we probably need to keep, for example in Mercury.connect() he added a check if `connecting` is true or if there is a `_connectPromises` entry for that session id - in these cases connect() returns early. We need these checks now too.

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -111,63 +99,50 @@ const Mercury = WebexPlugin.extend({
...
         isShutdownSwitchover: true,
         attemptOptions: {
           isShutdownSwitchover: true,
           onSuccess: (newSocket, webSocketUrl) => {
             this.logger.info(
-              `${this.namespace}: [shutdown] switchover connected, url: ${webSocketUrl} for ${sessionId}`
+              `${this.namespace}: [shutdown] switchover connected, url: ${webSocketUrl}`
             );
 
+            const oldSocket = this.socket;
```
marcin-bazyl: hmm not sure about this.... feels like doing this earlier, before _connectWithBackoff() is called is better

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -208,180 +178,51 @@ const Mercury = WebexPlugin.extend({
...
-        this.logger.info(`${this.namespace}: aborting shutdown switchover connection ${sessionId}`);
-        shutdownSwitchoverBackoffCall.abort();
-        this._shutdownSwitchoverBackoffCalls.delete(sessionId);
-      }
-      // Clean up any pending connection promises
-      if (this._connectPromises) {
-        this._connectPromises.delete(sessionId);
+      if (this.backoffCall) {
+        this.logger.info(`${this.namespace}: aborting connection`);
+        this.backoffCall.abort();
```
marcin-bazyl, line 235: `this.backoffCall` is not declared anywhere, it would be nicer to have it declared properly

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -208,180 +178,51 @@ const Mercury = WebexPlugin.extend({
...
-    return Promise.all(disconnectPromises).then(() => {
-      this.connected = false;
-      this.sockets.clear();
-      this.backoffCalls.clear();
-      // Clear connection promises to prevent stale promises
-      if (this._connectPromises) {
-        this._connectPromises.clear();
+      if (this.socket) {
+        this.socket.removeAllListeners('message');
+        this.once('offline', resolve);
```
marcin-bazyl, line 245: this doesn't seem right - what's the point of calling resolve when offline event is fired if we are already calling resolve in the next line (line 222) immediately so the promise will always already be resolved. Calling resolve again on the same promise does nothing

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -473,53 +314,58 @@ const Mercury = WebexPlugin.extend({
         }
 
         webSocketUrl.query.clientTimestamp = Date.now();
-        delete webSocketUrl.search;
```
marcin-bazyl, line 476: why is this line deleted?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -657,32 +489,35 @@ const Mercury = WebexPlugin.extend({
...
+      // eslint gets confused about whether or not call is actually used
       // eslint-disable-next-line prefer-const
       let call;
-      const onComplete = (err, sid = sessionId) => {
+      const onComplete = (err) => {
+        // Clear state flags based on connection type
         if (isShutdownSwitchover) {
-          this._shutdownSwitchoverBackoffCalls.delete(sid);
+          this._shutdownSwitchoverInProgress = false;
+          this._shutdownSwitchoverBackoffCall = undefined;
```
marcin-bazyl, line 538: feels like having all 3 of these:
_shutdownSwitchoverInProgress 
_shutdownSwitchoverBackoffCall 
backoffCall 
is a bit redundant, we only need 2, right? EIther a `backoffCall` and a flag to indicate of it's a shutdownSwitchover or not or the other option is to just have `backoffCall ` and `_shutdownSwitchoverBackoffCall`

but this is probably better to be done in next PR to keep this one small.

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -782,55 +592,34 @@ const Mercury = WebexPlugin.extend({
...
-
-        this.trigger(`${eventName}${suffix}`, ...rest);
-      } else {
-        // Old usage: _emit(eventName, ...args)
-        this.trigger(...args);
-      }
+      this.trigger(...args);
     } catch (error) {
-      // Safely handle errors without causing additional issues during cleanup
-      try {
```
marcin-bazyl, line 815: this try catch for errors during logging is removed - are you sure we want to remove it? I think it's better to keep it

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -850,40 +639,36 @@ const Mercury = WebexPlugin.extend({
...
-      const sessionSocket = this.sockets.get(sessionId);
-      let socketUrl;
-      event.sessionId = sessionId;
 
-      const isActiveSocket = sourceSocket === sessionSocket;
-      if (sourceSocket) {
+      let socketUrl;
+      if (isActiveSocket && this.socket) {
+        // Active socket closed - get URL from current socket reference
+        socketUrl = this.socket.url;
```
marcin-bazyl: why do we need this code here? if `isActiveSocket` is true, then this.socket and sourceSocket are the same thing, so what's the point of having this code if the code form line 656 would do the same job? or am I missing something?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6018,9 +6017,10 @@ export default class Meeting extends StatelessWebexPlugin {
     }
     switch (e.data.relayType) {
       case REACTION_RELAY_TYPES.REACTION:
+        // @ts-ignore - config coming from registerPlugin
         if (
           // @ts-ignore - config coming from registerPlugin
-          (this.config.receiveReactions || options.receiveReactions) &&
+          this.config.receiveReactions &&
```
marcin-bazyl, line 6051: nice find..... so basically `options` was always undefined, right? so potentially if `this.config.receiveReactions` was false this would throw an exception, right? and this is all easily caught by typescript, but we have the ts-ignore........

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -0,0 +1,312 @@
...
+  public registerAndConnect(
+    locusUrl: string,
+    datachannelUrl: string,
+    datachannelToken?: string,
+    sessionId: string = LLM_DEFAULT_SESSION
+  ): Promise<void> {
+    const channel = this.getOrCreateSession(sessionId);
+
+    // Deduplicate concurrent calls for the same session while a connection is in-flight.
+    if (channel.connectingPromise) {
```
marcin-bazyl: this check and the checks from lines 70-76 and also other code like lines 81, 84 - all of this belongs better inside the LLMChannel class rather than here

packages/@webex/internal-plugin-llm/src/llm.ts
shnaaz: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-827598

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: added other undeclared properties too

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: happened because of revert, added it back

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-827618

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: Added back, was removed due to revert

packages/@webex/internal-plugin-mercury/src/mercury.js
shnaaz: fixed, happened due to revert

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
shnaaz: fixing connect in mercury.js has fixed the issue where connect was getting hanged, removed this now

packages/@webex/internal-plugin-mercury/src/socket/socket-base.js
shnaaz: After making fixes in mercury.js, I did not see this issue, hence removed it

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
```diff
@@ -25,199 +24,135 @@ describe('plugin-llm', () => {
...
           binding: 'binding',
           webSocketUrl: 'wss://example.com/socket',
         },
       });
-      const sockets = new Map();
-
-      llmService.connect = sinon.stub().callsFake((url, sessionId) => {
-        sockets.set(sessionId, {connected: true});
-        llmService.getSocket = sinon.stub().callsFake((sid) => sockets.get(sid));
+      llmChannel.connect = sinon.stub().callsFake(() => {
```
marcin-bazyl, line 36: this doesn't seem right - we shouldn't be mocking the functionality of LLMChannel in tests of LLMChannel

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
```diff
@@ -25,199 +24,135 @@ describe('plugin-llm', () => {
...
-              webSocketUrl: 'wss://example.com/socket',
-            },
-          };
-        });
-
-        await llmService.registerAndConnect();
-        assert.equal(llmService.isConnected(), false);
-      });
-
-      it('registers connection with token', async () => {
```
marcin-bazyl, line 85: why this test and others around it are removed?

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
```diff
@@ -227,189 +162,98 @@ describe('plugin-llm', () => {
...
-            }
-
-            return this.disconnect(options, sessionId).then(() => {
-              this.connections.delete(sessionId);
-            });
-          },
-        };
-      });
-
-      it('calls disconnect and clears session connection (including token stored in session)', async () => {
```
marcin-bazyl, line 300: again - some removed tests, aren't they needed? when we're refactoring stuff, the unit tests are our safety net for regressions, so we should always try to keep as many of them as possible

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
shnaaz: This is mocking Mercury.connect(), LLMChannel extends Mercury so connect, disconnect, request is inherited from Mercury, hence stubbing them is correct.
We are testing registerAndConnect, disconnectLLM, token handling without opening network connections

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
shnaaz: Testing register which is private via registerAndConnect.
- old registers connection - new test is 'registers and connects'
- old registers connection with token - new 'sends requests with token header when token enabled and provided'

Missed does not register for invalid input - added that now

packages/@webex/internal-plugin-llm/test/unit/spec/llm.js
shnaaz: These were when we had multisession architecture. I have now added tests for llm-plugin as session mapping is there so that testing is covered for now

marcin-bazyl (approved):

PR #5049, opened 2026-06-25
fix: resolve reachability protocols

antsukanova:
# COMPLETES [SPARK-827697](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-827697)

Instant Connect was choosing poor media nodes because it fell back to GeoIP  instead of real reachability data. TCP and TLS reachability checks were gated behind `experimental` flags that defaulted to **off**, so clients on UDP-restricted networks had no TCP/TLS reachability results to report, leaving the backend to guess via GeoIP and sometimes pick suboptimal nodes, hurting media quality.

- **TCP and TLS reachability checks now run by default** (alongside UDP), so node
  selection is based on real reachability instead of GeoIP fallback.
- **Unified all reachability configuration** under the existing
  `meetings.enableReachabilityChecks` option:
  - `true` (default) → UDP + TCP + TLS
  - `false` → all checks disabled
  - `{ tcp, tls }` → toggle the TCP/TLS fallback transports (UDP is always tested,
    since it's the primary media transport)
- **Removed the now-redundant pieces**: the `experimental.enableTcpReachability` /
  `experimental.enableTlsReachability` flags and two unused private toggle methods.
- Updated the browser meetings sample to use the new config shape.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- `@webex/plugin-meetings` unit tests for reachability and meetings (new cases for
  default-on behavior, the unified config resolver, and config-driven protocol
  gating).
- ESLint and the TypeScript build (`build:src`) on the affected package.
- tested with sample app: 
Screenshot 2026-06-26 at 15 56 50
Screenshot 2026-06-26 at 15 57 21
Screenshot 2026-07-02 at 11 44 58

- [ ] This PR is related to
  - [x] Defect fix

### I certified that


k-wasniowski (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -977,6 +942,25 @@ export default class Meetings extends WebexPlugin {
...
+   * API to set which media reachability protocols are checked. Accepts `true`/`false`
+   * to enable/disable all checks, or an object to toggle TCP/TLS individually (UDP is
+   * always tested unless reachability is disabled entirely).
+   *
+   * @param {EnableReachabilityChecksConfig} config new enableReachabilityChecks value
+   * @private
+   * @memberof Meetings
+   * @returns {undefined}
+   */
+  private _setReachabilityConfig(config: EnableReachabilityChecksConfig) {
```
marcin-bazyl: if we're removing the toggles in web app, then I don't think we even need this method, nobody will call it. web app can hardcode the sdk config and it won't need to change it dynamically

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -1,4 +1,43 @@
...
+ * @returns {ResolvedReachabilityProtocols} resolved per-protocol flags
+ */
+export function resolveReachabilityProtocols(
+  config: EnableReachabilityChecksConfig | undefined
+): ResolvedReachabilityProtocols {
+  if (config === false) {
+    return {udp: false, tcp: false, tls: false};
+  }
+
+  if (config === undefined || config === true) {
```
marcin-bazyl: personally I feel that having the logic of "undefined means enabled" buried deep in the code here is not ideal and would be better to just have a boolean in the config that's always there and has a default value true in the sdk default config, but it's not a blocker comment, just my opinion

antsukanova: Good idea, I feel will be clearer as well, changed!

docs/samples/browser-plugin-meetings/app.js
```diff
@@ -121,8 +121,10 @@ function generateWebexConfig({credentials}) {
         enableMediaNegotiatedEvent: false,
         enableUnifiedMeetings: true,
         enableAdhocMeetings: true,
-        enableTcpReachability: tcpReachabilityConfigElm.checked,
-        enableTlsReachability: tlsReachabilityConfigElm.checked,
+      },
+      enableReachabilityChecks: {
+        tcp: tcpReachabilityConfigElm.checked,
+        tls: tlsReachabilityConfigElm.checked,
       },
```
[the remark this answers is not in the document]
antsukanova: There was no rule that a feature PR must rebuild/commit a bundle. They're listed in .gitignore.

packages/@webex/plugin-meetings/src/reachability/util.ts
```diff
@@ -1,4 +1,40 @@
...
+): ResolvedReachabilityProtocols {
+  if (typeof enabled === 'object') {
+    return {
+      udp: true,
+      tcp: enabled.tcp ?? true,
+      tls: enabled.tls ?? true,
+    };
+  }
+
+  return {udp: enabled, tcp: enabled, tls: enabled};
```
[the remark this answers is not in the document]
antsukanova: updated in `packages/@webex/plugin-meetings/src/reachability/index.ts`

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: As agreed no longer in PR, thanks for this idea

SomeBody16 (approved):

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -499,7 +500,9 @@ export default class Reachability extends EventsScope {
         reachable = Object.values(reachabilityResults).some(
           (result) =>
             !result.isVideoMesh &&
-            (result.udp?.result === 'reachable' || result.tcp?.result === 'reachable')
+            (result.udp?.result === 'reachable' ||
+              result.tcp?.result === 'reachable' ||
+              result.xtls?.result === 'reachable')
```
marcin-bazyl: video mesh doesn't support TURN TLS reahability checks, that's why we weren't checking their result here. Even TCP is I think also not supoorted.

marcin-bazyl (approved):
looks fine, but I didn't test it

packages/@webex/plugin-meetings/src/reachability/index.ts
```diff
@@ -905,18 +915,14 @@ export default class Reachability extends EventsScope {
       const cluster = clusterList[key];
 
       // Linus doesn't support TCP reachability checks on video mesh nodes
-      const includeTcpReachability =
-        // @ts-ignore
-        this.webex.config.meetings.experimental.enableTcpReachability && !cluster.isVideoMesh;
+      const includeTcpReachability = protocols.tcp && !cluster.isVideoMesh;
```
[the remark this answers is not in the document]
antsukanova: I don't think this is a real regression.

packages/@webex/plugin-meetings/src/reachability/index.ts
antsukanova: Thnak you for pointing that. My understanding of why I added xtls here: this function only evaluates public clusters (!result.isVideoMesh), so video-mesh results are filtered out before the protocol checks. And since TCP and TLS now flow the same way by default, that's why here as well now xtls was placed.

Was the comment more of a general heads-up about the video-mesh reachability flow?

marcin-bazyl: ah, yes, sorry, misread the code I didn't notice the `!` - but anyway the intention of this method is to check if public clusters are reachable without using TURN-TLS, just directly with UDP/TCP, so better to leave the code as it is

Shreyas281299 (approved):

packages/@webex/plugin-meetings/src/reachability/index.ts
antsukanova: Got you, great point, I have removed TLS here!

packages/@webex/plugin-meetings/src/reachability/util.ts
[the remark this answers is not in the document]
antsukanova: discussed above https://github.com/webex/webex-js-sdk/pull/5049#discussion_r3506834131

PR #5043, opened 2026-06-23
feat(tooling): manual deploy — phase 1 (authorize + resolution)

bhabalan:
# COMPLETES #< INSERT LINK TO ISSUE >

## This pull request addresses

Phase 1 of the manually-triggered SDK deploy workflow. This PR establishes the **authorization and resolution** half only — it authorizes the actor, parses and validates the dispatch inputs, resolves the publishable package set and deploy version, and prints the resolved plan to the run logs. **It performs no publishing.** Phase 2 (a follow-up PR) layers the NPM publish, documentation, and tag jobs on top of this PR's `resolve` job outputs.

Splitting the work this way lets the authorization gate and the version/package resolution land and be reviewed independently before the irreversible publishing steps are introduced.

## by making the following changes

Adds a new workflow file `.github/workflows/manual-deploy.yml` plus the design doc `docs/superpowers/specs/2026-06-11-manual-deploy-package-resolution-design.md`:

- **Triggered manually** via `workflow_dispatch` with `version` (semver), `dist-tag` (latest/next/beta/alpha), and `dry-run` (boolean) inputs
- **Three sequential jobs:** authorize → initialize → resolve
- **`authorize`** — gates the run behind active membership of `@webex/cypher-sdk-release-team` (via `ORG_READ_TOKEN`); no usernames hardcoded
- **`initialize`** — validates the input version format and the dist-tag / auto-version rules, then installs dependencies
- **`resolve`** — builds the `*-tools` workspaces, resolves the publishable package set (workspaces defining `deploy:npm`), resolves the deploy version (operator-supplied or auto minor-bump of current stable), and logs the full plan to the step output and `$GITHUB_STEP_SUMMARY`
- Exposes `packages`, `packages-glob`, and `version` as `resolve` job outputs — the contract Phase 2 consumes

**Deliberately out of scope (Phase 2):** no build-for-publish, no `yarn npm publish`, no documentation/changelog force-push, no git tag. The `dry-run` input is parsed and logged here; it is enforced by the publish jobs in Phase 2.

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Tooling change
- [ ] Internal code refactor

## The following scenarios were tested

- YAML validated with the `js-yaml` parser (parses cleanly; jobs resolve to `authorize`, `initialize`, `resolve`).
[...]

PR #5079, opened 2026-07-06
feat(tooling): manual deploy — phase 2 (publish + docs + tag)

bhabalan:

Phase 2 of the manually-triggered SDK deploy workflow, building on Phase 1 (#5043, merged). Phase 1 landed the **authorization and resolution** half — it authorizes the actor, validates the dispatch inputs, and resolves the publishable package set and deploy version. This PR layers the **publishing** half on top of Phase 1's `resolve` job outputs.

Extends `.github/workflows/manual-deploy.yml`:

- **`resolve`** now also builds the publishable packages (`Build Webex Core` + `Build Other Packages --topological`) and caches the `dist` outputs so the publish jobs can restore rather than rebuild.
- **`publish-npm`** — sets the resolved version across the publishable set (`npm pkg set`), builds for publish, and publishes with `--access public --tag <dist-tag> --tolerate-republish` (idempotent recovery reruns). A dry run validates packaging via `yarn pack --dry-run` without uploading.
- **`publish-documentation`** — rebuilds docs, regenerates the changelog, and force-pushes the `documentation` branch with `--force-with-lease`. Skipped on dry runs.
- **`publish-tag`** — creates and pushes the `v$VERSION` git tag, skipping cleanly when it already exists. Skipped on dry runs.

Job flow: `authorize → initialize → resolve → publish-npm → publish-documentation → publish-tag`.

### Codex review comments addressed

Two comments from the original combined PR (#5026) that were never addressed before the phase split:

- **Serialize deploys across versions before pushing docs** — the concurrency group is now a single stable `manual-deploy` lock instead of a per-version key. The old key let two different-version runs race the shared `documentation` force-push and git tag; the loser failed only after `publish-npm` had already uploaded packages.
- **Skip the current release tag as the changelog base** — the changelog base now excludes the version being deployed (`git describe --tags --abbrev=0 --exclude="v$VERSION"`). On a recovery rerun where `v$VERSION` already exists on the checked-out commit, `git describe` would otherwise return it and `git log <base>..HEAD` would be empty, dropping the recovered commits.

The remaining codex comment (changelog version derived from the registry dist-tag rather than the requested version) is inherent to `package-tools changelog` and correct in the normal flow, where `publish-npm` moves the dist-tag before docs run. It is documented inline as a recovery-rerun caveat.

### Known limitation (out of scope)

The `manual-deploy` concurrency lock only serializes manual deploys against each other. The automated `deploy.yml` (CD) also force-pushes the `documentation` branch and has no concurrency group, so a manual deploy racing a push to `next` can still contend on the docs branch. Closing that gap requires a shared concurrency group across both workflows — a separate change.

### Change Type

- [x] Tooling change

- YAML validated with the `js-yaml` parser; jobs resolve to `authorize`, `initialize`, `resolve`, `publish-npm`, `publish-documentation`, `publish-tag` with correct `needs`/`if` wiring.
- Confirmed root scripts (`build:publish`, `build:script`, `build:local`, `build:docs`) exist and that `deploy:npm` is `yarn npm publish` across all publishable packages, so the forwarded `--access`/`--tag`/`--tolerate-republish` flags are valid.
- Verified `git describe --tags --abbrev=0 --exclude=<tag>` returns the previous reachable tag.
- Confirmed `package-tools changelog` derives the version from the registry dist-tag (matches the inline caveat).
- End-to-end execution can be validated post-merge with a dry run.

- [x] Tool used for AI assistance
  - [x] Other - Please Specify: Claude (Anthropic)
- [x] This PR is related to
  - [x] Tech Debt
  - [x] Automation

🤖 Generated with [Claude Code](https://claude.com/claude-code)


akulakum (approved):
Looks good — approving. A few minor nits below; none are blocking.

### `.github/workflows/manual-deploy.yml`

1. **Job-flow diagram (lines 9–11)** — The diagram implies `publish-tag` branches directly from `publish-npm`, but the actual `needs` chain is linear: `publish-tag` depends on both `publish-npm` *and* `publish-documentation`. Consider simplifying to:
   ```
   authorize → initialize → resolve → publish-npm → publish-documentation → publish-tag
   ```

2. **`Cache Distributables` in `resolve` (line ~309)** — The `dist` cache saved here is largely overwritten downstream (`publish-npm` runs `build:publish` + `build:script` after version-set; `publish-documentation` runs `build:local`). Not wrong (mirrors `deploy.yml`), but provides limited CI time savings. Worth a comment noting it's a partial optimization, or revisiting in a follow-up.

3. **Action version drift (lines ~119+)** — This workflow uses `actions/checkout@v4`, `setup-node@v4`, and `cache@v4`, while `deploy.yml` has moved to v6/v5. Aligning versions would reduce maintenance surface — not blocking.

4. **Changelog recovery caveat (lines ~493–500)** — The inline note about `package-tools changelog` reading the registry dist-tag is great. Small enhancement: when `inputs.version` is explicitly set (recovery scenario), consider also echoing a warning into the run summary in `Summarize resolved deploy plan` so operators see it in the Actions UI without digging into step logs.

### PR description

- Please fill in the `COMPLETES #< INSERT LINK TO ISSUE >` placeholder before merge.

### Follow-up (out of scope)

- Consider a shared concurrency group between `manual-deploy.yml` and `deploy.yml` to prevent docs-branch races when a manual deploy coincides with a push to `next` — you've already documented this limitation well.

rsarika (approved):

PR #5091, opened 2026-07-09
test(calling): add voicemail playwright coverage

Jahnavi314:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7879

Adds Playwright E2E coverage for the Webex Calling voicemail flow. The test validates that an unanswered call can create a voicemail, the callee can list and play the voicemail, and the read state is reflected in both the UI and the refetched voicemail record before cleanup.

- Added `SET_VOICEMAIL` Playwright project and user-set configuration.
- Added voicemail suite/test group for the unanswered-call voicemail journey.
- Added voicemail Playwright helpers for list fetching, UI row lookup, playback, read/unread handling, settings restore, and cleanup.

### Change Type

- [ ] Documentation update
- [x] Tooling change
- [ ] Internal code refactor

- Automated Playwright E2E:
  - User 1 records voicemail and ends the call.
  - User 2 fetches voicemail list and sees the voicemail from User 1.
  - User 2 plays the voicemail and waits until audio playback finishes.
  - Voicemail read state is reflected in the UI and the refetched voicemail record.

  - [ ] Github Copilot
  - [ ] Other - ChatGPT
- [x] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [x] Automation
  
### I certified that

### DIFF
--- a/packages/calling/playwright.config.ts
+++ b/packages/calling/playwright.config.ts
@@ -8,8 +8,8 @@ import {USER_SETS} from './playwright/test-data';
 dotenv.config({path: path.resolve(__dirname, '../../.env')});
 
 const BASE_URL = process.env.PW_BASE_URL || 'https://localhost:8000';
-// Shared with cc_playwright voicemail/call flows; required for deterministic mic input in VM-CALL-001.
-const dummyAudioPath = path.resolve(__dirname, '../../cc_playwright/wav/dummyAudio.wav');
+// Deterministic mic input for VM-CALL-001; owned by packages/calling/playwright.
+const dummyAudioPath = path.resolve(__dirname, './playwright/fixtures/dummyAudio.wav');
 
 // Browser selection via PW_BROWSER env var: 'chrome' (default), 'firefox', 'edge', 'safari'
 const PW_BROWSER = process.env.PW_BROWSER || 'chrome';
--- a/packages/calling/playwright.config.ts
+++ b/packages/calling/playwright.config.ts
@@ -7,6 +7,7 @@ import {USER_SETS} from './playwright/test-data';
 dotenv.config({path: path.resolve(__dirname, '../../.env')});
 
 const BASE_URL = process.env.PW_BASE_URL || 'https://localhost:8000';
+const dummyAudioPath = path.resolve(__dirname, '../../cc_playwright/wav/dummyAudio.wav');
 
 // Browser selection via PW_BROWSER env var: 'chrome' (default), 'firefox', 'edge', 'safari'
 const PW_BROWSER = process.env.PW_BROWSER || 'chrome';
@@ -19,6 +20,7 @@ const chromiumArgs = [
   '--allow-file-access-from-files', // Allow file:// protocol access
   '--use-fake-ui-for-media-stream', // Auto-grant camera/mic permissions without prompt
   '--use-fake-device-for-media-stream', // Use synthetic audio/video instead of real hardware
+  `--use-file-for-fake-audio-capture=${dummyAudioPath}`, // Feed deterministic audio into WebRTC calls
   '--disable-extensions', // Prevent extensions from interfering with tests
   '--disable-plugins', // Prevent plugins from interfering with tests
   '--ignore-certificate-errors', // Accept self-signed certs from local dev server
@@ -142,18 +144,33 @@ export default defineConfig({
       testMatch: USER_SETS.SET_CALL_HISTORY.testSuite,
       use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
     },
+    // Voicemail message tests use USER_1+USER_2, so run after call history.
+    {
+      name: 'SET_VOICEMAIL - PROD',
+      dependencies: ['SET_CALL_HISTORY - PROD'],
+      testDir: './playwright/suites',
+      testMatch: USER_SETS.SET_VOICEMAIL.testSuite,
+      use: browserOptions[PW_BROWSER],
+    },
+    {
+      name: 'SET_VOICEMAIL - INT',
+      dependencies: ['SET_CALL_HISTORY - INT'],
+      testDir: './playwright/suites',
+      testMatch: USER_SETS.SET_VOICEMAIL.testSuite,
+      use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
+    },
 
-    // 3-user transfer tests — waits for call history because both suites use USER_1/USER_2.
+    // 3-user transfer tests — waits for call history and voicemail because they use USER_1/USER_2.
     {
       name: 'SET_CALL_TRANSFER_CONSULT - PROD',
-      dependencies: ['SET_CALL - PROD', 'SET_CALL_HISTORY - PROD'],
+      dependencies: ['SET_CALL - PROD', 'SET_CALL_HISTORY - PROD', 'SET_VOICEMAIL - PROD'],
       testDir: './playwright/suites',
       testMatch: USER_SETS.SET_CALL_TRANSFER_CONSULT.testSuite,
       use: browserOptions[PW_BROWSER],
     },
     {
       name: 'SET_CALL_TRANSFER_CONSULT - INT',
-      dependencies: ['SET_CALL - INT', 'SET_CALL_HISTORY - INT'],
+      dependencies: ['SET_CALL - INT', 'SET_CALL_HISTORY - INT', 'SET_VOICEMAIL - INT'],
       testDir: './playwright/suites',
       testMatch: USER_SETS.SET_CALL_TRANSFER_CONSULT.testSuite,
       use: {...browserOptions[PW_BROWSER], testEnv: 'int'} as any,
--- a/packages/calling/playwright.config.ts
+++ b/packages/calling/playwright.config.ts
@@ -1,5 +1,6 @@
 import {defineConfig, devices} from '@playwright/test';
 import dotenv from 'dotenv';
+import fs from 'fs';
 import path from 'path';
 import {USER_SETS} from './playwright/test-data';
 
@@ -7,12 +8,13 @@ import {USER_SETS} from './playwright/test-data';
 dotenv.config({path: path.resolve(__dirname, '../../.env')});
 
 const BASE_URL = process.env.PW_BASE_URL || 'https://localhost:8000';
+// Shared with cc_playwright voicemail/call flows; required for deterministic mic input in VM-CALL-001.
 const dummyAudioPath = path.resolve(__dirname, '../../cc_playwright/wav/dummyAudio.wav');
 
 // Browser selection via PW_BROWSER env var: 'chrome' (default), 'firefox', 'edge', 'safari'
 const PW_BROWSER = process.env.PW_BROWSER || 'chrome';
 
-const chromiumArgs = [
+const baseChromiumArgs = [
   '--disable-site-isolation-trials', // Allow cross-origin iframes in the same process
   '--disable-web-security', // Bypass CORS for local dev server
   '--no-sandbox', // Required for CI containers without root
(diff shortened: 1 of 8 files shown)


packages/calling/playwright/utils/voicemail.ts
```diff
@@ -0,0 +1,427 @@
...
+export const fetchVoicemailPageFromUi = async (
+  page: Page,
+  offset: number,
+  limit: number
+): Promise<VoicemailRow[]> => {
+  await clearVoicemailTable(page);
+  await page.locator(CALLING_SELECTORS.VOICEMAIL_OFFSET_INPUT).fill(String(offset));
+  await page.locator(CALLING_SELECTORS.VOICEMAIL_OFFSET_LIMIT_INPUT).fill(String(limit));
+  await page.locator(CALLING_SELECTORS.VOICEMAIL_FETCH_LIST_BTN).click({timeout: AWAIT_TIMEOUT});
+  await expect(page.locator(CALLING_SELECTORS.VOICEMAIL_TABLE_ROWS)).toHaveCount(limit, {
```
bhabalan: **Pagination step looks like it can't pass against the current sample app.**

`fetchVoicemailPageFromUi` clicks `#fetchVoicemailListButton` and then waits for `#voicemailTable tbody tr` to reach `limit` rows. But in `docs/samples/calling/app.js`, the `fetchVoicemailList()` handler wired to that button only calls `getVoicemailList(...)` and `console.log(response)` — it doesn't `await` it and never renders rows into `#voicemailTable`:

```js
function fetchVoicemailList() {
  const offset = document.getElementById('offset').value;
  const offsetLength = document.getElementById('offsetLength').value;
  const response = voicemail.getVoicemailList(parseInt(offset, 10), parseInt(offsetLength, 10), voicemailSort);
  console.log(response);
}
```

Since `clearVoicemailTable()` empties the table first, `toHaveCount(limit)` will find 0 rows and time out at `AWAIT_TIMEOUT`. Unlike `createVoiceMail()` (the `#Voice-mail` button), this handler has no rendering path. This PR doesn't touch the sample app, so either the sample's `fetchVoicemailList()` needs to render rows (in this PR) or the pagination assertion needs to be reworked. Can you confirm how this step passed locally?

bhabalan: Minor: asserting `toHaveCount(limit)` exactly couples the assertion to there being at least `limit` messages on the server. It's fine for `limit=1` here, but if this helper is reused with a larger limit than the mailbox holds it'll flake. Consider asserting `> 0` (or `<= limit`) unless an exact count is the thing under test.

packages/calling/playwright/utils/voicemail.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  if (typeof read === 'string') {
+    return read === 'true';
+  }
+
+  if (read && '$' in read) {
+    return read.$ === 'true';
+  }
+
+  return Boolean(read && Object.keys(read).length === 0);
```
bhabalan: This final fallback treats an empty object `{}` as **read = true**, which is a surprising heuristic (an empty/self-closing `read` element arguably means the opposite, or is simply absent). Right now `isVoicemailRead` only feeds `attachVoicemailSummary` (a logged artifact), so it's low-impact — but if it's ever used for an assertion this branch could silently invert the result. Worth a comment explaining the WxC payload shape this is handling, or dropping the branch.

packages/calling/playwright/utils/voicemail.ts
```diff
@@ -0,0 +1,427 @@
...
+  const statusCode = await page.evaluate(async (id) => {
+    const response = await (window as any).voicemail.voicemailMarkAsUnread(id);
+
+    return response?.statusCode;
+  }, messageId);
+
+  expect([200, 204]).toContain(statusCode);
+};
+
+export const waitForVoicemailReadState = async (
```
bhabalan, line 289: `waitForVoicemailReadState` is exported but not referenced anywhere in the suite (the test verifies read state via `expectVoicemailReadInUi` in the UI instead). If you don't plan to use it, consider removing it to avoid dead code; if it's intended for the data-level read assertion the test currently lacks, wiring it in would make the "marked read" claim in the test title actually verified at the API level.

packages/calling/playwright.config.ts
```diff
@@ -19,6 +20,7 @@ const chromiumArgs = [
   '--allow-file-access-from-files', // Allow file:// protocol access
   '--use-fake-ui-for-media-stream', // Auto-grant camera/mic permissions without prompt
   '--use-fake-device-for-media-stream', // Use synthetic audio/video instead of real hardware
+  `--use-file-for-fake-audio-capture=${dummyAudioPath}`, // Feed deterministic audio into WebRTC calls
```
bhabalan: Heads up: this is added to the shared `chromiumArgs`, so `--use-file-for-fake-audio-capture` now feeds the dummy WAV into **every** calling suite (call, call-history, transfer, settings), not just voicemail. That's probably harmless (and arguably better/deterministic), but it's a global behavior change riding along in a voicemail PR — worth calling out in the description, or scoping to the voicemail project's `use` if you want to limit the blast radius. Also note `dummyAudioPath` points outside the package at `../../cc_playwright/wav/dummyAudio.wav`; a missing-file guard or comment on that dependency would help future readers.

packages/calling/playwright/test-groups/voicemail.ts
```diff
@@ -0,0 +1,184 @@
...
+        await expectVoicemailVisibleInUi(calleePage, voicemailRecord);
+
+        await playVoicemailFromUi(calleePage, messageId, VOICEMAIL_PLAYBACK_TIMEOUT_MS);
+        const readRecord = await getVoicemailRecords(calleePage, {offset: 0, limit: 20}).then(
+          (records) => records.find((record) => getVoicemailMessageId(record) === messageId)
+        );
+        if (!readRecord) {
+          throw new Error(`Expected voicemail ${messageId} after play`);
+        }
+        await attachVoicemailSummary(testInfo, 'read-after-play', readRecord);
```
bhabalan: The test title says the voicemail can be "marked read", but after playback the only read-state check is the UI one inside `playVoicemailFromUi` (`expectVoicemailReadInUi`). Here you re-fetch `readRecord` but only assert it *exists* — you never assert `isVoicemailRead(readRecord) === true`. Adding that (or using the unused `waitForVoicemailReadState`) would make the data-level read transition actually part of the assertion rather than just the UI display.

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: That step couldn't pass against the current sample app as written. I've reworked it: the helper no longer clicks #fetchVoicemailListButton or waits on table row count from that handler. It fills offset/limit in the UI and validates pagination via getVoicemailRecords(offset, limit) at the SDK level, then uses openVoicemailList() for the play step.

Jahnavi314: Fixed — the assertion is now records.length > 0 and records.length <= limit instead of an exact toHaveCount(limit).

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: Wired it into VM-CALL-001 after playback to verify read state at the API level via waitForVoicemailReadState(page, messageId, true).

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: Fixed — dropped that branch. Empty or missing read is treated as unread. UCM Read: 'true'|'false' is handled separately.

packages/calling/playwright.config.ts
Jahnavi314: Scoped --use-file-for-fake-audio-capture to the SET_VOICEMAIL project only. Added an fs.existsSync guard on cc_playwright/wav/dummyAudio.wav.

packages/calling/playwright/test-groups/voicemail.ts
Jahnavi314: I tried wiring waitForVoicemailReadState after playback, but the backend read flag did not reliably update even after the UI read state changed. The test now verifies the read transition through the sample UI inside playVoicemailFromUi, where the row is marked read only after voicemailMarkAsRead succeeds. I removed the unused waitForVoicemailReadState helper to avoid dead code/flaky API-level assertion.

packages/calling/playwright/test-groups/voicemail.ts
```diff
@@ -0,0 +1,178 @@
...
```
Shreyas281299: Could we assert the refetched record is actually marked read before attaching it? `playVoicemailFromUi` verifies the sample app toggled the buttons after `voicemailMarkAsRead` returned 200/204, but this refetch only checks that the record still exists. A backend regression where the read state is not persisted would still pass; polling `isVoicemailRead(readRecord) === true` (or restoring the read-state helper) would make the "marked read" part of the test meaningful.

packages/calling/playwright.config.ts
```diff
@@ -1,17 +1,20 @@
...
+import fs from 'fs';
 import path from 'path';
 import {USER_SETS} from './playwright/test-data';
 
 // .env lives at repo root
 dotenv.config({path: path.resolve(__dirname, '../../.env')});
 
 const BASE_URL = process.env.PW_BASE_URL || 'https://localhost:8000';
+// Shared with cc_playwright voicemail/call flows; required for deterministic mic input in VM-CALL-001.
+const dummyAudioPath = path.resolve(__dirname, '../../cc_playwright/wav/dummyAudio.wav');
```
Shreyas281299: This makes the calling Playwright config depend on `cc_playwright/wav/dummyAudio.wav`, which is owned by a separate contact-center-style test tree. Architecturally, that couples the calling suite to an unrelated fixture location and makes future cleanup or reorganization of `cc_playwright` risky for calling tests. Could we move or copy this fixture under `packages/calling/playwright` (or a shared test-fixtures directory with explicit ownership) and reference that instead?

packages/calling/playwright/test-groups/voicemail.ts
```diff
@@ -0,0 +1,178 @@
...
+        await expectVoicemailVisibleInCurrentUi(calleePage, paginatedRecords[0]);
+        await playVoicemailFromUi(
+          calleePage,
+          getVoicemailMessageId(paginatedRecords[0]),
+          VOICEMAIL_PLAYBACK_TIMEOUT_MS
+        );
+      } finally {
+        await cleanupActiveCalls(callerPage);
+        await loadSettings(calleePage).catch(() => {});
+        await setVoicemailSendUnansweredCalls(calleePage, false).catch(() => {});
```
Shreyas281299: The cleanup only disables `sendUnansweredCalls`, but setup above also enables voicemail if it was disabled, forces `sendAllCalls` and `sendBusyCalls` to false, and changes the unanswered ring count to `2`. Since these are persistent account settings on shared USER_1/USER_2 test accounts, the suite can leave the account in a different state for later projects or for the next run if it fails mid-flow. Could we snapshot the original voicemail settings after `loadSettings` and restore all mutated fields in `finally`?

packages/calling/playwright/utils/voicemail.ts
```diff
@@ -0,0 +1,415 @@
...
+        intervals: VOICEMAIL_POLL_INTERVALS,
+        message: `Expected new voicemail from caller ${callerNumber}`,
+      }
+    )
+    .toBe(true);
+
+  return matchingRecord as VoicemailRecord;
+};
+
+const clearVoicemailTable = async (page: Page): Promise<void> => {
```
Shreyas281299: Architecture/helper boundary: `openVoicemailList()` starts by mutating the sample app DOM, re-enabling the button and clearing the table from Playwright, rather than using app behavior. Because this suite is intended to validate the sample app + SDK flow, direct DOM changes can mask regressions where the button remains disabled or stale rows are not cleared by `createVoiceMail()`. Can we keep helpers at the user/API boundary and fix/reset the sample app through app code or a dedicated test hook instead?

packages/calling/playwright/utils/voicemail.ts
```diff
@@ -0,0 +1,415 @@
...
+  await waitForVoicemailPlaybackToEnd(page, messageId, playbackTimeoutMs);
+  await expectVoicemailReadInUi(page, rowIndex);
+};
+
+/**
+ * Validates voicemail pagination via the SDK and fills offset/limit in the sample app.
+ * Does not use #fetchVoicemailListButton — that handler only logs and does not render rows
+ * (sample-app fix tracked separately from this Playwright PR).
+ */
+export const fetchVoicemailPage = async (
```
Shreyas281299: Architecture/test responsibility: this helper fills the sample app pagination controls, but then validates pagination by calling `window.voicemail.getVoicemailList()` directly and explicitly bypasses the sample-app fetch button. That makes VM-CALL-001 look like sample-app pagination coverage while it is really SDK API coverage, so a broken `#fetchVoicemailListButton` or rendering path can still pass. Can we either move this assertion to SDK/API-level coverage, or have the Playwright sample-app test drive the UI control once the sample-app behavior is fixed?

packages/calling/playwright/test-groups/voicemail.ts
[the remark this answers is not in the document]
Jahnavi314: Addressed. The test now snapshots the original voicemail settings after loadSettings() and restores enabled, sendAllCalls, sendBusyCalls, sendUnansweredCalls, and unanswered ring count in finally.

packages/calling/playwright/test-groups/voicemail.ts
Jahnavi314: Done. I replaced the existence-only refetch check with polling for the refetched voicemail record to be marked read before attaching the summary. I also updated isVoicemailRead() to handle the XSI read: {} payload correctly.

packages/calling/playwright.config.ts
Jahnavi314: I copied the fixture under packages/calling/playwright/fixtures/dummyAudio.wav and updated packages/calling/playwright.config.ts to reference that calling-owned fixture. The original cc_playwright/wav/dummyAudio.wav is unchanged.

packages/calling/playwright/test-groups/voicemail.ts
Jahnavi314: Done. The test now captures the original voicemail settings after loadSettings() and restores all fields changed by the flow in finally, including enabled, sendAllCalls, sendBusyCalls, sendUnansweredCalls, and unanswered ring count.

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: Removed the direct DOM mutation from openVoicemailList(). It now uses the sample app Voice Mail button and waits for the rendered voicemail rows instead of manually enabling the button or clearing the table.

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: Done. Removed fetchVoicemailPage() and the SDK-only pagination assertion from VM-CALL-001, so the Playwright UI test no longer looks like it validates sample-app pagination while bypassing the fetch button.

packages/calling/playwright/utils/voicemail.ts
bhabalan: This doesn't seem to be addressed correctly. Check again.

Shreyas281299 (approved):

packages/calling/playwright/utils/voicemail.ts
Jahnavi314: Kept the empty-object handling because WxC/XSI returns a self-closing <read/> element as read: {} for read messages, while unread messages omit the read marker. Added a code comment to make that payload shape explicit.

bhabalan (approved):

PR #5092, opened 2026-07-10
Removed sessionId usage

shnaaz:

Before refactor:
Main Session:
Uses webex.internal.llm singleton with session ID 'llm-default-session'
URL from [locusInfo.info.datachannelUrl]
Events: ['event:relay.event'] ['locus.llm'], 'online'

Practice Session:
Uses  same  webex.internal.llm singleton with session ID 'llm-practice-session'
URL from [locusInfo.info.practiceSessionDatachannelUrl]
Events: ['event:relay.event:llm-practice-session'], ['locus.llm:llm-practice-session']
Waits for main session 'online' before connecting

Voicea:
Single [webex.internal.voicea] singleton shared by both sessions
[announce()] / [turnOnCaptions()]operates on whichever session is active
Caption state isCaptionBoxOn is global

Refactor:
Main Session:

[meeting.llmChannel]= dedicated [LLMChannel] instance owned by meeting
[meeting.voiceaChannel]= dedicated [VoiceaChannel] instance bound to [llmChannel]
Events emitted directly on the channel instance (no suffix)

Practice Session:
[webinar.practiceSessionLLMChannel] = dedicated [LLMChannel] instance owned by webinar
[webinar.practiceSessionVoiceaChannel]= dedicated [VoiceaChannel] instance bound to PS channel
Still waits for main session 'online' before connecting

In both architectures, when Practice Session is active:

- Main session channel stays connected
- Practice session channel connects after main session is online
- Both receive relay events, routed by binding in event header
- Services like annotation switch to PS channel during practice session

### Channel Ownership Summary

Regular Meeting:
meeting.llmChannel
meeting.voiceaChannel → listens to meeting.llmChannel

Webinar (main session only):
meeting.llmChannel
meeting.voiceaChannel → listens to meeting.llmChannel

Webinar (practice session active):
meeting.llmChannel - main session (stays connected)
webinar.practiceSessionLLMChannel - practice session (ADDITIONAL)
meeting.voiceaChannel → SWITCHES to practiceSessionLLMChannel
                        (preserves caption state, re-announces)

On practice session exit:
meeting.voiceaChannel.switchLLMChannel(meeting.llmChannel) → switches back
webinar.practiceSessionLLMChannel → disconnected & cleaned up

< DESCRIBE YOUR CHANGES >

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
+++ b/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
@@ -1,255 +1,88 @@
 /* eslint-disable require-jsdoc */
 import {WebexPlugin} from '@webex/webex-core';
 import LLMChannel, {config} from './llm';
-import {DATA_CHANNEL_WITH_JWT_TOKEN, LLM_DEFAULT_SESSION} from './constants';
-import {DataChannelTokenType} from './llm.types';
+import {DATA_CHANNEL_WITH_JWT_TOKEN} from './constants';
 
 /**
  * LLMPlugin — registered as `webex.internal.llm`.
  *
- * Maintains a Map<sessionId, LLMChannel> so multiple simultaneous LLM
- * connections (e.g. default session + practice session) can coexist.
- * All existing callers continue to work unchanged via the session-keyed API.
+ * Factory for creating LLMChannel instances. Each Meeting creates and owns
+ * its own LLMChannel(s), allowing multiple independent connections without
+ * the need for session IDs or ownership tracking.
  *
- * sessionId values are the LLM_DEFAULT_SESSION / LLM_PRACTICE_SESSION constants,
- * which match the DataChannelTokenType enum values so token keys and session keys
- * are the same namespace.
+ * This is a breaking API change from the previous design where the plugin
+ * maintained a Map of sessions and exposed session-keyed methods.
+ *
+ * Old usage (no longer supported):
+ *   webex.internal.llm.isConnected()
+ *   webex.internal.llm.registerAndConnect(url, dcUrl, token, sessionId)
+ *
+ * New usage:
+ *   const llm = webex.internal.llm.createConnection();
+ *   await llm.registerAndConnect(url, dcUrl, token);
+ *   llm.isConnected();
+ *   // When done:
+ *   await llm.disconnect();
  */
 export class LLMPlugin extends (WebexPlugin as any) {
   namespace = 'llm';
 
-  private sessions = new Map<string, LLMChannel>();
+  /**
+   * Registry of active LLM channels for interceptor lookup.
+   * Channels are registered when created and unregistered on disconnect.
+   */
+  private channels = new Set<LLMChannel>();
 
-  private getOrCreateSession(sessionId: string): LLMChannel {
-    let channel = this.sessions.get(sessionId);
+  /**
+   * Creates a new LLMChannel instance. The caller owns the channel and is
+   * responsible for connecting, disconnecting, and cleaning it up.
+   *
+   * @example
+   * const llm = webex.internal.llm.createConnection();
+   * llm.setRefreshHandler(() => meeting.refreshDataChannelToken());
+   * await llm.registerAndConnect(locusUrl, datachannelUrl, token);
+   *
+   * // Subscribe to events directly on the channel
+   * llm.on('event:relay.event', handler);
+   * llm.on('online', onlineHandler);
+   *
+   * // When done
+   * await llm.disconnect();
+   *
+   * @returns {LLMChannel} A new LLM connection instance
+   */
+  public createConnection(): LLMChannel {
+    // @ts-ignore — WebexPlugin children require {parent: this.webex}
+    const channel = new LLMChannel({parent: this.webex});
 
-    if (!channel) {
-      // @ts-ignore — WebexPlugin children require {parent: this.webex}
-      channel = new LLMChannel({parent: this.webex});
-      // Forward all events emitted by the channel up through the plugin so that
-      // callers doing llm.on('event:relay.event', ...) or llm.on('online', ...)
-      // receive events from whichever session channel emits them.
-      // Non-default sessions emit events with :<sessionId> suffix so that
-      // practice-session consumers can subscribe to session-specific names
-      // like `event:relay.event:llm-practice-session`.
-      channel.on('all', (eventName: string, ...args: any[]) => {
-        if (sessionId === LLM_DEFAULT_SESSION) {
-          this.trigger(eventName, ...args);
-        } else {
-          this.trigger(`${eventName}:${sessionId}`, ...args);
-        }
-      });
-      this.sessions.set(sessionId, channel);
-    }
+    this.channels.add(channel);
+
+    // Auto-unregister when disconnected
+    channel.on('offline', () => {
+      this.channels.delete(channel);
+    });
 
     return channel;
   }
 
-  private getSession(sessionId: string): LLMChannel | undefined {
-    return this.sessions.get(sessionId);
+  /**
+   * Returns true if the data channel token feature flag is enabled.
+   * This is a global check, not per-connection.
+   * @returns {Promise<boolean>}
+   */
+  public isDataChannelTokenEnabled(): Promise<boolean> {
+    // @ts-ignore
+    return this.webex.internal.feature.getFeature('developer', DATA_CHANNEL_WITH_JWT_TOKEN);
   }
 
-  // Before webex.internal.llm was LLChannel instance which extended Mercury so it was directly available
-  get hasEverConnected(): boolean {
-    for (const ch of this.sessions.values()) {
-      if (ch.hasEverConnected) return true;
-    }
-
-    return false;
-  }
-
-  public registerAndConnect(
-    locusUrl: string,
-    datachannelUrl: string,
-    datachannelToken?: string,
-    sessionId: string = LLM_DEFAULT_SESSION
-  ): Promise<void> {
-    const channel = this.getOrCreateSession(sessionId);
-
-    return channel.registerAndConnect(locusUrl, datachannelUrl, datachannelToken);
-  }
-
-  public disconnectLLM(
-    options: {code: number; reason: string},
-    // eslint-disable-next-line default-param-last
-    sessionId: string = LLM_DEFAULT_SESSION,
-    ownerMeetingId?: string
-  ): Promise<boolean | void> {
-    const channel = this.getSession(sessionId);
-
-    if (!channel) return Promise.resolve();
-
-    const {isOwner} = this.resolveSessionOwnership(ownerMeetingId, sessionId);
-
-    if (!isOwner) {
-      this.logger.info(`llm#disconnectLLM --> skipping, not owner of session ${sessionId}`);
-
-      return Promise.resolve(false);
-    }
-
-    return channel.disconnect(options).then(() => {
-      this.sessions.delete(sessionId);
-
-      return true;
-    });
-  }
-
-  public disconnectAllLLM(options?: {code: number; reason: string}): Promise<void> {
-    const promises = Array.from(this.sessions.entries()).map(([sessionId, channel]) =>
-      channel.disconnect(options).then(() => this.sessions.delete(sessionId))
-    );
-
-    return Promise.all(promises).then(() => undefined);
-  }
-
-  public isConnected(sessionId: string = LLM_DEFAULT_SESSION): boolean {
-    const session = this.getSession(sessionId);
-    const connected = session?.isConnected() ?? false;
-
-    return connected;
-  }
-
-  public getBinding(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
-    return this.getSession(sessionId)?.getBinding();
-  }
-
-  public getSocket(sessionId: string = LLM_DEFAULT_SESSION): any {
-    return this.getSession(sessionId)?.socket;
-  }
-
-  // Backwards-compatibility: callers that access llm.socket directly
-  // (e.g. voicea's getPublishTransport) get the default session socket.
-  get socket(): any {
-    return this.getSocket(LLM_DEFAULT_SESSION);
-  }
-
-  public getLocusUrl(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
-    return this.getSession(sessionId)?.getLocusUrl();
-  }
-
-  public getDatachannelUrl(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
-    return this.getSession(sessionId)?.getDatachannelUrl();
-  }
-
-  // tokenKey IS the sessionId (DataChannelTokenType enum values equal LLM_*_SESSION constants)
-
-  public getDatachannelToken(
-    // eslint-disable-next-line default-param-last
-    tokenKey: string = LLM_DEFAULT_SESSION,
-    ownerMeetingId?: string
-  ): string | undefined {
-    const channel = this.getSession(tokenKey);
-
-    if (!channel) return undefined;
-
-    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
-
-    if (!isOwner) {
-      this.logger.info(
-        `llm#getDatachannelToken --> skip read for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
-      );
-
-      return undefined;
-    }
-
-    return channel.getDatachannelToken();
-  }
-
-  public setDatachannelToken(
-    datachannelToken: string,
-    ownerMeetingId?: string,
-    tokenKey: string = LLM_DEFAULT_SESSION
-  ): void {
-    const channel = this.getOrCreateSession(tokenKey);
-    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
-
-    if (!isOwner) {
-      this.logger.info(
-        `llm#setDatachannelToken --> skip write for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
-      );
-
-      return;
-    }
-
-    channel.setDatachannelToken(datachannelToken);
-  }
-
-  public clearDatachannelToken(tokenKey: string, ownerMeetingId: string): void {
-    const channel = this.getSession(tokenKey);
-
-    if (!channel) return;
-
-    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, tokenKey);
-
-    if (!isOwner) {
-      this.logger.info(
-        `llm#clearDatachannelToken --> skip clear for session ${tokenKey}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
-      );
-
-      return;
-    }
-
-    channel.clearDatachannelToken();
-  }
-
-  public setRefreshHandler(
-    handler: () => Promise<{
-      body: {datachannelToken: string; datachannelTokenType: DataChannelTokenType};
-    }>,
-    ownerMeetingId?: string,
-    sessionId: string = LLM_DEFAULT_SESSION
-  ): void {
-    const channel = this.getOrCreateSession(sessionId);
-    const {isOwner, currentOwner} = this.resolveSessionOwnership(ownerMeetingId, sessionId);
-
-    if (!isOwner) {
-      this.logger.info(
-        `llm#setRefreshHandler --> skip write for session ${sessionId}; owned by ${currentOwner}, candidate ${ownerMeetingId}`
-      );
-
-      return;
-    }
-
-    channel.setRefreshHandler(handler);
-  }
-
-  public refreshDataChannelToken(sessionId: string = LLM_DEFAULT_SESSION): Promise<any> {
-    const channel = this.getSession(sessionId);
-
-    if (!channel) {
-      this.logger.warn(`llm#refreshDataChannelToken --> no channel for session ${sessionId}`);
-
-      return Promise.resolve(null);
-    }
-
-    return channel.refreshDataChannelToken();
-  }
-
-  public setOwnerMeetingId(
-    ownerMeetingId: string | undefined,
-    sessionId: string = LLM_DEFAULT_SESSION
-  ): void {
-    const channel = this.getSession(sessionId);
-
-    if (channel) channel.ownerMeetingId = ownerMeetingId;
-  }
-
-  public getOwnerMeetingId(sessionId: string = LLM_DEFAULT_SESSION): string | undefined {
-    return this.getSession(sessionId)?.ownerMeetingId;
-  }
-
-  public resolveSessionOwnership(
-    ownerMeetingId?: string,
-    sessionId: string = LLM_DEFAULT_SESSION
-  ): {currentOwner: string | undefined; isOwner: boolean} {
-    const currentOwner = this.getOwnerMeetingId(sessionId);
-    const isOwner = !currentOwner || !ownerMeetingId || currentOwner === ownerMeetingId;
-
-    return {currentOwner, isOwner};
-  }
-
+  /**
+   * Find a connection by its datachannel URL. Used by the interceptor to
+   * route token refresh requests to the correct channel.
+   * @param {string} url - The request URL to match
+   * @returns {LLMChannel | undefined}
+   */
   public getConnectionByDatachannelUrl(url: string): LLMChannel | undefined {
-    for (const channel of this.sessions.values()) {
+    for (const channel of this.channels) {
       const datachannelUrl = channel.getDatachannelUrl();
 
       if (datachannelUrl && LLMChannel.matchesDatachannelRequestUrl(url, datachannelUrl)) {
@@ -260,37 +93,37 @@ export class LLMPlugin extends (WebexPlugin as any) {
     return undefined;
   }
 
+  /**
+   * Find a locus URL by matching a request URL to an active connection's
+   * datachannel URL. Used by the interceptor.
+   * @param {string} requestUrl - The request URL to match
+   * @returns {string | undefined}
+   */
   public getLocusUrlByDatachannelUrl(requestUrl: string): string | undefined {
-    for (const channel of this.sessions.values()) {
-      const datachannelUrl = channel.getDatachannelUrl();
+    const channel = this.getConnectionByDatachannelUrl(requestUrl);
 
-      if (datachannelUrl && LLMChannel.matchesDatachannelRequestUrl(requestUrl, datachannelUrl)) {
-        return channel.getLocusUrl();
-      }
-    }
-
-    return undefined;
+    return channel?.getLocusUrl();
   }
 
-  public getSessionIdByDatachannelUrl(requestUrl: string): string | undefined {
-    for (const [sessionId, channel] of this.sessions.entries()) {
-      const datachannelUrl = channel.getDatachannelUrl();
-
-      if (datachannelUrl && LLMChannel.matchesDatachannelRequestUrl(requestUrl, datachannelUrl)) {
-        return sessionId;
-      }
-    }
-
-    return undefined;
+  /**
+   * Get all active connections. Useful for diagnostics/debugging.
+   * @returns {Set<LLMChannel>}
+   */
+  public getAllConnections(): Set<LLMChannel> {
+    return new Set(this.channels);
   }
 
-  public isDataChannelTokenEnabled(): Promise<boolean> {
-    // @ts-ignore
-    return this.webex.internal.feature.getFeature('developer', DATA_CHANNEL_WITH_JWT_TOKEN);
-  }
+  /**
+   * Disconnect all active connections. Useful for cleanup on logout.
+   * @param {object} [options] - Disconnect options
+   * @param {number} [options.code] - WebSocket close code
+   * @param {string} [options.reason] - WebSocket close reason
+   * @returns {Promise<void>}
+   */
+  public async disconnectAll(options?: {code: number; reason: string}): Promise<void> {
+    const promises = Array.from(this.channels).map((channel) => channel.disconnect(options));
 
-  public getAllConnections(): Map<string, LLMChannel> {
-    return new Map(this.sessions);
+    await Promise.all(promises);
   }
 }
 
--- a/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
+++ b/packages/@webex/internal-plugin-llm/src/llm-plugin.ts
@@ -57,11 +57,6 @@ export class LLMPlugin extends (WebexPlugin as any) {
 
     this.channels.add(channel);
 
-    // Auto-unregister when disconnected
-    channel.on('offline', () => {
-      this.channels.delete(channel);
-    });
-
     return channel;
   }
 
@@ -91,18 +86,6 @@ export class LLMPlugin extends (WebexPlugin as any) {
     }
 
     return undefined;
-  }
-
-  /**
-   * Find a locus URL by matching a request URL to an active connection's
-   * datachannel URL. Used by the interceptor.
-   * @param {string} requestUrl - The request URL to match
-   * @returns {string | undefined}
-   */
-  public getLocusUrlByDatachannelUrl(requestUrl: string): string | undefined {
-    const channel = this.getConnectionByDatachannelUrl(requestUrl);
-
-    return channel?.getLocusUrl();
   }
 
   /**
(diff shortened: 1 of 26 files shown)


packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -1,255 +1,88 @@
...
+   */
+  public createConnection(): LLMChannel {
+    // @ts-ignore — WebexPlugin children require {parent: this.webex}
+    const channel = new LLMChannel({parent: this.webex});
+
+    this.channels.add(channel);
+
+    // Auto-unregister when disconnected
+    channel.on('offline', () => {
+      this.channels.delete(channel);
```
marcin-bazyl: hmmm not too sure about that..... why are we doing it like this? shouldn't the owner be the one that decides when to delete it? I think an LLM connection can go offline temporarily and come back online, right? or is it not possible to come back from offline and a new ocnnetion always has to be created?

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -260,37 +93,37 @@ export class LLMPlugin extends (WebexPlugin as any) {
     return undefined;
   }
 
+  /**
+   * Find a locus URL by matching a request URL to an active connection's
+   * datachannel URL. Used by the interceptor.
+   * @param {string} requestUrl - The request URL to match
+   * @returns {string | undefined}
+   */
   public getLocusUrlByDatachannelUrl(requestUrl: string): string | undefined {
```
marcin-bazyl: I think this function is not called from anywhere now anymore, right? so can be removed...

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -149,6 +156,12 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
    */
   public isConnected = (): boolean => this.connected;
 
+  /**
+   * Get the underlying WebSocket
+   * @returns {any} socket
+   */
+  public getSocket = (): any => this.socket;
```
marcin-bazyl: socket is a Mercury prop, so I'm wondering if maybe Mercury would be a better place for this method? I mean if we are exposing it as public on LLM then why not also on Mercury?

packages/@webex/internal-plugin-llm/src/llm.types.ts
```diff
@@ -3,63 +3,87 @@ export enum DataChannelTokenType {
...
-  ) => string | undefined;
-  setDatachannelToken: (
-    datachannelToken: string,
-    tokenKey?: DataChannelTokenKey,
-    ownerMeetingId?: string
-  ) => void;
-  clearDatachannelToken: (tokenKey: DataChannelTokenKey, ownerMeetingId: string) => void;
+
+  /** Get the stored datachannel token. */
+  getDatachannelToken: () => string | undefined;
```
marcin-bazyl, line 47: this method and some other ones are still documented in the llm plugin readme file as being accessible through `webex.internal.llm` - we need to update the readme file with all the new api changes

packages/@webex/internal-plugin-voicea/test/unit/spec/voicea.js
```diff
@@ -177,39 +205,39 @@ describe('plugin-voicea', () => {
...
         // seqNum should increment
         assert.equal(voiceaService.seqNum, 2);
       });
 
       it('does not send if not connected', () => {
-        voiceaService.webex.internal.llm.isConnected.returns(false);
+        // Replace the channel with a disconnected one
+        const disconnectedChannel = createMockChannel({isConnected: false});
+        voiceaService.unregisterChannel('default');
+        voiceaService.registerChannel(disconnectedChannel, 'default');
```
marcin-bazyl: I must be missing something obvious, but I don't see a registerChannel() method on the voice plugin, so how does this work?

packages/@webex/plugin-meetings/src/annotation/index.ts
```diff
@@ -38,6 +47,67 @@ class AnnotationChannel extends WebexPlugin implements IAnnotationChannel {
...
+    this.channelHandlers.delete(type);
+  }
+
+  /**
+   * Get the active LLM channel (prefers practice session if connected)
+   * @returns {LLMChannel | undefined}
+   */
+  private getActiveChannel(): LLMChannel | undefined {
+    const practiceChannel = this.channels.get('practice-session');
+    if (practiceChannel?.isConnected()) {
```
marcin-bazyl: just wondering.... does annotation need to know about both practice session and default channels? I think it will only ever use one of them at a time, right? so maybe better for it to have just a single reference to one currently active channel?
in an edge case where we are in a practice session and for some reason the PS channel is disconnected temporarily this code will fall back to using the default channel and that's not correct, right?

packages/@webex/plugin-meetings/src/annotation/index.ts
```diff
@@ -38,6 +47,67 @@ class AnnotationChannel extends WebexPlugin implements IAnnotationChannel {
...
+
+  /**
+   * Indicates whether any registered LLM channel is connected.
+   * @returns {boolean}
+   */
+  private isLLMConnected(): boolean {
+    const defaultChannel = this.channels.get('default');
+    const practiceChannel = this.channels.get('practice-session');
+
+    return defaultChannel?.isConnected() || practiceChannel?.isConnected() || false;
```
marcin-bazyl: kind of same as above - if we are in a practice session, then what use do we have from knowing that default channel is connected if PS one is not? shouldn't isLLMConnected() return false in such case?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -253,16 +254,44 @@ const Breakouts = WebexPlugin.extend({
...
    * @returns {void}
    */
   listenToBroadcastMessages() {
-    if (!this.webex.internal.llm.isConnected() || this.hasSubscribedToMessage) {
+    if (!this._llmChannel?.isConnected() || this.hasSubscribedToMessage) {
       return;
     }
 
-    this.listenTo(this.webex.internal.llm, 'event:breakout.message', (event) => {
+    this.listenTo(this._llmChannel, 'event:breakout.message', (event) => {
```
marcin-bazyl, line 302: not sure, but maybe we should check if `this._llmChannel` is defined before this call?

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
shnaaz: right owner is already doing the clean up, removed this

packages/@webex/internal-plugin-voicea/test/unit/spec/voicea.js
shnaaz: Tests are broken, pls ignore them.
The flow is Meeting creates LLMChanne, Meeting creates VoiceaChannel by passing its LLMChannel, VoiceaChannel constructor subscribes to LLM events, Meeting subscribes to VoiceaChannel events, When LLM receives relay events, VoiceaChannel processes them and emits higher-level events like New_CAPTION

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -40,93 +41,64 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+          );
+        });
+
         if (meeting) {
-          result = await meeting.refreshDataChannelToken();
-        } else {
-          // @ts-ignore
-          result = await this.internal.llm.refreshDataChannelToken(sessionId);
-        }
+          const result = await (meeting as any).refreshDataChannelToken();
```
marcin-bazyl: meeting will be of type `Meeting` - we should use that instead of casting to `any` (same for lines 94, 95)

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -40,93 +41,64 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+      // 3) If no channel matches, fall back to finding the meeting by locusUrl
       refreshDataChannelToken: async (requestUrl?: string) => {
-        let sessionId;
-        let meeting;
-
-        if (typeof requestUrl === 'string') {
-          // @ts-ignore
-          sessionId = this.internal.llm.getSessionIdByDatachannelUrl?.(requestUrl);
+        // @ts-ignore
+        const channel = this.internal.llm.getConnectionByDatachannelUrl?.(requestUrl);
```
marcin-bazyl: why the `?.` ? I don't think it's needed

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5993,12 +5999,26 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
         if (!this.areVoiceaEventsSetup) {
           this.setUpVoiceaListeners();
         }
+        await this.voiceaChannel.turnOnCaptions(options?.spokenLanguage);
 
-        // @ts-ignore
-        await this.webex.internal.voicea.turnOnCaptions(options?.spokenLanguage);
+        // Also enable captions on practice session channel if in practice session
+        if (this.webinar?.practiceSessionVoiceaChannel) {
```
marcin-bazyl: this is the only place we're accessing practiceSessionVoiceaChannel outside of `Webinar` class. Maybe it would be cleaner to keep it private within the Webinar class and just expose a `turnOnCaptions` method or something like this that would do the `turnOnCaptions()` call on the voicea practice session voicea if it exists

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6047,16 +6069,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
-      ? llm.getBinding(LLM_PRACTICE_SESSION)
-      : llm.getBinding();
+    const expectedBinding = this.llmChannel?.getBinding();
 
     if (!expectedBinding || route === expectedBinding) {
       return true;
     }
 
+    // Also check practice session binding for webinars
+    const practiceSessionBinding = this.webinar?.practiceSessionLLMChannel?.getBinding();
```
marcin-bazyl: similar to other comment - here is the only place where practiceSessionLLMChannel is used outside Webinar

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6047,16 +6069,19 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+    if (practiceSessionBinding && route === practiceSessionBinding) {
```
marcin-bazyl: overall when I compare how isRelayEventRouteValid() looks and behaves in this PR compared to `next` branch, I feel that the logic is changed and possibly broken, can you check this in more detail please.

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -812,6 +812,27 @@ export default class Meeting extends StatelessWebexPlugin {
...
+
+  /**
+   * Pending datachannel token saved from join response, used when creating the LLM channel.
+   */
+  private _pendingDatachannelToken?: string;
+
+  /**
+   * Pending practice session datachannel token, passed to webinar for its LLM channel.
+   */
+  private _pendingPracticeSessionDatachannelToken?: string;
```
marcin-bazyl: would it make more sense to keep this inside Webinar class?

packages/@webex/plugin-meetings/src/breakouts/index.ts
shnaaz: checkd in line above

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: Have tested meetings, webinars, practice session, annotations, cc, whiteboards, breakouts

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -1,255 +1,83 @@
...
-          this.trigger(`${eventName}:${sessionId}`, ...args);
-        }
-      });
-      this.sessions.set(sessionId, channel);
-    }
+  /**
+   * Registry of active LLM channels for interceptor lookup.
+   * Channels are registered when created and unregistered on disconnect.
+   */
+  private channels = new Set<LLMChannel>();
```
marcin-bazyl, line 34: entries are never removed from this set, should they maybe be removed on disconnect?  or do we want to keep them for some debugging info?

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -1,255 +1,83 @@
...
+   * // Subscribe to events directly on the channel
+   * llm.on('event:relay.event', handler);
+   * llm.on('online', onlineHandler);
+   *
+   * // When done
+   * await llm.disconnect();
+   *
+   * @returns {LLMChannel} A new LLM connection instance
+   */
+  public createConnection(): LLMChannel {
```
marcin-bazyl, line 54: the function is called "createConnection" but it returns an LLMChannel - it would be better to settle on one name and use it consistently everywhere - either "connection" or "channel"

note that there are more places that use "connection" like getConnectionByDatachannelUrl, getAllConnections, etc

packages/@webex/internal-plugin-llm/src/llm.ts
marcin-bazyl: I see you added it to Mercury, but this one is still here, it can be removed now

packages/@webex/internal-plugin-voicea/src/voicea-plugin.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * VoiceaPlugin - factory for creating VoiceaChannel instances
+ * @export
+ * @class VoiceaPlugin
+ */
+export class VoiceaPlugin extends WebexPlugin {
+  namespace = VOICEA;
+
+  /**
+   * Creates a VoiceaChannel bound to the given LLMChannel
```
marcin-bazyl: might be worth adding a comment that Voicea does not take ownership of the LLMChannel, that LLMChannel's ownership stays with the caller.

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -23,41 +24,59 @@ import {
...
- * @description VoiceaChannel to hold single instance of LLM
+ * @description VoiceaChannel — handles voicea/transcription functionality for a single LLM connection.
+ * Created via `webex.internal.voicea.createChannel(llmChannel)`. The caller owns the
+ * channel and is responsible for its lifecycle.
  * @export
  * @class VoiceaChannel
  */
-export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
-  namespace = VOICEA;
+export class VoiceaChannel extends (EventEmitter as any) implements IVoiceaChannel {
```
marcin-bazyl: why are we casting EventEmitter  as any?

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -23,41 +24,59 @@ import {
...
-   * @param {Object} e
-   * @returns {undefined}
+   * Creates a VoiceaChannel bound to the given LLMChannel
+   * @param {LLMChannel} llmChannel - The LLM channel to use
+   * @param {Object} webex - The webex instance for making requests
    */
+  constructor(llmChannel: LLMChannel, webex: any) {
+    super();
+    this.llmChannel = llmChannel;
+    this.webex = webex;
```
marcin-bazyl: I'm not a big fan of storing a reference to the whole webex, because you can then access any plugin you want and this can create hidden circular dependencies. We only use webex here for sending the requests, we instead of whole webex, we could just pass webex.request in the constructor args and store just that

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -481,20 +432,20 @@ export class VoiceaChannel extends WebexPlugin implements IVoiceaChannel {
    * is announce processing
    * @returns {boolean}
    */
-  private isAnnounceProcessing = () =>
+  public isAnnounceProcessing = (): boolean =>
```
marcin-bazyl: why is this public now? (same for isAnnounceProcessed and isCaptionProcessing) - it's not called from outside the class, it doesn't feel like something that should be public

packages/@webex/plugin-meetings/src/annotation/index.ts
```diff
@@ -24,11 +25,17 @@ class AnnotationChannel extends WebexPlugin implements IAnnotationChannel {
 
   private seqNum: number;
 
-  hasSubscribedToEvents: boolean;
+  hasSubscribedToEvents!: boolean;
 
-  approvalUrl: string;
-  locusUrl: string;
-  deviceUrl: string;
+  approvalUrl!: string;
```
marcin-bazyl: are the extra `!` added because existing code already had warnings that you're fixing this way or are these new warnings as a result of the refactor?

packages/@webex/plugin-meetings/src/breakouts/index.ts
```diff
@@ -253,16 +254,44 @@ const Breakouts = WebexPlugin.extend({
...
+   * @returns {void}
+   */
+  registerLLMChannel(channel: LLMChannel) {
+    // If already subscribed to a previous channel, clean it up
+    if (this.hasSubscribedToMessage && this._llmChannel) {
+      this.stopListening(this._llmChannel);
+      this.hasSubscribedToMessage = false;
+    }
+
+    this._llmChannel = channel;
```
marcin-bazyl, line 277: _llmChannel is not declared anywhere properly

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: ok I see you moved it, but it's now public and name starting with "_" - probably best to make it private and add a public setter

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -53,6 +43,25 @@ const Webinar = WebexPlugin.extend({
     meetingId: 'string',
   },
 
+  /**
+   * LLM channel for practice session, owned by this webinar instance.
+   * @type {LLMChannel|undefined}
+   */
+  practiceSessionLLMChannel: undefined as LLMChannel | undefined,
```
marcin-bazyl: this should be private

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -53,6 +43,25 @@ const Webinar = WebexPlugin.extend({
...
+
+  /**
+   * Voicea channel for practice session, bound to practiceSessionLLMChannel.
+   * @type {VoiceaChannel|undefined}
+   */
+  practiceSessionVoiceaChannel: undefined as VoiceaChannel | undefined,
```
marcin-bazyl: this should be private

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -53,6 +43,25 @@ const Webinar = WebexPlugin.extend({
...
+
+  /**
+   * Pending practice session datachannel token, saved from join response.
+   * Used when creating the practice session LLM channel.
+   * @type {string|undefined}
+   */
+  _pendingPracticeSessionDatachannelToken: undefined as string | undefined,
```
marcin-bazyl, line 58: this should be private and a public setter added

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -405,97 +400,109 @@ const Webinar = WebexPlugin.extend({
...
+        // Emit event so clients can register listeners on the new channel
+        Trigger.trigger(
+          meeting,
+          {file: 'webinar/index', function: 'updatePSDataChannel'},
+          EVENT_TRIGGERS.MEETING_PRACTICE_SESSION_VOICEA_CHANNEL_READY,
+          {practiceSessionVoiceaChannel: this.practiceSessionVoiceaChannel}
+        );
+
+        // Announce and enable captions on practice session channel
+        this.practiceSessionVoiceaChannel?.announce?.();
```
marcin-bazyl: the last `?.` is not needed here, right? same for the call to turnOnCaptions() in line 460 - I know AI often adds it unnecessarily, can you check if there are any other places AI added it

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
```diff
@@ -40,93 +42,66 @@ export default class DataChannelAuthTokenInterceptor extends Interceptor {
...
+            throw new Error('DataChannel token refresh returned no payload');
+          }
+
+          const {datachannelToken} = result.body;
 
-        if (!result?.body) {
-          throw new Error('DataChannel token refresh returned no payload');
+          // Store token on the meeting's LLM channel if available
+          // llmChannel is private, cast required for cross-module access
+          const {llmChannel} = meeting as any;
```
marcin-bazyl: why do we need the cast "as any" here? the meeting is from `allMeetings` that's declared as `Record<string, Meeting>` - I hope we can have it without any case or in the worst case cast it as `Meeting`

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6715,133 +6670,71 @@ export default class Meeting extends StatelessWebexPlugin {
...
+        // Register event listeners on the channel
+        this.llmChannel.off('event:relay.event', this.processRelayEvent);
+        this.llmChannel.on('event:relay.event', this.processRelayEvent);
+        this.llmChannel.off(LOCUS_LLM_EVENT, this.processLocusLLMEvent);
+        this.llmChannel.on(LOCUS_LLM_EVENT, this.processLocusLLMEvent);
+        this.llmChannel.off('online', this.handleLLMOnline);
+        this.llmChannel.on('online', this.handleLLMOnline);
+
+        // Register annotation channel
+        this.annotation.registerChannel(this.llmChannel);
```
marcin-bazyl: I see annotation.registerChannel() is called from here (within updateLLMConnection()) for the main llm channel and also from Webinar class for PS channel. I'm not sure, but I think updateLLMConnection() is called in many different cases. Is it possible that updateLLMConnection() might be called while we are in a practice session and this call here could overwrite the PS llm channel that was set from Webinar class?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -812,6 +812,22 @@ export default class Meeting extends StatelessWebexPlugin {
...
+  /**
+   * The LLM channel owned by this meeting for datachannel communication.
+   * Created in updateLLMConnection, cleaned up in cleanupLLMConneciton.
+   */
+  private llmChannel?: LLMChannel;
+
+  /**
+   * The Voicea channel for transcription/captions, bound to this meeting's LLM channel.
+   */
+  voiceaChannel?: VoiceaChannel;
```
marcin-bazyl, line 828: should this maybe be private?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6545,113 +6548,68 @@ export default class Meeting extends StatelessWebexPlugin {
...
 
-    if (practiceSessionDatachannelToken) {
-      // @ts-ignore
-      this.webex.internal.llm.setDatachannelToken(
-        practiceSessionDatachannelToken,
-        this.id,
-        LLM_PRACTICE_SESSION
-      );
+    // Practice session token is handled by webinar
+    if (practiceSessionDatachannelToken && this.webinar) {
```
marcin-bazyl: this.webinar is set in constructor so it's always there, no need for the check `&& this.webinar` here

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -5993,12 +5994,20 @@ export default class Meeting extends StatelessWebexPlugin {
...
+            'Meeting:index#startTranscription --> voiceaChannel not available'
+          );
+
+          return;
+        }
+
         if (!this.areVoiceaEventsSetup) {
           this.setUpVoiceaListeners();
         }
+        await this.voiceaChannel.turnOnCaptions(options?.spokenLanguage);
```
marcin-bazyl, line 6154: question: just wondering.... if we are in a practice session, the call to `this.webinar?.turnOnCaptions()` from the next line will turn on the PS captions, but will the fact that we also do this call here cause any problem?

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: but the behaviour of this function has changed: events routed to the non‑active binding (main while in practice session, or practice while in main session) are now accepted (function returns true) where on "next" branch it returned false.

packages/@webex/plugin-meetings/src/webinar/index.ts
```diff
@@ -405,97 +400,109 @@ const Webinar = WebexPlugin.extend({
...
+   * @returns {void}
+   */
+  setupPracticeSessionVoiceaListeners(meeting) {
+    if (!this.practiceSessionVoiceaChannel || !meeting) {
+      return;
+    }
+
+    // Forward voicea events from practice session channel to meeting's voicea callbacks
+    Object.keys(meeting.voiceaListenerCallbacks).forEach((eventName) => {
+      this.practiceSessionVoiceaChannel.on(eventName, meeting.voiceaListenerCallbacks[eventName]);
```
marcin-bazyl: I'm not sure where in the code this is undone when we stop practice session. I see a call to this.practiceSessionVoiceaChannel.deregisterEvents() in cleanupPSDataChannel() but I'm not sure if this is enough - it unregisters the LLM event but I don't see the code that unregisters these voicea events

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: okay restored the original

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
shnaaz: have removed on disconnect

packages/@webex/internal-plugin-voicea/src/voicea.ts
shnaaz: for tests, made private again, they are covered indirectly though announce and tunrOnCaptions

packages/@webex/plugin-meetings/src/annotation/index.ts
shnaaz: existing code had warnings,reverted

packages/@webex/plugin-meetings/src/webinar/index.ts
shnaaz: has been removed since we now do switch

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: good catch, i will add a guard to register when not connected to PS

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: that line has been removed as we moved to switch llm channel

packages/@webex/plugin-meetings/src/webinar/index.ts
shnaaz: This has bee removed now

packages/@webex/plugin-meetings/src/interceptors/dataChannelAuthToken.ts
shnaaz: this is because llmChannel is private, added setLLMChannelDataToken method to do this instead

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: I see the setter added, but the prop is still public

packages/@webex/plugin-meetings/src/meeting/index.ts
marcin-bazyl: I see this is now private, but I also see it's being accessed in lots of places by the web app in the web app PR, so feels like this shouldn't be private after all? or we need some public getter

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
```diff
@@ -56,6 +56,7 @@ export class LLMPlugin extends (WebexPlugin as any) {
     const channel = new LLMChannel({parent: this.webex});
 
     this.channels.add(channel);
+    channel.onDisconnect = () => this.channels.delete(channel);
```
marcin-bazyl: This is not great. As onDisconnect is public anyone else can overwrite it and the LLM plugin's one won't get called. We should do this properly via an event registration and emit

packages/@webex/internal-plugin-voicea/src/voicea.ts
```diff
@@ -124,6 +126,45 @@ export class VoiceaChannel extends (EventEmitter as any) implements IVoiceaChann
...
+
+    // Reset announcement state for new connection
+    this.announceStatus = ANNOUNCE_STATUS.IDLE;
+    this.captionStatus = TURN_ON_CAPTION_STATUS.IDLE;
+    this.captionServiceId = undefined;
+    this.areCaptionsEnabled = false;
+
+    // Re-announce and re-enable captions if they were on
+    if (captionsWereOn) {
+      await this.turnOnCaptions(spokenLanguage);
```
marcin-bazyl, line 166: question: just wondering if maybe we need to do something with `currentCaptionLanguage` too?

packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -6726,8 +6733,11 @@ export default class Meeting extends StatelessWebexPlugin {
         this.llmChannel.off('online', this.handleLLMOnline);
         this.llmChannel.on('online', this.handleLLMOnline);
 
-        // Register annotation channel
-        this.annotation.registerChannel(this.llmChannel);
+        // Register annotation channel only if not in practice session
+        // (practice session manages its own annotation channel via updatePSDataChannel)
+        if (!this.webinar?.isPracticeSessionLLMChannelConnected()) {
```
marcin-bazyl: shouldn't this be only checking if we're in practice session no matter if the PS LLM channel is connected or not? I mean there can be a case of being in PS but with disconnected LLM channel and in that case we still want annotation to stay on the PS LLM channel and not use the main one, right?

marcin-bazyl (approved):
approving, so that you're not blocked when I'm on PTO, but there are 6 open comment threads right now, so please either address them by replying in this PR if no code change is needed or do the code change in another PR and reply here with just the PR link

packages/@webex/internal-plugin-llm/src/llm-plugin.ts
shnaaz: Will be done in separate PR

packages/@webex/plugin-meetings/src/meeting/index.ts
shnaaz: changed to public

packages/@webex/internal-plugin-voicea/src/voicea.ts
shnaaz: During switchLLMChannel(): We reset captionServiceId to undefined, but currentCaptionLanguage stays intact

When the new channel announces: The voicea service responds with a new captionServiceId

In onCaptionServiceIdUpdate() (lines 486-495):
if (this.captionServiceId !== serviceId) {
  this.captionServiceId = serviceId;
  if (this.currentCaptionLanguage) {
    this.requestLanguage(this.currentCaptionLanguage);  // Auto-restores!
  }
}

packages/@webex/internal-plugin-llm/src/llm.ts
```diff
@@ -257,7 +264,7 @@ export default class LLMChannel extends (Mercury as any) implements ILLMChannel
     this.datachannelUrl = undefined;
     this.datachannelToken = undefined;
     this.refreshHandler = undefined;
-    this.ownerMeetingId = undefined;
+    this.emit('disconnected');
```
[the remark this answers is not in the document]
shnaaz: covered by https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-850929

PR #5097, opened 2026-07-15
feat(meetings): emit WASM runtime performance metric on meeting create

antsukanova:
# COMPLETES [SPARK-832121](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-832121)

We can tell whether a browser *has* WebAssembly, but not whether it runs it *fast enough*.
Some browsers run WASM through a slow interpreter instead of the JIT, where real-time WASM
effects (e.g. background noise removal) perform poorly. We currently have no telemetry to
see how often real users are in that state.

- Added `emitWasmRuntimePerformance` in `meetings/index.ts`, which runs the
  `WasmRuntimeProbe.check()` from `@webex/web-capabilities` after a meeting is created and
  reports the result as a behavioral metric (`status`, `ratio`, `wasmMs`, `jsMs`,
  `correlation_id`).
- Also logs the probe result (`status`, `ratio`, `wasmMs`, `jsMs`) via
  `LoggerProxy.logger.log`, so an affected user's runtime state is visible in their uploaded
  client logs during triage — not only in aggregate metrics.
- It's wrapped in lodash `once`, so it runs at most once per page regardless of how many
  meetings are created; probe failures are logged and swallowed so they never affect
  `create()`.
- Added the `WASM_RUNTIME_PERFORMANCE` (`js_sdk_wasm_runtime_performance`) behavioral
  metric constant.
- Pins `@webex/web-capabilities` to `^1.12.0`, the first published version that ships
  `WasmRuntimeProbe`.

Examples from test app:
Screenshot 2026-07-15 at 19 04 00
Screenshot 2026-07-15 at 19 03 12

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Unit tests added in `test/unit/spec/meetings/index.js`:
- Emits `js_sdk_wasm_runtime_performance` once after a meeting is created, with the probe
  result and correlation id.
- Logs the WASM runtime status after a meeting is created.
- Emits only once even when `create()` is called multiple times.
- A probe failure is swallowed — `create()` still resolves the meeting and the error is logged.
- The probe does not re-run when `create()` returns an already-created meeting (e.g. rejoin).

  - [ ] Github Copilot
  - [x] Other - Please Specify: Cursor
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that


yarn.lock
```diff
@@ -8922,7 +8922,7 @@ __metadata:
     "@webex/test-helper-retry": "workspace:*"
     "@webex/test-helper-test-users": "workspace:*"
     "@webex/ts-sdp": ^1.8.1
-    "@webex/web-capabilities": ^1.10.0
+    "@webex/web-capabilities": "link:../../../../web-capabilities"
```
szotrabh: local changes to remove

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -210,6 +211,31 @@ export default class Meetings extends WebexPlugin {
...
   registrationStatus: MeetingRegistrationStatus;
 
+  /**
+   * Emits a metric describing how well this browser runs WebAssembly, used to spot browsers
+   * where real-time WASM effects (e.g. background noise removal) run poorly.
+   *
+   * @param {string} [correlationId] - correlation id to report the result against
+   * @returns {void}
+   */
+  private emitWasmRuntimePerformance = once((correlationId?: string): void => {
```
szotrabh: Just general comment, it might be useful to log status of WasmRuntimeProbe.check() in the log file.

szotrabh (approved):

yarn.lock
antsukanova: yep, thank you, I have attached as well in description that, and will update 👍
```
⚠️ Draft / do not merge yet. package.json currently points
@webex/web-capabilities at a local link:../../../../web-capabilities for testing
against the unpublished probe. This must be swapped to the real published
@webex/web-capabilities version before merge.
```

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: good point, thank you, updated with log for successfull case as well

fnowakow (approved):

k-wasniowski (approved):

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -210,6 +211,35 @@ export default class Meetings extends WebexPlugin {
...
+        return Metrics.sendBehavioralMetric(BEHAVIORAL_METRICS.WASM_RUNTIME_PERFORMANCE, {
+          status: result.status,
+          ratio: result.ratio,
+          wasmMs: result.wasmMs,
+          jsMs: result.jsMs,
+          correlation_id: correlationId,
+        });
+      })
+      .catch((error) => {
+        LoggerProxy.logger.error(
```
k-wasniowski, line 239: When will it throw? Maybe we should also emit a metric here

antsukanova: > When will it throw? 

In practice, it won't because it's designed so every failure mode resolves to a status instead of rejecting. So this `.catch` isn't really guarding the probe.

What it's actually guarding is the synchronous metric-send in the `.then`because it runs synchronously and reads `webex.internal.metrics / webex.config.metrics`, so if the metrics plugin/config isn't ready when create() runs, that can throw (not sure how often this happens but seems like better to be safe). 

I'd keep it as a log rather than a metric here in `catch`: the thing we'd be reporting is a failure to send a metric, so emitting another metric from that same path would likely tell us anything usefull. 

Thank for questions and happy to discuss if needed more!

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
```diff
@@ -1875,6 +1876,98 @@ describe('plugin-meetings', () => {
...
+          afterEach(() => {
+            probeCheckStub.restore();
+            metricsSpy.restore();
+          });
+
+          it('emits js_sdk_wasm_runtime_performance once after a meeting is created', async () => {
+            await webex.meetings.create(test1, test2);
+            await testUtils.flushPromises();
+
+            assert.calledOnceWithExactly(metricsSpy, 'js_sdk_wasm_runtime_performance', {
```
Kesari3008, line 1909: I see assertion for checking probeCheckStub not being called but when it is supposed to called there are no assertions added for that. Also address comment added by codex reviewer bot

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -210,6 +211,37 @@ export default class Meetings extends WebexPlugin {
...
   registrationStatus: MeetingRegistrationStatus;
 
+  /**
+   * Emits a metric describing how well this browser runs WebAssembly, used to spot browsers
+   * where real-time WASM effects (e.g. background noise removal) run poorly.
+   *
+   * @param {string} [correlationId] - correlation id to report the result against
+   * @returns {void}
+   */
+  private emitWasmRuntimePerformance = once((correlationId?: string): void => {
```
Kesari3008: Why are we keeping correlationId optional ? As of now I see that it only is being invoked in one place and that has correlationId being passed

Kesari3008 (approved):

packages/@webex/plugin-meetings/test/unit/spec/meetings/index.js
antsukanova: Good point, added it, thanks for the sharp eye on that

packages/@webex/plugin-meetings/src/meetings/index.ts
antsukanova: good point, shoud be ok to not have correlationId optional here yes, updated

PR #5123, opened 2026-07-27
fix(meetings): always emit inbound audio issue metric and bump media-core

SomeBody16:

Inbound audio issue telemetry could be skipped in multi-member meetings when no other member was currently unmuted, which caused under-reporting of inbound audio problems.

The branch also updates internal media dependency alignment to pick up the latest internal media-core/runtime fixes used by meetings-related packages.

- Updated inbound audio issue handling in meeting stats processing so the inbound audio issue metric is emitted consistently whenever the inbound audio issue event is detected.
- Preserved existing event-trigger behavior for the specific case where at least one non-self, non-paired member is unmuted.
- Updated unit tests to reflect and validate the new metric-emission behavior.
- Bumped @webex/internal-media-core from 2.26.2 to 2.28.0 in:
  - @webex/plugin-meetings
  - @webex/media-helpers
  - calling
- Refreshed lockfile entries accordingly, including transitive web client media engine update from 3.40.2 to 3.41.1.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Ran targeted unit tests for meetings:
  - yarn workspace @webex/plugin-meetings test:unit --targets meeting/index.js
- Result:
  - 758 passing
  - Exit code: 0

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

### DIFF
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -8482,6 +8482,7 @@ export default class Meeting extends StatelessWebexPlugin {
             data.issueSubType,
             this.correlationId
           );
+
           Trigger.trigger(
             this,
             {file: 'meeting/index', function: 'createStatsAnalyzer'},
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -8476,13 +8476,12 @@ export default class Meeting extends StatelessWebexPlugin {
           return !member.isSelf && !member.isPairedWithSelf && !member.isAudioMuted;
         });
 
-        this.mediaProperties.sendMediaIssueMetric(
-          'inbound_audio',
-          data.issueSubType,
-          this.correlationId
-        );
-
         if (atLeastOneUnmutedOtherMember) {
+          this.mediaProperties.sendMediaIssueMetric(
+            'inbound_audio',
+            data.issueSubType,
+            this.correlationId
+          );
           Trigger.trigger(
             this,
             {file: 'meeting/index', function: 'createStatsAnalyzer'},
--- a/packages/@webex/plugin-meetings/src/meeting/index.ts
+++ b/packages/@webex/plugin-meetings/src/meeting/index.ts
@@ -8476,13 +8476,13 @@ export default class Meeting extends StatelessWebexPlugin {
           return !member.isSelf && !member.isPairedWithSelf && !member.isAudioMuted;
         });
 
-        if (atLeastOneUnmutedOtherMember) {
-          this.mediaProperties.sendMediaIssueMetric(
-            'inbound_audio',
-            data.issueSubType,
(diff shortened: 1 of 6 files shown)


packages/@webex/plugin-meetings/src/meeting/index.ts
```diff
@@ -8476,13 +8476,13 @@ export default class Meeting extends StatelessWebexPlugin {
           return !member.isSelf && !member.isPairedWithSelf && !member.isAudioMuted;
         });
 
-        if (atLeastOneUnmutedOtherMember) {
-          this.mediaProperties.sendMediaIssueMetric(
-            'inbound_audio',
-            data.issueSubType,
-            this.correlationId
-          );
+        this.mediaProperties.sendMediaIssueMetric(
```
antsukanova: Could you please remind me why it's useful to send this metric differently in the `fromTrigger.trigger` part? It might also be a good idea to document this logic or add a quick explanation in a test case description because it is not obvious.

SomeBody16: `sendMediaIssueMetric` = telemetry side effect (throttled)
`Trigger.trigger` = local event emission side effect (app callback-oriented)

Currently, if issue will happen in 5s time span with members unmuted but they will mute themselfs moment before emitting event (for example they muting then unmuting trying to fix the problem), metric will be completly ignored for that occurance

I believe we should send metric on every occurance on event, and call `Trigger.trigger` (for ui notification) when there are unmuted members on meeting

Sending metric on every occurance:

* they are not send on method call, but accumulated

> Metrics are throttled in case some users are in a bad state and StatsMonitor emits them every 5s, so at most we'll send 1 metric every 5 minutes with counters saying how many times each media issue happened

* metric will show true numbers of occurances that can be later verified with logs, instead of possibly skipping some of them based on mute state

Sending `Trigger.trigger` only if there are unmuted members

* this triggers notification "Are you experiencing any audio issues?" so it make sense to only show it when users are unmuted in case of false positives and have some own logic in web-client to be shown only once and will trigger additional metrics
* web-client filters subtypes to only trigger notification for `ALL_PACKETS_DISCARDED` and `DECODE_RESULTS_IN_ZERO_AUDIO_LEVEL` so we won't introduce dialog for first audio packet by mistake

antsukanova: Thank you, good with that logic

antsukanova (approved):

packages/@webex/plugin-meetings/src/meeting/index.ts
SomeBody16: acutally, I'll change it because with current flow it would always send first audio frame delay and would not check for other members, I'll restore previous state as it makes sense, maybe we can improve it later with additional discussions, but for now let just deliever it

PR #5128, opened 2026-07-28
fix: mercury reconnections fail forever

marcin-bazyl:
# COMPLETES #[SPARK-826088](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-826088)

Users end up in a state where they have no Mercury connection, so they miss notifications about meeting being ended or other things and when they try to do some action like for example start screen sharing Locus fails, because the participant is no longer in the meeting

The issue is that Web app registers its own interceptor that modifies the websocket url. On reconnection SDK is reading the closed socket's url and trying to use it, but because it's the modified one, it fails the check against the u2c catalog, so reconnection fails.

Fixed this by storing the url used for the socket connection - that's the url that passes the catalog check and is later modified by web app's interceptor. When socket is closed we now use that stored url for the reconnection so it can pass the catalog check and then gets modified by the interceptor.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

unit tests and manual testing by forcing a reconnect via these console commands:
```
// first run the web app with the SDK and crosslaunch it:
https://co.webex.com/meet/<your_CEC>?launchApp=true&webClientEndpoint=https://localhost.webex.com
// then join a meeting and do this:
webex.internal.samesiteproxy.enableWebsocketProxy()
x=webex.meetings.getActiveWebrtcMeeting()
x.reconnect({networkDisconnect: true})
// at this point mercury is reconnected using the web app proxy
x.reconnect({networkDisconnect: true})
// now it's reconnected again, before the fix it would fail here complaining about not finding the url in the catalog
```

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation
  
### I certified that

---


packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -664,6 +669,14 @@ const Mercury = WebexPlugin.extend({
 
         this.logger.info(`${this.namespace} ${logPrefix} url for ${sessionId}: ${webSocketUrl}`);
 
+        // Remember the resolved URL used to open this socket so that
+        // reconnection in _onclose re-derives from a catalog-valid host. This is
+        // captured before socket.open(), which a lower layer (e.g. the same-site
+        // websocket proxy done with interceptors) may rewrite into a non-catalog host.
+        // Reusing the post-proxy native socket URL would fail
+        // host-catalog validation in _prepareUrl and block reconnection.
+        this.sessionWebSocketUrls.set(sessionId, webSocketUrl);
```
chrisadubois, line 690: " As long as a client feels it should be connected to mercury, it should retry connections to mercury until it has a stable websocket. The exact error, except for a couple of notable exceptions, doesn't matter. The client should reconnect with these general rules:
	•	Do not immediately reconnect in a loop. Have a back off algorithm on reconnects so the client is not in a fast loop trying to reconnect when it keeps failing (for whatever reason).
	•	Obey 429 errors following the standard ideas about rate limiting for any rest server returning a 429.
	•	If a 404 is received during the websocket setup call, that means the client's registration is gone/expired and the client needs to get a new registration through WDM and try again.
	•	Every other error, the client should retry following the rules above. There are no special meanings beyond 429 and 404."

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -857,13 +870,18 @@ const Mercury = WebexPlugin.extend({
...
-        socketUrl = sourceSocket.url;
-      }
+      // Reconnect using the resolved URL this session's socket was
+      // opened with, captured in _prepareAndOpenSocket. Do NOT reuse
+      // sourceSocket.url: that reflects the URL the native WebSocket was
+      // actually opened with, which a lower layer may have rewritten
+      // (e.g. via interceptors) into a host that is not in the SDK host
+      // catalog. Feeding such a URL back into _prepareUrl fails host-catalog
+      // validation and permanently blocks reconnection.
+      const socketUrl = this.sessionWebSocketUrls.get(sessionId);
```
chrisadubois, line 902: I'm not 100% sure how we are going to monitor this change to make sure it didn't cause something more catastrophic than it fixed. Is there some telemetry or at the very least, some logging we can wire up about this (I'm sure it partially already exists) / be very explicit in the PR about what we need to look at and check on to ensure this is not a detrimental change ?

packages/@webex/internal-plugin-mercury/src/mercury.js
```diff
@@ -44,6 +44,10 @@ const Mercury = WebexPlugin.extend({
       default: () => new Map(),
       type: 'object',
     },
+    sessionWebSocketUrls: {
```
chrisadubois, line 47: question: we could make this a config value (same with all the rest, backoffCalls, localClusterServiceUrls, etc since these control behavior, and aren't necessarily just instance variables. and then we could actually have some unit tests of "before + after" - plus it allows consumers to control how this change would go. 

That said, this can create a tangled mess, and is out of scope for the fix.

chrisadubois (approved):
Approving, but I think we need to be explicit about the telemetry we are using to monitor this and try to isolate the change in case there is something else we missed...

packages/@webex/internal-plugin-mercury/src/mercury.js
marcin-bazyl: not sure if it's worth it and for sure not now as we're doing a refactor of Mercury/LLM on a branch so I'd rather avoid any unnecessary changes on next branch

packages/@webex/internal-plugin-mercury/src/mercury.js
marcin-bazyl: we don't have good metrics about mercury right now. I can add some now in another PR and maybe wait a few days to collect them before merging this PR

marcin-bazyl: I was wrong, we have existing metrics:
https://app.eu.amplitude.com/analytics/cisco-cross-project/chart/e-6nzaq5sw?sharingId=lVCg3UzE&source=copy+url
so we can monitor them once this PR is deployed

jor-row (approved):
it was a little harder to repro the issue on prod than i thougth (didn't straight away fail) but i was eventually able to and then test this change successfully

PR #5139, opened 2026-08-03
fix: skip heartbeat check for main dataset if no LLM channel

WeijuanShao:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-841885

skip heartbeat check for main dataset if no LLM channel

Added a meeting-scoped LLM expected callback into HashTreeParser callbacks.
Updated heartbeat watchdog logic so main dataset skips timer setup only when:
the callback is provided, and the callback explicitly reports that current-meeting LLM is not expected.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -568,6 +569,31 @@ export default class LocusInfo extends EventsScope {
...
         locusInfoUpdateCallback: this.updateFromHashTree.bind(this, locusUrl),
         syncLatencyTracker: this.callbacks.syncLatencyTracker,
+        isLlmConnected: () => {
+          const llm = this.webex?.internal?.llm;
+
+          if (!llm?.isConnected?.()) {
+            return false;
+          }
+
+          const ownership = llm.resolveSessionOwnership?.(this.meetingId, LLM_DEFAULT_SESSION);
```
marcin-bazyl: what about practice session? (same for line 575 - isConnected() takes sessionId as an argument

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -568,6 +569,31 @@ export default class LocusInfo extends EventsScope {
...
+
+          if (
+            ownership?.currentOwner &&
+            ownership.currentOwner !== this.meetingId &&
```
marcin-bazyl: are you sure we need this check to have all these conditions? isn't checking `isOwner` enough? the checks here look the same like what is already done inside resolveSessionOwnership()

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -568,6 +569,31 @@ export default class LocusInfo extends EventsScope {
...
+            !ownership.isOwner
+          ) {
+            return false;
+          }
+
+          const connectedLocusUrl = llm.getLocusUrl?.();
+
+          if (connectedLocusUrl && connectedLocusUrl !== locusUrl) {
+            return false;
```
marcin-bazyl: just to double check: is it guaranteed that llm url always changes when moving between breakouts?

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1837,6 +1838,19 @@ class HashTreeParser {
         continue;
       }
 
+      if (
+        dataSet.name === DataSetNames.MAIN &&
```
marcin-bazyl: are we sure it's only MAIN dataset? what about others datasets that are sent over LLM like atd-unmuted, atd-active?
We already have LLM_DATASET_NAMES defined that contains them, maybe it should be used here?

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -1837,6 +1838,19 @@ class HashTreeParser {
...
+        this.callbacks.isLlmConnected &&
+        !this.callbacks.isLlmConnected()
+      ) {
+        LoggerProxy.logger.info(
+          `HashTreeParser#resetHeartbeatWatchdogs --> ${this.debugId} skipping heartbeat watchdog timer for data set "${dataSet.name}" because LLM is disconnected`
+        );
+
+        // eslint-disable-next-line no-continue
+        continue;
```
marcin-bazyl, line 1895: this code here handles the case when we call resetHeartbeatWatchdogs() for main and LLM is not connected, but what about the case when LLM is not connected yet, but it is connecting and later when it becomes connected, we will end up with no heartbeat watchdog. I think we need to check if we need to restart any heartbeat watchdogs when LLM changes from "not connected" to "connected".

packages/@webex/plugin-meetings/src/locus-info/index.ts
WeijuanShao: The practice session uses a separate LLM, so it shouldn't affect the connection for `LLM_DEFAULT_SESSION`. Since the default parameter of `isConnected()` is `LLM_DEFAULT_SESSION`, so isConnected with empty parameter is OK. And I think we don't need to consider the practice session connection here.

packages/@webex/plugin-meetings/src/locus-info/index.ts
WeijuanShao: Yes, both the `locusUrl` and `dataChannelUrl` change when switching breakout sessions.

packages/@webex/plugin-meetings/src/locus-info/index.ts
```diff
@@ -568,6 +569,27 @@ export default class LocusInfo extends EventsScope {
       callbacks: {
         locusInfoUpdateCallback: this.updateFromHashTree.bind(this, locusUrl),
         syncLatencyTracker: this.callbacks.syncLatencyTracker,
+        isLlmConnected: () => {
+          const llm = this.webex?.internal?.llm;
+
+          if (!llm?.isConnected?.()) {
```
marcin-bazyl: as discussed 1-1 on webex, we'll need to change this to check if we expect LLM to be connected instead of checking if it is actually connected

xinhyao (approved):

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
marcin-bazyl: as discussed offline, I think this comment is still valid, but now the edge case is like this:
1. we receive a message with dataset A, but isLLMExpected() returns false, so we don't start a watchdog timer
2. later state changes to JOINED so now isLLMExpected() will start returning true, but we still don't have a watchdog set
3. we don't receive any more messages for dataset A -> we never do any sync for it, because we have no watchdog

this can be handled in another PR

packages/@webex/plugin-meetings/src/hashTree/hashTreeParser.ts
```diff
@@ -110,6 +110,7 @@ export type HashTreeParserCallbacks = {
   locusInfoUpdateCallback: LocusInfoUpdateCallback;
   syncLatencyTracker?: SyncLatencyTracker;
   generateTrackingId?: GenerateTrackingId;
+  isLlmExpected?: () => boolean;
```
marcin-bazyl: there is no need for this to be optional

marcin-bazyl (approved):

PR #5141, opened 2026-08-03
fix(calling): config for reduced ice gathering time for ice-lite

rarajes2:
# COMPLETES #< Ad-hoc >

### What
When a remote peer advertises itself as an ICE-lite agent (`a=ice-lite`) in its ROAP offer, it only provides host candidates and performs no connectivity checks — so waiting the full `ICE_CANDIDATES_TIMEOUT` (3s) for local candidate gathering just stalls media negotiation. This PR lets the SDK detect ice-lite offers and apply a much shorter timeout, gated behind a new **opt-in** SDK config.

### Config
New optional `iceGathering` block on `CallingClientConfig`:

```javascript
createClient(webex, {
  iceGathering: {
    reduceTimeoutForIceLite: true, // opt-in; defaults to false (preserves 3s behavior)
    iceLiteTimeout: 500,           // optional override; defaults to ICE_LITE_CANDIDATES_TIMEOUT
  },
});
```

- `IceGatheringConfig` type defined in `calling/types.ts`, re-exported via `CallingClient/types.ts`, and exposed publicly from `index.ts`.
- Config threaded through `CallingClient` → `getCallManager`/`CallManager` → `createCall` → `Call` (positional param, no signature-object refactor).
- `getCallManager` seeds the config on first construction and refreshes it on the existing singleton (order-independent).
- `Call.getIceCandidatesTimeout()` returns the reduced timeout only when the feature is enabled **and** the buffered remote offer is ice-lite; otherwise the default 3s is used.

### Behavior
Fully backward compatible — with no config (or `reduceTimeoutForIceLite` unset), the existing 3s timeout is unchanged.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Parametrized `call.test.ts` coverage: enabled+ice-lite → reduced, custom `iceLiteTimeout` → custom, disabled → default, no config → default, enabled+non-ice-lite → default.
- `callManager.test.ts`: verifies the singleton config refresh and that created calls receive it.
- All affected suites green (`call`, `callManager`, `CallingClient`); ESLint clean (only pre-existing warnings).
- Also did a sanity test by making/receiving call, mute/unmute, hold/resume

  - [ ] Github Copilot
  - [x] Other - Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -359,7 +359,10 @@ async function initCalling(e) {
   const callingConfig = {
     clientConfig : clientConfig,
     callingClientConfig: callingClientConfig,
-    logger:loggerConfig
+    logger:loggerConfig,
+    iceGathering: {
+      reduceTimeoutForIceLite: true
+    }
   }
 
   calling = await Calling.init({webexConfig, callingConfig});
--- a/docs/samples/calling/app.js
+++ b/docs/samples/calling/app.js
@@ -346,6 +346,9 @@ async function initCalling(e) {
     },
     serviceData,
     jwe: jwtTokenForDestElm.value,
+    iceGathering: {
+      reduceTimeoutForIceLite: true,
+    },
   };
 
   if (callingClientConfig.discovery.country === 'Country') {
@@ -360,9 +363,6 @@ async function initCalling(e) {
     clientConfig : clientConfig,
     callingClientConfig: callingClientConfig,
     logger:loggerConfig,
-    iceGathering: {
-      reduceTimeoutForIceLite: true
-    }
   }
 
   calling = await Calling.init({webexConfig, callingConfig});
--- a/packages/calling/src/CallingClient/calling/call.ts
+++ b/packages/calling/src/CallingClient/calling/call.ts
@@ -44,6 +44,7 @@ import {
   DEVICES_ENDPOINT_RESOURCE,
   HOLD_ENDPOINT,
   ICE_CANDIDATES_TIMEOUT,
+  ICE_LITE_CANDIDATES_TIMEOUT,
   INITIAL_SEQ_NUMBER,
   MAX_CALL_KEEPALIVE_RETRY_COUNT,
   MEDIA_ENDPOINT_RESOURCE,
@@ -78,6 +79,7 @@ import {
   ICall,
   IceCandidateErrorEventPayload,
   IceEventPayload,
+  IceGatheringConfig,
   MediaContext,
   MidCallCallerId,
   MidCallEvent,
@@ -185,6 +187,8 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
   private callKeepaliveRetryCount = 0;
 
   private apiRequest: APIRequest;
+
+  private iceGatheringConfig?: IceGatheringConfig;
 
   private handleMediaRoapEvent = async (event: RoapMessageEvent) => {
     log.info(
@@ -455,7 +459,8 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     lineId: string,
     deleteCb: DeleteRecordCallBack,
     indicator: ServiceIndicator,
-    destination?: CallDetails
+    destination?: CallDetails,
+    iceGatheringConfig?: IceGatheringConfig
   ) {
     super();
     this.destination = destination;
@@ -464,6 +469,7 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
     this.deviceId = deviceId;
     this.serviceIndicator = indicator;
     this.lineId = lineId;
+    this.iceGatheringConfig = iceGatheringConfig;
 
     /* istanbul ignore else */
     if (!this.sdkConnector.getWebex()) {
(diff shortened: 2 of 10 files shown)


packages/calling/src/CallingClient/calling/callManager.ts
```diff
@@ -518,15 +528,35 @@ export class CallManager extends Eventing<CallEventTypes> implements ICallManage
...
-export const getCallManager = (webex: WebexSDK, indicator: ServiceIndicator): ICallManager => {
+export const getCallManager = (
+  webex: WebexSDK,
+  indicator: ServiceIndicator,
+  iceGatheringConfig?: IceGatheringConfig
+): ICallManager => {
   if (!callManager) {
-    callManager = new CallManager(webex, indicator);
+    callManager = new CallManager(webex, indicator, iceGatheringConfig);
+  } else if (iceGatheringConfig) {
```
Kesari3008, line 558: Why are we adding else if condition here like this ? This can be handle at call manager constructor level

packages/calling/src/CallingClient/constants.ts
```diff
@@ -130,6 +130,13 @@ export const MOBIUS_EU_INT = 'mobius-eu-central-1.int.infra.webex.com';
 export const FAILOVER_CACHE_PREFIX = 'wxc-failover-state';
 export const ACTIVE_MOBIUS_STORAGE_KEY = 'wxc-active-mobius';
 export const ICE_CANDIDATES_TIMEOUT = 3000;
+/*
+ * When the remote peer advertises itself as an ICE-lite agent (a=ice-lite), it only offers
```
Kesari3008: We do not need to give such detailed description for this constant. Just saying that a reduced timeout constant introduced is enough

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2365,6 +2371,39 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+   * shorter timeout is returned to avoid stalling media negotiation for ~3 seconds.
+   *
+   * This optimization is opt-in: it only applies when the SDK is initialized with
+   * `iceGathering.reduceTimeoutForIceLite` set to `true`. Otherwise the default
+   * ICE_CANDIDATES_TIMEOUT is always used.
+   *
+   * @returns The ICE candidates timeout in milliseconds.
+   */
+  private getIceCandidatesTimeout(): number {
+    if (!this.iceGatheringConfig?.reduceTimeoutForIceLite) {
```
Kesari3008, line 2389: If this flag is true, aren't we supposed to use reduced timeout so why are we returning ICE_CANDDIDATES_TIMEOUT here

packages/calling/src/CallingClient/calling/call.ts
```diff
@@ -2365,6 +2371,39 @@ export class Call extends Eventing<CallEventTypes> implements ICall {
...
+      return ICE_CANDIDATES_TIMEOUT;
+    }
+
+    const remoteSdp = this.remoteRoapMessage?.sdp;
+
+    if (remoteSdp && /^a=ice-lite[\r\n]*$/im.test(remoteSdp)) {
```
Kesari3008, line 2395: If we are supposed to figure out the timeout based on the SDP having ice-cite then why do we need the flag: reduceTimeoutForIceLite

Kesari3008, line 2395: Can we use some other function here instead of test to find if ice-lite is present in SDP

Kesari3008 (changes requested):

packages/calling/src/CallingClient/calling/callManager.ts
rarajes2: CallManager is a module-level singleton, so its config can only be set once at construction unless there's an explicit refresh path.

**Note**: getCallManager is called from three places, and only the first caller actually constructs the singleton -- every later call just receives the already-cached instance.

packages/calling/src/CallingClient/constants.ts
rarajes2: Done

packages/calling/src/CallingClient/calling/call.ts
rarajes2: The check is with negation

packages/calling/src/CallingClient/calling/call.ts
rarajes2: We are not opening this feature for everyone. We have kept this behind this config so that it can be executed in controlled manner and once customer confirm that this is not causing issues anymore, we will remove this config part.

By default the old behaviour will still stay.

rarajes2: We are using regular expression here so, `.test()` seems appropriate. Let me know if you have different thought

packages/calling/src/CallingClient/calling/callManager.ts
```diff
@@ -518,15 +528,35 @@ export class CallManager extends Eventing<CallEventTypes> implements ICallManage
...
+    (callManager as CallManager).setIceGatheringConfig(iceGatheringConfig);
```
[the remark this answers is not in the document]
rarajes2: This is expected behavior. `iceGatheringConfig` will be provided/omitted initially during initialization and with that the CallManager instance will be created. Further call to getCallManager will again require the config explicitly to have any change in the behavior.

packages/calling/src/CallingClient/calling/callManager.ts
Kesari3008: The config that exists in CallingClient will be passed while the CallManager object creation the very first time and we will keep returning the same instance everytime which will have the icegathering config. All the calls that are made after CallingClient initialization can use the same icegathering config

I still feel this elseIf block is not needed

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: So customers who set this config for now while initializing calling, we will perform the ice lite check only for them. let's say I dont se the config, then ice lite check that we are performing in SDP will not be done ?

packages/calling/src/CallingClient/calling/call.ts
Kesari3008: My bad, missed the negation

Kesari3008, line 2389: This still may not needed because the final return is ICE_CANDIDATES_TIMEOUT. So check should be if config is true, test the remote SDP and if test passes then returns reduced timeout else just just return ICE_CANDIDATES_TIMEOUT

Kesari3008 (approved):
Approving the PR keeping the urgency in mind but 2 changes that are being suggested seem still valid.

CallManager check and call.ts config with negation check

packages/calling/src/CallingClient/calling/call.ts
rarajes2: Correct

packages/calling/src/CallingClient/calling/call.ts
rarajes2: This will be required. If we don't keep this check the ice-lite check will still be performed irrespective of the config flag. Here we are following early return pattern

packages/calling/src/CallingClient/calling/callManager.ts
rarajes2: This will handle edge case where if next call to getCallManager need to change the behavior. Again, all these checks will go away once customer confirm that reducing the ice gathering timeout helps them with the faster call answer else we will find some other way.

rarajes2:
> Approving the PR keeping the urgency in mind but 2 changes that are being suggested seem still valid.
> 
> CallManager check and call.ts config with negation check

Addressed

packages/calling/src/CallingClient/calling/call.ts
mkesavan13, line 2395: Ideally, should have tested via `ts-sdp` but it's okay to go this way. When we make it default, we should take the ts-sdp route

mkesavan13 (approved):

docs/samples/calling/app.js
```diff
@@ -359,7 +359,10 @@ async function initCalling(e) {
   const callingConfig = {
     clientConfig : clientConfig,
     callingClientConfig: callingClientConfig,
-    logger:loggerConfig
+    logger:loggerConfig,
+    iceGathering: {
+      reduceTimeoutForIceLite: true
+    }
```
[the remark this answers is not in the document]
rarajes2: Done, Thanks

PR #5144, opened 2026-08-03
fix(webex-core): match allowed domains on DNS label boundaries

chrisadubois:
# COMPLETES #https://jira-eng-gpk2.cisco.com/jira/browse/FPV-688

`ServiceCatalog.findAllowedDomain` compared a request's host against each configured allowed domain using a substring test, rather than matching on DNS label boundaries. As a result a hostname that merely contained an allowed domain could be treated as allowed.

The result of this check is used by the auth interceptor to decide whether a request carries an `Authorization` header, so the match needs to be exact.

Match on DNS label boundaries, so a hostname is allowed only when it is the allowed domain itself or a subdomain of it:

```ts
host === domain || host.endsWith(`.${domain}`)
```

Both catalogs (`services` and `services-v2`) now share a single helper, `lib/domains.ts`. The helper also owns the url parsing, so the two catalogs cannot resolve a host differently.

Configured domains are canonicalized once when they are set, rather than on every request, so the stored list is canonical and `getAllowedDomains()` returns predictable values. Entries that are not usable hostnames are discarded at that point. Matching still normalizes defensively, because `allowedDomains` is a public property that can be assigned directly.

### Compatibility

Entries in `allowedDomains` are now matched as complete hostnames. An entry that is only a fragment of a domain — for example `example` intended to cover `example.com` — no longer matches. An entry that is empty, or otherwise not a usable hostname, is discarded when the allowed domains are set.

Every configuration shipped in this repo, and every value supplied at runtime from meeting preferences, is already a complete hostname, so no in-tree configuration changes behaviour. Downstream configurations that rely on fragments will need to use complete hostnames. The `allowedDomains` example in `packages/@webex/webex-core/README.md` predated boundary matching and has been corrected.

### Change Type

- [x] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Automated: `yarn workspace @webex/webex-core test:unit` — 33 suites, 738 tests passing. The catalog specs are seeded with fixtures shaped like a real catalog (the commercial domains, a multi-part suffix such as `broadcloud.com.au`, sites added at runtime, an entry that overlaps another, and an unset entry) and exercised with real service hostnames such as `u2c-a.wbx2.com`, `conv-a.wbx2.com`, `foobar.ciscospark.com` and `idbroker.webex.com`. Cases cover exact and subdomain matches, first-match ordering for overlapping entries, explicit ports, mixed case, trailing dots, partial labels, suffix and middle-label matching, userinfo, encoded and unusual separators in the authority, empty allowlist entries, and unparseable urls.
- Manual: verified the commercial, FedRAMP, controlled and ITE `allowedDomains` configurations against real `Services` instances — every legitimate service url still resolves to its allowed domain, no environment regresses, and a configuration supplied with mixed case, duplicates and an unset entry collapses to a single canonical entry.
- Automated: `yarn workspace @webex/plugin-meetings test:unit` — 214 passing. No `plugin-meetings` source changes; only the `fetchUserPreferredWebexSite` expectations move, because the catalog no longer stores an empty entry.
- Automated: the allowed-domain check is covered end to end through `AuthInterceptor.onRequest()` against a real `Services` and `ServicesV2` catalog, asserting the header is attached for a url under an allowed domain and that the token is never requested for one that is not. Verified these fail if the matcher is reverted to a substring comparison.
- Lint and `tsconfig.typecheck.json` clean.

  - [x] Github Copilot
  - [ ] Other - Please Specify
- [x] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1372,18 +1372,9 @@ export default class Meetings extends WebexPlugin {
           return this.request.getMeetingPreferences().then((res) => {
             if (res) {
               const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
-
-              // always take the value from this response, so that a re-register
-              // that no longer names a site clears the previous one instead of
-              // leaving a stale site in place and skipping the fallback below
               this.preferredWebexSite = preferredWebexSite;
-
-              // preferences may not name a site, and an empty value identifies
-              // no host to allow
-              if (preferredWebexSite) {
-                // @ts-ignore
-                this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
-              }
+              // @ts-ignore
+              this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
             }
 
             // fall back to getting the preferred site from the user information
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1372,9 +1372,14 @@ export default class Meetings extends WebexPlugin {
           return this.request.getMeetingPreferences().then((res) => {
             if (res) {
               const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
-              this.preferredWebexSite = preferredWebexSite;
-              // @ts-ignore
-              this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
+
+              // preferences may not name a site, in which case there is nothing
+              // to prefer and nothing to allow
+              if (preferredWebexSite) {
+                this.preferredWebexSite = preferredWebexSite;
+                // @ts-ignore
+                this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
+              }
             }
 
             // fall back to getting the preferred site from the user information
--- a/packages/@webex/plugin-meetings/src/meetings/index.ts
+++ b/packages/@webex/plugin-meetings/src/meetings/index.ts
@@ -1373,10 +1373,14 @@ export default class Meetings extends WebexPlugin {
             if (res) {
               const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
 
-              // preferences may not name a site, in which case there is nothing
-              // to prefer and nothing to allow
+              // always take the value from this response, so that a re-register
+              // that no longer names a site clears the previous one instead of
+              // leaving a stale site in place and skipping the fallback below
+              this.preferredWebexSite = preferredWebexSite;
+
+              // preferences may not name a site, and an empty value identifies
+              // no host to allow
               if (preferredWebexSite) {
-                this.preferredWebexSite = preferredWebexSite;
                 // @ts-ignore
                 this.webex.internal.services._getCatalog().addAllowedDomains([preferredWebexSite]);
               }
--- a/packages/@webex/webex-core/src/lib/domains.ts
+++ b/packages/@webex/webex-core/src/lib/domains.ts
@@ -1,0 +1,73 @@
+import Url from 'url';
+
+// Lowercase and drop any leading/trailing dots so `.Example.com` and `example.com.`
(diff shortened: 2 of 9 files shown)


packages/@webex/webex-core/src/lib/domains.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Canonicalise a list of configured allowed domains, discarding any entry that
+ * is not a usable hostname. Callers normalise on the way in so the stored list
+ * is already canonical, rather than re-deriving it on every request.
+ *
+ * @param {Array<string>} allowedDomains - The configured allowed domains.
+ * @returns {Array<string>} - Normalized, de-duplicated, non-empty entries.
+ */
+export const normalizeAllowedDomains = (allowedDomains: Array<string>): Array<string> =>
+  uniq(
+    (Array.isArray(allowedDomains) ? allowedDomains : []).map(normalizeHostname).filter(Boolean)
```
robstax, line 31: issue: The stored-list normalization is useful, but the affected meetings caller and tests need to be updated in this PR. In `plugin-meetings/src/meetings/index.ts`, only assign/add the parsed `preferredWebexSite` when it is non-empty. Then update the unit-test expectations so `['']` becomes `[]` and `['', 'site.webex.com']` becomes `['site.webex.com']`. The targeted plugin-meetings suite currently reports 13 failures from these expectations. Once addressed, the “plugin-meetings can pass an unset preferred site” item can be removed from Known follow-ups.

packages/@webex/webex-core/test/unit/spec/services-v2/service-catalog.ts
```diff
@@ -110,18 +127,68 @@ describe('webex-core', () => {
...
+        ['https://go.example-a.com/resource/id', 'example-a.com'],
+        // an explicit port must not defeat the match
+        ['https://example-a.com:8443/resource/id', 'example-a.com'],
+        ['https://u2c-a.wbx2.com:8000/resource/id', 'wbx2.com'],
+        // hostname comparison is case insensitive
+        ['https://U2C-A.WBX2.COM/resource/id', 'wbx2.com'],
+        // a trailing dot is the same name
+        ['https://u2c-a.wbx2.com./resource/id', 'wbx2.com'],
+        // a sibling registrable domain must not match
+        ['https://notwebex.com/resource/id', undefined],
```
robstax, line 154: issue: Please add this coverage to `test/unit/spec/interceptors/auth.js`, where the Authorization header is actually added. Instantiate a real `Services`/`ServicesV2`, set its catalog allowlist to `['webex.com']`, and assign it to `webex.internal.services`. Keep the real `hasAllowedDomains()` and `isAllowedDomainUrl()` methods—do not use the current `uri.includes(host)` test stub. Then call `interceptor.onRequest()` and assert:

- `https://notwebex.com` and `https://webex.com.unrelated.example` return without an `authorization` header and never call `credentials.getUserToken()`.
- `https://api.webex.com` receives the expected authorization header.

This verifies that the new catalog matcher is connected to the code that actually sends the token.

robstax (commented):
looks mostly good. two issues:
1. due to mocks/stubs, the full flow isn't tested
2. this broke a bunch of meetings plugin unit tests due to the boolean filter

packages/@webex/plugin-meetings/src/meetings/index.ts
```diff
@@ -1372,9 +1372,18 @@ export default class Meetings extends WebexPlugin {
           return this.request.getMeetingPreferences().then((res) => {
             if (res) {
               const preferredWebexSite = MeetingsUtil.parseDefaultSiteFromMeetingPreferences(res);
+
+              // always take the value from this response, so that a re-register
+              // that no longer names a site clears the previous one instead of
+              // leaving a stale site in place and skipping the fallback below
               this.preferredWebexSite = preferredWebexSite;
```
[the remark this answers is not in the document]
chrisadubois: removed src changes here for scope

chrisadubois:
@robstax have addressed but without the src changes to plugin-meetings intentionally should be in a separate PR

robstax (approved):
we should probably add checks for protocol. not a regression, but allowedDomains will accept `http`. we probably only want to accept `https` imo

Kesari3008 (approved):

Issue #11, opened 2016-07-31
Document spark.request


ianwremmel:
spark.request is defined in a package that doesn't currently get picked up by the doc generator

PR #5162, opened 2026-08-12
feat(web-extension-bridge): adds @webex/web-extension-bridge

rarajes2:
# COMPLETES #< [CAI-8448](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-8448) >

Adds `@webex/web-extension-bridge`, a new TypeScript workspace package that lets a web
application and a Chrome Manifest V3 extension exchange typed, validated messages over a
hardened `window.postMessage` + `chrome.runtime` path. Implements M1–M4 of the intake spec
(added here as `WEB-EXTENSION-BRIDGE-INTAKE-SPEC.md`).

### What it provides

Four subpath exports, one per execution context, so a web build never pulls in extension code:

| Import | Context | Surface |
| --- | --- | --- |
| `@webex/web-extension-bridge/web` | page | `createWebBridge` — publish, request handlers, connection events |
| `@webex/web-extension-bridge/extension/content` | content script | relay only, no product logic |
| `@webex/web-extension-bridge/extension/background` | MV3 service worker | `createExtensionBridge` — subscribe, request, connections, buffer |
| `@webex/web-extension-bridge/extension/client` | popup / options / side panel | `createExtensionClient`, proxies to the worker |

Core capabilities: fire-and-forget push from page to extension (buffered while the popup is
closed), on-demand pull from extension to page with timeouts and abort, connection lifecycle
events, per-topic rate limiting, and counters for observability.

### Security posture

- Strict origin allow-listing on both sides; sender verification in the worker
  (`sender.id`, `sender.tab`, `sender.origin`) rather than relying on manifest `matches`.
- Envelope validation with a fixed schema, size caps, replay protection (LRU + TTL) and
  session binding; prototype-pollution-safe parsing.
- Coded, redacted errors — no stack traces, handler messages or payloads cross the boundary,
  and no log field can carry a payload or session token.
- Least-privilege manifest (`storage` only), no `web_accessible_resources`, content script
  pre-bundled into a single classic script.
- Threat model T1–T14 has one-to-one regression tests in `test/unit/spec/security/threats.ts`.

### Samples

`docs/samples/web-extension-bridge` (web app) and `docs/samples/web-extension-bridge-extension`
(MV3 extension), built with esbuild into git-ignored `vendor/` directories. The extension
manifest is generated from a template so the allowed origin is configurable and carries a
"local testing only" banner. See the README for the click-path.

### Tests

575 unit tests (mocha/sinon/chai via legacy-tools) covering core protocol, web and extension
adapters, an integration suite that wires both halves through fake browser seams, and the
security regression suite. Lint and typecheck are clean.

### Not in this PR

Playwright end-to-end tests, the CI workflow, and release/SBOM tooling are deferred.

## Test plan

- [ ] `yarn workspace @webex/web-extension-bridge test` (lint + 575 unit tests)
- [ ] `yarn workspace @webex/web-extension-bridge build:src`
- [ ] `yarn workspace @webex/web-extension-bridge build:samples`, load the unpacked extension
      and walk the README click-path: handshake, push, pull, `HANDLER_ERROR`, `TIMEOUT`,
      `NO_HANDLER`

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Vidcast - https://app.vidcast.io/share/69f00dd6-68be-4d85-b0ff-dd476c171f83

  - [ ] Github Copilot
  - [x] Other - Cursor (Opus 5 High)
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/web-extension-bridge/src/extension/content.ts
```diff
@@ -0,0 +1,388 @@
...
+        markPageConnected();
+        break;
+
+      case EnvelopeKind.HELLO_ACK:
+        markPageConnected();
+        break;
+
+      case EnvelopeKind.PUSH:
+        if (!pushLimiter.allow(rateLimitKey(undefined, envelope.topic))) {
+          logger.warn('push rate limited', {channel, topic: envelope.topic});
```
akulakum, line 294: **P2 — Silent push loss at content relay rate limit**

After `publish()` succeeds via `postMessage`, this branch drops over-limit pushes with only a log warning — the page gets no signal and the extension never receives the message.

README says `publish()` "never silently drops" (validation/connection errors at the page API). README documents worker-eviction loss as an accepted risk, but not relay-side rate-limit drops.

Options:
1. Document relay→worker drops as best-effort (and add a counter), or
2. Propagate backpressure to the page, or
3. Expose drops via `getCounters()` on the page side.

T4 tests worker-side limiting but not this hop. Please add a regression test.

packages/@webex/web-extension-bridge/src/extension/content.ts
```diff
@@ -0,0 +1,388 @@
...
+        channel,
+        kind,
+        source: EnvelopeSource.EXTENSION,
+        topic: CONTROL_TOPIC,
+        id: nextId(),
+        session: token,
+      })
+    );
+  };
+
```
akulakum, line 154: **P2 — Swallowed worker notification failures**

All `runtime.sendMessage` rejections are caught and ignored here. After extension reload or worker eviction, the page can remain "connected" while CONNECT/PUSH never reach the worker.

README "Push is best-effort" covers worker eviction, not relay→worker delivery failure. Consider:
- Marking the page disconnected after N consecutive failures
- Incrementing a counter observable via debug/counters
- Documenting in the accepted-risks table

Please add a test simulating rejected `sendMessage` during CONNECT/PUSH.

packages/@webex/web-extension-bridge/src/core/validate.ts
```diff
@@ -0,0 +1,198 @@
...
+  if (!payloadCheck.ok) {
+    if (payloadCheck.rejection === PayloadRejection.TOO_LARGE) {
+      return drop(DropReason.PAYLOAD_TOO_LARGE);
+    }
+
+    if (payloadCheck.rejection === PayloadRejection.RESERVED_KEY) {
+      return drop(DropReason.RESERVED_KEY);
+    }
+
+    return drop(DropReason.PAYLOAD_NOT_SERIALISABLE);
```
akulakum, line 229: **P3 — `RESPONSE` error sub-object not validated**

For `RESPONSE` with `ok: false`, the `error` field shape is not validated here. `fromWireError()` mitigates by reading only `code`, but the validator should enforce `{code, message}` and reserved-key rules for consistency with the wire contract.

Please validate the error object when `ok === false`, or explicitly reject unknown keys.

packages/@webex/web-extension-bridge/src/extension/content.ts
```diff
@@ -0,0 +1,388 @@
...
+  if (existing) {
+    return existing;
+  }
+
+  const relay = createContentRelay(resolvePageWindow(), resolveChrome(), options);
+
+  started.set(channel, relay);
+
+  return relay;
```
akulakum, line 487: **P3 — `startContentRelay()` returns dead relay after `destroy()`**

`destroy()` removes listeners but does not remove the relay from the `started` map. A subsequent `startContentRelay()` returns a dead handle — common during extension hot-reload in dev.

Please call `started.delete(channel)` in `destroy()`, and add a test: destroy → startContentRelay → verify messages still forward.

packages/@webex/web-extension-bridge/src/extension/senders.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param origin - Reported sender origin.
+ * @returns Whether the origin is acceptable.
+ */
+export function isOriginAllowed(
+  allowed: Set<string> | undefined,
+  origin: string | undefined
+): boolean {
+  if (!allowed) {
+    return true;
+  }
```
akulakum: **P3 — Runtime origin allow-list optional by default**

When `allowed` is unset, all origins pass this check and the worker defers entirely to manifest `matches`. README says runtime `allowedOrigins` is "strongly recommended" — a broad manifest plus no runtime list weakens T1 defenses.

Consider requiring `allowedOrigins` in production, or emitting a one-time dev warning when unset.

packages/@webex/web-extension-bridge/src/core/validate.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  if (
+    correlationId !== null &&
+    (typeof correlationId !== 'string' ||
+      correlationId.length < 1 ||
+      correlationId.length > MAX_ID_LENGTH)
+  ) {
+    return drop(DropReason.INVALID_CORRELATION_ID);
+  }
+
```
akulakum, line 174: **P3 — Non-null `correlationId` allowed on non-RESPONSE kinds**

This block validates `correlationId` type but does not require `correlationId === null` for non-`RESPONSE` kinds. `REQUEST`/`PUSH`/`HELLO` with a stray correlation ID pass validation today.

Please require `correlationId === null` unless `kind === RESPONSE` to keep the protocol contract strict.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+flowchart LR
+    subgraph page ["Page context (untrusted)"]
+        app["your app code"]
+        wb["/web<br/>createWebBridge()"]
+    end
+    subgraph ext ["Extension (privileged)"]
+        relay["/extension/content<br/>relay — TRUST BOUNDARY"]
+        bg["/extension/background<br/>createExtensionBridge()"]
+        cli["/extension/client<br/>createExtensionClient()"]
+    end
```
akulakum, line 172: **Documentation — Link or vend the author gist**

The supplementary gist (core-concepts, overview, sdk-anatomy — manifest→SDK mapping, sample walkthrough) is excellent but lives externally and is not linked from the README.

Please either:
- Link it from this Architecture section, or
- Move content into `packages/@webex/web-extension-bridge/docs/` so it versions with releases.

packages/@webex/web-extension-bridge/src/extension/messages.ts
```diff
@@ -0,0 +1,173 @@
...
+  command: ClientCommand;
+  topic?: string;
+  payload?: JsonValue;
+  tabId?: number;
+  timeoutMs?: number;
+  limit?: number;
+}
+
+export type ClientCommandValue =
+  | JsonValue
```
akulakum, line 78: **Supporting fix — `GET_BUFFERED` command missing `topic`**

`ClientCommandMessage` has `limit` but no `topic` field. This forces the client to filter by topic after the worker applies a global limit (separate Codex thread on `client.ts`).

When fixing buffered messages, please add `topic?: string` here and handle it in the worker's `GET_BUFFERED` handler before applying `limit`.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+A handler receives `(payload, meta)` where `meta` is a frozen
+`{topic, messageId, receivedAt}`. A handler's return value is size- and
+serialisability-checked too: an oversized or circular result fails as `INVALID_PAYLOAD`
+rather than throwing inside `postMessage`.
+
+### `createExtensionBridge(options?): ExtensionBridge`
+
+| `ExtensionBridgeOptions` | Type | Default | Notes |
+| --- | --- | --- | --- |
+| `channel` | `string` | `'webex-bridge'` | Must match the page and the client. |
```
akulakum, line 327: **Documentation — `publish()` delivery guarantee is incomplete**

"Never silently drops" is accurate for validation/connection errors at the page API boundary, but misleading for:
- Content relay rate-limit drops (after successful `postMessage`)
- Worker eviction (partially covered in accepted risks, not here)

Please align this row and §2 Push flow with what "best-effort / fire-and-forget" means at each hop: page→postMessage, relay→worker, worker→buffer/subscribers.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+```bash
+nvm use 22.14
+yarn workspace @webex/web-extension-bridge build:src
+yarn workspace @webex/web-extension-bridge build:samples
+yarn samples:serve
+```
+
+`build:samples` bundles the SDK into `docs/samples/*/vendor/` and generates
+`docs/samples/web-extension-bridge-extension/manifest.json` for `https://localhost:8000`. Both
```
akulakum, line 253: **Process — Sample bootstrap is two-step**

Developers must run `build:src` then `build:samples` before samples work; `vendor/` and manifest are gitignored. The intake spec mentioned a unified `dev` script — optional nice-to-have: single command for build + watch + serve to reduce first-run friction.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+A `MessagePort` handshake was evaluated as a way to move post-handshake traffic off the
+shared `window`. It is not a security control — the port is transferred over a
+`postMessage` event every page listener receives, so it can be raced — and it is not
+implemented. If it is ever added, it will be as noise reduction, not isolation.
+
+Threats T1–T14 from the intake spec each have a dedicated regression test in
+[`test/unit/spec/security/threats.ts`](test/unit/spec/security/threats.ts). A new threat
+needs a new failing test before the fix is accepted.
+
```
akulakum, line 465: **Process — Track deferred Playwright E2E (M5)**

Playwright end-to-end tests are explicitly deferred. Manual click-path and unit/integration coverage are strong, but please open a follow-up ticket for M5 so real-browser E2E is not lost post-merge.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+| `subscribeTopic` | `(topic, listener) => () => void` | Topic-filtered `subscribe`. |
+| `request` | `<T>(topic, payload?, opts?) => Promise<T>` | FR2. `opts` is `{tabId?, timeoutMs?, signal?}`. Always settles. |
+| `listConnections` | `() => Promise<Connection[]>` | FR5. `{tabId, origin, url?, connectedAt}` per attached tab. |
+| `getBufferedMessages` | `(opts?: {topic?, limit?}) => Promise<BufferedMessage[]>` | FR8. Oldest first. |
+| `getCounters` | `() => Promise<Record<string, number>>` | Asynchronous because the counters live in the worker. |
+
+Requires `"permissions": ["storage"]`. It deliberately does **not** require the `tabs`
+permission: the active tab is resolved with `tabs.query({active: true, currentWindow:
+true})`, which needs no permission. Without host permission for a tab, `Connection.url`
+is simply absent while `origin` remains available.
```
akulakum, line 351: **Documentation — Rate-limit claim needs update**

This documents rate limiting as "Token bucket per `(tabId, topic)`" as DoS protection. Until the topic-cycling bypass is fixed (separate Codex thread on `rateLimit.ts`), a page cycling unique topics can evict buckets and get fresh token budgets per topic.

After the fix, please update this to mention an aggregate per-tab cap as well.

packages/@webex/web-extension-bridge/README.md
```diff
@@ -0,0 +1,425 @@
...
+| Wildcards | No `'*'` in `allowedOrigins` (the SDK rejects it) and no `'*'` `targetOrigin`. Confirm by lint. |
+| Extension resources | No `web_accessible_resources`. If unavoidable, a single named file with `use_dynamic_url: true`. |
+| Permissions | `permissions` is `["storage"]` or narrower. No `tabs`. No `externally_connectable`. |
+| CSP | `extension_pages` CSP set on the extension; a strict CSP, ideally with Trusted Types, on the web app. |
+| Logging | `debug: false`. Verify no payloads or tokens reach any log sink. |
+| Payload validation | Every `requestHandler` topic validates its payload against a strict schema, and every push topic is validated on receipt. |
+| Data classification | No secrets, tokens or credentials traverse the bridge. Where authenticity matters, sign server-side. |
+| Rendering | Extension and web UI render untrusted values with `textContent` only. |
+| Limits | Review `maxPayloadBytes`, `defaultTimeoutMs`, rate limits and buffer caps against your workload. |
+| Supply chain | Lockfile committed, audit clean at high severity, SBOM archived, published with provenance. |
```
akulakum, line 481: **Process — Add CI gate for package tests**

These commands are documented locally but there is no `.github/workflows` entry running `yarn workspace @webex/web-extension-bridge test` on PRs. With 575 claimed unit tests, please add a CI job (or extend `pull-request.yml`) so regressions are caught automatically.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,78 @@
+{
```
akulakum, line 1: **Process — Branch is behind `next`**

The PR is mergeable but behind the base branch. Please rebase onto current `next` and re-run the full test suite before merge.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,78 @@
...
+      "types": "./dist/types/extension/client.d.ts",
+      "import": "./dist/extension/client.js",
+      "require": "./dist/extension/client.js"
+    },
+    "./package": "./package.json"
+  },
+  "files": [
+    "dist",
+    "README.md",
+    "SECURITY.md"
```
akulakum, line 58: **Process — CHANGELOG missing for new public package**

The intake spec (§12) lists CHANGELOG. Acceptable to defer to first release, but please confirm it will be added as part of the publish workflow or track explicitly.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,78 @@
...
+  ],
+  "scripts": {
+    "prebuild": "rimraf dist",
+    "build": "yarn run -T tsc --declaration true --declarationDir ./dist/types",
+    "build:src": "webex-legacy-tools build -dest \"./dist\" -src \"./src\" -js -ts -maps && yarn build",
+    "build:samples": "node ./scripts/build-samples.mjs && node ./scripts/prepare-samples.mjs",
+    "deploy:npm": "yarn npm publish",
+    "test": "yarn test:style && yarn test:unit",
+    "test:style": "eslint \"./src/**/*.ts\" && eslint --no-eslintrc --no-ignore -c ./.eslintrc.samples.js ../../../docs/samples/web-extension-bridge/app.js ../../../docs/samples/web-extension-bridge-extension/popup.js ../../../docs/samples/web-extension-bridge-extension/service-worker.js",
+    "test:unit": "webex-legacy-tools test --unit --runner mocha",
```
akulakum, line 68: **Process — Coverage thresholds not enforced**

Intake spec targets ≥90% core / ≥80% overall coverage. `package.json` has `test:unit` but no `coverage` script or CI threshold. Please add coverage enforcement or track as an explicit follow-up before first release.

akulakum (changes requested):
## Summary

Strong architecture, security model, and documentation (README + author gist). Requesting changes on the items below **in addition to** the six existing Codex threads.

**Non-Codex findings:** silent relay-side push drops, swallowed worker notifications, stale relay registry after destroy, validator hardening gaps, doc/code mismatches, and repo integration follow-ups (CI, rebase, M5 E2E).

**Process:** Please rebase onto `next` before merge.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "url": "https://github.com/webex/webex-js-sdk.git",
+    "directory": "packages/@webex/web-extension-bridge"
+  },
+  "main": "dist/index.js",
+  "types": "dist/types/index.d.ts",
+  "devMain": "src/index.ts",
+  "sideEffects": [
+    "./dist/extension/content.js"
+  ],
+  "exports": {
```
rsarika, line 17: [P1] The primary page API should be exported from `@webex/web-extension-bridge`, not an implementation-shaped `/web` subpath. This is a new package, so make `.` the web API and use one `./extension` facade for extension APIs. Consumers should import product concepts, not source layout.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,78 @@
...
+      "types": "./dist/types/index.d.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    },
+    "./web": {
+      "types": "./dist/types/web/webBridge.d.ts",
+      "import": "./dist/web/webBridge.js",
+      "require": "./dist/web/webBridge.js"
+    },
+    "./extension/background": {
```
rsarika, line 38: [P2] Please collapse the public background/client/content subpaths into `@webex/web-extension-bridge/extension`. Keep a dedicated `./content-script` entry only for manifest wiring. This makes the supported surface discoverable without coupling consumers to execution-directory names.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,639 @@
...
+  chromeApi: ChromeLike,
+  options: ExtensionBridgeOptions = {}
+): ExtensionBridge {
+  const channel = options.channel ?? DEFAULT_CHANNEL;
+
+  if (typeof channel !== 'string' || !CHANNEL_PATTERN.test(channel)) {
+    throw new BridgeError('INSECURE_CONFIG', 'channel must match ^[a-zA-Z0-9._:-]{1,128}$');
+  }
+
+  const allowedOrigins = resolveAllowedOrigins(options.allowedOrigins);
```
rsarika, line 123: [P1] `allowedOrigins` is optional here, so this runtime check accepts every own-extension content-script sender and falls back to manifest matching. The intake spec requires `sender.origin` to be checked against a configured exact allow-list. Reject absent or empty allow-lists at construction.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,639 @@
...
+    // Synchronous limiter check, before any storage round trip, so a flood cannot
+    // queue up asynchronous work faster than it is drained.
+    if (!pushLimiter.allow(rateLimitKey(tabId, envelope.topic))) {
+      counters.increment(CounterName.RATE_LIMITED, envelope.topic);
+      logger.warn('push rate limited', {channel, tabId, topic: envelope.topic});
+
+      return;
+    }
+
+    void (async () => {
```
rsarika: [P1] This async work is launched with `void`, while the runtime message listener later returns without holding the event/response channel open. An MV3 worker can be suspended before storage and buffering finish, losing CONNECT/DISCONNECT/PUSH state. Await relay handling, return `true`, and respond once persistence has settled.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,639 @@
...
+    inFlight.releaseAll(tabId);
+
+    const settled = pending.settleAll('DISCONNECTED', tabId);
+
+    if (settled > 0) {
+      logger.debug('settled in-flight requests on disconnect', {channel, tabId, reason});
+    }
+  };
+
+  const appendToBuffer = (entry: StoredBufferEntry): Promise<StoredBufferEntry[]> =>
```
rsarika, line 325: [P1] The buffer has an entry cap but no total-byte cap. With the configurable 1 MiB payload limit, 200 entries can exceed `chrome.storage.session` quota; errors are swallowed, so buffering then silently stops. Track serialized bytes and enforce a fixed total-byte budget.

packages/@webex/web-extension-bridge/src/extension/content.ts
```diff
@@ -0,0 +1,388 @@
...
+
+// Importing this module in a content script is enough to relay the default channel,
+// which is what `import '@webex/web-extension-bridge/extension/content'` relies on.
+// Outside a page-plus-extension context (a unit test, a service worker) it does nothing.
+if (
+  typeof window !== 'undefined' &&
+  typeof (globalThis as {chrome?: unknown}).chrome === 'object' &&
+  (globalThis as {chrome?: unknown}).chrome !== null
+) {
+  startContentRelay();
```
rsarika: [P2] Please move this auto-start side effect to a dedicated `content-script` entry point. Keep this module pure and expose `startContentRelay()` through the proposed `./extension` facade, so popup, worker, and content consumers can import the API without hidden startup behavior.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/**
+ * @internal Test seam.
+ *
+ * @param chromeApi - Extension platform object.
+ * @param options - Client options.
+ * @returns A bridge-shaped client.
+ */
+export function createExtensionClientWith(
```
rsarika, line 52: [P2] `createExtensionClientWith` is a test seam but is published in the subpath declaration surface. The analogous `*With` helpers should be internal/test-only so consumers cannot construct bridges with injected platform objects.

packages/@webex/web-extension-bridge/src/core/rateLimit.ts
```diff
@@ -0,0 +1,158 @@
...
+  private readonly buckets = new Map<string, Bucket>();
+
+  private readonly perSecond: number;
+
+  private readonly maxKeys: number;
+
+  private readonly now: () => number;
+
+  public constructor(options: RateLimiterOptions = {}) {
+    this.perSecond = Math.max(options.perSecond ?? DEFAULT_PUSHES_PER_SECOND, 1);
```
rsarika: [P2] `Math.max(NaN, 1)` remains `NaN`, which makes token comparisons fail open and disables rate limiting. `Infinity` also permits effectively unbounded limits elsewhere. Validate every numeric configuration value as finite, integral, and within explicit bounds before constructing limiters/buffers.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,212 @@
...
+        topic,
+        ...(payload === undefined ? {} : {payload}),
+        ...(typeof opts.tabId === 'number' ? {tabId: opts.tabId} : {}),
+        ...(typeof opts.timeoutMs === 'number' ? {timeoutMs: opts.timeoutMs} : {}),
+      };
+
+      // The worker owns the request timer. An abort here stops the caller waiting; it
+      // cannot recall a request already in flight in the page.
+      if (opts.signal) {
+        return (await Promise.race([send(command), abortRejection(opts.signal, topic)])) as T;
```
rsarika: [P2] When `send(command)` wins this race, the listener installed by `abortRejection()` remains on a long-lived signal. Repeated requests retain closures until the signal eventually aborts. Use a settling helper that removes the abort listener on either resolution or rejection.

rsarika (changes requested):
Request changes: the package needs a cleaner consumer-facing API and the security/lifecycle issues below addressed before merge.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,212 @@
...
+    },
+
+    async getBufferedMessages(
+      opts: {topic?: string; limit?: number} = {}
+    ): Promise<BufferedMessage[]> {
+      const value = await send({
+        __webexBridgeClient: true,
+        channel,
+        command: ClientCommand.GET_BUFFERED,
+        ...(typeof opts.limit === 'number' ? {limit: opts.limit} : {}),
```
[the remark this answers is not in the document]
rarajes2: Fixed. `getBufferedMessages` now sends the validated `topic` to the worker, and the worker filters before applying `limit`. Malformed topics are rejected with `INVALID_TOPIC`; the client no longer post-filters a globally truncated result. Added regressions covering an older matching message buried behind newer messages from another topic, per-topic limit handling, and malformed-topic rejection.

packages/@webex/web-extension-bridge/src/core/serialize.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+export function checkPayload(payload: unknown, maxBytes: number): PayloadCheck {
+  if (payload === undefined) {
+    return {ok: true, bytes: 0};
+  }
+
+  let serialised: string;
+
+  try {
+    serialised = JSON.stringify(payload) as string;
```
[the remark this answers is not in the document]
rarajes2: Fixed. `inspectJson()` now recursively enforces the `JsonValue` grammar, reserved-key rule, depth bound, accessor rejection, array-shape rules, cycle detection, and an expansion budget before `JSON.stringify` measures size. It rejects functions, `undefined`, symbols, non-finite numbers, BigInt, class instances, sparse arrays, extra array keys, accessors, and oversized DAG expansions. Shared subtrees remain accepted when they are not cycles. `postMessage` failures are normalized to `BridgeError('INVALID_PAYLOAD')`. Added regressions for the grammar cases, exponential DAGs, deep sharing, array-index accessors, non-enumerables, and structured-clone failures. One proxy TOCTOU limitation remains documented as outbound-only.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,639 @@
...
+        },
+        tabId
+      );
+
+      // The relay round trip settles the promise; the timer inside `pending` settles it
+      // if the relay never answers. Whichever happens first wins, and the loser is a
+      // no-op because correlation is single-use.
+      void Promise.resolve(tabsApi.sendMessage(tabId, relayRequest)).then(
+        (response) => settleFromRelay(id, topic, connection.session, response),
+        () => pending.reject(id, new BridgeError('NOT_CONNECTED', undefined, topic))
```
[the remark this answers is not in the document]
rarajes2: Fixed. A rejected `tabsApi.sendMessage` now rejects the request with `NOT_CONNECTED` and then calls `dropConnection(tabId, 'send-failed')`, so stale records stop appearing in `listConnections()` and active-tab targeting. The order is intentional: dropping first would settle all pending requests as `DISCONNECTED` and lose the accurate error for the request that actually failed. Added a regression asserting both the error code and an empty connection list after the failed send.

packages/@webex/web-extension-bridge/src/core/json.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param value - Value to inspect. Arrays and plain objects are walked.
+ * @param reserved - Key names to reject.
+ * @returns The offending key, or `undefined` when the value is clean.
+ */
+export function findReservedKey(value: unknown, reserved: readonly string[]): string | undefined {
+  const seen = new WeakSet<object>();
+
+  const walk = (node: unknown, depth: number): string | undefined => {
+    if (typeof node !== 'object' || node === null || depth > MAX_WALK_DEPTH) {
+      return undefined;
```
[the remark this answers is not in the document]
rarajes2: Fixed. `inspectJson()` now returns a discriminated result, so depth overflow cannot be confused with “no reserved key.” It surfaces as `PayloadRejection.TOO_DEEP` and `DropReason.PAYLOAD_TOO_DEEP`, with separate accounting from unserialisable payloads. Added tests for overflow, reserved keys hidden below the bound, distinct error reporting, and a payload just inside the limit.

packages/@webex/web-extension-bridge/src/core/rateLimit.ts
```diff
@@ -0,0 +1,158 @@
...
+   * @returns Whether the message is within budget. A rejected message consumes no
+   *   token, so the caller can retry after a backoff.
+   */
+  public allow(key: string): boolean {
+    const at = this.now();
+    const existing = this.buckets.get(key);
+
+    if (!existing) {
+      this.evictIfFull();
+      this.buckets.set(key, {tokens: this.perSecond - 1, updatedAt: at});
```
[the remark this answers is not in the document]
rarajes2: Fixed. Rate limiting now has two levels sharing the same bounded bucket implementation: per `(tabId, topic)` and an aggregate per-tab bucket (default aggregate: `4 × pushesPerSecond`). Both are checked before either is charged, so rejected topic pushes do not drain the aggregate allowance. Added regressions for 500 unique topics, tab isolation, refill behavior, no token spend on rejection, and map bounding. This closes the topic-cycling bypass while retaining multi-topic headroom.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,639 @@
...
+        }
+
+        logger.debug('tab attached', {channel, tabId, origin});
+        void upsertConnection(record);
+        break;
+      }
+
+      case RelayKind.DISCONNECT:
+        logger.debug('tab detached', {channel, tabId, reason: relay.reason});
+        dropConnection(tabId, relay.reason ?? 'bye');
```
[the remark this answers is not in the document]
rarajes2: Fixed. Relay `DISCONNECT` now uses a session-scoped removal path that only removes the connection if the stored session still matches the relay token; stale disconnects are logged and ignored. Tab lifecycle events remain unconditional because they describe the tab itself. Added regressions for both stale-token ignore and current-session removal, including protection of the replacement session’s in-flight requests.

packages/@webex/web-extension-bridge/src/extension/content.ts
rarajes2: Fixed using options 1 and 3. Relay-side rate-limit drops are now counted in `relayDropped.RATE_LIMITED`, exposed through `ContentRelay.getCounters()`, and documented in the per-hop delivery/accepted-risk tables. I did not add backpressure because `publish()` has already returned after `postMessage`; making it await a relay round trip would change the fire-and-forget contract. Added a regression confirming one forwarded push and two counted drops at a one-per-second limit.

packages/@webex/web-extension-bridge/src/extension/content.ts
rarajes2: Fixed all three requested observability/lifecycle pieces. `notifyWorker` counts failures as `relaySendFailed.<kind>`, resets the consecutive-failure count after a success, and calls `markPageGone('worker-unreachable', true)` after three consecutive failures. The `tellPage` path sends `BYE`, so the page no longer reports connected after the relay gives up; the counter is cleared before that notification to avoid recursion. README documents the behavior and a regression covers rejected CONNECT/PUSH notifications.

packages/@webex/web-extension-bridge/src/core/validate.ts
rarajes2: Fixed. `isValidWireError()` now requires a non-array object with a non-empty string `code`, an optional string `message`, no reserved keys, and no unknown keys. Invalid failed responses are dropped as `DropReason.INVALID_ERROR` instead of being normalized later by `fromWireError()`. Added nine malformed-error cases plus regressions for a valid failed response and for successful responses that correctly omit `error`.

packages/@webex/web-extension-bridge/src/extension/content.ts
rarajes2: Fixed with an identity-safe variant of the requested change. `destroy()` now calls `forgetStartedRelay(relay)`, which removes only that exact relay. Deleting by channel could evict a newer replacement if an older relay is destroyed late. Added a regression that verifies repeated starts reuse the live handle, destroy removes it, and the next start returns a fresh handle/session that forwards normally.

packages/@webex/web-extension-bridge/src/extension/senders.ts
rarajes2: Fixed by making the allow-list required rather than warning. `isOriginAllowed` now takes a required set with no pass-through branch, and missing sender origins are rejected. `createExtensionBridge` rejects missing, non-array, or empty `allowedOrigins` with `INSECURE_CONFIG`; the TypeScript option is required too. The tab-URL fallback was removed because it would provide a weaker alternative to the runtime sender-origin check. This also addresses the higher-priority duplicate finding in #25.

packages/@webex/web-extension-bridge/README.md
rarajes2: Fixed. The README now has a seven-row push-delivery table covering page API validation, relay inbound validation, relay rate limiting, relay→worker send failures, worker checks/buckets, buffer eviction, and storage failures. Each hop names its counter key. “Never silently drops” is scoped to the page API boundary; downstream drops are observable but not recoverable from the page. The observability and accepted-risk sections use the same keys.

packages/@webex/web-extension-bridge/README.md
rarajes2: Tracked and intentionally left out of this package-level change. Adding a job to `.github/workflows` changes repo-wide gating and should land with the CI owners. README §9 now records the missing package test gate and the offer to prepare a separate `pull-request.yml` change.

packages/@webex/web-extension-bridge/package.json
rarajes2: Not changed here deliberately. Rebasing an open branch rewrites history and would move the anchors for these 31 review threads, making the response mapping harder to audit. The current branch was re-tested (643 passing, ESLint clean, `tsc` clean). Please rebase onto `next` as the final merge step and rerun `yarn workspace @webex/web-extension-bridge test`.

packages/@webex/web-extension-bridge/package.json
rarajes2: Tracked, using the deferred-to-first-release option. README §9 now records that `CHANGELOG.md` will be added with the first published release. The current `package.json#files` list is intentionally asserted by the T12 test, so adding it later will be a deliberate packaging and test update rather than an accidental tarball change.

packages/@webex/web-extension-bridge/package.json
rarajes2: Tracked with the CI gate rather than adding a script that nobody runs. README §9 records both the missing coverage command/threshold and the dependency on repo-level CI ownership. This change adds 68 tests (575 → 643), with 67 targeting the review-identified paths, but that does not substitute for enforced thresholds.

packages/@webex/web-extension-bridge/package.json
rarajes2: Fixed. The package root now exports `createWebBridge` plus the existing public errors/types, making `.` the page API. `createWebBridge` accesses `window` only when called, so importing the root remains safe in workers and Node type-checking. `./web` remains as a documented deprecated alias to preserve compatibility.

packages/@webex/web-extension-bridge/package.json
rarajes2: Fixed as requested. New `./extension` exposes `createExtensionBridge`, `createExtensionClient`, `startContentRelay`, and public types; `./content-script` is reserved for manifest wiring. The old subpaths remain deprecated aliases. Sample bundling still uses context-specific source modules so each bundle ships only its required slice and does not widen the runtime surface.

packages/@webex/web-extension-bridge/src/extension/background.ts
rarajes2: Fixed as specified. `resolveAllowedOrigins` now returns a required `Set<string>` and throws `INSECURE_CONFIG` for missing, non-array, or empty lists. `createExtensionBridge(options)` requires options, sender origins are checked against the configured set, and absent origins are rejected. The `tab.url` fallback was removed so it cannot bypass the runtime allow-list. `ExtensionBridgeOptions.allowedOrigins` is required at compile time, too; this also resolves #11.

packages/@webex/web-extension-bridge/src/extension/background.ts
rarajes2: Fixed. `handleRelay` and `handleRelayPush` now return promises; CONNECT, DISCONNECT, and PUSH await their persistence/buffering work. `onRuntimeMessage` returns `true` for relay messages and calls `sendResponse({ok:true})` only after completion, or `{ok:false}` on failure while incrementing `dropped.<code>`. Regressions verify that the acknowledgement is not sent until connection state and buffered pushes are visible.

packages/@webex/web-extension-bridge/src/extension/background.ts
rarajes2: Fixed both the byte cap and the swallowed-error problem. Entries record serialized `bytes` once; eviction is oldest-first until both entry and `buffer.maxBytes` limits hold. The new default is 4 MiB, bounded to 1 KiB–8 MiB, while a single entry larger than the budget is retained because it is already within `maxPayloadBytes`. `SessionStore.onWriteError` increments `storageWriteFailed.<store>` and logs a warning. Added byte-eviction and oversized-entry regressions.

packages/@webex/web-extension-bridge/src/extension/content.ts
rarajes2: Fixed as specified. `src/content-script.ts` now owns guarded auto-start and manifest wiring; `src/extension/content.ts` is pure; and `startContentRelay()` is exported through `./extension` without re-exporting the side-effect entry. `package.json#sideEffects` names only the content-script bundle, the deprecated alias preserves legacy auto-start behavior, and sample builds use the new entry.

packages/@webex/web-extension-bridge/src/extension/client.ts
rarajes2: Fixed through the facade/export-map change. Public facades expose only `createExtensionBridge`, `createExtensionClient`, and `startContentRelay`; deprecated aliases resolve to those facades, so published specifiers no longer reach the `*With` seams. The helpers remain source-level exports for relative-path unit tests, and the facade docs explain why they must not be added back.

packages/@webex/web-extension-bridge/src/core/rateLimit.ts
rarajes2: Fixed as specified. `requireBoundedInteger` now rejects non-finite, non-integral, or out-of-range limiter/buffer values with `INSECURE_CONFIG`, covering rates, bucket sizes, in-flight limits, and buffer limits. Payload sizes/timeouts remain clamped because that is safe intent handling; limiter/buffer values are rejected because coercion can fail open. Regressions cover NaN, Infinity, invalid fractions, bounds, and the replacement of the old clamping expectation.

packages/@webex/web-extension-bridge/src/extension/client.ts
rarajes2: Fixed. `raceAbort(work, signal, topic)` removes its abort listener in `finally` on both resolution and rejection and handles an already-aborted signal before installing a listener. This prevents long-lived signals from retaining closures after successful requests while preserving `ABORTED` behavior when cancellation wins. Added regressions for listener cleanup across repeated requests, cancellation wins, and already-aborted signals.

packages/@webex/web-extension-bridge/src/core/validate.ts
rarajes2: Fixed. Validation now rejects every non-`RESPONSE` envelope carrying a non-null `correlationId` with `INVALID_CORRELATION_ID`, while still accepting `null` for `PUSH`, `REQUEST`, and `HELLO`. Added regressions for stray IDs on each non-response kind and for the valid null case.

packages/@webex/web-extension-bridge/README.md
rarajes2: Fixed by vendoring the material rather than linking an external gist. Added and linked `docs/core-concepts.md`, `docs/overview.md`, and `docs/sdk-anatomy.md`, covering the envelope, validation hops, session token, push/pull contracts, bounds, integration shape, context mapping, and end-to-end flows. The docs are source-grounded and are not included in the published tarball. Please diff them against the original gist and fold in any rationale, history, or diagrams that only the author can supply.

packages/@webex/web-extension-bridge/src/extension/messages.ts
rarajes2: Fixed. `topic?: string` already existed on `ClientCommandMessage` for `REQUEST`; the issue was that `GET_BUFFERED` did not populate it. The client now sends the validated topic, the worker applies topic filtering before `limit`, and the JSDoc explains command-specific meaning and ordering. Tests are covered by the buffered-message regressions in #1.

packages/@webex/web-extension-bridge/README.md
rarajes2: Fixed. The README now has a seven-row push-delivery table covering page API validation, relay inbound validation, relay rate limiting, relay→worker send failures, worker checks/buckets, buffer eviction, and storage failures. Each hop names its counter key. “Never silently drops” is scoped to the page API boundary; downstream drops are observable but not recoverable from the page. The observability and accepted-risk sections use the same keys.

packages/@webex/web-extension-bridge/README.md
rarajes2: Tracked rather than changed. A unified build/watch/serve command would reduce friction, but choosing a watcher and dev server adds a dependency decision to a package whose zero-dependency posture is a documented security control. The follow-up is now listed in README §9 so it remains visible.

packages/@webex/web-extension-bridge/README.md
rarajes2: Tracked rather than changed. README §9 now records the deferred M5 Playwright E2E work, its current coverage (unit, integration, security, and manual click-path checks), and the need for a follow-up ticket. I cannot open the Jira ticket from this change; please raise M5 and replace the placeholder with its link before merge.

rarajes2 (commented):
Addressed the 31 inline review comments using the response map provided by the author.

packages/@webex/web-extension-bridge/src/core/serialize.ts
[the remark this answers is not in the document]
rarajes2: Time-of-check to time-of-use (TOCTOU)

akulakum (approved):
Approved — The architecture, security hardening, test coverage (643 tests), documentation, and vendored docs are in good shape. The fixes for the original Codex threads and our review comments look solid.

## Before merge (required)

Please fix the **three open Codex P2 threads** (Aug 31) before this is merged:

1. **Session-scoped send-failure cleanup** (`background.ts`) — use session-scoped removal on `tabs.sendMessage` rejection instead of unconditional `dropConnection(tabId)`, so a stale failure cannot tear down a newer session on the same tab.

2. **Client payload validation** (`client.ts`) — call `assertPayload` (or equivalent) before `runtime.sendMessage`, so invalid payloads surface as `BridgeError` rather than Chrome transport errors.

3. **Abort before send** (`client.ts`) — check `signal.aborted` before invoking `send(command)`; do not rely on `raceAbort` alone when `send(command)` is evaluated eagerly.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,884 @@
...
+
+          // A rejection here means the tab has no listener: the content script is gone
+          // after an extension reload, a discarded tab, or a failed injection, while
+          // the stored record still says otherwise. Rejecting this one request and
+          // leaving the record in place would keep `listConnections()` advertising a
+          // dead tab — and keep default active-tab targeting routing to it — until a
+          // navigation or removal event happened along. The failed send *is* that
+          // event, so it is treated as one.
+          logger.debug('dropping unreachable connection', {channel, tabId});
+          dropConnection(tabId, 'send-failed');
```
rsarika: Non-blocking follow-up: this cleanup is tab-scoped, while the request was issued against `connection.session`. If a page reloads and establishes a newer session before this older send fails, this could remove the new connection. Consider passing the expected session into cleanup, matching the stale-DISCONNECT handling above.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,241 @@
...
+        }
+      });
+    },
+
+    async request<T = JsonValue>(
+      topic: string,
+      payload?: JsonValue,
+      opts: RequestOptions = {}
+    ): Promise<T> {
+      assertTopic(topic);
```
rsarika, line 134: Non-blocking follow-up: validating the payload here before `runtime.sendMessage` would give consumers the documented `BridgeError(INVALID_PAYLOAD)` for circular or non-cloneable values, instead of a Chrome transport error.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,241 @@
...
+        topic,
+        ...(payload === undefined ? {} : {payload}),
+        ...(typeof opts.tabId === 'number' ? {tabId: opts.tabId} : {}),
+        ...(typeof opts.timeoutMs === 'number' ? {timeoutMs: opts.timeoutMs} : {}),
+      };
+
+      // The worker owns the request timer. An abort here stops the caller waiting; it
+      // cannot recall a request already in flight in the page.
+      if (opts.signal) {
+        return (await raceAbort(send(command), opts.signal, topic)) as T;
```
rsarika, line 166: Non-blocking follow-up: `send(command)` is evaluated before `raceAbort()` observes an already-aborted signal. An early `signal.aborted` check before calling `send` would avoid dispatching work that the caller has already cancelled.

packages/@webex/web-extension-bridge/package.json
```diff
@@ -0,0 +1,88 @@
...
+      "types": "./dist/types/index.d.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    },
+    "./extension": {
+      "types": "./dist/types/extension/index.d.ts",
+      "import": "./dist/extension/index.js",
+      "require": "./dist/extension/index.js"
+    },
+    "./content-script": {
```
rsarika, line 28: Small API-cleanup follow-up: since this is a new package and we do not need compatibility aliases, consider removing the layout-shaped legacy subpaths and retaining only `.`, `./extension`, and `./content-script` as the published contract.

packages/@webex/web-extension-bridge/src/extension/background.ts
```diff
@@ -0,0 +1,884 @@
...
+      // Oldest-out: the buffer is a bounded convenience, so a chatty page evicts its
+      // own history rather than growing the worker's footprint.
+      let start = Math.max(fresh.length - bufferMaxEntries, 0);
+      let bytes = 0;
+
+      for (let index = start; index < fresh.length; index += 1) {
+        bytes += fresh[index]?.bytes ?? 0;
+      }
+
+      // A single entry larger than the whole budget is still kept: it is already
```
rsarika, line 340: Non-blocking follow-up: `buffer.maxBytes` is described as a total bound, but a single larger entry is intentionally retained. Either drop/avoid persisting that entry while still broadcasting it live, or document the single-entry exception so consumers can size storage predictably.

rsarika (approved):
Approved. The main review feedback has been addressed well, especially the consumer-facing entry points, origin enforcement, persistence handling, limits, and regression coverage. I left a few non-blocking follow-ups for future hardening.

packages/@webex/web-extension-bridge/src/extension/background.ts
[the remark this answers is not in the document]
rarajes2: Confirmed, and fixed. The rejection handler now calls removeConnectionForSession(tabId, connection.session, 'send-failed'), so a delayed failure can remove only the session that issued the request, not a newer relay session on the same tab. The failed request still rejects with NOT_CONNECTED, while the replacement session and its in-flight requests remain intact. dropConnection remains tab-scoped for lifecycle events such as tabs.onRemoved/onUpdated, where the tab itself is gone. Added an interleaving regression that holds the send open, attaches session B, then fails session A’s send.

packages/@webex/web-extension-bridge/src/extension/client.ts
```diff
@@ -0,0 +1,241 @@
...
+      opts: RequestOptions = {}
+    ): Promise<T> {
+      assertTopic(topic);
+
+      const command: ClientCommandMessage = {
+        __webexBridgeClient: true,
+        channel,
+        command: ClientCommand.REQUEST,
+        topic,
+        ...(payload === undefined ? {} : {payload}),
```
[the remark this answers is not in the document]
rarajes2: Confirmed, and fixed. The client now calls assertPayload(payload, MAX_PAYLOAD_BYTES_CEILING, topic) before constructing and sending the chrome.runtime.sendMessage command. Cyclic, BigInt, nested-function, and other non-JSON payloads therefore produce the documented BridgeError('INVALID_PAYLOAD') and never reach Chrome transport. The client uses the hard 1 MiB ceiling because it has no worker-specific maxPayloadBytes option; the worker still enforces its own, possibly smaller cap. Added a regression asserting a self-referential object rejects and runtime.sendMessage is never called.

packages/@webex/web-extension-bridge/src/extension/client.ts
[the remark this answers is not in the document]
rarajes2: Confirmed, and fixed. send(command) was eagerly evaluated before raceAbort() could inspect an already-aborted signal, so a cancelled caller could still start worker/page side effects. After topic and payload validation, the client now checks opts.signal?.aborted and throws BridgeError('ABORTED') before building or sending the command. The guard inside raceAbort remains because it is also a standalone helper. Added a regression asserting ABORTED and zero runtime.sendMessage calls, while retaining the existing immediate-rejection test for the caller-visible contract.

packages/@webex/web-extension-bridge/src/extension/background.ts
rarajes2: Confirmed, and fixed. The rejection handler now calls removeConnectionForSession(tabId, connection.session, 'send-failed'), so a delayed failure can remove only the session that issued the request, not a newer relay session on the same tab. The failed request still rejects with NOT_CONNECTED, while the replacement session and its in-flight requests remain intact. dropConnection remains tab-scoped for lifecycle events such as tabs.onRemoved/onUpdated, where the tab itself is gone. Added an interleaving regression that holds the send open, attaches session B, then fails session A’s send.

packages/@webex/web-extension-bridge/src/extension/client.ts
rarajes2: Confirmed, and fixed. The client now calls assertPayload(payload, MAX_PAYLOAD_BYTES_CEILING, topic) before constructing and sending the chrome.runtime.sendMessage command. Cyclic, BigInt, nested-function, and other non-JSON payloads therefore produce the documented BridgeError('INVALID_PAYLOAD') and never reach Chrome transport. The client uses the hard 1 MiB ceiling because it has no worker-specific maxPayloadBytes option; the worker still enforces its own, possibly smaller cap. Added a regression asserting a self-referential object rejects and runtime.sendMessage is never called.

packages/@webex/web-extension-bridge/src/extension/client.ts
rarajes2: Confirmed, and fixed. send(command) was eagerly evaluated before raceAbort() could inspect an already-aborted signal, so a cancelled caller could still start worker/page side effects. After topic and payload validation, the client now checks opts.signal?.aborted and throws BridgeError('ABORTED') before building or sending the command. The guard inside raceAbort remains because it is also a standalone helper. Added a regression asserting ABORTED and zero runtime.sendMessage calls, while retaining the existing immediate-rejection test for the caller-visible contract.

packages/@webex/web-extension-bridge/package.json
rarajes2: Agreed, and done. Because the package has not been published, there is no released contract requiring compatibility aliases; keeping them would ship a deprecation cycle on day one. Removed ./web, ./extension/background, ./extension/client, and ./extension/content, retaining only ., ./extension, ./content-script, and ./package. Updated README §1 and the src/index.ts module docs to remove the deprecated-specifier text while preserving the note about hidden *With test seams. Verified after yarn build:src that every surviving export-map path resolves to an emitted file, and the existing wildcard-subpath security test still passes. The intake spec remains unchanged because it records the original request, not the final published contract.

packages/@webex/web-extension-bridge/src/extension/background.ts
rarajes2: Documented, with behavior intentionally unchanged. The prior wording implied buffer.maxBytes was an absolute total, but the implementation retains the newest entry even when it alone exceeds the budget. That exception is deliberate, bounded, and covered by the existing characterization test: storage can exceed maxBytes by at most one entry, and that entry is still limited by maxPayloadBytes (≤1 MiB). README §5 now states when it occurs (maxBytes < maxPayloadBytes), the resulting bound (max(maxBytes, one entry)), and how consumers should size chrome.storage.session. Dropping such entries would make a small buffer silently discard every oversized push, so no code change was made.

PR #5171, opened 2026-08-14
feat[contact-center]: Mid call and Post Call summaries 

rsarika:
# COMPLETES [WXCC-9127](https://jira-eng-gpk2.cisco.com/jira/browse/WXCC-9127)

WxCC generates AI call summaries — a **mid-call** summary the agent reviews before a consult or
transfer, and a **post-call** summary the agent reviews during wrapup — but the JS SDK had no
surface for them. Custom desktops built on this SDK could not request a summary, show it to the
agent, capture the agent's edits and feedback, or hand a summary to the receiving agent on a
transfer.

The delivery model makes this awkward for consumers to do themselves. Summaries are requested over
one channel and delivered asynchronously over Contact Center realtime as a separate inbound frame,
so a consumer would have to correlate request to response by `conversationId`, impose its own
timeout, and cope with frames that arrive *before* the task they belong to is registered (both the
receiving agent's summary and the per-interaction feature flags can land early). Each consumer
re-implementing that correlation is where the bugs live.

**Task API** — four methods on `ITask`, each returning a promise that settles on the matching
inbound frame:

| Method | Purpose |
| --- | --- |
| `requestPostCallSummary()` | Resolves with `PostCallSummaryEventPayload` |
| `sendPostCallSummaryResponse(payload)` | Sends the agent's edits, counters, feedback and wrapup code |
| `requestMidCallSummary(actionType)` | `'CONSULT' \| 'TRANSFER'`; resolves with `MidCallSummaryEventPayload` |
| `sendMidCallSummaryResponse(payload, actionType)` | Sends the agent's edits, counters and feedback |

**Two new task events**, so consumers do not maintain their own `interactionId → state` maps:

- `task:midCallSummaryForReceivingAgent` — the summary that travelled with a transfer, replayed on
  the receiving agent's task even when the frame arrived before that task existed.
- `task:featureEnablement` — per-interaction `midCallEnabled` / `postCallEnabled` flags used to gate
  the summary UI. Fires at registration time if the frame arrived first.

**`AISummaryCoordinator`** (new) owns the request/response correlation the consumer would otherwise
write: pending-request maps keyed by conversation, a 15s timeout per request
(`AI_SUMMARY_DURATION_MS`), buffering of early receiver summaries and orphan feature-flag
frames with matching retention windows, cancellation on task teardown, and rejection of a duplicate
in-flight request with `AI_SUMMARY_REQUEST_ALREADY_PENDING`.

**Transport** — `ApiAiAssistant.sendSummaryGetEvent()` / `sendSummaryResponseEvent()`, with routing
for the new inbound frames through `CC_AI_SUMMARY_EVENTS`.

**Failure modes are typed, not thrown strings** — `AI_SUMMARY_ERROR_CODES` covers
`POST_CALL_SUMMARY_DISABLED`, `MID_CALL_SUMMARY_DISABLED`, `POST_CALL_SUMMARY_TIMEOUT`,
`MID_CALL_SUMMARY_TIMEOUT`, `AI_ASSISTANT_BASE_URL_NOT_AVAILABLE` and
`AI_SUMMARY_REQUEST_ALREADY_PENDING`, so a consumer can distinguish "feature off" from "timed out"
without string matching.

**Behavioral metrics** for each get/response success and failure, feature-enablement receipt, and
inbound frames dropped with no matching task.

**Public types exported** from the package entry point: `PostCallSummaryEventPayload`,
`MidCallSummaryEventPayload`, `MidCallSummaryReceivingAgentPayload`, `FeatureEnablementEventPayload`,
`PostCallSummarySections`, `MidCallSummarySections`, the response payload types, `SummaryCounters`,
`AISummaryActionType`, `AISummaryFeedback`, and the summary state unions — plus
`CC_AI_SUMMARY_EVENTS`, `AI_SUMMARY_ERROR_CODES` and `AIAssistantEventName`.

**Sample app** (`docs/samples/contact-center/`) exercises the whole surface end to end: summary
panels for consult, transfer, incoming-transfer and wrapup; read-only and edit modes rendered from
the payload's `sections` map with a `summaryText` fallback; a delta payload so only edited sections
are sent; thumbs up/down feedback and view/edit/copy counters; suggested wrapup codes applied to the
dropdown; and `sessionStorage` persistence so a refresh mid-wrapup does not lose the summary.

Also in this branch: minified sample bundles (`docs/samples/*.min.js`) are removed from git tracking
and ignored, and the `ai-docs` under `contact-center/src` are updated to match the new flows.

https://app.vidcast.io/share/c826de62-f149-49ac-83f1-8cfa32a9ffc3

### Change Type

- [x] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

**Automated** — `yarn jest` in `packages/@webex/contact-center`: **34 suites, 1088 tests, all
passing.** Unit tests added or extended for every new unit:
`AISummaryCoordinator`, `Task`, `TaskManager`, `TaskUtils`, `ApiAiAssistant`, `cc`, package `index`,
`config`, `core/Utils` and `task/voice/Voice`. Coverage includes:

- post-call request → resolve → `sendPostCallSummaryResponse` with edits, counters and wrapup code
- mid-call request for both `CONSULT` and `TRANSFER`, and the response for each
- summary frame arriving **before** the task is registered (receiver buffer and orphan
  feature-enablement retention), and the buffer expiring
- request timeout at 15s for both summary types, and cancellation on task teardown
- duplicate in-flight request rejected with `AI_SUMMARY_REQUEST_ALREADY_PENDING`
- feature-disabled and missing-base-URL rejections
- inbound frames with no matching task counted as dropped
- metrics emitted on each success and failure path

**Manual, against a live org** (walkthrough linked above):

- consult and transfer from the initiating agent — summary requested, reviewed, edited, sent
- receiving agent sees the transferred summary on task assignment
- post-call summary during wrapup — reviewed, edited, feedback given, submitted with the wrapup code
- summary panels gated correctly when `midCallEnabled` / `postCallEnabled` are off
- agent wraps up before the summary arrives — response still sent once it resolves
- summary never arrives — `NOT_RECEIVED` state sent; summary ignored — `IGNORED` state sent
- refresh mid-wrapup restores the pending post-call summary from `sessionStorage`

  - [ ] Github Copilot
  - [x] Other - Please Specify

    Claude Code

- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


rarajes2 (commented):
One additional inline finding from this pass.

docs/samples/contact-center/app.js
```diff
@@ -2369,6 +3074,8 @@ function registerTaskListeners(task) {
 
     // Clean up task creation time tracking
     taskCreationTimes.delete(task.data.interactionId);
+    summaryFeatureMap.delete(task.data.interactionId);
```
rarajes2, line 3038: The termination cleanup removes the in-memory feature flags but leaves this interaction's persisted post-call summary in `sessionStorage`. If a task ends before `wrapupCall()` (for example, a non-wrapup termination), the generated summary remains available for the rest of the browser session and the storage map grows indefinitely. Call `clearSummaryStorageState(task.data.interactionId)` here alongside the other task-end cleanup.

packages/@webex/contact-center/src/services/config/types.ts
```diff
@@ -141,10 +141,18 @@ export const CC_TASK_EVENTS = {
...
-  MID_CALL_SUMMARY: 'MID_CALL_SUMMARY',
-  /** Event emitted when a post-call summary is available */
+} as const;
+
+/**
+ * AI summary events received through Contact Center realtime delivery.
+ * @enum {string}
+ * @public
+ */
+export const CC_AI_SUMMARY_EVENTS = {
```
rarajes2: [P1] Preserve the legacy `CC_TASK_EVENTS.MID_CALL_SUMMARY` and `.POST_CALL_SUMMARY` members (for example as aliases). Moving them exclusively to `CC_AI_SUMMARY_EVENTS` makes existing SDK consumers receive `undefined` at runtime after upgrade, which is a breaking public-API change.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -85,49 +190,498 @@ export default class TaskManager extends EventEmitter {
...
+    }
+
+    const featurePayload = payload as FeatureEnablementEventPayload;
+    const matchingTask = Object.values(this.taskCollection).find((task) => {
+      const correlation = this.getAISummaryCorrelationForTask(task, 'feature-presence-scan');
+
+      return correlation?.interactionId === featurePayload.interactionId;
+    });
+
+    this.trackFeatureEnablementReceived({
```
rarajes2, line 424: [P2] Keep telemetry non-blocking here. If `trackFeatureEnablementReceived()` throws, this valid frame is never retained or emitted, so the summary APIs later reject as disabled. Store/deliver the flags first, or isolate metrics failures.

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +303,510 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+        'POST_CALL_SUMMARY',
+        'POST_CALL_SUMMARY_TIMEOUT',
+        conversationId,
+        interactionId,
+        AIAssistantEventName.GET_POST_CALL_SUMMARY
+      );
+
+      this.postCallSummaryResponseContext = {conversationId, interactionId};
+
+      this.trackAISummaryOperation(
```
rarajes2: [P2] Isolate this metric from the product operation. A `MetricsManager.trackEvent()` failure after a summary has already resolved sends control to the catch block and rejects the public API; response callers may then retry an event the backend already accepted.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -85,49 +190,498 @@ export default class TaskManager extends EventEmitter {
...
+        'late-or-uncorrelated',
+        CC_AI_SUMMARY_EVENTS.MID_CALL_SUMMARY,
+        {
+          conversationId: payload.conversationId,
+        }
+      );
+    }
+  }
+
+  private handleReceivingAgentSummaryEvent(payload: Record<string, unknown>): void {
```
rarajes2, line 487: [P2] Validate the complete receiving-summary payload before casting and emitting it. With only these two checks, `{conversationId: 'c', summaryText: {invalid: true}}` is delivered as `MidCallSummaryReceivingAgentPayload` even though consumers are promised a string summary.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -1131,11 +1700,15 @@ export default class TaskManager extends EventEmitter {
...
       this.wrapupData,
       this.agentId,
+      this.agentName,
       this.answerCallOnWebexService
     );
 
+    this.configureTaskAISummary(task);
     this.setupTaskListeners(task);
     this.taskCollection[payload.interactionId] = task;
+    this.retainFeatureEnablementForTask(task);
```
rarajes2, line 1826: [P2] This replays retained feature flags before the lifecycle pipeline publishes the reserved task to consumers. A host therefore cannot yet attach `task:featureEnablement`, and the only notification is lost. Replay after the task has been surfaced.

rarajes2 (commented):
Defect-focused review findings.

docs/samples/contact-center/app.js
rarajes2, line 3038: [P2] Also clear `sessionStorage` here. A task that terminates outside `wrapupCall()` leaves its generated post-call summary persisted for the rest of the session, which retains stale contact content and grows the storage map.

docs/samples/contact-center/app.js
```diff
@@ -658,10 +687,483 @@ async function showInitiateConsultDialog() {
...
   initiateConsultDialog.close();
+  if (midCallSummary.actionType !== 'CONSULT') return;
+  const interactionId = currentTask?.data?.interactionId;
+  const features = summaryFeatureMap.get(interactionId) || {};
+  if (!features.midCallEnabled || !interactionId) return;
+  try {
+    if (midCallSummary.payload) {
+      await currentTask.sendMidCallSummaryResponse({
+        summaryReceived: true,
+        summary: buildSummaryPayload('consult-summary'),
```
rarajes2, line 750: [P2] Capture the task that opened this flow instead of using mutable `currentTask` when the dialog closes. If another task is assigned while a summary is pending, this response can be sent with the second task's interaction/conversation identifiers.

rsarika: Fixed in 275300fa88. Consult and transfer flows now capture the task that opened the flow and use it for summary requests, responses, and the operation, instead of relying on mutable currentTask.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: Fixed in 275300fa88. Success telemetry is now isolated from the summary response operation, so a metrics exception cannot reject a response already accepted by the backend.

docs/samples/contact-center/app.js
rsarika: Addressed in 275300fa88 by removing the AI-summary sessionStorage persistence from the sample app entirely. Since no summary or task state is stored now, task:end has no persisted summary to clear.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: Fixed in 275300fa88. Feature-enablement telemetry is now best effort; a metrics exception cannot prevent valid flags from being retained or delivered to the task.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: Fixed in 275300fa88. Retained feature flags are replayed after the public task lifecycle event, with duplicate-delivery protection and reservation re-key handling.

docs/samples/contact-center/app.js
rsarika: Addressed in 275300fa88 by removing the AI-summary sessionStorage persistence from the sample app entirely. There is no longer stale generated summary state to retain after task termination.

packages/@webex/contact-center/src/services/config/types.ts
rsarika: Intentional. These members were placeholders for this feature and had no consumers, so they were removed intentionally. We are not restoring them.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: No code change was needed. The existing isNonEmptyString(payload.summaryText) check rejects objects, arrays, empty strings, and other non-string values before the payload is cast or routed.

packages/@webex/contact-center/src/services/task/constants.ts
```diff
@@ -30,6 +30,8 @@ export const CAMPAIGN_PREVIEW_SKIP = '/skip';
 export const CAMPAIGN_PREVIEW_REMOVE = '/remove';
 /** 80-second timeout for accepting preview contact (outbound call setup takes longer than default 20s) */
 export const TIMEOUT_PREVIEW_ACCEPT = 80000;
```
rarajes2, line 32: [P2] Align this timeout with the feature contract. The PR describes and tests a 30-second request/early-frame retention window, but all coordinator request, receiver-buffer, and feature-flag timers use this 15-second value. A valid summary or task arriving between 15 and 30 seconds is therefore discarded or rejected a full 15 seconds earlier than the documented behavior.

rarajes2 (commented):
Re-review finding on the current head.

docs/samples/calling.min.js
Kesari3008, line 1: Why do we have deletion of these files: .min.js appearing in the PR ?

packages/@webex/contact-center/src/metrics/constants.ts
```diff
@@ -186,6 +186,18 @@ export const METRIC_EVENT_NAMES = {
     'AI Assistant Fetch Historic Transcripts Success',
   AI_ASSISTANT_FETCH_HISTORIC_TRANSCRIPTS_FAILED: 'AI Assistant Fetch Historic Transcripts Failed',
 
+  // AI summary events
+  AI_SUMMARY_GET_POST_CALL_SUCCESS: 'AI Summary Get Post Call Success',
```
Kesari3008: nitpick: More appropriate name would be calling PostCall Summary imstead of AI Summary, same applicable for mid call summary

packages/@webex/contact-center/src/services/config/types.ts
Kesari3008: Why do we need to add them as separate constant objects, why can't they be part of TASK_EVENTS and if we are planning to keep them separate then all AI Assistant feature events should be part of same object

packages/@webex/contact-center/src/services/task/constants.ts
```diff
@@ -30,6 +30,8 @@ export const CAMPAIGN_PREVIEW_SKIP = '/skip';
...
+export const AI_SUMMARY_DURATION_MS = 15_000;
```
Kesari3008: There is a typo here I believe, remove the underscore

packages/@webex/contact-center/src/services/task/AISummaryCoordinator.ts
```diff
@@ -0,0 +1,332 @@
...
+  AISummaryRequestCoordinator,
+  AISummaryTimeoutCodeByInboundType,
+  FeatureEnablementEventPayload,
+  ITask,
+  MidCallSummaryReceivingAgentPayload,
+  TASK_EVENTS,
+} from './types';
+import {createAISummaryError} from '../AISummaryUtils';
+
+type PendingAISummaryRequest<T extends AISummaryInboundType> = {
```
Kesari3008: These types should exist in types file not here. Also why do we have a need to create a separate class for handling summary

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -46,11 +56,81 @@ import {
...
+import {AIAssistantEventName} from '../../types';
+import type {AISummaryResponseTransportPayload} from '../../types';
+import {getAISummaryCorrelation} from './TaskUtils';
+import {createAISummaryError, isFiniteNonNegativeNumber, isNonEmptyString} from '../AISummaryUtils';
 
 type UIControlConfigInput = Omit<UIControlConfig, 'channelType'> & {
   channelType?: UIControlConfig['channelType'];
 };
 
+type AISummaryAdapter = Pick<ApiAIAssistant, 'sendSummaryGetEvent' | 'sendSummaryResponseEvent'>;
```
Kesari3008: Same comment here

packages/@webex/contact-center/src/cc.ts
```diff
@@ -693,34 +773,18 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
-        this.services.rtdWebSocketManager.close(false, 'Unregistering the RTD websocket');
-      }
-
       // Clear any cached agent configuration
       this.agentConfig = null;
 
       if (wxAppPublishError) {
         throw wxAppPublishError;
       }
-
```
Kesari3008, line 710: Why is this change introduced ? What was the need for moving it to catch block and also adding few more failures conditions

packages/@webex/contact-center/src/cc.ts
```diff
@@ -693,34 +773,18 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
       if (!this.services.webSocketManager.isSocketClosed) {
         this.services.webSocketManager.close(false, 'Unregistering the SDK');
       }
-      if (this.services.rtdWebSocketManager && !this.services.rtdWebSocketManager.isSocketClosed) {
```
Kesari3008, line 696: Why are we removing both ? Looks like they were duplicates but keeping one would still be needed

packages/@webex/contact-center/src/cc.ts
```diff
@@ -650,23 +734,19 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
-      this.taskManager.off(TASK_EVENTS.TASK_MULTI_LOGIN_HYDRATE, this.handleTaskMultiLoginHydrate);
-      this.taskManager.off(
-        TASK_EVENTS.TASK_CAMPAIGN_PREVIEW_RESERVATION,
-        this.handleCampaignPreviewReservation
-      );
+      this.unregisterTaskManagerEventForwarders();
       this.taskManager.unregisterIncomingCallEvent();
 
       this.services.webSocketManager.off('message', this.handleWebsocketMessage);
-      this.services.rtdWebSocketManager.off('message', this.handleRTDWebsocketMessage);
```
Kesari3008, line 669: Why is this removed ?

packages/@webex/contact-center/src/cc.ts
```diff
@@ -515,12 +517,92 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    };
+  }
+
   /**
    * Sets up event listeners for incoming tasks and task hydration
    * Subscribes to task events from the task manager
    * @private
    */
-  private incomingTaskListener() {
+  private unregisterTaskManagerEventForwarders(): void {
```
Kesari3008, line 604: Why do need new method for this ? Do we have multiple invocations for this method ?

packages/@webex/contact-center/src/cc.ts
```diff
@@ -81,6 +81,8 @@ import type {
   ContactServiceQueueSearchParams,
 } from './types';
 
+type CapturedFailure = {captured: false; error?: never} | {captured: true; error: unknown};
```
Kesari3008: Belongs in type file and what is captured failure

packages/@webex/contact-center/src/cc.ts
```diff
@@ -515,12 +517,92 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
...
+    this.taskManager.off(TASK_EVENTS.TASK_HYDRATE, this.handleTaskHydrate);
+    this.taskManager.off(TASK_EVENTS.TASK_MULTI_LOGIN_HYDRATE, this.handleTaskMultiLoginHydrate);
+    this.taskManager.off(TASK_EVENTS.TASK_MERGED, this.handleTaskMerged);
+    this.taskManager.off(
+      TASK_EVENTS.TASK_CAMPAIGN_PREVIEW_RESERVATION,
+      this.handleCampaignPreviewReservation
+    );
+  }
+
+  private refreshTaskManagerEventForwarders(): void {
```
Kesari3008, line 620: If we are already switching off these listerners as part of cleanup during dergeister why is there a need of refresh

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -2,25 +2,97 @@ import {v4 as uuidv4} from 'uuid';
...
 import {
   AI_ASSISTANT_BASE_URL_TEMPLATE,
   AI_ASSISTANT_ENV_MAP,
   AI_ASSISTANT_API_URLS,
   WCC_API_GATEWAY,
 } from './constants';
 import {AIFeatureFlags} from './config/types';
+import {isFiniteNonNegativeNumber, isNonEmptyString} from './AISummaryUtils';
+
+export const AI_SUMMARY_HTTP_TIMEOUT_MS = TIMEOUT_REQ;
```
Kesari3008: Move these to types and constants file

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -40,16 +112,11 @@ export class ApiAIAssistant {
...
     const wccApiGatewayUrl = this.webex.internal.services.get(WCC_API_GATEWAY) || '';
 
     if (!wccApiGatewayUrl) {
-      const {error: detailedError} = getErrorDetails(
-        new Error('AI_ASSISTANT_BASE_URL_NOT_AVAILABLE'),
-        METHODS.GET_BASE_URL,
-        CC_FILE
-      );
-      throw detailedError;
+      return undefined;
```
Kesari3008, line 68: Why are we returning undefined and why was this change needed ?

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -69,7 +143,265 @@ export class ApiAIAssistant {
...
+    ) {
+      throw this.createSummaryError(
+        AI_SUMMARY_TRANSPORT_ERROR_CODES.VALIDATION_FAILED,
+        METHODS.SEND_SUMMARY_GET_EVENT
+      );
+    }
+
+    const now = Date.now();
+    const baseUrl = this.getSummaryBaseUrl();
+    const body = this.buildSummaryEventEnvelope({
```
Kesari3008: Why do we need 3 different methods to send HTTP request and handle the response? Is it really worth splitting the business logic like how its done ? Are we ever going to use the private methods created to build the body and send the HTTP POST request for any other feature because only then it makes sense to have this sort of implementation else one public method which builds the payload based on params received and sending the request should be enough. I also don't see the use of response where we have created dependency on the promise resolve and reject

packages/@webex/contact-center/src/services/AISummaryUtils.ts
```diff
@@ -0,0 +1,15 @@
+export type AISummaryError = Error & {data?: Record<string, unknown>};
+
+export const isNonEmptyString = (value: unknown): value is string =>
+  typeof value === 'string' && value.length > 0;
+
+export const isFiniteNonNegativeNumber = (value: unknown): value is number =>
+  typeof value === 'number' && Number.isFinite(value) && value >= 0;
+
+export const createAISummaryError = (errorCode: string): AISummaryError => {
```
Kesari3008: We have created createSummaryError and then this method also. Having one method created should be enough and it should be placed here

packages/@webex/contact-center/src/services/task/AISummaryCoordinator.ts
```diff
@@ -0,0 +1,332 @@
...
+  ): 'resolved' | 'not-found' {
+    const entries = this.getPendingMap(eventType);
+    const removed = this.removeTimedEntry(entries, conversationId, (entry) => {
+      entry.resolve(payload);
+    });
+
+    return removed ? 'resolved' : 'not-found';
+  }
+
+  public routeReceivingSummary(
```
Kesari3008: Where are we using the methods created in this file and why do we need so many methods ? AISummaryUtils.ts can be combined with this file and any utility method which helps in simplifying some business logic or helper methods which would be commons used in multiple places can be placed in it. Please update the structure a bit

packages/@webex/contact-center/src/services/core/Utils.ts
```diff
@@ -16,6 +16,18 @@ import {
 import {PARTICIPANT_TYPES, STATE_CONSULT} from './constants';
 import {DialPlan} from '../config/types';
 
+type GetErrorDetailsOptions = {
```
Kesari3008: Move this a type file and why do we need to create a new method for upload logs

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +303,514 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+          ...fields,
+        } as never,
+        ['operational']
+      );
+    } catch {
+      // Metrics are best effort and must not affect the accepted response.
+    }
+  }
+
+  private async requestPostCallSummaryInternal(
```
Kesari3008: I would like to understand why have we created so many layers within Task.ts as well as APIAIAssistant.ts to just invoke one API.
requestPostCallSummary invokes requestPostCallSummaryInternal which invokes requestAISummary which invokes sendUSmmaryGetEvent and there are other methods also involved which is helping to validate and build and the final request and handle. I believe this design can be simplified to a great extent and there is no need for so much layering. Also what would be the public facing API here. ideally we have created APIAiAssistant as public facing object for all AI assistant feature so the main public API exposed for this feature should be present in that file and not Task.ts. These features are tracked and maintained per task object basis so this sort of relation we have to create for these summaries as well

Kesari3008 (changes requested):
Please schedule a sit down review if possible for this. I think we need discussion for the design itself

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1629,6 +1635,189 @@ export type ParticipantBooleanKey =
  */
 export type TaskResponse = AgentContact | Error | void;
 
+export type AISummaryActionType = 'CONSULT' | 'TRANSFER';
```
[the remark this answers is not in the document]
rsarika: Addressed in commit e30f20b783. Added @public documentation to the exported AI-summary types in packages/@webex/contact-center/src/services/task/types.ts.

packages/@webex/contact-center/src/services/core/RtdRequestResolver.ts
```diff
@@ -0,0 +1,149 @@
+import type {RtdRequestOptions, RtdRequestRegistration} from './types';
+
+/**
+ * Coordinates an outbound HTTP request with the RTD event that completes it.
+ *
+ * The HTTP response is only an acknowledgement. The caller receives the result
+ * when the matching RTD event is resolved by the RTD WebSocket owner.
+ */
+type PendingRtdRequest<T> = {
```
rarajes2, line 9: Let's discuss this file. I am not able to understand it's use-case

packages/@webex/contact-center/src/services/core/WebexRequest.ts
```diff
@@ -15,10 +15,23 @@ import {
...
+
+        return existingInstance;
+      }
+
+      const instance = new WebexRequest(options);
+      WebexRequest.instances.set(options.webex, instance);
+      WebexRequest.instance = instance;
+
+      return instance;
     }
```
rarajes2, line 35: Why are we allowing multiple instance creation ?

packages/@webex/contact-center/src/services/core/WebexRequest.ts
```diff
@@ -30,20 +43,26 @@ class WebexRequest {
...
   }): Promise<IHttpResponse> {
-    const {service, resource, method, body, headers} = options;
+    const {service, resource, uri, method, body, headers, addAuthHeader, timeout} = options;
 
     return this.webex.request({
-      service,
-      resource,
+      ...(service !== undefined ? {service} : {}),
+      ...(resource !== undefined ? {resource} : {}),
+      ...(uri !== undefined ? {uri} : {}),
```
rarajes2, line 60: type of uri is string, I don't think we should use empty object as in case it's undefined. Please cross check other fields as well

packages/@webex/contact-center/src/services/core/Utils.ts
```diff
@@ -16,6 +16,18 @@ import {
...
+/**
+ * Uploads diagnostic logs without allowing an upload failure to affect the original operation.
+ * This is shared by the error helpers because both paths intentionally use best-effort logging.
+ */
+const uploadLogsFireAndForget = (correlationId?: string) => {
+  const uploadLogsPromise = WebexRequest.getInstance().uploadLogs({
+    correlationId,
+  });
+
+  uploadLogsPromise?.catch?.(() => undefined);
```
rarajes2, line 28: Should we at least log this ?

packages/@webex/contact-center/src/services/core/Utils.ts
```diff
@@ -160,9 +178,9 @@ export const getErrorDetails = (error: any, methodName: string, moduleName: stri
       trackingId: failure?.trackingId,
     });
     // we can add more conditions here if not needed for specific cases eg: silentReLogin
-    WebexRequest.getInstance().uploadLogs({
-      correlationId: failure?.trackingId,
-    });
+    if (shouldUploadLogs) {
+      uploadLogsFireAndForget(failure?.trackingId);
```
rarajes2, line 182: Instead of creating a new method, why don't we update the uploadLog method itself to handle the error silently ?

packages/@webex/contact-center/src/services/core/WebexRequest.ts
```diff
@@ -15,10 +15,23 @@ import {
 class WebexRequest {
   private webex: WebexSDK;
   private static instance: WebexRequest;
+  private static instances = new WeakMap<WebexSDK, WebexRequest>();
```
rarajes2, line 18: `key` of the map is the whole WebexSDK ?

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -65,7 +65,8 @@ export default class Voice extends Task implements IVoice {
     data: TaskData,
     callOptions?: VoiceUIControlOptions,
     wrapupData?: WrapupData,
-    agentId?: string
+    agentId?: string,
+    agentName?: string
```
rarajes2, line 69: Q. I see we have added agentName everywhere, why is it required ?

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +257,462 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+
+  public requestPostCallSummary(): Promise<PostCallSummaryEventPayload> {
+    const operation = this.requestPostCallSummaryInternal();
+
+    operation.catch(() => undefined);
+
+    return operation;
+  }
+
+  private async requestPostCallSummaryInternal(): Promise<PostCallSummaryEventPayload> {
```
rarajes2, line 280: Please see if we can move out this whole logic to a separate file and import a few methods in this file to keep this file clean and readable.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -44,6 +59,111 @@ const MAIN_INTERACTION_CORRELATED_EVENTS = new Set<CC_EVENTS>([
...
+  | 'late-or-uncorrelated'
+  | 'sdk-deregistered'
+  | 'ambiguous-receiver'
+  | 'receiver-buffer-expired';
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  Boolean(value) && typeof value === 'object' && !Array.isArray(value);
+
+const hasValidOptionalString = (payload: Record<string, unknown>, key: string): boolean =>
+  payload[key] === undefined || typeof payload[key] === 'string';
```
rarajes2, line 87: Same here, see if we can move out these logic to a separate file and import only a few public method in this file. will keep the file readable

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1669,6 +1692,207 @@ export type ParticipantBooleanKey =
...
+  timeoutId?: ReturnType<typeof setTimeout>;
+};
+
+/** @internal AI-summary event names handled by the realtime-delivery path. */
+export type AISummaryRealtimeEventType =
+  | typeof import('../config/types').CC_TASK_EVENTS.POST_CALL_SUMMARY
+  | typeof import('../config/types').CC_TASK_EVENTS.MID_CALL_SUMMARY
+  | typeof import('../config/types').CC_TASK_EVENTS.FEATURE_ENABLEMENT
+  | typeof import('../config/types').CC_TASK_EVENTS.MID_CALL_SUMMARY_RESPONSE_SUBSEQUENT_AGENT;
+
```
rarajes2, line 1895: See of this does not create a cyclic dependency

packages/@webex/contact-center/src/services/AISummaryUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+        ...(context?.statusCode !== undefined ? {statusCode: context.statusCode} : {}),
+        details: {
+          data: {
+            reason: errorCode,
+            methodName,
+            ...(context?.eventName ? {eventName: context.eventName} : {}),
+            ...(context?.agentId ? {agentId: context.agentId} : {}),
+            ...(context?.orgId ? {orgId: context.orgId} : {}),
+            ...(context?.interactionId ? {interactionId: context.interactionId} : {}),
+            ...(context?.conversationId ? {conversationId: context.conversationId} : {}),
```
rarajes2, line 33: looks messy, see if you can destructure context

packages/@webex/contact-center/src/services/AISummaryUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+import type {AISummaryFailureContext} from '../types';
+import {getErrorDetails} from './core/Utils';
+
+export type AISummaryError = Error & {data?: Record<string, unknown>};
+
+export const isNonEmptyString = (value: unknown): value is string =>
+  typeof value === 'string' && value.length > 0;
+
+export const isFiniteNonNegativeNumber = (value: unknown): value is number =>
+  typeof value === 'number' && Number.isFinite(value) && value >= 0;
```
rarajes2, line 11: Do we have lodash or any other 3rd party lib for these kind of logic ? I think this might become repetitive

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -69,7 +92,227 @@ export class ApiAIAssistant {
...
+    };
+
+    if (input.kind === 'response') {
+      data.action = eventName;
+      data.summary = input.payload.summary;
+      data.numberOfTimesViewed = input.payload.numberOfTimesViewed;
+      data.numberOfTimesEdited = input.payload.numberOfTimesEdited;
+      data.numberOfTimesCopied = input.payload.numberOfTimesCopied;
+      data.feedback = input.payload.feedback;
+      data.state = input.payload.state;
```
rarajes2, line 137: see if you can use spread here, might be done in 1 line

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -69,7 +92,227 @@ export class ApiAIAssistant {
...
+
+    try {
+      const requestPromise = this.webexRequest.request({
+        uri: `${baseUrl}${AI_ASSISTANT_API_URLS.EVENT}`,
+        method: HTTP_METHODS.POST,
+        addAuthHeader: true,
+        body,
+        timeout: AI_SUMMARY_HTTP_TIMEOUT_MS,
+      });
+      requestPromise.catch(() => undefined);
```
rarajes2, line 174: should we add logs for all the silent failure ?

rarajes2 (changes requested):

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: Reviewed this. The summary logic stays in Task because it maps task state to the existing request and event APIs. Moving it into another file would add another layer without changing behavior, so no change is planned.

packages/@webex/contact-center/src/services/core/Utils.ts
rsarika: We are keeping this as a small fire-and-forget helper. It makes the call site explicit and ensures diagnostic upload failures do not alter the original error flow. Changing uploadLogs itself could affect its other callers. No change is needed.

packages/@webex/contact-center/src/services/core/Utils.ts
rsarika: This is intentionally silent. The primary error is already logged by getErrorDetails; an upload-log failure is only a secondary best-effort diagnostic failure, and logging it would add noise around the original error. No functional change is needed.

packages/@webex/contact-center/src/services/task/types.ts
rsarika: Checked this. The new references are type-level and do not introduce a runtime import cycle. The contact-center source build has also passed with this arrangement, so no change is needed.

packages/@webex/contact-center/src/services/core/WebexRequest.ts
rsarika: Multiple instances are intentional. WebexRequest is cached in a WeakMap keyed by the WebexSDK instance, so separate SDK instances get isolated request wrappers. The static fallback preserves existing singleton usage. No code change is needed.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
rsarika: agentName is needed in the outbound mid-call summary response event so the backend can identify the agent associated with the response. The constructor field remains optional for backward compatibility, and Task includes it when available. No code change is needed.

packages/@webex/contact-center/src/services/core/WebexRequest.ts
rsarika: The WeakMap key is the WebexSDK object identity, not serialized SDK data. This gives each SDK instance its own request wrapper and avoids retaining instances unnecessarily. No issue was found.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: Addressed in the latest local changes: the optional response fields now use object spread, removing the nested assignment branches while preserving the same payload behavior.

packages/@webex/contact-center/src/services/core/WebexRequest.ts
rsarika: Already addressed. Optional request fields use !== undefined, so an empty string is preserved rather than replaced with an empty object. No additional change is needed.

packages/@webex/contact-center/src/services/AISummaryUtils.ts
rsarika: Addressed in the latest local changes: the summary failure context is destructured once and the fields are reused, which keeps the error payload construction easier to read.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: Reviewed this. TaskManager owns the task lookup and RTD correlation, so keeping that logic together makes the flow easier to follow. Moving it into another file would add a layer without changing behavior, so no change is planned.

packages/@webex/contact-center/src/services/AISummaryUtils.ts
rsarika: Kept the native implementations. These are small, local checks and Set membership, so adding lodash would add unnecessary dependency and indirection for this feature.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: This catch is intentional. It prevents a delayed unhandled rejection when the timeout race settles first; the actual request error is normalized at the operation boundary. Logging here would duplicate or obscure the primary error, so no change is needed.

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +257,428 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+  public async requestPostCallSummary(): Promise<PostCallSummaryEventPayload> {
+    const metricFields: Record<string, unknown> = {
+      operation: METHODS.REQUEST_POST_CALL_SUMMARY,
+    };
+
+    try {
+      this.metricsManager.timeEvent([
+        METRIC_EVENT_NAMES.AI_SUMMARY_GET_POST_CALL_SUCCESS,
+        METRIC_EVENT_NAMES.AI_SUMMARY_GET_POST_CALL_FAILED,
+      ]);
```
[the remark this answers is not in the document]
rsarika: Not addressed. This PR intentionally follows the existing MetricsManager.timeEvent()/trackEvent() pattern used throughout the SDK. Making timing invocation-local would require a broader MetricsManager API change and is outside this PR's minimal scope.

packages/@webex/contact-center/src/services/core/WebexRequest.ts
```diff
@@ -15,10 +15,23 @@ import {
...
+
+      if (existingInstance) {
+        WebexRequest.instance = existingInstance;
+
+        return existingInstance;
+      }
+
+      const instance = new WebexRequest(options);
+      WebexRequest.instances.set(options.webex, instance);
+      WebexRequest.instance = instance;
```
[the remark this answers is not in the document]
rsarika: Addressed in commit 482def9. getErrorDetails now accepts the client-scoped WebexRequest, and ApiAiAssistant passes its own instance, so diagnostic uploads stay with the originating Webex client. Legacy no-context callers retain the existing fallback.

packages/@webex/contact-center/src/types.ts
```diff
@@ -1005,6 +1011,64 @@ export const AIAssistantEventName = {
...
+
+export type AISummaryResponseTransportPayload =
+  | (Contact.PostCallSummaryResponsePayload &
+      AISummaryResponseTransportIdentifiers & {
+        eventName: typeof AIAssistantEventName.POST_CALL_SUMMARY_RESPONSE;
+      })
+  | AISummaryMidCallResponseTransportPayload;
+
+export type AISummaryEnvelopeInput =
+  | {
```
Kesari3008: Is pipe symbol needed here at the start ?

packages/@webex/contact-center/src/types.ts
```diff
@@ -1005,6 +1011,64 @@ export const AIAssistantEventName = {
  */
 export type AIAssistantEventName = Enum<typeof AIAssistantEventName>;
 
+export type AISummaryGetEventName =
```
Kesari3008: Why do we need separate enum for Summary event names ? Also just calling them AIAssistantEvents is enough. We don't have to put name in the name itself

packages/@webex/contact-center/src/cc.ts
```diff
@@ -81,6 +81,8 @@ import type {
   ContactServiceQueueSearchParams,
 } from './types';
 
+type DeregisterFailure = {captured: false; error?: never} | {captured: true; error: unknown};
```
Kesari3008: This doesn't belong in this file

Kesari3008: I believe we should revert all the changes done in this file that has nothing to do with Summary feature and if they are supposed to be improvements, then should be address separately so we can review and discuss those change properly. It's too chaotic and difficult to address them in the same PR where a feature is being implemented

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
Kesari3008: This comment is not addressed

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+        {webexRequest: this.webexRequest}
       );
       throw detailedError;
     }
 
-    return AI_ASSISTANT_BASE_URL_TEMPLATE.replace('%s', resolvedEnv);
+    return baseUrl;
+  }
+
+  private getSummaryBaseUrl(): string {
```
Kesari3008: Why do we need separate method for fetch url for summary api, seems to be invoking the same method: resolveBaseUrl only

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+        clearTimeout(timeoutId);
+      }
+    }
+  }
+
+  public async sendSummaryGetEvent(
+    agentId: string,
+    interactionId: string,
+    conversationId: string,
+    eventName: AISummaryGetEventName
```
Kesari3008: We already have a type existing for this, why aren't we using that:  AIAssistantEventName

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+        {...context, ...(statusCode !== undefined ? {statusCode} : {})}
+      );
+    } finally {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
+    }
+  }
+
+  public async sendSummaryGetEvent(
```
Kesari3008: This method and below method, both are doing exactly same what we have in sendEvent pre existing method. So why do we need 2 new methods. Only thing that needs to be adjusted is payload which can be done easily in the same method

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+        METHODS.GET_BASE_URL
+      );
+    }
+  }
+
+  private async sendSummaryEvent(
+    input: AISummaryEnvelopeInput,
+    context: AISummaryFailureContext
+  ): Promise<void> {
+    const eventName = input.kind === 'get' ? input.eventName : input.payload.eventName;
```
Kesari3008: eventName is already being received, why we need condition for input.kind...not needed

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+    const interactionId = input.kind === 'get' ? input.interactionId : input.payload.interactionId;
```
Kesari3008: This also shouldn't change, we should just receive interaction as it is irrespective of whether it is getpostcallsummary or getpostcallsummaryresponse

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -60,16 +76,247 @@ export class ApiAIAssistant {
...
+    };
+    const baseUrl = this.getSummaryBaseUrl();
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+    const timeoutMarker = {code: AI_SUMMARY_TRANSPORT_ERROR_CODES.TIMEOUT};
+
+    const timeoutPromise = new Promise<never>((_resolve, reject) => {
+      timeoutId = setTimeout(() => reject(timeoutMarker), AI_SUMMARY_HTTP_TIMEOUT_MS);
+    });
+
+    try {
```
Kesari3008: As mentioned sendEvent already does it. For the same API we shouldn't create 2 different methods.
This is how the design should be:

One single public API available to be invoked from widgets for each request(4 request 2 for post call summary and its response and mid call summary and its response). Widgets will send us data like which event and based on that we form the payload and invoke sendEvent and pass the relevant details

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -716,6 +718,27 @@ export enum TASK_EVENTS {
    * ```
    */
   TASK_CAMPAIGN_CONTACT_UPDATED = 'task:campaignContactUpdated',
+
+  /**
+   * Triggered when a mid-call summary is available for the receiving agent.
+   */
+  TASK_MID_CALL_SUMMARY_FOR_RECEIVING_AGENT = 'task:midCallSummaryForReceivingAgent',
```
Kesari3008: We dont need to add for receiving agent in the event name

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1669,6 +1692,207 @@ export type ParticipantBooleanKey =
  */
 export type TaskResponse = AgentContact | Error | void;
 
+/** AI-assisted mid-call summary action. @public */
+export type AISummaryActionType = 'CONSULT' | 'TRANSFER';
```
Kesari3008, line 1695: I believe some of these types and payload can be simplified and redundancy can be avoided, please check

Kesari3008 (changes requested):
We will need a call to discuss TaskManager.ts, Task.ts  changes and how RTDRequestresolver would be getting used and has a usecase

rarajes2 (commented):
Blocking finding remains: the sample keeps post-call summary state and the pending request in global variables. If two tasks enter wrap-up concurrently, the later task resets and replaces that shared state; the earlier request can then populate the wrong task’s UI or cause the later task to send an incorrect `IGNORED`/cross-interaction response. Scope the post-call summary state and pending request to the owning task before merging.

rarajes2 (approved):
Approved: I found no remaining blocking findings on the current head. Existing non-blocking follow-ups (including duplicate RTD-request settlement and invocation-local summary metrics) should be addressed in a follow-up before or alongside merge as appropriate.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -26,6 +27,14 @@ import {
 } from './constants';
 import {AIFeatureFlags} from './config/types';
 
+type PendingRtdRequest<T> = {
```
Kesari3008: Type file shouldn't exist here

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
```diff
@@ -257,23 +373,27 @@ export class ApiAIAssistant {
     try {
       const baseUrl = this.getBaseUrl();
       const orgId = this.webex.credentials.getOrgId();
+      const {actionTimeStamp, ...eventData} = eventMetaData ?? {};
```
Kesari3008: Why are these changes needed? Language code is something we expect to receive from the method parameters, same goes for trackingID. And action timestamp we are supposed to have current time so calculating it using String(Date.now() should be applicable for all types of AI Assistant APIs

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -716,6 +717,27 @@ export enum TASK_EVENTS {
    * ```
    */
   TASK_CAMPAIGN_CONTACT_UPDATED = 'task:campaignContactUpdated',
+
+  /**
+   * Triggered when a mid-call summary is received for the task.
+   */
+  TASK_MID_CALL_SUMMARY_RECEIVED = 'task:midCallSummaryReceived',
```
Kesari3008: I don't see this event being used anywhere, please check and remove if it's unused

packages/@webex/contact-center/src/constants.ts
```diff
@@ -82,3 +82,13 @@ export const METHODS = {
   GET_CALL_DETAILS_ON_WEBEX: 'getCallDetailsOnWebex',
   ENSURE_WXAPP_MERCURY_CONNECTED: 'ensureWxAppMercuryConnected',
 };
+
+export const AI_SUMMARY_ERROR_CODES = {
```
Kesari3008: I don't see these error code being used anywhere

docs/samples/contact-center/style.css
```diff
@@ -1077,3 +1077,35 @@ legend {
...
+  outline: 2px solid currentColor;
+  background-color: #e0f0ff;
+  font-weight: bold;
+}
+
+/* AI summary per-section editing */
+.summary-sections-container label {
+  display: block;
+  font-weight: bold;
+  margin-top: 8px;
```
Kesari3008: There is a common practice we follow to use rem instead of px. Not sure how strict it is for SDK but please cross-check

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -1669,6 +1691,179 @@ export type ParticipantBooleanKey =
  */
 export type TaskResponse = AgentContact | Error | void;
 
+/** AI-assisted mid-call summary action. @public */
+export type AISummaryActionType = 'CONSULT' | 'TRANSFER';
```
Kesari3008: As discussed, please cleanup the types

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -45,12 +59,17 @@ import {
...
+import {AIFeatureFlags, WrapupData} from '../config/types';
+import {AIAssistantEventName, AIAssistantEventType} from '../../types';
+import {createSummaryError} from './TaskUtils';
+import type ApiAIAssistant from '../ApiAiAssistant';
 
 type UIControlConfigInput = Omit<UIControlConfig, 'channelType'> & {
   channelType?: UIControlConfig['channelType'];
 };
 
+const AI_SUMMARY_INVALID_RESPONSE_PAYLOAD = 'AI_SUMMARY_INVALID_RESPONSE_PAYLOAD';
```
Kesari3008: This should be in constants file

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +248,266 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+        eventName,
+        eventMetaData: {
+          conversationId,
+          clientType: AI_ASSISTANT_CLIENT_TYPE,
+        },
+        timeout: AI_SUMMARY_DURATION_MS,
+      });
+
+      if (isPostCall) {
+        this.postCallSummaryResponseContext = {conversationId, interactionId};
```
Kesari3008: Both objects are recording same values so does it matter if it's for midcall or postcall. This condition feels unnecessary

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -216,6 +248,266 @@ export default abstract class Task extends EventEmitter implements ITask {
...
+
+  public clearFeatureEnablement(): void {
+    this.aiSummaryCapabilities = {
+      midCallEnabled: false,
+      postCallEnabled: false,
+    };
+    this.pendingFeatureEnablement = undefined;
+  }
+
+  public requestPostCallSummary(): Promise<PostCallSummaryEventPayload> {
```
Kesari3008: If we are having dedicated methods for each of the API invocation then we can just directly call sendEvent from here instead of requestSummary and sendSummaryResponseMethod

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -109,6 +115,41 @@ export default class TaskManager extends EventEmitter {
...
+
+          this.handleFeatureEnablementEvent(payload?.data?.data);
+
+          return;
+        }
+        case CC_TASK_EVENTS.POST_CALL_SUMMARY:
+        case CC_TASK_EVENTS.MID_CALL_SUMMARY: {
+          if (!this.aiSummaryInboundActive || !payload?.data?.data?.conversationId) return;
+
+          this.apiAIAssistant?.resolveFromRtdEvent(
```
Kesari3008: I am assuming we have these methods: resolveFromRtdEvent and selectReceivingSummaryTasks due to the promise logic. If we simply emit  the summary event as we receive, wont that simplify the logic.

packages/@webex/contact-center/src/services/task/TaskManager.ts
```diff
@@ -138,11 +181,173 @@ export default class TaskManager extends EventEmitter {
     }
   }
 
+  private removeTimedAISummaryEntry<T extends {timeoutId?: ReturnType<typeof setTimeout>}>(
```
Kesari3008: Please check the usecase for these methods

packages/@webex/contact-center/src/cc.ts
```diff
@@ -652,6 +651,14 @@ export default class ContactCenter extends WebexPlugin implements IContactCenter
    * @private
    */
   private incomingTaskListener() {
+    this.taskManager.off(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
```
Kesari3008: As discussed, cleanup these changes as they are not realted to the summary feature

Kesari3008 (changes requested):
Some of the .md files seem to have changes more than required. Ideally they should only get updated for additional changes done but it also has modified existing details

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: Partially addressed in the pending update: PendingRtdRequest has been moved from ApiAiAssistant.ts to services/core/types.ts. I kept the Promise-based correlation because requestPostCallSummary and requestMidCallSummary publicly return Promise<AISummary>; the Promise settles only after the event request and matching RTD result complete. Replacing this with event-only delivery would change that public contract, so I am leaving that design decision open for confirmation.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: languageCode and trackingId still come from the sendEvent method parameters; the current code only omits those properties when they are undefined. The remaining question is actionTimeStamp: summary responses currently accept actionTimeStamp and publishTimestamp from the caller and use the current time only as a fallback. Can you confirm whether the backend requires sendEvent to always generate String(Date.now()) for summary response events as well? I am keeping this open pending that wire-contract confirmation.

packages/@webex/contact-center/src/services/task/types.ts
rsarika: Checked the complete flow. TASK_MID_CALL_SUMMARY_RECEIVED is emitted by TaskManager when a subsequent-agent summary is routed to the selected task, consumed by the contact-center sample through task:midCallSummaryReceived, and covered by TaskManager, cc, and root-contract tests. It is the receiving-agent delivery event, so it is used and should remain.

packages/@webex/contact-center/src/constants.ts
rsarika: Checked all entries. AI_ASSISTANT_BASE_URL_NOT_AVAILABLE is used by ApiAiAssistant.getBaseUrl; the post-call and mid-call disabled and timeout codes are used by Task request handling; and the exported object is pinned by the root public-contract test. These codes are used and should remain.

docs/samples/contact-center/style.css
rsarika: Cross-checked the sample stylesheet. It currently uses both px and rem extensively, and there is no rem-only rule in this package. The new 2px outline and small pixel margins follow nearby sample styles, so I have retained them. We can normalize the complete sample stylesheet separately if the team wants a strict rem convention.

packages/@webex/contact-center/src/services/task/types.ts
rsarika: Addressed in commit 9dd3d1360c. The flow-specific duplicate summary payload and response types were consolidated into the shared AISummary, AISummaryResponse, AISummaryAction, AISummaryFeatureEnablement, AISummarySections, AISummaryState, and AISummaryFeedback types. This diff thread is now outdated.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: Addressed in the pending update. AI_SUMMARY_INVALID_RESPONSE_PAYLOAD has been moved to src/constants.ts as a standalone internal constant. It was intentionally not added to the exported AI_SUMMARY_ERROR_CODES object because that object has a pinned public shape.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: The two objects have the same shape but retain different request lifecycles. A post-call request and a mid-call request can capture different interaction or conversation context before their responses are sent; one shared slot could overwrite the other. A map keyed by summary type would be equivalent but not simpler. I have kept the separate contexts unless the API contract guarantees these flows can never overlap.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: The four public methods are intentionally thin, typed entry points. requestSummary centralizes feature gating, event selection, RTD correlation, timeout handling, and metrics; sendSummaryResponse centralizes retained context, response serialization, timeout, and metrics. Calling sendEvent directly from each public method would duplicate this behavior, and request methods still need to await the matching RTD result. No change is planned here.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: resolveFromRtdEvent is required by the current Promise-returning request APIs so the matching RTD result can settle the caller request. selectReceivingSummaryTasks serves a separate path: it selects the correct task for an unsolicited subsequent-agent summary and is not caused by Promise correlation. An event-only initiating API would be a public contract change. Please confirm if that breaking redesign is intended; otherwise the two paths should remain.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: Checked the use cases. removeTimedAISummaryEntry is used across feature-enablement retention, receiving-summary buffering, task updates/removal, and full cleanup. It atomically clears the timeout, removes the map entry, and returns the entry when the caller needs it, preventing stale timers and duplicated cleanup logic. It should remain.

packages/@webex/contact-center/src/cc.ts
rsarika: Addressed in commit 180defe061. The unrelated cc.ts lifecycle and reconnection changes were removed, including restoration of the existing RTD close guards and log messages. The remaining cc.ts diff is limited to generated-summary state cleanup, agent-name propagation, and enabling the RTD connection when generated summaries are configured. This thread is now outdated.

rsarika:
Follow-up on the Markdown review: addressed in the pending update. The broad rewrites across eight legacy/reference documents have been removed, along with the stale RTD ownership edits. The final documentation diff against the PR base is limited to the canonical metrics-spec.md and task-spec.md files: 32 additions and 7 replacements covering only AI summary APIs, events, metrics, and catalog accuracy. I am leaving the review open until this pending update is pushed.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: Updated the type placement earlier: PendingRtdRequest and RtdRequestOptions now live in services/core/types.ts. After checking the Contact Center API pattern, I kept the Promise contract. Initiated summary methods return Promise<AISummary> and settle on the matching RTD response, while unsolicited subsequent-agent summaries remain event-driven. This keeps correlation, timeout, and cleanup inside the SDK instead of every consumer.

packages/@webex/contact-center/src/services/ApiAiAssistant.ts
rsarika: Updated in 15f3da706d. languageCode and trackingId still come from the sendEvent parameters, and sendEvent now always sets actionTimeStamp using String(Date.now()), overriding any metadata value. I also removed actionTimeStamp and publishTimestamp from the public AISummaryResponse; the SDK generates the publish timestamp at send time. Agent Desktop follows the same ownership model and does not pass these timestamps from UI callers.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: Updated in 15f3da706d. I replaced the two fields and conditional with one summaryResponseContexts map keyed by summary type. Independent entries are still needed because Agent Desktop supports overlapping post-call consult flows and TaskManager can re-key the same task before responses are sent; one unkeyed context could overwrite the other request correlation. A focused overlap test now covers both retained contexts.

packages/@webex/contact-center/src/services/task/Task.ts
rsarika: I kept the shared private helpers intentionally. The four public methods remain dedicated typed entry points, while the helpers avoid duplicating feature gating, event selection, RTD correlation, retained context, timeout handling, serialization, and metrics. The request methods also cannot directly call only sendEvent because their Promise must settle with the matching RTD summary.

packages/@webex/contact-center/src/services/task/TaskManager.ts
rsarika: Confirmed after tracing both paths: resolveFromRtdEvent supports the Promise-returning initiated request APIs. selectReceivingSummaryTasks is separate from that Promise logic; it routes an unsolicited subsequent-agent summary to the correct task, including task creation and re-key timing races, and that path already emits TASK_MID_CALL_SUMMARY_RECEIVED. So the requester remains Promise-based and the receiving-agent path remains event-based.

PR #5172, opened 2026-08-17
fix(calling): handle 409 error for keepalive

rarajes2:
# COMPLETES #< [SPARK-832708](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-832708) >

When Mobius answers a keepalive (`device_status`) with `409 Conflict`, the device registration
is gone but the same user still holds an active registration elsewhere — typically calling
opened in a second browser tab, which superseded this one. Previously a 409 fell through to the
`default` branch of `handleRegistrationErrors`, so the SDK re-registered. That unregistered the
other tab, which then re-registered and unregistered this one, producing a registration
ping-pong between tabs.

A 409 on keepalive is now a hard stop: the keepalive worker is terminated, no re-registration is
attempted, the Mobius WebSocket is closed, and the app is notified that its session was
superseded.

- `register.ts` — the `KEEPALIVE_FAILURE` branch short-circuits to a new
  `handle409KeepaliveFailure` before `handleRegistrationErrors` runs, so no retry, failover, or
  restore path can execute. The worker is stopped up front (before waiting on the shared mutex)
  so no further keepalive can be sent.
- `performRegistrationDownCleanup` → `performHardStopCleanup(caller, hardStop)`, now shared with
  the 409 path. A `HardStop` discriminated union (`registration/types.ts`) makes it impossible to
  signal a superseded session without the `LineError` the consumer receives, and its
  `HARD_STOP_REASON` values double as the log label. Teardown is unchanged: timers cleared,
  transient flags and failover cache reset, status `INACTIVE`, WebSocket closed with the existing
  permanent code `{code: 3050, reason: 'done (permanent)'}`.
- Observability: submits the existing `KEEPALIVE_ERROR` / `KEEPALIVE_FAILURE` registration metric
  and calls `uploadLogs()`.

### Public API (additive)

- New `LINE_EVENTS.SESSION_SUPERSEDED = 'session_superseded'`, typed in `LineEventTypes` as
  `(error: LineError) => void`.
- New `ERROR_TYPE.SESSION_SUPERSEDED` and `ERROR_CODE.CONFLICT = 409`.
- For backward compatibility `unregistered` still fires first, followed by `session_superseded`
  with the error. `Line.lineEmitter` re-emits it through the same guarded branch as `error`.

No existing event, payload, or status transition changed. Keepalive `404` / `429` / `5xx` and the
registration, restoration, failover, and failback paths are untouched — the short-circuit is
scoped to the keepalive worker's failure branch only.

### Notes

The `LineError` message comes from a new `SESSION_SUPERSEDED_MESSAGE` constant rather than
Mobius's `data.message`: the keepalive worker's failure envelope forwards only headers/status, and
adding the response body would have silently changed the 403 device-limit branch in
`handleRegistrationErrors`. Detection relies solely on the status code.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

Vidcast - https://app.vidcast.io/share/d6bda2af-24a9-4cee-ab69-d9a2da283aa3

**Further Testing**

User registers in different session with old session having active call - with http/mercury and without 409 handling
https://app.vidcast.io/share/7fbc2f34-ec31-459c-bfc2-5a92c3a11383

User registers in different session with old session having active call - with mobius socket and 409 handling
https://app.vidcast.io/share/a8a4ea5c-792d-45f1-b343-59128d55e903

  - [ ] Github Copilot
  - [x] Other - Cursor
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1193,6 +1253,16 @@ export class Registration implements IRegistration {
                 logContext
               );
 
+              if (Number(error.statusCode) === ERROR_CODE.CONFLICT) {
```
Kesari3008: Why are we not adding a separate switchcase in handleRegistrationError itself where we are addressing all the error codes for this error as well. We should move the handling there. Anyway it seems all we need to do is invoke the handler

packages/calling/src/CallingClient/line/index.ts
```diff
@@ -205,6 +205,7 @@ export default class Line extends Eventing<LineEventTypes> implements ILine {
         this.emit(event);
         break;
       case LINE_EVENTS.ERROR:
+      case LINE_EVENTS.SESSION_SUPERSEDED:
```
Kesari3008: Why do we need to define new event for just handling this error ? If 409 is received for keepalive, we are to assume that in this tab registration doesn't exist so sending Unregistered event  back to the application should be enough. We can add reason for deregistration in the payload for Unregistered event but a new event just for keepalive failure is not required

packages/calling/src/CallingClient/line/types.ts
```diff
@@ -18,6 +18,7 @@ export enum LINE_EVENTS {
   RECONNECTING = 'reconnecting',
   REGISTERED = 'registered',
   UNREGISTERED = 'unregistered',
+  SESSION_SUPERSEDED = 'session_superseded',
```
Kesari3008: Not needed

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1387,27 +1457,33 @@ export class Registration implements IRegistration {
...
+   * `LINE_EVENTS.UNREGISTERED` so the SDK consumer is notified. A superseded session
+   * additionally emits `LINE_EVENTS.SESSION_SUPERSEDED` with the reason.
    *
    * Runs under the shared mutex to avoid racing with other registration flows.
    *
    * @param caller - Identifier of the caller, used for logs.
+   * @param hardStop - Why the registration is being torn down as {@link HardStop}.
    */
-  private async performRegistrationDownCleanup(caller: string): Promise<void> {
+  private async performHardStopCleanup(caller: string, hardStop: HardStop): Promise<void> {
```
Kesari3008: If we really wanna rename this method, registrationCleanup is enough. Also the changes with respect to new event are not needed. Noticed that we are already emitting Unregistered event here after cleaning everything up and after sedning unregistered, we send Session_Superseded again which is not ok by design. After receiving Unregistered event, no events should be emitted and application wont be handling it coz ideally they are supposed to destroy the line object post receiving Unregistered event

Kesari3008 (changes requested):
Let's discuss the comments if needed

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -386,6 +388,64 @@ export class Registration implements IRegistration {
...
+      serverType,
+      error.headers?.trackingid ?? '',
+      keepaliveRetryCount,
+      lineError
+    );
+
+    await this.performHardStopCleanup(METHODS.HANDLE_409_KEEPALIVE_FAILURE, {
+      reason: HARD_STOP_REASON.SESSION_SUPERSEDED,
+      error: lineError,
+    });
```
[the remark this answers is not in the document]
rarajes2: It's already handled. SDK receives "normal_disconnect" event from mercury and mobius-socket as soon as the same user registers in a different session. Vidcast is attached in the PR description.

packages/calling/src/CallingClient/registration/register.ts
rarajes2: Done

packages/calling/src/CallingClient/line/index.ts
rarajes2: Done

packages/calling/src/CallingClient/line/types.ts
rarajes2: Done

packages/calling/src/CallingClient/registration/register.ts
rarajes2: Renamed to `registrationCleanup `

rsarika (approved):

packages/calling/src/common/Utils.ts
```diff
@@ -345,15 +351,32 @@ export async function handleRegistrationErrors(
...
   let shouldDisconnect = false;
+  let handledByCallback = false;
   const lineError = createLineError('', {}, ERROR_TYPE.DEFAULT, RegistrationStatus.INACTIVE);
 
   const errorCode = Number(err.statusCode);
   let finalError = false;
+
+  /* Treatment for a status code this handler has no specific rule for: notify the
+   * consumer and leave the retry decision to the caller. */
+  const handleUnknownError = () => {
```
Kesari3008, line 367: Using this for 409 error and stating unknown error doesn't seem correct. We should dedicated line error message in line error context and same goes with log.warn as well and we don't need separate method just to cover 2 switch cases

packages/calling/src/common/Utils.ts
```diff
@@ -407,6 +430,33 @@ export async function handleRegistrationErrors(
       break;
     }
 
+    case ERROR_CODE.CONFLICT: {
+      /* Mobius answers 409 when this device is gone while the same user still holds an
+       * active registration elsewhere, typically calling opened in another browser tab.
+       * Only a flow that can hard stop the session opts into this handling; the remaining
+       * flows keep treating 409 as an unknown error. */
+      if (!sessionSupersededCb) {
+        handleUnknownError();
```
Kesari3008, line 439: We can just breakout of it instead of stating it as unknown error. Just logging it should be ok

packages/calling/src/common/Utils.ts
```diff
@@ -473,7 +523,7 @@ export async function handleRegistrationErrors(
         emitterCb(lineError, finalError);
         shouldDisconnect = serverCount > 1;
 
-        return {finalError, shouldDisconnect};
```
Kesari3008, line 476: Why do we need to add this flag in other error codes ?

packages/calling/src/CallingClient/registration/register.ts
```diff
@@ -1212,10 +1269,25 @@ export class Registration implements IRegistration {
...
+                    this.handle409KeepaliveFailure(
+                      clientError,
+                      error,
+                      serverType,
+                      event.data.keepAliveRetryCount
+                    ),
+                }
               );
 
+              if (handledByCallback) {
```
Kesari3008, line 1285: We can just use the status code to return, we don't need a new callback

packages/calling/src/CallingClient/registration/ai-docs/registration-spec.md
```diff
@@ -217,6 +253,7 @@ sequenceDiagram
...
+| REGISTRATION-R-010 | A keepalive answered with `409 Conflict` is a hard stop on first occurrence: `handleRegistrationErrors` classifies the status code and delegates to the keepalive-only session-superseded handler, the keepalive worker is terminated, every re-registration path is skipped, hard-stop cleanup closes WSS and sets status to `INACTIVE`, and the consumer receives a single `LINE_EVENTS.UNREGISTERED` carrying a `SESSION_SUPERSEDED` `LineError` as the reason; a `KEEPALIVE_ERROR` metric is submitted and logs are uploaded. | Mobius returns `409` when the same user registered elsewhere (typically a second browser tab), so re-registering would unregister that other device and start a registration ping-pong between the two. | `src/CallingClient/registration/register.ts` | `src/CallingClient/registration/register.test.ts` | none identified | PRESENT |
```
Kesari3008: 010 should come after 009 or change the number

packages/calling/src/CallingClient/registration/ai-docs/ARCHITECTURE.md
```diff
@@ -532,6 +533,41 @@ sequenceDiagram
 
 > **Note:** The synthetic `MOBIUS_SOCKET_4001_EVENT` envelope emitted when the server closes the socket with code `4001` carries `eventType: 'registration.down'` and therefore drives the **same** cleanup path as a server-pushed async `registration.down`. See [`mobius-socket/ai-docs/ARCHITECTURE.md`](../../../mobius-socket/ai-docs/ARCHITECTURE.md) for the close-code matrix.
 
+### Keepalive `409 Conflict` — Session Superseded
```
Kesari3008: This is redundant, it's present here as well as in registration-spec.md

packages/calling/ai-docs/patterns/error-handling-patterns.md
```diff
@@ -389,6 +393,7 @@ Key behaviors by status code:
 - **401 Unauthorized** — final error, emits token error
 - **403 Forbidden** — inspects `errorCode` in body for device-limit-exceeded (triggers `restoreRegCb`), device-creation-disabled (final error), or device-creation-failed (non-final)
 - **404 Device Not Found** — final error; on keepalive, triggers `handle404KeepaliveFailure` which re-attempts registration
+- **409 Conflict** — final error when the caller passed `sessionSupersededCb`: the handler owns the hard stop and `handledByCallback` is returned, so no event is emitted from here. Callers without that handler keep the unknown-error treatment
```
Kesari3008: Would be good to specify that this error handling is only for keepalive

packages/calling/ai-docs/patterns/error-handling-patterns.md
```diff
@@ -343,44 +343,48 @@ export type LineErrorEmitterCallback = (err: LineError, finalError?: boolean) =>
 
 ### handleRegistrationErrors
 
-Handles registration and keepalive error flows by mapping HTTP status codes to `LineError` and deciding whether to emit a final error, trigger a retry, or restore registration. Used by `Registration` for both initial registration failures and keepalive failures received from the web worker. Returns `Promise<boolean>` indicating whether to abort. See `src/common/Utils.ts`.
+Handles registration and keepalive error flows by mapping HTTP status codes to `LineError` and deciding whether to emit a final error, trigger a retry, restore registration, or hard stop a superseded session. Used by `Registration` for both initial registration failures and keepalive failures received from the web worker. See `src/common/Utils.ts`.
```
Kesari3008: Instead of saying hard stop a superseded session, we should say or do registration cleanup.

Kesari3008 (approved):
Approving the PR keep the deadline in mind but please ensure to address these comments before merge. They need addressing

packages/calling/src/common/Utils.ts
rarajes2: This is not just for the 409 error. This is the cleaner way to handle it --- the logic is extracted from the default case and re-used.

packages/calling/src/common/Utils.ts
rarajes2: This is just a safety net for when callback for handling the superseded session is not provided.

packages/calling/src/common/Utils.ts
rarajes2: This was done during the mobius-socket development --- To ensure if we need to disconnect the socket or keep it connected.

packages/calling/src/CallingClient/registration/register.ts
rarajes2: This is not a new callback. However based on your previous comment about delegating the error handling to the handleRegistrationError method, the logic was encapsulated inside a callback and passed along just the way we have for the handle429, etc

Reference - https://github.com/webex/webex-js-sdk/pull/5172#discussion_r3810184784

packages/calling/src/CallingClient/registration/ai-docs/registration-spec.md
rarajes2: Done

packages/calling/src/CallingClient/registration/ai-docs/ARCHITECTURE.md
rarajes2: Removed

packages/calling/ai-docs/patterns/error-handling-patterns.md
rarajes2: Done

packages/calling/ai-docs/patterns/error-handling-patterns.md
rarajes2: Done

packages/calling/src/common/Utils.ts
Kesari3008: I wasn't talking about shouldDisconnect flag but handedbyCallback flag. It seems to have been added in some switchcases

packages/calling/src/common/Utils.ts
Kesari3008: When the callback for 409 isn't passed and we receive 409 and we match we would be marking the error as unknown as per this code which is not needed. We can just ignore such cases so this additional method was not needed and unknonw block could remain same

packages/calling/src/CallingClient/registration/register.ts
Kesari3008: My bad...it was a typo. I meant to say we don't need to maintain a new flag. You already have existing information, marking final error as true and error status code. Even though if you have the block inside handleRegistrationErrors since we need to prevent it from running the remianing logic, we need to return from here which can be done using existing values instead of introducing new flag

packages/calling/src/common/Utils.ts
Kesari3008: Yeah and it's ok the to have if but we need not to invoke steps inside hanlde unknown error. Just breaking out of the case should be enough

PR #5174, opened 2026-08-18
Participant drop functionality and tests

ajitkumar0828-spec:
# COMPLETES # https://jira-eng-sjc12.cisco.com/jira/browse/CAI-8464

Added Drop Button functionality for agent desktop

< DESCRIBE YOUR CHANGES >

https://app.vidcast.io/share/2861c66e-0857-41f4-8ea5-6daeeb0702a3

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

  - [ ] Github Copilot
  - [x] Other - Please Specify
- [ ] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---


packages/@webex/contact-center/.sdd/manifest.json
```diff
@@ -51,7 +51,7 @@
       "canonical_spec": "ai-docs/contact-center-spec.md",
       "contracts": {
         "provides": [
-          "Published @webex/contact-center package; ContactCenter WebexPlugin; public SDK methods, events, types, Task, AddressBook, ApiAIAssistant, UserPreference, cc.userPreference, exported user-preference request/response types, and preview-campaign methods acceptPreviewContact/skipPreviewContact/removePreviewContact"
+          "Published @webex/contact-center package; ContactCenter WebexPlugin; public SDK methods, events, types, Task, AddressBook, ApiAIAssistant, UserPreference, cc.userPreference, exported user-preference request/response types, preview-campaign methods acceptPreviewContact/skipPreviewContact/removePreviewContact, and DropConferenceParticipantPayload/task.dropConferenceParticipant"
```
Kesari3008: Why are we adding specifically about this feature in the manifest file. I dont consult, tranfer or conference flows explicitly mentioned here so why only this feature

packages/@webex/contact-center/src/services/core/aqm-reqs.ts
```diff
@@ -117,15 +117,24 @@ export default class AqmReqs {
...
+                  module: AQM_REQS_FILE,
+                  method: METHODS.CREATE_PROMISE,
+                });
+              } else {
+                LoggerProxy.log(`Routing request failed: ${JSON.stringify(msg)}`, {
+                  module: AQM_REQS_FILE,
+                  method: METHODS.CREATE_PROMISE,
+                });
+              }
               const eerr = new Err.Details(notifFail.errId, msg as any);
```
Kesari3008: There is a typo here for this error variable, we can fix it. Also why do we need to add this condition ?

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -350,6 +409,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
               actions: ['updateTaskData', 'clearConsultState', 'emitTaskConsultEnd'],
             },
           ],
+          [TaskEvent.PARTICIPANT_LEAVE]: [
```
Kesari3008, line 423: This state transition is not following the definition of state machine. target and actions are the property it takes which is missing here. Why is a new event being introduced like this ?

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -827,6 +828,90 @@ export default class Voice extends Task implements IVoice {
...
+  /**
+   * Removes another participant from an active conference.
+   * Completion is correlated with ParticipantLeftConference by the AQM request layer.
+   * @param payload - Participant identifier to remove
+   * @returns Promise<TaskResponse>
+   */
+  public async dropConferenceParticipant(
+    payload: DropConferenceParticipantPayload
+  ): Promise<TaskResponse> {
+    const participantId = payload?.participantId;
```
Kesari3008: payload is mandatory param in the method definition so why is ? present marking it optionally being present

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -827,6 +828,90 @@ export default class Voice extends Task implements IVoice {
...
   }
 
+  /**
+   * Removes another participant from an active conference.
+   * Completion is correlated with ParticipantLeftConference by the AQM request layer.
+   * @param payload - Participant identifier to remove
+   * @returns Promise<TaskResponse>
+   */
+  public async dropConferenceParticipant(
+    payload: DropConferenceParticipantPayload
```
Kesari3008, line 1189: This payload seems to be be having just one property, why have created new type for it. We can directly say in the definition participantId: string and also if we are specifically defining it to be string why is the string check in the if block needed

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -827,6 +828,90 @@ export default class Voice extends Task implements IVoice {
...
+
+    if (typeof participantId !== 'string' || participantId.trim().length === 0) {
+      throw new Error('participantId must be a non-empty string');
+    }
+
+    const requestInteractionId =
+      this.data.interaction?.mainInteractionId || this.data.interactionId;
+
+    try {
+      LoggerProxy.info('Dropping conference participant', {
```
Kesari3008: This logger can be present at the start of the method, not here. Having it inside try block is not needed

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -827,6 +828,90 @@ export default class Voice extends Task implements IVoice {
...
+
+      const response = await this.contact.dropConferenceParticipant({
+        interactionId: requestInteractionId,
+        data: {participantId},
+      });
+
+      this.metricsManager.trackEvent(
+        METRIC_EVENT_NAMES.TASK_CONFERENCE_PARTICIPANT_DROP_SUCCESS,
+        {
+          taskId: this.data.interactionId,
```
Kesari3008, line 1228: This value and requestInteractId are the same. Just one filed in the metric is enough. Same comment for failure metric metadata as well

packages/@webex/contact-center/src/services/task/voice/Voice.ts
```diff
@@ -827,6 +828,90 @@ export default class Voice extends Task implements IVoice {
...
+          failureType: failureTrackingFields.failureType,
+          reasonCode: failureTrackingFields.reasonCode,
+        },
+        ['operational', 'behavioral', 'business']
+      );
+
+      LoggerProxy.error('Failed to drop conference participant', {
+        module: CC_FILE,
+        method: METHODS.DROP_CONFERENCE_PARTICIPANT,
+        interactionId: requestInteractionId,
```
Kesari3008, line 1268: Print trackingId here as well

packages/@webex/contact-center/src/services/task/constants.ts
```diff
@@ -73,6 +73,7 @@ export const METHODS = {
   TRANSFER: 'transfer',
   CONSULT_TRANSFER: 'consultTransfer',
   CONSULT_CONFERENCE: 'consultConference',
+  DROP_CONFERENCE_PARTICIPANT: 'dropConferenceParticipant',
```
Kesari3008, line 76: I believe this is too lengthy. dropParticipant is enough for a name but its a nitpick. Applied everywhere for the name

packages/@webex/contact-center/src/services/task/contact.ts
```diff
@@ -458,6 +458,38 @@ export default function routingContact(aqm: AqmReqs) {
       })
     ),
 
+    /*
+     * Drop another participant from an active conference
+     */
+    dropConferenceParticipant: aqm.req(
+      (p: {interactionId: string; data: Contact.DropConferenceParticipantPayload}) => ({
```
Kesari3008: Same comment here about the payload type

packages/@webex/contact-center/src/services/task/Task.ts
```diff
@@ -141,6 +142,15 @@ export default abstract class Task extends EventEmitter implements ITask {
     this.unsupportedMethodError('consultConference');
   }
 
+  public async dropConferenceParticipant(
+    payload: DropConferenceParticipantPayload
+  ): Promise<TaskResponse> {
+    if (payload) {
```
Kesari3008: This If is not needed

webpack.config.js
```diff
@@ -76,20 +76,31 @@ module.exports = (env = {NODE_ENV: process.env.NODE_ENV || 'production'}) => ({
   },
   mode: env && env.NODE_ENV === 'development' ? 'development' : 'production',
   output: {
-    path: path.resolve(__dirname),
+    path: path.resolve(__dirname, env && env.WEBPACK_SERVE ? 'docs/samples' : ''),
```
Kesari3008, line 81: What are these changes about ?

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -705,35 +793,9 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
-                'clearConsultState',
-                'emitTaskParticipantLeft',
-                'emitTaskEnd',
-              ],
-            },
-            {
-              guard: (params) =>
-                !guards.didCurrentAgentLeaveConference(params) &&
-                guards.shouldDowngradeConferenceToConnected(params),
+              guard: guards.shouldDowngradeConferenceToConnected,
```
Kesari3008: We have removed this guard: didCurrentAgentLeaveConference for transition to Connected state. Was this intentional change, did we test if this doesn't break anything

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -58,6 +58,64 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
...
+        'updateTaskData',
+        'markEnded',
+        'clearConsultState',
+        'emitTaskConsultEnd',
+        'emitTaskEnd',
+      ],
+    },
+  ];
+
+  const survivingParticipantLeaveTransition = {
```
Kesari3008: This optimization is unncessary

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
Kesari3008, line 423: Why would be any Agent in Connected state if they are in conference. Connected state is only for 1:1 call with customer. And when a participant leave happens, it can only happen in conference state. This transition may not be valid

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
```diff
@@ -401,6 +464,10 @@ export function getTaskStateMachineConfig(uiControlConfig: UIControlConfig) {
             target: TaskState.CONNECTED,
             actions: ['updateTaskData'],
           },
+          [TaskEvent.PARTICIPANT_LEAVE]: [
```
Kesari3008, line 492: Same comment here. We will have to evaluate if drop is something being attempted in HOLD_INITIATING state or RESUME_INITIATING state

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
```diff
@@ -515,12 +515,12 @@ export const actions: TaskActionsMap = {
   }),
   handleConferenceStarted: assign({consultInitiator: false}),
 
-  setConsultDestination: assign(({event}: TaskActionArgs) => {
+  setConsultDestination: assign(({context, event}: TaskActionArgs) => {
```
Kesari3008, line 518: Why have we introduced this change ?

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
```diff
@@ -383,12 +384,34 @@ export const guards = {
       return true;
     }
 
-    //    For EP-DN agents the backend removes the leaving participant entirely
-    //    from the participants map (rather than setting hasLeft). If this task
-    //    is in CONFERENCING (implied by the guard being evaluated here) but the
-    //    agent is absent from the updated participants, they have left.
-    const participants = taskData?.interaction?.participants;
-    if (participants && !(selfAgentId in participants)) {
+    const currentParticipant = taskData?.interaction?.participants?.[selfAgentId];
```
Kesari3008, line 389: Would like to discuss this change ?

Kesari3008 (changes requested):
Except test files and .md files, reviewed the code

Kesari3008:
Update Jira description with JIRA link, changes done and vidcast link for testing done in SDK

packages/@webex/contact-center/src/services/core/aqm-reqs.ts
ajitkumar0828-spec: Fixed the eerr typo and simplified the logging branches. The condition is required because Participant Drop URLs and routing events can contain participant identifiers or phone numbers. When redactSensitiveLogs is enabled, logging the raw message or constructed error could expose those values. Normal AQM requests retain the existing detailed logs, while sensitive requests emit one generic redacted message. A unit assertion was added for this behavior.

packages/@webex/contact-center/.sdd/manifest.json
ajitkumar0828-spec: Agreed. I removed the Participant Drop-specific details

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
ajitkumar0828-spec: PARTICIPANT_LEAVE is not a new event; the existing ParticipantLeftConference routing event is mapped to it by TaskManager and it was already handled in the conference states. XState permits an event to use an ordered array of transition configurations. The shared guarded entries each contain target and actions. The final action-only entry intentionally has no target because it updates a surviving task without changing its current state. I added comments and the event mapping to make this explicit.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
ajitkumar0828-spec: I removed it and now validate the payload explicitly before directly accessing participantId. The runtime guard remains necessary because the published SDK can also be called from untyped JavaScript, where callers can still pass undefined. The public TypeScript signature remains mandatory and unchanged.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
ajitkumar0828-spec: The named payload type is intentional and follows existing Contact Center SDK conventions, including one-property types such as HoldResumePayload, ResumeRecordingPayload, and declinePayload. It also provides one shared public contract across ITask, base Task, and Voice, and is already consumed by widgets as {participantId}.  
The string validation is a runtime boundary check. TypeScript types are not present at runtime, and JavaScript callers can still pass an invalid or blank value. We validate before metrics and AQM registration so an invalid identifier cannot reach the routing request. I have removed the confusing optional chaining and made the runtime payload check explicit.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
ajitkumar0828-spec: Moved the informational log outside the try block. It remains after input validation and interaction-ID resolution so invalid requests do not produce a Drop-attempt log. The try block now covers only metrics and the AQM Drop operation.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
ajitkumar0828-spec: These fields match for normal tasks but differ in supported child/main-interaction flows. taskId identifies the SDK task instance, which can remain child-interaction keyed after consult or EP-DN handling, while requestInteractionId is the current main interaction sent to AQM. This follows the existing metric convention used by consult() and end(). I added an explanatory comment and strengthened both success and failure coverage with differing IDs.

packages/@webex/contact-center/src/services/task/voice/Voice.ts
ajitkumar0828-spec: Added the tracking ID from the already-sanitized AQM failure tracking fields to the participant Drop failure log. Updated tests cover both structured failures with a tracking ID and rejection values where it is unavailable.

packages/@webex/contact-center/src/services/task/constants.ts
ajitkumar0828-spec: api have the link /conference/participants/ so added the conference to make it uniform with api endpoint.

packages/@webex/contact-center/src/services/task/contact.ts
ajitkumar0828-spec: Simplified the internal contact-service request to {interactionId, participantId}. The participant ID is a URL path parameter rather than request-body data, so the nested payload type was unnecessary here. The public task payload remains unchanged.

packages/@webex/contact-center/src/services/task/Task.ts
ajitkumar0828-spec: if (payload) block has been removed

webpack.config.js
ajitkumar0828-spec: This aligns webpack-dev-server’s in-memory output with the URLs used by the sample pages. The Contact Center sample requests /samples/contact-center.min.js; previously the development bundle was exposed under a mismatched nested path, resulting in Webex is not defined. Production output paths remain unchanged. I added an inline explanation and removed the unrelated overlay customization.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
ajitkumar0828-spec: The removal was intentional because shouldDowngradeConferenceToConnected already requires the current agent to remain on the main-call media. However, I restored the explicit negation using the newer didCurrentAgentLeaveMainInteraction guard to make the invariant clear and provide defense in depth. Added tests verify both the surviving-agent downgrade and current-agent termination paths.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
ajitkumar0828-spec: Removed the shared survivingParticipantLeaveTransition object and inlined the action-only transition at each participant-leave handler. This makes the state definitions explicit without changing behavior.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
ajitkumar0828-spec: CONNECTED is the normal steady state for a 1:1 call. This handler covers an event-ordering race in child-task/EP-DN flows where ParticipantLeftConference can arrive before the local actor has moved out of CONNECTED. Without handling it, an explicit self-departure is ignored and the stale call UI remains until refresh. The guarded transition terminates only when the event contains authoritative evidence that the current agent left; another participant’s leave preserves the CONNECTED state. A regression test now covers this scenario.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
ajitkumar0828-spec: Drop does not need to be initiated by the agent whose task is in HOLD_INITIATING or RESUME_INITIATING. A primary agent can remotely drop that agent while their local hold/unhold request is awaiting its routing acknowledgment. In that race, ParticipantLeftConference must still terminate the removed agent’s task. The transition only terminates with explicit current-agent departure evidence; an event for another participant updates the task while preserving the transient state. Regression tests now cover both behaviors.

packages/@webex/contact-center/src/services/task/state-machine/actions.ts
ajitkumar0828-spec: CONSULT is a user-action event and does not carry taskData. The previous implementation therefore replaced the existing state-machine task snapshot with only {destAgentId}. This removed the prior main-call participant/media information required to process an immediately following ParticipantLeftConference or AgentConsultEnded event. The action now explicitly preserves context.taskData while adding the consult destination. A regression assertion verifies that the interaction snapshot remains intact.

packages/@webex/contact-center/src/services/task/state-machine/guards.ts
ajitkumar0828-spec: This guard was introduced for the cross-channel EP-DN scenario where Agent 2 is removed from the main interaction while consulting Agent 3. Depending on event shape, routing may represent departure by participantId, hasLeft, removal from the participant map, or—for AgentConsultEnded—removal from mainCall. Agent Desktop applies the same participant/main-call membership checks when deciding whether to remove the task.  
The SDK guard is deliberately narrower: an event naming another participant cannot terminate the current agent, missing/partial snapshots are non-terminal, and main-call membership fallback is limited to a conference-originated CONSULT_END with corroborating active participant-map and consult-leg membership. This prevents false termination while fixing the stale cross-channel call UI.

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
Kesari3008: THese comments need revisiting

packages/@webex/contact-center/src/services/task/state-machine/TaskStateMachine.ts
Kesari3008: THese comments need revisiting

webpack.config.js
```diff
@@ -76,13 +76,20 @@ module.exports = (env = {NODE_ENV: process.env.NODE_ENV || 'production'}) => ({
   },
   mode: env && env.NODE_ENV === 'development' ? 'development' : 'production',
   output: {
-    path: path.resolve(__dirname),
+    // Sample pages load bundles from /samples/*.min.js. Align webpack-dev-server's
+    // in-memory output with that URL while preserving the existing production paths.
+    path: path.resolve(__dirname, env && env.WEBPACK_SERVE ? 'docs/samples' : ''),
```
Kesari3008, line 81: Revert this change before merge if this has nothing to do with drop feature

Kesari3008 (approved):
Approving the PR considering the time crunch but few changes need revisiting and cleanup

akulakum (approved):

webpack.config.js
ajitkumar0828-spec: reverted

PR #5175, opened 2026-08-18
fix(calling): neutralize SCIM filter injection and restrict Mobius WSS hosts (CAI-8461)

mkesavan13:
# COMPLETES https://jira-eng-sjc12.cisco.com/jira/browse/CAI-8461

Three P0 security findings in `@webex/calling`:

- **U-01** — SCIM filter injection via SIP `externalId` (`CallerId/index.ts:103-122`): attacker-controlled value from a SIP header was interpolated raw into `id eq "..."` filter expressions, allowing injection of arbitrary SCIM filter clauses.
- **U-02** — SCIM filter injection via `userExternalId` (`Utils.ts:1552-1565`): same pattern in `resolveContact` — the `callingPartyInfo.userExternalId` field was interpolated without escaping.
- **U-03** — Mobius WSS socket with no host allowlist (`Utils.ts:139-223`, `socket-base.ts:238-317`): discovery-provided WSS URIs were trusted as socket destinations without validation, enabling a malicious discovery response to exfiltrate the Mobius bearer token to an arbitrary host.

**Root cause:** Attacker-influenceable identifiers were interpolated raw into SCIM filter expressions, and discovery-provided WSS hosts were trusted as socket destinations without an allowlist. The existing `encodeURIComponent(filter)` transport-layer encoding does not prevent SCIM filter grammar injection; a value-layer escape is required. No host validation existed before token transmission over WebSocket.

- **`packages/calling/src/common/constants.ts`** — Added `MOBIUS_WSS_ALLOWED_DOMAINS` allowlist (`webex.com`, `wbx2.com`, `webexapis.com`, `cisco.com`)
- **`packages/calling/src/common/Utils.ts`** — Added `escapeScimFilterValue` helper (escapes `\` then `"`); applied to `resolveContact` `userExternalId`; added host-allowlist filter `isTrustedMobiusWssHost` to `filterMobiusUris` dropping non-allowlisted WSS URIs
- **`packages/calling/src/CallingClient/calling/CallerId/index.ts`** — Applied `escapeScimFilterValue` to SIP `externalId` before SCIM filter interpolation in `parseRemotePartyInfo`
- **`packages/calling/src/mobius-socket/socket/socket-base.ts`** — Added pre-WebSocket host allowlist gate in `open()` — rejects with typed `ConnectionError` before `authorize(token)` runs
- **`packages/calling/src/common/Utils.test.ts`** — Added UT-1 (`resolveContact` injection) and UT-4 (`filterMobiusUris` untrusted host); updated existing tests
- **`packages/calling/src/CallingClient/calling/CallerId/index.test.ts`** — Added UT-2 (`parseRemotePartyInfo` injection)
- **`packages/calling/src/mobius-socket/socket.test.ts`** — Added UT-3 (socket.open non-allowlisted rejection + allowlisted acceptance)
- **`packages/calling/jest.config.js`** — Updated Jest module name mapper to use new stub mocks
- **`packages/calling/package.json`** — Added pre-build of `@webex/legacy-tools` to resolve compile dependency
- **`packages/calling/test/mocks/`** — Added `internal-plugin-metrics-stub.js`, `webex-core-stub.js`, `empty-stub.js` for isolated unit test execution
- **`packages/calling/ai-docs/`** — Updated `SECURITY.md`, `GETTING_STARTED.md`, `caller-id-spec.md`, `mobius-socket-spec.md` to document the new controls

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- Tests added: 4 new unit tests (UT-1 through UT-4) across 4 test files
- Gate 1 verification: passed — `yarn workspace @webex/calling build:src` compiled successfully
- Gate 2 verification: passed — `yarn workspace @webex/calling test:unit` 857/857 tests passed
- Gate 3 verification: not run
- Coverage outcome: passed

**Targeted test results (4 new security-focused suites):**
```
yarn workspace @webex/calling test:unit \
  --testPathPattern="Utils.test.ts|CallerId/index.test.ts|socket.test.ts|mobius-socket.test.ts"

Test Suites: 4 passed, 4 total
Tests:       252 passed, 252 total
```

## Acceptance Criteria

| ID | Criterion | Source | JiraToPr status | Evidence |
|---|---|---|---|---|
| AC-1 | Attacker-controlled SIP externalId and userExternalId values can no longer alter SCIM filter grammar: values are escaped/validated before being interpolated into `id eq "..."`, so injected filter clauses (e.g. a value containing a double-quote) are neutralized while legitimate identifiers still resolve. | Jira findings U-01, U-02 | Unit validated | `escapeScimFilterValue` applied at `CallerId/index.ts:116` and `Utils.ts:1600-1603`; UT-1 and UT-2 confirmed passing in Gate 2 |
| AC-2 | The Mobius bearer token is only transmitted over a WSS socket whose host is on a trusted allowlist: discovery-provided WSS URIs with non-allowlisted hosts are dropped, and socket open rejects (before authorize) any URL whose host is not allowlisted, so a malicious discovery response cannot exfiltrate the token to an arbitrary host. | Jira findings U-03 | Unit validated | `MOBIUS_WSS_ALLOWED_DOMAINS` constant + `isTrustedMobiusWssHost` + `filterMobiusUris` drop + `socket-base.ts` pre-connect guard; UT-3 and UT-4 confirmed passing in Gate 2 |

## Contract Discovery Warnings

- Manifest reference discovery capped at 100 strings

  - [ ] Github Copilot
  - [x] Other - Please Specify: JiraToPr (Claude Sonnet 4.5 via Anthropic API)
- [x] This PR is related to
  - [ ] Feature
  - [x] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

Jira: https://jira-eng-sjc12.cisco.com/jira/browse/CAI-8461

### DIFF
--- a/packages/calling/jest.config.js
+++ b/packages/calling/jest.config.js
@@ -43,9 +43,20 @@ const jestConfig = {
   },
   coverageDirectory: 'coverage',
   coverageReporters: ['clover', 'json', 'lcov'],
-  transformIgnorePatterns: ['/node_modules/(?!(@webex/internal-media-core)/)'],
+  transformIgnorePatterns: [
+    '/node_modules/(?!(@webex/internal-media-core|@webex/media-helpers|@webex/test-helper-mock-web-socket|@webex/common-timers)/)/',
+  ],
   testMatch: ['<rootDir>/src/**/*.test.[jt]s'],
-  moduleNameMapper: {'^uuid$': 'uuid'},
+  moduleNameMapper: {
+    '^uuid$': 'uuid',
+    '^@webex/media-helpers$': '<rootDir>/../../node_modules/@webex/media-helpers/src/index.ts',
+    '^@webex/internal-plugin-metrics$': '<rootDir>/test/mocks/internal-plugin-metrics-stub.js',
+    '^@webex/internal-plugin-device$': '<rootDir>/test/mocks/empty-stub.js',
+    '^@webex/internal-plugin-feature$': '<rootDir>/test/mocks/empty-stub.js',
+    '^@webex/(test-helper-[^/]+)$': '<rootDir>/../../node_modules/@webex/$1/src/index.js',
+    '^@webex/common$': '<rootDir>/../../node_modules/@webex/common/src/index.js',
+    '^@webex/common-timers$': '<rootDir>/../../node_modules/@webex/common-timers/src/index.ts',
+  },
   reporters: [
     'default',
     [
--- a/packages/calling/jest.config.js
+++ b/packages/calling/jest.config.js
@@ -44,7 +44,7 @@ const jestConfig = {
   coverageDirectory: 'coverage',
   coverageReporters: ['clover', 'json', 'lcov'],
   transformIgnorePatterns: [
-    '/node_modules/(?!(@webex/internal-media-core|@webex/media-helpers|@webex/test-helper-mock-web-socket|@webex/common-timers)/)/',
+    '/node_modules/(?!(@webex/internal-media-core|@webex/media-helpers|@webex/test-helper-mock-web-socket|@webex/common-timers)/)',
   ],
   testMatch: ['<rootDir>/src/**/*.test.[jt]s'],
   moduleNameMapper: {
--- a/packages/calling/jest.config.js
+++ b/packages/calling/jest.config.js
@@ -51,8 +51,6 @@ const jestConfig = {
     '^uuid$': 'uuid',
     '^@webex/media-helpers$': '<rootDir>/../../node_modules/@webex/media-helpers/src/index.ts',
     '^@webex/internal-plugin-metrics$': '<rootDir>/test/mocks/internal-plugin-metrics-stub.js',
-    '^@webex/internal-plugin-device$': '<rootDir>/test/mocks/empty-stub.js',
-    '^@webex/internal-plugin-feature$': '<rootDir>/test/mocks/empty-stub.js',
     '^@webex/(test-helper-[^/]+)$': '<rootDir>/../../node_modules/@webex/$1/src/index.js',
     '^@webex/common$': '<rootDir>/../../node_modules/@webex/common/src/index.js',
     '^@webex/common-timers$': '<rootDir>/../../node_modules/@webex/common-timers/src/index.ts',
--- a/packages/calling/package.json
+++ b/packages/calling/package.json
@@ -31,7 +31,7 @@
     "prebuild": "rimraf dist",
     "build": "tsc",
     "build:src": "tsc",
-    "compile": "yarn workspaces foreach -tR --from @webex/calling run build:src",
+    "compile": "yarn workspaces foreach -tR --from @webex/legacy-tools run build:src && yarn workspaces foreach -tR --from @webex/calling run build:src",
     "test:unit": "jest --config=jest.config.js --runInBand",
     "test:style": "eslint 'src/**/*.ts'",
     "fix:lint": "eslint 'src/**/*.ts' --fix",
--- a/packages/calling/package.json
+++ b/packages/calling/package.json
@@ -31,7 +31,7 @@
     "prebuild": "rimraf dist",
     "build": "tsc",
     "build:src": "tsc",
-    "compile": "yarn workspaces foreach -tR --from @webex/legacy-tools run build:src && yarn workspaces foreach -tR --from @webex/calling run build:src",
+    "compile": "yarn workspaces foreach -tR --from @webex/calling run build:src",
     "test:unit": "jest --config=jest.config.js --runInBand",
     "test:style": "eslint 'src/**/*.ts'",
     "fix:lint": "eslint 'src/**/*.ts' --fix",
--- a/packages/calling/src/CallingClient/calling/CallerId/index.test.ts
+++ b/packages/calling/src/CallingClient/calling/CallerId/index.test.ts
@@ -256,4 +256,40 @@ describe('CallerId tests', () => {
     expect(callerId['callerInfo'].name).toStrictEqual('Bob Marley');
     expect(callerId['callerInfo'].num).toStrictEqual('5888');
   });
+
+  it('parseRemotePartyInfo escapes SCIM filter injection in externalId', async () => {
+    const injectionPayload = '69fde5ad-fb8b-4a1b-9998-b0999e95719b" or userName pr "';
+    const dummyCallerId = {
+      'x-broadworks-remote-party-info': `userId="nkjwuovmbo@64941297.int10.bcld.webex.com";userDn="tel:+12142865888;ext=5888;country-code=1";externalId=${injectionPayload}`,
+    };
+
+    webex.request.mockResolvedValue({
+      statusCode: 200,
+      body: dummyScimResponse,
+    });
+
+    const localCallerId = createCallerId(webex, () => {});
+
+    localCallerId['callerInfo'] = {} as DisplayInformation;
+    localCallerId.fetchCallerDetails(dummyCallerId);
+    await waitForMsecs(50);
+
+    // The request should have been made
+    expect(webex.request).toHaveBeenCalled();
+
+    // The raw unescaped injection payload must NOT appear in the request URI
+    const lastCall = (webex.request as jest.Mock).mock.calls[
+      (webex.request as jest.Mock).mock.calls.length - 1
+    ];
+    const requestedUri: string = lastCall[0].uri;
+    const rawFilter = encodeURIComponent(`id eq "${injectionPayload}"`);
+
+    expect(requestedUri).not.toContain(rawFilter);
+
+    // The escaped form must appear instead
+    const escaped = injectionPayload.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+    const escapedFilter = encodeURIComponent(`id eq "${escaped}"`);
+
+    expect(requestedUri).toContain(escapedFilter);
+  });
 });
--- a/packages/calling/src/CallingClient/calling/CallerId/index.test.ts
+++ b/packages/calling/src/CallingClient/calling/CallerId/index.test.ts
@@ -112,9 +112,9 @@ describe('CallerId tests', () => {
    * We will check whether the priority order is followed or not.
    */
   it(' When PA-ID, From header is present along with x-broad-works, ', async () => {
+    const injectionPayload = '69fde5ad-fb8b-4a1b-9998-b0999e95719b" or userName pr "';
     const dummyCallerId = {
-      'x-broadworks-remote-party-info':
-        'userId="nkjwuovmbo@64941297.int10.bcld.webex.com";userDn="tel:+12142865888;ext=5888;country-code=1";externalId=69fde5ad-fb8b-4a1b-9998-b0999e95719b',
+      'x-broadworks-remote-party-info': `userId="nkjwuovmbo@64941297.int10.bcld.webex.com";userDn="tel:+12142865888;ext=5888;country-code=1";externalId=${injectionPayload}`,
       'p-asserted-identity': '"John O\'Connor - ( Guest )" <sip:5888@10.155.4.7;user=phone>',
       from: '"Alice" <sip:5889@64941297.int10.bcld.webex.com>;tag=1932136170-1654008881246',
     };
@@ -131,6 +131,18 @@ describe('CallerId tests', () => {
     expect(callerId['callerInfo'].id).toStrictEqual(dummyScimResponse.Resources[0].id);
     expect(callerId['callerInfo'].name).toStrictEqual('Cathy');
     expect(callerId['callerInfo'].num).toStrictEqual('5008');
+
+    /* resolveCallerId must receive the escaped externalId, not the raw SCIM-injection payload (AC-1 / U-01) */
+    const lastCall = (webex.request as jest.Mock).mock.calls[
+      (webex.request as jest.Mock).mock.calls.length - 1
+    ];
+    const requestedUri: string = lastCall[0].uri;
+    const rawFilter = encodeURIComponent(`id eq "${injectionPayload}"`);
+    const escaped = injectionPayload.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+    const escapedFilter = encodeURIComponent(`id eq "${escaped}"`);
+
+    expect(requestedUri).not.toContain(rawFilter);
+    expect(requestedUri).toContain(escapedFilter);
   });
 
   it(' When PA-ID ,From header is present along with x-broad-works , but name in PAI is missing', async () => {
@@ -255,41 +267,5 @@ describe('CallerId tests', () => {
     expect(callerId['callerInfo'].id).toBe(undefined);
     expect(callerId['callerInfo'].name).toStrictEqual('Bob Marley');
     expect(callerId['callerInfo'].num).toStrictEqual('5888');
-  });
-
-  it('parseRemotePartyInfo escapes SCIM filter injection in externalId', async () => {
-    const injectionPayload = '69fde5ad-fb8b-4a1b-9998-b0999e95719b" or userName pr "';
-    const dummyCallerId = {
-      'x-broadworks-remote-party-info': `userId="nkjwuovmbo@64941297.int10.bcld.webex.com";userDn="tel:+12142865888;ext=5888;country-code=1";externalId=${injectionPayload}`,
-    };
-
-    webex.request.mockResolvedValue({
-      statusCode: 200,
-      body: dummyScimResponse,
-    });
-
-    const localCallerId = createCallerId(webex, () => {});
-
-    localCallerId['callerInfo'] = {} as DisplayInformation;
-    localCallerId.fetchCallerDetails(dummyCallerId);
-    await waitForMsecs(50);
-
-    // The request should have been made
-    expect(webex.request).toHaveBeenCalled();
-
-    // The raw unescaped injection payload must NOT appear in the request URI
-    const lastCall = (webex.request as jest.Mock).mock.calls[
-      (webex.request as jest.Mock).mock.calls.length - 1
-    ];
-    const requestedUri: string = lastCall[0].uri;
-    const rawFilter = encodeURIComponent(`id eq "${injectionPayload}"`);
-
-    expect(requestedUri).not.toContain(rawFilter);
-
-    // The escaped form must appear instead
-    const escaped = injectionPayload.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
-    const escapedFilter = encodeURIComponent(`id eq "${escaped}"`);
-
-    expect(requestedUri).toContain(escapedFilter);
   });
 });
--- a/packages/calling/src/common/Utils.ts
+++ b/packages/calling/src/common/Utils.ts
@@ -128,6 +128,7 @@ import {
   WEBEX_API_PROD,
   WEBEX_API_BTS,
   BW_XSI_ENDPOINT_VERSION_WITH_SLASH,
+  MOBIUS_WSS_ALLOWED_DOMAINS,
 } from './constants';
 import {Model, WDMDevice, WebexSDK} from '../SDKConnector/types';
 import SDKConnector from '../SDKConnector';
@@ -135,6 +136,23 @@ import {CallSettingResponse} from '../CallSettings/types';
 import {ContactResponse} from '../Contacts/types';
 import {LineErrorEmitterCallback} from '../CallingClient/line/types';
 import {LineError, createLineError} from '../Errors/catalog/LineError';
+
+/**
+ * Returns true when the hostname of the given WSS URI ends with one of the
+ * trusted Mobius/Webex domain suffixes.  URIs with unparseable hostnames are
+ * dropped for safety.
+ */
+function isTrustedMobiusWssHost(wssUri: string): boolean {
+  try {
+    const host = new URL(wssUri).hostname;
+
+    return MOBIUS_WSS_ALLOWED_DOMAINS.some(
+      (domain) => host === domain || host.endsWith(`.${domain}`)
+    );
+  } catch {
+    return false;
+  }
+}
 
 export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl: string) {
   const logContext = {
@@ -157,7 +175,12 @@ export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl:
   if (mobiusServers?.primary?.wss) {
     log.info('Adding Primary wss uris', logContext);
     for (const wssUri of mobiusServers.primary.wss) {
-      wssArrayPrimary.push(wssUri);
+      /* Drop WSS URIs whose host is not in the trusted allowlist (AC-2) */
+      if (isTrustedMobiusWssHost(wssUri)) {
+        wssArrayPrimary.push(wssUri);
+      } else {
+        log.warn(`Dropping untrusted primary WSS URI: host not in allowed domains`, logContext);
+      }
     }
   }
 
@@ -171,7 +194,12 @@ export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl:
   if (mobiusServers?.backup?.wss) {
     log.info('Adding Backup wss uris', logContext);
     for (const wssUri of mobiusServers.backup.wss) {
-      wssArrayBackup.push(wssUri);
+      /* Drop WSS URIs whose host is not in the trusted allowlist (AC-2) */
+      if (isTrustedMobiusWssHost(wssUri)) {
+        wssArrayBackup.push(wssUri);
+      } else {
+        log.warn(`Dropping untrusted backup WSS URI: host not in allowed domains`, logContext);
+      }
     }
   }
 
@@ -1545,6 +1573,22 @@ export async function resolveCallerIdByName(name: string) {
 }
 
 /**
+ * Escapes SCIM filter-grammar metacharacters in an opaque identifier value.
+ *
+ * Prevents filter injection when a value is interpolated into `id eq "..."`.
+ * Only the backslash and double-quote characters are significant in SCIM
+ * quoted-string literals; all other characters are kept as-is so that
+ * legitimate identifiers (UUIDs, SIP addresses, etc.) resolve unchanged.
+ *
+ * @param value - The raw identifier value to escape.
+ * @returns The escaped value safe for use inside a SCIM quoted-string literal.
+ */
+export function escapeScimFilterValue(value: string): string {
+  // Escape backslash first (order matters: must precede quote escaping)
+  return value.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+}
+
+/**
  * Resolve the contact information.
  *
  * @param callingPartyInfo - Calling Party Info.
@@ -1553,8 +1597,10 @@ export async function resolveContact(
   callingPartyInfo: CallingPartyInfo
 ): Promise<DisplayInformation | null> {
   if (callingPartyInfo.userExternalId && callingPartyInfo.userExternalId.$) {
-    /* SCIM Search */
-    return resolveCallerIdDisplay(`id eq "${callingPartyInfo.userExternalId.$}"`);
+    /* SCIM Search — escape the value to prevent filter-grammar injection (AC-1 / U-02) */
+    return resolveCallerIdDisplay(
+      `id eq "${escapeScimFilterValue(callingPartyInfo.userExternalId.$)}"`
+    );
   }
   if (callingPartyInfo.name && callingPartyInfo.name.$) {
     /* People Search */
--- a/packages/calling/src/common/Utils.ts
+++ b/packages/calling/src/common/Utils.ts
@@ -128,7 +128,7 @@ import {
   WEBEX_API_PROD,
   WEBEX_API_BTS,
   BW_XSI_ENDPOINT_VERSION_WITH_SLASH,
-  MOBIUS_WSS_ALLOWED_DOMAINS,
+  isMobiusWssHostAllowed,
 } from './constants';
 import {Model, WDMDevice, WebexSDK} from '../SDKConnector/types';
 import SDKConnector from '../SDKConnector';
@@ -144,11 +144,7 @@ import {LineError, createLineError} from '../Errors/catalog/LineError';
  */
 function isTrustedMobiusWssHost(wssUri: string): boolean {
   try {
-    const host = new URL(wssUri).hostname;
-
-    return MOBIUS_WSS_ALLOWED_DOMAINS.some(
-      (domain) => host === domain || host.endsWith(`.${domain}`)
-    );
+    return isMobiusWssHostAllowed(new URL(wssUri).hostname);
   } catch {
     return false;
   }
--- a/packages/calling/src/common/constants.ts
+++ b/packages/calling/src/common/constants.ts
@@ -48,6 +48,20 @@ export const SCIM_ENTERPRISE_USER = 'urn:ietf:params:scim:schemas:extension:ente
 export const WEBEX_API_PROD = 'https://webexapis.com';
 export const WEBEX_API_BTS = 'https://integration.webexapis.com';
 export const WEBEX_API_FEDRAMP = 'https://api-usgov.webex.com';
+
+/**
+ * Trusted domain suffixes for Mobius WSS connections (AC-2).
+ *
+ * A bearer token may only be transmitted over a WSS socket whose hostname
+ * ends with one of these domains.  The list covers primary/backup Mobius
+ * infrastructure, the Webex API gateway, and FedRAMP variants.
+ */
+export const MOBIUS_WSS_ALLOWED_DOMAINS: readonly string[] = [
+  'webex.com',
+  'wbx2.com',
+  'webexapis.com',
+  'cisco.com',
+];
 export const WEBEX_API_CONFIG_INT_URL = `${WEBEX_API_BTS}/v1/uc/config`;
 export const WEBEX_API_CONFIG_PROD_URL = `${WEBEX_API_PROD}/v1/uc/config`;
 export const WEBEX_API_CONFIG_FEDRAMP_URL = `${WEBEX_API_FEDRAMP}/v1/uc/config`;
--- a/packages/calling/src/common/constants.ts
+++ b/packages/calling/src/common/constants.ts
@@ -62,6 +62,17 @@ export const MOBIUS_WSS_ALLOWED_DOMAINS: readonly string[] = [
   'webexapis.com',
   'cisco.com',
 ];
+
+/**
+ * Shared predicate for the Mobius WSS host allowlist (AC-2). Used by both
+ * `Socket#open` and `filterMobiusUris` so the two enforcement points cannot
+ * drift apart.
+ */
+export function isMobiusWssHostAllowed(host: string): boolean {
+  return MOBIUS_WSS_ALLOWED_DOMAINS.some(
+    (domain) => host === domain || host.endsWith(`.${domain}`)
+  );
+}
 export const WEBEX_API_CONFIG_INT_URL = `${WEBEX_API_BTS}/v1/uc/config`;
 export const WEBEX_API_CONFIG_PROD_URL = `${WEBEX_API_PROD}/v1/uc/config`;
 export const WEBEX_API_CONFIG_FEDRAMP_URL = `${WEBEX_API_FEDRAMP}/v1/uc/config`;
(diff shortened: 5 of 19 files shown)


mkesavan13 (commented):
## QA Harness — Automated Code Review

**Verdict:** Comment

**Findings:** 2 — 0 blocking · 0 major · 1 minor · 1 nit

### Pull request overview

This PR neutralizes SCIM filter injection by routing user-controllable `externalId`/`userExternalId` values through a new `escapeScimFilterValue` helper, and restricts Mobius WSS connections to an allowlisted set of host suffixes at both the discovery-filter layer (`filterMobiusUris`) and the socket layer (`Socket#open`). The escaping and host-allowlist logic are consistent across their call sites and the added tests exercise the injection, host-drop, and allowlist paths. One thing to verify: the new `transformIgnorePatterns` regex in `jest.config.js` gained a trailing slash that appears to change what it matches (noted inline). Separately, the two sibling SCIM query builders in `Contacts/ContactsClient.ts` still interpolate `id eq "${...}"` without the new helper; those ids are backend-issued so risk is lower, but routing them through `escapeScimFilterValue` would keep the defense uniform.

### Reviewed changes

QA Harness reviewed 18 out of 18 changed files and generated 2 comments.

### Test & CI gaps

- No test covers the unparseable-host branch: `filterMobiusUris` should drop a malformed `wss` entry (e.g. `'wss://'` or `'not a url'`) via `isTrustedMobiusWssHost`'s catch returning false, and `Socket#open` should reject when `new URL(url)` throws and `this.domain` falls back to the raw string. Add cases asserting the malformed entry is filtered out / the open promise rejects.

### Findings not anchored to a diff line

- `packages/calling/src/mobius-socket/socket/socket-base.ts:242` — When `new URL(url)` throws, `this.domain` falls back to the raw `url` string. That is safe here (a bad string won't match the allowlist and the connection is rejected), but the fallback value is then also used in log lines and the reject `reason`. Consider leaving `this.domain` as a sentinel (e.g. `'unknown-domain'`) on parse failure so an unparseable/attacker-supplied string is not echoed verbatim into logs.

_Generated by the quality-assurance harness. This review is advisory._

packages/calling/jest.config.js
```diff
@@ -43,9 +43,20 @@ const jestConfig = {
   },
   coverageDirectory: 'coverage',
   coverageReporters: ['clover', 'json', 'lcov'],
-  transformIgnorePatterns: ['/node_modules/(?!(@webex/internal-media-core)/)'],
+  transformIgnorePatterns: [
+    '/node_modules/(?!(@webex/internal-media-core|@webex/media-helpers|@webex/test-helper-mock-web-socket|@webex/common-timers)/)/',
```
mkesavan13: **minor · correctness**

The negative-lookahead now ends with `...common-timers)/)/` — an extra literal `/` after the lookahead's closing paren, which the previous single-package pattern did not have. Because the lookahead is zero-width, the trailing `/` forces the next character after `/node_modules/` to be another `/`, i.e. it only matches the substring `/node_modules//`, which never occurs in a real module path. The result is that this ignore pattern matches nothing, so Jest will attempt to transform *all* of `node_modules` rather than ignoring everything except the four listed `@webex/*` ESM packages. That is slower and can surface transform errors on unrelated deps. If the intent was selective ignore, drop the trailing slash: `'/node_modules/(?!(@webex/internal-media-core|@webex/media-helpers|@webex/test-helper-mock-web-socket|@webex/common-timers)/)'`.

packages/calling/src/CallingClient/calling/CallerId/index.test.ts
```diff
@@ -256,4 +256,40 @@ describe('CallerId tests', () => {
     expect(callerId['callerInfo'].name).toStrictEqual('Bob Marley');
     expect(callerId['callerInfo'].num).toStrictEqual('5888');
   });
+
+  it('parseRemotePartyInfo escapes SCIM filter injection in externalId', async () => {
```
Kesari3008: There would be already an existing test for resolveCallerId method and since it's just another method invocation, existing testcase should be modified to verify invocation of this new method instead of adding new testcases everytime for every change

packages/calling/src/common/constants.ts
```diff
@@ -48,6 +48,20 @@ export const SCIM_ENTERPRISE_USER = 'urn:ietf:params:scim:schemas:extension:ente
...
 export const WEBEX_API_FEDRAMP = 'https://api-usgov.webex.com';
+
+/**
+ * Trusted domain suffixes for Mobius WSS connections (AC-2).
+ *
+ * A bearer token may only be transmitted over a WSS socket whose hostname
+ * ends with one of these domains.  The list covers primary/backup Mobius
+ * infrastructure, the Webex API gateway, and FedRAMP variants.
+ */
+export const MOBIUS_WSS_ALLOWED_DOMAINS: readonly string[] = [
```
Kesari3008, line 59: We don't need to define this constant. There is already a domain list defined in services plugin which is used to validate allowed domains. Use isAllowedDomainUrl method from services plugin for validation instead of defining new array and new methods

packages/calling/src/common/Utils.ts
```diff
@@ -136,6 +137,23 @@ import {ContactResponse} from '../Contacts/types';
...
+/**
+ * Returns true when the hostname of the given WSS URI ends with one of the
+ * trusted Mobius/Webex domain suffixes.  URIs with unparseable hostnames are
+ * dropped for safety.
+ */
+function isTrustedMobiusWssHost(wssUri: string): boolean {
+  try {
+    const host = new URL(wssUri).hostname;
+
+    return MOBIUS_WSS_ALLOWED_DOMAINS.some(
```
Kesari3008, line 149: Make use of existing method isAllowedDomainUrl. Also check if fedramp domain needs to be imcluded as based on this list, fedramp domain is missing

packages/calling/src/common/Utils.ts
```diff
@@ -157,7 +175,12 @@ export function filterMobiusUris(mobiusServers: MobiusServers, defaultMobiusUrl:
   if (mobiusServers?.primary?.wss) {
     log.info('Adding Primary wss uris', logContext);
     for (const wssUri of mobiusServers.primary.wss) {
-      wssArrayPrimary.push(wssUri);
+      /* Drop WSS URIs whose host is not in the trusted allowlist (AC-2) */
+      if (isTrustedMobiusWssHost(wssUri)) {
+        wssArrayPrimary.push(wssUri);
+      } else {
+        log.warn(`Dropping untrusted primary WSS URI: host not in allowed domains`, logContext);
```
Kesari3008, line 182: @rarajes2 I believe warning alone wont be enough here, if this push doesn't happen arrays can be empty which will lead to registration failures. We need to discuss the error handling here

packages/calling/src/mobius-socket/socket/socket-base.ts
```diff
@@ -260,6 +261,22 @@ export default class Socket extends EventEmitter {
 
       checkRequired(['forceCloseDelay', 'token', 'trackingId', 'logger'], resolvedOptions);
 
+      /* Guard: reject before WebSocket creation if the target host is not trusted (AC-2 / U-03) */
+      const isHostTrusted = MOBIUS_WSS_ALLOWED_DOMAINS.some(
```
Kesari3008, line 265: Here also we can make use of isAllowedDomainUrl

packages/calling/package.json
```diff
@@ -31,7 +31,7 @@
     "prebuild": "rimraf dist",
     "build": "tsc",
     "build:src": "tsc",
-    "compile": "yarn workspaces foreach -tR --from @webex/calling run build:src",
+    "compile": "yarn workspaces foreach -tR --from @webex/legacy-tools run build:src && yarn workspaces foreach -tR --from @webex/calling run build:src",
```
Kesari3008, line 34: This change is unnecessary and doesn't seem to be needed. Explain why this is added and if it has nothing to do wit security vulnerability fixes then remove these changes

packages/calling/jest.config.js
Kesari3008: Same comment here: Why are these changes required and what's the relevance with security vulnerability fix. If no relevance, please remove

packages/calling/test/mocks/empty-stub.js
```diff
@@ -0,0 +1,2 @@
+// Empty stub for side-effect-only webex plugin imports
+module.exports = {};
```
Kesari3008, line 2: Why are these stub files required as part of this PR. Check the relevance wit the vulnerability fix

Kesari3008 (changes requested):

packages/calling/src/common/Utils.ts
rarajes2: I don't know if we can do much here if we only get this list one time and there is no way this can be remediated through another API call.

- We can't throw any error from here
- We can log the untrusted URIs in the warning so that we can tell what caused the registration failure
- Maybe we can emit some event if the list is empty for primary/backup URIs from `attemptRegistrationWithServers` (this we can discuss further)

I am open for better suggestions.

mkesavan13:
Addressed in commit `6a7146e5a4e4`.

Reason: Added Utils.test.ts case covering isTrustedMobiusWssHost's catch branch (malformed wss URIs dropped) and socket.test.ts case covering Socket#open's new URL() throw fallback (domain=raw string, still rejected). RED: temporarily broke each safeguard, tests failed. GREEN: restored, tests pass. Regression: full Utils.test.ts (109) and socket.test.ts (51) suites pass.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175

packages/calling/jest.config.js
mkesavan13: Addressed in commit `6a7146e5a4e4`.

Reason: Removed stray trailing slash in packages/calling/jest.config.js:47 transformIgnorePatterns regex. RED: node regex check showed the buggy pattern matched nothing (no ignore effect). GREEN: fixed regex correctly ignores node_modules except the 4 allowlisted @webex/* packages. Regression: Utils.test.ts (108 tests, exercises the ts-jest transform pipeline) still passes.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3846589697

packages/calling/src/CallingClient/calling/CallerId/index.test.ts
mkesavan13: Addressed in commit `6a7146e5a4e4`.

Reason: Removed standalone injection test; folded its escaping assertions into the existing PA-ID/From/x-broadworks resolveCallerId test case per reviewer's suggestion, avoiding a new test per change. RED: temporarily removed escapeScimFilterValue call, assertion failed. GREEN: restored, passes. Regression: full CallerId/index.test.ts suite (7 tests) passes.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949791955

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: constants.ts:59 defines a new MOBIUS_WSS_ALLOWED_DOMAINS array; services.js:1049 exposes an existing isAllowedDomainUrl(url) backed by the services catalog. Whether that catalog-based list is semantically equivalent for the WSS trust check is a design call for maintainers, not something resolvable by static inspection alone.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949860538

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Same reuse-existing-utility suggestion for filterMobiusUris. The 'fedramp domain missing' claim isn't clearly correct on inspection: 'api-usgov.webex.com' already ends with '.webex.com', which is in MOBIUS_WSS_ALLOWED_DOMAINS — unless the real FedRAMP Mobius WSS host differs. Confirming the actual FedRAMP host set needs maintainer/ops knowledge I lack.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949871955

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: filterMobiusUris only log.warn()s and drops untrusted WSS entries; if all discovery hosts are untrusted, primary/backup wss arrays end up empty with no escalation beyond a log line, risking silent registration failures. Author (rarajes2) explicitly acknowledges this is unresolved and asks for discussion, confirming a human decision on error-handling strategy is needed.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949899550

packages/calling/src/mobius-socket/socket/socket-base.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Same reuse-isAllowedDomainUrl suggestion at socket-base.ts:265. Socket is a generic transport-only class with no import of SDKConnector/webex.internal.services, so using isAllowedDomainUrl would require threading the services plugin into this low-level class — an architectural change needing maintainer sign-off, not a drop-in swap.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949945780

packages/calling/package.json
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: package.json:34 adds/changes a 'compile' script. The package's .sdd/manifest.json already records this exact command as verified 2026-08-05, before this PR's head commit (2026-08-18), suggesting it may predate the PR — but I have no git-diff/history access in this read-only pass to confirm what this PR actually changed, so I can't settle the reviewer's scope concern.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949976249

packages/calling/jest.config.js
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Same scope question for jest.config.js (moduleNameMapper entries for internal-plugin-device/feature routed to empty-stub.js, plus transformIgnorePatterns). mobius-socket/index.ts already side-effect-imports both plugins, so some mocking need is plausible, but confirming whether these entries were introduced by this PR or predate it needs diff/CI context unavailable here.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949977728

packages/calling/test/mocks/empty-stub.js
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: test/mocks/empty-stub.js is a trivial mock referenced by the jest.config.js moduleNameMapper entries flagged above. Its necessity is tied to the same open scope question about whether these test-infra changes are required by this security fix or are unrelated additions — the PR author needs to clarify.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949986484

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: PR author's (rarajes2) own reply continuing the Utils.ts:182 discussion about empty wss arrays after untrusted-host filtering. It proposes options (can't throw, log untrusted URIs, emit an event) and asks for input — an in-progress design discussion, not a completed fix, requiring a human decision before any code change.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3950436579

packages/calling/src/common/constants.ts
Kesari3008: Compare the items added in MOBIUS_WSS_ALLOWED_DOMAIN list, if those items already exist in catalog-based list that means we can use that

packages/calling/src/common/Utils.ts
Kesari3008: If fedramp domains are covered then ignore this comment

packages/calling/package.json
Kesari3008: Remove the change

packages/calling/jest.config.js
Kesari3008: Remove these changes

packages/calling/test/mocks/empty-stub.js
Kesari3008: remove the changes

packages/calling/src/mobius-socket/socket/socket-base.ts
Kesari3008: If directly accessing webex.internal.services is not possible then in common utils folder create new method like a wrapper which will make use of isAllowedDomainUrl method from services layer and use that in socket-base file and other place as well

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Reviewer asks the author to compare the newly hardcoded MOBIUS_WSS_ALLOWED_DOMAINS list against the dynamic service-catalog list to see if the hardcoded list is redundant/can be derived from the catalog instead. Resolving this requires knowledge of the live services catalog contents and a security-architecture call (static allowlist vs. dynamic catalog-derived list) that can't be settled by static code inspection alone.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965369905

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; no code change was required.

Reason: Comment is conditional: 'if fedramp domains are covered then ignore this comment'. constants.ts defines WEBEX_API_FEDRAMP as https://api-usgov.webex.com, whose host ends with '.webex.com', which is already in MOBIUS_WSS_ALLOWED_DOMAINS (constants.ts:60). SECURITY.md (ai-docs) documents this suffix-based allowlist as the intended AC-2 mitigation, so the fedramp case is covered and no further code change is needed.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965372922

packages/calling/package.json
mkesavan13: Addressed in commit `ccaa155f403a`.

Reason: Reverted the 'compile' script in packages/calling/package.json to drop the unrelated '@webex/legacy-tools' build step. Confirmed via repo-wide grep that this script is not invoked by CI or any tooling (CI builds *-tools/webex-core/all packages directly via 'workspaces foreach'), so removing it has no build or test impact.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965380475

packages/calling/jest.config.js
mkesavan13: Addressed in commit `ccaa155f403a`.

Reason: Removed the jest.config.js moduleNameMapper entries for '@webex/internal-plugin-device' and '@webex/internal-plugin-feature' that routed to the deleted stub. Verified safety by building both packages exactly as CI's pre-test build step does, then rerunning the full suite against the real built modules.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965383691

packages/calling/test/mocks/empty-stub.js
mkesavan13: Addressed in commit `ccaa155f403a`.

Reason: Deleted test/mocks/empty-stub.js. Confirmed safe (not just deleted blindly) by building @webex/internal-plugin-device and @webex/internal-plugin-feature the same way CI's pre-test build step does, then rerunning the full calling suite with the real dist output instead of the stub.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965386411

packages/calling/src/mobius-socket/socket/socket-base.ts
mkesavan13: Addressed in commit `ccaa155f403a`.

Reason: Extracted the duplicated Mobius WSS host-allowlist logic into one exported isMobiusWssHostAllowed(host) helper in common/constants.ts. Utils.ts's isTrustedMobiusWssHost and socket-base.ts's Socket#open now both call it, eliminating the risk of the two enforcement points drifting apart.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965395543

packages/calling/src/CallingClient/calling/CallerId/index.test.ts
mkesavan13: Reviewed; no code change was required.

Reason: The existing test 'When PA-ID, From header is present along with x-broad-works' in CallerId/index.test.ts (now lines 114-146) was modified in place to assert the SCIM-escaping behavior of resolveCallerId/resolveCallerIdDisplay, rather than a brand-new test case being appended. This matches the reviewer's request to extend the existing test rather than add a new one each time.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949791955

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: constants.ts:59-75 still defines a standalone MOBIUS_WSS_ALLOWED_DOMAINS list and isMobiusWssHostAllowed() rather than delegating to webex-core's catalog-backed services.isAllowedDomainUrl as requested. Adopting isAllowedDomainUrl would require plumbing a webex/services reference into the Socket layer and verifying the service catalog actually covers all Mobius WSS hosts (incl. FedRAMP) — an architecture/verification call the team hasn't made yet.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949860538

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Utils.ts:145-151 (isTrustedMobiusWssHost) still calls the custom isMobiusWssHostAllowed helper instead of services.isAllowedDomainUrl, and the FedRAMP coverage question raised (whether api-usgov.webex.com is properly covered by the canonical catalog list vs. the ad-hoc one) has not been investigated/resolved in code.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949871955

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: filterMobiusUris (Utils.ts:171-200) only logs a warning when dropping untrusted WSS URIs; no fallback is populated for wssArrayPrimary/backup. CallingClient.connectToMobiusSocket() returns early without throwing if primaryWssMobiusUris ends up empty (confirmed in CallingClient.ts:708 and ai-docs/ARCHITECTURE.md), so the reviewer's concern about silent registration/connection failure is real and still unresolved — no error-handling strategy has been implemented or agreed on.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949899550

packages/calling/src/mobius-socket/socket/socket-base.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: socket-base.ts:265 still guards Socket#open with the custom isMobiusWssHostAllowed() rather than services.isAllowedDomainUrl, same unresolved consolidation request as the constants.ts/Utils.ts comments.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949945780

packages/calling/package.json
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: packages/calling/package.json:23-26 still contains the 'browser' field remapping socket.js/socket.ts to socket.shim.js/socket.shim.ts. It may be legitimately needed to support environment-specific WebSocket constructor injection introduced by this PR, but the reviewer's request to justify or remove it as out-of-scope for a security-only PR was never answered in the diff.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949976249

packages/calling/jest.config.js
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: jest.config.js still contains moduleNameMapper/transformIgnorePatterns entries (now mapping to test/mocks/internal-plugin-metrics-stub.js) that are unrelated to the SCIM/WSS security fix. The stub was renamed and commented for clarity but the reviewer's underlying question of whether this test-infra change belongs in a security-vulnerability PR was never explicitly resolved.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949977728

packages/calling/test/mocks/empty-stub.js
mkesavan13: Reviewed; no code change was required.

Reason: packages/calling/test/mocks/empty-stub.js no longer exists in the current tree; it was removed/replaced by the purpose-named and commented test/mocks/internal-plugin-metrics-stub.js in a follow-up commit, addressing the reviewer's concern about the specific stub file's relevance.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949986484

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: The PR author's own reply in this thread explicitly states the error-handling gap (empty primary/backup URI arrays after untrusted-host filtering) can't be fully remediated with a simple fix and invites further discussion/suggestions ('I am open for better suggestions'). No resolution has been implemented since, confirming this is an open design question rather than a closed item.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3950436579

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: This Sept 9 follow-up softens the earlier ask to 'compare MOBIUS_WSS_ALLOWED_DOMAIN items against the catalog-based list' before deciding to reuse isAllowedDomainUrl. That comparison/investigation has not been done in the code (constants.ts:59 unchanged), so the underlying consolidation decision remains open.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965369905

packages/calling/package.json
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Despite the explicit 'Remove the change' instruction on Sept 9 (before the head commit at 08:24 UTC the same day), the package.json 'browser' field remapping is still present at head. Whether it is required infrastructure for the WSS host-validation fix or out-of-scope scope-creep has not been settled between author and reviewer.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965380475

packages/calling/test/mocks/empty-stub.js
mkesavan13: Reviewed; no code change was required.

Reason: The specific file flagged, packages/calling/test/mocks/empty-stub.js, no longer exists at head — it was removed/replaced by the properly named and documented internal-plugin-metrics-stub.js, satisfying the reviewer's explicit 'remove the changes' request for that artifact.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3965386411

packages/calling/src/CallingClient/calling/CallerId/index.test.ts
mkesavan13: Reviewed; no code change was required.

Reason: Verified in CallerId/index.test.ts:114-146: the injection-escaping assertions (rawFilter vs escapedFilter, requestedUri checks) are folded into the existing PA-ID/From/x-broadworks test case exactly as claimed, and escapeScimFilterValue's effect is still exercised there. Claim of 'addressed' holds.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951049689

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Confirmed constants.ts:59 still defines a hardcoded MOBIUS_WSS_ALLOWED_DOMAINS array while services.js:1049 exposes a separate catalog-backed isAllowedDomainUrl. Whether the WSS trust check should be static or catalog-derived is an unresolved security-architecture choice that static inspection cannot settle; genuinely needs maintainer sign-off.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951049809

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Verified WEBEX_API_FEDRAMP = 'https://api-usgov.webex.com', whose host does end with '.webex.com' (already in the allowlist), so the reviewer's 'fedramp domain missing' claim isn't clearly correct as-is. Whether the real FedRAMP Mobius WSS host differs from the API gateway host requires ops/maintainer knowledge not resolvable by code inspection alone.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050013

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Confirmed in Utils.ts:171-200 that filterMobiusUris only log.warn()s on untrusted WSS hosts and can leave primary/backup wss arrays empty with no escalation. This is a real behavioral gap whose remediation strategy (throw, event, etc.) is an open design question the PR author explicitly asked for input on.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050170

packages/calling/src/mobius-socket/socket/socket-base.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Confirmed socket-base.ts has no import of SDKConnector or webex.internal.services (only @webex/common, @webex/common-timers, lodash, and local constants). Swapping isMobiusWssHostAllowed for isAllowedDomainUrl would require threading the services plugin into a low-level transport class — a genuine architectural change needing maintainer approval.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050300

packages/calling/package.json
mkesavan13: Reviewed; no code change was required.

Reason: This comment (posted 2026-09-07) flagged uncertainty about the package.json 'compile' script's scope. The later commit ccaa155f403a reverted this script to drop the unrelated @webex/legacy-tools step (confirmed in current package.json:34, which now reads 'yarn workspaces foreach -tR --from @webex/calling run build:src' with no legacy-tools reference). The scope concern is resolved by that follow-up fix, making this earlier open question moot.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050407

packages/calling/jest.config.js
mkesavan13: Reviewed; no code change was required.

Reason: Current jest.config.js has no moduleNameMapper entries for internal-plugin-device/internal-plugin-feature and no empty-stub.js reference in transformIgnorePatterns — the disputed test-infra additions this comment questioned are no longer present in the file, so the scope concern is now moot.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050502

packages/calling/test/mocks/empty-stub.js
mkesavan13: Reviewed; no code change was required.

Reason: Verified test/mocks/empty-stub.js no longer exists in the worktree (glob returned no matches), confirming this file was removed and the scope/necessity question this comment raised is resolved.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050618

packages/calling/src/common/Utils.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Same underlying open issue as rgD26 (Utils.ts:171-200, silent-drop behavior with no escalation when all WSS hosts are filtered out). This is the PR author's own in-progress design discussion proposing several remediation options; still unresolved in current code and requires a maintainer decision.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3951050756

packages/calling/src/common/constants.ts
mkesavan13: Reviewed; a human decision is required before changing code.

Reason: Same underlying open issue as rgDxR, re-raised on 2026-09-09 after the PR was updated: whether the hardcoded MOBIUS_WSS_ALLOWED_DOMAINS list (constants.ts:59) is redundant with or should derive from the dynamic services catalog. Still unresolved in current code; requires maintainer knowledge of the live catalog and a security-architecture decision.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3966206947

packages/calling/package.json
mkesavan13: Reviewed; no code change was required.

Reason: Verified current packages/calling/package.json:34 'compile' script reads 'yarn workspaces foreach -tR --from @webex/calling run build:src' with no @webex/legacy-tools step, matching the claimed revert. Claim of 'addressed' holds.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3966207272

packages/calling/test/mocks/empty-stub.js
mkesavan13: Reviewed; no code change was required.

Reason: Verified test/mocks/empty-stub.js is absent from the current worktree, matching the claimed deletion. Claim of 'addressed' holds.

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3966207636

packages/calling/src/common/constants.ts
mkesavan13: @Kesari3008

Reviewed; a human decision is required before changing code.

Reason: Swapping a static allowlist for the catalog-backed isAllowedDomainUrl() is a security-architecture choice (static vs dynamically-discovered trust boundary) requiring maintainer knowledge of live catalog contents, not a drop-in change.

Decision needed: Should the WSS host check use the static MOBIUS_WSS_ALLOWED_DOMAINS list or the dynamic catalog-backed isAllowedDomainUrl(), and is the runtime catalog's domain set equivalent for this trust boundary?

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949860538

packages/calling/src/common/Utils.ts
mkesavan13: @Kesari3008

Reviewed; a human decision is required before changing code.

Reason: The 'fedramp domain missing' claim doesn't hold for the known FedRAMP API host, but confirming whether a distinct FedRAMP Mobius WSS host exists needs information outside the repo.

Decision needed: Is the real production FedRAMP Mobius WSS hostname distinct from '*.webex.com' (already covered by the allowlist), and should the check reuse isAllowedDomainUrl() instead?

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949871955

packages/calling/src/common/Utils.ts
mkesavan13: @Kesari3008

Reviewed; a human decision is required before changing code.

Reason: Real, still-unresolved gap confirmed in current code with no chosen remediation strategy; the PR author's own reply asks for input on options, confirming a human policy decision is needed.

Decision needed: When all primary/backup WSS URIs are dropped as untrusted, should the code throw, emit a client-facing event, or take some other action instead of silently leaving the arrays empty?

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949899550

packages/calling/src/mobius-socket/socket/socket-base.ts
mkesavan13: @Kesari3008

Reviewed; a human decision is required before changing code.

Reason: Socket has no access to the services plugin today; adopting isAllowedDomainUrl() here is an architectural coupling change needing maintainer approval, not a mechanical swap.

Decision needed: Should the low-level Socket transport class be given access to webex.internal.services so it can call isAllowedDomainUrl(), and is that coupling acceptable?

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#discussion_r3949945780

mkesavan13:
Compare the items added in MOBIUS_WSS_ALLOWED_DOMAIN list, if those items already exist in catalog-based list that means we can use that

mkesavan13:
@mkesavan13

Reviewed; a human decision is required before changing code.

Reason: Comment asks if MOBIUS_WSS_ALLOWED_DOMAINS overlaps an existing catalog list for reuse. Found COMMERCIAL_ALLOWED_DOMAINS (webex-core constants.js), unioned into services.allowedDomains for general HTTP auth-header gating, a broader/different threat model. 3 of 4 entries overlap, but that list includes unrelated broadcloud* domains and lacks cisco.com. Reusing it would drop cisco.com or broaden the WSS bearer-token trust boundary. No stated intent or policy resolves this.

Decision needed: Should MOBIUS_WSS_ALLOWED_DOMAINS be replaced/unioned with webex-core's COMMERCIAL_ALLOWED_DOMAINS, accepting it would drop cisco.com or add unrelated broadcloud* domains to hosts trusted to carry the bearer [redacted] over WSS?

Original feedback: https://github.com/webex/webex-js-sdk/pull/5175#issuecomment-5765258739

PR #5179, opened 2026-08-19
feat(contact-center): centralize consult transfer list policy (CAI-8354)

rsarika:
# COMPLETES [CAI-8354](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-8354)

Contact Center consumers need consistent consult and transfer destination eligibility, visibility, and backend ordering through the existing SDK surfaces.

## Changes

- Keep getBuddyAgents, getQueues, and getEntryPoints as the public list methods.
- Apply action-aware buddy-agent defaults while preserving explicit existing options.
- Apply queue eligibility, profile-view, ordering, and cache-safety defaults through getQueues.
- Map profile-scoped dial-number rows through getEntryPoints and expose the dialled number as EntryPointRecord.number.
- Preserve backend response order and pagination metadata.
- Expose ordered Consult and Transfer destination arrays on root Task UI controls for all task types.
- Route supported entry-point transfers through the existing vteam transfer path.
- Update the Contact Center sample, focused unit coverage, and coordinated feature specification.

## Demo

[Watch the Vidcast](https://app.vidcast.io/share/e4cb181c-71d9-4669-bc0a-1cd068ff99f3)

## Impact

Consumers can use the existing list methods and read one root Task control field without recreating reusable destination decisions. Existing explicit list options remain available for consumers that need supported overrides.

## Validation

- Contact Center unit tests: 38 suites and 1,024 tests passed.
- Contact Center build passed.
- Style checks completed with 0 errors; 40 existing warnings remain.
- Diff-scoped specification drift check passed.

## Change Type

- [x] Bug fix
- [x] SDK behavioral defaults
- [x] Documentation update

## GAI Coding Policy

- [x] GAI was used to create a draft that was subsequently customized or modified.


packages/@webex/contact-center/src/services/config/constants.ts
```diff
@@ -258,17 +258,14 @@ export const endPointMap = {
...
    * @returns The endpoint URL string.
-   * @public
-   * @example
-   * const url = endPointMap.entryPointList('org123', 'page=0&pageSize=10');
    * @ignore
    */
-  entryPointList: (orgId: string, queryParams: string) =>
-    `/organization/${orgId}/v2/entry-point?${queryParams}`,
+  entryPointDialNumberList: (orgId: string, queryParams: string) =>
+    `/organization/${orgId}/v3/dial-number?${queryParams}`,
```
rarajes2, line 275: Do we not need the v2 at all ?

packages/@webex/contact-center/src/services/config/Util.ts
```diff
@@ -238,6 +238,9 @@ function parseAgentConfigs(profileData: {
     callVariablesSuppressed: tenantData.callVariablesSuppressed,
     agentDbId: userData.dbId,
     allowConsultToQueue: agentProfileData.consultToQueue,
+    accessQueue: agentProfileData.accessQueue as 'ALL' | 'SPECIFIC' | 'NONE',
+    accessEntryPoint: agentProfileData.accessEntryPoint as 'ALL' | 'SPECIFIC' | 'NONE',
+    accessBuddyTeam: agentProfileData.accessBuddyTeam as 'ALL' | 'SPECIFIC' | 'NONE',
```
rarajes2: Since we have created types for these, we should use them.

packages/@webex/contact-center/src/services/core/WebexRequest.ts
```diff
@@ -34,14 +34,16 @@ class WebexRequest {
...
   }): Promise<IHttpResponse> {
-    const {service, resource, method, body} = options;
+    const {service, resource, method, body, headers} = options;
 
     return this.webex.request({
       service,
       resource,
       method,
       body,
+      ...(headers ? {headers} : {}),
```
rarajes2, line 46: Check once if the spread is working as expected

packages/@webex/contact-center/src/services/task/state-machine/uiControlsComputer.ts
```diff
@@ -97,6 +105,53 @@ export function getDefaultUIControls(): TaskUIControls {
...
+
+  const isVoice = config.channelType === TASK_CHANNEL_TYPE.VOICE;
+  const interaction = taskData?.interaction;
+  const direction = interaction?.contactDirection?.type?.toUpperCase();
+  const destinations: DestinationControls = {
+    consult: [],
+    transfer: [],
+  };
+
+  if (destinationConfig.accessBuddyTeam?.toUpperCase() !== 'NONE') {
```
rarajes2: Instead of using the hardcoded strings here, we should use the created constants. Applicable everywhere

packages/@webex/contact-center/src/services/task/types.ts
```diff
@@ -67,6 +67,23 @@ export const CONSULT_TRANSFER_DESTINATION_TYPE = {
  */
 export type ConsultTransferDestinationType = Enum<typeof CONSULT_TRANSFER_DESTINATION_TYPE>;
 
+/**
+ * Desktop Profile access level for a consult/transfer destination category.
+ * @internal
+ */
+export type ConsultTransferCollaborationAccess = 'ALL' | 'SPECIFIC' | 'NONE';
```
rarajes2: This is repeated in this file as well - packages/@webex/contact-center/src/services/config/types.ts

Please search through the codebase and centralize it

packages/@webex/contact-center/src/services/EntryPoint.ts
```diff
@@ -2,12 +2,46 @@ import {HTTP_METHODS, WebexSDK} from '../types';
...
+  `fields=in=("entryPointName","dialledNumber");value=="${escapeCmsSearchFilterValue(search)}"`;
+
+const mergeEntryPointAttributes = (attributes?: string): string => {
+  const mergedAttributes = new Set(DEFAULT_ENTRY_POINT_ATTRIBUTES.split(','));
+
+  attributes
+    ?.split(',')
+    .map((attribute) => attribute.trim())
+    .filter(Boolean)
+    .forEach((attribute) => mergedAttributes.add(attribute));
```
rarajes2, line 28: Q. How does the `.filter(Boolean)` work ?

packages/@webex/contact-center/src/types.ts
```diff
@@ -570,25 +570,28 @@ export type RequestBody =
...
- * @ignore
  */
 export type BuddyAgents = {
   /**
-   * The media type channel to filter buddy agents.
-   * Determines which channel capability the returned agents must have.
+   * The media type channel to filter buddy agents. Defaults to telephony when omitted.
    */
-  mediaType: 'telephony' | 'chat' | 'social' | 'email';
+  mediaType?: 'telephony' | 'chat' | 'social' | 'email';
```
rarajes2, line 582: Is this not a required field anymore ?

rarajes2 (commented):
Please attach a vidcast to showcase the feature.

packages/@webex/contact-center/src/services/config/constants.ts
rsarika: The v3 dial-number endpoint is intentional for this flow because it supports desktopProfileFilter, returns the entry-point mapping and dialled number, and matches the list required by the consumer. The previous v2 entry-point endpoint has no remaining call site. If we later need raw entry-point entities, that should be handled as a separate use case.

packages/@webex/contact-center/src/services/config/Util.ts
rsarika: Agreed. The collaboration-access type should be defined once and reused by DesktopProfileResponse, Profile, the task destination configuration, and this mapping instead of repeating the literal union. addressed

packages/@webex/contact-center/src/services/core/WebexRequest.ts
rsarika: Confirmed. When headers is provided, the spread adds {headers} to the request; when it is undefined, no headers property is added. The WebexRequest unit test verifies forwarding the header object, and the EntryPoint tests verify the required headers reach webex.request()

packages/@webex/contact-center/src/services/task/state-machine/uiControlsComputer.ts
rsarika: Agreed. I’ll use a shared collaboration-access constant for NONE and the corresponding direction constants instead of spelling these values inline.

packages/@webex/contact-center/src/types.ts
rsarika: Correct—it is intentionally optional now. getBuddyAgents() defaults it to telephony, allowing calls such as getBuddyAgents({action: 'Consult'}). Digital consumers can still pass chat, email, or social explicitly. The default behavior is covered by a unit test.

packages/@webex/contact-center/src/services/EntryPoint.ts
rsarika: After splitting and trimming the attributes, .filter(Boolean) removes empty strings. For example, "id, ,name," becomes ["id", "name"] before the values are added to the Set.

packages/@webex/contact-center/src/services/task/types.ts
rsarika: Agreed. This is the same access domain used by the profile and task configuration. I’ll centralize the constant/type in the configuration types and reuse it from the task state-machine types.

rarajes2 (approved):
LGTM

akulakum (approved):

PR #5205, opened 2026-09-03
feat(meetings): add endSimultaneousInterpretation to interpretation plugin

xinhyao:
# COMPLETES #<https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-858727>

add endSimultaneousInterpretation method in SDK 

- Add `endSimultaneousInterpretation()` to the `SimultaneousInterpretation` plugin. It sends `PATCH {locusUrl}/controls` with `{interpretation: {siEnabled: false}}`, letting the host end Simultaneous Interpretation for the whole meeting. Locus clears SI subscriptions and removes interpreters and their interpreter roles; SI cannot be re-enabled for the rest of the meeting. The Locus DTO in the response is fed back through `MeetingUtil.updateLocusFromApiResponse`, matching the existing `updateInterpreters()` / `changeDirection()` shape.
- Add a readable `siEnabled` prop (sourced from Locus controls) reflecting whether SI is currently on. **This is consumer-facing state**: webex-web-client reads `meeting.simultaneousInterpretation.siEnabled` (refreshed via the existing `MEETING_INTERPRETATION_UPDATE` trigger) as the public signal for "SI is on".
- Update `updateInterpretation()` so `siEnabled` is only written when the incoming payload actually carries it, so the meeting-info path (which passes just `{siLanguages}`) cannot clobber the value maintained by Locus controls.
- Add unit tests covering the new method and the conditional `siEnabled` update.

### Change Type

- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

**Automated**
- `yarn workspace @webex/plugin-meetings test:unit --targets interpretation/index.ts` → all passing, including the new cases for `endSimultaneousInterpretation()` and the conditional `siEnabled` update.
- `eslint` on `src/interpretation/index.ts` → clean.

**Manual (via webex-web-client consuming this build)** 
- Host applies interpreters → `siEnabled` becomes `true` → SI control appears.
- Host ends SI via `endSimultaneousInterpretation()` → Locus broadcasts `siEnabled:false` → `siEnabled` becomes `false` → SI control hides for the host and all participants.
- Meeting-info refresh (e.g. permission-token refresh) while SI is on does **not** flip `siEnabled` off.

  - [ ] Github Copilot
  - [ ] Other - Please Specify
- [ ] This PR is related to
  - [ ] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### I certified that

---

### DIFF
--- a/packages/@webex/plugin-meetings/src/interpretation/index.ts
+++ b/packages/@webex/plugin-meetings/src/interpretation/index.ts
@@ -33,6 +33,7 @@ const SimultaneousInterpretation = WebexPlugin.extend({
     meetingSIEnabled: 'boolean', // appears the meeting support SI feature
     hostSIEnabled: 'boolean', // appears the meeting host/interpreter feature of SI enabled
     selfIsInterpreter: 'boolean', // current user is interpreter or not
+    siEnabled: 'boolean', // appears SI is currently enabled in the meeting (from locus controls)
   },
   derived: {
     shouldQuerySupportLanguages: {
@@ -118,6 +119,11 @@ const SimultaneousInterpretation = WebexPlugin.extend({
    */
   updateInterpretation(interpretation) {
     this.siLanguages.set(interpretation?.siLanguages || []);
+    // Only update siEnabled when it's actually provided. The meeting-info path calls this with
+    // just {siLanguages}, and an unconditional set would wipe the siEnabled carried by locus controls.
+    if (interpretation && 'siEnabled' in interpretation) {
+      this.set('siEnabled', !!interpretation.siEnabled);
+    }
   },
   /**
    * Update self's interpretation information (self is interpreter)
@@ -203,6 +209,37 @@ const SimultaneousInterpretation = WebexPlugin.extend({
       })
       .catch((error) => {
         LoggerProxy.logger.error('Meeting:interpretation#updateInterpreters failed', error);
+        throw error;
+      });
+  },
+  /**
+   * End simultaneous interpretation in the meeting. Only the host is allowed to call this api.
+   * Locus clears SI subscriptions, removes interpreters and their interpreter roles.
+   * It cannot be re-enabled during the current meeting.
+   * @returns {Promise}
+   */
+  endSimultaneousInterpretation() {
+    const meeting = this.webex.meetings.meetingCollection.getByKey('locusUrl', this.locusUrl);
+
+    return this.request({
+      method: HTTP_VERBS.PATCH,
+      uri: `${this.locusUrl}/controls`,
+      body: {
+        interpretation: {
+          siEnabled: false,
+        },
+      },
+    })
+      .then((response) => {
+        MeetingUtil.updateLocusFromApiResponse(meeting, response);
+
+        return response;
+      })
+      .catch((error) => {
+        LoggerProxy.logger.error(
+          'Meeting:interpretation#endSimultaneousInterpretation failed',
+          error
+        );
         throw error;
       });
   },
--- a/packages/@webex/plugin-meetings/test/unit/spec/interpretation/index.ts
+++ b/packages/@webex/plugin-meetings/test/unit/spec/interpretation/index.ts
@@ -159,6 +159,27 @@ describe('plugin-meetings', () => {
         interpretation.updateInterpretation({siLanguages: [{languageName: 'en', languageCode: 1}]});
         checkSILanguage(interpretation.siLanguages.en, {languageName: 'en', languageCode: 1});
       });
+      it('stores siEnabled from the interpretation control', () => {
+        interpretation.updateInterpretation({siEnabled: true, siLanguages: []});
+        assert.equal(interpretation.siEnabled, true);
+
+        interpretation.updateInterpretation({siEnabled: false, siLanguages: []});
+        assert.equal(interpretation.siEnabled, false);
+      });
+      it('leaves siEnabled untouched when the key is not provided', () => {
+        interpretation.updateInterpretation({siEnabled: true, siLanguages: []});
+        assert.equal(interpretation.siEnabled, true);
+
+        // The meeting-info path passes only siLanguages; siEnabled must not be clobbered to false.
+        interpretation.updateInterpretation({siLanguages: [{languageName: 'en', languageCode: 1}]});
+        assert.equal(interpretation.siEnabled, true);
+
+        interpretation.updateInterpretation({});
+        assert.equal(interpretation.siEnabled, true);
+
+        interpretation.updateInterpretation(undefined);
+        assert.equal(interpretation.siEnabled, true);
+      });
     });
 
     describe('#updateSelfInterpretation', () => {
@@ -356,6 +377,47 @@ describe('plugin-meetings', () => {
         assert.calledOnceWithExactly(
           LoggerProxy.logger.error,
           'Meeting:interpretation#updateInterpreters failed',
+          mockError
+        );
+      });
+    });
+
+    describe('#endSimultaneousInterpretation', () => {
+      it('makes the request as expected', async () => {
+        const mockResponse = {body: {locus: {url: 'locusUrl'}}};
+        webex.request.returns(Promise.resolve(mockResponse));
+
+        await interpretation.endSimultaneousInterpretation();
+        assert.calledOnceWithExactly(webex.request, {
+          method: 'PATCH',
+          uri: 'locusUrl/controls',
+          body: {
+            interpretation: {
+              siEnabled: false,
+            },
+          },
+        });
+        assert.calledOnceWithExactly(
+          mockMeeting.locusInfo.handleLocusAPIResponse,
+          mockMeeting,
+          mockResponse.body
+        );
+      });
+
+      it('rejects with error', async () => {
+        const mockError = new Error('something wrong');
+        webex.request.returns(Promise.reject(mockError));
+        LoggerProxy.logger.error = sinon.stub();
+
+        await assert.isRejected(
+          interpretation.endSimultaneousInterpretation(),
+          mockError,
+          'something wrong'
+        );
+
+        assert.calledOnceWithExactly(
+          LoggerProxy.logger.error,
+          'Meeting:interpretation#endSimultaneousInterpretation failed',
           mockError
         );
       });


WeijuanShao (approved):

mickelr (commented):
## Review — blocking items

Thanks for this! The logic is sound and stylistically consistent with `updateInterpreters()` / `changeDirection()`.

**Verified locally:** applied the diff and ran `yarn workspace @webex/plugin-meetings test:unit --targets interpretation/index.ts` → 43 passing (including the 4 new tests); `eslint` on `src/interpretation/index.ts` is clean.

Three blocking items — one process, two design. Inline comments below for the code ones.

---

### 🔴 Blocker 1 — PR description is not filled in

The template placeholders are still present:

- `## by making the following changes` → `< DESCRIBE YOUR CHANGES >`
- `## The following scenarios were tested` → `< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >`

The template states this explicitly, for FedRAMP compliance purposes:

> **IMPORTANT** FAILING TO FILL OUT THIS TEMPLATE WILL RESULT IN REJECTION OF YOUR PULL REQUEST

Additionally, `[x] Code was generated entirely by GAI` is checked — the GAI Coding Policy section requires the Copyright Annotation Best Practices followed to be listed below that checkbox, and the AI tool used to be specified. Please complete both.

Also worth flagging: `[ ] I have updated the documentation accordingly` is unchecked, and this PR adds new public API surface (`meeting.simultaneousInterpretation.endSimultaneousInterpretation()` plus a new readable `siEnabled` prop) with no docs, types, or sample-app coverage. Adding a button to the meetings sample would double as the manual test evidence for the section above.

### 🔴 Blocker 2 — `siEnabled` is written but never read

See inline comment on `src/interpretation/index.ts` L36.

### 🔴 Blocker 3 — "preserve when the key is missing" can strand `siEnabled: true`

See inline comment on `src/interpretation/index.ts` L124.

---

### Non-blocking nits

1. **`siEnabled` has no default** → it is `undefined` until the first Locus control arrives, so `if (siEnabled === false)` and `if (!siEnabled)` diverge for consumers. Consider `siEnabled: {type: 'boolean', default: false}`, plus a test asserting the initial value.
2. **No host-permission guard.** The JSDoc says *"Only the host is allowed to call this api"* but nothing enforces it. The plugin already tracks `canManageInterpreters` (set from `SELF_ROLES.MODERATOR` in `meeting/index.ts:4066`). A fail-fast reject would match `changeDirection()` / `handoffInterpreter()` and save a guaranteed-failing round-trip. (`updateInterpreters()` doesn't guard either, so this is a consistency question, not a regression.)
3. **No `locusUrl` guard** — if `this.locusUrl` is undefined the request goes to `"undefined/controls"`. Same pre-existing gap as `updateInterpreters`, but cheap to add here.
4. **Local state isn't reset after success.** `siLanguages`, `selfIsInterpreter`, `receiveLanguage`, `targetLanguage`, `sourceLanguage` and `hostSIEnabled` all rely entirely on the Locus DTO in the PATCH response flowing back through `handleLocusAPIResponse`. Please confirm the PATCH response actually carries the cleared `controls.interpretation` — the test mocks it as `{body: {locus: {url: 'locusUrl'}}}`, which doesn't exercise that at all. If it doesn't, clients will render stale interpreter UI until the next Locus event.
5. **Missing types.** `src/interpretation/interpretation.types.ts` gains no entry for an `Interpretation` control shape carrying `siEnabled`; `updateInterpretation(interpretation)` stays implicitly `any`.
6. **Test coverage gap:** no case for `meetingCollection.getByKey` returning `undefined`. `MeetingUtil.updateLocusFromApiResponse` handles a null meeting, but that branch is untested for this method.

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -33,6 +33,7 @@ const SimultaneousInterpretation = WebexPlugin.extend({
     meetingSIEnabled: 'boolean', // appears the meeting support SI feature
     hostSIEnabled: 'boolean', // appears the meeting host/interpreter feature of SI enabled
     selfIsInterpreter: 'boolean', // current user is interpreter or not
+    siEnabled: 'boolean', // appears SI is currently enabled in the meeting (from locus controls)
```
mickelr, line 36: **Blocking — `siEnabled` is written but never read.**

Nothing in this diff (or anywhere else in the repo — I grepped; `siEnabled` did not exist before this PR) consumes `this.siEnabled`. It is set in `updateInterpretation()` and that's it. Could you clarify the intent?

**If it's meant as consumer-facing state** (readable as `meeting.simultaneousInterpretation.siEnabled`, refreshed via the existing `MEETING_INTERPRETATION_UPDATE` trigger), then that contract needs to be stated in the PR description and ideally in the types/docs — as it stands a reader has no way to know this prop is the public signal for "SI is on".

**If it's meant as an internal guard**, then it isn't wired up. The JSDoc on the new method says:

> It cannot be re-enabled during the current meeting.

…but nothing enforces that. As written a client can call `endSimultaneousInterpretation()` repeatedly and just collect server errors, and `updateInterpreters()` / `querySupportLanguages()` / `changeDirection()` / `handoffInterpreter()` all remain callable after SI has ended. If the guard is the intent, those should early-reject when `siEnabled === false`.

Either answer is fine — but right now this is dead state, and dead state tends to quietly rot. Please pick one and wire it through (or drop the prop from this PR).

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -118,6 +119,11 @@ const SimultaneousInterpretation = WebexPlugin.extend({
    */
   updateInterpretation(interpretation) {
     this.siLanguages.set(interpretation?.siLanguages || []);
+    // Only update siEnabled when it's actually provided. The meeting-info path calls this with
+    // just {siLanguages}, and an unconditional set would wipe the siEnabled carried by locus controls.
+    if (interpretation && 'siEnabled' in interpretation) {
```
mickelr: **Blocking — this key-presence heuristic can strand `siEnabled: true` permanently.**

The Locus controls path (`meeting/index.ts:3108`) forwards the **raw** `controls.interpretation` object straight into this method:

```ts
this.locusInfo.on(
  LOCUSINFO.EVENTS.CONTROLS_MEETING_INTERPRETATION_UPDATED,
  ({interpretation}) => {
    this.simultaneousInterpretation.updateInterpretation(interpretation);
```

So if Locus ever **omits** `siEnabled` once SI is off — rather than sending an explicit `false` — the flag sticks at `true` for the rest of the meeting with no recovery path. That's a realistic failure mode for Locus control DTOs, especially on deltas, where cleared or unchanged sub-fields are commonly dropped. Note also that `controlsUtils.ts:242` only emits the event when `!isEqual(previous?.interpretation, current?.interpretation)`, so a dropped field still looks like a change but carries no corrective value.

**Please confirm against the Locus contract that `siEnabled` is always present in `controls.interpretation`.** If it isn't guaranteed, gate on the *caller* rather than on key presence — that makes the intent explicit and removes the dependency on server payload shape:

```ts
updateInterpretation(interpretation, {preserveSiEnabled = false} = {}) {
  this.siLanguages.set(interpretation?.siLanguages || []);
  if (!preserveSiEnabled) {
    this.set('siEnabled', !!interpretation?.siEnabled);
  }
}
```

…with `MeetingUtil.parseInterpretationInfo` passing `{preserveSiEnabled: true}`.

---

Minor, on the comment itself — it says the meeting-info path

> calls this with just `{siLanguages}`, and an unconditional set would wipe the siEnabled carried by locus controls

That caller (`meeting/util.ts:1010`) is already guarded by `if (!meeting.simultaneousInterpretation?.siLanguages?.length)`, so it only ever runs *before* any languages exist — i.e. before Locus controls could have set `siEnabled` in the first place. The defensive check is still reasonable, but the comment overstates the risk, and it's pointing at a caller whose object shape is a known constant. That's precisely the case where an explicit parameter reads better than sniffing the payload.

packages/@webex/plugin-meetings/src/interpretation/index.ts
```diff
@@ -206,6 +212,37 @@ const SimultaneousInterpretation = WebexPlugin.extend({
         throw error;
       });
   },
+  /**
+   * End simultaneous interpretation in the meeting. Only the host is allowed to call this api.
+   * Locus clears SI subscriptions, removes interpreters and their interpreter roles.
+   * It cannot be re-enabled during the current meeting.
+   * @returns {Promise}
+   */
+  endSimultaneousInterpretation() {
```
mickelr, line 221: Non-blocking, but grouping the smaller asks for this method here:

1. **No permission guard** despite the JSDoc saying host-only. `this.canManageInterpreters` is already maintained on this plugin (set from `SELF_ROLES.MODERATOR` in `meeting/index.ts:4066`) — a `Promise.reject(new Error('Missing permission to manage interpreters'))` would match the defensive style of `changeDirection()` / `handoffInterpreter()` and avoid a guaranteed-failing round-trip.
2. **No `locusUrl` guard** — if `this.locusUrl` is undefined this PATCHes `"undefined/controls"`.
3. **Idempotency** — tied to the `siEnabled` discussion above: once SI has ended, repeat calls should arguably short-circuit rather than hit Locus again.

The `MeetingUtil.updateLocusFromApiResponse` / logger / rethrow shape matches `updateInterpreters()` exactly — nice and consistent 👍

packages/@webex/plugin-meetings/src/interpretation/index.ts
xinhyao: webex-web-client reads meeting.simultaneousInterpretation.siEnabled (refreshed via the existing MEETING_INTERPRETATION_UPDATE trigger) as the public "SI is on" signal, and mirrors it into Redux (isOn) to show/hide the SI control. Two read sites there:

sanitizeSI.ts: isOn: !!meeting?.simultaneousInterpretation?.siEnabled
si/actions/index.ts (updateSIInterpretation): const newIsSIOnFromSdk = !!meeting?.simultaneousInterpretation?.siEnabled

packages/@webex/plugin-meetings/src/interpretation/index.ts
mickelr: Thanks @xinhyao — that resolves Blocker 2. 👍 Confirming for the record:

- `siEnabled` is **consumer-facing state**, read by webex-web-client in `sanitizeSI.ts` (`isOn: !!meeting?.simultaneousInterpretation?.siEnabled`) and `si/actions/index.ts` (`updateSIInterpretation`).
- Both read sites coerce with `!!`, so my nit about the missing `default: false` is harmless for that consumer — `undefined` → `false`. I'd still add the default plus a JSDoc line saying "public signal, mirrored by clients", so the next person doesn't delete it as dead state. Not blocking.

Please put that explanation in the PR description too — the "who reads this" context is the single most useful thing for a future reader, and right now the description is still the unedited template.

---

**However, your answer makes Blocker 3 materially worse, and I've now reproduced it.**

If webex-web-client uses `siEnabled` to show/hide the SI control, then a stranded `siEnabled: true` means **the SI control stays visible after SI has ended** — the exact state this PR exists to prevent.

I traced the full chain in `next` and verified each link:

**1. `ControlsUtils.parse` is a shallow spread** (`controlsUtils.ts:22`), so `interpretation` passes through raw and untouched — no normalisation, no defaults:

```ts
ControlsUtils.parse = (controls: any) => {
  const parsedControls = {...controls};
  // ...only record / meetingContainer / etc. get normalised
  return parsedControls;
};
```

**2. `locus-info` replaces controls wholesale** — `this.controls = controls` at `locus-info/index.ts:2540`. There is **no deep merge** of a delta's `controls` into the previous one. `onDeltaLocus` → `updateLocusInfo` → `updateControls(locus.controls, locus.self)` passes the delta's controls object straight through, and `current` comes from `parse()` of exactly that object.

So for a delta DTO whose `controls` carries some other control but not `interpretation`, I confirmed with a scratch test against `ControlsUtils.getControls`:

```
--> hasInterpretationChanged = true
--> current.interpretation   = undefined
```

…and for a delta that carries `interpretation` without the `siEnabled` key:

```
--> hasInterpretationChanged = true
--> "siEnabled" in current.interpretation = false
```

**3. `meeting/index.ts:3108` forwards `current.interpretation` verbatim** into `updateInterpretation()`.

**4. Result** — scratch test run against this PR's branch:

```
interpretation.updateInterpretation({siEnabled: true, siLanguages: [{languageName: 'en', languageCode: 1}]});
interpretation.updateInterpretation(undefined);   // the delta above

--> siLanguages.length = 0      // languages WERE cleared
--> siEnabled          = true   // but siEnabled stayed true
```

Both assertions pass. The two lines of `updateInterpretation` now disagree about what a missing payload means:

```ts
this.siLanguages.set(interpretation?.siLanguages || []);          // absent -> RESET
if (interpretation && 'siEnabled' in interpretation) { ... }      // absent -> PRESERVE
```

That asymmetry is the bug. It yields "SI is on, with zero languages" — a state that cannot be reached legitimately and that no consumer is written to handle. Worth noting that **every other control handler in `updateControls` resets on absence** (`current?.x?.enabled !== previous?.x?.enabled`); this PR introduces the only preserve-on-absence control in the file, which is why it deserves explicit justification rather than an inline comment.

**What I'm asking for (one of):**

1. **Confirm with the Locus team, in writing on this PR, that `controls.interpretation.siEnabled` is always present whenever `controls.interpretation` is present, AND that `controls.interpretation` is always present whenever `controls` is present in a delta.** Both halves are required for the current code to be safe. If that holds, keep the code and add a comment citing the contract.
2. **Or** drop the key-presence sniffing and gate on the caller, so behaviour no longer depends on server payload shape:

   ```ts
   updateInterpretation(interpretation, {preserveSiEnabled = false} = {}) {
     this.siLanguages.set(interpretation?.siLanguages || []);
     if (!preserveSiEnabled) {
       this.set('siEnabled', !!interpretation?.siEnabled);
     }
   }
   ```

   with `MeetingUtil.parseInterpretationInfo` passing `{preserveSiEnabled: true}` — that caller is the only one with a known-partial shape, and it's already guarded by `if (!meeting.simultaneousInterpretation?.siLanguages?.length)` so it only runs before Locus controls have landed.

Option 2 is ~4 lines, removes the dependency on an unverified server contract, and keeps both lines of the method agreeing on the same semantics. I'd take it unless the Locus contract is confirmed.

Either way, please add a regression test for the delta case — `updateInterpretation(undefined)` after SI was enabled — asserting `siLanguages` and `siEnabled` end up **consistent** with each other. The existing test `leaves siEnabled untouched when the key is not provided` currently locks in the inconsistent behaviour as intended.

---

Everything else from my first pass stands as non-blocking. CI is green across all 10 checks, and I re-ran the package unit suite locally on this branch — 43 passing.

xinhyao: 1. Tested this: ending SI always sends an explicit siEnabled: false, so the presence check can't strand it at true — and if an unrelated delta ever drops the field while SI is on, keeping the existing value is exactly what we want (an unconditional set would flip a live SI off).

2. The !siLanguages?.length guard doesn't fully cover this either: if siLanguages is empty while SI is on (e.g. a mid-meeting meeting-info refresh), the meeting-info path would set siEnabled: false without the check — so it's a deliberate guard against an unintended state flip.

packages/@webex/plugin-meetings/src/interpretation/index.ts
xinhyao: > Non-blocking, but grouping the smaller asks for this method here:
> 
> 1. **No permission guard** despite the JSDoc saying host-only. `this.canManageInterpreters` is already maintained on this plugin (set from `SELF_ROLES.MODERATOR` in `meeting/index.ts:4066`) — a `Promise.reject(new Error('Missing permission to manage interpreters'))` would match the defensive style of `changeDirection()` / `handoffInterpreter()` and avoid a guaranteed-failing round-trip.
> 2. **No `locusUrl` guard** — if `this.locusUrl` is undefined this PATCHes `"undefined/controls"`.
> 3. **Idempotency** — tied to the `siEnabled` discussion above: once SI has ended, repeat calls should arguably short-circuit rather than hit Locus again.
> 
> The `MeetingUtil.updateLocusFromApiResponse` / logger / rethrow shape matches `updateInterpreters()` exactly — nice and consistent 👍

1.Permission guard — the caller in webex-web-client already gates this so only the host can invoke endSimultaneousInterpretation(), so the host-only contract is enforced at the call site. Adding a plugin-level guard would just duplicate that.

2.locusUrl guard — none of the other methods in this file guard locusUrl either; it's a foundational precondition for any Locus interaction. If locusUrl were undefined, the whole Locus communication premise is already broken, so I don't think that assumption holds in practice.

3.Idempotency — by design, once SI is ended the "End SI" button is hidden and can't be re-enabled for the rest of the meeting (only after ending the meeting and starting again). So there's no path to call endSimultaneousInterpretation() twice within the same meeting.

xinhyao:
> ### Non-blocking nits
> 1. **`siEnabled` has no default** → it is `undefined` until the first Locus control arrives, so `if (siEnabled === false)` and `if (!siEnabled)` diverge(分开) for consumers. Consider `siEnabled: {type: 'boolean', default: false}`, plus a test asserting the initial value.
> 2. **No host-permission guard.** The JSDoc says _"Only the host is allowed to call this api"_ but nothing enforces it. The plugin already tracks `canManageInterpreters` (set from `SELF_ROLES.MODERATOR` in `meeting/index.ts:4066`). A fail-fast reject would match `changeDirection()` / `handoffInterpreter()` and save a guaranteed-failing round-trip. (`updateInterpreters()` doesn't guard either, so this is a consistency(一致性) question, not a regression.)
> 3. **No `locusUrl` guard** — if `this.locusUrl` is undefined the request goes to `"undefined/controls"`. Same pre-existing gap as `updateInterpreters`, but cheap to add here.
> 4. **Local state isn't reset after success.** `siLanguages`, `selfIsInterpreter`, `receiveLanguage`, `targetLanguage`, `sourceLanguage` and `hostSIEnabled` all rely entirely(全部地) on the Locus DTO in the PATCH response flowing back through `handleLocusAPIResponse`. Please confirm the PATCH response actually carries the cleared `controls.interpretation` — the test mocks it as `{body: {locus: {url: 'locusUrl'}}}`, which doesn't exercise that at all. If it doesn't, clients will render stale interpreter UI until the next Locus event.
> 5. **Missing types.** `src/interpretation/interpretation.types.ts` gains no entry for an `Interpretation` control shape carrying `siEnabled`; `updateInterpretation(interpretation)` stays implicitly `any`.
> 6. **Test coverage gap(豁缝):** no case for `meetingCollection.getByKey` returning `undefined`. `MeetingUtil.updateLocusFromApiResponse` handles a null meeting, but that branch is untested for this method.

nit1: All 15 props in this class use the shorthand form without an explicit default, so siEnabled stays consistent with the rest. Consumers already read it defensively (!!meeting?.simultaneousInterpretation?.siEnabled), so undefined-until-set is harmless here. 
nit2 and 3: see the reply above.
nit4: after endSimultaneousInterpretation() succeeds, Locus returns the cleared controls.interpretation on the reflow, which resets the state correctly. So we don't need to clear it optimistically on the client side.
nit5: interpretation here is the raw Locus controls DTO, which carries many fields that drift over time. We only ever touch siLanguages and siEnabled via optional access, so typing the whole payload would add a partial type that's easy to let rot. I'd prefer to keep any and access defensively rather than maintain a type we don't fully own.
nit6: added a test covering the case where meetingCollection.getByKey returns undefined.

packages/@webex/plugin-meetings/src/interpretation/index.ts
xinhyao: > Thanks @xinhyao — that resolves Blocker 2. 👍 Confirming for the record:
> 
> * `siEnabled` is **consumer-facing state**, read by webex-web-client in `sanitizeSI.ts` (`isOn: !!meeting?.simultaneousInterpretation?.siEnabled`) and `si/actions/index.ts` (`updateSIInterpretation`).
> * Both read sites coerce with `!!`, so my nit about the missing `default: false` is harmless for that consumer — `undefined` → `false`. I'd still add the default plus a JSDoc line saying "public signal, mirrored by clients", so the next person doesn't delete it as dead state. Not blocking.
> 
> Please put that explanation in the PR description too — the "who reads this" context is the single most useful thing for a future reader, and right now the description is still the unedited template.
> 
> **However, your answer makes Blocker 3 materially worse, and I've now reproduced it.**
> 
> If webex-web-client uses `siEnabled` to show/hide the SI control, then a stranded `siEnabled: true` means **the SI control stays visible after SI has ended** — the exact state this PR exists to prevent.
> 
> I traced the full chain in `next` and verified each link:
> 
> **1. `ControlsUtils.parse` is a shallow spread** (`controlsUtils.ts:22`), so `interpretation` passes through raw and untouched — no normalisation, no defaults:
> 
> ```ts
> ControlsUtils.parse = (controls: any) => {
>   const parsedControls = {...controls};
>   // ...only record / meetingContainer / etc. get normalised
>   return parsedControls;
> };
> ```
> 
> **2. `locus-info` replaces controls wholesale** — `this.controls = controls` at `locus-info/index.ts:2540`. There is **no deep merge** of a delta's `controls` into the previous one. `onDeltaLocus` → `updateLocusInfo` → `updateControls(locus.controls, locus.self)` passes the delta's controls object straight through, and `current` comes from `parse()` of exactly that object.
> 
> So for a delta DTO whose `controls` carries some other control but not `interpretation`, I confirmed with a scratch test against `ControlsUtils.getControls`:
> 
> ```
> --> hasInterpretationChanged = true
> --> current.interpretation   = undefined
> ```
> 
> …and for a delta that carries `interpretation` without the `siEnabled` key:
> 
> ```
> --> hasInterpretationChanged = true
> --> "siEnabled" in current.interpretation = false
> ```
> 
> **3. `meeting/index.ts:3108` forwards `current.interpretation` verbatim** into `updateInterpretation()`.
> 
> **4. Result** — scratch test run against this PR's branch:
> 
> ```
> interpretation.updateInterpretation({siEnabled: true, siLanguages: [{languageName: 'en', languageCode: 1}]});
> interpretation.updateInterpretation(undefined);   // the delta above
> 
> --> siLanguages.length = 0      // languages WERE cleared
> --> siEnabled          = true   // but siEnabled stayed true
> ```
> 
> Both assertions pass. The two lines of `updateInterpretation` now disagree about what a missing payload means:
> 
> ```ts
> this.siLanguages.set(interpretation?.siLanguages || []);          // absent -> RESET
> if (interpretation && 'siEnabled' in interpretation) { ... }      // absent -> PRESERVE
> ```
> 
> That asymmetry is the bug. It yields "SI is on, with zero languages" — a state that cannot be reached legitimately and that no consumer is written to handle. Worth noting that **every other control handler in `updateControls` resets on absence** (`current?.x?.enabled !== previous?.x?.enabled`); this PR introduces the only preserve-on-absence control in the file, which is why it deserves explicit justification rather than an inline comment.
> 
> **What I'm asking for (one of):**
> 
> 1. **Confirm with the Locus team, in writing on this PR, that `controls.interpretation.siEnabled` is always present whenever `controls.interpretation` is present, AND that `controls.interpretation` is always present whenever `controls` is present in a delta.** Both halves are required for the current code to be safe. If that holds, keep the code and add a comment citing the contract.
> 2. **Or** drop the key-presence sniffing and gate on the caller, so behaviour no longer depends on server payload shape:
>    ```ts
>    updateInterpretation(interpretation, {preserveSiEnabled = false} = {}) {
>      this.siLanguages.set(interpretation?.siLanguages || []);
>      if (!preserveSiEnabled) {
>        this.set('siEnabled', !!interpretation?.siEnabled);
>      }
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    with `MeetingUtil.parseInterpretationInfo` passing `{preserveSiEnabled: true}` — that caller is the only one with a known-partial shape, and it's already guarded by `if (!meeting.simultaneousInterpretation?.siLanguages?.length)` so it only runs before Locus controls have landed.
> 
> Option 2 is ~4 lines, removes the dependency on an unverified server contract, and keeps both lines of the method agreeing on the same semantics. I'd take it unless the Locus contract is confirmed.
> 
> Either way, please add a regression test for the delta case — `updateInterpretation(undefined)` after SI was enabled — asserting `siLanguages` and `siEnabled` end up **consistent** with each other. The existing test `leaves siEnabled untouched when the key is not provided` currently locks in the inconsistent behaviour as intended.
> 
> Everything else from my first pass stands as non-blocking. CI is green across all 10 checks, and I re-ran the package unit suite locally on this branch — 43 passing.

fixed.

mickelr (approved):
## Re-review of `f328c15` — Blocker 3 resolved ✅

Thanks for taking Option 2, @xinhyao — and for adding both the regression test and the "meeting not found" case. This addresses everything blocking.

### Verified the fix independently

`updateInterpretation` now has both lines agreeing on the same semantics:

```ts
updateInterpretation(interpretation, {preserveSiEnabled = false} = {}) {
  this.siLanguages.set(interpretation?.siLanguages || []);
  if (!preserveSiEnabled) {
    this.set('siEnabled', !!interpretation?.siEnabled);
  }
}
```

I re-ran my original repro against this head, driving it through the real `ControlsUtils.getControls` chain rather than calling the plugin directly:

```
END-TO-END: delta without interpretation no longer strands siEnabled
  --> siLanguages.length = 0
  --> siEnabled          = false        (was: true — the stranded state)

END-TO-END: delta with interpretation but no siEnabled key also resets
  --> siEnabled = false

meeting-info path still preserves siEnabled
  --> siEnabled = true, siLanguages.length = 1
```

All four scenarios behave correctly. The "SI on with zero languages" state is no longer reachable, and the SI control can no longer get stuck visible in webex-web-client after SI ends.

Also confirmed:

- **Both call sites handled correctly** — `meeting/util.ts:1012` passes `{preserveSiEnabled: true}` (the only known-partial payload), `meeting/index.ts:3108` passes nothing so the Locus controls path resets. Those are the only two callers in `src`.
- **Behaviour now matches every other control handler** in `updateControls` — reset-on-absence. No more preserve-on-absence outlier in the file.
- **The risk no longer depends on an unverified Locus contract**, which was the main thing I wanted. Worth noting for the record: since `siLanguages` has *always* reset on absence, if Locus did routinely send partial `controls`, SI languages would already be visibly breaking in production today — decent empirical evidence the reset path is the safe default.
- `yarn workspace @webex/plugin-meetings test:unit --targets interpretation/index.ts` → **45 passing**; `meeting/utils.js` → **179 passing**.
- `eslint` on both changed src files → clean (the one `require-jsdoc` warning on `util.ts:998` is the pre-existing `renameKey` helper — I confirmed it reports identically on unmodified `next`).
- CI green across all 10 checks on `f328c15`.

Nice touch replacing `leaves siEnabled untouched when the key is not provided` with `resets siEnabled and siLanguages consistently for a delta without interpretation` — the old test was locking in the buggy behaviour as intended, so swapping it out was exactly right.

**LGTM.** 👍

---

### Two loose ends before merge (both quick, neither blocking the code)

**1. The PR description now describes the old implementation.** This bullet is stale:

> Update `updateInterpretation()` so `siEnabled` is **only written when the incoming payload actually carries it**, so the meeting-info path (which passes just `{siLanguages}`) cannot clobber the value maintained by Locus controls.

That's the key-presence version that was replaced. The current behaviour is the opposite by default — Locus controls always write `siEnabled`, and only the meeting-info caller opts out via `{preserveSiEnabled: true}`. Since this is the one bullet explaining the mechanism, it's worth correcting so it doesn't mislead the next reader. Something like:

> Update `updateInterpretation()` to accept a `{preserveSiEnabled}` option. By default `siEnabled` is reset from the payload, keeping it consistent with `siLanguages`; the meeting-info path (which carries only `{siLanguages}`) opts out so it cannot clobber the value owned by Locus controls.

The "Automated" test bullet has the same stale wording ("the conditional `siEnabled` update").

**2. The GAI section is still incomplete.** `[x] Code was generated entirely by GAI` is checked, but the template marks the annotation as mandatory:

> **MANDATORY** If Yes, Mention the GAI Coding Policy Copyright Annotation Best Practices followed separated by a comma below the yes checkbox

The best-practices list, the `Tool used for AI assistance` checkbox, and `This PR is related to` are all still unfilled. Worth completing given this is the FedRAMP-motivated part of the template.

---

### Remaining non-blocking nits (fine as follow-ups)

Carried over from my first pass, none of which gate this PR:

1. **`siEnabled` has no default** — still `undefined` until the first Locus control lands (I confirmed: `initial siEnabled = undefined`). Harmless since both webex-web-client read sites coerce with `!!`, but `{type: 'boolean', default: false}` would make the prop self-documenting.
2. **No host-permission guard** on `endSimultaneousInterpretation()` despite the host-only JSDoc — `canManageInterpreters` is already available.
3. **No `locusUrl` guard** — would PATCH `"undefined/controls"`.
4. **Idempotency** — repeat calls after SI ended still hit Locus.
5. **No types** for the interpretation control shape in `interpretation.types.ts`.

Happy for these to land separately, or not at all — your call.

